PR #4264, opened 2025-04-25
Update NodeJS to v20

diegocurbelo:
Fixes #<github_issue_id>

This PR updates NodeJS from version v16 to v20.

- [x] Updates the dependencies
- [x] Updates the dev dependencies
- [x] Migrate Webpack configuration from v4 to v5
- [x] Fix jest tests
- [x] Fix build warnings (SASS/SCSS deprecation warnings)
- [x] Fix linting rules

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


.npmrc
```diff
@@ -1,2 +1,5 @@
-legacy-peer-deps=true
-engine-strict=true
+save-exact = true
```
diegocurbelo, line 1: Config copied from Guteberg

.prettierrc.js
```diff
@@ -1,3 +1,11 @@
...
-module.exports = require( '@wordpress/prettier-config' );
+module.exports = {
+	...require( '@wordpress/prettier-config' ),
+	overrides: [
+		{
+			files: [ 'changelog.txt' ],
+			options: { parser: 'markdown' },
+		},
+	],
+};
```
diegocurbelo: Config copied from Guteberg

bin/update-pot-file-references.php
```diff
@@ -89,7 +89,7 @@ function load_js_transpiling_source_maps(): array {
 		}
 
 		foreach ( $file_json[ 'sources' ] as $source ) {
-			$source = preg_replace( '%^webpack:///\./(client/.*)$%', '${1}', $source );
+			$source = preg_replace( '%^webpack://\./(client/.*)$%', '${1}', $source );
```
diegocurbelo: In the Webpack 5 config, we are manually setting `devtoolModuleFilenameTemplate`, so we can replace the previous triple slash prefixes with normal URL-like paths `webpack://[resource-path]`

.eslintrc.js
```diff
@@ -62,6 +62,12 @@ module.exports = {
 		'testing-library/no-unnecessary-act': 'off',
 		'@typescript-eslint/no-empty-function': 'off',
 		'@typescript-eslint/no-var-requires': 'off',
+		'@wordpress/i18n-text-domain': [
+			'error',
+			{
+				allowedTextDomain: 'woocommerce-gateway-stripe',
+			},
+		],
```
daledupreez: `@wordpress/i18n-text-domain` was reporting a _lot_ of issues. I chose to specify our expected text domain, but we could also keep to the defaults.

.eslintrc.js
```diff
@@ -82,4 +88,5 @@ module.exports = {
 			'@wordpress/data',
 		],
 	},
+	ignorePatterns: [ 'phpunit-html/**' ],
```
daledupreez: I tripped over a local folder left over from testing and realised it should always be ignored for linting.

daledupreez:
## Progress

### Linting

Linting is now working, but there are still 13 errors stemming from the following:
* 6 errors: React 18 deprecated `ReactDOM.render()` in favour of `createRoot()` ([docs link](https://reactjs.org/link/switch-to-createroot)). If someone wants to tackle that, feel free. 😁 
* 5 errors: `client/entrypoints/express-checkout/index.js` is failing a `@typescript-eslint/no-use-before-define` rule because it looks like there are circular dependencies between some functions.
* 2 errors: `client/data/payment-gateway/hooks.js` and `client/data/settings/hooks.js` both claim to have unnecessary dependencies in the some hooks.

### Javascript Unit Tests

Many JS tests are not even running, mostly due to one of the following three errors:
```
TypeError: Cannot read properties of undefined (reading 'zone')
TypeError: Cannot read properties of undefined (reading 'add')
TypeError: (0 , _data.createSelector) is not a function
```
For the first two errors, the root cause seems to be an issue relating to the interplay between `moment` and `moment-timezone` from within the `@wordpress/date` package, where we have multiple versions of the package in use across the dependency tree. 😐 

The final error stems from an import of `createSelector` from `@wordpress/data`. Not sure how that's getting borked! 🙃 

There are then a handful of what look like legitimate test failures. At least some are related to code trying to test async or delayed loading.

### SASS warnings

It looks like we would be able to avoid some of these by updating `@wordpress/scripts`, but we may simply need to disable/silence the warnings for now. I'll explore whether we can update that tomorrow (unless someone else gets to it first).

daledupreez:
As of dea439c02ef3b4573a955d9f44cb0610c6763e9a, I've now addressed all the remaining linting issues, so the linter is clean.

daledupreez:
SASS warnings are addressed by c1dbefebdee447d1ccfe1df2614f6a4dce4e9d5e and a337c11323beea2f19444e93a42136db575611db.

That required two main areas of work:
 * Switch our actual SCSS files to use `@use` and namespaced imports instead of `@import`
 * Switch our root/abstract SCSS files to use `@forward` to include baseline variables, and ensures that we have deduped values - duplicate forwarded variables and mixins trigger errors, e.g. `$black` is exported from `base-styles/_colors.scss` and `base-styles/_colors.native.scss`.

.npmrc
daledupreez, line 1: While working on various dependencies, I noticed that this results in the default `^` prefix for compatible versions being removed. I am not sure how I feel about that, but I wanted to flag it as a key side-effect.

.npmrc
```diff
@@ -1,2 +1,5 @@
...
+engine-strict = true
+legacy-peer-deps = true
```
daledupreez, line 3: Do we still need this? Maybe something for a follow-up PR.

daledupreez (commented):
Also, before I forget to note this, we almost certainly need to look at extracting some of these changes into smaller PRs that we can test and merge independently. Based on the work I did, I can think of a few areas that might qualify for this:
 * Switch from `@import` to `@use` in the SCSS files
 * Re-order some SCSS that was split by nested rules
 * Re-order some JS files to ensure that code is defined before it is used
 * Moving some dependencies to devDependencies (though they would still need to get their version bumped separately)
 * Remove some unnecessary React hook deps
 * There are a _lot_ of linting changes that might be ignored/accepted by the older linter - if we can extract some of these out, this patch will be _much_ more manageable

laurendavissmith:
Pushed up my changes which don't resolve the jest issue but nudge things in the right direction. Individual tests are still handling too much of their own mocking which is causing a lot of headaches. The mocks setup needs to be rethought entirely and shift as much as possible to handling with global mocks and overrides in the test files only when absolutely necessary.

I also experimented with a `jest-early-mocks.js` file which can be configured in the the `setupFiles` part of the jest config. I didn't end up needing it but if you run into race conditions that can be a useful way out.

client/api/blocks.js
```diff
@@ -109,3 +97,15 @@ const getRequiredFieldDataFromCheckoutForm = ( data ) => {
 
 	return data;
 };
+
+export const createOrder = ( sourceEvent, paymentRequestType ) => {
+	let data = normalizeOrderData( sourceEvent, paymentRequestType );
+	data = getRequiredFieldDataFromCheckoutForm( data );
+
+	return $.ajax( {
```
laurendavissmith: I haven't looked to see how this gets utilized but I don't love that we return a promise with no default error state. It forces every instance of createOrder to implement an error catch since there is no fallback.

wjrosa:
I took the freedom to rebase the PR again with the `develop` branch. I hope it helps to keep tracking the required changes.

wjrosa:
I have fixed all the unit tests here as well. I will move the fixes to a separate PR. Edit: PR is here https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4607

wjrosa:
2 required fixes for the unit tests are out of [that PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4607):
- `registerStore` is now `register` (`@wordpress/data` package). This requires the new Node version to work
- Removal of the `@wordpress/notices` package. We can safely remove as part of legacy code removal work

wjrosa:
After the latest updates in the `develop` branch, I started to have issues when building this branch:
Screenshot 2025-09-01 at 12 59 20

Following the URL mentioned in the error (https://github.com/postcss/postcss/wiki/PostCSS-8-for-end-users), I have included the `postcss` packages as dev dependencies. That fixed the issue for me.

wjrosa:
@daledupreez I spent some free time today reverting all the unnecessary changes here. It looks much smaller than before.

daledupreez:
Thanks for the cleanup, @wjrosa! That is definitely much smaller. That said, I am still wondering whether it may make sense to create a new branch off `develop` and then start by bumping the `node` and `npm` versions only. There are enough other version bumps that may or may not be needed for us to upgrade node.

wjrosa:
@daledupreez I think it is worth it, yes. And it is better than using this PR, as we would start fresh. If you do decide to create a new PR, feel free to request my review 👍

wjrosa:
@daledupreez I took the freedom to create the new PR I suggested here https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4670

PR #4265, opened 2025-04-25
Fix gateway name on non payments pages

### LINKED ISSUE #4204: Replacing title on payment settings page affects other settings pages
**Describe the bug**
The stripe class replaces the gateway title with `# payment method(s)` on the payment settings page. 

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/ed221de4cb17d862a5057d8c38a54066a4bb6241/includes/class-wc-gateway-stripe.php#L143-L154

However the condition is too broad and replaces the title on all settings pages rather than just the payment listings page, `/wp-admin/admin.php?page=wc-settings&tab=checkout`.

Therefore other extensions attempting to display a list of gateways display `# payment method(s)` rather than `Stripe`.

**Recommendation**

In the code block above, include the condition `$GET['tab'] === 'checkout'` (with an `isset`).

**To Reproduce**
1. Install Deposits
2. Enable Stripe (Credit cards and other gateways)
3. Go to the Deposits settings screen, `/wp-admin/admin.php?page=wc-settings&tab=products&section=deposits`
4. Open the list of options in the setting `Disable Payment Gateways`
5. Observe Stripe displays as `# payment method(s)`
6. Observe there is no apparent setting for Stripe Credit Cards

**Expected behavior**

The Stripe gateway should be titled as such on other settings screens.

**Screenshots**

Image

 - WordPress Version: 6.9-alpha-60122
 - WooCommerce Version: 9.7.1
 - Stripe Plugin Version: trunk @ 7f0e8783c4fbe164b382d9f7f6f62045cc24a428
 - Browser [e.g. chrome, safari] and Version: N/A
 - Any other plugins installed
    - Deposits (trunk)
    - Query Monitor
    - User Switching
    - Various MU helper plugins for development (speed up cron jobs, enable live site by default, avoid order number conficts, etc)

**Additional context**

malithsen:
Fixes STRIPE-394
Fixes #4204

This PR adds an extra check to the gateway name generation logic so that we only show the payment method count when in the Settings -> Payments page. All other pages should see "Stripe" as the gateway name.

## Testing instructions

1. Install Deposits.
2. Enable Stripe (Credit cards and other gateways).
3. Go to the Deposits settings screen, `/wp-admin/admin.php?page=wc-settings&tab=products&section=deposits`.
4. Open the list of options in the setting `Disable Payment Gateways`.
5. Observe that the gateway name shows up as Stripe instead of `# payment method(s)`.
6. Go to WooCommerce -> Settings -> Payments page.
7. Stripe gateway should still show up as "Stripe - # payment method(s). 

Before
CleanShot 2025-04-25 at 11 15 32@2x

After
CleanShot 2025-04-25 at 11 16 27@2x

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -17,6 +17,7 @@
 * Dev - Improve SPE e2e tests to reduce flakiness
 * Fix - Prevents fatal errors for cases where we fail to load product details
 * Fix - Address an edge case with webhook URL comparisons
+* Fix - Show correct gateway name in non payments settings pages.
 
 = 9.4.1 - 2025-04-17 =
 * Dev - Forces rollback of version 9.4.0.
--- a/includes/class-wc-gateway-stripe.php
+++ b/includes/class-wc-gateway-stripe.php
@@ -145,7 +145,7 @@ class WC_Gateway_Stripe extends WC_Stripe_Payment_Gateway {
 	 */
 	public function get_title() {
 		// Change the title on the payment methods settings page to include the number of enabled payment methods.
-		if ( ! WC_Stripe_Feature_Flags::is_upe_checkout_enabled() && isset( $_GET['page'] ) && 'wc-settings' === $_GET['page'] ) {
+		if ( ! WC_Stripe_Feature_Flags::is_upe_checkout_enabled() && isset( $_GET['page'] ) && 'wc-settings' === $_GET['page'] && isset( $_GET['tab'] ) && 'checkout' === $_GET['tab'] ) {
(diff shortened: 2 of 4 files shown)


daledupreez (approved):
I think this is fairly safe, but I am only seeing the original issue appear when Legacy checkout is enabled. With the new checkout experience enabled, I don't see a dedicated "Stripe" entry. It may be worth specifying that for the testing instructions.

But I think the restrictions apply in either case.

malithsen:
I see the issue on both Legacy and new checkout. In any case, the check added here should be safe and applies to both.

PR #4268, opened 2025-04-25
Fix subscription failure with Sepa Debit

Mayisha:
In `develop` branch, subscription purchase with Sepa debit fails with a fatal error.

```
Uncaught Error: Call to a member function create_payment_token_for_user() on null in /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php:2689
```


Mayisha:
Confirming that the fatal error has been fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4280.
Closing this PR.

PR #4289, opened 2025-05-02
Determine WordPress and WooCommerce plugin versions dynamically via GitHub Actions

daledupreez:

While working on some tests, I noticed that our unit test matrix for WordPress is currently missing 6.8 (and had 6.6 duplicated due to WooCommerce <-> WordPress compatibility), and we also haven't picked up the most recent minor release of WooCommerce. I then spent some time exploring whether we could automate building the most recent versions using GitHub Actions, and put together the actions in this PR to get the _n_ most recent releases of WordPress and any specified plugin, and then wired them into our PHP unit test testing matrix.

As a side effect, we've added WordPress `6.8.1` instead of `6.6`, and we picked up the recent WooCommerce patch release.

It's possible that both actions will be useful to more folks in the WordPress community, so I plan to kick off a discussion about publishing these actions for broader internal use.

### Implementation notes

The approach for both actions is fairly similar:
 * Make an API call to `api.wordpress.org` to get (WordPress releases | plugin details) in JSON format
 * Use `jq` to filter that down to a list of releases, and filter out suffixed releases, e.g. `-beta`, `-rc1` etc
 * Use `jq` to group by `major.minor` version, and then pick only the most recent release for each `major.minor` version, e.g. we group WordPress `6.7.2`, `6.7.1`, and `6.7.0`, but we then filter down to `6.7.2`
 * Output the requested number of versions as both JSON and newline-separated text

The one minor difference (for now) is that the WordPress list uses an hourly file cache. That may or may not be something that makes sense in the long run.

## Testing instructions

The new actions and their wiring are already in place!

* Check the `PHP tests / get-wordpress-versions` log in the tests below, and verify that the logs show us fetching the most recent WordPress versions
* Check the `PHP tests / get-woocommerce-versions` log in the tests below, and verify that the logs show us fetching the  set of recent WooCommerce releases
* Check the various `PHP tests / Stable *` logs to verify that we're running with the right versions of 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


annemirasol (approved):
This looks good! 🚢 

If we plan to publish broadly, we should probably sanitize `plugin-slug`.

PR #4292, opened 2025-05-05
Fix invalid IP address during mandate data creation

### LINKED ISSUE #3543: Invalid IP address with the new checkout
**Describe the bug**<br>With the new Stripe checkout, users are encountering an "Invalid IP address" error. The specific error message is: "Payment failed: Invalid IP address. Order status changed from Pending payment to Failed." This issue does not occur with the legacy checkout

**To Reproduce**<br>This issue is affecting many users, and a [thread](<https://wordpress.org/support/topic/no-ip-address-supplied/#post-18093119>) has already been started regarding it. However, no one has been able to find a solution yet

Adding another [topic](<https://wordpress.org/support/topic/payment-failed-invalid-ip-address-with-the-new-stripe-checkout/#new-topic-0>)

**Expected behavior**<br>With the new checkout, this error should not come.

* WordPress Version - 6.2.2
* WooCommerce Version - 8.8.5
* Stripe Plugin Version - 8.7.0
* On every browser

annemirasol:
Fixes STRIPE-233
Fixes #3543 

While waiting for [the fix](https://github.com/woocommerce/woocommerce/pull/57284) in core to land (WooCommerce 9.9.0), we will handle cases where `WC_Geolocation::get_ip_address();` returns multiple, comma-separated IP addresses.

## Testing instructions
1. Enable "Log error messages" in WooCommerce > Payments > Stripe > Settings.
2. Smoke test mandate data creation by purchasing any item using [any of the payment methods listed here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/0a9c6b9e2757b2f57689e0e57bdfb021464a18cf/includes/class-wc-stripe-intent-controller.php#L1059). For example, ACH Direct Debit.
3. The order should be successful.
4. Check your PHP logs for errors.
5. Check WooCommerce > Status > Logs > woocommerce-gateway-stripe for errors.
 
---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -1600,11 +1600,11 @@ class WC_Stripe_Helper {
 		$url_parts         = wp_parse_url( $url );
 		$webhook_url_parts = wp_parse_url( $webhook_url );
 
-		$url_host     = $url_parts['host'] ?? '';
-		$url_path     = $url_parts['path'] ?? '';
-		$url_query    = $url_parts['query'] ?? '';
-		$webhook_host = $webhook_url_parts['host'] ?? '';
-		$webhook_path = $webhook_url_parts['path'] ?? '';
+		$url_host      = $url_parts['host'] ?? '';
+		$url_path      = $url_parts['path'] ?? '';
+		$url_query     = $url_parts['query'] ?? '';
+		$webhook_host  = $webhook_url_parts['host'] ?? '';
+		$webhook_path  = $webhook_url_parts['path'] ?? '';
 		$webhook_query = $webhook_url_parts['query'] ?? '';
 
 		if ( $url_host !== $webhook_host || $url_path !== $webhook_path ) {
@@ -1624,7 +1624,7 @@ class WC_Stripe_Helper {
 			return false;
 		}
 
-		$url_query_parts = [];
+		$url_query_parts     = [];
 		$webhook_query_parts = [];
 
 		parse_str( $url_query, $url_query_parts );
@@ -1727,6 +1727,11 @@ class WC_Stripe_Helper {
 	 */
 	public static function add_mandate_data( $request ) {
 		$ip_address = WC_Geolocation::get_ip_address();
+
+		// Handle cases where WC_Geolocation::get_ip_address() returns multiple, comma-separated IP addresses.
+		// TODO: Remove this block when WooCommerce 9.9.0 is released.
+		$ip_address = trim( current( preg_split( '/,/', sanitize_text_field( wp_unslash( $ip_address ) ) ) ) );
+
 		self::maybe_log_ip_issues( $ip_address );
 
 		$request['mandate_data'] = [
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -1729,8 +1729,11 @@ class WC_Stripe_Helper {
 		$ip_address = WC_Geolocation::get_ip_address();
 
 		// Handle cases where WC_Geolocation::get_ip_address() returns multiple, comma-separated IP addresses.
+		// This will be addressed upstream in WooCommerce 9.9.0 as of (https://github.com/woocommerce/woocommerce/pull/57284).
 		// TODO: Remove this block when WooCommerce 9.9.0 is released.
-		$ip_address = trim( current( preg_split( '/,/', sanitize_text_field( wp_unslash( $ip_address ) ) ) ) );
+		if ( str_contains( $ip_address, ',' ) ) {
+			$ip_address = trim( current( preg_split( '/,/', $ip_address ) ) );
+		}
 
 		self::maybe_log_ip_issues( $ip_address );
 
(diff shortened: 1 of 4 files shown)


tests/phpunit/test-wc-stripe-helper.php
```diff
@@ -394,9 +394,9 @@ public function provide_is_wallet_payment_method(): array {
 	 * @return void
 	 */
 	public function test_handle_main_stripe_settings() {
-		WC_Stripe_Helper::update_main_stripe_settings( [ 'test' => 'test' ] );
+		WC_Stripe_Helper::update_main_stripe_settings( [ 'test' => 'abc' ] );
 		$current_settings = WC_Stripe_Helper::get_stripe_settings();
-		$this->assertSame( [ 'test' => 'test' ], $current_settings );
+		$this->assertSame( $current_settings['test'], 'abc' );
```
annemirasol, line 399: I updated this to take into account hooks on `pre_update_option_woocommerce_stripe_settings` that may inject extra settings.

includes/class-wc-stripe-helper.php
```diff
@@ -1727,6 +1727,11 @@ public static function get_klarna_preferred_locale( $store_locale, $billing_coun
 	 */
 	public static function add_mandate_data( $request ) {
 		$ip_address = WC_Geolocation::get_ip_address();
+
+		// Handle cases where WC_Geolocation::get_ip_address() returns multiple, comma-separated IP addresses.
+		// TODO: Remove this block when WooCommerce 9.9.0 is released.
+		$ip_address = trim( current( preg_split( '/,/', sanitize_text_field( wp_unslash( $ip_address ) ) ) ) );
```
daledupreez: I have two thoughts on this approach:
1. We should link to the upstream PR in the comments so we can easily pick up the thread and context later
2. Given that we're not expecting this to happen, it feels like we're going to do a lot of work here for every call. Is it worth adding a simple gate that checks for the `,` character before triggering all the cleanup code?
3. I don't think we need to unslash or sanitize the value in our code, as it should have been sanitized already in `get_ip_address()`. AFAICT, we're cleaning up the values in that function, so we shouldn't need to re-process the values in this code.
```suggestion
		// Handle cases where WC_Geolocation::get_ip_address() returns multiple, comma-separated IP addresses.
		// This will be addressed upstream in WooCommerce 9.9.0 as of (https://github.com/woocommerce/woocommerce/pull/57284).
		// TODO: Remove this block when WooCommerce 9.9.0 is released.
		if ( str_contains( $ip_address, ',' ) ) {
			$ip_address = trim( current( preg_split( '/,/', $ip_address ) ) );
		}
```

annemirasol: @daledupreez Good call, committed your suggestion.

tests/phpunit/test-wc-stripe-helper.php
```diff
@@ -507,4 +507,42 @@ public function test_add_stripe_methods_in_woocommerce_gateway_order() {
...
+	 * @return void
+	 */
+	public function test_add_mandate_data( $index, $value, $expected ) {
+		unset( $_SERVER['REMOTE_ADDR'] );
+		unset( $_SERVER['HTTP_X_REAL_IP'] );
+		unset( $_SERVER['HTTP_X_FORWARDED_FOR'] );
+
+		$_SERVER[ $index ] = $value;
+		$request           = WC_Stripe_Helper::add_mandate_data( [] );
+		$ip_address        = $request['mandate_data']['customer_acceptance']['online']['ip_address'];
```
daledupreez: Nit: Given that this is a unit test, it's OK if this falls over with a fatal exception, but I would prefer to make sure that a failure is reported directly rather than indirectly.
```suggestion
		$this->assertTrue( isset( $request['mandate_data']['customer_acceptance']['online']['ip_address'] ) );
		$ip_address        = $request['mandate_data']['customer_acceptance']['online']['ip_address'];
```

tests/phpunit/test-wc-stripe-helper.php
```diff
@@ -507,4 +507,42 @@ public function test_add_stripe_methods_in_woocommerce_gateway_order() {
...
+	 * @param string $expected The expected IP address.
+	 * @dataProvider provider_test_add_mandate_data
+	 * @return void
+	 */
+	public function test_add_mandate_data( $index, $value, $expected ) {
+		unset( $_SERVER['REMOTE_ADDR'] );
+		unset( $_SERVER['HTTP_X_REAL_IP'] );
+		unset( $_SERVER['HTTP_X_FORWARDED_FOR'] );
+
+		$_SERVER[ $index ] = $value;
```
daledupreez: Non-blocking nit: the variable names `$index`, `$value`, and `$expected` are all super-generic. It would be nice to use more specific variable names so it's easier to grok what each value is doing and represents.
```suggestion
	/**
	 * Test for `add_mandate_data`.
	 *
	 * @param string $server_variable_key   The key of the server variable to set.
	 * @param string $server_variable_value The value to set the server variable to.
	 * @param string $expected_ip_address    The expected IP address.
	 * @dataProvider provider_test_add_mandate_data
	 * @return void
	 */
	public function test_add_mandate_data( $server_variable_key, $server_variable_value, $expected_ip_address ) {
		unset( $_SERVER['REMOTE_ADDR'] );
		unset( $_SERVER['HTTP_X_REAL_IP'] );
		unset( $_SERVER['HTTP_X_FORWARDED_FOR'] );

		$_SERVER[ $server_variable_key ] = $server_variable_value;
```

daledupreez (approved):
Changes look good and test well. I have two minor, non-blocking comments in the unit tests that you may or may not want to address.

PR #4294, opened 2025-05-07
Fix payment method names in the new react-based settings page.

malithsen:
Fixes STRIPE-417

This PR addresses an issue where Stripe payment methods (such as SEPA, Bancontact, etc.) were not being recognized or displayed in the WooCommerce edit order page when the "New payments settings experience" is enabled. Previously, the `hide_gateways_on_settings_page` function was unconditionally removing these alternative payment methods from the global payment gateways list. This payment method removal is needed for the current settings pages, as we don't want to show every payment method offered under the Stripe gateway as a separate gateway. However, with the new React-based settings page, this appears to be no longer needed (Admittedly, I haven't found where in the WooCommerce code this is now handled ). 

With the new React-based payments settings, WooCommerce expects all available payment methods to be present in the gateways list for correct display and order management. Hiding these methods causes them to appear as undeclared or with generic IDs (e.g., stripe_sepa_debit) in the admin order view.

This PR proposes a simple conditional check at the start of `hide_gateways_on_settings_page` to detect if the new payments settings experience is enabled and to return early.

## Testing instructions

1. As a shopper, purchase a subscription using SEPA.
2. As merchant, go to WooCommerce -> Settings -> Advanced -> Features and enable "Enable the new payments settings experience".
3. Go to order edit page of the subscription purchased in step 1.
4. Make sure payment method name is correctly displayed.
5. Under "Related orders" click the subscription order. Make sure Payment method shows up as "SEPA Direct Debit" instead of "Manual Renewal"
6. (Regression test) Go to WooCommerce -> Settings -> Payments page, make sure the gateway list doesn't list the individual payment methods offered under Stripe as gateways.

Before
CleanShot 2025-05-07 at 14 24 27@2x

CleanShot 2025-05-07 at 14 36 21@2x

Now
CleanShot 2025-05-07 at 14 23 43@2x

CleanShot 2025-05-07 at 14 37 13@2x

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-settings-controller.php
+++ b/includes/admin/class-wc-stripe-settings-controller.php
@@ -273,7 +273,7 @@ class WC_Stripe_Settings_Controller {
 	 */
 	public static function hide_gateways_on_settings_page() {
 		// Prevent hiding gateways in the new payments settings experience (React-based UI).
-		if ( 'yes' === get_option( 'woocommerce_feature_reactify-classic-payments-settings_enabled' ) ) {
+		if ( class_exists( '\Automattic\WooCommerce\Utilities\FeaturesUtil' ) && \Automattic\WooCommerce\Utilities\FeaturesUtil::feature_is_enabled( 'reactify-classic-payments-settings' ) ) {
 			return;
 		}
 
--- a/includes/admin/class-wc-stripe-settings-controller.php
+++ b/includes/admin/class-wc-stripe-settings-controller.php
@@ -273,7 +273,7 @@ class WC_Stripe_Settings_Controller {
 	 */
 	public static function hide_gateways_on_settings_page() {
 		// Prevent hiding gateways in the new payments settings experience (React-based UI).
-		if ( get_option( 'woocommerce_feature_reactify-classic-payments-settings_enabled' ) === 'yes' ) {
+		if ( 'yes' === get_option( 'woocommerce_feature_reactify-classic-payments-settings_enabled' ) ) {
 			return;
 		}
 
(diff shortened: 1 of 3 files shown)


includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -272,6 +272,11 @@ public function admin_scripts( $hook_suffix ) {
 	 * to display the payment gateways on the WooCommerce Settings page.
 	 */
 	public static function hide_gateways_on_settings_page() {
+		// Prevent hiding gateways in the new payments settings experience (React-based UI).
+		if ( 'yes' === get_option( 'woocommerce_feature_reactify-classic-payments-settings_enabled' ) ) {
```
daledupreez: Side note: it feels like we shouldn't need to depend on the raw option name to work out whether this feature is enabled or not. I'll kick off an internal thread to see whether a wrapper is reasonable, and I may well move to an issue in WooCommerce.

daledupreez (approved):
Changes look good to me and test well with the payment settings experience enabled and disabled.

includes/admin/class-wc-stripe-settings-controller.php
daledupreez: It looks like we may be able to do this in a way that includes default values etc:
```suggestion
		$woo_feature_controller = wc_get_container()->get( \Automattic\WooCommerce\Internal\Features\FeaturesController::class );
		if ( $woo_feature_controller && $woo_feature_controller->feature_is_enabled( 'reactify-classic-payments-settings' ) ) {
```

malithsen: @daledupreez Thanks for pointing me in the right direction!

Looks like there's a [utility class](https://github.com/woocommerce/woocommerce/blob/02669dcfc2e7a379ac8795966815a4e36d9cd30a/plugins/woocommerce/src/Utilities/FeaturesUtil.php#L41) we can use, I've updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4294/commits/2ee1c4d0b4890efdab58b9c57c4b530a316ad90e

PR #4297, opened 2025-05-08
Only store appearance styles on checkout pages

### LINKED ISSUE #2417: UPE design is not consistent if appearance transient gets created from my account or pay for order page
**Describe the bug**
UPE design is not consistent if appearance transient gets created from my account or pay for order page.

This is because if you don't have the `wc_stripe_upe_appearance` created before visiting My account or the Pay for order page, it will get created without essential stylesheets derived from elements from the checkout page.

**To Reproduce**
1. Delete `wc_stripe_upe_appearance` transient.
2. Visit My account > Add payment method > Stripe.
3. Notice how the UPE element looks different.
4. Delete `wc_stripe_upe_appearance` transient again.
5. Visit the checkout page.
6. Notice how the UPE element looks how it should be.

**Expected behavior**
I'm not sure how exactly we could fix this, but perhaps we should not create the transient from a page other than checkout.

malithsen:
Fixes STRIPE-114
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2417

This PR updates the logic for persisting the appearance settings. Previously, the frontend would always store the appearance styles regardless of the context. With this change, we only store the appearance data from checkout pages. 
Other pages(Add payment method, change payment method, pay for order) will compute the appearance data or reuse any existing data (same as before).

## Testing instructions

1. Remove appearance transients (I do this directly on the db by deleting the results of `select * from wp_options where option_name like '%appearance%'`)
2. Visit My account > Add payment method > Stripe.
3. Visit the shortcode checkout page.
4. Notice the styles of the payment element.
5. Remove the transient again.
6. This time, visit the checkout page (without visiting the my account -> add payment)
7. Notice the styles of the payment element, it should look exactly as before (Step 4)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/stripe-utils/utils.js
+++ b/client/stripe-utils/utils.js
@@ -158,18 +158,6 @@ export const getUPETerms = ( value = 'always' ) => {
 };
 
 /**
- * `storageKeys` object contains keys for storing values related to the appearance
- *  settings of Stripe UPE (Universal Payment Element) in different contexts.
- */
-export const storageKeys = {
-	// Key to store the appearance settings for Stripe UPE in the general WooCommerce context
-	UPE_APPEARANCE: 'wc_stripe_upe_appearance',
-
-	// Key to store the appearance settings for Stripe UPE when used within WooCommerce Blocks context
-	WC_BLOCKS_UPE_APPEARANCE: 'wc_stripe_wc_blocks_upe_appearance',
-};
-
-/**
  * Sets a key-value pair in the localStorage along with a time-to-live (TTL) value, which specifies
  * the time (in milliseconds) after which the item will be considered expired.
  *
--- a/client/stripe-utils/utils.js
+++ b/client/stripe-utils/utils.js
@@ -616,14 +616,15 @@ export const initializeUPEAppearance = ( api, isBlockCheckout = 'false' ) => {
 			: getStripeServerData()?.appearance;
 
 	const data = getStripeServerData();
-	const isValidContext =
+
+	const isValidUpdateContext =
 		isBlockCheckout === 'true' ||
 		( data.isCheckout && ! data.isOrderPay && ! data.isChangingPayment );
 
-	// If appearance is empty, only save it if in a valid context.
 	if ( ! appearance ) {
 		appearance = getAppearance( isBlockCheckout === 'true' );
-		if ( isValidContext ) {
+		// If we have re-built the appearance, only update the settings in the checkout context
+		if ( isValidUpdateContext ) {
 			api.saveAppearance( appearance, isBlockCheckout );
 		}
 	}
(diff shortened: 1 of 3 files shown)


client/stripe-utils/utils.js
```diff
@@ -627,10 +615,17 @@ export const initializeUPEAppearance = ( api, isBlockCheckout = 'false' ) => {
...
+		isBlockCheckout === 'true' ||
+		( data.isCheckout && ! data.isOrderPay && ! data.isChangingPayment );
+
+	// If appearance is empty, only save it if in a valid context.
 	if ( ! appearance ) {
 		appearance = getAppearance( isBlockCheckout === 'true' );
-		api.saveAppearance( appearance, isBlockCheckout );
+		if ( isValidContext ) {
+			api.saveAppearance( appearance, isBlockCheckout );
+		}
```
daledupreez: I think we should update the naming for `isValidContext` to `isValidUpdateContext`, as "isValidContext" implies that we're not in a valid context, which is pretty broad.
```suggestion
	const isValidUpdateContext =
		isBlockCheckout === 'true' ||
		( data.isCheckout && ! data.isOrderPay && ! data.isChangingPayment );

	if ( ! appearance ) {
		appearance = getAppearance( isBlockCheckout === 'true' );
		// If we have re-built the appearance, only update the settings in the checkout context
		if ( isValidUpdateContext ) {
			api.saveAppearance( appearance, isBlockCheckout );
		}
```

At a more subtle level, we only care about `isValidUpdateContext` if we have ref-fetched the content, so I have a slight preference for pushing the definition into that code block so the logic is only running when we need it. That's mostly from a code grokking perspective -- I don't like having to read through a complex assignment only to realise it is actually irrelevant later in the code. (This suggestion is non-blocking, to be clear.)

daledupreez (approved):
I have some minor quibbles about the naming and the code grouping, but they're not blocking, and this is working otherwise.

I also updated the description to flag two things:
1. I linked the GitHub issue as well as the Linear issue
2. I specified that shortcode checkout needs to be visited in the testing instructions -- we cache appearance data for blocks in a separate transient. 😁

client/stripe-utils/utils.js
malithsen: Great suggestions. Github didn't allow me to apply your suggestion directly, so I did it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4297/commits/9c12069374d4911c8f9d3eaa1fd80280b9f297b3

Also moved the definition into the branch in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4297/commits/02805ab7b77020b54dae167f7b844f38a0e5b4b0

PR #4299, opened 2025-05-09
Show recurring icon in payment methods

### LINKED ISSUE #2831: [Settings] Missing information about the payment method types' support for Automatic Recurring Payments
**Describe the bug**
When [Woo Subscriptions](https://woo.com/products/woocommerce-subscriptions/) is enabled, and the classic checkout experience (non-UPE) is enabled, a column titled “Automatic Recurring Payments” is added to the Payment Methods table under WooCommerce > Settings > Payments page.

This column displays the capability of each payment method to process automatic recurring payments for subscriptions. We’re not retaining this information in the new interface, under the WooCommerce > Settings > Payments > Stripe settings > Payment Methods tab.

**To Reproduce**

**Expected behavior**

**Screenshots**
Before and after:

**Additional context**

Mayisha:
Fixes STRIPE-192
Fixes #2831 

- Added a recurring icon with a tooltip in payment methods that supports subscription.

## Testing instructions
- Go to the Stripe settings page in your wp-admin.
- Confirm that there is a new recurring icon beside the payment methods that supports subscriptions with a tooltip.

Screenshot 2025-05-09 at 3 29 38 PM

Screenshot 2025-05-09 at 3 29 11 PM


changelog.txt
```diff
@@ -29,6 +29,7 @@
 * Dev - Minor fix to e2e setup code
 * Dev - Make PHP error log from Docker container available in docker/logs/php/error.log
 * Fix - Fix invalid IP address error encountered during mandate data creation.
+* Add - Show an icon beside the payment methods that supports automattic recurring payments.
```
daledupreez: Two nits:
* `automattic` => `automattic` 😁 
* I think "supports" should be "support" because the payment methods are a plural
```suggestion
* Add - Show an icon beside the payment methods that support automatic recurring payments.
```

readme.txt
```diff
@@ -139,5 +139,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Dev - Minor fix to e2e setup code
 * Dev - Make PHP error log from Docker container available in docker/logs/php/error.log
 * Fix - Fix invalid IP address error encountered during mandate data creation.
+* Add - Show an icon beside the payment methods that supports automattic recurring payments.
```
daledupreez: As above RE the changelog message.
```suggestion
* Add - Show an icon beside the payment methods that support automatic recurring payments.
```

client/payment-methods-map.js
```diff
@@ -45,6 +47,7 @@ const paymentMethodsMap = {
 		Icon: icons.card,
 		currencies: [],
 		allows_manual_capture: true,
+		supports_recurring: true,
```
daledupreez: Nit: given that this is a Javascript value, can we use `supportsRecurring` as a variable name?
Also, to keep noise down, it may be worth only setting the value when it is true or needs to be computed explicitly.

client/components/recurring-payment-icon/index.js
```diff
@@ -0,0 +1,64 @@
...
+			width: auto;
+		}
+	}
+`;
+
+const IconComponent = ( { children, ...props } ) => (
+	<span { ...props }>
+		<Icon src={ icon } alt="" />
+		{ children }
+	</span>
```
daledupreez: This `span` element means that we're not getting automatic alignment from the `div` with `inline-flex` that wraps the elements, and we end up with slightly offset/weird vertical alignment.

More specifically, I think the icon should be horizontally aligned with the text, and the current layout has it slightly above that.

client/components/recurring-payment-icon/index.js
```diff
@@ -0,0 +1,64 @@
+import { __ } from '@wordpress/i18n';
+import React, { useState } from 'react';
+import styled from '@emotion/styled';
+import { Popover as PopoverComponent } from '@wordpress/components';
+import icon from './icon.svg';
+
+const Icon = styled.img`
+	height: 12px;
+	width: 12px;
```
daledupreez: Can we make these the same size as the text? At present, I think the text size is `14px` by default.

client/components/recurring-payment-icon/index.js
```diff
@@ -0,0 +1,64 @@
...
+const StyledPopover = styled( PopoverComponent )`
+	top: -11px !important;
+
+	.components-popover__content {
+		border-radius: 4px;
+		box-shadow: 0px 2px 6px 0px rgba( 0, 0, 0, 0.05 );
+		padding: 5px 10px;
+		background-color: #000000;
+		color: #ffffff;
+	}
```
daledupreez: The styles here don't match what we're doing for the currency (i) popover. We should use the same styles or component.
Screenshot 2025-05-13 at 13 56 17

daledupreez (changes requested):
This is really close, but I think we have a few small visual aspects to polish.

client/components/recurring-payment-icon/index.js
Mayisha: Waiting on confirmation from design about this.

client/payment-methods-map.js
Mayisha: Fixed in [2227fb5](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4299/commits/2227fb591e013626640964430996d1028098391d)

client/components/recurring-payment-icon/index.js
Mayisha: Done in [97d80d7](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4299/commits/97d80d7a55c7d5f09138a9fcb7bc13a3677b1a6d)

client/components/recurring-payment-icon/index.js
Mayisha: As per the confirmation from design, we are keeping these different as one if a tooltip and another is a popover. However, in the previous implementation, I was using the popover component for showing `Supports recurring payments` message. I have updated the component in [facce12](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4299/commits/facce12b5604d9f84ce10561b5b120093df918ce)

client/components/recurring-payment-icon/index.js
Mayisha: Initially fixed it in [97d80d7](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4299/commits/97d80d7a55c7d5f09138a9fcb7bc13a3677b1a6d) but the changes are not relevant anymore after updating the code to use the tooltip component [facce12](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4299/commits/facce12b5604d9f84ce10561b5b120093df918ce).

Mayisha:
Thanks @daledupreez, for the feedback. I have made some adjustments in my recent commits and it's ready for review again. 👀

client/components/recurring-payment-icon/index.js
```diff
@@ -0,0 +1,31 @@
+import { __ } from '@wordpress/i18n';
+import React from 'react';
+import styled from '@emotion/styled';
+import icon from './icon.svg';
+import Tooltip from 'wcstripe/components/tooltip';
+
+const Icon = styled.img`
+	height: 14px;
+	width: 14px;
+	cursor: pointer;
```
daledupreez: Do we need this? I am not sure that the item is clickable, so adding a pointer cursor doesn't feel right. Also, it doesn't seem to be visible for me! 🙃

daledupreez (approved):
Overall, this is working well, but I think we can remove the one `cursor` CSS rule that came up.

PR #4304, opened 2025-05-12
Fix payment method title when using the shortcode checkout with OC enabled

wjrosa:
Fixes STRIPE-424

This PR fixes the payment method title for the shortcode checkout page when the Optimized Checkout is enabled. Previously, not only would the OC container display the custom title (which defaults to "Stripe"), but the title would also be added as the payment method title inside the order metas (and shown in the order details page).

The fix consists of changing the main title logic on the backend, removing the specific shortcode part, and adding a frontend part to handle that correctly.

## Testing instructions

- Checkout and build this branch on your test environment (`fix/shortcode-checkout-payment-method-title-when-oc-is-enabled`)
- Connect your Stripe account
- Enable the OC feature flag (`_wcstripe_feature_oc`). You can do it by either hardcoding the return value of `is_oc_available` to `true` or by running `npm run wp option update _wcstripe_feature_oc 'yes'`
- As a merchant, disable the legacy checkout experience in settings (`wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings`)
- Define a custom title for the container below the setting

### Shortcode checkout

- Go to the classic checkout page
- Confirm the custom title is displayed above the Optimized Checkout container
- Complete the purchase
- Confirm the "thank you" page displays the correct payment method (and not the custom title)

### Block checkout

- Confirm the custom title is displayed above the Optimized Checkout container
- Complete the purchase
- Confirm the "thank you" page displays the correct payment method (and not the custom title)

### Pay for order

- As a merchant, create a new order for your existing shopper user
- As the shopper, try to pay for it using the "My Account" page
- Confirm the custom title is displayed above the Optimized Checkout container
- Complete the purchase
- Confirm the "thank you" page displays the correct payment method (and not the custom title)

### Order details

- As a merchant, open any paid order using this branch
- Confirm the payment method is correctly displayed:
Screenshot 2025-05-13 at 09 25 32

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-method-cc.php
```diff
@@ -41,27 +41,23 @@ public function __construct() {
...
-			return $this->oc_title;
+			// Block checkout and pay for order page.
+			if ( has_block( 'woocommerce/checkout' ) || ! empty( $_GET['pay_for_order'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+				return $this->oc_title;
+			}
+		}
+
+		// Default implementation
+		$wallet_type = $payment_details->card->wallet->type ?? null;
+		if ( $payment_details && $wallet_type ) {
```
annemirasol: Should this be just `if ( $wallet_type )`, since `$payment_details` being set is already implied?

annemirasol (changes requested):
- When using a 3DS2 card, I still get the OC title as the payment method title in the order confirmation page. I used `4000002500003155` to test. Block and classic checkout behave the same.

- When using Google Pay, I notice the payment method in the order confirmation page says "Credit / Debit Card" instead of "Google Pay (Stripe)", which is the behavior in develop.

includes/payment-methods/class-wc-stripe-upe-payment-method-cc.php
wjrosa: Totally! Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4304/commits/d0517b7571dfb5b07fc3b1cf57aa10b5339699db

wjrosa:
Thanks for the review, Anne! I have fixed both issues in my latest commits. I am not super happy with the code; it feels a lot like a "short blanket" situation, so I am open to any suggestions.

annemirasol:
> I am not super happy with the code; it feels a lot like a "short blanket" situation, so I am open to any suggestions.

Which part are you not super happy with?

annemirasol:
Flagging your attention to several issues I encountered with Optimized Checkout that I believe are out of scope for this PR:

1. WeChat not working in block checkout. Unable to check classic checkout due to issue #2.
Screenshot 2025-05-16 at 9 37 02 AM


2. Not all enabled payment methods appear in classic checkout. Screenshot is classic vs block, same payment method configuration. Notice how only credit card is available in classic checkout.
Screenshot 2025-05-16 at 9 25 54 AM


3. "Save payment information to my account" is appearing even for logged out users.
Screenshot 2025-05-16 at 9 21 34 AM


4. When changing the OC title, it doesn't let me delete the text entirely. I can only change it if I highlight, then type to overwrite.
https://github.com/user-attachments/assets/ddc7c114-e62c-4142-a378-d3218d32e26b

annemirasol (approved):
I'm gonna go ahead and approve this PR, as it does fix the issue described.

I left a comment about other surfaced issues, which I believe are out of scope and should be handled in a separate PR.

✅ Tested classic checkout using normal card, 3DS card, CashApp and ACH -- payment method titles are correct
✅ Tested block checkout with normal card, 3DS card, Affirm -- payment method titles are correct
✅ Tested Google Pay -- payment method title is correct

wjrosa:
Thanks for flagging those issues, Anne! I will create new issues to handle them. For the "3", the saving checkbox shows up for you even if the "Create an account" checkbox is not checked?

annemirasol:
> For the "3", the saving checkbox shows up for you even if the "Create an account" checkbox is not checked?

@wjrosa Nope, it wasn't. The option to create an account on checkout was not even enabled.

wjrosa:
@annemirasol just to close the loop regarding the issues you found, I have created them now (STRIPE-463 to STRIPE-465 in Linear).

> Which part are you not super happy with?

Basically, the whole `get_optimized_checkout_title` method. I developed it trying to think of all possible ways it should not return the `$this->oc_title` value. So, whenever the custom title appears somewhere, that method must be updated. It does not feel much right to me.

PR #4307, opened 2025-05-13
Update payment_button_request_size upgrade checks

### LINKED ISSUE #4290: Not able to save anything in Stripe settings with "Error Save Settings" message.
After I uncheck Legacy checkout experience, Stripe setting start to bug since that time. I now have "Pay by Link" button on my page, that im not able to get rid of. Every time I try to edit and save anything in the setting, the is "Error Save Settings" popup message and nothing happens. Im also not able to save Log error messages to see where is the problem. Please help me with this problem, thank you.

daledupreez:
Fixes STRIPE-426
Fixes #4290

This PR makes some small logic changes to our upgrade code for the `payment_request_button_size` field in our plugin settings. Prior to this change, we would only try to migrate values when we detect that the previously installed version of the gateway was older than 7.8.0 _and_ we successfully loaded the gateway instance.

As indicated by #4290, we have at least one merchant who still had the now-invalid value of `'medium'` for that field. This PR modifies the logic slightly by ensuring that we migrate the `'medium'` value regardless of which version we're working with.

At an implementation level, the logic for checking the migration on update now does the following:
 * If we don't have a previous version, do nothing.
 * If there is no value for the `payment_request_button_size` field, do nothing.
 * If the previous version is >= 7.8.0:
   - If the value for the `payment_request_button_size` field is anything other than `'medium'`, do nothing.
     * The implicit "else" for this condition, where we fall through to the remaining logic, is the main change in this PR.
 * If we can't load the gateway, do nothing.
 * If the `payment_request_button_size` field is `'default'`, change the value to `'small'`
 * If the `payment_request_button_size` field is `'medium'`, change the value to `'default'`

I would also love some feedback on the unit tests -- does it make sense to have the tests grouped by function (`Version_Updates`) vs by the underlying class?

## Testing instructions

I believe that the unit tests should give us confidence in making this change, but I think we can discuss whether manual tests should also be tackled to confirm this is working as expected.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


annemirasol (commented):
> does it make sense to have the tests grouped by function (Version_Updates) vs by the underlying class?

I think there's argument to be made for either approach. I personally prefer grouping by class, if only to be consistent with the rest of the codebase. Ideally, the class itself is a grouping of some function anyway. In this case, a grouping of functions related to PRBs.

And speaking of consistency, what do you think about using `WC_Stripe_Logger::log()` instead of `error_log()`?

daledupreez:
Thanks for the feedback, @annemirasol! I've updated the tests and I think this code is ready for another look.

annemirasol (approved):
Logic looks solid, LGTM 🚢

PR #4310, opened 2025-05-14
Fix error in setting default payment method

### LINKED ISSUE #2863: Deprecate/adjust set_default_source
**Describe the bug**

I was looking through the logs on a bunch of sites we host, and I noticed that every transaction generates a series of API requests to Stripe, one of which always responds with a `400 ERR` error code:

Screenshot 2024-01-26 at 2 41 01 PM
(example)

The error is not phrased well, but `default_source` param of the API does not accept PaymentMethod `pm_abcd1234` tokens. In fact [that field appears to be deprecated in general](https://github.com/stripe/stripe-dotnet/issues/1843#issuecomment-556122912). [This appears to be the offending bit of code.](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/ee0ea892855b8c0614d163fc06d17f82e6323543/includes/class-wc-stripe-customer.php#L530)

**To Reproduce**
1. Purchase something using WooCommerce with the Stripe Gateway as a new customer.
2. Go to Stripe. Find the customer and go to their "Events & Logs" tab.
3. Observe 4 requests to the API which respond with `200` and one which responds with `400`:

Screenshot 2024-01-26 at 3 00 07 PM

**Expected behavior**
Probably makes the most sense to [use `default_payment_method` when the token starts with `pm_` ](https://stripe.com/docs/api/customers/update#update_customer-invoice_settings-default_payment_method) and `default_source` when the token starts with `src_`

**Screenshots**
See above

 - WordPress Version: latest
 - WooCommerce Version: 8.5.2
 - Stripe Plugin Version: 7.9.1
 - Any other plugins installed: A variety, but shouldn't be relevant here.

Mayisha:
Fixes STRIPE-196
Fixes #2863 

- Calling the correct `set_default_` function based on the token type. If the saved token is a payment method, set default using `set_default_payment_method`. Otherwise, use `set_default_source`.
- These changes are to support the special case, if any store reverts to the legacy experience.

## Testing instructions
- Enable the new checkout experience. 
- As a new shopper, log in and go to **My Account > Payment Method**
- Add a few payment methods.
- Now enable the legacy checkout experience.
- Use `develop` branch.
- As shopper, go to **My Account > Payment Method** and set another payment method as default.
- Check your Stripe logs (https://dashboard.stripe.com/test/logs). You should see a failed API request (status code 400) to `/customers/<id>` with a message `No such source: 'pm_xx`. 
- Now checkout this branch.
- As a shopper, set another payment method as the default.
- Check your Stripe logs (https://dashboard.stripe.com/test/logs) and confirm no failed API request. 
- Check the latest `/customer/<id>` call and confirm that the request was made with `default_payment_method`.


includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -522,11 +522,10 @@ public function woocommerce_payment_token_set_default( $token_id ) {
 		$stripe_customer = new WC_Stripe_Customer( get_current_user_id() );
 
 		try {
-			if ( WC_Stripe_Feature_Flags::is_upe_checkout_enabled() ) {
-				if ( WC_Stripe_UPE_Payment_Gateway::ID === $token->get_gateway_id() ) {
-					$stripe_customer->set_default_payment_method( $token->get_token() );
-				}
-			} elseif ( WC_Gateway_Stripe::ID === $token->get_gateway_id() || WC_Gateway_Stripe_Sepa::ID === $token->get_gateway_id() ) {
+			// If the token is a payment method, set it as the default payment method. If the token is a source, set it as the default source.
+			if ( strpos( $token->get_token(), 'pm_' ) === 0 ) {
```
annemirasol, line 526: Do you see us needing this logic anywhere else? If yes, it might be good to move these to helper functions, e.g. `is_source_token( $token )`, `is_payment_method_token( $token )`

annemirasol (approved):
Left a non-blocking suggestion.

- Was able to reproduce the issue
- Verified this branch fixes it -- no more 400 error

includes/payment-tokens/class-wc-stripe-payment-tokens.php
Mayisha: I didn't see any token check when I skimmed through the codebase. Keeping it as it is for now.

PR #4312, opened 2025-05-14
Fix methods not available in classic checkout for OC

wjrosa:
Fixes STRIPE-425

Currently, the classic/shortcode checkout does not support some payment methods when the Optimized Checkout feature is enabled. That's because OC renders all methods dynamically in the same payment element container, and some methods might not work correctly when the `setupFutureUsage` option is absent during Stripe initialization (option required to flag when the method must be saved for future usage/subscriptions). The classic checkout implementation does not provide an `update` method to change Stripe options on the fly. To avoid such issues, we hardcoded `setupFutureUsage` to `off_session`.

The fix here consists of removing the hardcoded value and remounting the whole Stripe payment element every time the "Save payment method" value checkbox changes.

## Testing instructions

- Checkout and build this branch on your test environment (`fix/methods-not-available-in-oc-when-saving-them`)
- Use a Polish Stripe account to make testing easier, with my steps here
- On your Stripe dashboard, enable the Przelewy24 payment method
- Connect your Stripe account to your store
- Enable the OC feature flag (`_wcstripe_feature_oc`). You can do it by either hardcoding the return value of `is_oc_available` to `true` or by running `npm run wp option update _wcstripe_feature_oc 'yes'`
- As a merchant, disable the legacy checkout experience in settings (`wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings`)
- Set your currency to EUR

### Regular product

- As a shopper, add any non-subscription product to your cart
- Go to the shortcode checkout page
- Confirm all payment methods are available
- Click the "Save payment method" checkbox
- Confirm P24 is not available anymore

### Subscription product

- As a shopper, add any non-subscription product to your cart
- Go to the shortcode checkout page
- Confirm P24 is not available

If you want, you can test this with other payment methods that do not support reuse: Alipay, Boleto, EPS, etc.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/classic/upe/payment-processing.js
```diff
@@ -161,6 +162,16 @@ async function createStripePaymentElement( api, paymentMethodType ) {
...
+	const setupFutureUsage =
+		document.getElementById( 'wc-stripe-new-payment-method' ).checked ||
+		getStripeServerData()?.cartContainsSubscription;
+	if ( setupFutureUsage ) {
+		options = {
+			...options,
+			setupFutureUsage: 'off_session',
+		};
+	}
+
```
annemirasol: By placing this here, we're also changing the options setup for non-OC. Is this intentional?

wjrosa: Before the OC implementation, the shortcode checkout had no `setupFutureUsage` option set. This could potentially fix other issues with non-OC transactions. But I will bind it to OC for now to avoid unintended changes and keep the scope of the PR. Edit: done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4312/commits/001e29e51b531601a20963cc6d2e9a73de6c95a2

annemirasol (approved):
LGTM!

Tested for classic checkout:
- Smoke-tested non-OC flows
- OC: Able to see all enabled payment methods
- OC: Able to see only reusable payment methods, when "save payment method" is checked
- OC: Smoke-tested completing the purchase (card, CashApp)
- OC: Smoke-tested completing the purchase + saving payment method (card, CashApp)

PR #4319, opened 2025-05-15
Fix order currency fatal error

wjrosa:

This PR fixes a possible fatal error that happens when we check if a payment method is available on the checkout using a specific order ID:
```
PHP Fatal error:  Uncaught Error: Call to a member function get_currency() on bool in /srv/htdocs/wp-content/plugins/woocommerce-gateway-stripe/includes/payment-methods/class-wc-stripe-upe-payment-method.php:291
```

This error was caught with Grafana with Stripe version 9.5.0.

## Testing instructions

This seems to be an edge case, so code review should be enough. Check if the tests are still passing.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


annemirasol (approved):
Changes look solid, LGTM!

PR #3863, opened 2025-02-10
Adding admin notices for detached subscriptions

wjrosa:
See: p1739145321956079-slack-C7U3Y3VMY

## Changes proposed in this Pull Request:

This PR adds new admin notices to alert merchants when subscriptions payment methods are detached, preventing renewals from happening. This should help them be more proactive in identifying and fixing this kind of issue.

## Testing instructions

- Checkout and build this branch on your test environment (`add/admin-notices-for-detached-subscriptions`)
- Connect your Stripe account
- Install the subscriptions extension
- Create a subscription product
- As a shopper, purchase the subscription
- As a merchant, access the Stripe dashboard and delete the payment methods bound to the shopper account above
- Confirm that you can see the new notice in wp-admin:
Screenshot 2025-02-11 at 13 27 39

(you can also just set `$payment_method->customer` to `null` in the `subscriptions_check_environment` method)

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [x] Added changelog entry **in both** `changelog.txt` and `readme.txt` (or does not apply)
-   [ ] Tested on mobile (or does not apply)

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4320, opened 2025-05-15
Removing the detached subscriptions notice feature

wjrosa:
See p1747316165642649-slack-C7YPW6K40?thread_ts=1747242351.267229&cid=C7YPW6K40

I am removing the detached subscriptions notice feature (for admins) due to constant issues with it when retrieving the payment method information from Stripe's API. The feature was introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3863. It does not seem that helpful for our merchants, so not worth it keeping it.

## Testing instructions

You can also check out this branch on your test environment and confirm we are no longer trying to retrieve the list of subscriptions with payment method information to build the admin notice.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
+* Fix - Removes the subscription detached notice feature to avoid issues when requesting the Stripe's API to retrieve payment method details.
```
daledupreez: This changelog entry is a bit confusing to me. Maybe we can phrase it as follows?
```suggestion
* Fix - Stop checking for detached subscriptions for admin users, as it was slowing down wp-admin
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -467,8 +462,11 @@ public function payment_methods_check_environment() {
 	 * Environment check for subscriptions.
 	 *
 	 * @return void
+	 *
+	 * @deprecated 9.6.0 This method is no longer used and will be removed in a future version.
```
daledupreez, line 466: Rather than marking this as deprecated, is it worth making this runnable via a manual check/button? I can still see value in getting the data based on a specific demand.

readme.txt
```diff
@@ -112,6 +112,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 
 = 9.6.0 - xxxx-xx-xx =
 
+* Fix - Removes the subscription detached notice feature to avoid issues when requesting the Stripe's API to retrieve payment method details.
```
daledupreez: As above RE the changelog entry.

tests/phpunit/compat/test-class-wc-stripe-subscriptions-helper.php
```diff
@@ -17,55 +17,4 @@ class WC_Stripe_Subscriptions_Helper_Test extends WP_UnitTestCase {
...
-		$expected = [
-			[
-				'id'                        => $subscription_id,
-				'customer_id'               => $customer_id,
-				'change_payment_method_url' => $subscription->get_change_payment_method_url(),
-			],
-		];
-		$this->assertEquals( $expected, WC_Stripe_Subscriptions_Helper::get_some_detached_subscriptions() );
-
-		remove_filter( 'pre_http_request', $test_request, 10, 3 );
```
daledupreez, line 66: Nit: We should remove the filter before we run the assertEquals(), as we won't clean up if the assertion fails.

tests/phpunit/compat/test-class-wc-stripe-subscriptions-helper.php
```diff
@@ -17,55 +17,4 @@ class WC_Stripe_Subscriptions_Helper_Test extends WP_UnitTestCase {
 	public function test_is_subscriptions_enabled() {
 		$this->assertTrue( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() );
 	}
-
-	/**
-	 * Test for `get_some_detached_subscriptions`.
-	 *
-	 * @return void
-	 */
-	public function test_get_some_detached_subscriptions() {
```
daledupreez, line 26: If we decide to keep the functionality as a manual call/check, we may want to keep this test.

daledupreez (commented):
I have some initial feedback based on inspection.

annemirasol:
Noting here that this line may be what's causing the caching feature to fail: https://github.com/woocommerce/woocommerce-gateway-stripe/blob/0f60a6605a70371ac839e2aef946dbbcbc67452d/includes/compat/class-wc-stripe-subscriptions-helper.php#L34

If the store does not have detached subscriptions, or our sampling did not return any, we cache an empty array which will pass the "is empty" check.

wjrosa:
> Noting here that this line may be what's causing the caching feature to fail:
> 
> https://github.com/woocommerce/woocommerce-gateway-stripe/blob/0f60a6605a70371ac839e2aef946dbbcbc67452d/includes/compat/class-wc-stripe-subscriptions-helper.php#L34
> 
> If the store does not have detached subscriptions, or our sampling did not return any, we cache an empty array which will pass the "is empty" check.

Good catch, Anne! Makes total sense to me

readme.txt
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4320/commits/96e49ea4dccc93e880fec527eb49874dc502012a

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: My initial idea was to be proactive for merchants reaching out to us all the time to report issues with their subscriptions, with no payment method attached. But no one ever asked for this. And since this is requiring constant work and causing new issues, I don't think it is worth keeping it. But happy to rework this again on another PR if you disagree. We can just revert the deprecation later if needed (as we did a couple times).

annemirasol (approved):

PR #4322, opened 2025-05-15
Track charge completed via webhooks in order note

Mayisha:
Added `(via webhook)` in the order note related to charge completion. This will help track the status change done via webhook vs synchronous payment request.
The main intention of this addition is to help investigate the duplicate charge/order note issues.

## Testing instructions
- Code review.


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1235,6 +1237,7 @@ protected function handle_deferred_payment_intent_succeeded( $order, $intent_id
 		WC_Stripe_Logger::log( "Processing Stripe PaymentIntent {$intent_id} for order {$order->get_id()} via deferred webhook." );
 
 		do_action( 'wc_gateway_stripe_process_payment', $charge, $order );
+		$charge->is_webhook_response = true;
```
diegocurbelo, line 1240: We are getting this error here during tests:
> Attempt to assign property 'is_webhook_response' of non-object

diegocurbelo (changes requested):
Some tests are failing because of the new `is_webhook_response` property. 

Maybe we can create a DTO for the charge, or add a `is_webhook` param (or similar name) to `process_response( $charge, $order )`

Mayisha:
Thanks @diegocurbelo for the feedback.

> Maybe we can create a DTO for the charge, or add a is_webhook param (or similar name) to process_response( $charge, $order )

In my initial implementation, I added an `is_webhook` param to `process_response( $charge, $order )`. I changed it after Anne pointed out to 3PP integration in p1747834069601559/1747831256.786549-slack-C055WHLA98D. 

I have fixed the mocks in [bcd5273](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4322/commits/bcd5273e900d4e22dbaed3f062da4d1b55ac46bd). I noticed that this `get_latest_charge_from_intent` function is mocked in [another file](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/trunk/tests/phpunit/payment-methods/test-class-wc-stripe-upe-payment-gateway.php#L911) and we are mocking the data as an object. 

Let me know what you think about both of the points above.

diegocurbelo (approved):
Thanks for the 3PP integration context 👍🏼 

Screenshot 2025-05-27 at 17 35 31

LGTM 🚢

PR #4323, opened 2025-05-15
Prevent Stripe API calls after detecting the API keys are not valid (401 response from API)

### LINKED ISSUE #4270: Unauthorized API requests tried far to many times
**Describe the bug**

A failed authorization attempt to the API results in a large number of retries, after a test account of mine was disabled 119 failed attempts were made to the stripe gateway, 119 in total:

* 96 to the `v1/payment_method_configurations` endpoint
* 21 to the `v1/account` endpoint 
* 1 to the `v1/payment_method_domains` endpoint
* 1 o the `v1/webhook_endpoints` endpoint

It's possible these were jobs catching up over the course of a week as I'd been on some PTO but the requests were blocking causing the page to apparently stall.

**To Reproduce**
1. Connect to a test account
2. Install and Activate Query Monitor
3. Go in to the database to modify the account details to be incorrect
4. Hit the Dashboard
5. The QM HTTP requests tab will show many failed attempts

**Expected behavior**

Re-authorization should be limited to one or two requests.

**Screenshots**

Image

 - WordPress Version 6.9-alpha-60170
 - WooCommerce Version 9.8.2
 - Stripe Plugin Version dab50f1780d225314dab985e53d285c7d17b065b (feature branch PR)
 - Browser [e.g. chrome, safari] and Version N/A
 - Any other plugins installed
    - Deposits
    - Query Montor
    - Various development focused MU-plugins: speed up cron jobs, prevent order number conflicts, etc.

**Additional context**

This occurred after my account was shutdown during a tidy up (it was intended to be against a woopayments account but they got the wrong one)

diegocurbelo:
Fixes #4270
Relates to STRIPE-445

This PR implements a mechanism to prevent unnecessary Stripe API calls when the API keys are invalid (401 response). 
- Added an option to track invalid API keys state
- Added automatic clearing of invalid API keys state when new keys are saved
- Updated the notice messages, and the state in the connection status check to show the Account as disconnected when API keys are invalid

This helps reduce unnecessary API calls and improves error notices/UI for merchants with invalid API credentials:

Screenshot 2025-05-15 at 16 16 00

Screenshot 2025-05-16 at 10 42 29

Screenshot 2025-05-16 at 12 50 36

## Testing instructions

1. Navigate to `WP-Admin > WooCommerce > Settings > Payments`, and select the `Stripe` option from the list
2. Go to the `Settings` tab and make sure the account is correctly connected:
    Screenshot 2025-05-16 at 13 10 22
3. Replace the API keys with an invalid value:
4. Navigate to the `WP-Admin` dashboard and check that Stripe notice is shwon:
    Screenshot 2025-05-16 at 13 13 54
5. Navigate to `WP-Admin > WooCommerce > Settings > Payments`, and select the `Stripe` option from the list 
6. Go to the `Settings` tab and make sure the account status has the connection error message:
    Screenshot 2025-05-16 at 13 16 09
7. Click `Configure connection` and make sure the Account status is `Disconnected`
    Screenshot 2025-05-16 at 13 16 59
8. Click the `Create or connect a test account` button, a re-connect your account
9. Check that after returning to the WP-Admin dasboard, the top notice is no longere present, and the acount status section has been restored:
    Screenshot 2025-05-16 at 13 25 43

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,7 +7,7 @@
 * Update - Remove BACS from the unsupported 'change payment method for subscription' page.
 * Fix - Fix payment method title display when new payment settings experience is enabled
 * Fix - Prevent styles from non-checkout pages affecting the appearance of Stripe element.
-* Fix - Prevert further Stripe API calls if API keys are invalid (401 response).
+* Fix - Prevent further Stripe API calls if API keys are invalid (401 response).
 
 = 9.5.0 - 2025-05-13 =
 * Fix - Fixes the listing of payment methods on the classic checkout when the Optimized Checkout is enabled.
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,6 +7,7 @@
 * Update - Remove BACS from the unsupported 'change payment method for subscription' page.
 * Fix - Fix payment method title display when new payment settings experience is enabled
 * Fix - Prevent styles from non-checkout pages affecting the appearance of Stripe element.
+* Fix - Prevert further Stripe API calls if API keys are invalid (401 response).
 
 = 9.5.0 - 2025-05-13 =
 * Fix - Fixes the listing of payment methods on the classic checkout when the Optimized Checkout is enabled.
--- a/client/settings/account-details/index.js
+++ b/client/settings/account-details/index.js
@@ -133,11 +133,11 @@ const AccountDetails = () => {
 					{ createInterpolateElement(
 						isTestModeEnabled
 							? __(
-									"Seems like the test API keys we've saved for you are no longer valid. If you recently updated them, use the <strong>Configure Connection</strong> button below to reconnect.",
+									"We couldn't connect to your account, it seems like the test API keys we've saved for you are no longer valid. Please use the <strong>Configure connection</strong> button below to reconnect.",
 									'woocommerce-gateway-stripe'
 							  )
 							: __(
-									"Seems like the live API keys we've saved for you are no longer valid. If you recently updated them, use the <strong>Configure Connection</strong> button below to reconnect.",
+									"We couldn't connect to your account, it seems like the live API keys we've saved for you are no longer valid. Please use the <strong>Configure connection</strong> button below to reconnect.",
 									'woocommerce-gateway-stripe'
 							  ),
 						{
--- a/includes/class-wc-stripe-api.php
+++ b/includes/class-wc-stripe-api.php
@@ -17,6 +17,20 @@ class WC_Stripe_API {
 	const STRIPE_API_VERSION = '2024-06-20';
 
 	/**
+	 * The test mode invalid API keys transient key.
+	 *
+	 * @var string
+	 */
+	const TEST_MODE_INVALID_API_KEYS_TRANSIENT_KEY = 'wcstripe_test_invalid_api_keys_detected';
+
+	/**
+	 * The live mode invalid API keys transient key.
+	 *
+	 * @var string
+	 */
+	const LIVE_MODE_INVALID_API_KEYS_TRANSIENT_KEY = 'wcstripe_live_invalid_api_keys_detected';
+
+	/**
 	 * Secret API Key.
 	 *
 	 * @var string
@@ -233,6 +247,13 @@ class WC_Stripe_API {
 	public static function retrieve( $api ) {
 		WC_Stripe_Logger::log( "{$api}" );
 
+		// If we have a transient indicating that the secret key is not valid, we dont't attempt the API call and we return an error.
+		$transient_key = WC_Stripe_Mode::is_test() ? self::TEST_MODE_INVALID_API_KEYS_TRANSIENT_KEY : self::LIVE_MODE_INVALID_API_KEYS_TRANSIENT_KEY;
+		$invalid_api_keys_detected = get_transient( $transient_key );
+		if ( $invalid_api_keys_detected ) {
+			return new WP_Error( 'stripe_error', __( 'The Stripe API keys are not valid.', 'woocommerce-gateway-stripe' ) );
+		}
+
 		$response = wp_safe_remote_get(
 			self::ENDPOINT . $api,
 			[
@@ -241,6 +262,13 @@ class WC_Stripe_API {
 				'timeout' => 70,
 			]
 		);
+
+		// If we get a 401 error, we know the secret key is not valid, we save a transient to avoid making calls until the secrect key gets updated.
+		if ( ! empty( $response['response']['code'] ) && 401 === $response['response']['code'] ) {
+			$transient_key = WC_Stripe_Mode::is_test() ? self::TEST_MODE_INVALID_API_KEYS_TRANSIENT_KEY : self::LIVE_MODE_INVALID_API_KEYS_TRANSIENT_KEY;
+			set_transient( $transient_key, true );
+			return new WP_Error( 'stripe_error', __( 'The Stripe API keys are not valid.', 'woocommerce-gateway-stripe' ) );
+		}
 
 		if ( is_wp_error( $response ) || empty( $response['body'] ) ) {
 			WC_Stripe_Logger::log( 'Error Response: ' . print_r( $response, true ) );
--- a/includes/class-wc-stripe-api.php
+++ b/includes/class-wc-stripe-api.php
@@ -17,14 +17,14 @@ class WC_Stripe_API {
 	const STRIPE_API_VERSION = '2024-06-20';
 
 	/**
-	 * The test mode invalid API keys transient key.
+	 * The test mode invalid API keys option key.
 	 *
 	 * @var string
 	 */
 	const TEST_MODE_INVALID_API_KEYS_OPTION_KEY = 'wc_stripe_test_invalid_api_keys_detected';
 
 	/**
-	 * The live mode invalid API keys transient key.
+	 * The live mode invalid API keys option key.
 	 *
 	 * @var string
 	 */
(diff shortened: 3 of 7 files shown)


includes/class-wc-stripe-api.php
```diff
@@ -233,6 +247,13 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
 	public static function retrieve( $api ) {
 		WC_Stripe_Logger::log( "{$api}" );
 
+		// If we have a transient indicating that the secret key is not valid, we dont't attempt the API call and we return an error.
+		$transient_key = WC_Stripe_Mode::is_test() ? self::TEST_MODE_INVALID_API_KEYS_TRANSIENT_KEY : self::LIVE_MODE_INVALID_API_KEYS_TRANSIENT_KEY;
+		$invalid_api_keys_detected = get_transient( $transient_key );
+		if ( $invalid_api_keys_detected ) {
+			return new WP_Error( 'stripe_error', __( 'The Stripe API keys are not valid.', 'woocommerce-gateway-stripe' ) );
+		}
+
```
daledupreez: I think we should do this check before we trigger the logger, otherwise we're going to flood the logs.
```suggestion
		// If we have a transient indicating that the secret key is not valid, we don't attempt the API call and we return an error.
		$transient_key = WC_Stripe_Mode::is_test() ? self::TEST_MODE_INVALID_API_KEYS_TRANSIENT_KEY : self::LIVE_MODE_INVALID_API_KEYS_TRANSIENT_KEY;
		$invalid_api_keys_detected = get_transient( $transient_key );
		if ( $invalid_api_keys_detected ) {
			return new WP_Error( 'stripe_error', __( 'The Stripe API keys are not valid.', 'woocommerce-gateway-stripe' ) );
		}

		WC_Stripe_Logger::log( "{$api}" );

```

includes/class-wc-stripe-api.php
```diff
@@ -242,6 +263,13 @@ public static function retrieve( $api ) {
 			]
 		);
 
+		// If we get a 401 error, we know the secret key is not valid, we save a transient to avoid making calls until the secrect key gets updated.
+		if ( ! empty( $response['response']['code'] ) && 401 === $response['response']['code'] ) {
```
daledupreez: We should guard against `$response` being a `WP_Error` instance.
```suggestion
		if ( is_array( $response ) && ! empty( $response['response']['code'] ) && 401 === $response['response']['code'] ) {
```

includes/class-wc-stripe-api.php
```diff
@@ -242,6 +263,13 @@ public static function retrieve( $api ) {
...
+			$transient_key = WC_Stripe_Mode::is_test() ? self::TEST_MODE_INVALID_API_KEYS_TRANSIENT_KEY : self::LIVE_MODE_INVALID_API_KEYS_TRANSIENT_KEY;
```
daledupreez: We should still have the local variable from the top of the function available. Maybe we can call it `$invalid_key_transient_key` and re-use the value from above?

includes/class-wc-stripe-api.php
```diff
@@ -242,6 +263,13 @@ public static function retrieve( $api ) {
...
+			set_transient( $transient_key, true );
```
daledupreez: Do we want to do this immediately in all cases? Or only after we've triggered multiple failures in a short period of time?

I worry that there may be other situations where one or two 401 responses could occur -- I wouldn't want to stop a merchant's payments if that occurs.

Also, do we want to record the timestamp that we took the action? That may be useful when debugging and trying to understand what is going on.

daledupreez (commented):
I think this is the general approach should work well, with a concern about triggering the code after one 401 response.

More broadly, I think we also need to clear the transient when we save/update the Stripe keys, which feels like it ought to occur in `WC_Stripe_Connect->save_stripe_keys()` and whatever code we use to disconnect/clear keys.

includes/class-wc-stripe-api.php
```diff
@@ -242,6 +263,12 @@ public static function retrieve( $api ) {
 			]
 		);
 
+		// If we get a 401 error, we know the secret key is not valid, we save a transient to avoid making calls until the secrect key gets updated.
+		if ( is_array( $response ) && ! empty( $response['response']['code'] ) && 401 === $response['response']['code'] ) {
+			set_transient( $invalid_api_keys_transient_key, true );
+			return json_decode( '' ); // The UI expects this empty response in case of invalid API keys.
```
daledupreez: I believe `json_decode( '' ) === null`. Can we just return `null` instead?

includes/class-wc-stripe-api.php
```diff
@@ -231,6 +245,13 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
 	 * @param string $api
 	 */
 	public static function retrieve( $api ) {
+		// If we have an option flag indicating that the secret key is not valid, we dont't attempt the API call and we return an error.
```
daledupreez: Nit: typo
```suggestion
		// If we have an option flag indicating that the secret key is not valid, we don't attempt the API call and we return an error.
```

includes/class-wc-stripe-api.php
```diff
@@ -242,6 +263,18 @@ public static function retrieve( $api ) {
 			]
 		);
 
+		// If we get a 401 error, we know the secret key is not valid.
+		if ( is_array( $response ) && ! empty( $response['response']['code'] ) && 401 === $response['response']['code'] ) {
```
daledupreez: Minor thought: my code in #4327 is extremely defensive in this check, just to make sure that we have all the data we're expecting. Not blocking, but it may be worth being super-defensive to make sure we don't break in unexpected ways.
```suggestion
		if ( is_array( $response ) && isset( $response['response'] ) && is_array( $response['response'] ) && isset( $response['response']['code'] ) && 401 === $response['response']['code'] ) {
```

changelog.txt
```diff
@@ -7,6 +7,7 @@
 * Update - Remove BACS from the unsupported 'change payment method for subscription' page.
 * Fix - Fix payment method title display when new payment settings experience is enabled
 * Fix - Prevent styles from non-checkout pages affecting the appearance of Stripe element.
+* Fix - Prevert further Stripe API calls if API keys are invalid (401 response).
```
daledupreez: Nit: typo
```suggestion
* Fix - Prevent further Stripe API calls if API keys are invalid (401 response).
```

readme.txt
```diff
@@ -118,6 +118,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Update - Remove BACS from the unsupported 'change payment method for subscription' page.
 * Fix - Fix payment method title display when new payment settings experience is enabled
 * Fix - Prevent styles from non-checkout pages affecting the appearance of Stripe element.
+* Fix - Prevert further Stripe API calls if API keys are invalid (401 response).
```
daledupreez: Typo
```suggestion
* Fix - Prevent further Stripe API calls if API keys are invalid (401 response).
```

client/settings/account-details/index.js
```diff
@@ -133,11 +133,11 @@ const AccountDetails = () => {
 					{ createInterpolateElement(
 						isTestModeEnabled
 							? __(
-									"Seems like the test API keys we've saved for you are no longer valid. If you recently updated them, use the <strong>Configure Connection</strong> button below to reconnect.",
+									"We couldn't connect to your account, it seems like the test API keys we've saved for you are no longer valid. Please use the <strong>Configure connection</strong> button below to reconnect.",
```
ebinnion, line 136: Not urgent in this case. But, will this break internationalization of this string in the patch release?

daledupreez: Good call.

daledupreez (commented):
Overall, this is looking good and tests as described.

I do think we need to fix the typos in the changelogs, and I had one minor comment about possibly making the path to identify the response code more defensive.

client/settings/account-details/index.js
daledupreez: That said, I think the trade-off for correctly communicating the status is probably worth it.

includes/class-wc-stripe-api.php
```diff
@@ -242,6 +263,18 @@ public static function retrieve( $api ) {
 			]
 		);
 
+		// If we get a 401 error, we know the secret key is not valid.
```
wjrosa, line 266: Maybe we could consider later changing this to check for anything different from 200, to cover more possible issues.

client/settings/account-details/index.js
diegocurbelo: Correct, new strings need to be translated in the community translations site: https://translate.wordpress.org/projects/wp-plugins/woocommerce-gateway-stripe/

includes/class-wc-stripe-api.php
daledupreez: My proposed changes in #4327 take that approach when looking at responses. 😁

daledupreez (approved):
Thanks for the cleanup, @diegocurbelo! 🚀

includes/class-wc-stripe-api.php
malithsen: @diegocurbelo @daledupreez Did we address this concern somewhere? I wonder if one-off 401s are contributing to [this potential false positive issue](https://wordpress.org/support/topic/version-9-5-1-falsely-thinks-account-is-disconnected/) a user mentioned in forums

diegocurbelo: API Keys can not be temporary disabled, only revoked; so getting one 401 means that we need new keys (there is no way for them to become valid again).

The user in the forum mentions that payments are still coming through, so we are not blocking API calls because of a 401 (the 401 flag is not set). It seems to have been a problem with the account cache; if the account can't be read (because of a timeout or a 500, 429 errors etc) the UI assumes the account is disconnected and show the notices.

We show refactor the account settings (and global notices) to account for the new flows we added to the Stripe API connector (which might cause the account/PM list to be null temporary), but that might be to big of a change for a patch release.

malithsen: That makes more sense. Thanks Diego! I see that it was resolved after clearing cache/transients.

Issue #4317, opened 2025-05-15
Prevent runaway request failures when we have Stripe API issues


daledupreez:
### Description

This issue is related to https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4315, but not quite the same. In that issue, we specifically tackle the issue where our keys are invalid. This issue is aimed at addressing a slightly more general case: what happens if we're getting other repeated errors from the Stripe API? We should try to avoid unnecessary API calls when we can detect that there is a problem with our API requests to Stripe. The problem is that we end up making plenty of requests to Stripe that slow the site down, and we don't have a way to limit the number of calls that we can reasonably expect to fail.

### Steps to reproduce

We don't have specific steps to reproduce this issue specifically, as we want to tackle key invalidation in https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4315. But we should be able to replicate a similar situation by using those same steps, and then disabling any handling specific to HTTP 401 errors.

Those steps are:
1. If your site has a connection to Stripe, disconnect it
2. Connect (or reconnect) to Stripe in live mode
3. Switch to test mode
4. See repeated API calls to Stripe

### Expected behavior

When we are having issues communicating with Stripe, we should apply some level of rate limiting that prevents us from making API calls for a short amount of time.

daledupreez:
[This comment](https://wordpress.org/support/topic/very-slow-backend-50-calls-to-the-stripe-api/#post-18467542) in one of our forum threads specifically shows a number of 429 errors being returned from Stripe. We need to handle those gracefully, as we need to detect when we're being rate-limited, and correctly stop calling the API when that's the case.

PR #4327, opened 2025-05-16
Enforce plugin rate limit when Stripe returns 429 errors

daledupreez:
Relates to STRIPE-445
Relates to:
 * https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4317
 * https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4323

This PR implements some post-request checks to detect when we receive an HTTP 429 response from Stripe, and then impose a 30 second period where we won't make further `GET` requests via `WC_Stripe_API::retrieve()`. When we enter into a rate-limited mode, we'll take the following actions:
 * Log an error via `WC_Stripe_Logger`
 * Log an error to the PHP error log via `error_log()`
 * Add the expiry time for the current rate limit to the respective `wc_stripe_live_api_rate_limit` or `wc_stripe_test_api_rate_limit` option
 * Add details for the current rate limit period to the appropriate `wc_stripe_live_api_rate_limit_history` or `wc_stripe_test_api_rate_limit_history` option (but we'll only retain 20 of these to avoid unnecessary option size).

When `WC_Stripe_API::retrieve()` is called while the rate limit is in effect, we'll return a `null` response immediately, without logging anything extra.

When `WC_Stripe_API::retrieve()` is called after the rate limit expires, we'll remove the previous rate limit entry from the appropriate option, and proceed.

There are some possible nuances we may want to handle differently:
 * I am not yet pushing up a notice when this occurs -- I think we should add one
 * The logic for detecting error responses will now detect all 4xx and 5xx response codes, where we would only treat responses with an empty body as an error before
 * The rate limit period is only 30 seconds - this may need to be shorter or longer
 * I don't know how useful the history is/will be, but I think it's a good idea
   - This may be useful for triggering an email after 2 rate limits within a period of time, for example.
 * The logic has a lot of overlap with the work @diegocurbelo is doing in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4323 - we will need to confirm our approach

## Testing instructions

* Create a build using the code from this branch, install it on a site, and ensure that you have a Stripe connection configured
* Add a plugin or tool that allows you to run snippets of PHP code
* Set up the following code to run on your site - it modifies an HTTP response from Stripe to simulate an HTTP 429 response:
```php
function simulate_http_429_from_stripe_api( $response, $parsed_args, $url ) {
	if ( ! str_starts_with( $url, 'https://api.stripe.com/v1/' ) ) {
		return $response;
	}
	
	if ( is_array( $response ) && isset( $response['response'] ) && is_array( $response['response'] ) && isset( $response['response']['code'] ) ) {
		$response['response']['code'] = 429;
	}
	
	return $response;
}

add_filter( 'http_response', 'simulate_http_429_from_stripe_api', 10, 3 );
```
* Navigate to _WooCommerce_ > _Settings_ > _Payments_, and select the Stripe option from the list
* Navigate to the Payment Methods page
* Confirm that no items in the 
* Use WP-CLI or some other tool to verify that the `wc_stripe_test_api_rate_limit` option has a value (or the `_live_` equivalent) has a value
  - WP-CLI command: `wp option get wc_stripe_test_api_rate_limit`
* Use WP-CLI or some other tool to verify that the `wp option get wc_stripe_test_api_rate_limit_history` (or the `_live_` equivalent) has readable data

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-api.php
```diff
@@ -250,6 +278,80 @@ public static function retrieve( $api ) {
...
+	 * @return void
+	 */
+	protected static function check_stripe_api_error_response( $response ) {
+		// If we don't have an array for $response, return early, as we won't have an HTTP status code.
+		if ( ! is_array( $response ) ) {
+			return;
+		}
+
+		// We specifically want to check $response['response']['code']. If it's not present, return early.
+		if ( ! isset( $response['response'] ) || ! is_array( $response['response'] ) || ! isset( $response['response']['code'] ) ) {
```
annemirasol: Nit, only for readability: I think it's sufficient to just check `! isset( $response['response']['code'] )` here?

```
❯ php -a
Interactive shell

php > $a = [];
php > echo $a['b']['c'];
PHP Warning:  Undefined array key "b" in php shell code on line 1

Warning: Undefined array key "b" in php shell code on line 1
PHP Warning:  Trying to access array offset on null in php shell code on line 1

Warning: Trying to access array offset on null in php shell code on line 1
php > var_dump( isset( $a['b']['c'] ) );
bool(false)
```

includes/class-wc-stripe-api.php
```diff
@@ -250,6 +278,80 @@ public static function retrieve( $api ) {
...
+			$is_test_mode = WC_Stripe_Mode::is_test();
+
+			$timestamp = time();
+			$rate_limit_option_key = $is_test_mode ? self::TEST_MODE_STRIPE_API_RATE_LIMIT_OPTION_KEY : self::LIVE_MODE_STRIPE_API_RATE_LIMIT_OPTION_KEY;
+			update_option( $rate_limit_option_key, $timestamp + self::STRIPE_API_RATE_LIMIT_DURATION );
+
+			$mode = $is_test_mode ? 'test' : 'LIVE';
+			$message = "Stripe {$mode} mode API has been rate limited, disabling API calls for " . self::STRIPE_API_RATE_LIMIT_DURATION . ' seconds.';
+
+			error_log( 'woocommerce-gateway-stripe: WARNING: ' . $message );
```
annemirasol, line 338: Did we mean to have a separate `error_log()` here, in addition to `WC_Stripe_Logger::error( $message );` in the following line?

includes/class-wc-stripe-api.php
```diff
@@ -16,6 +16,28 @@ class WC_Stripe_API {
...
+	 * @var string
+	 */
+	public const TEST_MODE_STRIPE_API_RATE_LIMIT_OPTION_KEY = 'wc_stripe_test_api_rate_limit';
+
+	/**
+	 * Duration we will use to disable Stripe API calls if we have been rate limited.
+	 *
+	 * @var int
+	 */
+	public const STRIPE_API_RATE_LIMIT_DURATION = 30;
```
annemirasol: Is this seconds? Or minutes?

At first I thought it would be minutes, but from the code below, it turned out to be seconds.

includes/class-wc-stripe-api.php
```diff
@@ -250,6 +278,80 @@ public static function retrieve( $api ) {
...
+
+		$status_code = $response['response']['code'];
+
+		if ( 429 === $status_code ) {
+			// Stripe has rate limited us, so disable API calls for a period of time.
+			$is_test_mode = WC_Stripe_Mode::is_test();
+
+			$timestamp = time();
+			$rate_limit_option_key = $is_test_mode ? self::TEST_MODE_STRIPE_API_RATE_LIMIT_OPTION_KEY : self::LIVE_MODE_STRIPE_API_RATE_LIMIT_OPTION_KEY;
+			update_option( $rate_limit_option_key, $timestamp + self::STRIPE_API_RATE_LIMIT_DURATION );
```
annemirasol: See question about seconds vs minutes above.

includes/class-wc-stripe-api.php
daledupreez: We could also split it into multiple lines, but the issue is if `$response['response']` is an object, even if we use `isset()`.
```
$ php -a
Interactive shell

php > $b = (object) [ 'c' => 'c' ]; $a = [ 'b' => $b ];
php > echo isset( $a['b']['c'] );
PHP Warning:  Uncaught Error: Cannot use object of type stdClass as array in php shell code:1
Stack trace:
#0 {main}
  thrown in php shell code on line 1

Warning: Uncaught Error: Cannot use object of type stdClass as array in php shell code:1
Stack trace:
#0 {main}
  thrown in php shell code on line 1
```

includes/class-wc-stripe-api.php
daledupreez: Yes. I want to make sure that we log _something_ even if logging is disabled at the plugin level -- someone needs to know about this no matter what. And I checked `WC_Stripe_Logger::error()` won't log if plugin-level logging is not enabled.

includes/class-wc-stripe-api.php
daledupreez: I've split this into a multiline OR condition to help with the readability in 4e638b87.

includes/class-wc-stripe-api.php
daledupreez: That is fair. If you had to think twice, the name needs to be better! I've added `_IN_SECONDS` to the constant name in 84e5af2 so it's super-explicit.

includes/class-wc-stripe-api.php
```diff
@@ -250,6 +278,84 @@ public static function retrieve( $api ) {
...
+			! isset( $response['response'] )
+			|| ! is_array( $response['response'] )
+			|| ! isset( $response['response']['code'] )
+		) {
+			return;
+		}
+
+		$status_code = $response['response']['code'];
+
+		if ( 429 === $status_code ) {
```
wjrosa, line 327: Based on https://docs.stripe.com/rate-limits#rate-limited-requests, it is better to check for the header. It is possible that a 429 may not be related to rate-limiting. Both work fine, but checking the header would be more consistent.

daledupreez:
Closing this PR as stale.

PR #4330, opened 2025-05-16
Adding a fetch cooldown to PM config retrieval endpoint calls

wjrosa:
See p1747425481186179-slack-C08SUDB2YKC?thread_ts=1747424546.927019&cid=C08SUDB2YKC

This PR adds a fetch cooldown (handled using a new option) to the calls we do to retrieve the merchant's payment method configuration. This mitigates possible rate-limiting issues, blocking requests for a while (1 minute). It was suggested by @ebinnion .

## Testing instructions

- Verify that `/payment_method_configurations` is not called more than once per minute, e.g. add debugging code inside WC_Stripe_API::retrieve(), and repeatedly refreshing checkout page or settings page
- Simulate unreliable transient cache by force returning null or setting a super short expiry, and verify that a) still no excessive API calls and b) there are no unexpected behavior with enabled payment methods

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

cc @ebinnion @woocommerce/quark

### DIFF
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -78,7 +78,7 @@ class WC_Stripe_Payment_Method_Configurations {
 			return self::$primary_configuration;
 		}
 
-		$cache_key = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
+		$cache_key                    = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
 		$cached_primary_configuration = get_transient( $cache_key );
 		if ( false === $cached_primary_configuration || null === $cached_primary_configuration ) {
 			return null;
@@ -114,10 +114,17 @@ class WC_Stripe_Payment_Method_Configurations {
 	 * @return object|null
 	 */
 	private static function get_payment_method_configuration_from_stripe() {
+		// Only allow fetching payment configuration once per minute.
+		$fetch_cooldown = get_option( 'wcstripe_payment_method_config_fetch_cooldown', 0 );
+		if ( $fetch_cooldown > time() ) {
+			return null;
+		}
+
 		$result         = WC_Stripe_API::get_instance()->get_payment_method_configurations();
 		$configurations = $result->data ?? null;
 
 		if ( ! $configurations ) {
+			update_option( 'wcstripe_payment_method_config_fetch_cooldown', time() + MINUTE_IN_SECONDS );
 			return null;
 		}
 
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -66,16 +66,13 @@ class WC_Stripe_Payment_Method_Configurations {
 				return $cached_primary_configuration;
 			}
 
-			// If we are in cooldown and cache is not available, return fallback configuration
-			// using legacy settings.
-			$pmc_configuration_from_legacy_settings = self::get_fallback_pmc_configuration();
-			if ( ! empty( $pmc_configuration_from_legacy_settings ) ) {
-				error_log( 'Returning PMC configuration from legacy settings: ' . print_r( $pmc_configuration_from_legacy_settings, true ) );
-				return $pmc_configuration_from_legacy_settings;
+			$fallback_cache = self::get_payment_method_configuration_from_cache( true );
+			if ( $fallback_cache ) {
+				return $fallback_cache;
 			}
 
 			// Final fallback: return minimal card configuration.
-			return self::get_fallback_pmc_configuration( true );
+			return self::get_minimal_card_config();
 		}
 
 		update_option( 'wcstripe_payment_method_config_fetch_cooldown', time() + MINUTE_IN_SECONDS );
@@ -105,42 +102,11 @@ class WC_Stripe_Payment_Method_Configurations {
 	}
 
 	/**
-	 * Get fallback payment method configuration, using legacy settings.
-	 *
-	 * @param bool $card_only Whether to enable only the card payment method.
-	 * @return array
-	 */
-	private static function get_fallback_pmc_configuration( $card_only = false ) {
-		$stripe_settings   = WC_Stripe_Helper::get_stripe_settings();
-		$pmc_configuration = [];
-
-		foreach ( WC_Stripe_UPE_Payment_Gateway::UPE_AVAILABLE_METHODS as $payment_method_id => $payment_method_name ) {
-			$is_enabled = $card_only ?
-				WC_Stripe_Payment_Methods::CARD === $payment_method_id :
-				! empty( $stripe_settings['upe_checkout_experience_accepted_payments'] ) &&
-					is_array( $stripe_settings['upe_checkout_experience_accepted_payments'] ) &&
-					in_array( $payment_method_id, $stripe_settings['upe_checkout_experience_accepted_payments'], true );
-
-			$pmc_configuration[ $payment_method_id ] = (object) [
-				'parent'             => WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_PARENT_ID : self::LIVE_MODE_CONFIGURATION_PARENT_ID,
-				'available'          => $is_enabled,
-				'display_preference' => (object) [
-					'overridable' => null,
-					'preference'  => $is_enabled ? 'on' : 'off',
-					'value'       => $is_enabled ? 'on' : 'off',
-				],
-			];
-		}
-
-		return $pmc_configuration;
-	}
-
-	/**
 	 * Get the payment method configuration from cache.
 	 *
 	 * @return object|null
 	 */
-	private static function get_payment_method_configuration_from_cache() {
+	private static function get_payment_method_configuration_from_cache( $use_fallback = false ) {
 		if ( null !== self::$primary_configuration ) {
 			return self::$primary_configuration;
 		}
@@ -148,6 +114,9 @@ class WC_Stripe_Payment_Method_Configurations {
 		$cache_key                    = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
 		$cached_primary_configuration = get_transient( $cache_key );
 		if ( false === $cached_primary_configuration || null === $cached_primary_configuration ) {
+			if ( $use_fallback ) {
+				return get_option( $cache_key );
+			}
 			return null;
 		}
 
@@ -162,6 +131,7 @@ class WC_Stripe_Payment_Method_Configurations {
 		self::$primary_configuration = null;
 		$cache_key                   = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
 		delete_transient( $cache_key );
+		delete_option( $cache_key );
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -114,10 +114,17 @@ private static function set_payment_method_configuration_cache( $configuration )
...
+		$fetch_cooldown = get_option( 'wcstripe_payment_method_config_fetch_cooldown', 0 );
+		if ( $fetch_cooldown > time() ) {
+			return null;
+		}
+
 		$result         = WC_Stripe_API::get_instance()->get_payment_method_configurations();
 		$configurations = $result->data ?? null;
 
 		if ( ! $configurations ) {
+			update_option( 'wcstripe_payment_method_config_fetch_cooldown', time() + MINUTE_IN_SECONDS );
```
malithsen: We set the cooldown when there are no configurations (Stripe limit is already reached). Instead should we always set the cooldown to prevent multiple calls within a minute?

wjrosa: I think Eric and Anne had the same concern over Slack. Can you check the latest changes they suggested, please? I think that may fix the issue

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -197,6 +258,13 @@ public static function get_upe_enabled_payment_method_ids( $force_refresh = fals
 			}
 		}
 
+		// Save in options, so we can use this as a fallback.
+		if ( $force_refresh ) {
+			$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
+			$stripe_settings['upe_checkout_experience_accepted_payments'] = $enabled_payment_method_ids;
+			WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
+		}
+
```
annemirasol: Not happy with this. This means we only get a copy when `$force_refresh` is true, ie. when user visits the Stripe settings page.

If we remove the check though, we will do it far too often.

annemirasol: Forgot to include express checkout methods too -- Apple Pay, Google Pay, Link.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -58,16 +58,83 @@ class WC_Stripe_Payment_Method_Configurations {
...
 			$cached_primary_configuration = self::get_payment_method_configuration_from_cache();
 			if ( $cached_primary_configuration ) {
 				return $cached_primary_configuration;
 			}
+
+			// If we are in cooldown and cache is not available, return fallback configuration
+			// using legacy settings.
+			$pmc_configuration_from_legacy_settings = self::get_fallback_pmc_configuration();
+			if ( ! empty( $pmc_configuration_from_legacy_settings ) ) {
+				error_log( 'Returning PMC configuration from legacy settings: ' . print_r( $pmc_configuration_from_legacy_settings, true ) );
```
ebinnion: Is this intended to be here?

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -115,11 +155,7 @@ private static function set_payment_method_configuration_cache( $configuration )
 	 */
 	private static function get_payment_method_configuration_from_stripe() {
 		$result         = WC_Stripe_API::get_instance()->get_payment_method_configurations();
-		$configurations = $result->data ?? null;
-
-		if ( ! $configurations ) {
-			return null;
-		}
```
annemirasol, line 122: I don't know why this is removed. Consider restoring this.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -58,29 +58,65 @@ class WC_Stripe_Payment_Method_Configurations {
 	 * @return object|null
 	 */
 	private static function get_primary_configuration( $force_refresh = false ) {
-		if ( ! $force_refresh ) {
+		// Only allow fetching payment configuration once per minute.
+		$fetch_cooldown = defined( 'IS_TESTING' ) ? 0 : get_option( 'wcstripe_payment_method_config_fetch_cooldown', 0 );
+		if ( ! $force_refresh && $fetch_cooldown > time() ) {
```
ebinnion: I think that we should use `||` here. I understand that this may result in the admin page being a bit stale. But:

1. We can find ways to bust this cache with other methods.
2. Without it, do we have any rate limiting of the requests here? So, what happens if someone's browser is just sitting there making a ton of  /wp-json/wc/v3/stripe/settings.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -115,11 +155,7 @@ private static function set_payment_method_configuration_cache( $configuration )
 	 */
 	private static function get_payment_method_configuration_from_stripe() {
 		$result         = WC_Stripe_API::get_instance()->get_payment_method_configurations();
-		$configurations = $result->data ?? null;
-
-		if ( ! $configurations ) {
```
malithsen, line 120: Any reason for removing this check?

Without this, if we don't get configurations, it will throw a warning when in foreach below

```
foreach() argument must be of type array|object, null given 
```

annemirasol: I don't know why it was removed, but yeah we should restore

malithsen: nvm. Misread the change, I see we're falling back to an array. All good.

annemirasol: @malithsen I'm seeing they (maybe Wesley?) changed it to an array fallback

`$configurations = $result->data ?? [];`

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -58,29 +58,65 @@ class WC_Stripe_Payment_Method_Configurations {
...
+					'value'       => 'on',
+				],
+			],
+		];
+	}
+
 	/**
 	 * Get the payment method configuration from cache.
 	 *
 	 * @return object|null
```
ebinnion, line 99: Minor nit: Need to update docblock now.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -58,29 +58,71 @@ class WC_Stripe_Payment_Method_Configurations {
...
+			// If we are hitting the API too much, and our main cache is not working, use the fallback cache.
+			if ( $is_in_cooldown ) {
+				$fallback_cache = self::get_payment_method_configuration_from_cache( true );
+				if ( $fallback_cache ) {
+					return $fallback_cache;
+				}
+
+				// If even the fallback cache is not available, return minimal card configuration.
+				return self::get_minimal_card_config();
+			}
```
annemirasol: I updated this to wrap the fallback logic in `if ( $is_in_cooldown )` because,

1. frontend may trigger call to API, i.e. `$force_refresh` is false
2. transient may already be expired, i.e. `$cached_primary_configuration` is empty
3. if we are not in cooldown, i.e. hitting the API too much, we want to retrieve a fresh PMC and not use stale fallbacks

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -58,29 +58,47 @@ class WC_Stripe_Payment_Method_Configurations {
 	 * @return object|null
 	 */
 	private static function get_primary_configuration( $force_refresh = false ) {
-		if ( ! $force_refresh ) {
+		// Only allow fetching payment configuration once per minute.
+		$fetch_cooldown = defined( 'IS_TESTING' ) ? 0 : get_option( 'wcstripe_payment_method_config_fetch_cooldown', 0 );
```
daledupreez: Nit: we should be checking that it is defined and true
```suggestion
		$fetch_cooldown = defined( 'WC_STRIPE_IS_TESTING' ) && WC_STRIPE_IS_TESTING ? 0 : get_option( 'wcstripe_payment_method_config_fetch_cooldown', 0 );
```

phpunit.xml.dist
```diff
@@ -25,5 +25,7 @@
 			<html outputDirectory="phpunit-html"/>
 		</report>
 	</coverage>
-
+	<php>
+		<const name="IS_TESTING" value="1"/>
```
daledupreez: It feels like we should use a more distinct constant name to prevent collisions in the wild. `IS_TESTING` is very generic, and could be set for many other reasons.
```suggestion
		<const name="WC_STRIPE_IS_TESTING" value="1"/>
```

daledupreez: Thinking on this more, I think we should aim to rely on constants set by the WordPress Core test setup/bootstrap code. Based on the code section below, I think we might want to rely on `if ( defined( 'WP_TESTS_TITLE' ) && defined( 'WP_TESTS_EMAIL' ) )` in our code, rather than a custom/local env variable. This is especially true for cases where we end up running alongside other plugins' unit tests. Maybe we should use both using OR?

https://github.com/WordPress/wordpress-develop/blob/8b466fce082b5b8731cebca85469ae1978318504/tests/phpunit/includes/bootstrap.php#L189-L201

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -58,29 +58,47 @@ class WC_Stripe_Payment_Method_Configurations {
...
 				return $cached_primary_configuration;
 			}
+
+			// If we are hitting the API too much, and our main cache is not working, use the fallback cache.
+			if ( $is_in_cooldown ) {
+				$fallback_cache = self::get_payment_method_configuration_from_cache( true );
+				if ( $fallback_cache ) {
+					return $fallback_cache;
+				}
+			}
```
daledupreez, line 81: I think it's worth adding an explicit comment to explain that we are intentionally falling through here, and ~what we expect to happen~ why we are doing so. _(Edited)_
```suggestion
			}
			// Intentionally fall through to fetching the data from Stripe if we don't have it locally, 
			// even when $force_refresh = false and/or $is_in_cooldown is true.
			// We _need_ the payment method configuration for things to work as expected, 
			// so we will fetch it if we don't have anything locally.
```

daledupreez (commented):
I have some minor feedback from inspection, mostly around the constant we're using for detecting the test environment, and adding some comments to the code.

I will only get a chance to fully test this a bit later today.

wjrosa:
> I have some minor feedback from inspection, mostly around the constant we're using for detecting the test environment, and adding some comments to the code.
> 
> I will only get a chance to fully test this a bit later today.

@daledupreez Yeah, I did introduce the constant to identify the test environment. I am aware the code should be agnostic of the unit tests (I said that on Slack). I suggested we revisit this next week and remove it

includes/class-wc-stripe-payment-method-configurations.php
wjrosa: Yes, that was me. At least for me, when I run the tests without this change I get the error Malith mentioned:
```
foreach() argument must be of type array|object, null given
```
I kept the change just because I think it makes sense type-wise. But feel free to remove it if you want

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -104,8 +133,11 @@ public static function clear_payment_method_configuration_cache() {
 	 */
 	private static function set_payment_method_configuration_cache( $configuration ) {
 		self::$primary_configuration = $configuration;
-		$cache_key = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
+		$cache_key                   = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
 		set_transient( $cache_key, $configuration, self::CONFIGURATION_CACHE_TRANSIENT_EXPIRATION );
+
+		// To be used as fallback if we are in API cooldown and the transient is not available.
+		update_option( $cache_key, $configuration );
```
annemirasol, line 140: Should we set `$autoload` to `false` here?

annemirasol: `strlen( serialize( $configuration ) ) )` is `5816`

wjrosa: I think so 👍 . I understand that the `true` value is recommended for values used on the frontend.

annemirasol: We may end up using it for frontend calls though, e.g. checkout. 

> Accepts a boolean, or null to stick with the initial value or, if no initial value is set, to leave the decision up to default heuristics in WordPress.

I think we can leave it be.

annemirasol (approved):

Issue #4182, opened 2025-04-02
Settings Synchronization: Cache enabled payment methods for non-admin pages


hsingyuc:
We fetch the Stripe payment method configuration on every frontend page, which can potentially generate many requests to the configuration API. We can update the cache when we fetch new data on the admin pages.

Dev: [We can set transient here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/ce62542bf4d4a686415a7a2843defd4193fa07ea/includes/class-wc-stripe-payment-method-configurations.php#L75C2-L88C3)

Discussion [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4091#discussion_r2014959167)

PR #4331, opened 2025-05-16
Implement custom database cache

### LINKED ISSUE #4247: Repeated payment configuration API calls slow down site
**Describe the bug**

Version 9.4.0 of the WooCommerce Stripe payment gateway causes lots of API calls to Stripe, slowing down the store and causing other problems, including cases where no payment methods would be shown as available. The core problem was described at a high level in https://developer.woocommerce.com/2025/04/18/developer-advisory-stripe-payment-gateway-9-4-1-release/.

**To Reproduce**

Ensure you are running version `9.4.0` - the changes were rolled back in `9.4.1`, so the specific version is needed.

Add the following code to `WC_Stripe_API::get_all_payment_method_configurations()` on line 473 in `includes/class-wc-stripe-api.php`:
```php
		WC_Stripe_Logger::log( 'Calling get_all_payment_method_configurations' );
```

1. Ensure you are using the new checkout method, and not legacy checkout.
2. Add one or more products to your cart
3. Navigate to checkout
4. In a separate, logged-in window, navigate to _WooCommerce_ -> _Status_ -> _Logs_ and open the  log for `woocommerce-gateway-stripe` with the current date
5. Reload the checkout page.
6. Verify that we triggered the log message you added above at least once by refreshing the log page
7. Reload the checkout page again.
8. Verify that we triggered the log message you added above at least one more time by refreshing the log page.

**Expected behavior**

We should only fetch the updated configuration when needed, either when configuring payment methods in the settings page, or when we don't have the details locally yet.

 - WordPress Version - 6.7.2
 - WooCommerce Version - 9.8.1
 - Stripe Plugin Version - 9.4.0

**Additional context**

This issue was flagged in multiple support forum threads after the release, including the following:
 * https://wordpress.org/support/topic/9-4-0-no-payment-appearing-on-checkout-breaks-site/
 * https://wordpress.org/support/topic/version-9-4-0-update-critcal-error-for-wordpress-6-8/
 * https://wordpress.org/support/topic/sorry-it-seems-that-there-are-no-available-payment-methods-after-update-9-4-0/
 * https://wordpress.org/support/topic/9-4-0-critical-error/
 * https://wordpress.org/support/topic/urgent-credit-card-payment-does-not-appear-anymore-on-my-checkout-after-update/

diegocurbelo:
Fixes #4247, #4182

This PR implements a custom database cache layer for persistent caching with in-memory optimization. It's based on the WooPayments Database Cache implementation class.

It's intended as a drop-in replacement of the current transient-based implementation; future PRs will move the live/test logic to the cache layer.

NOTE: It keeps the Fetch cooldown (but adds a key suffix) for now; it will be replaced with a 5-minute cache for admin pages in a future PR.

## Testing instructions

> To easily test the cache edit the file includes/class-wc-stripe-api.php add a WC_Stripe_Logger::log( 'get_payment_method_configurations' ) to the function get_payment_method_configurations (line 505).
>
> Or add a breakpoint on line 506 and enable Xdebug.

1. Go to the plugin settings: `WP-Admin > WooCommerce > Settings`
2. Go to Payments and open the `Stripe Gateway` details page
3. Go to the `Payment Methods` tab
4. Check that only one call to the Configurations API was made
5. Reload the page, and check that a new Configurations API call was made (always get the latest data for the settings page)
6. Go to the Store, add any product to the cart, go to the cart, and then go to the checkout
7. Check that no new Configurations API call was made

The new Cache implementation uses an `option`, you can check it with:
`wp option get wcstripe_test_payment_method_configuration_cache`
It should show something like this:
```
% wp option get wcstripe_test_payment_method_configuration_cache
array (
  'data' =>
  (object) array(
     'id' => 'pmc_1P9Z7lH8j9v0pLPbZq3PvAta',
     'object' => 'payment_method_configuration',
     'active' => true,
     'affirm' =>
    (object) array(
       'available' => false,
       'display_preference' =>
      (object) array(
         'overridable' => true,
         'preference' => 'off',
         'value' => 'off',
      ),
    ),
  ...
  'card' =>
    (object) array(
       'available' => true,
       'display_preference' =>
      (object) array(
         'overridable' => true,
         'preference' => 'on',
         'value' => 'on',
      ),
    ),
  ...
  ),
  'updated' => 1747680270,
  'ttl' => 600,
)
```

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -103,11 +103,11 @@ private static function get_payment_method_configuration_from_cache( $use_fallba
...
 
-		$cache_key                    = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
-		$cached_primary_configuration = get_transient( $cache_key );
+		$cache_key                    = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_KEY;
+		$cached_primary_configuration = WC_Stripe_Database_Cache::get( $cache_key );
 		if ( false === $cached_primary_configuration || null === $cached_primary_configuration ) {
 			if ( $use_fallback ) {
-				return get_option( $cache_key );
+				return get_option( $cache_key . '_fallback' );
 			}
```
annemirasol, line 111: To be clear, I'm 100% for removing this fallback cache. Just not the cooldown period.

I'm also okay with removing it in a following PR.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -133,11 +134,11 @@ public static function clear_payment_method_configuration_cache() {
 	 */
 	private static function set_payment_method_configuration_cache( $configuration ) {
 		self::$primary_configuration = $configuration;
-		$cache_key                   = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
-		set_transient( $cache_key, $configuration, self::CONFIGURATION_CACHE_TRANSIENT_EXPIRATION );
+		$cache_key                   = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_KEY;
+		WC_Stripe_Database_Cache::set( $cache_key, $configuration, self::CONFIGURATION_CACHE_EXPIRATION );
 
 		// To be used as fallback if we are in API cooldown and the transient is not available.
```
annemirasol:
```suggestion
		// To be used as fallback if we are in API cooldown and the main cache is not available.
```

annemirasol: We're removing the secondary cache in this PR or in a following PR anyways -- but just in case it's staying.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -121,9 +121,10 @@ private static function get_payment_method_configuration_from_cache( $use_fallba
 	 */
 	public static function clear_payment_method_configuration_cache() {
 		self::$primary_configuration = null;
-		$cache_key                   = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_TRANSIENT_KEY;
-		delete_transient( $cache_key );
+		$cache_key                   = WC_Stripe_Mode::is_test() ? self::TEST_MODE_CONFIGURATION_CACHE_KEY : self::LIVE_MODE_CONFIGURATION_CACHE_KEY;
+		WC_Stripe_Database_Cache::delete( $cache_key );
 		delete_option( $cache_key );
```
annemirasol:
```suggestion
```

I think this is redundant -- `delete_option( $cache_key )` is already performed by the preceding line.

annemirasol (commented):
- Verified no excessive PMC fetches
- Verified able to update payment methods in wp-admin and immediately see them in the store front

Found a possibly redundant call to `delete_option()`.

includes/class-wc-stripe-payment-method-configurations.php
diegocurbelo: We decided to leave it in place until after we merge this and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4334, so we have the 1-minute fallback as an extra precaution.

includes/class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,151 @@
...
+	 * @param int     $ttl     The TTL of the cache.
+	 *
+	 * @return void
+	 */
+	private static function write_to_cache( $key, $data, $ttl ) {
+		// Add the  data and expiry time to the array we're caching.
+		$cache_contents            = [];
+		$cache_contents['data']    = $data;
+		$cache_contents['updated'] = time();
+		$cache_contents['ttl']     = $ttl;
```
daledupreez: Non-blocking nit: using multiple assignments into the array feels noisy to me, and makes me read a bit harder to fully get that we're writing a single array and what shape it has. I would find it easier to set the contents as a single array, but that's truly a non-blocking stylistic preference.
```suggestion
		$cache_contents            = [
			'data'    => $data,
			'ttl'     => $ttl,
			'updated' => time(),
		];
```

includes/class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,151 @@
...
+		self::$in_memory_cache[ $key ] = $cache_contents;
+
+		// Create or update the DB option cache.
+		// Note: Since we are adding the current time to the option value, WP will ALWAYS write the option because
+		// the cache contents value is different from the current one, even if the data is the same.
+		// A `false` result ONLY means that the DB write failed.
+		// Yes, there is the possibility that we attempt to write the same data multiple times within the SAME second,
+		// and we will mistakenly think that the DB write failed. We are OK with this false positive,
+		// since the actual data is the same.
+		$result = update_option( $key, $cache_contents, 'no' );
```
daledupreez: I have multiple comments here:
* I am not sure why we're using autoload = 'no', as I believe we're expecting to need this cache regularly, so not autoloading the value is unexpected for me.
  - It may have to do with other processes being more likely to fetch the value from the database later, but I still feel we should be OK with there being small timing gaps.
* It looks like we should use a boolean value in more recent versions of WordPress per the PHPDoc for `update_option`:
https://github.com/WordPress/wordpress-develop/blob/9ebba7a9aedc57d5492e409cd0aae0b6958d3642/src/wp-includes/option.php#L831-L847
* I think we should explicitly comment on why we're making whatever choice we want for this flag.

includes/class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * A class for caching data as an option in the database.
+ */
+class WC_Stripe_Database_Cache {
+
+	/**
+	 * In-memory cache for the duration of a single request.
+	 *
+	 * This is used to avoid multiple database reads for the same data and as a backstop in case the database write fails,
+	 * thus ensuring the cache generator is not called multiple times (which would mean multiple API calls).
```
daledupreez: I think the "cache generator" aspect of this should be removed -- it looks like it's replicated from the original code.

includes/class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,151 @@
...
+			wp_cache_delete( $key, 'options' );
+		}
+	}
+
+	/**
+	 * Get the cache contents for a certain key.
+	 *
+	 * @param string $key The cache key.
+	 *
+	 * @return array|false The cache contents (array with `data`, `fetched`, and `errored` entries).
```
daledupreez: I think we're returning `data`, `ttl`, and `updated` entries, AFAICT.
```suggestion
	 * @return array|false The cache contents (array with `data`, `ttl`, and `updated` entries).
```

includes/class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,151 @@
...
+	 * @param string $key            The cache key.
+	 * @param array  $cache_contents The cache contents.
+	 *
+	 * @return boolean True if the contents are expired. False otherwise.
+	 */
+	private static function is_expired( $key, $cache_contents ) {
+		$expires = $cache_contents['updated'] + $cache_contents['ttl'];
+		$now     = time();
+
+		return apply_filters( 'wcstripe_database_cache_is_expired', $expires < $now, $key, $cache_contents );
```
daledupreez, line 166: While we shouldn't _expect_ bad data to be written to the cache, I think we should guard against it, especially because we're not limiting which keys can be passed to the functions in the class.
```suggestion
		if ( ! is_array( $cache_contents ) || ! isset( $cache_contents['updated'] ) || ! isset( $cache_contents['ttl'] ) {
			// Treat bad/invalid cache contents as expired
			return true;
		}
		// Double-check that we have integers for `updated` and `ttl`.
		if ( ! is_int( $cache_contents['updated'] ) || ! is_int( $cache_contents['ttl'] ) ) {
			return true;
		}

		$expires = $cache_contents['updated'] + $cache_contents['ttl'];
		$now     = time();

		return apply_filters( 'wcstripe_database_cache_is_expired', $expires < $now, $key, $cache_contents );
```

tests/phpunit/test-class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,171 @@
...
+			'null'    => null,
+		];
+
+		foreach ( $test_cases as $type => $data ) {
+			$key    = "test_{$type}";
+			WC_Stripe_Database_Cache::set( $key, $data );
+			$result = WC_Stripe_Database_Cache::get( $key );
+			$this->assertEquals( $data, $result, "Failed to cache {$type} data type" );
+		}
+	}
```
daledupreez: Nit: It feels like using a dataProvider could make it slightly easier to "see" what failed from the test descriptions when things go wrong. I am 1000% in favour of using test runners with different inputs, so the quibble here is mostly around later ease in seeing failures from logs.
```suggestion
	/**
	 * Data provider for {@see test_cache_with_data_type()}.
	 *
	 * @return array Array of test cases.
	 */
	public function provide_data_type_test_cases() {
		return [
			'string'  => [ 'string', 'test string' ],
			'integer' => [ 'integer', 123 ],
			'array'   => [ 'array', [ 'key' => 'value' ] ],
			'object'  => [ 'object', (object) [ 'property' => 'value' ] ],
			'boolean' => [ 'boolean', true ],
			'null'    => [ 'null', null ],
		];
	}

	/**
	 * Test cache with different data types.
	 *
	 * @dataProvider provide_data_type_test_cases
	 */
	public function test_cache_with_different_data_types( $key_suffix, $data ) {
		$key    = "test_{$key_suffix}";
		WC_Stripe_Database_Cache::set( $key, $data );
		$result = WC_Stripe_Database_Cache::get( $key );
		$this->assertEquals( $data, $result, "Failed to cache and fetch data type" );
	}
```

tests/phpunit/test-class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,171 @@
...
+	/**
+	 * Test cache with different data types.
+	 */
+	public function test_cache_with_different_data_types() {
+		$test_cases = [
+			'string'  => 'test string',
+			'integer' => 123,
+			'array'   => [ 'key' => 'value' ],
+			'object'  => (object) [ 'property' => 'value' ],
+			'boolean' => true,
```
daledupreez: I think we should also have a test case for `false` to make sure that works, not just `true`. False is more likely to break or behave weirdly! 🙃

tests/phpunit/test-class-wc-stripe-database-cache.php
daledupreez: As a separate thought, it may be worth combining these data type tests with `test_cache_expiration()` and `test_in_memory_cache()`. I think it would be good to make sure that all layers are working correctly across all types.

tests/phpunit/test-class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,171 @@
...
+	 * Clean up after each test.
+	 */
+	public function tearDown(): void {
+		parent::tearDown();
+		// Clean up any test data.
+		WC_Stripe_Database_Cache::delete( 'test_key' );
+		WC_Stripe_Database_Cache::delete( 'expiring_key' );
+		WC_Stripe_Database_Cache::delete( 'delete_key' );
+		WC_Stripe_Database_Cache::delete( 'memory_key' );
+		WC_Stripe_Database_Cache::delete( 'custom_ttl_key' );
```
daledupreez: This cleanup is incomplete, especially if you consider the datatype tests. Maybe it would be simpler if we added a public, static `get_cached_keys()` helper to the cache class that returns `array_keys( self::$in_memory_cache )`.

tests/phpunit/test-class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,171 @@
...
+/**
+ * WC_Stripe_Database_Cache_Test Class.
+ */
+class WC_Stripe_Database_Cache_Test extends WP_UnitTestCase {
+	/**
+	 * Setup test.
+	 */
+	public function setUp(): void {
+		parent::setUp();
+	}
```
daledupreez: We should remove this -- it's a no-op.
```suggestion
```

includes/class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,151 @@
...
+	 *
+	 * @param string $key  The key to store the value under.
+	 * @param mixed  $data The value to store.
+	 * @param int    $ttl  The TTL of the cache. Dafault 1 hour.
+	 *
+	 * @return void
+	 */
+	public static function set( $key, $data, $ttl = HOUR_IN_SECONDS ) {
+		self::write_to_cache( $key, $data, $ttl );
+	}
```
daledupreez, line 44: I think we may need to require or add a key prefix so we don't unintentionally overwrite or delete "real" options in the database. Requiring callers to provide valid keys would make it more work for callers, while adding a prefix would reduce the key length that we can support, and could lead to redundant option key names. That said, having a key prefix would allow for clearing all persistent cached data via a database query.
```suggestion
	public static function set( $key, $data, $ttl = HOUR_IN_SECONDS ) {
		// Requiring a key prefix
		if ( ! self::is_permitted_key( $key ) ) {
		 	// Either return early and do nothing, or throw an exception
		 	return;
		 	throw new InvalidArgumentException( 'Unsupported key' );
		}
		// Add key prefix -- that function would simply do something like return 'wc_stripe_db_cache_' . $key;
		$db_key = self::get_db_key( $key );

		self::write_to_cache( $key, $data, $ttl );
	}
```

daledupreez (commented):
This is testing well with one minor note: we've kept the cooldown logic in effect, so we won't re-fetch data while we're in the cooldown. (This impacts step 5 in the test instructions.)

I have a lot of smaller feedback items, but I want to call out two:
 * We need to make a call on whether we want prefixing for the options, and how that should work.
 * We should add some guards for bad data in `is_expired()`.

It's probably worth adding an explicit link to the WooCommerce Payments code we're using to start with: https://github.com/Automattic/woocommerce-payments/blob/4b084af108cac9c6bd2467e52e5cdc3bc974a951/includes/class-database-cache.php

includes/class-wc-stripe-database-cache.php
diegocurbelo: To reduce the changes needed in the WC_Stripe_Payment_Method_Configurations class, I just kept the prefixes we had in place; I agree we should add a prefix for the cache (wcstripe_cache_*) but we should also handle the test/live mode internally, so we can remove the multiple WC_Stripe_Mode::is_test() ternary ifs we have everytime we use the cache. I'll make a new PR for that improvement.

includes/class-wc-stripe-database-cache.php
diegocurbelo: > Autoloading too many options can lead to performance problems, especially if the
> options are not frequently used. For options which are accessed across several places
> in the frontend, it is recommended to autoload them, by using true.
> For options which are accessed only on few specific URLs, it is recommended
> to not autoload them, by using false.

The behavior was copied from the WooPayments implementation. I think it makes sense to set the autoload = false, as we are implementing this to be a general cache for the plugin; some values like the PMC are going to be needed frequently, but it might be better to allow specific values to set it to true (we can do this when we implement the prefix and test/live mode detection).
I don't have a strong opinion either way. I replaced the 'no' with false for now, but I'm happy to remove the parameter and let WordPress determine the autoload value.

annemirasol (approved):
Looks overall solid to me.

Tested, while logging when fallback cache is used:
- when cache is expired, we fetch the PMC
- within cooldown, no fetches observed, even when reloading the payment methods page
- when cache is not expired, storefront does not trigger a PMC fetch
- after updating payment methods, fetch is observed
- no indication fallback cache had to be used

includes/class-wc-stripe-database-cache.php
daledupreez: I don't have a super-strong opinion either, but it feels like we're fetching the payment method configurations a lot, so they may qualify for a `true` value. But that does feel like something that is worth tackling in a follow-up.

includes/class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * Class WC_Stripe_Database_Cache
+ */
+
+/**
+ * A class for caching data as an option in the database.
+ *
+ * Based on the WooCommerce Payments Database_Cache class implementation.
+ * https://github.com/Automattic/woocommerce-payments/blob/4b084af108cac9c6bd2467e52e5cdc3bc974a951/includes/class-database-cache.php
```
daledupreez: Non-blocking nit: maybe this should use `@see` to be super-explicit.
```suggestion
 * @see https://github.com/Automattic/woocommerce-payments/blob/4b084af108cac9c6bd2467e52e5cdc3bc974a951/includes/class-database-cache.php
```

includes/class-wc-stripe-database-cache.php
```diff
@@ -0,0 +1,176 @@
...
+	 * Wraps the data in the cache metadata and stores it.
+	 *
+	 * @param string  $key     The key to store the data under.
+	 * @param mixed   $data    The data to store.
+	 * @param int     $ttl     The TTL of the cache.
+	 *
+	 * @return void
+	 */
+	private static function write_to_cache( $key, $data, $ttl ) {
+		// Add the  data and expiry time to the array we're caching.
```
daledupreez: Nit: extra space
```suggestion
		// Add the data and expiry time to the array we're caching.
```

tests/phpunit/test-class-wc-stripe-database-cache.php
daledupreez: We should still remove this method (but it's not blocking).

daledupreez (approved):
I think we can get this shipped as-is, and then iterate from there.

Noting that I didn't retest, but I did check the changes since my last review and they look good. Also, [Anne (re)tested](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4331#pullrequestreview-2855708752), so I feel confident in the code.

includes/class-wc-stripe-database-cache.php
daledupreez: I just added this in cc434a2c81227e102cce61a9df4a15cd82cf6830.

includes/class-wc-stripe-database-cache.php
daledupreez: I fixed this in cc434a2c81227e102cce61a9df4a15cd82cf6830.

PR #4334, opened 2025-05-19
Disable Settings Sync when no valid PMC is available

diegocurbelo:
Fixes #<github_issue_id>

This PR disables the Payment Method Configuration sync if the response from the Stripe API does not contain a PMC that inherits from the Platform PMC.

It uses the same `pmc_enabled` configuration setting:
- **empty/unset**: PMC temporarly enabled, migration needed
- **yes**: PMC enabled
- **no**: PMC disabled

The migration needed scenario only happens on the first call to either `get_upe_available_payment_method_ids` or `get_upe_enabled_payment_method_ids`, after that, `pmc_enabled` is set to either `yes` or `no`.

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -159,6 +159,8 @@ private static function get_payment_method_configuration_from_stripe() {
 			}
 		}
 
+		// If we don't have a Payment Method Configuration that inherits from the WooCommerce Platform, disable Payment Method Configuration sync.
+		self::disable_payment_method_configuration_sync();
```
daledupreez, line 163: I think we need to guard against cases where we receive `[]` or `null` from the `get_payment_method_configurations()` call, which could easily be the case if we've been rate limited.

daledupreez: Also, might it make sense to add a lower-level cache to the raw `get_payment_method_configurations()` API calls? (I don't think this blocks the logic here, but that feels like it could also avoid unnecessary API calls.)

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -328,8 +330,13 @@ function ( $id ) use ( $is_test_mode ) {
...
 		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
-		$key             = WC_Stripe_Mode::is_test() ? 'test_connection_type' : 'connection_type';
-		return isset( $stripe_settings[ $key ] ) && 'connect' === $stripe_settings[ $key ];
+
+		$connection_type_key              = WC_Stripe_Mode::is_test() ? 'test_connection_type' : 'connection_type';
+		$is_account_connected_to_platform = isset( $stripe_settings[ $connection_type_key ] ) && 'connect' === $stripe_settings[ $connection_type_key ];
+
+		$is_pmc_disabled = isset( $stripe_settings['pmc_enabled'] ) && 'no' === $stripe_settings['pmc_enabled'];
+
+		return $is_account_connected_to_platform && ! $is_pmc_disabled;
```
daledupreez: Non-blocking nit: I think this could be easier to read if we used early returns.
```suggestion
		// If the account is not a Connect account, we should not sync payment methods.
		if ( ! isset( $stripe_settings[ $connection_type_key ] ) || 'connect' !== $stripe_settings[ $connection_type_key ] ) {
			return false;
		}

		// If we don't have the pmc_enabled flag, or it is set to no, we should not sync payment methods.
		if ( ! isset( $stripe_settings['pmc_enabled'] ) || 'no' === $stripe_settings['pmc_enabled'] ) {
			return false;
		}

		return true;
```

tests/phpunit/test-class-wc-stripe-payment-method-configurations.php
```diff
@@ -11,4 +11,121 @@ class WC_Stripe_Payment_Method_Configurations_Test extends WP_UnitTestCase {
...
+		WC_Stripe_Helper::update_main_stripe_settings( $initial_settings );
+		$property->setValue( null, null );
+	}
+
+	/**
+	 * Tests that pmc_enabled is not set to 'no' when valid PMC data exists.
+	 *
+	 * @return void
+	 */
+	public function test_pmc_enabled_not_disabled_with_valid_data() {
```
daledupreez, line 89: It feels like we should add a test where we have configurations with other parent IDs as well.

Mayisha:
I have added a few changes.
- [3231d25](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4334/commits/3231d2582399ad2e8579c04e6adc5f05f7d7aa8a): added necessary mocks to fix the failing tests.
- [fbeb4e2](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4334/commits/fbeb4e2bbabe571a541c932ab67eb60b9a7d6123), [8485c4f](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4334/commits/8485c4f6e54c5877885c074fb03e6c2771350b9f): This fixes the following scenario-
    - Simulate a scenario that disables the pmc. I tested by manually changing the parent configuration id which set `pmc_enabled` to `no` when I authenticated to my Stripe account.
    - Now change the parent configuration id as before.
    - Connect to the same or another Stripe account.
    - Re-auth does not reset or clear the `no` value from `pmc_enabled` and so the syncing never starts even after connecting under the platform account.
In these 2 commits, I have cleared `pmc_enabled` where necessary to trigger reassessing the value on reconnect.

includes/class-wc-stripe-payment-method-configurations.php
diegocurbelo: > I think we need to guard against cases where we receive `[]` or `null` from the `get_payment_method_configurations()` call, which could easily be the case if we've been rate-limited.

I think we should default to disabling sync if anything unexpected happens, and once we stabilize the rate limiting problem, we can attempt to re-enable sync (with a retry limit or similar, like try 4 times with an interval of 8n hours)

daledupreez: Noting that @diegocurbelo and I discussed this yesterday, and we landed on adding a comment as a starting point.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -163,7 +163,7 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 			$options[ $prefix . 'publishable_key' ]     = $publishable_key;
 			$options[ $prefix . 'secret_key' ]          = $secret_key;
 			$options[ $prefix . 'connection_type' ]     = $type;
-
+			$options['pmc_enabled']                     = 'connect' === $type ? '' : 'no'; // When not connected via oauth, the PMC is disabled. Otherwise, set to empty string which will be set to 'yes' after the migration in 'maybe_migrate_payment_methods_from_db_to_pmc'.
```
daledupreez, line 166: This feels like it might be risky: what happens if we connect in test mode, but toggle the main `pmc_enabled` flag? Should we just tackle using a prefixed version of the flag?

daledupreez (commented):
I have comments from inspection, will loop through some other reviews and then loop back to test this. (My comments a pretty minor overall.)

daledupreez (approved):
I think this change is ready to ship, though I think it would be worth logging when this happens, either via the error log or via the Stripe logs.

For reference, I tested this by modifying `WC_Stripe_API_>get_payment_method_configurations()` to return the following:
* A list of payment methods where the `parent` had `'_BAD'` added as a suffix
* An empty list of payment methods
* A normal list of payment methods from Stripe

In the first two cases, the `pmc_enabled` flag was set to "no", while in the third it was left as `'yes'`.

includes/class-wc-stripe-payment-method-configurations.php
diegocurbelo: Fixed in [d353bd0](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4334/commits/d353bd0d6df8762c1528ef9c2607aecb36a6ba6a).
The pmc_enabled check needs to only disable the use of PMCs if the `pmc_enabled` flag is set and it's `no`. When not set (or empty), it means the migration was not attempted yet (that will result in a 'yes' or 'no'). I added a comment explaining this use case.

```
// If we have the pmc_enabled flag and it's set to no, we should not use payment methods configurations.
if ( isset( $stripe_settings['pmc_enabled'] ) && 'no' === $stripe_settings['pmc_enabled'] ) {
	return false;
}
```

tests/phpunit/test-class-wc-stripe-payment-method-configurations.php
diegocurbelo: Added in [1924bf0](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4334/commits/1924bf07ddfead702664fa572dbd942bb7dd8bf5)

includes/connect/class-wc-stripe-connect.php
diegocurbelo: All plugin settings (other than keys and webhooks) are shared between _live_ and _test_ mode; I think that having two Payment Method lists would be confusing 🤷🏼

PR #4347, opened 2025-05-22
Fix reconnect loop and PMC sync after reconnect

Mayisha:
Fixes #<github_issue_id>

- Removes the code that updates PMC with data from `upe_checkout_experience_accepted_payments` on connect/re-connect. When `pmc_enabled` is empty, this is handled in the `maybe_migrate_payment_methods_from_db_to_pmc` function
- Prevents disabling pmc when `get_payment_method_configurations` returns an api error by bailing early.

## Testing instructions

1. Simulate an invalid connection configuration
    `wp option patch update woocommerce_stripe_settings pmc_enabled 'no'`
2. Go to `WP-Admin > WooCommerce > Settings > Payments` and select `Stripe` 
3. Go to the `Settings` and `Refresh account details`:
    Screenshot_2025-05-21_at_21_05_32 
4. Check that the connection error message is displayed:
    Screenshot 2025-05-21 at 21 09 28
5. Now reconnect the account and check that the connection succeeds.
6. Check `wp option pluck woocommerce_stripe_settings pmc_enabled`. It should be set to `yes`.

### DIFF
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -146,7 +146,13 @@ class WC_Stripe_Payment_Method_Configurations {
 	 * @return object|null
 	 */
 	private static function get_payment_method_configuration_from_stripe() {
-		$result         = WC_Stripe_API::get_instance()->get_payment_method_configurations();
+		$result = WC_Stripe_API::get_instance()->get_payment_method_configurations();
+
+		// If the result does not have a data property, it means we received an error from the API.
+		if ( ! isset( $result->data ) ) {
+			return null;
+		}
+
 		$configurations = $result->data ?? [];
 
 		// When connecting to the WooCommerce Platform account a new payment method configuration is created for the merchant.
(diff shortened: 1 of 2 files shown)


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -146,7 +146,13 @@ private static function set_payment_method_configuration_cache( $configuration )
 	 * @return object|null
 	 */
 	private static function get_payment_method_configuration_from_stripe() {
-		$result         = WC_Stripe_API::get_instance()->get_payment_method_configurations();
+		$result = WC_Stripe_API::get_instance()->get_payment_method_configurations();
+
+		// If the result does not have a data property, it means we received an error from the API.
+		if ( ! isset( $result->data ) ) {
+			return null;
+		}
```
daledupreez, line 154: AFAICT, `WC_Stripe_API::retrieve()` can also return `null`, so we should likely check for that case too.
```suggestion
		// If the result does not have a data property, it means we received an error from the API.
		if ( ! is_object( $result ) ||  ! isset( $result->data ) ) {
			return null;
		}
```

wjrosa (approved):
Great fix, Mayisha! Working perfectly for me:

Error UI:
Screenshot 2025-05-22 at 10 13 48

After reconnecting:
Screenshot 2025-05-22 at 10 16 25

`pmc_enabled` value:
Screenshot 2025-05-22 at 10 17 24
(I am constantly getting that translation error, but it is not related to this change or feature)

includes/class-wc-stripe-payment-method-configurations.php
wjrosa: (took the freedom to fix your suggestion, `isobejct` > `is_object`)

PR #4348, opened 2025-05-22
Ensure we clear the in-memory Stripe API keys for key situations

daledupreez:
Fixes STRIPE-461, STRIPE-462, STRIPE-470

In 9.5.x, we've seeng situations where users end up in authentication loops: they keep reconnecting to Stripe, but we continue to disable their access. This PR aims to resolve that issue by ensuring that the static, in-memory key used by `WC_Stripe_API` is cleared when we update specific settings that mean we should use a new key. Those settings are:
 * When `testmode` is modified
 * When `secret_key` is modified
 * When `test_secret_key` is modified

In terms of implementation, the logic is triggered by the existing `WC_Stripe->gateway_settings_update()` function, which itself is hooked into the `pre_update_option_woocommerce_stripe_settings` filter. It specifically checks if those keys exist in both the new and old settings arrays, and then calls `WC_Stripe_API::set_secret_key( '' )` to reset `WC_Stripe_API::$secret_key`. That way future calls from the current process will look up the correct key on the next attempt to call Stripe.

### Reproduce the problem

Using a current/recent build without this patch, do the following:
* Navigate to _WooCommerce_ > _Settings_ > _Payments_ and select the Stripe option from the list
* If you are not already connected to Stripe, connect using a test account
* Switch to the "Payment Methods" tab
* Ensure that the "Apple Pay / Google Pay" option is enabled, and save the changes if needed
* Switch back to the "Settings" tab
* Run the following WP-CLI commands to simulate breaking the Stripe connection:
```
wp option patch update woocommerce_stripe_settings test_secret_key 'sk_test_INVALID'
wp option patch update woocommerce_stripe_settings pmc_enabled 'no'
```
* In the "Account status" section, click on the kebab menu, and select "Refresh account details"
stripe_refresh_details
* Verify that you see the connection error displayed
stripe_account_not_connected
* If you click on the "Configure connection" button and reconnect the account, you keep getting stuck in this state

### Verify the fix

Now apply the changes from this branch.
* Navigate to _WooCommerce_ > _Settings_ > _Payments_ and select the Stripe option from the list
* Switch to the "Settings" tab
* Click on "Configure connection" button and work through the flow to reconnect the account
* The account connection should succeed

### Smoke test `testmode` changes

This is less likely to be problematic, given that the entities for live and test mode are not distinct for the most part, but I think it's good hygiene regardless. It does require you to have an eligible, live Stripe account to connect, however.
 * Get the site back into the broken state.
 * On the "Settings" tab, toggle the "Enable test mode" setting and save the changes
 * Nothing should break. (Feel free to toggle back too!)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,6 +11,7 @@
 * Add - Implement custom database cache for persistent caching with in-memory optimization.
 * Update - Remove feature that flags 401s and proactively blocks subsequent API calls until the store has reauthenticated.
 * Fix - Disable payment settings sync when we receive unsupported payment method configurations.
+* Fix - Ensure that we use current Stripe API keys after settings updates
 
 = 9.5.1 - 2025-05-17 =
 * Fix - Add a fetch cooldown to the payment method configuration retrieval endpoint to prevent excessive requests.
--- a/readme.txt
+++ b/readme.txt
@@ -122,5 +122,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Add - Implement custom database cache for persistent caching with in-memory optimization.
 * Update - Remove feature that flags 401s and proactively blocks subsequent API calls until the store has reauthenticated.
 * Fix - Disable payment settings sync when we receive unsupported payment method configurations.
+* Fix - Ensure that we use current Stripe API keys after settings updates
 
 [See changelog for full details across versions](https://raw.githubusercontent.com/woocommerce/woocommerce-gateway-stripe/trunk/changelog.txt).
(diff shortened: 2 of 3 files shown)


wjrosa (approved):
Looks great and works as expected:

Error UI:
Screenshot 2025-05-22 at 11 24 59

After reconnection:
Screenshot 2025-05-22 at 11 26 37

`pmc_enabled` value:
Screenshot 2025-05-22 at 11 26 57

annemirasol (approved):
Discussed over Slack huddle.

PR #4355, opened 2025-05-23
Void intent when cancelling an uncaptured order

malithsen:
Fixes STRIPE-456

This PR ensures that uncaptured Stripe PaymentIntents are voided when the associated order is canceled.

Previously, `process_refund()` treated all zero-amount refund requests as no-ops. However, for uncaptured payments (pre-authorizations), the correct behavior is to void the PaymentIntent rather than return early. This change updates the logic so that zero-amount refunds will only be skipped if the charge was already captured. For uncaptured charges, the method now proceeds to void the intent.

Ideally, we'd want to decouple voiding and refunding into separate methods and [cancel_payment](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/d547ff3ab28ddf394e004f3d3314fe6a5a482716/includes/class-wc-stripe-order-handler.php#L394) should call the correct method (refund vs void) appropriately. That's left as an exercise for another time :)  

## Testing instructions

1. As a merchant, ensure that Stripe is set to authorize first and capture later from settings panel `wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings`
2. As a customer, go through the purchase process and complete checkout.
3. As a merchant, in the edit order screen, change the order status to cancelled.
4. Order notes should indicate that the order was voided.
CleanShot 2025-05-23 at 12 33 55
6. View the transaction in the Stripe dashboard and ensure the payment intent was cancelled
CleanShot 2025-05-23 at 13 37 25

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,6 +11,7 @@
 * Fix - Prevent styles from non-checkout pages affecting the appearance of Stripe element.
 * Fix - Send correct attribute when setting the default payment method.
 * Dev - Build dynamic WordPress and WooCommerce dependencies for unit tests.
+* Fix - Void intent when cancelling an uncaptured order
 
 = 9.5.2 - 2025-05-22 =
 * Add - Implement custom database cache for persistent caching with in-memory optimization.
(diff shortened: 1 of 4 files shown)


diegocurbelo (approved):
Looks good, and tests well 🚢 

> Ideally, we'd want to decouple voiding and refunding into separate methods 
Should we create a tech debt issue for this?

malithsen:
> Should we create a tech debt issue for this?

@diegocurbelo Created https://linear.app/a8c/issue/STRIPE-479/ :)

PR #3691, opened 2025-01-03
Using Blocks API to process ECE orders

wjrosa:
Fixes #1573

## Changes proposed in this Pull Request:

This PR changes how ECE orders are processed. Those will now use the Blocks API. To do that, I have introduced multiple new methods on the front end that request the API endpoint (`/wc/store/v1/checkout/`) and format the body according to the new structure.

The new flow is controlled by the `express_checkout_use_blocks_api` Stripe setting.

## Testing instructions

- Checkout and build this branch on your test environment (`tweak/using-blocks-api-to-process-ece-orders`)
- Connect your Stripe account and enable express payment methods (Google Pay, Apple Pay, Link)
- Enable the use of Blocks API for ECE. You can do it by hardcoding `true` as the returning value of `use_blocks_api`
- Access your store by a public-facing address (with either JN website or something like Ngrok)
- As a shopper, purchase any product using express methods (on both blocks checkout and shortcode checkout)
- Confirm you can complete the transaction

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [x] Added changelog entry **in both** `changelog.txt` and `readme.txt` (or does not apply)
-   [ ] Tested on mobile (or does not apply)

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

cc @reykjalin

PR #4357, opened 2025-05-26
ECE: Fix state mapping

Mayisha:
Fixes STRIPE-478, STRIPE-442, STRIPE-455

When a shopper pays with Google/Apple Pay, for a few countries, the express payment methods return the billing and shipping address in a different format than WooCommerce core. For example, if the billing country is Ireland and the state is Kildare, Google Pay returns `Co. Kildare` but Woocommerce expects `KE` for the same state.

We had the state mapping in place and it was used in the create_order [ajax request](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php#L319). As we moved away from this ajax and using [Blocks API now](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3691/files), the mapping is ignored now. As a result when the shipping/billing address has one of these states, the checkout fails with `Invalid parameter(s): billing_address, shipping_address` error.

- Added an Ajax call to normalize the address fields before the checkout request. (I have tried some possible filters/actions, but the error happens during the REST params validation step before reaching to any hooks. Another alternative would be to move the whole mapping on the JS side)
- Tweaked the existing helper methods to be used with the Ajax method based on the current format of the request params.
- Removed the unused `create_order` Ajax method.

## Testing instructions
- In your Google Pay, create a payment method with the billing address in Ireland. (I have tested with hard-coded address. You can set the following billing address hard-coded [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/trunk/client/express-checkout/utils/normalize.js#L56-L61)

```
country: 'IE',
city: 'Straffan',
state: 'Co. Kildare', 
postcode: 'W23 K6D7',
```
- Add a product to your cart and go to the checkout page.
- In `develop` branch, try to pay with Google Pay. You should get the error `Invalid parameter(s): billing_address`
- In this branch, the payment should be successful. If you check the order, the billing address should have the correct state, city and country.


includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -140,6 +140,21 @@ public function ajax_clear_cart() {
 		exit;
 	}
 
+	/**
+	 * Normalizes address fields in WooCommerce supported format.
+	 */
+	public function ajax_normalized_address() {
```
malithsen: nit: rename function to match current convention
```suggestion
	public function ajax_normalize_address() {
```

malithsen (approved):
Nice work!
Replicated the issue on `develop` and confirmed it's fixed in this branch with hard-coded data.

> Another alternative would be to move the whole mapping on the JS side

We might have to do this, if performance becomes an issue, but for now, this should be good enough, especially given that it's a very light-weight call.

I left a suggestion on renaming a function, but it's not a blocker for merging.

PR #4367, opened 2025-05-27
Add country shipping restrictions to Express Checkout

### LINKED ISSUE #4365: Express checkout doesn't honor shipping country restrictions
## Description

When setting up a store to only allow shipping to a specific country, the express checkout buttons don't pass those restrictions on to Link, Apple Pay, or Google Pay.

## To Reproduce

1. Set up a WooCommerce store to only allow selling to and shipping to a specific country
2. Add a product to the store that requires shipping
3. Ensure that a shipping option exists for the country specified above
4. Ensure that one or more express checkout option is enabled
5. Add your product to the cart while in an incognito window
6. Pick one of the express payment methods, and specify a shipping address in another country
7. The purchase is allowed to proceed

### Expected behavior

The payment should not be permitted when an invalid country is specified.

### Additional context

This issue was originally reported in the support forums:
 * https://wordpress.org/support/topic/link-payment-does-not-obey-woo-country-restrictions/

daledupreez:
Fixes STRIPE-481
Fixes #4365

This PR ensures that when merchants have configured their WooCommerce store to ship only to a specified set of countries, and we have a product that needs to be shipped, we specify those countries in the `allowedShippingCountries` option for express checkout elements ([Stripe docs](https://docs.stripe.com/js/elements_object/create_express_checkout_element#express_checkout_element_create-options-allowedShippingCountries)). Note that this will only be in effect when we can also compute a shipping rate for the product, so we need some level of location default for users.

The new logic is only effective when all of the following are true:
 * There are one or more products on the page or in the cart that require shipping
 * We can compute one or more shipping rates for the product(s)
 * Express checkout is enabled
 * The store is configured to ship to a limited set of countries

Note that the basic requirement for shipping rates to be available for shippable products already existed, this PR adds on the country restrictions when they can be determined.

### Implementation notes

The approach I've taken in this PR is fairly simple:
 * I've added a `get_allowed_shipping_countries()` helper method to `WC_Stripe_Express_Checkout_Helper`, which wraps the WooCommerce Core logic in `WC()->countries->get_shipping_countries()`, and extracts only the country codes from the array keys.
 * I've added a new `allowed_shipping_countries` field to the express checkout data we return, and then use that data to populate the `allowedShippingCountries` option for the express checkout buttons.

#### Concerns and possible gaps

I am not 100% sure as to how I ought to be handling certain situations, including the following:
 * When shipping is disabled, should we return `[]` to ensure we don't allow any countries to be shipped to? Or is that likely to be handled by us not allowing shippable products to be purchased?
 * Should we be more proactive when we can tell that no countries are eligible for shipping?
 * ~I am not sure how to test or configure Link for testing~

## Testing instructions

This features needs a dedicated site for testing so Apple Pay and Google Pay can be fully configured (at least in test mode). What that means is the following:
 * WooCommerce needs to be installed
 * WooCommerce needs to be configured to allow shipping to a small number of countries -- it can be just one or a few
 * At least one shipping method needs to be enabled for each of the permitted countries
   - For better visibility, it may be worth using flat rate shipping so you can see when it is included vs not -- that's harder to see with free shipping
 * The store needs at least one product that requires shipping and is eligible for the shipping method defined above

I successfully testing this with a jurassic.ninja site, for what it's worth.

With all of that in place, you need to create a build of the plugin from this branch, and upload it to your site.

* Navigate to _WooCommerce_ > _Settings_ > _Payments_
* Click on the Stripe option
* Click on the option to connect a test account, and work through the Stripe connection flow to connect a test account
* Ensure that Apple Pay, Google Pay, and Link are all enabled

### Google Pay

* Navigate to your shippable product in the store front-end, using either a desktop or mobile device
* Verify that you see at least two of the express checkout payment methods enabled in the front end
* Click on the "GPay" button
* This should open the Google Pay modal in test mode
* Choose a test address for a country we are currently allowing
* Verify that you see the shipping included and no errors, something like the following:
Google Pay - accepted shipping address

* Now change the address to a country that is not on the allowed list of countries.
* Verify that you see an error and any included shipping charge is dropped from the price:
Google Pay - invalid shipping address

_I tested the above using desktop Chrome and Safari on iOS._

### Apple Pay

* Use Safari to navigate to your shippable product in the store front-end
  - You may also need to be on a Mac or iOS -- I tested on MacOS and iOS
* Verify that you see at least two of the express checkout payment methods enabled in the front end
* Click on the Apple Pay button
* Verify that you get an Apple Pay modal
  - I already had Apple Pay configured, so the modal came up without any issues
* Either pick an address in a country that doesn't allow shipping, or add a new address in a country that doesn't allow shipping
* Verify that you see an error like the following for the invalid shipping country. (Note that the shipping price is removed from the total price.)
Apple Pay shipping address error

#### Ensure your email address is available in the Link test/sandbox environment

* Add a product to your cart, and proceed to checkout
* While in checkout, enter credit card details and fill out the section to save your details with Link, including the checkbox to save your details
* Make the purchase

#### Actually using link

* Navigate to your shippable product in the store front-end
* Verify that you see the Link express checkout method on the page
* Click on the Link button
* Choose or enter a test address for a country that allows shipping
* Verify that you don't see an error
* Now change the address to be in a different country that doesn't allow shipping
* Verify that you see an error about the shipping address
Link shipping address error

_I tested the above in Chrome on MacOS and Safari on iOS._

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


daledupreez:
I updated the testing instructions to include Link. 🎉

malithsen (approved):
Thanks for the fix and detailed testing instructions @daledupreez!

Tested all three express methods on product, cart and checkout pages.

I want to note that (although out of scope of this PR) with Google Pay, the shopper is allowed to click on the "Pay now" button even with the `Invalid shipping address.` error. Clicking on the Pay now button closes the model and another error is shown on the page. It would be nice to disable the button when there's an error but I am not certain if we have direct control over it.

CleanShot 2025-05-30 at 12 11 10

daledupreez:
> I want to note that (although out of scope of this PR) with Google Pay, the shopper is allowed to click on the "Pay now" button even with the `Invalid shipping address.` error. Clicking on the Pay now button closes the model and another error is shown on the page. It would be nice to disable the button when there's an error but I am not certain if we have direct control over it.

Thanks for the extensive testing, @malithsen! This does feel like it's an issue with the Google Pay UX rather than ours -- we definitely don't have much control over the content of the modal.

PR #4368, opened 2025-05-28
Fix division by zero fatal error

wjrosa:

This PR fixes a possible fatal error that occurs when the order item quantity is zero (causing an attempt to divide by zero). This error was seen on our Grafana logs:
```
PHP Fatal error:  Uncaught DivisionByZeroError: Division by zero in /srv/htdocs/wp-content/plugins/woocommerce-gateway-stripe/includes/abstracts/abstract-wc-stripe-payment-gateway.php:1456
```

The fix consists of changing the quantity amount to 1. Another possibility I considered was throwing an exception, but I decided on the other approach since this information is sent to Stripe to create the payment intent.

## Testing instructions

Since this is an edge case, code review should be enough.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


malithsen (commented):
@wjrosa Thanks for working on this!

While setting the quantity to 1 when it's 0 prevents the fatal error, I am worried that it masks the root cause and potentially introduces data integrity errors. It's possible that the quantity is >1, but gets set to 0 somewhere. Could also be a mishandled flow with inventory management.

What do you think of logging the order ID and the quantity when this happens, so we can try to get to the root of it?

wjrosa:
> @wjrosa Thanks for working on this!
> 
> While setting the quantity to 1 when it's 0 prevents the fatal error, I am worried that it masks the root cause and potentially introduces data integrity errors. It's possible that the quantity is >1, but gets set to 0 somewhere. Could also be a mishandled flow with inventory management.
> 
> What do you think of logging the order ID and the quantity when this happens, so we can try to get to the root of it?

Thank you for your review. And that's a great idea! I have added the log to my latest commits 👍

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -1456,9 +1457,15 @@ function ( $item ) use ( $currency ) {
...
 				$product_description = substr( $item->get_name(), 0, 26 );
 				$quantity            = $item->get_quantity();
-				$unit_cost           = WC_Stripe_Helper::get_stripe_amount( ( $subtotal / $quantity ), $currency );
-				$tax_amount          = WC_Stripe_Helper::get_stripe_amount( $item->get_total_tax(), $currency );
-				$discount_amount     = WC_Stripe_Helper::get_stripe_amount( $subtotal - $item->get_total(), $currency );
+				if ( ! $quantity ) {
+					WC_Stripe_Logger::error(
+						"Stripe Level 3 data: Order item with ID {$item->get_id()} from order ID {$order_id} has no quantity set. Defaulting to 1."
+					);
+					$quantity = 1; // Default to 1 if quantity is not set.
```
malithsen: @wjrosa Thanks for adding the logs.

What do you think of not setting it to one and letting it error out? I wonder if we will be creating data integrity errors down the line. The real quantity could be greater than one, and due to some unknown issue, it's zero here. If we assume it's 1, we will be sending incorrect data to Stripe.

wjrosa: Oh right. Do you mean closing this PR and leave it as-is, or adding an exception to be thrown on these cases? This could at least make our code safer against the fatals. I have added the exception for now. Let me know if you think that is a good idea 👍

malithsen: I was thinking of just keeping the logs without `quantity=1`, but adding an exception makes more sense along with the logs. Thanks Wesley!

malithsen (approved):

PR #4328, opened 2025-05-16
Adding new banners for BNPL to show upon legacy checkout deprecation

wjrosa:
Fixes STRIPE-430
P2: pcUrMV-b0q-p2
Figma: nFI8jCCAQnd19PNwScqVU6-fi

## Changes proposed in this Pull Request:

We are introducing a new promotional banner surface to promote the Buy Now, Pay Later payment methods. This banner will show up for merchants using the new checkout experience that don't have BNPLs enabled, and after the legacy checkout is deprecated.

The banner should be dismissible. And should send merchants to `https://woocommerce.com/document/stripe/setup-and-configuration/additional-payment-methods/` upon clicking the "Learn more" link.

Preview:
Screenshot 2025-05-29 at 18 13 42

I am also taking the opportunity to improve how we organize the banner components' files.

## Testing instructions

- Checkout and build this branch on your test environment (`add/new-banners-for-bnpl-upon-legacy-checkout-deprecation`)
- As a merchant, enable the legacy checkout experience

### Rendering

- Confirm the banner is not shown yet
- Disable the legacy checkout experience
- Disable BNPL methods (Klarna, Affirm, Afterpay/Clearpay)
- Confirm the new banner is displayed in the payment methods page (`wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=methods`)
- Enable any BNPL method
- Confirm the banner is gone

### Visuals

- Confirm the visuals match the Figma file and the preview above

### Dismissing

- Click the "Dismiss" link
- Confirm the banner is gone

### Learn more link

- Click the "Learn more" link
- Confirm it sends you to: https://woocommerce.com/document/stripe/setup-and-configuration/additional-payment-methods/

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4382, opened 2025-05-30
Adding a legacy deprecation feature flag

wjrosa:

This PR just introduces a new feature flag based on the current Stripe extension version (9.6.0+) to determine whether the legacy checkout experience is deprecated. This will be useful to implement code specific to the deprecation work, such as the [new promotional banner](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4328).

## Testing instructions

Code review should be enough. This feature flag has not yet been implemented in the code.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-feature-flags.php
```diff
@@ -189,4 +194,13 @@ public static function is_spe_available() {
...
 		return 'yes' === self::get_option_with_default( self::OC_FEATURE_FLAG_NAME );
 	}
+
+	/**
+	 * Checks whether the legacy checkout is deprecated.
+	 *
+	 * @return bool
+	 */
+	public static function is_legacy_checkout_deprecated() {
+		return WC_STRIPE_VERSION >= self::LEGACY_CHECKOUT_DEPRECATION_VERSION;
```
malithsen:
```suggestion
		return version_compare( WC_STRIPE_VERSION, self::LEGACY_CHECKOUT_DEPRECATION_VERSION, '>=' );
```

Let's use `version_compare()` rather than a direct string numeric comparison. This should guards against weird edge cases, especially with patch versions.

malithsen: On second thoughts, why is the version comparison necessary? Could we simply return `true` since we will be deprecating legacy features in a major version that includes this change?

wjrosa: This just allows us to ship this change before the 9.6 version. But I am OK with waiting for the specific version. In that case, instead of hardcoding the value, I would just not include this change at all (and [display the banner by default](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4328), for example).

malithsen: > This just allows us to ship this change before the 9.6 version

Ah thanks, that's the context I was missing. Let's just update it to use `version_compare` then.

wjrosa: Alright. I applied the change already in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4382/commits/cc0d05f4d2758262c9436e625f2ee36c5e6bfdd2

malithsen:
@wjrosa Sorry I missed your review request earlier. 
The diff doesn't look right anymore, maybe we want to rebase these changes. 
Also some of the required tests are failing.
The flag itself makes sense to me if we are planning on making another release before the legacy deprecation.

wjrosa:
I forgot to share an update here. I chatted with @Mayisha in p1748620729762169-slack-C055WHLA98D, and we decided to move forward with the banner enabled by default on the target version. So this PR is no longer necessary. Sorry for taking some of your time!

PR #4385, opened 2025-06-02
Migrate to new checkout experience

Mayisha:
Fixes STRIPE-492

- Updated the `is_upe_checkout_enabled` function to always return `true`
- On plugin update,
    - the value of `upe_checkout_experience_enabled` is getting set as `yes`
    - tracking the migration 

## Testing instructions
- Create a JN site and install the latest Stripe plugin.
- In the settings page, enable the legacy checkout experience.
- As a shopper, place an order with card payment method.
- As a shopper, purchase a subscription with a card
- Now build this branch locally and upload to your JN site.
- Go to the Stripe settings page and confirm that you now have the new checkout experience. 
- As a shopper, place another order with card payment method. It should be successful.
- As an admin, renew the previous subscription from the subscription edit page. It should be successful.
- Add the following filter in your site

```
add_filter( 'wc_stripe_is_upe_checkout_enabled', '__return_false' );
```

- Go to the Stripe settings page and confirm that you now have the legacy checkout experience. 
- As a shopper, place another order with card payment method. It should be successful.

### Note
- Legacy E2E tests are removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4422


includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -402,66 +402,64 @@ public function payment_methods_check_environment() {
 		$is_stripe_settings_page = isset( $_GET['page'], $_GET['section'] ) && 'wc-settings' === $_GET['page'] && 0 === strpos( $_GET['section'], 'stripe' );
 		$currency_messages       = '';
 
-		foreach ( $payment_methods as $method => $class ) {
-			$gateway = new $class();
+		if ( WC_Stripe_Feature_Flags::is_upe_checkout_enabled() ) {
```
Mayisha, line 405: Wrapped the existing code in if/else block. No other changes.

tests/phpunit/Admin/WC_REST_Stripe_Settings_Controller_Test.php
```diff
@@ -261,45 +261,27 @@ public function test_enum_fields( $rest_key, $option_name, $original_valid_value
...
 	public function test_individual_payment_method_settings() {
-		// Disable UPE and set up EPS gateway.
 		update_option(
 			'woocommerce_stripe_settings',
 			[
-				'enabled'     => 'yes',
-				'title'       => 'Credit card',
-				'description' => 'Pay with Credit card',
-				WC_Stripe_Feature_Flags::UPE_CHECKOUT_FEATURE_ATTRIBUTE_NAME => 'no',
+				'enabled' => 'yes',
```
Mayisha, line 267: The test was written for legacy before. Changed it to run for upe.

tests/phpunit/WC_Stripe_REST_UPE_Flag_Toggle_Controller_Test.php
```diff
@@ -1,104 +0,0 @@
-<?php
```
Mayisha, line 1: This REST controller is not needed anymore. Removing only the tests for now.

tests/phpunit/WC_Stripe_Inbox_Notes_Test.php
```diff
@@ -59,13 +59,6 @@ public function tear_down() {
 		parent::tear_down();
 	}
 
-	public function test_create_upe_availability_note() {
```
Mayisha, line 62: This note used to introduce UPE as a new feature. Removing the test as it is unnecessary now.

includes/class-wc-stripe-feature-flags.php
```diff
@@ -145,9 +145,14 @@ public static function is_upe_preview_enabled() {
...
 	 */
 	public static function is_upe_checkout_enabled() {
-		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
-		return ! empty( $stripe_settings[ self::UPE_CHECKOUT_FEATURE_ATTRIBUTE_NAME ] )
-			&& 'yes' === $stripe_settings[ self::UPE_CHECKOUT_FEATURE_ATTRIBUTE_NAME ];
+		/**
+		 * Temporary filter to allow rollback to legacy checkout experience.
+		 *
+		 * @since 9.6.0
+		 * @deprecated This filter will be removed in a future version. New checkout experience will always be enabled.
```
wjrosa: Do we have the specific version in which the filter will be removed? I think it would be nice to use it here.

wjrosa (approved):
Nice job! Code is good and it works as expected:

> * Go to the Stripe settings page and confirm that you now have the new checkout experience.

Screenshot 2025-06-23 at 10 12 44

> * As a shopper, place another order with card payment method. It should be successful.

Screenshot 2025-06-23 at 10 14 36

> * As an admin, renew the previous subscription from the subscription edit page. It should be successful.

Screenshot 2025-06-23 at 10 20 29

Screenshot 2025-06-23 at 10 21 23

> * Go to the Stripe settings page and confirm that you now have the legacy checkout experience.

Screenshot 2025-06-23 at 10 27 09

> * As a shopper, place another order with card payment method. It should be successful.

Screenshot 2025-06-23 at 10 28 41

includes/admin/class-wc-stripe-admin-notices.php
annemirasol: Optional: we can probably just remove the Sofort notice as it's been almost 2 years since.

annemirasol (approved):
Not a blocker, but I noticed that plugin update/legacy forced migration will not migrate payment method settings -- it will follow whatever the PMC says (if the store is PMC enabled, which we expect for most stores to be). 

For example, I had card and Apple Pay/Google Pay enabled before the plugin update. Afterwards, I had card and Klarna enabled, because it's what my PMC has.

And maybe we want it that way... but wondering if we should communicate in some way, post-plugin update, that merchants need to check their payment method settings?

The same behavior would be observed had merchants manually switched to UPE, by the way. But they would already be in the settings page, and there is a flash notice at the bottom about checking their payment methods.

I also noticed there's [code hooked into options pre-update](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/51016cab4fb4764d1cd1b75ae7a7a84431a4f779/includes/class-wc-stripe.php#L602), and at first I thought we can and should remove it, but I realized we should keep it in the event the store cannot do PMC sync.

includes/admin/class-wc-stripe-admin-notices.php
Mayisha: Removed in [ffd8c45](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4385/commits/ffd8c45668ed7c70afa68a022204f4e283961ddc)

includes/class-wc-stripe-feature-flags.php
Mayisha: Added in [17682c5](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4385/commits/17682c5235c54ce77675b7f6b46bf82e7438dd90)

Mayisha:
Thanks @annemirasol for the PMC details. I will check that separately and merge this PR first. 👀

PR #4388, opened 2025-06-02
Fix payment method detachment when deleting user via CLI

### LINKED ISSUE #4335: Check WP environment when detaching payment methods after deleting users in CLI
Currently when a user is deleted, the plugin detaches their payment methods via the Stripe API.  To prevent this causing issues on staging environments which are still connected to the live Stripe account, there is a check to see if we are in admin section, and if so if we are in a production environment.

If we are deleting users via WP CLI, the check to `is_admin()` will return false, and the user will be detached regardless of the environment.  This is not ideal, as I have just found out while bulk deleting users from a staging environment.

In `class-wc-stripe-api.php::479` we have:

```
	public static function should_detach_payment_method_from_customer() {
		// If we are in test mode, we can always detach the payment method.
		if ( WC_Stripe_Mode::is_test() ) {
			return true;
		}

		// Return true for the delete user request from the admin dashboard when the site is a production site
		// and return false when the site is a staging/local/development site.
		// This is to avoid detaching the payment method from the live production site.
		// Requests coming from the customer account page i.e delete payment method, are not affected by this and returns true.
		if ( is_admin() ) {
			if ( 'production' === wp_get_environment_type() ) {
				return true;
			} else {
				return false;
			}
		}

		return true;
	}
```

I suggest checking for `( defined( 'WP_CLI' ) && WP_CLI )` as well as the check for `is_admin()` before checking for environment:

```
		if ( is_admin() ||  ( defined( 'WP_CLI' ) && WP_CLI )) {
			if ( 'production' === wp_get_environment_type() ) {
				return true;
			} else {
				return false;
			}
		}
```

malithsen:
Fixes STRIPE-466
Fixes #4335

This PR addresses an issue where payment methods were detached from Stripe customers when deleting users via WP-CLI in non-production environments. Previously, the code only checked for `is_admin()` to determine if the request was coming from the admin interface, which meant WP-CLI requests would bypass the environment check and always detach payment methods.
The fix adds a check for WP-CLI requests using `defined( 'WP_CLI' ) && WP_CLI` alongside the existing `is_admin()` check. 

## Testing instructions
1. Add `define( 'WP_ENVIRONMENT_TYPE', 'staging' );` to your `wp-config.php`.
2. If your site is in test mode, disable [this](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/ab4fbe62c431150882aaf658252a2cc4f22539b9/includes/class-wc-stripe-api.php#L456) check.
3. As a shopper, create a new account. 
4. Go to My Account -> Payment methods and add a new payment method.
5. On your Stripe dashboard, under Customers, make sure the customer shows up with their payment method.
6. Remove the user using wp cli `wp user delete USER_ID`
7. On your Stripe dashboard, the payment method should not be removed from the shopper.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -29,6 +29,7 @@
 * Fix - Ensure that we migrate payment_request_button_size=medium on upgrade
 * Fix - Apply shipping country restrictions to Express Checkout
 * Dev - Prevent changelog entries with trailing periods
+* Fix - When the user is deleted via WP CLI, take into account the environment type before detaching their payment methods
 
 = 9.5.2 - 2025-05-22 =
 * Add - Implement custom database cache for persistent caching with in-memory optimization.
--- a/includes/class-wc-stripe-api.php
+++ b/includes/class-wc-stripe-api.php
@@ -457,11 +457,11 @@ class WC_Stripe_API {
 			return true;
 		}
 
-		// Return true for the delete user request from the admin dashboard when the site is a production site
+		// Return true for the delete user request from the admin dashboard or WP-CLI when the site is a production site
 		// and return false when the site is a staging/local/development site.
 		// This is to avoid detaching the payment method from the live production site.
 		// Requests coming from the customer account page i.e delete payment method, are not affected by this and returns true.
-		if ( is_admin() ) {
+		if ( is_admin() || ( defined( 'WP_CLI' ) && WP_CLI ) ) {
 			if ( 'production' === wp_get_environment_type() ) {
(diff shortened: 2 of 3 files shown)


Mayisha (approved):
Working as described 🎉 

- In `develop`, the payment method gets detached when I delete a user via cli.
- In this branch, the payment method is not detached when I delete the user via cli.

PR #4391, opened 2025-06-03
Adding the subscriptions detached notice as a WC debug tool

wjrosa:
Fixes STRIPE-499

As part of STRIPE-483, this PR introduces a new WooCommerce Debug Tool to list subscriptions with detached payment methods. This feature was previously used to [render an admin notice on every page load](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3863) and [removed due to performance issues](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4320).

Tool display in the WooCommerce -> Status -> Tools tab:
Screenshot 2025-06-03 at 18 50 27

Display of detached subscriptions:
Screenshot 2025-06-03 at 18 50 10

## Testing instructions

- Checkout this branch on your test environment (`add/subscriptions-detached-tool`)
- Connect your Stripe account
- Go to WooCommerce -> Status -> Tools and confirm you can see the new "List Stripe subscriptions with detached payment method" tool
- Create some subscriptions for any user (perform some purchases or add using the admin interface) *
- Comment out the lines handling the cache (`includes/compat/class-wc-stripe-subscriptions-helper.php:53` and `includes/compat/class-wc-stripe-subscriptions-helper.php:88`)
- Comment out the lines checking for the detachment, so all subscriptions can be returned (`includes/compat/class-wc-stripe-subscriptions-helper.php:74` and `includes/compat/class-wc-stripe-subscriptions-helper.php:77`, also comment out the closing brackets to avoid breaking the block) *
- In the Tools tab, click the "List subscriptions" button
- Confirm the newly added subscriptions are correctly listed inside a notice

\* You can also purchase subscriptions as a shopper and detach their payment method using the Stripe dashboard instead.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -24,23 +31,33 @@ public static function is_subscriptions_enabled() {
...
+	 * Loads all subscriptions, and attempts to return those that are detached from the customer.
+	 *
+	 * @param int $limit The maximum number of subscriptions to retrieve. Use -1 for no limit (default).
+	 * @return array
+	 */
+	public static function get_detached_subscriptions( $limit = -1 ) {
 		// Check if we have a cached result.
 		$cached_subscriptions = get_transient( self::DETACHED_SUBSCRIPTIONS_TRANSIENT_KEY );
-		if ( ! empty( $cached_subscriptions ) ) {
+		if ( is_array( $cached_subscriptions ) ) {
```
wjrosa, line 59: As @annemirasol pointed out, this should fix the lack of caching for an empty list.

includes/class-wc-stripe-status.php
```diff
@@ -194,4 +195,61 @@ public function render_status_report_section() {
...
+	public function debug_tools( $tools ) {
+		if ( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() ) {
+			$tools['list_detached_subscriptions'] = [
+				'name'     => __( 'List Stripe subscriptions with detached payment method', 'woocommerce-gateway-stripe' ),
+				'button'   => __( 'List subscriptions', 'woocommerce-gateway-stripe' ),
+				'desc'     => sprintf(
+					'%1$s<br/><strong class="red">%2$s</strong> %3$s',
+					__( 'This tool will list all Stripe subscriptions with detached payment methods.', 'woocommerce-gateway-stripe' ),
+					__( 'Note:', 'woocommerce-gateway-stripe' ),
+					__( 'Be aware this will perform a Stripe request for each subscription on your store. This can be very costly, depending on how many subscriptions you have.', 'woocommerce-gateway-stripe' ),
```
Mayisha: How about we rephrase this to something similar to this (AI rephrase)

`This tool will make an API request to Stripe for each subscription in your store. For stores with many subscriptions, this may temporarily impact performance.`

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -24,23 +31,33 @@ public static function is_subscriptions_enabled() {
...
+
+	/**
+	 * Loads all subscriptions, and attempts to return those that are detached from the customer.
+	 *
+	 * @param int $limit The maximum number of subscriptions to retrieve. Use -1 for no limit (default).
+	 * @return array
+	 */
+	public static function get_detached_subscriptions( $limit = -1 ) {
 		// Check if we have a cached result.
 		$cached_subscriptions = get_transient( self::DETACHED_SUBSCRIPTIONS_TRANSIENT_KEY );
```
Mayisha: We have a new cache class `WC_Stripe_Database_Cache`. Do you think we should use it here?

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -75,4 +89,43 @@ public static function get_some_detached_subscriptions() {
...
+			$detached_messages .= sprintf(
+			/* translators: %1$s is the subscription ID. %2$s is a customer payment method page. %3$s is the customer's page on Stripe */
+				__( '#%1$s: %2$s | %3$s<br/>', 'woocommerce-gateway-stripe' ),
+				$subscription['id'],
+				$customer_payment_method_link,
+				$customer_stripe_page
+			);
+		}
+		if ( ! empty( $detached_messages ) ) {
+			$detached_messages = __( 'Some subscriptions are missing payment methods, <strong>preventing renewals</strong>. Share the payment method page link with the customer to update it or manually set the Stripe Payment Method ID meta field in the subscriptions details\' "Billing" section to another from the customer\'s page on Stripe. Below are the last subscriptions affected and the links as mentioned earlier:<br />', 'woocommerce-gateway-stripe' ) . $detached_messages;
```
Mayisha: The action item part of the message looks confusing at first pass. I asked AI and got the following rephrase-

```
To fix this, either: 1) Share the payment method update link with the customer, or 2) Manually update the payment method in the subscription\'s billing details using a valid payment method from the customer\'s Stripe account. Below are the affected subscriptions and their update links:
```
Even if we do not use these exact words, the numerical pointing seems like a good idea to me.

Mayisha: Also how about we mention the count/number somewhere in the message.

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -24,23 +31,33 @@ public static function is_subscriptions_enabled() {
...
 			return $cached_subscriptions;
 		}
 
 		$subscriptions = wcs_get_subscriptions(
 			[
-				'subscriptions_per_page' => 50,
+				'subscriptions_per_page' => $limit,
 				'page'                   => 1,
 				'orderby'                => 'date',
 				'order'                  => 'DESC',
```
Mayisha: Given the recent API issues, what do you think about limiting the detachment checks to the subscriptions that have renewal due in the next 2 weeks or a month during this bulk check?

Mayisha: Also, should we check the active subscriptions only, discarding the on-hold and pending-cancel ones? They won't go through auto renewal anyway so no chance of failure.

Mayisha (commented):
I have a few questions/suggestions. Apart from that the functionality is working as described.

Screenshot 2025-06-05 at 8 59 39 PM

includes/class-wc-stripe-status.php
daledupreez: What about showing a dialog or confirm box for sites with subscriptions beyond a threshold, say 100 subscriptions?
Also, what happens if some subscriptions are not with Stripe?
Maybe we should have a recommendation _not_ to use the tool in this way if you have more than 100 or 200 subscriptions, and then recommend that they use the bulk selection tools.

wjrosa: Thanks for the suggestions, addressed them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/88b76cbe5f9216c9ad872fabb1b3d84d6fda2717.

> Also, what happens if some subscriptions are not with Stripe?

Those are ignored. We look for the `_stripe_source_id` meta.

> then recommend that they use the bulk selection tools

I will update the paragraph to mention it once we have this other part (in another PR).

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: That's a great idea! Replaced the transients in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/fa8632003f47ff3ec321fcd46ed34ac1e0adb398

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: > The action item part of the message looks confusing at first pass. I asked AI and got the following rephrase-

I liked the suggestion! Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/6265cb9e4a37f0f7bb74056a452dbcc9c33406e6

> Also how about we mention the count/number somewhere in the message.

Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/a142fa48370d3eaf4b845d47159feebed91d5805

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: Other great ideas!

> Given the recent API issues, what do you think about limiting the detachment checks to the subscriptions that have renewal due in the next 2 weeks or a month during this bulk check?

Put the 1 month limit in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/13738880e504bf91ff54fdf5e688bdb801e27b75

> Also, should we check the active subscriptions only, discarding the on-hold and pending-cancel ones? They won't go through auto renewal anyway so no chance of failure.

Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/2bf8f9326a38c1e457e47503fdac4b260c50ab6f

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -24,27 +31,37 @@ public static function is_subscriptions_enabled() {
...
+	 *
+	 * @param int $limit The maximum number of subscriptions to retrieve. Use -1 for no limit (default).
+	 * @return array
+	 */
+	public static function get_detached_subscriptions( $limit = -1 ) {
 		// Check if we have a cached result.
-		$cached_subscriptions = get_transient( self::DETACHED_SUBSCRIPTIONS_TRANSIENT_KEY );
-		if ( ! empty( $cached_subscriptions ) ) {
+		$cached_subscriptions = WC_Stripe_Database_Cache::get( self::DETACHED_SUBSCRIPTIONS_CACHE_KEY );
+		if ( is_array( $cached_subscriptions ) ) {
```
daledupreez: This implementation is going to be weird if we call this function with different values for `$limit`, as we'll inherit the previous value of `$limit`, regardless of whether it was larger or smaller than the current value of `$limit`. I think the cached data needs to include the value of `$limit`, and we need some logic in this code block to return the correct number of rows.

Also, we should return early when `$limit === 0`.
```suggestion
		if ( $limit === 0 ) {
			return array();
		}
		$cached_subscription_data = WC_Stripe_Database_Cache::get( self::DETACHED_SUBSCRIPTIONS_CACHE_KEY );
		if ( is_array( $cached_subscription_data ) && is_int( $cached_subscription_data['limit'] ?? null ) && is_array( $cached_subscription_data['subscriptions'] ?? null ) ) {
			$stored_limit = $cached_subscription_data['limit'];
			// If we have the same limit, or found no results, return the cached data
			if ( $limit === $stored_limit || ( array() === $cached_subscription_data['subscriptions'] ) ) {
				return $cached_subscription_data['subscriptions'];
			}
			// If the new limit is less than the stored limit, return the first $limit results from the cache.
			if ( $limit > 0 && ( $stored_limit === -1 || $limit < $stored_limit ) ) {
				return array_slice( $cached_subscription_data['subscriptions'], 0, $limit );
			}
			// Otherwise, $limit > $stored_limit, so we need to fetch more data
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -63,16 +85,66 @@ public static function get_some_detached_subscriptions() {
...
-					if ( count( $detached_subscriptions ) >= 5 ) {
-						break;
-					}
 				}
 			}
 		}
 
 		// Cache the result for a day.
-		set_transient( self::DETACHED_SUBSCRIPTIONS_TRANSIENT_KEY, $detached_subscriptions, DAY_IN_SECONDS );
+		WC_Stripe_Database_Cache::set( self::DETACHED_SUBSCRIPTIONS_CACHE_KEY, $detached_subscriptions, DAY_IN_SECONDS );
```
daledupreez: As noted above, we should store the value of `$limit` alongside the cached data.
```suggestion
		WC_Stripe_Database_Cache::set( self::DETACHED_SUBSCRIPTIONS_CACHE_KEY, array( 'limit' => $limit, 'subscriptions' => $detached_subscriptions ), DAY_IN_SECONDS );
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -63,16 +85,66 @@ public static function get_some_detached_subscriptions() {
...
+				_n(
+					'%s subscription is missing the payment method, <strong>preventing renewals</strong>. ',
+					'%s subscriptions are missing payment methods, <strong>preventing renewals</strong>. ',
+					count( $subscriptions ),
+					'woocommerce-gateway-stripe'
+				),
+				count( $subscriptions )
+			); /* phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped */
+			// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
+			$intro_message    .= __( 'To fix this, either: 1) Share the payment method page link with the customer to update it, or 2) Manually update the payment method in the subscription\'s billing details using a valid payment method from the customer\'s Stripe account. ', 'woocommerce-gateway-stripe' );
```
daledupreez: Style question: is it worth using `"` quotes to avoid using `\'` in the string contents? The `\` characters make it harder to search for string matches.
```suggestion
			$intro_message    .= __( "To fix this, either: 1) Share the payment method page link with the customer to update it, or 2) Manually update the payment method in the subscription's billing details using a valid payment method from the customer's Stripe account. ", 'woocommerce-gateway-stripe' );
```

daledupreez: Separately, might it be worth adding explicit `<br />` elements so the actions are super-clear?
```suggestion
			$intro_message    .= __( 'To fix this, either: <br />1) Share the payment method page link with the customer to update it, or <br />2) Manually update the payment method in the subscription\'s billing details using a valid payment method from the customer\'s Stripe account. ', 'woocommerce-gateway-stripe' );
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -63,16 +85,66 @@ public static function get_some_detached_subscriptions() {
...
 	}
+
+	/**
+	 * Builds a string containing messages about subscriptions that are detached from the customer.
+	 *
+	 * @param $subscriptions array An array of subscriptions that are detached from the customer.
+	 * @return string A string containing the messages to be displayed in the admin interface.
+	 */
+	public static function build_subscriptions_detached_messages( $subscriptions = [] ) {
+		$detached_messages = '';
```
daledupreez, line 160: Non-blocking thought: is it worth adding an early return when `$subscriptions` is empty?
```suggestion
		$detached_messages = '';
		if ( empty( $subscriptions ) || ! is_array( $subscriptions ) ) {
			return '';
		}
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -54,6 +71,11 @@ public static function get_some_detached_subscriptions() {
 				continue;
 			}
 
+			// Filter subscriptions not renewing in the next month
```
daledupreez, line 93: Not blocking: if we plan to allow users to pre-select subscriptions, e.g. from the admin, we probably want to separate this filtering step from the logic that checks each subscription. So we'd probably want an `is_subscription_detached( $subscription )` function and a `get_detached_subscription_data( $subscription )` function.

But it may be worth tackling that in the PR where we add that functionality.

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -54,6 +71,11 @@ public static function get_some_detached_subscriptions() {
...
+			if ( $subscription->get_time( 'next_payment' ) > ( time() + MONTH_IN_SECONDS ) ) {
```
daledupreez: Not blocking: Just to be extra safe, should we check for today + 1 month + 1 day? That way we catch subscriptions that renew in one month but later in the day.
```suggestion
			if ( $subscription->get_time( 'next_payment' ) > ( time() + MONTH_IN_SECONDS + DAY_IN_SECONDS ) ) {
```

tests/phpunit/Compat/WC_Stripe_Subscriptions_Helper_Test.php
```diff
@@ -21,4 +23,124 @@ class WC_Stripe_Subscriptions_Helper_Test extends WP_UnitTestCase {
...
+	 *
+	 * @return void
+	 */
+	public function test_get_detached_subscriptions() {
+		add_filter(
+			'wc_stripe_unit_test_get_subscription_time_next_payment_date',
+			function () {
+				return strtotime( '+1 week' );
+			}
+		);
```
daledupreez: Style-wise, might it be clearer to use a local variable so it's super-explicit that we're adding and removing the same function to the filter?
```suggestion
		$return_now_plus_one_week = function () {
			return strtotime( '+1 week' );
		};
		add_filter(
			'wc_stripe_unit_test_get_subscription_time_next_payment_date',
			$return_now_plus_one_week
		);
```

tests/phpunit/Compat/WC_Stripe_Subscriptions_Helper_Test.php
```diff
@@ -21,4 +23,124 @@ class WC_Stripe_Subscriptions_Helper_Test extends WP_UnitTestCase {
...
+			];
+		};
+
+		add_filter( 'pre_http_request', $test_request, 10, 3 );
+
+		$expected = [
+			[
+				'id'                        => $subscription_id,
+				'customer_id'               => $customer_id,
+				'change_payment_method_url' => $subscription->get_change_payment_method_url(),
```
daledupreez: Is it worth testing that we get a specific value for this, possibly by modifying the value in the `wcs_get_change_payment_method_url` filter?

daledupreez (commented):
I haven't retested, but I have some high-level concerns that I want to flag before testing.

#### The admin/merchant experience

I think we should assume people will run this tool on hundreds of subscriptions. That may or may not be intentional, but the user experience will be bad if we're making hundreds of API requests to Stripe, and the HTTP request either reaches a server response timeout or looks like it stalled.

I think we should make this work by default, even when run on larger sites. So I think we should do something like the following:
* Work out how long these API requests to Stripe end up taking, on average
* Use that average to set a default maximum number of subscriptions to check so we get a response in less than 60 seconds -- we should be conservative here, as some sites will have additional latency when making requests to Stripe's APIs
* Add a filter that uses that number as the default maximum number of subscriptions to check, e.g. `wc_stripe_detached_subscriptions_maximum_count`
* Add a filter that also applies a maximum timeout of 55 seconds, and make sure we respond within that timeframe when processing detached subscriptions, e.g. `wc_stripe_detached_subscriptions_maximum_time`
  - If we can detect the server response timeout, we could use 5 seconds less than that value.
* We should specify these limits and the related filters in the upfront message, so users' expectations are set correctly
* If we exceed the maximum number or maximum time, we can include messaging to state that we did so

The cumulative effect of all these changes should be as follows:
 * If people run the tool on a site that has too many subscriptions or too much latency, we trip over the respective limit, and give them feedback about the limit we hit, instead of failing and giving them nothing but frustration at waiting for something that times out.
 * If more experienced merchants or developers want to increase either limit, they can do so via the filters, but they are then making explicit changes that we can warn them against.
 * When we implement logic to allow bulk subscription checks via the subscriptions UI, we can re-use the filters above to continue keeping the experience usable rather than complete.

#### `Fixes` keyword in description

I just checked, and STRIPE-483 includes multiple sub-items -- we shouldn't use `Fixes` for any one of the related PRs. It may be better to create sub-issues for each piece of the work, and then mark each of those as complete via the respective PRs.

tests/phpunit/Compat/WC_Stripe_Subscriptions_Helper_Test.php
wjrosa: I personally don't think it makes much sense: the testing for `get_change_payment_method_url` should be done by the subscriptions extension. But I added it anyway in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/cb91b22731853d113620ef501bd76aa8f37ef117

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: Yeah, I will tackle this on another PR. I will probably not use this method anyway (since we will retrieve a list of specific IDs).

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: > Style question: is it worth using " quotes to avoid using \' in the string contents? The \ characters make it harder to search for string matches.

Previously, I remember we had always used single quotes for strings. I don't know if that changed here or if it was in another repository. But I have updated it manually now anyway in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/8ae96ebffbc75e7ee0bab52110552a73a84bd14d (git does not allow me to apply changes to outdated files)

> Separately, might it be worth adding explicit <br /> elements so the actions are super-clear?

Makes sense! Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/4ee7ff65f805d52a807c7ba593cbb343203c15e3

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: Your comment made me realise the limit implementation was incorrect. The limit should refer to the returned subscriptions, not the number retrieved from the database. I have fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/be1c230e1ba06ea16c871e4d92888334b2f89c8d

I have also used a better approach in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/a6c4bef5a86f940df813273329dd3e18c4a869d6, using the cache key as a prefix and appending the limit to it instead. I think it is better because we don't need to store and retrieve everything in a single key and fetch the whole array every time.

wjrosa:
> I haven't retested, but I have some high-level concerns that I want to flag before testing.
> 
> #### The admin/merchant experience
> I think we should assume people will run this tool on hundreds of subscriptions. That may or may not be intentional, but the user experience will be bad if we're making hundreds of API requests to Stripe, and the HTTP request either reaches a server response timeout or looks like it stalled.
> 
> I think we should make this work by default, even when run on larger sites. So I think we should do something like the following:
> 
> * Work out how long these API requests to Stripe end up taking, on average
> * Use that average to set a default maximum number of subscriptions to check so we get a response in less than 60 seconds -- we should be conservative here, as some sites will have additional latency when making requests to Stripe's APIs
> * Add a filter that uses that number as the default maximum number of subscriptions to check, e.g. `wc_stripe_detached_subscriptions_maximum_count`
> * Add a filter that also applies a maximum timeout of 55 seconds, and make sure we respond within that timeframe when processing detached subscriptions, e.g. `wc_stripe_detached_subscriptions_maximum_time`
>   
>   * If we can detect the server response timeout, we could use 5 seconds less than that value.
> * We should specify these limits and the related filters in the upfront message, so users' expectations are set correctly
> * If we exceed the maximum number or maximum time, we can include messaging to state that we did so
> 
> The cumulative effect of all these changes should be as follows:
> 
> * If people run the tool on a site that has too many subscriptions or too much latency, we trip over the respective limit, and give them feedback about the limit we hit, instead of failing and giving them nothing but frustration at waiting for something that times out.
> * If more experienced merchants or developers want to increase either limit, they can do so via the filters, but they are then making explicit changes that we can warn them against.
> * When we implement logic to allow bulk subscription checks via the subscriptions UI, we can re-use the filters above to continue keeping the experience usable rather than complete.
> 
> #### `Fixes` keyword in description
> I just checked, and [STRIPE-483](https://linear.app/a8c/issue/STRIPE-483) includes multiple sub-items -- we shouldn't use `Fixes` for any one of the related PRs. It may be better to create sub-issues for each piece of the work, and then mark each of those as complete via the respective PRs.

I have created a new issue to handle the filters you suggested: STRIPE-502. I have included it under the main issue STRIPE-483, and added sub-issues for each part of the change 👍

tests/phpunit/Compat/WC_Stripe_Subscriptions_Helper_Test.php
```diff
@@ -21,4 +23,134 @@ class WC_Stripe_Subscriptions_Helper_Test extends WP_UnitTestCase {
...
+	/**
+	 * Test for `get_detached_subscriptions`.
+	 *
+	 * @return void
+	 */
+	public function test_get_detached_subscriptions() {
+		$return_now_plus_one_week = function () {
+			return strtotime( '+1 week' );
+		};
+		add_filter(
```
Mayisha: Instead of adding a test-specific filter for setting the `next_payment` date, we could use `$subscription->set_next_payment_date()`.

includes/class-wc-stripe-status.php
```diff
@@ -194,4 +195,62 @@ public function render_status_report_section() {
...
+		if ( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() ) {
+			$tools['list_detached_subscriptions'] = [
+				'name'     => __( 'List Stripe subscriptions with detached payment method', 'woocommerce-gateway-stripe' ),
+				'button'   => __( 'List subscriptions', 'woocommerce-gateway-stripe' ),
+				'desc'     => sprintf(
+					'%1$s<br/><strong class="red">%2$s</strong> %3$s<br/><strong>%4$s</strong>',
+					__( 'This tool will list all Stripe subscriptions with detached payment methods.', 'woocommerce-gateway-stripe' ),
+					__( 'Note:', 'woocommerce-gateway-stripe' ),
+					__( 'This tool will make an API request to Stripe for each Stripe subscription in your store. For stores with many subscriptions, this may temporarily impact performance.', 'woocommerce-gateway-stripe' ),
+					__( 'Not recommended if you have more than 100 subscriptions.', 'woocommerce-gateway-stripe' ),
```
Mayisha: This 'not recommended' message is a bit misleading. In general, most of the sites are supposed to have more than 100 subscriptions and this note might demotivate them from using this tool.
We are checking only the `active` subscriptions with the next payment date within the next month. I think, you meant the number of those subscriptions only (correct me if I am wrong).
How about adding some more details like - 'Not recommended if you have more than 100 active subscriptions due for renewal within 30 days'

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -24,36 +36,65 @@ public static function is_subscriptions_enabled() {
...
-		);
+		$subscriptions = [];
+		$page          = 1;
+		$per_page      = self::MAX_SUBSCRIPTIONS_PER_PAGE;
+
+		do {
+			$batch             = wcs_get_subscriptions(
+				[
+					'subscriptions_per_page' => $per_page,
+					'page'                   => $page,
```
Mayisha:
```suggestion
					'paged'                   => $page,
```

The argument name should be `paged` ([ref](https://github.com/woocommerce/woocommerce-subscriptions/blob/19b8edeff75b558c15ebc8aa9f285736e11573f1/includes/core/wcs-functions.php#L420C4-L420C9))

I discovered it as I changed the `MAX_SUBSCRIPTIONS_PER_PAGE` value to `10`. This caused a fatal error in my site.

Screenshot 2025-06-23 at 2 01 45 PM

Mayisha (changes requested):
Mostly working well.

Screenshot 2025-06-23 at 1 35 18 PM

I have a couple of minor comments and a solution to fix a fatal error. The fatal should be fixed as sites having more than `MAX_SUBSCRIPTIONS_PER_PAGE` and requiring pagination will encounter the fatal error.

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: Nice catch, thanks! Suggestion applied ✅

includes/class-wc-stripe-status.php
wjrosa: You're right, Mayisha. I have updated the message in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/b0d290586b2c1bb072fbc762fd1d296bc691b5f6

tests/phpunit/Compat/WC_Stripe_Subscriptions_Helper_Test.php
wjrosa: That's much smarter! Implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/e35e1e74f259b8059460eb5a1b64e301123f3c7e

Mayisha (approved):
Thanks for the updates @wjrosa. 👍

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
+* Add - New WooCommerce Debug Tool to list subscriptions without a payment method attached.
```
daledupreez: Nit: trailing `.`
```suggestion
* Add - New WooCommerce Debug Tool to list subscriptions without a payment method attached
```

includes/class-wc-stripe-status.php
```diff
@@ -194,4 +195,62 @@ public function render_status_report_section() {
...
 	}
+
+	/**
+	 * Add Stripe tools to the Woo debug tools.
+	 *
+	 * @param array $tools List of current available tools.
+	 */
+	public function debug_tools( $tools ) {
+		if ( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() ) {
+			$tools['list_detached_subscriptions'] = [
```
daledupreez: Nit: should we prefix this key to make collisions way less likely?
```suggestion
			$tools['wc_stripe_list_detached_subscriptions'] = [
```

includes/class-wc-stripe-status.php
```diff
@@ -194,4 +195,62 @@ public function render_status_report_section() {
...
+				'name'     => __( 'List Stripe subscriptions with detached payment method', 'woocommerce-gateway-stripe' ),
+				'button'   => __( 'List subscriptions', 'woocommerce-gateway-stripe' ),
+				'desc'     => sprintf(
+					'%1$s<br/><strong class="red">%2$s</strong> %3$s<br/><strong>%4$s</strong>',
+					__( 'This tool will list all Stripe subscriptions with detached payment methods.', 'woocommerce-gateway-stripe' ),
+					__( 'Note:', 'woocommerce-gateway-stripe' ),
+					__( 'This tool will make an API request to Stripe for each Stripe subscription in your store. For stores with many subscriptions, this may temporarily impact performance.', 'woocommerce-gateway-stripe' ),
```
daledupreez: Should we refine this message a bit to make it clearer which subscriptions we're going to be checking?
```suggestion
					__( 'This tool will make an API request to Stripe for each active Stripe subscription in your store that is due to renew in the next month. For stores with many subscriptions, this may temporarily impact performance.', 'woocommerce-gateway-stripe' ),
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -24,36 +36,65 @@ public static function is_subscriptions_enabled() {
...
 			}
 
+			// Filter subscriptions not renewing in the next month
+			if ( $subscription->get_time( 'next_payment' ) > ( time() + MONTH_IN_SECONDS + DAY_IN_SECONDS ) ) {
+				continue;
+			}
+
 			$source_id = $subscription->get_meta( '_stripe_source_id' );
 			if ( $source_id ) {
 				$payment_method = WC_Stripe_API::get_payment_method( $source_id );
```
daledupreez: Should we cache this value so we're not doing unnecessary work we run this multiple times?

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -14,6 +14,18 @@ class WC_Stripe_Subscriptions_Helper {
 	 */
 	private const DETACHED_SUBSCRIPTIONS_TRANSIENT_KEY = 'wcstripe_detached_subscriptions';
```
daledupreez: Given that we're now using the database cache, should we rename the constant and remove the prefix?
```suggestion
	private const DETACHED_SUBSCRIPTIONS_CACHE_PREFIX = 'detached_subscriptions';
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -103,4 +141,59 @@ public static function is_manual_renewal_enabled() {
...
+				esc_url( self::STRIPE_CUSTOMER_PAGE_BASE_URL . $subscription['customer_id'] ),
+				esc_html(
+				/* translators: this is a text for a link pointing to the customer's page on Stripe */
+					__( 'Stripe customer page &rarr;', 'woocommerce-gateway-stripe' )
+				)
+			);
+			$detached_messages .= sprintf(
+			/* translators: %1$s is the subscription ID. %2$s is a customer payment method page. %3$s is the customer's page on Stripe */
+				__( '#%1$s: %2$s | %3$s<br/>', 'woocommerce-gateway-stripe' ),
+				$subscription['id'],
```
daledupreez: I think we should wrap this value in `esc_html()` just to be safe.
```suggestion
				esc_html( $subscription['id'] ),
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -103,4 +141,59 @@ public static function is_manual_renewal_enabled() {
...
+
+		// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
+		$intro_message = sprintf(
+		/* translators: %s: subscriptions count */
+			_n(
+				'%s subscription is missing the payment method, <strong>preventing renewals</strong>. ',
+				'%s subscriptions are missing payment methods, <strong>preventing renewals</strong>. ',
+				count( $subscriptions ),
+				'woocommerce-gateway-stripe'
+			),
```
daledupreez: Can we approach this in a different way that allows us to HTML-escape the output? It feels like we should always  escape the output, even if it is "just" translations.

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -103,4 +141,59 @@ public static function is_manual_renewal_enabled() {
...
+			count( $subscriptions )
+		); /* phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped */
+		// phpcs:disable WordPress.Security.EscapeOutput.OutputNotEscaped
+		$intro_message .= __( "To fix this, either: <br />1) Share the payment method page link with the customer to update it, or <br />2) Manually update the payment method in the subscription's billing details using a valid payment method from the customer's Stripe account. ", 'woocommerce-gateway-stripe' );
+		$intro_message .= __( 'Below are the affected subscriptions and their update links:<br />', 'woocommerce-gateway-stripe' );
```
daledupreez: As above, it feels like we should separate out the formatting and translation, because it's currently preventing us from escaping the output.

readme.txt
```diff
@@ -112,6 +112,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 
 = 9.6.0 - xxxx-xx-xx =
 
+* Add - New WooCommerce Debug Tool to list subscriptions without a payment method attached.
```
daledupreez: Nit: trailing `.`
```suggestion
* Add - New WooCommerce Debug Tool to list subscriptions without a payment method attached
```

tests/phpunit/Compat/WC_Stripe_Subscriptions_Helper_Test.php
```diff
@@ -43,4 +45,122 @@ public function test_is_manual_renewal_enabled() {
...
+		$expected = [
+			[
+				'id'                        => $subscription_id,
+				'customer_id'               => $customer_id,
+				'change_payment_method_url' => $mocked_payment_method_url,
+			],
+		];
+		$this->assertEquals( $expected, WC_Stripe_Subscriptions_Helper::get_detached_subscriptions() );
+
+		remove_filter( 'pre_http_request', $test_request, 10, 3 );
```
daledupreez: To avoid failed assertions leaving the `pre_http_request` filter set up, might it make sense to handle the cleanup before the assertion?
```suggestion
		$result = WC_Stripe_Subscriptions_Helper::get_detached_subscriptions();

		// Clean up before we run any assertions
		remove_filter( 'pre_http_request', $test_request, 10, 3 );

		$this->assertEquals( $expected, $result );

```

tests/phpunit/WC_Stripe_Status_Test.php
```diff
@@ -51,4 +56,115 @@ public function test_render_status_report_section() {
...
+
+		$account = $this->getMockBuilder( WC_Stripe_Account::class )
+			->disableOriginalConstructor()
+			->getMock();
+
+		$status = new WC_Stripe_Status( $gateway, $account );
+
+		$tools = $status->debug_tools( [] );
+
+		$this->assertArrayHasKey( 'list_detached_subscriptions', $tools );
```
daledupreez: Is it worth checking for any additional values or data in `$tools['list_detached_subscriptions']`? It feels like it could be good to check some basic fields.

tests/phpunit/WC_Stripe_Status_Test.php
```diff
@@ -51,4 +56,115 @@ public function test_render_status_report_section() {
...
+
+		ob_start();
+		$status->list_detached_subscriptions();
+		$output = ob_get_clean();
+
+		$this->assertStringContainsString( $expected_output, $output );
+
+		WC_Subscriptions_Helpers::$wcs_get_subscriptions = null;
+
+		remove_filter( 'pre_http_request', $test_request, 10 );
```
daledupreez, line 146: As above, is it worth running the cleanup before we run any assertions?

daledupreez (commented):
Changes are working as described, but I am still super-concerned about this shipping without _any_ way to prevent nasty user-facing issues due to script timeouts.

tests/phpunit/WC_Stripe_Status_Test.php
wjrosa: I think so. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/4fbce0d48ba2d2dd3d28421752372712e9ccbf56

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: That's a good idea. Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/eb5e31d93f030f80a882bb5975483a61a1a191a8

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: I think the PHPCS ignoring comments was there because of a previous implementation. I removed it and I don't see any warnings. Is that your concern? If not, could you give me an example of what you expect?

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: Updated the formatting in my latest commits 👍

tests/phpunit/WC_Stripe_Status_Test.php
wjrosa: Added additional checks in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/5ba27af554b81ca2d9f09248cf064ebd6a6f6701

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -97,7 +97,11 @@ public static function get_detached_subscriptions( $limit = -1 ) {
 
 			$source_id = $subscription->get_meta( '_stripe_source_id' );
 			if ( $source_id ) {
-				$payment_method = WC_Stripe_API::get_payment_method( $source_id );
+				$payment_method = WC_Stripe_Database_Cache::get( 'payment_method_for_source_' . $source_id );
```
daledupreez, line 100: Do we have any APIs that allow for payment methods to be detached from within the plugin? If so, we should also make sure that database updates triggered from that code clear this cache.

includes/class-wc-stripe-status.php
```diff
@@ -194,4 +195,62 @@ public function render_status_report_section() {
...
+		return $tools;
+	}
+
+	/**
+	 * Lists Stripe subscriptions with detached payment methods.
+	 *
+	 * @return void
+	 */
+	public function list_detached_subscriptions() {
+		$subscriptions     = WC_Stripe_Subscriptions_Helper::get_detached_subscriptions();
```
daledupreez: As a safety measure, might it be worth passing in a number like `1000` here rather than defaulting to `-1`?
```suggestion
		$subscriptions     = WC_Stripe_Subscriptions_Helper::get_detached_subscriptions( 1000 );
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -103,4 +145,59 @@ public static function is_manual_renewal_enabled() {
...
+				'%s subscriptions are missing payment methods, <strong>preventing renewals</strong>. ',
+				count( $subscriptions ),
+				'woocommerce-gateway-stripe'
+			),
+			count( $subscriptions )
+		);
+		$intro_message .= __( 'To fix this, either:', 'woocommerce-gateway-stripe' ) . '<br />';
+		$intro_message .= __( '1) Share the payment method page link with the customer to update it, or', 'woocommerce-gateway-stripe' ) . '<br />';
+		$intro_message .= __( "2) Manually update the payment method in the subscription's billing details using a valid payment method from the customer's Stripe account. ", 'woocommerce-gateway-stripe' );
+		$intro_message .= __( 'Below are the affected subscriptions and their update links:', 'woocommerce-gateway-stripe' ) . '<br />';
```
daledupreez: This comment follows on from [this thread](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/files#r2164315732), and shows an example of making super-sure that our output is fully HTML-escaped, as most of this content was not being escaped, despite being intended for inclusion in a notice.

I double-checked the WP i18n docs and have now suggested `wp_kses()` rather than trying to interpolate the `<strong>` element later.

```suggestion
		$intro_message = sprintf(
			wp_kses
			/* translators: %s: subscriptions count */
				_n(
					'%s subscription is missing the payment method, <strong>preventing renewals</strong>. ',
					'%s subscriptions are missing payment methods, <strong>preventing renewals</strong>. ',
					count( $subscriptions ),
					'woocommerce-gateway-stripe'
				),
				[ 'strong' => array() ]
			),
			count( $subscriptions )
		);
		$intro_message .= esc_html__( 'To fix this, either:', 'woocommerce-gateway-stripe' ) . '<br />';
		$intro_message .= esc_html__( '1) Share the payment method page link with the customer to update it, or', 'woocommerce-gateway-stripe' ) . '<br />';
		$intro_message .= esc_html__( "2) Manually update the payment method in the subscription's billing details using a valid payment method from the customer's Stripe account. ", 'woocommerce-gateway-stripe' );
		$intro_message .= esc_html__( 'Below are the affected subscriptions and their update links:', 'woocommerce-gateway-stripe' ) . '<br />';
```

includes/class-wc-stripe-status.php
daledupreez: And maybe we can add a notice (or an option to show a notice?) when a limit is supplied and additional subscriptions exist beyond that number? My only concern on this front is that we might not have that iteration in place before we ship 9.6.0 next week.

daledupreez (approved):
This is working as described, but I would still like to have more guard rails in place before we ship to users, as this could easily break or behave poorly on sites with lots of subscriptions.

So I am wary of merging this before we ship 9.6.0, as I think we should improve the failure mode(s) before we roll this out.

daledupreez (changes requested):
I forgot about my request to HTML-escape things. 🤦 I am going to shift back to requesting some changes, though they should be pretty simple.

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: That's a good point. Added the cache deletion in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391/commits/4d3c986724b7f86ddb40e55453274b51e922f646

includes/class-wc-stripe-status.php
wjrosa: The problem I see with that is that we would need to query for all subscriptions to check how many might exist beyond 1000.

> My only concern on this front is that we might not have that iteration in place before we ship 9.6.0 next week.

It was not my intention to target 9.6.0. As I mentioned during the meetup, it was preferred that 9.6.0 focus on legacy 👍 . The idea was to ship this before or after it.

daledupreez (approved):
The changes look good to me and test well with the updates.

daledupreez:
Thanks for your patience in dealing with all my suggestions and complaints, @wjrosa!

PR #4393, opened 2025-06-04
Subscription detached notice when viewing subscription

wjrosa:
Fixes STRIPE-500

As part of STRIPE-483, this PR adds a new notice to the subscription edit page in wp-admin when the subscription payment method is detached. It also includes information about the new tool introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391

Notice preview:
Screenshot 2025-06-04 at 17 32 17

I am also taking the opportunity to introduce the new page helper class I created a POC for [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4314).

## Testing instructions

- Checkout this branch on your test environment (`add/single-subscription-detached-notice`)
- Create a subscription for any user (perform a purchase or add using the admin interface) *
- Comment out the lines handling the cache (`includes/compat/class-wc-stripe-subscriptions-helper.php:53` and `includes/compat/class-wc-stripe-subscriptions-helper.php:88`)
- Comment out the lines checking for the detachment, so all subscriptions can be returned (`includes/compat/class-wc-stripe-subscriptions-helper.php:74` and `includes/compat/class-wc-stripe-subscriptions-helper.php:77`, also comment out the closing brackets to avoid breaking the block) *
- Attempt to edit the subscription
- Confirm the notice is displayed

\* You can also purchase subscriptions as a shopper and detach their payment method using the Stripe dashboard instead.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -458,6 +463,47 @@ public function payment_methods_check_environment() {
...
+			$theorder = wcs_get_subscription( $GLOBALS['post']->ID );
+		}
+
+		if ( ! isset( $theorder ) || ! $theorder instanceof WC_Subscription ) {
+			return;
+		}
+
+		$subscription = $theorder;
+		if ( WC_Stripe_Subscriptions_Helper::is_subscription_payment_method_detached( $subscription ) ) {
+			$detached_message  = __( 'The payment method for this subscription has been detached, <strong>preventing renewals</strong>. Share the payment method page link with the customer to update it or manually set the Stripe Payment Method ID meta field in the subscriptions details\' "Billing" section to another from the customer\'s page on Stripe. Below are the last subscriptions affected and the links as mentioned earlier:<br />', 'woocommerce-gateway-stripe' );
```
Mayisha: Same suggestion as the [other PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4391#discussion_r2129111891). 
Additionally, on this subscription edit page, we are only talking about this specific subscription. `Below are the last subscriptions affected and the links as mentioned earlier:` This part talks about more than one subscriptions which is a bit off in this context.

includes/class-wc-stripe-page-helper.php
```diff
@@ -0,0 +1,37 @@
...
+ * Class to identify current page/screen.
+ */
+class WC_Stripe_Page_Helper {
+	/**
+	 * Checks if the current page is a subscription edit page in wp-admin.
+	 *
+	 * @return bool
+	 */
+	public static function is_subscription_edit_page() {
+		if ( apply_filters( 'wc_stripe_is_subscription_edit_page', false ) ) {
```
Mayisha: Curious about why we are using this `wc_stripe_is_subscription_edit_page` filter in 3 places and what is the thought behind introducing this filter.

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Good point. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/205da17f04c94aee2c4dd8eb5122a25090a90799 and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/941bda2a9748de412ebbcd490a094a34f3fde0ac

includes/class-wc-stripe-page-helper.php
wjrosa: It is basically used for unit tests. For some reason, I could not overwrite the check for the page using the query params. I tried multiple ways, but none worked. So I decided to use this filter instead.

I have improved the code in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/e4e48e1252bd4ec0fab54e941beeb9e32b8fc191

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -114,6 +110,26 @@ public static function get_detached_subscriptions( $limit = -1 ) {
...
+	 * @param $subscription WC_Subscription The subscription object to check.
+	 * @return bool True if the payment method is detached, false otherwise.
+	 */
+	public static function is_subscription_payment_method_detached( $subscription ) {
+		$source_id = $subscription->get_meta( '_stripe_source_id' );
+		if ( ! $source_id ) {
+			return false;
+		}
+
+		$payment_method = WC_Stripe_API::get_payment_method( $source_id );
```
Mayisha: We should wrap it in a try/catch block.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -464,6 +469,49 @@ public function payment_methods_check_environment() {
...
+			);
+			$detached_message .= '<br />' . sprintf(
+				/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag */
+				__( 'To list all your current subscriptions with payment methods detached, go to WooCommerce -> Status -> %1$sTools%2$s -> <strong>List Stripe subscriptions with detached payment method</strong>.', 'woocommerce-gateway-stripe' ),
+				'<a href="' . esc_url( admin_url( 'admin.php?page=wc-status&tab=tools' ) ) . '">',
+				'</a>'
+			);
+			$this->add_admin_notice( 'subscription_detached', 'notice notice-error', $detached_message );
+		}
+	}
```
Mayisha, line 513: @wjrosa as we are on the subscription edit page and there is only 2 links to show, what do you think about showing the links (action item) beside the points itself?

**Current**
Screenshot 2025-06-21 at 12 08 28 AM


**Proposed**
Screenshot 2025-06-21 at 12 03 34 AM

includes/class-wc-stripe-page-helper.php
Mayisha: Sharing what I found as I was also looking for alternate solutions, I found one tweak that worked for me.

I had to change the `is_custom_orders_table_enabled` as following-

```
public static function is_custom_orders_table_enabled() {
		return get_option( wc_get_container()->get( CustomOrdersTableController::class )::CUSTOM_ORDERS_TABLE_USAGE_ENABLED_OPTION ) === 'yes';
	}
```

I tried the following tests in `WC_Stripe_Page_Helper_Test`

```
public function test_is_subscription_edit_page_with_custom_orders_table() {
		update_option( 'woocommerce_custom_orders_table_enabled', 'yes' );

		// Test valid subscription edit page
		$_GET['page'] = 'wc-orders--shop_subscription';
		$_GET['id'] = '123';
		$this->assertTrue( WC_Stripe_Page_Helper::is_subscription_edit_page() );

		// Test invalid page
		$_GET['page'] = 'wc-orders--shop_order';
		$this->assertFalse( WC_Stripe_Page_Helper::is_subscription_edit_page() );

		// Test missing ID
		$_GET['page'] = 'wc-orders--shop_subscription';
		unset( $_GET['id'] );
		$this->assertFalse( WC_Stripe_Page_Helper::is_subscription_edit_page() );

		remove_filter( 'wc_stripe_woo_compat_utils_is_custom_orders_table_enabled', '__return_true' );
		unset( $_GET['page'], $_GET['id'] );
	}

public function test_is_subscription_edit_page_with_custom_order_tables_disabled() {
		$subscription = wp_insert_post( [
			'post_type' => 'shop_subscription',
			'post_title' => 'Test Subscription',
			'post_status' => 'publish'
		] );
		$_GET['post'] = $subscription;
		$_GET['action'] = 'edit';
		$this->assertTrue( WC_Stripe_Page_Helper::is_subscription_edit_page() );

		// Test wrong action
		$_GET['action'] = 'view';
		$this->assertFalse( WC_Stripe_Page_Helper::is_subscription_edit_page() );

		// Test wrong post type
		$regular_post = wp_insert_post( [ 
			'post_title' => 'Test Post',
			'post_status' => 'publish'
		] );
		$_GET['post'] = $regular_post;
		$_GET['action'] = 'edit';
		$this->assertFalse( WC_Stripe_Page_Helper::is_subscription_edit_page() );

		unset( $_GET['post'], $_GET['action'] );
	}
```

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: `get_payment_method` calls `retrieve` and none of them throws exceptions, but they can return a `WP_Error` object. I updated the method to check for that in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/6c1822a0eaac71956d8cb177b0b4c0887e623825

wjrosa: Added specific unit tests in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/609571c29a6ccc63cd93385ffd71bc66daa60093

includes/class-wc-stripe-page-helper.php
wjrosa: Thanks for sharing what you found! I have removed the filter and created unit tests based on your draft above in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/ad034fd1114e09dca9cd3f8f23b1381309516601

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Good idea! Thank you. I have implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/b963054a5aeac9759abdd241425b6e341463cac9. Here's a preview:

Screenshot 2025-06-25 at 09 47 13

wjrosa:
Since I started to move some of our functions checking for the current page to WooCommerce core, I removed the `WC_Stripe_Page_Helper` class in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/2e6eb2a7a2ac0fa4fa47563f0ae59cce7cbaa382

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -114,6 +110,39 @@ public static function get_detached_subscriptions( $limit = -1 ) {
...
+		if ( is_wp_error( $payment_method ) ) {
```
Mayisha: We should check here if the response has any error object as well (`isset( $payment_method->error )`

I coincidentally came across an old subscription on my local site while testing this PR. I see the subscription detached notice on the subscription edit page but there is no errors in my logs. When I logged the `$payment_method` variable to check, I found the following value-

`
{"error":{"code":"resource_missing","doc_url":"https:\/\/stripe.com\/docs\/error-codes\/resource-missing","message":"No such PaymentMethod: 'pm_id'","param":"payment_method","request_log_url":"https:\/\/dashboard.stripe.com\/acct_id\/test\/logs\/req_IUgF67UoEFgkAA?t=1751034251","type":"invalid_request_error"}}
`

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -114,6 +110,39 @@ public static function get_detached_subscriptions( $limit = -1 ) {
...
+			// If we can't retrieve the payment method, assume it's detached.
+			WC_Stripe_Logger::error(
+				sprintf(
+					/* translators: %1$s is the subscription ID, %2$s is the error message */
+					__( 'Error retrieving payment method for subscription %1$s: %2$s', 'woocommerce-gateway-stripe' ),
+					$subscription->get_id(),
+					$payment_method->get_error_message()
```
Mayisha: Noting for caution, when the error is not a wp_error object but the error I mentioned in the above comment (`$payment_method->error`), this line will throw a fatal error for undefined `get_error_message` function.

Mayisha (changes requested):
Thanks @wjrosa for the changes. The notice looks perfect. Just one last bit to fix that I commented below.

To reproduce the error, I think these steps should work.
- Purchase a subscription as a shopper.
- Renew the subscription once as an admin.
- Connect to a different (non-US) Stripe account.
- View the subscription edit page.

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: Thanks! Added the check in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/f0b77614152b648bcb718a54fad8b168b9c1075f (included a fallback)

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: It should be safe now with the extra check in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/f0b77614152b648bcb718a54fad8b168b9c1075f

Mayisha (approved):
Thanks for addressing the last piece. LGTM 🎉

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -558,6 +618,24 @@ public function stripe_updated() {
...
+
+	/**
+	 * Checks if the current page is a subscription edit page in wp-admin.
+	 *
+	 * This should be removed once WooCommerce provides a way to check for subscription edit pages.
+	 *
+	 * @return bool
+	 */
+	private static function is_subscription_edit_page() {
+		$query_params = wp_unslash( $_GET ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
```
daledupreez: Might it be worth checking `$_REQUEST` to handle `POST` requests to the page as well?
```suggestion
		$query_params = wp_unslash( $_REQUEST ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -8,18 +8,18 @@
  */
 class WC_Stripe_Subscriptions_Helper {
 	/**
-	 * Transient key for detached subscriptions.
+	 * Stripe customer page base URL.
 	 *
 	 * @var string
 	 */
-	private const DETACHED_SUBSCRIPTIONS_CACHE_PREFIX = 'detached_subscriptions';
+	public const STRIPE_CUSTOMER_PAGE_BASE_URL = 'https://dashboard.stripe.com/customers/';
```
daledupreez, line 15: Non-blocking thought: should we send users to the test dashboard when test mode is enabled?

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.7.0 - xxxx-xx-xx =
+* Add - Introduces a new notice that might be displayed when editing a subscription that has no payment method attached
```
daledupreez: Maybe we can simplify this a bit?
```suggestion
* Add - Show a notice when editing a subscription that has no payment method attached
```

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -118,6 +110,45 @@ public static function get_detached_subscriptions( $limit = -1 ) {
...
 	}
 
+	/**
+	 * Checks if a subscription's payment method is detached from the customer.
+	 *
+	 * @param $subscription WC_Subscription The subscription object to check.
+	 * @return bool True if the payment method is detached, false otherwise.
+	 */
+	public static function is_subscription_payment_method_detached( $subscription ) {
+		$source_id = $subscription->get_meta( '_stripe_source_id' );
```
daledupreez, line 124: Should we validate that we have a valid subscription? We're implicitly checking for an order object by calling `get_meta()`, but that would not prevent normal orders from flowing through to the logic below.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -444,6 +449,61 @@ public function payment_methods_check_environment() {
...
+		if ( empty( $theorder ) && ! empty( $GLOBALS['post']->ID ) ) {
+			$theorder = wcs_get_subscription( $GLOBALS['post']->ID );
+		}
+
+		if ( ! isset( $theorder ) || ! $theorder instanceof WC_Subscription ) {
+			return;
+		}
+
+		$subscription = $theorder;
+		if ( WC_Stripe_Subscriptions_Helper::is_subscription_payment_method_detached( $subscription ) ) {
```
daledupreez, line 481: Should we not check the subscription status before running these checks? I don't think we should run the checks if the subscription was cancelled or put on hold. I am less sure about pending subscriptions, but that may be similar.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -444,6 +449,61 @@ public function payment_methods_check_environment() {
...
+			$detached_message .= __( 'To fix this, either: <br />', 'woocommerce-gateway-stripe' );
+			$detached_message .= __( '1) Share the payment method page link with the customer to update it: ', 'woocommerce-gateway-stripe' ) . $customer_payment_method_link . '<br />';
+			$detached_message .= __( ' or <br />', 'woocommerce-gateway-stripe' );
+			$detached_message .= __( "2) Manually update the payment method in the subscription's billing details using a valid payment method from the customer's Stripe account: ", 'woocommerce-gateway-stripe' ) . $customer_stripe_page . '<br />';
+			$detached_message .= '<br />' . sprintf(
+				/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag */
+				__( 'To list all your current subscriptions with payment methods detached, go to WooCommerce -> Status -> %1$sTools%2$s -> <strong>List Stripe subscriptions with detached payment method</strong>.', 'woocommerce-gateway-stripe' ),
+				'<a href="' . esc_url( admin_url( 'admin.php?page=wc-status&tab=tools' ) ) . '">',
+				'</a>'
+			);
```
daledupreez: I think we should use the same string for the tool name so we don't end up with different translations. Ideally, we should add a helper method or similar to get the name so we're using exactly the same code to get the string in all cases, but that may be worth tackling in a follow-up.
```suggestion
			$detached_message .= '<br />' . sprintf(
				/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag 3) The already-translated title of the tool*/
				__( 'To list all your current subscriptions with payment methods detached, go to WooCommerce -> Status -> %1$sTools%2$s -> <strong>%3$s</strong>.', 'woocommerce-gateway-stripe' ),
				'<a href="' . esc_url( admin_url( 'admin.php?page=wc-status&tab=tools' ) ) . '">',
				'</a>',
				__( 'List Stripe subscriptions with detached payment method', 'woocommerce-gateway-stripe' ),
			);
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -444,6 +449,61 @@ public function payment_methods_check_environment() {
...
+		// If $theorder is empty (i.e. non-HPOS), fallback to using the global post object.
+		if ( empty( $theorder ) && ! empty( $GLOBALS['post']->ID ) ) {
+			$theorder = wcs_get_subscription( $GLOBALS['post']->ID );
+		}
+
+		if ( ! isset( $theorder ) || ! $theorder instanceof WC_Subscription ) {
+			return;
+		}
+
+		$subscription = $theorder;
```
daledupreez: Populating `$theorder` using local code feels risky, especially if other code/plugins have checks driven by `$theorder` being defined. I think we should stick to a local variable.
```suggestion
		global $theorder;

		$subscription = null;

		if ( isset( $theorder ) ) {
			$subscription = $theorder;
		}
		// If $theorder is empty (i.e. non-HPOS), fallback to using the global post object.
		elseif ( empty( $theorder ) && ! empty( $GLOBALS['post']->ID ) ) {
			$subscription = wcs_get_subscription( $GLOBALS['post']->ID );
		}

		if ( ! isset( $subscription ) || ! $subscription instanceof WC_Subscription ) {
			return;
		}

```

tests/phpunit/Admin/Migrations/WC_Stripe_Subscriptions_Repairer_Legacy_SEPA_Tokens_Test.php
```diff
@@ -431,7 +431,7 @@ private function create_orders_and_subscriptions() {
 	 * @return WC_Subscription The created subscription.
 	*/
 	private function create_subscription( $payment_method, $customer_id, $source_id ) {
-		$subscription = new \WC_Subscription();
```
daledupreez, line 434: Nit: why are we including this unrelated change in this PR?

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -531,4 +531,57 @@ public function options_to_notices_map() {
...
+
+		$this->assertCount( 1, $notices->notices );
+		$this->assertArrayHasKey( 'subscription_detached', $notices->notices );
+		$this->assertMatchesRegularExpression( '/The payment method for this subscription has been detached/', $notices->notices['subscription_detached']['message'] );
+
+		remove_filter( 'pre_http_request', $test_request, 10, 3 );
+
+		// Clean up.
+		unset( $_GET );
+		update_option( 'woocommerce_custom_orders_table_enabled', 'no' );
```
daledupreez: We should also restore `$theorder`. And should we run the cleanup before we run any assertions?

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -531,4 +531,57 @@ public function options_to_notices_map() {
...
+		};
+
+		add_filter( 'pre_http_request', $test_request, 10, 3 );
+
+		$notices = new WC_Stripe_Admin_Notices();
+		$notices->subscription_check_detachment();
+
+		$this->assertCount( 1, $notices->notices );
+		$this->assertArrayHasKey( 'subscription_detached', $notices->notices );
+		$this->assertMatchesRegularExpression( '/The payment method for this subscription has been detached/', $notices->notices['subscription_detached']['message'] );
```
daledupreez: Do we really need a regular expression for this check? Would `assertStringContainsString()` or `assertEquals()` not be sufficient?

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -531,4 +531,57 @@ public function options_to_notices_map() {
...
+		update_option( 'woocommerce_custom_orders_table_enabled', 'yes' );
+
+		$subscription = new WC_Subscription();
+		$subscription->set_id( 123 );
+		$subscription->save();
+
+		$subscription->update_meta_data( '_stripe_source_id', 'src_123' );
+		$subscription->save_meta_data();
+
+		$theorder = $subscription;
```
daledupreez: It feels like we should also test the other case where HPOS is not enabled and the `$post` global is defined.

tests/phpunit/Compat/WC_Stripe_Subscriptions_Helper_Test.php
```diff
@@ -174,4 +174,124 @@ public function provide_test_build_subscriptions_detached_messages() {
 			],
 		];
 	}
+
+	/**
+	 * Tests for `is_subscription_payment_method_detached`.
```
daledupreez: To be more specific, might it be worth using `@see` for this reference?
```suggestion
	 * Tests for {@see WC_Stripe_Subscriptions_Helper::is_subscription_payment_method_detached()}.
```

daledupreez (changes requested):
This is working as expected, but I noticed a number of issues that I think we need to address.

tests/phpunit/Compat/WC_Stripe_Subscriptions_Helper_Test.php
wjrosa: I followed the existing standards. But that works too 👍

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
wjrosa: Makes sense! Did both changes in:
https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/ff234091c7d6a5c3a207bb8034859caf57094c91
https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/deb7e09b9ff62d5fb2ffe3f741da26b06fb7826a

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
wjrosa: You're right. That's much simpler. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/1cedb6cf8eb4b8c744a9350c9ba4dce49106fe8c

tests/phpunit/Admin/Migrations/WC_Stripe_Subscriptions_Repairer_Legacy_SEPA_Tokens_Test.php
wjrosa: I just saw it while working on the main changes here, and that is not worth a standalone PR (too small and hard to justify).

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
wjrosa: It makes sense, Dale! Added it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/ed9b6f1b57308d58151007799b0d970a5e208e10

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: That's a good point. I have included a check for the `active` status only (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/68845b82f1b60c33c62688d9d690fb31c555b2f7) to match what we did for the debug tool. There, we decided to keep only active because the other statuses would not renew anyway.

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: Ideally, this should be done by a strict type check. But since we don't usually do that, I added the extra check in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/4c2d5ebce27fb89215c7924684df6b16d6885e23

includes/compat/class-wc-stripe-subscriptions-helper.php
wjrosa: Stripe automatically redirects the user based on the current test mode setting 👍 (validated it now, dashboard is the same no matter the URL)

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Good idea. Suggestion applied 👍

changelog.txt
```diff
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 9.7.0 - xxxx-xx-xx =
-* Add - Introduces a new notice that might be displayed when editing a subscription that has no payment method attached
+* Add - Show a notice when editing a subscription that has no payment method attached
```
daledupreez: Non-blocking: we may want to specify that we only look at active subscriptions.
```suggestion
* Add - Show a notice when editing an active subscription that has no payment method attached
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -450,7 +450,7 @@ public function payment_methods_check_environment() {
 	}
 
 	/**
-	 * Adds a notice to the order details page if the subscription payment method has been detached.
+	 * Adds a notice to the order details page if the (active) subscription payment method has been detached.
```
daledupreez: I think it would be clearer to express this as follows, because I find it hard to understand what the `(active)` applies to.
```suggestion
	 * Adds a notice to the subscription details page if we are looking at an active subscription and the payment method has been detached.
```

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -574,14 +589,68 @@ public function test_subscription_check_detachment() {
...
-		unset( $_GET );
+		unset( $_REQUEST );
+		if ( count( $post_globals ) > 0 ) {
+			foreach ( $post_globals as $key => $value ) {
+				unset( $GLOBALS[ $key ] );
+			}
+		}
+
+		$theorder = $original_order;
 		update_option( 'woocommerce_custom_orders_table_enabled', 'no' );
```
daledupreez: Should we set this to `no` or reset it to what it was when the tests started?

tests/phpunit/Helpers/WC_Subscriptions.php
```diff
@@ -25,11 +28,15 @@ class WC_Subscriptions {
...
 	public static $wcs_get_subscription = null;
 
-	public static function set_wcs_get_subscription( $function ) {
-		self::$wcs_get_subscription = $function;
+	/**
+	 * @param callable $calback Function to call when wcs_get_subscription is called.
+	 * @return void
+	 */
+	public static function set_wcs_get_subscription( $calback ) {
+		self::$wcs_get_subscription = $calback;
```
daledupreez: Nit: there is a typo in `$calback`.
```suggestion
	 * @param callable $callback Function to call when wcs_get_subscription is called.
	 * @return void
	 */
	public static function set_wcs_get_subscription( $callback ) {
		self::$wcs_get_subscription = $callback;
```

includes/compat/class-wc-stripe-subscriptions-helper.php
daledupreez: I am not sure that we can include `WC_Subscription` as the argument type anyway, as the subscriptions plugin may not be installed. So this looks good to me!

tests/phpunit/Admin/Migrations/WC_Stripe_Subscriptions_Repairer_Legacy_SEPA_Tokens_Test.php
daledupreez: Thanks for confirming.

daledupreez (approved):
Thanks for tackling all those small improvements!

I've re-tested and things are still looking good. I had some minor feedback on comments, but none of them are blocking.

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
wjrosa: When I added this line, I double-checked the value of the setting before running the test, and it was set to `no`. But I have now improved it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4393/commits/30138d4ef3485b057b09fd79d48d27406adc8f3a by storing the original value before we start testing the class

PR #4395, opened 2025-06-05
Fix: express checkout displays wrong amount for JPY

Mayisha:
Fixes STRIPE-494

When the store currency is set to JPY, the express checkout modals/pages display 100x the amount. This only happens on the shortcode cart and shortcode checkout pages.

We are using the `currency_decimals` field to calculate price in [the frontend](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/client/express-checkout/transformers/wc-to-stripe.js#L19) but its value is never set on the backend. So by default `2` is used for every currency during the calculation.
This PR sets the `currency_decimals` in params to be used in the frontend.

## Testing instructions
- Set store currency to JPY and set the `Number of decimals` to `0`.

Screenshot 2025-06-05 at 5 44 01 PM

- As a shopper, add a product to your cart and go to the shortcode checkout page.
- In `develop`, click the `Link` button and notice that the amount displayed is 100x more than the actual total price.

Screenshot 2025-06-06 at 12 08 26 AM

- In this branch, click the `Link` button and confirm that the correct amount.

Screenshot 2025-06-06 at 12 08 36 AM

- Check Google Pay and Apple Pay modals and confirm that they are showing the correct amount as well.
- Purchase the product with express checkout and confirm that correct amount has been charged in Stripe.
- Change the store currency to USD/Euro and `Number of decimals` to `2`. Confirm that the amount displayed in express checkout is correct for this currency.


annemirasol (approved):
Code LGTM!

- Was able to reproduce the issue\*
- Verified this PR fixes it

| Before | After |
|--------|-------|
| Screenshot 2025-06-09 at 10 24 02 AM | Screenshot 2025-06-09 at 10 25 47 AM |



\* I can only reproduce the issue when my Link is logged out. After logging in, it fixes the price. When logged in, the price is always correct.

In `develop`:
| Logged out | After logging in |
|------------|------------------|
| Screenshot 2025-06-09 at 10 24 02 AM | Screenshot 2025-06-09 at 10 24 23 AM |

PR #4379, opened 2025-05-30
Fix legacy deprecation notice displayed on new plugin installs

Mayisha:
Reported in p1748557211532439-slack-C055WHLA98D 

## Changes proposed in this Pull Request:
- Bailing early when Stripe is not enabled from `display_legacy_deprecation_notice` method.

## Testing instructions
- Create a new JN site.
- Install WooCommerce, and install the latest Stripe plugin. Activate the plugins.
- Notice that, on the plugin settings page, the legacy deprecation notice is displayed, although you haven't enabled the Stripe payment method yet.

Screenshot 2025-05-30 at 2 21 19 PM

- Now deactivate and delete the plugin from the JN site.
- Build this branch and upload the zip to your JN site. Install and activate the Stripe plugin from the zip.
- Confirm that, on the plugin settings page, the legacy deprecation notice is not displayed.
- Enable Stripe and connect to your test account.
- Confirm that, on the plugin settings page, the legacy deprecation notice is not displayed.
- Enable legacy experience in your Stripe settings page. Confirm that, on the plugin settings page, now the legacy deprecation notice is displayed.

PR #4396, opened 2025-06-05
Fix plugin list page links

Mayisha:
@malithsen discovered this while reviewing https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4379

In a recent refactor these code have been moved from the `woocommerce-gateway-stripe.php` file (the base file) to this new file. So in this new class, we need to use the `WC_STRIPE_MAIN_FILE` constant that defines the plugin base file name instead of `_FILE_` which returns the current file name

Fixed the file name to retrieve the correct base file of the plugin. 

## Testing instructions
- Go to the `Plugins` page in your WordPress dashboard.
- In `develop` branch, notice that the `Settings`, `Docs` and `Support` links are missing for Stripe plugin.

Screenshot 2025-06-05 at 5 13 46 PM

- In this branch, notice that the `Settings`, `Docs` and `Support` links are present.

Screenshot 2025-06-05 at 5 07 36 PM

**Note**
Skipping a new changelog entry as the refactor was done in the same version as well.


malithsen (approved):
Looks great! Thanks for fixing this.

PR #4397, opened 2025-06-05
Unit tests for the banners refactoring

wjrosa:
Base PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4328

On the base PR I did a considerable refactoring of the promotional banners. I am adding unit tests for all the new JS files here to make the code more reliable.

## Testing instructions

Code review. Check if the tests make sense and are still passing.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/settings/payment-settings/promotional-banner/__tests__/re-connect-account-banner.test.js
```diff
@@ -0,0 +1,84 @@
...
+		);
+
+		expect( window.location.assign ).toHaveBeenCalledWith( oauthUrl );
+
+		// Set the original function back to keep further tests working as expected.
+		Object.defineProperty( window, 'location', {
+			value: { assign },
+		} );
+	} );
+	it( 'should create error notice when oauth URLs are invalid', () => {
```
daledupreez: Can we test that we throw an error when the expected live and test URLs are null separately?

client/settings/payment-settings/promotional-banner/new-checkout-experience-apms-banner.js
```diff
@@ -87,7 +87,7 @@ export const NewCheckoutExperienceAPMsBanner = ( {
 							'woocommerce-gateway-stripe'
 						) }
 					</h4>
-					<p>
+					<p data-testid="intro-new-checkout-apms">
```
daledupreez: Do we really need these test IDs for the tests? Can we not check for the text strings we're adding to the content?

daledupreez (commented):
Things are looking good enough to me, with the exception of using the test IDs.

That said, might it be worth aiming for more integration-y tests that validate we see the expected banners in the main UI, and not just rely on lower-level unit tests? (That work could be tackled in a follow-up PR.)

client/settings/payment-settings/promotional-banner/new-checkout-experience-apms-banner.js
wjrosa: TestIds are faster, but I imagine you don't like them because they require testing impacting written code, which should not be the case (code should be agnostic of tests). Is that your thought? Just to understand it. Updated it all with my latest commits.

client/settings/payment-settings/promotional-banner/__tests__/re-connect-account-banner.test.js
wjrosa: I'm sorry, but I'm not sure if I understand. Do you mean move it to a separate `describe` clause?

Mayisha (approved):
Individual banner tests look good to me. 👍 
I find the layout test `client/settings/payment-settings/promotional-banner/__tests__/banner-layout.test.js` not valuable. We are directly adding the elements and then checking if they are present, which does not add much value and asserts no behavioral changes. You can consider removing it.

client/settings/payment-settings/promotional-banner/__tests__/re-connect-account-banner.test.js
daledupreez: Sorry about the delayed response.

My thinking was that we are currently testing the following, where `testOauthUrl` and `oauthUrl` are both null:
```js
		const { getByText } = render(
			<ReConnectAccountBanner testOauthUrl={ null } oauthUrl={ null } />
		);
```

I think we should also have the following test cases:
 * `testOauthUrl` is null and `oauthUrl` is valid
 * `testOauthUrl` is valid and `oauthUrl` is null

This is not blocking.

client/settings/payment-settings/promotional-banner/new-checkout-experience-apms-banner.js
daledupreez: This is more of a philosophical question, largely because Testing Library explicitly advocates for tests that mirror how users interact with the UI. Test IDs are more of an implementation detail that are faster to write and a bit easier to maintain, but are a little bit abstracted from the actual user experience. I do think it's worth it to use the text explicitly, which does have slightly more maintenance burden when we update the text, as it means we're closer to testing the UI.

daledupreez (approved):
All good from my side -- none of my comments are blocking.

wjrosa:
@daledupreez thanks for the review!

> I think we should also have the following test cases:

Even though that's not a blocker, I added them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4397/commits/e432439b1ece374d68e2c9145372b824cb406e40

daledupreez:
@wjrosa, it looks like you haven't added a changelog entry and haven't explicitly noted that you won't add one, so the changelog checks are going to fail. Not sure which path you want for this PR.

PR #4340, opened 2025-05-21
Hide future payments message from payment element when manual renewal is required

Mayisha:
Fixes STRIPE-460
Fixes #4332 

## Changes proposed in this Pull Request:
- Checking if automatic renewal is disabled for the site. 
- When automatic renewal is disabled and manual renewal is required,
    - Prevent showing the terms stating future usage conditions.
    - Prevent saving the payment method.

## Testing instructions
- Install the WooCommerce Subscriptions plugin and create a subscription product.
- Go to `WooCommerce > Settings > Subscriptions` and disable auto renewal.

Screenshot 2025-05-21 at 12 07 13 PM

- As a shopper, add the subscription product to your cart and go to the checkout page. (Test on both shortcode and block checkout)
- Confirm that the Stripe terms about saving the payment method for future usage are not displayed.

Shortcode Checkout | Block Checkout
--- | ---
Screenshot 2025-05-23 at 3 36 33 PM | Screenshot 2025-05-23 at 3 36 11 PM

- Go to `My Account > Payment methods` page and confirm that the card has not been saved.
- Go to `My Account > Subscriptions` and view the subscription. Confirm that the subscription is set to manual renewal.
- Regression test
    - Enable automatic payment.
    - As a shopper, purchase a subscription.
    - Confirm that on the checkout page, the Stripe terms about saving the payment method for future usage are displayed.

    Screenshot 2025-05-23 at 3 24 45 PM

    - Confirm that the card is saved in `My Account > Payment methods` page.

PR #4401, opened 2025-06-06
Allow non-reusable methods to be used with manual renewals

### LINKED ISSUE #4217: OXXO no longer appears as a payment method at checkout when a subscription product is in the cart
**Describe the bug**
OXXO no longer appears as a payment method at checkout when a subscription product is in the cart. This worked previously (manual renewals were processed successfully), but the option is now missing after introducing the new checkout experience.

**To Reproduce**
1. Enable OXXO in Stripe and WooCommerce Stripe Gateway settings.
2. Ensure store currency is set to MXN and store location is Mexico.
3. Create a subscription product.
4. Add subscription product to cart and proceed to checkout.
5. Observe that OXXO is not displayed as an available payment method.

**Expected behavior**
OXXO should appear as an available payment method at checkout for subscription purchases.

**Additional context**
	•	The issue only occurs with subscription products in the cart.
	•	OXXO works fine for simple products.
	•	This is likely a regression caused by changes in the new checkout experience.

Reported in collaboration with @ixistudio — thanks to Alberto Torres for the detailed testing, documentation, and screenshots that made it possible to identify this issue clearly.

wjrosa:
Fixes STRIPE-403
Fixes #4217
Woo Doc https://woocommerce.com/document/subscriptions/store-manager-guide/#accept-manual-renewals

Currently, voucher payment methods (Multibanco, Boleto, and Oxxo) cannot be used to pay for subscriptions. That's because they don't allow auto-renewal. According to the Woo docs, those should be accepted when manual renewals are enabled or required.

This PR fixes this behavior by checking for both settings when displaying and processing orders.

## Testing instructions

- Checkout and build this branch on your test environment (`add/allow-non-reusable-methods-to-be-used-with-manual-renewal`)
- Connect your Stripe account
- Add a subscription product
- Enable manual renewals in WooCommerce -> Settings -> Subscriptions -> "Accept Manual Renewals"
- Make sure the legacy checkout and the optimized checkout are disabled
- Configure your store to follow the specific rules for [Oxxo](https://docs.stripe.com/payments/oxxo), [Multibanco](https://docs.stripe.com/payments/multibanco), and [Boleto](https://docs.stripe.com/payments/boleto)
- As a shopper, add the subscription product to your cart
- Confirm the voucher payment methods are now available in both the block checkout and shortcode checkout

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
-* Add - Voucher payment methods (Boleto, Multibanco, and Oxxo) can now be used when purchasing subscriptions if manual renewals are enabled or required.
+* Add - Voucher payment methods (Boleto, Multibanco, and Oxxo) can now be used when purchasing subscriptions if manual renewals are enabled or required
 * Add - Adds a new promotional banner to promote the BNPL payment methods (Klarna, Afterpay, and Affirm) on the settings page.
 * Fix - Adds an exception to be thrown when the order item quantity is zero, during the retrieval of level 3 data from an order.
 * Dev - Deprecates the WC_Stripe_Order class and removes its inclusion call.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
+* Add - Voucher payment methods (Boleto, Multibanco, and Oxxo) can now be used when purchasing subscriptions if manual renewals are enabled or required.
 * Add - Adds a new promotional banner to promote the BNPL payment methods (Klarna, Afterpay, and Affirm) on the settings page.
 * Fix - Adds an exception to be thrown when the order item quantity is zero, during the retrieval of level 3 data from an order.
 * Dev - Deprecates the WC_Stripe_Order class and removes its inclusion call.
--- a/client/blocks/__tests__/utils.test.js
+++ b/client/blocks/__tests__/utils.test.js
@@ -57,7 +57,7 @@ describe( 'Blocks Utils', () => {
 			};
 		} );
 
-		test( 'has auto renewal subscription', () => {
+		test( 'cart has auto renewal subscription', () => {
 			mockGetSetting.mockReturnValue( {
 				cartContainsSubscription: true,
 				subscriptionManualRenewalEnabled: false,
@@ -65,11 +65,11 @@ describe( 'Blocks Utils', () => {
 			expect( shouldSetupOffSessionPayment( false, false ) ).toBeTruthy();
 		} );
 
-		test( 'shows save option', () => {
+		test( 'showSaveOption is true', () => {
 			expect( shouldSetupOffSessionPayment( true, true ) ).toBeTruthy();
 		} );
 
-		test( 'either', () => {
+		test( 'cart does not have auto renewal subscription and showSaveOption is false', () => {
 			expect( shouldSetupOffSessionPayment( false, false ) ).toBeFalsy();
 		} );
 	} );
--- a/client/blocks/__tests__/utils.test.js
+++ b/client/blocks/__tests__/utils.test.js
@@ -2,6 +2,7 @@ import { render } from '@testing-library/react';
 import {
 	extractOrderAttributionData,
 	populateOrderAttributionInputs,
+	shouldSetupOffSessionPayment,
 } from 'wcstripe/blocks/utils';
 
 describe( 'Blocks Utils', () => {
@@ -41,6 +42,35 @@ describe( 'Blocks Utils', () => {
 			expect(
 				global.wc_order_attribution.setOrderTracking
 			).toHaveBeenCalledWith( true );
+		} );
+	} );
+
+	describe( 'shouldSetupOffSessionPayment', () => {
+		let mockGetSetting;
+
+		beforeEach( () => {
+			mockGetSetting = jest.fn().mockReturnValue( {} );
+			global.wc = {
+				wcSettings: {
+					getSetting: mockGetSetting,
+				},
+			};
+		} );
+
+		test( 'has auto renewal subscription', () => {
+			mockGetSetting.mockReturnValue( {
+				cartContainsSubscription: true,
+				subscriptionManualRenewalEnabled: false,
+			} );
+			expect( shouldSetupOffSessionPayment( false, false ) ).toBeTruthy();
+		} );
+
+		test( 'shows save option', () => {
+			expect( shouldSetupOffSessionPayment( true, true ) ).toBeTruthy();
+		} );
+
+		test( 'either', () => {
+			expect( shouldSetupOffSessionPayment( false, false ) ).toBeFalsy();
 		} );
 	} );
 } );
--- a/client/blocks/utils.js
+++ b/client/blocks/utils.js
@@ -20,14 +20,20 @@ export const getBlocksConfiguration = () => {
  * Determines if off-session payment should be set up.
  *
  * @param {boolean} showSaveOption - Whether to show the save option.
+ * @param {boolean} paymentMethodReusable - Whether the payment method is reusable.
  * @return {boolean} True if off-session payment should be set up, false otherwise.
  */
-export const shouldSetupOffSessionPayment = ( showSaveOption ) => {
+export const shouldSetupOffSessionPayment = (
+	showSaveOption,
+	paymentMethodReusable
+) => {
 	const config = getBlocksConfiguration();
+	const methodNotReusableButManualRenewalEnabled =
+		! paymentMethodReusable && config?.subscriptionManualRenewalEnabled;
 	const hasAutoRenewingSubscription =
 		config?.cartContainsSubscription &&
 		! config?.subscriptionRequiresManualRenewal &&
-		! config?.subscriptionManualRenewalEnabled;
+		! methodNotReusableButManualRenewalEnabled;
 	return hasAutoRenewingSubscription || showSaveOption;
 };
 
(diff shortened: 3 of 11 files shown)


client/blocks/__tests__/utils.test.js
```diff
@@ -43,4 +44,33 @@ describe( 'Blocks Utils', () => {
...
+
+		test( 'has auto renewal subscription', () => {
+			mockGetSetting.mockReturnValue( {
+				cartContainsSubscription: true,
+				subscriptionManualRenewalEnabled: false,
+			} );
+			expect( shouldSetupOffSessionPayment( false, false ) ).toBeTruthy();
+		} );
+
+		test( 'shows save option', () => {
```
Mayisha: I think 'shouldSetupOffSessionPayment' only determines what value should be sent in the `setupFututreUsage` field. It is not related the display of the save checkbox.

client/blocks/utils.js
```diff
@@ -20,13 +20,20 @@ export const getBlocksConfiguration = () => {
...
+ * @param {boolean} paymentMethodReusable - Whether the payment method is reusable.
  * @return {boolean} True if off-session payment should be set up, false otherwise.
  */
-export const shouldSetupOffSessionPayment = ( showSaveOption ) => {
+export const shouldSetupOffSessionPayment = (
+	showSaveOption,
+	paymentMethodReusable
+) => {
 	const config = getBlocksConfiguration();
+	const manualRenewalRequired =
```
Mayisha: Let's add a comment explaining the logic.

includes/compat/trait-wc-stripe-subscriptions-utilities.php
```diff
@@ -121,7 +121,7 @@ public function is_subscription_item_in_cart() {
 	 */
 	public function is_manual_renewal_required() {
 		if ( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() ) {
-			return ( class_exists( 'WCS_Manual_Renewal_Manager' ) && WCS_Manual_Renewal_Manager::is_manual_renewal_required() ) || $this->cart_contains_renewal();
+			return WC_Stripe_Subscriptions_Helper::is_manual_renewal_required() || $this->cart_contains_renewal();
```
Mayisha: Now that we have the `is_manual_renewal_required` function in the helper class, we should remove this function and use the helper method [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/add/allow-non-reusable-methods-to-be-used-with-manual-renewal/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L521).

client/blocks/__tests__/utils.test.js
wjrosa: The test name here refers to the `showSaveOption` being passed as `true`. Not the expected result. I will think of a different name 🤔

wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4401/commits/ad8474d6b9092409c9982fe5e890b1aa07d73add

client/blocks/utils.js
wjrosa: Makes sense. Improved it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4401/commits/c7a8c705ffcf8cc4a2264a1e9ac4a2976f198b67

includes/compat/trait-wc-stripe-subscriptions-utilities.php
wjrosa: I left this block because this method checks for `$this->cart_contains_renewal()` and the new helper method does not. I am not sure if I should add `$this->cart_contains_renewal()` to the helper method actually 🤔

Mayisha: Good point about `$this->cart_contains_renewal()`. 
I did some functional tests again, I realized that this condition is not required at all. I added this in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4340 but checking if the cart has any renewal isn't needed here. 
We are good to remove this and using the one from the helper is sufficient.

Mayisha (approved):
Thanks @wjrosa for implementing this. When manual renewal is enabled, I can now purchase a subscription and then it is set to manual renewal. 

✅ Card: renewal is set as automatic.
✅ iDeal: renewal is set as automatic.
✅ Multibanco: renewal is set as manual.
✅ EPS: renewal is set as manual.

I have one suggestion left about cleanup https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4401#discussion_r2155161619. I have tested the steps in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4340 and confirmed that using the helper method is sufficient.

client/blocks/utils.js
```diff
@@ -20,14 +20,16 @@ export const getBlocksConfiguration = () => {
...
-	const hasAutoRenewingSubscription =
-		config?.cartContainsSubscription &&
-		! config?.subscriptionRequiresManualRenewal;
-	return hasAutoRenewingSubscription || showSaveOption;
+export const shouldSetupOffSessionPayment = (
+	showSaveOption,
+	paymentMethodReusable
+) => {
+	return (
+		hasAutoRenewingSubscription( paymentMethodReusable ) || showSaveOption
```
daledupreez: Non-blocking nit: I think it's easier to read if we check the local boolean before calling the function.
```suggestion
		showSaveOption || hasAutoRenewingSubscription( paymentMethodReusable )
```

daledupreez: Another small, non-blocking nit: I would also prefer the variable naming to make it super-clear that the values are booleans.
```suggestion
export const shouldSetupOffSessionPayment = (
	shouldShowSaveOption,
	isPaymentMethodReusable
) => {
	return (
		hasAutoRenewingSubscription( isPaymentMethodReusable ) || shouldShowSaveOption
```

client/blocks/utils.js
```diff
@@ -180,3 +182,36 @@ export const getStripeImageUrl = ( imageName ) => {
...
+ * @param {boolean} paymentMethodReusable Indicates if the payment method is reusable.
+ * @return {boolean} True if the cart contains an auto-renewing subscription, false otherwise.
+ */
+const hasAutoRenewingSubscription = ( paymentMethodReusable ) => {
+	const config = getBlocksConfiguration();
+	return (
+		config?.cartContainsSubscription &&
+		! isManualRenewalRequired( paymentMethodReusable )
+	);
+};
```
daledupreez: Another non-blocking naming suggestion for the boolean argument.
```suggestion

/**
 * Whether manual renewal is required based on the payment method's reusability.
 *
 * It is considered required if:
 * - The payment method is not reusable and manual renewal is enabled in the configuration.
 * - The configuration explicitly requires manual renewal.
 *
 * @param {boolean} isReusablePaymentMethod
 * @return {boolean} True if manual renewal is required, false otherwise.
 */
const isManualRenewalRequired = ( isReusablePaymentMethod ) => {
	const config = getBlocksConfiguration();
	return (
		( ! isReusablePaymentMethod &&
			config?.subscriptionManualRenewalEnabled ) ||
		config?.subscriptionRequiresManualRenewal
	);
};

/**
 * Checks if the cart contains an auto-renewing subscription.
 *
 * @param {boolean} isReusablePaymentMethod Indicates if the payment method is reusable.
 * @return {boolean} True if the cart contains an auto-renewing subscription, false otherwise.
 */
const hasAutoRenewingSubscription = ( isReusablePaymentMethod ) => {
	const config = getBlocksConfiguration();
	return (
		config?.cartContainsSubscription &&
		! isManualRenewalRequired( isReusablePaymentMethod )
	);
};
```

readme.txt
```diff
@@ -112,6 +112,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 
 = 9.6.0 - xxxx-xx-xx =
 
+* Add - Voucher payment methods (Boleto, Multibanco, and Oxxo) can now be used when purchasing subscriptions if manual renewals are enabled or required.
```
daledupreez: Nit: trailing `.`
```suggestion
* Add - Voucher payment methods (Boleto, Multibanco, and Oxxo) can now be used when purchasing subscriptions if manual renewals are enabled or required
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
+* Add - Voucher payment methods (Boleto, Multibanco, and Oxxo) can now be used when purchasing subscriptions if manual renewals are enabled or required.
```
daledupreez: Nit: trailing `.`
```suggestion
* Add - Voucher payment methods (Boleto, Multibanco, and Oxxo) can now be used when purchasing subscriptions if manual renewals are enabled or required
```

daledupreez (commented):
I have some minor, non-blocking feedback from inspection; I will try to test tomorrow.

includes/compat/trait-wc-stripe-subscriptions-utilities.php
wjrosa: Thanks for the functional tests, Mayisha! I have updated the implementation in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4401/commits/457b9baa407c6d6dbe6b354f9a65d0cc145cdd08

client/blocks/utils.js
wjrosa: I could not auto-apply your suggestions, but I manually did them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4401/commits/e53a28ee1cbc17fc0b661c8e74a45c24c31564cb

daledupreez (approved):
Changes LGTM and test well. I didn't cover shortcode + block checkout for all three payment methods, but I confirmed that purchases work for all three payment methods, and I checked the behaviour with both "allow manual" and "require manual" with Multibanco.

PR #4402, opened 2025-06-07
Refactor Account to use the new Database Cache

diegocurbelo:

This PR replaces the `transient` used for the Stripe Account Data cache with the new `WC_Stripe_Database_Cache` implementation.

It also replaces the previous constants for test/live mode transient keys, with just one key, as the new cache automatically handles the plugin mode.

## Testing instructions

1. Go to `WP-Admin > WooCommerce > Payments`
2. Select the Stripe payment gateway (click `Manage`)
3. Go to `Dev Tools` Database Cache tab, and check that the cache entry (`wc_stripe_cache_test_account_data`) is present, and note the `Updated` column value for it.
    Screenshot 2025-07-02 at 18 00 52
4. Go back to the Stripe Gateway Settings (or refresh the page if open in another tab).
5. Go to `Dev Tools` Database Cache tab, and check that the updated value for the cache entry `wc_stripe_cache_test_account_data` hasn't changed.
6. Delete the cache entry (using the `Delete` button in that row)
7. Go back to the Stripe Gateway Settings (or refresh the page if open in another tab).
8. Go back to the `Dev Tools` Database Cache tab, and check that a new cache entry was added for `wc_stripe_cache_test_account_data` with the TTL defined in the constant ACCOUNT_CACHE_EXPIRATION (2 hours)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (approved):
Thanks for the refactor. LGTM.

✅ Account data is now stored in the database cache.
✅ Test and live data are correctly stored with their corresponding keys.
✅ Account refresh updates the cache.

Screenshot 2025-07-03 at 3 44 08 PM

includes/class-wc-stripe-account.php
```diff
@@ -99,7 +110,7 @@ public function get_cached_account_data( $mode = null ) {
 	 * @return array empty when no data found in transient, otherwise returns cached data
 	 */
 	private function read_account_from_cache( $mode = null ) {
```
malithsen: nit: `$mode` is no longer used. We can remove that parameter here and update line 97 above which appears to be the only place we call this function.

malithsen (approved):
LGTM. Commented on an additional code cleanup we can do, but that's non-blocking.

includes/class-wc-stripe-account.php
diegocurbelo: Nice catch... removed in [3355f10](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4402/commits/3355f1098911983923731581cab85843ae3449e1)

malithsen (approved):
LGTM. Thanks for the refactor Diego!

PR #3063, opened 2024-04-10
Fixes orders that require manual review due to risk/fraud rules being marked as processing/completed before approval

mattallan:
Fixes #2679 

## Changes proposed in this Pull Request:

> [!NOTE]
> This PR requires a bit more testing but I'm pushing up my current changes to get an initial review of the approach taken. I've only tested basic flows like processing high-risk fraud payments, payments flagged for manual review via custom rules, and regular non-risk payments. Areas I still need to test:
> - [x] how subscription purchases and renewals are handled
> - [x] 3DS cards
> - [x] APM that requires redirect
> - [x] `review.open` webhook sent before and after payment_intent/charge.succeeded webhook

The way we handle payments/charges that have been flagged for manual review by Stripe is we do the following:
1. Process the `payment_intent.succeeded` webhook and mark the order as processing/completed
2. If the charge requires manual review, Stripe will send an `review.open` event, which we listen for and put that order on-hold and store its current status in `_stripe_status_before_hold` metadata.
3. Store manager reviews the risky payment and approves it from their Stripe dashboard
4. We then process the `review.closed` webhook by restoring the status of the order back to its original status stored in `_stripe_status_before_hold`.

This flow relies heavily on the order in which these webhooks are sent/processed.
The succeeded event needs to come first and then `review.open`.

The issue described in #2679 is what happens when the `review.open` webhook is sent/processed first, the order is put on-hold and then the `payment_intent.succeeded` event is handled soon after, which then updates the status to processing. This results in the order in WooCommerce being processing/completed and ready to ship, but still waiting for manual approval in Stripe.

The expected behaviour is for orders to remain on-hold until the payment risk has been manually looked and approved, regardless of the order in which the webhooks are processed.

## Testing instructions

1. Make sure you have Stripe webhooks set up and working with your test store (I personally use ngrok on my local site)
2. Set up a custom review rule in Stripe Dashboard by going to https://dashboard.stripe.com/test/settings/radar/rules
3. Click on `Add Rule` within the Review Rules section:
image
4. Set the "Review if" condition to something that will cause purchases by your test customer to be flagged by the rule. For example I created a rule with **`Review if :email_domain: = 'example.com'`**
5. Go back to your store and attempt to purchase an product using any card.
6. View the order from the WP Admin > Edit Order page and confirm the order is on-hold and has note: image
7. View the order meta and confirm `_stripe_status_before_hold` is set to processing.
8. Go to https://dashboard.stripe.com/test/radar/reviews to see active payments that require review.
[...]

PR #4403, opened 2025-06-09
Fix status and notification for failed refunds

### LINKED ISSUE #4405: Failed refunds flagged as successful
Originated from STRIPE-503

### Describe the bug

When a refund is attempted on an order using the WooCommerce Stripe Gateway and the refund fails (e.g., due to an expired or cancelled card), WooCommerce still marks the order status as "refunded" and sends out refund confirmation emails to both the customer and the admin — even though the refund did not go through.

#### Steps to reproduce the behavior:

Set up WooCommerce with the WooCommerce Stripe Gateway plugin in test mode.

Make a test order with "Asynchronous failure" card from Stripe test cards here: https://docs.stripe.com/testing?testing-method=card-numbers#refunds

Attempt a refund from the order page

Observe that:

The order status is changed to "refunded". (However, a new order note is stating that the refund failed) https://d.pr/i/nPdQcz

Email is sent to the customer indicating the refund was successful. https://d.pr/i/ad3UG5

#### Expected behavior

If a refund fails:

The order status should not be changed to "refunded" (it should remain "completed")

No refund confirmation email should be sent out.

Optionally, a different email could be sent to notify the customer that the refund attempt failed.

#### Screenshots

https://d.pr/i/nPdQcz

https://d.pr/i/6mIKst

#### Additional context

[SystemStatusReport_testsite14405.wpcomstaging.com_2025-06-09T09-32-56.txt](https://uploads.linear.app/f629fcde-6f9b-48e8-8af3-60eed8823bff/998cf098-f51a-470f-b32b-7540bc05d3fb/7d0351c9-58d0-4bd1-aa63-bb6b240707c7): https://uploads.linear.app/f629fcde-6f9b-48e8-8af3-60eed8823bff/998cf098-f51a-470f-b32b-7540bc05d3fb/7d0351c9-58d0-4bd1-aa63-bb6b240707c7

wjrosa:
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4405

Currently, when a refund attempt fails asynchronously, we still send the email notification saying it was successful and keep the "Refunded" order status. That can be confusing for both the merchant and the shopper. This PR fixes that behavior by checking for failures, sending additional notifications, and reverting the "Refunded" order status change.

I am also removing the parameters we used to provide to `WC_Stripe_Email_Failed_Authentication_Retry` because the class does not accept any.

## Testing instructions

- Checkout and build this branch on your test environment (`fix/refund-failed-status-and-notification`)
- Connect your Stripe account
- Make sure webhooks are configured, and you are accessing the test website thought a public-facing URL
- Go to any checkout page
- Complete the purchase using the async refund failure card (`4000000000005126`) ([docs](https://docs.stripe.com/testing?testing-method=card-numbers#refunds))
- As a merchant, attempt to refund the order
- Confirm that you received the refund successfully email, and the order is updated to `Refunded`
- Confirm that short time after it, you get another email communicating the failure, and that the order status was reverted (`Completed` or `Processing`)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -682,6 +682,24 @@ public function send_failed_order_email( $order_id, $status_update = [] ) {
...
+	 */
+	public function send_failed_refund_emails( $order ) {
+		$emails = WC()->mailer()->get_emails();
+
+		if ( empty( $emails ) || empty( $order ) || ! isset( $emails['WC_Stripe_Email_Admin_Failed_Refund'] ) || ! isset( $emails['WC_Stripe_Email_Customer_Failed_Refund'] ) ) {
+			return;
+		}
+
+		$emails['WC_Stripe_Email_Admin_Failed_Refund']->trigger( $order->get_id(), $order );
+		$emails['WC_Stripe_Email_Customer_Failed_Refund']->trigger( $order->get_id(), $order );
```
daledupreez: We should split the conditions up a bit here -- as things are written, we won't send any emails if one of the values is not set.
```suggestion
		if ( empty( $emails ) || empty( $order ) ) {
			return;
		}

		if ( isset( $emails['WC_Stripe_Email_Admin_Failed_Refund'] ) ) {
			$emails['WC_Stripe_Email_Admin_Failed_Refund']->trigger( $order->get_id(), $order );
		}

		if ( isset( $emails['WC_Stripe_Email_Customer_Failed_Refund'] ) ) {
			$emails['WC_Stripe_Email_Customer_Failed_Refund']->trigger( $order->get_id(), $order );
		}
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -820,7 +823,19 @@ public function process_webhook_refund_updated( $notification ) {
...
+					// Store the failure reason
+					if ( isset( $refund_object->failure_reason ) ) {
+						$order->update_meta_data( '_stripe_refund_failure_reason', $refund_object->failure_reason );
+					}
+
+					// Revert to previous status
+					$status_before_refund = $this->get_stripe_order_status_before_refund( $order );
+					if ( in_array( $status_before_refund, apply_filters( 'woocommerce_valid_order_statuses_for_payment_complete', [ OrderStatus::ON_HOLD, OrderStatus::PENDING, OrderStatus::FAILED, OrderStatus::CANCELLED ], $order ), true ) ) {
+						$status_before_refund = apply_filters( 'woocommerce_payment_complete_order_status', $order->needs_processing() ? OrderStatus::PROCESSING : OrderStatus::COMPLETED, $order->get_id(), $order );
+					}
```
daledupreez: I am finding this block of code really confusing and hard to follow. What I _think_ is happening is as follows:
* We are checking if `$status_before_refund` is in the set of order statuses that are valid for order completion
  - If it is in that set, we are then overwriting `$status_before_refund` to be either PROCESSING or COMPLETED
* We are then updating the order status to the value of $status_before_refund`

I think we need a `! in_array()`, as we should be overwriting the value to something that is permitted only if the value is _NOT_ in the set of valid completion status values.
```suggestion
					$valid_payment_complete_statuses = apply_filters( 'woocommerce_valid_order_statuses_for_payment_complete', [ OrderStatus::ON_HOLD, OrderStatus::PENDING, OrderStatus::FAILED, OrderStatus::CANCELLED ], $order );
					if ( ! in_array( $status_before_refund, $valid_payment_complete_statuses, true ) ) {
						$default_status = $order->needs_processing() ? OrderStatus::PROCESSING : OrderStatus::COMPLETED;
						$status_before_refund = apply_filters( 'woocommerce_payment_complete_order_status', $default_status, $order->get_id(), $order );
					}
```

As a side note, the need to hard-code the permitted status values being passed to the filter feels pretty suspect. But I don't see an API or helper for these values. Maybe we should add one?

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+	 */
+	abstract public function get_template_params();
+
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		$this->title   = __( 'Refund request failed', 'woocommerce-gateway-stripe' );
+		$this->heading = __( 'Refund request failed', 'woocommerce-gateway-stripe' );
+		$this->subject = __( '[{site_title}] Refund request failed for #{order_number}.', 'woocommerce-gateway-stripe' );
```
daledupreez: I think we should simplify the language here to say the refund failed, especially seeing as this email will be going to users.
```suggestion
		$this->title   = __( 'Refund failed', 'woocommerce-gateway-stripe' );
		$this->heading = __( 'Refund failed', 'woocommerce-gateway-stripe' );
		$this->subject = __( '[{site_title}] Refund failed for #{order_number}.', 'woocommerce-gateway-stripe' );
```

includes/class-wc-stripe.php
```diff
@@ -686,11 +686,16 @@ public function add_emails( $email_classes ) {
...
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/compat/class-wc-stripe-email-failed-preorder-authentication.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/compat/class-wc-stripe-email-failed-authentication-retry.php';
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/compat/class-wc-stripe-email-failed-refund.php';
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/compat/class-wc-stripe-email-admin-failed-refund.php';
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/compat/class-wc-stripe-email-customer-failed-refund.php';
 
 		// Add all emails, generated by the gateway.
 		$email_classes['WC_Stripe_Email_Failed_Renewal_Authentication']  = new WC_Stripe_Email_Failed_Renewal_Authentication( $email_classes );
 		$email_classes['WC_Stripe_Email_Failed_Preorder_Authentication'] = new WC_Stripe_Email_Failed_Preorder_Authentication( $email_classes );
-		$email_classes['WC_Stripe_Email_Failed_Authentication_Retry']    = new WC_Stripe_Email_Failed_Authentication_Retry( $email_classes );
```
daledupreez, line 715: Nit: we should not remove the `$email_classes` argument unless we document why.

includes/compat/class-wc-stripe-email-admin-failed-refund.php
```diff
@@ -0,0 +1,47 @@
...
+ * @since 9.6.0
+ */
+class WC_Stripe_Email_Admin_Failed_Refund extends WC_Stripe_Email_Failed_Refund {
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->id          = 'failed_refund_admin';
```
daledupreez: Should we use a more specific string to avoid possible collisions with other gateways?
```suggestion
		$this->id          = 'wc_stripe_failed_refund_admin';
```

includes/compat/class-wc-stripe-email-customer-failed-refund.php
```diff
@@ -0,0 +1,46 @@
...
+ * @since 9.6.0
+ */
+class WC_Stripe_Email_Customer_Failed_Refund extends WC_Stripe_Email_Failed_Refund {
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->id          = 'failed_refund_customer';
```
daledupreez: As above RE avoiding collisions.
```suggestion
		$this->id          = 'wc_stripe_failed_refund_customer';
```

templates/emails/plain/failed-refund-admin.php
```diff
@@ -0,0 +1,42 @@
...
+	exit;
+}
+
+echo esc_html( $email_heading ) . "\n\n";
+
+printf(
+	// translators: 1) an order number, 2) the customer's full name, 3) the reason for the failure.
+	esc_html_x(
+		'The refund request for order %1$s from %2$s has failed. Reason: %3$s. The customer was also notified.',
+		'In admin refund failed email',
```
daledupreez: Should we make the "customer was also notified" conditional on whether that email is hooked in at runtime? It feels like we could get this wrong if merchants or admins disable that email.

templates/emails/failed-refund-admin.php
```diff
@@ -0,0 +1,52 @@
...
+do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
+
+<p>
+	<?php
+
+	echo esc_html(
+		sprintf(
+			// translators: 1) an order number, 2) the customer's full name, 3) the reason for the refund request failure.
+			_x(
+				'The refund request for order %1$s from %2$s has failed. Reason:  %3$s. The customer was also notified.',
```
daledupreez: As elsewhere, I think we can simplify "refund request" to just "refund".
```suggestion
				'The refund for order %1$s from %2$s has failed. Reason:  %3$s. The customer was also notified.',
```

templates/emails/failed-refund-customer.php
```diff
@@ -0,0 +1,51 @@
...
+
+<p>
+	<?php
+
+	echo esc_html(
+		sprintf(
+			// translators: 1) an order number, 2) the reason for the refund request failure.
+			_x(
+				'The refund request for order %1$s. Reason:  %2$s.',
+				'In admin refund failed email',
```
daledupreez: Two things:
1. We can remove the "request"
2. This is a customer email
```suggestion
				'The refund for order %1$s. Reason:  %2$s.',
				'In customer refund failed email',
```

templates/emails/plain/failed-refund-customer.php
```diff
@@ -0,0 +1,41 @@
...
+	// translators: 1) an order number, 2) the reason for the failure.
+	esc_html_x(
+		'The refund request for order %1$s has failed. Reason: %2$s.',
+		'In customer refund failed email',
+		'woocommerce-gateway-stripe'
+	),
+	esc_html( $order->get_order_number() ),
+	esc_html( $reason )
+) . "\n\n";
+printf( esc_html__( 'The refunded order is as follows:', 'woocommerce-gateway-stripe' ) ) . "\n\n";
```
daledupreez: I think we should change the wording here. "The refunded order" implies that the order was refunded. I think we should just use "The original order" or "The order details are as follows:".

tests/phpunit/Compat/WC_Stripe_Email_Customer_Failed_Refund_Test.php
```diff
@@ -0,0 +1,52 @@
...
+		$params = $email->get_template_params();
+
+		$this->assertArrayHasKey( 'order', $params );
+		$this->assertArrayHasKey( 'reason', $params );
+		$this->assertArrayHasKey( 'email_heading', $params );
+		$this->assertArrayHasKey( 'sent_to_admin', $params );
+		$this->assertArrayHasKey( 'plain_text', $params );
+		$this->assertArrayHasKey( 'email', $params );
+
+		$this->assertEquals( $order, $params['order'] );
```
daledupreez, line 43: Can we also test some of the basic fields to make sure they are pulling from the expected source values?

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
```diff
@@ -700,4 +701,93 @@ public function provide_test_process_webhook_charge_succeeded_skipped_for_synchr
...
+
+		$this->mock_webhook_handler->process_webhook_refund_updated( $notification );
+
+		if ( ! empty( $expected_note ) ) {
+			// Check if the order note was added correctly.
+			$order_id = $notification->data->object->charge;
+			$order    = wc_get_order( $order_id );
+			if ( ! $order ) {
+				// If the order is not found, we cannot proceed with the assertions.
+				return;
```
daledupreez: Should we not return an error for this case? This feels like a failure we should flag rather than ignore.

daledupreez (commented):
I have a number of comments from inspection.

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
wjrosa: Added a forced failure in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/59581b775d9d7ee6e4db97b179097f2562143638

templates/emails/plain/failed-refund-customer.php
wjrosa: Thanks! I was struggling to find a good phrase. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/54cc55ff1c3c47eebc209a1f2de226625b4bd739

tests/phpunit/Compat/WC_Stripe_Email_Customer_Failed_Refund_Test.php
wjrosa: Can you give me some examples of what you expect to be tested here? Do you mean like the target email address? If so, I can create another test for those. Here, I am testing the `get_template_params` method specifically.

includes/compat/class-wc-stripe-email-admin-failed-refund.php
wjrosa: I think it makes sense. I used this format to follow the existing standard (all the other emails lack the `wc_stripe` prefix). I just wanted to note in case we want to update them in the future. I have applied your suggestion.

tests/phpunit/Compat/WC_Stripe_Email_Customer_Failed_Refund_Test.php
daledupreez: Sorry I wasn't more specific! I was wanting to make sure that the values actually are what we expect, and not just that the values are defined. The only value we are checking is `order`, and not any of the other fields.

includes/class-wc-stripe.php
wjrosa: Alright. I reverted the change in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/71e01c053a92ec5c5270ebb767dc2a88859a74b5. I just removed it here because it is not used (`WC_Stripe_Email_Failed_Authentication_Retry` constructor doesn't even accept parameters).

tests/phpunit/Compat/WC_Stripe_Email_Customer_Failed_Refund_Test.php
wjrosa: Gotcha! Updated the tests in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/7164648bd77771fb9c92959a6fd26866cf424ed2. I am now just checking for the value, since it will already fail if the keys don't exist.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Good catch! This was actually inherited from the `get_stripe_order_status_before_hold` logic, introduced [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3063/files). I think your reasoning makes sense, so I accepted the suggestion 👍 . We may need to update the `get_stripe_order_status_before_hold` logic as well, but I will do that on another PR to avoid non-related changes here.

wjrosa: > As a side note, the need to hard-code the permitted status values being passed to the filter feels pretty suspect. But I don't see an API or helper for these values. Maybe we should add one?

You mean like adding a constant to wrap those values or something like that? (`OrderStatus::PAYMENT_COMPLETE_STATUSES`)

daledupreez: > You mean like adding a constant to wrap those values or something like that? (`OrderStatus::PAYMENT_COMPLETE_STATUSES`)

Exactly. It would be a contribution to WooCommerce core that's not urgent, but it feels like it would be a small, helpful improvement in the long run.

includes/class-wc-stripe.php
daledupreez: We can keep it! We should just have a note like your comment above: the constructor does't accept arguments, so it was already a no-op.

includes/compat/class-wc-stripe-email-admin-failed-refund.php
daledupreez: Backwards compatibility is tough. 😐 

I vote we don't touch older emails, but we can probably align on plugin-specific IDs moving forward.

includes/class-wc-stripe-webhook-handler.php
wjrosa: I see! I left a draft for this idea here https://github.com/woocommerce/woocommerce/pull/58936

includes/class-wc-stripe.php
wjrosa: I have included it again in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/2d975a81c40d5844be3374740b79da5a87506eb7 with a simple comment (I wasn't much sure where to include the comment, let me know if you think it is OK that way).

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
+* Fix - Correctly notifies customers and merchants of a failed refund and reverts the refunded status.
```
daledupreez: Nit: can you remove the trailing `.` as per our recent changelog style discussions?
```suggestion
* Fix - Correctly notifies customers and merchants of a failed refund and reverts the refunded status
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -820,7 +823,21 @@ public function process_webhook_refund_updated( $notification ) {
 						$note = sprintf( __( 'Refund canceled for %1$s - Refund ID: %2$s - Reason: %3$s', 'woocommerce-gateway-stripe' ), $amount, $refund_object->id, $refund_object->failure_reason );
 					}
 
-					$order->add_order_note( $note );
+					// Store the failure reason
+					if ( isset( $refund_object->failure_reason ) ) {
+						$order->update_meta_data( '_stripe_refund_failure_reason', $refund_object->failure_reason );
+					}
```
daledupreez: Is it sufficient to store only `failure_reason`? I can see it might be useful to capture `$refund_object->status`, as that explicitly indicates why the refund was not processed.

includes/class-wc-stripe.php
daledupreez: I think it's OK to just have a comment on PR -- we don't need that in the code forever.

My complaint was mostly: this change looks unrelated to this PR, so it would be really good to make it clear to the reviewer (and future developers) why this change was made. We can remove the inline comment.

includes/compat/class-wc-stripe-email-admin-failed-refund.php
```diff
@@ -0,0 +1,47 @@
...
+ */
+class WC_Stripe_Email_Admin_Failed_Refund extends WC_Stripe_Email_Failed_Refund {
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->id          = 'wc_stripe_failed_refund_admin';
+		$this->description = __( 'Refund request failure emails are sent to chosen recipient(s) when an attempt to process refund fails.', 'woocommerce-gateway-stripe' );
```
daledupreez: Nit: We can simplify to "Refund failure", and I think we should add an `a` to the description.
```suggestion
		$this->description = __( 'Refund failure emails are sent to chosen recipient(s) when an attempt to process a refund fails.', 'woocommerce-gateway-stripe' );
```

includes/compat/class-wc-stripe-email-customer-failed-refund.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+class WC_Stripe_Email_Customer_Failed_Refund extends WC_Stripe_Email_Failed_Refund {
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->id          = 'wc_stripe_failed_refund_customer';
+		$this->description = __( 'Sent to a customer when a refund request fails. The email contains the original order information.', 'woocommerce-gateway-stripe' );
```
daledupreez: To match the details, I think it's worth simplifying and removing the "request". I also think we should call out that it occurs for cancellations, so merchants know that will happen.
```suggestion
		$this->description = __( 'Sent to a customer when a refund fails or is cancelled. The email contains the original order information.', 'woocommerce-gateway-stripe' );
```

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+	protected static function get_refund_reason_description( $refund_reason_key ) {
+		switch ( $refund_reason_key ) {
+			case 'charge_for_pending_refund_disputed':
+				return __( 'Customer disputed the charge while the refund is pending.', 'woocommerce-gateway-stripe' );
+			case 'declined':
+				return __( 'Declined by financial partners.', 'woocommerce-gateway-stripe' );
+			case 'expired_or_canceled_card':
+				return __( 'Payment method is canceled by the customer or expired by the merchant.', 'woocommerce-gateway-stripe' );
+			case 'insufficient_funds':
+				return __( 'Insufficient funds and has crossed the pending refund expiry window.', 'woocommerce-gateway-stripe' );
```
daledupreez: Is this an accurate description of the state? Is the only time we will show this when we've exceeded the refund expiry window?

Separately, I think we should phrase this more simply:
```suggestion
				return __( 'Insufficient funds and after the refund period ended.', 'woocommerce-gateway-stripe' );
```

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+	 * @return string
+	 */
+	protected static function get_refund_reason_description( $refund_reason_key ) {
+		switch ( $refund_reason_key ) {
+			case 'charge_for_pending_refund_disputed':
+				return __( 'Customer disputed the charge while the refund is pending.', 'woocommerce-gateway-stripe' );
+			case 'declined':
+				return __( 'Declined by financial partners.', 'woocommerce-gateway-stripe' );
+			case 'expired_or_canceled_card':
+				return __( 'Payment method is canceled by the customer or expired by the merchant.', 'woocommerce-gateway-stripe' );
```
daledupreez: Can we use "was" here? I think it more clearly communicates that the action by the customer or merchant already occurred. And how did the merchant expire the card? I think we should avoid specifying who or what performed the action, and simply communicate the end state.
```suggestion
				return __( 'The payment method has expired or was canceled.', 'woocommerce-gateway-stripe' );
```

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+			case 'insufficient_funds':
+				return __( 'Insufficient funds and has crossed the pending refund expiry window.', 'woocommerce-gateway-stripe' );
+			case 'lost_or_stolen_card':
+				return __( 'Failed due to loss or theft of the original card.', 'woocommerce-gateway-stripe' );
+			case 'merchant_request':
+				return __( 'Failed upon the merchant’s request.', 'woocommerce-gateway-stripe' );
```
daledupreez: I think we should be careful about the wording for this specific item. If we're sending this to customers, it feels like we should be careful about how we tell them about this item. How about we say something like: "Was cancelled by the merchant"?
```suggestion
				return __( 'Was canceled by the merchant.', 'woocommerce-gateway-stripe' );
```

readme.txt
```diff
@@ -112,6 +112,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 
 = 9.6.0 - xxxx-xx-xx =
 
+* Fix - Correctly notifies customers and merchants of a failed refund and reverts the refunded status.
```
daledupreez, line 115: Nit: another trailing `.`
```suggestion
* Fix - Correctly notifies customers and merchants of a failed refund and reverts the refunded status
```

templates/emails/failed-refund-admin.php
```diff
@@ -0,0 +1,52 @@
...
+				'woocommerce-gateway-stripe'
+			),
+			$order->get_order_number(),
+			$order->get_formatted_billing_full_name(),
+			$reason
+		)
+	);
+	?>
+</p>
+<p><?php esc_html_e( 'The renewal order is as follows:', 'woocommerce-gateway-stripe' ); ?></p>
```
daledupreez: I think we should use the same language as elsewhere.
```suggestion
<p><?php esc_html_e( 'The order details are as follows:', 'woocommerce-gateway-stripe' ); ?></p>
```

templates/emails/failed-refund-customer.php
```diff
@@ -0,0 +1,51 @@
...
+				'In customer refund failed email',
+				'woocommerce-gateway-stripe'
+			),
+			$order->get_order_number(),
+			$reason
+		)
+	);
+	?>
+</p>
+<p><?php esc_html_e( 'The refunded order is as follows:', 'woocommerce-gateway-stripe' ); ?></p>
```
daledupreez: As above RE consistency.
```suggestion
<p><?php esc_html_e( 'The order details are as follows:', 'woocommerce-gateway-stripe' ); ?></p>
```

templates/emails/plain/failed-refund-admin.php
```diff
@@ -0,0 +1,42 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+echo esc_html( $email_heading ) . "\n\n";
+
+printf(
+	// translators: 1) an order number, 2) the customer's full name, 3) the reason for the failure.
+	esc_html_x(
+		'The refund request for order %1$s from %2$s has failed. Reason: %3$s. The customer was also notified.',
```
daledupreez: Consistency: remove "request"
```suggestion
		'The refund for order %1$s from %2$s has failed. Reason: %3$s. The customer was also notified.',
```

templates/emails/failed-refund-customer.php
```diff
@@ -0,0 +1,51 @@
...
+do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
+
+<p>
+	<?php
+
+	echo esc_html(
+		sprintf(
+			// translators: 1) an order number, 2) the reason for the refund request failure.
+			_x(
+				'The refund for order %1$s. Reason:  %2$s.',
```
daledupreez: I think we need to include "has failed".
```suggestion
				'The refund for order %1$s has failed. Reason:  %2$s.',
```

templates/emails/plain/failed-refund-customer.php
```diff
@@ -0,0 +1,41 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+echo esc_html( $email_heading ) . "\n\n";
+
+printf(
+	// translators: 1) an order number, 2) the reason for the failure.
+	esc_html_x(
+		'The refund request for order %1$s has failed. Reason: %2$s.',
```
daledupreez: Consistency: remove "request"
```suggestion
		'The refund for order %1$s has failed. Reason: %2$s.',
```

tests/phpunit/Compat/WC_Stripe_Email_Failed_Refund_Test.php
```diff
@@ -0,0 +1,119 @@
...
+	/**
+	 * Provider for the `test_trigger` method.
+	 *
+	 * @return array
+	 */
+	public function provide_test_trigger() {
+		return [
+			'not enabled'  => [
+				'is enabled'     => false,
+				'recipient'      => '',
```
daledupreez: I think this test should include a recipient so we are more likely to verify that it is not sent based on the enabled flag.
```suggestion
				'recipient'      => 'test@example.com',
```

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
```diff
@@ -6,7 +6,11 @@
 use stdClass;
 use WC_Gateway_Stripe;
 use WC_Gateway_Stripe_Giropay;
+use WC_Order;
+use WC_Stripe;
 use WC_Stripe_Customer;
+use WC_Stripe_Email_Customer_Failed_Refund;
+use WC_Stripe_Email_Failed_Refund;
```
daledupreez: AFAICT, the changes in this file are only cosmetic. Am I missing something?

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
```diff
@@ -700,4 +701,91 @@ public function provide_test_process_webhook_charge_succeeded_skipped_for_synchr
...
+
+			$notes = wc_get_order_notes(
+				[
+					'order_id' => $order->get_id(),
+					'limit'    => 1,
+				]
+			);
+			$this->assertCount( 1, $notes );
+			$this->assertSame( $expected_note, $notes[0]->content );
+		}
```
daledupreez: Should we also verify that no notes are added when `$expected_note` is empty?

Also, I am not sure why we are (re)fetching the order, when all we need to look at are the notes for the order?
```suggestion
		$notes = wc_get_order_notes(
			[
				'order_id' => $order->get_id(),
				'limit' => 1,
			]
		);

		if ( empty( $expected_note ) ) {
			$this->assertEquals( [], $notes );
			return;
		}
		
		$this->assertCount( 1, $notes );
		$this->assertSame( $expected_note, $notes[0]->content );
```

daledupreez (commented):
I have more feedback from inspection, mostly around wording, TBH.

I'll switch to testing now

includes/class-wc-stripe-webhook-handler.php
wjrosa: Alright! Storing it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/5e00794e629574d3e7eb7fc64e7c9fb9b58a6b06

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
wjrosa: Yes. I tried to include a specific unit test here for `send_failed_refund_emails` (spent some hours on it), but for some reason, `WC()->mailer()->get_emails()` is not behaving as I wanted in tests. The `woocommerce_email_classes` filter is never hooked. So I gave up on adding it here, I may try again on another PR. I have reverted the changes now.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2268,24 +2291,69 @@ private function needs_ssl_setup() {
...
+			$current_status          = apply_filters( 'woocommerce_payment_complete_order_status', $payment_complete_status, $order->get_id(), $order );
+		}
+		if ( OrderStatus::ON_HOLD === $target_status ) {
+			$meta_key = '_stripe_status_before_hold';
+		}
+		if ( OrderStatus::REFUNDED === $target_status ) {
+			$meta_key = '_stripe_status_before_refund';
+		}
+		if ( empty( $meta_key ) ) {
+			WC_Stripe_Logger::log( 'Error: Unable to set the status before event. Meta key is not defined.' );
```
daledupreez: Nit: It feels like we should include more details for debugging purposes.
```suggestion
			$log_message = sprintf( 'Error: Unable to set the order status for order %d when transitioning from %s to %s.', $order->get_id(), $current_status, $target_status );
			WC_Stripe_Logger::log( $log_message );
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -820,7 +825,21 @@ public function process_webhook_refund_updated( $notification ) {
...
```
daledupreez: I think we should have this update and the update above occur in all cases, as it's possible that multiple attempts are made over time, and the two values should stay in sync, especially when the reason is empty.
```suggestion
					$order->update_meta_data( '_stripe_refund_status', $refund_object->status );
					if ( isset( $refund_object->failure_reason ) ) {
						$order->update_meta_data( '_stripe_refund_failure_reason', $refund_object->failure_reason );
					} else {
						$order->delete_meta_data( '_stripe_refund_failure_reason' );
					}
```

daledupreez (commented):
I have two small comments from inspection; switching to testing now.

Also, there is a legitimate linting error due to a trailing whitespace somewhere. 🙃

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+	 *
+	 * @return array
+	 */
+	abstract public function get_template_params();
+
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		$this->title   = __( 'Refund failed', 'woocommerce-gateway-stripe' );
```
daledupreez: After seeing this in context, I think we should be more specific with the title so it's clearer that it's a Stripe refund.
```suggestion
		$this->title   = __( 'Stripe refund failed', 'woocommerce-gateway-stripe' );
```

woo_email_settings

daledupreez (commented):
This seems to be working well for me overall, but I am not seeing the customer email. Will keep testing.

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+				return __( 'Payment method is canceled by the customer or expired by the merchant.', 'woocommerce-gateway-stripe' );
+			case 'insufficient_funds':
+				return __( 'Insufficient funds and has crossed the pending refund expiry window.', 'woocommerce-gateway-stripe' );
+			case 'lost_or_stolen_card':
+				return __( 'Failed due to loss or theft of the original card.', 'woocommerce-gateway-stripe' );
+			case 'merchant_request':
+				return __( 'Failed upon the merchant’s request.', 'woocommerce-gateway-stripe' );
+			case 'unknown':
+			default:
+				return __( 'Unknown reason', 'woocommerce-gateway-stripe' );
```
daledupreez: Nit: all the other reasons include a trailing `.`. We should also include one here, or remove the `.` from all the other values, just so we are consistent.
```suggestion
				return __( 'Unknown reason.', 'woocommerce-gateway-stripe' );
```

templates/emails/plain/failed-refund-admin.php
daledupreez: Another small issue: the reason values mostly include a trailing `.`, so we should not add another one.
```suggestion
		'The refund for order %1$s from %2$s has failed. Reason: %3$s The customer was also notified.',
```

templates/emails/plain/failed-refund-customer.php
```diff
@@ -0,0 +1,41 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+echo esc_html( $email_heading ) . "\n\n";
+
+printf(
+	// translators: 1) an order number, 2) the reason for the failure.
+	esc_html_x(
+		'The refund for order %1$s has failed. Reason: %2$s.',
```
daledupreez, line 11: The reason value has a trailing `.`. We should either fix the source strings or remove this trailing `.`.

templates/emails/failed-refund-customer.php
```diff
@@ -0,0 +1,51 @@
...
+do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
+
+<p>
+	<?php
+
+	echo esc_html(
+		sprintf(
+			// translators: 1) an order number, 2) the reason for the refund request failure.
+			_x(
+				'The refund for order %1$s has failed. Reason:  %2$s.',
```
daledupreez, line 19: Another double `.` situation.

templates/emails/failed-refund-admin.php
```diff
@@ -0,0 +1,52 @@
...
+do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
+
+<p>
+	<?php
+
+	echo esc_html(
+		sprintf(
+			// translators: 1) an order number, 2) the customer's full name, 3) the reason for the refund request failure.
+			_x(
+				'The refund for order %1$s from %2$s has failed. Reason:  %3$s. The customer was also notified.',
```
daledupreez, line 19: Another double `.` situation.

daledupreez (commented):
Things are looking good and testing well, but I think we can polish things a bit before we ship this.

daledupreez:
I was wrong -- I am still not seeing the customer-facing email. I received the refund email as the customer, and I received the admin email for the refund failed, but I did not receive the customer email. Something definitely seems weird/unexpected.

includes/compat/class-wc-stripe-email-failed-refund.php
wjrosa: I decided to go the other way around, and remove the `.` from all the refund failure reasons in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/b4c8ee24207961534a577c15b5fb8c22a8170360

templates/emails/plain/failed-refund-admin.php
wjrosa: Since I decided to remove the `.` from the refund failure reason strings, I just updated the messaging here by removing the `request` word in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/719e420582fc65916cb9397f9649ec4db5083db1

wjrosa:
> I was wrong -- I am still not seeing the customer-facing email. I received the refund email as the customer, and I received the admin email for the refund failed, but I did not receive the customer email. Something definitely seems weird/unexpected.

Can you try again, please? I have enforced the customer's recipient address in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/33227e7fb5a0c3369970655bba5bbb408ec85808. I did some tests here and it worked for me

includes/compat/class-wc-stripe-email-failed-refund.php
daledupreez: Works for me! I think not having it will be more flexible in the long run.

includes/compat/class-wc-stripe-email-admin-failed-refund.php
```diff
@@ -0,0 +1,55 @@
...
+			'email'         => $this,
+		];
+	}
+
+	/**
+	 * Trigger.
+	 *
+	 * @inheritDoc
+	 */
+	public function trigger( $order_id, $order = false ) {
```
daledupreez, line 50: These `trigger()` overrides feel like they may not be the best approach to forcing the right email/recipient, but they are working. Perhaps someone else on the team knows another way for us to tackle this. (Definitely not blocking for now.)

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+	 */
+	abstract public function get_template_params();
+
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		$this->title   = __( 'Stripe refund failed', 'woocommerce-gateway-stripe' );
+		$this->heading = __( 'Refund failed', 'woocommerce-gateway-stripe' );
+		$this->subject = __( '[{site_title}] Refund failed for #{order_number}.', 'woocommerce-gateway-stripe' );
```
daledupreez: After seeing this in the emails, I think we should adjust this slightly so it is more aligned with the other emails we are sending. The changes are to use `order #{order_number}` and to remove the trailing `.`.
```suggestion
		$this->subject = __( '[{site_title}] Refund failed for order #{order_number}', 'woocommerce-gateway-stripe' );
```

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+			case 'declined':
+				return __( 'Declined by financial partners', 'woocommerce-gateway-stripe' );
+			case 'expired_or_canceled_card':
+				return __( 'Payment method is canceled by the customer or expired by the merchant', 'woocommerce-gateway-stripe' );
+			case 'insufficient_funds':
+				return __( 'Insufficient funds and has crossed the pending refund expiry window', 'woocommerce-gateway-stripe' );
+			case 'lost_or_stolen_card':
+				return __( 'Failed due to loss or theft of the original card', 'woocommerce-gateway-stripe' );
+			case 'merchant_request':
+				return __( 'Failed upon the merchant’s request', 'woocommerce-gateway-stripe' );
```
daledupreez: I think we should clean these up in a follow-up PR. I would propose the following wording:
* `charge_for_pending_refund_disputed` - _The charge has been disputed_
* `declined` - _The refund was declined_ - "our financial partners" is the issuing bank/wallet/etc as Stripe's partner
* `expired_or_canceled_card` - _The original payment method has expired or was canceled_
* `insufficient_funds` - _We could not process the refund at this time_ - We should NEVER tell customers that the merchant has insufficient funds. 😬 
* `lost_or_stolen_card` - _The original payment method was lost or stolen_
* `merchant_request` - _We stopped processing the refund_ - I think this should be clearer for users.

daledupreez (approved):
The `trigger()` overrides worked for me! I think we have some follow-up work, but nothing blocking.

daledupreez:
@wjrosa, also the test failures are legitimate. I can pick this up tomorrow and get it over the finish line.

wjrosa:
> @wjrosa, also the test failures are legitimate. I can pick this up tomorrow and get it over the finish line.

It was simple to fix the tests, so I did it now. But I'd appreciate it if you could move this PR forward if there is more work needed 🙏

includes/class-wc-stripe.php
```diff
@@ -686,11 +686,16 @@ public function add_emails( $email_classes ) {
...
+		$email_classes['WC_Stripe_Email_Failed_Authentication_Retry']    = new WC_Stripe_Email_Failed_Authentication_Retry(); // $email_classes parameter removed. This class does not accept it.
```
Mayisha: QQ: Was this `// $email_classes parameter removed. This class does not accept it.` supposed to be a comment in the PR but mistakenly put in the code? 🤔

includes/compat/class-wc-stripe-email-admin-failed-refund.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+class WC_Stripe_Email_Admin_Failed_Refund extends WC_Stripe_Email_Failed_Refund {
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		parent::__construct();
+
+		$this->id          = 'wc_stripe_failed_refund_admin';
+		$this->description = __( 'Refund failure emails are sent to chosen recipient(s) when an attempt to process a refund fails.', 'woocommerce-gateway-stripe' );
```
Mayisha:
```suggestion
		$this->description = __( 'Refund failure emails are sent to the admin when an attempt to process a refund fails.', 'woocommerce-gateway-stripe' );
```
The description for the customer's one is specific, so this one could be specific too.

includes/compat/class-wc-stripe-email-failed-refund.php
Mayisha:
```suggestion
		$this->subject = __( '[{site_title}]: Refund failed for #{order_number}', 'woocommerce-gateway-stripe' );
```

For alignment we can put a `:` after the site name.

includes/compat/class-wc-stripe-email-failed-refund.php
```diff
@@ -0,0 +1,133 @@
...
+	public function trigger( $order_id, $order = false ) {
+		$this->object = $order;
+
+		if ( ! $this->is_enabled() || ! $this->get_recipient() ) {
+			return;
+		}
+
+		$this->find['order-number']    = '{order_number}';
+		$this->replace['order-number'] = $this->object->get_order_number();
+
```
Mayisha: I see in my editor that `$this->find` and `$this->replace` have been deprecated in 3.2.0. Not a blocker but we should consider using the latest way, may be in a refactor.

templates/emails/failed-refund-admin.php
Mayisha: Typo: `renewal` order.
Unresolved the comment as it was not addressed cc: @wjrosa

templates/emails/failed-refund-customer.php
Mayisha: Unresolved, as I think this should be updated as well @wjrosa. Let's not call it a `refunded` order.

Mayisha (commented):
I have a few minor comments/observations. 
Apart from that there is only one issue with partial refund. When a partial refund fails, it is not tracked in the order notes and no email is sent to the admin and customer.

includes/class-wc-stripe.php
wjrosa: Not really, @daledupreez asked for some documentation in code to explain why I removed the parameter. I could not find a good way to do it

includes/compat/class-wc-stripe-email-failed-refund.php
wjrosa: It makes sense, Mayisha. I have updated the code with the new standard already in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/acf140727a91e3cf21674faa8fd86353e2c0c42e

includes/compat/class-wc-stripe-email-failed-refund.php
wjrosa: Thanks for the suggestions, Dale! To speed up things a bit I already applied them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/962820ff90f6f98c8907247109c518d4dc1bf2fe

> We should NEVER tell customers that the merchant has insufficient funds.

Can you elaborate? Since these refund reasons are retrieved from Stripe, the merchant could just look there anyway.

wjrosa: Just to add to this, I got the original refund reasons from [their docs](https://docs.stripe.com/refunds#failed-refunds).

includes/class-wc-stripe.php
daledupreez: Docs in the PR is sufficient, @wjrosa! I must not have been sufficiently clear.  My main concern was that the parameter was just removed, but the PR didn't explain why.

includes/compat/class-wc-stripe-email-failed-refund.php
daledupreez: My main concern is that we call `get_reason()` for both the admin and the customer emails. So we need to be very careful about what we tell _customers_ about the reason for the refund. Also, we only have one helper method and no way to specify whether the recipient is a customer or admin/merchant, so I went with more generic messages.

We could also solve this by having a parameter to `get_reason()` that specifies whether the recipient is a customer.

includes/class-wc-stripe.php
wjrosa: Oh OK, that is much better. Removed the comment from the code in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/47a4d3cc78819edb33aac2db9606737a5e4bb56a and updated the PR description.

includes/compat/class-wc-stripe-email-failed-refund.php
wjrosa: That's a good point 👍 . Thank you for elaborating. I think we can leave it like this for now and maybe update the method later if needed to make it more straightforward for merchants.

Mayisha:
> It makes sense, Mayisha. I have updated the code with the new standard already in [acf1407](https://github.com/woocommerce/woocommerce-gateway-stripe/commit/acf140727a91e3cf21674faa8fd86353e2c0c42e)

@wjrosa the update you mentioned in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/files#r2161745838) is made on a different file instead of `includes/compat/class-wc-stripe-email-failed-refund.php`.

> Apart from that there is only one issue with partial refund. When a partial refund fails, it is not tracked in the order notes and no email is sent to the admin and customer.

This issue I mentioned in the [previous comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403#pullrequestreview-2950213963) is still present.

wjrosa:
> @wjrosa the update you mentioned in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/files#r2161745838) is made on a different file instead of `includes/compat/class-wc-stripe-email-failed-refund.php`.

Oops 🤦 ! Silly mistake, sorry! Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/e4de235d8515a6edb3e9c77e737e12cd057c508c

> This issue I mentioned in the [previous comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403#pullrequestreview-2950213963) is still present.

I forgot to address this one yesterday. Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4403/commits/de5e51fef9dc2fb70c975482e68651777839ad6c. The issue was that partial refunds don't change the order status, so the note was also ignored. But the emails were working for me already. Did you try using a JN website? They don't work locally.

Mayisha (approved):
Thanks for the fixes. LGTM :shipit:

PR #4406, opened 2025-06-11
Implementing composer autoloader without namespaces

wjrosa:
Fixes STRIPE-507
See peNR48-1sM-p2#comment-879

As part of the autoloader implementation, this PR changes the existing version to use Composer's classmap version instead. This should provide Composer with useful features such as the auto-generation of the classmap file and cache.

### Install

- Checkout to this branch on your local environment (`dev/implementing-psr-4-autoloader-without-namespaces`)
- Run `npm run build:deps`
- Confirm you get no fatal errors related to missing files upon accessing the admin interface or the checkout page as a shopper

### Build

- Run `npm run build`
- Confirm the extension .zip file is correctly generated, and it now includes the `vendor` folder

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
@daledupreez, will we drop the PSR-4 filenames or tackle it in another project? To know if I should revert the change in the unit test files.

woocommerce-gateway-stripe.php
```diff
@@ -81,10 +81,11 @@ function woocommerce_gateway_stripe() {
...
 
 	if ( ! isset( $plugin ) ) {
-		// Initialize our autoloader.
-		require_once __DIR__ . '/includes/class-wc-stripe-autoloader.php';
-		WC_Stripe_Autoloader::init();
-
+		// Includes the autoloader.
+		$autoload_filepath = __DIR__ . '/vendor/autoload.php';
+		if ( file_exists( $autoload_filepath ) ) {
+			require $autoload_filepath;
```
Mayisha, line 87: Do we need any fallback here?

Mayisha (commented):
Nice and clean 👍 

Quick questions: are we getting rid of the PSR-4 changes done in other PRs as well as the `WC_Stripe_Autoloader` class?

wjrosa:
> Nice and clean 👍
> 
> Quick questions: are we getting rid of the PSR-4 changes done in other PRs as well as the `WC_Stripe_Autoloader` class?

Thanks for the review, Mayisha! I asked @daledupreez about the PSR-4 changes [just above](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4406#issuecomment-2963965549). As for the `WC_Stripe_Autoloader`, yes. But it is better to deprecate it before.

woocommerce-gateway-stripe.php
wjrosa: Hum, maybe the `WC_Stripe_Autoloader` class?

Mayisha: If we are not removing the class then `WC_Stripe_Autoloader` can be the fallback. 
On second thought, a proper build/zip of the plugin should always contain the `autoload.php` in `vendor` directory. Maybe a fallback is not much needed.

wjrosa: I agree. But I added the fallback anyway just to make it safer in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4406/commits/4f1a49738f145179f1284d5d3563865c02bc6349

daledupreez (approved):
The changes LGTM, and I tested the code while we were at the meetup.

For the PSR-4 changes in the unit tests, I think it's worth discussing that on our next team call.

Mayisha (approved):
Looks good to me. 

✅ `vendor` directory contains necessary files after a build.
✅ I did a couple of quick smoke tests in the settings and checkout.

wjrosa:
Hey @daledupreez ! I am not sure what to do with this PR:
- Merge it to the base branch (your PR) so we can move forward when you decide to resume the autoloader work
- Close it and recreate once the work is resumed
- Keep it here as blocked and wait for the base branch to be merged

daledupreez:
@wjrosa, I vote we merge the PR into the base branch for now. We can work out what our plan should be from there, as I did have some concerns that I think we need to work through before we can make the official switch. (Those are in peNR48-1BY-p2 .)

wjrosa:
Sounds good 👍 . Thank you!

PR #4411, opened 2025-06-16
Restricting BNPL promotional banner to target version

wjrosa:
Fixes STRIPE-531
Base PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4328
See pcUrMV-b0q-p2#comment-32417

The banner promoting BNPLs will now go out with the 9.7.0 version. This PR checks the current extension version when deciding whether the banner will be shown.

## Testing instructions

Code review should be enough. You can play with lowering the current extension version on the backend (`includes/admin/class-wc-stripe-settings-controller.php#253`) and checking if the banner shows up by following the testing steps of [this PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4328).

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
+* Fix - Restricts the BNPLs promotional banner to only be displayed after version 9.7.0.
 * Add - Adds a new promotional banner to promote the BNPL payment methods (Klarna, Afterpay, and Affirm) on the settings page.
 * Fix - Adds an exception to be thrown when the order item quantity is zero, during the retrieval of level 3 data from an order.
 * Dev - Deprecates the WC_Stripe_Order class and removes its inclusion call.
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
-* Fix - Restricts the BNPLs promotional banner to only be displayed after version 9.7.0.
+* Fix - Restricts the BNPLs promotional banner to only be displayed after version 9.7.0
 * Add - Adds a new promotional banner to promote the BNPL payment methods (Klarna, Afterpay, and Affirm) on the settings page.
 * Fix - Adds an exception to be thrown when the order item quantity is zero, during the retrieval of level 3 data from an order.
 * Dev - Deprecates the WC_Stripe_Order class and removes its inclusion call.
--- a/client/settings/payment-settings/promotional-banner/get-promotional-banner-type.js
+++ b/client/settings/payment-settings/promotional-banner/get-promotional-banner-type.js
@@ -40,7 +40,8 @@ export const getPromotionalBannerType = (
 		isUpeEnabled &&
 		! hasBNPLEnabled &&
 		// eslint-disable-next-line camelcase
-		wc_stripe_settings_params.plugin_version >= '9.7.0'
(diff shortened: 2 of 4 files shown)


wjrosa:
There are other ways to do it, but I think that's the simplest one:
- Revert the main PR -> It would require another PR to re-implement the code again, having to solve many conflicts along the way, and block the other PR introducing unit tests just for all the banners
- Add a "flag" on the backend to control it -> Closer to this implementation, but it would require adding a new parameter to maintain
- Remove the condition to display the banner -> Requires us to create yet another PR to re-introduce the condition

client/settings/payment-settings/promotional-banner/get-promotional-banner-type.js
```diff
@@ -35,7 +36,12 @@ export const getPromotionalBannerType = (
 
 	if ( oauthConnected === false ) {
 		return RECONNECT_BANNER;
-	} else if ( isUpeEnabled && ! hasBNPLEnabled ) {
+	} else if (
+		isUpeEnabled &&
+		! hasBNPLEnabled &&
+		// eslint-disable-next-line camelcase
+		wc_stripe_settings_params.plugin_version >= '9.7.0'
```
annemirasol: Will `wc_stripe_settings_params` always be defined and an object?

annemirasol (approved):
Code LGTM. I have a non-blocking comment around whether it makes sense to use some conditional chaining.

Did a quick smoke test -- works as expected!

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
+* Fix - Restricts the BNPLs promotional banner to only be displayed after version 9.7.0.
```
daledupreez: Nit: we should remove the trailing `.` per our style guide

client/settings/payment-settings/promotional-banner/get-promotional-banner-type.js
daledupreez: I think we need to make this a bit more robust, as the `>=` comparison will break when we hit `10.0.0`, as that returns false for this comparison.

Maybe we can add the following checks to make extra-sure that we have a numeric major.minor that's greater than or equal to 9.7?
```suggestion
		wc_stripe_settings_params?.plugin_version &&
		parseFloat( String( wc_stripe_settings_params.plugin_version ).split( '.' ).slice( 0, 2 ).join( '.' ) ) >= 9.7
```

readme.txt
```diff
@@ -112,6 +112,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 
 = 9.6.0 - xxxx-xx-xx =
 
+* Fix - Restricts the BNPLs promotional banner to only be displayed after version 9.7.0.
```
daledupreez: Nit: another trailing `.`
```suggestion
* Fix - Restricts the BNPLs promotional banner to only be displayed after version 9.7.0
```

daledupreez (changes requested):
I think the logic is a bit buggy as written, so we need to fix that.

changelog.txt
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4411/commits/95ca3ee47a0743c8ed58deed0a2473f177f54313

daledupreez (approved):
Tested with the updated changes and things look good to me! 🚢

PR #4417, opened 2025-06-17
Allow ECE display when default customer location is set to geolocate

### LINKED ISSUE #4170: Re-enable ECE for virtual products using geolocate as default customer location

wjrosa:
Fixes STRIPE-378
Fixes #4170

This PR enables the display of the express checkout element buttons when taxes are enabled, the product is virtual, taxes are based on the billing address, but the default customer location is set to "Geolocate" or "Geolocate (with page caching support)".

Initially, it was disabled because we could not determine the customer's location with the setup above. But if location is geolocated, taxes should be accurate.

## Testing instructions

- Checkout and build this branch on your test environment (`tweak/allow-ece-display-when-default-customer-location-is-gelocated`)
- As a merchant, connect your Stripe account
- Go to `wp-admin/admin.php?page=wc-settings&tab=general` and set "Default customer location" to "Geolocate" or "Geolocate (with page caching support)"
- Check "Enable tax rates and calculations" and save
- Go to `wp-admin/admin.php?page=wc-settings&tab=tax` and set "Calculate tax based on" to "Customer billing address" and save
- Enable the express payment methods (Google Pay, Apple Pay, Link)
- As a shopper, access the store using a public-facing URL and add a virtual product to your cart
- Confirm the express methods are now shown:
Screenshot 2025-06-17 at 17 50 17

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


annemirasol (approved):
Works as described! 🚢 

> Initially, it was disabled because we could not determine the customer's location with the setup above. But if location is geolocated, taxes should be accurate.

Small nit on the PR description: Taxes won't always be accurate. For example, my credit card billing address is Texas, but I'm on a business trip in Florida and purchased something while there.

But after discussing this as a team, and given several merchant feedbacks, we decided to relax this restriction. Given there is also a filter available (`wc_stripe_should_hide_express_checkout_button_based_on_tax_setup`), any merchant who wishes to tweak the behavior can do so.

wjrosa:
> Works as described! 🚢
> 
> > Initially, it was disabled because we could not determine the customer's location with the setup above. But if location is geolocated, taxes should be accurate.
> 
> Small nit on the PR description: Taxes won't always be accurate. For example, my credit card billing address is Texas, but I'm on a business trip in Florida and purchased something while there.
> 
> But after discussing this as a team, and given several merchant feedbacks, we decided to relax this restriction. Given there is also a filter available (`wc_stripe_should_hide_express_checkout_button_based_on_tax_setup`), any merchant who wishes to tweak the behavior can do so.

Thanks for the review, Anne! And thanks for explaining it better. I may have understood it differently at first 👍

daledupreez:
I have some high-level feedback on this change: I have concerns that this change in behaviour doesn't have a configuration setting, and that it is being enabled by default, with the only mechanism to disable being a filter.

While some merchants have asked for this, pushing it on all merchants doesn't feel right, especially when we may be adding to their sales tax collection issues.

I think it would be ideal to have this controlled by a configuration setting, I think we need to flip this behaviour around and have it be the same as before by default, and we then make it possible to enable the new behaviour via a filter.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.6.0 - xxxx-xx-xx =
+* Tweak - Allow the display of express payment methods when default customer location is set to "Geolocate" or "Geolocate (with page caching support)".
```
daledupreez: Two concerns:
 * I think this should be flagged as an `Update`, not a `Tweak`
 * We should remove the trailing `.`
```suggestion
* Update - Allow the display of express checkout when default customer location is set to "Geolocate" or "Geolocate (with page caching support)"
```

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -719,11 +719,14 @@ private function should_hide_ece_based_on_tax_setup() {
...
+		$is_taxable               = $this->is_product_or_cart_taxable();
+		$needs_shipping           = $this->product_or_cart_needs_shipping();
+		$is_tax_based_on_billing  = 'billing' === get_option( 'woocommerce_tax_based_on' );
+		$default_customer_address = get_option( 'woocommerce_default_customer_address' );
+
+		if ( $is_taxable && $is_tax_based_on_billing && ! $needs_shipping
+			// Allow ECE to be displayed if the default customer address is set to `geolocation` or `geolocation_ajax`.
+			&& ! in_array( $default_customer_address, [ 'geolocation', 'geolocation_ajax' ], true ) ) {
 			return true;
 		}
```
daledupreez, line 731: I think we need a filter here rather than just defaulting on.

daledupreez (commented):
This is working as described, but I think we need the option off by default and available via a filter. Rather than iterating in this PR, I created https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4423 to implement my suggestions.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Since you added the filter in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4423, I am marking it as resolved.

daledupreez:
@wjrosa, do we want to close this PR as a result of the discussion in #4423? The proposal is in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4423#pullrequestreview-2946329679).

wjrosa:
> @wjrosa, do we want to close this PR as a result of the discussion in #4423? The proposal is in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4423#pullrequestreview-2946329679).

Yeah, let's do that 👍 . I will add a comment to the original issue

PR #4415, opened 2025-06-17
Use platform PMC Id constants for the Checkout rendering

diegocurbelo:
Fixes STRIPE-533

## Changes proposed in this Pull Request:

The parameter `paymentMethodConfigurationParentId` is only used for the Smart Checkout. Still, it's always present, which causes the checkout page to try fetching the PMC list even when the PMC flag is false.

This PR replaces the `get_parent_configuration_id` implementation with the constant for each mode (test or live), to avoid.

## Testing instructions

1. Disable settings sync (PMC)
    `wp option patch update woocommerce_stripe_settings pmc_enabled 'no'`
2. Checkout the `develop` branch
3. In the store, add a product to the cart and go to the checkout
4. Go to in the DevTools' Database Cache tab and check that the `wcstripe_cache_test_payment_method_configuration` entry was just updated (TTL == Expires):
    Screenshot 2025-06-17 at 17 25 45
5. Delete the entry, refresh the checkout, and check that a new cache entry was created.
6. Checkout this branch (`fix/533-use-pmc-constants-for-checkout-rendering`)
7. Wait a couple of minutes and refresh the checkout, check that the cache entry was not updated (TTL and Expires dont match)
    Screenshot 2025-06-17 at 17 29 25
8. Delete the `wcstripe_cache_test_payment_method_configuration` entry (again) and reload the checkout.
9. Check that the cache entry was not created.

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4419, opened 2025-06-18
Update payment method configuration tests after #4415

daledupreez:

This PR ensures the unit tests for payment method configurations are updated in response to #4415, as I saw unit tests failures in `release/9.5.3` stemming from those changes, as well as reproducing them when running a smaller set of unit tests locally.

## Testing instructions

Inspection plus the automated GitHub tests should be sufficient.

---

### Changelog entry

<details>

#### Comment 
This PR only updates unit tests.
</details>

**Post merge**

### DIFF
--- a/tests/phpunit/WC_Stripe_Payment_Method_Configurations_Test.php
+++ b/tests/phpunit/WC_Stripe_Payment_Method_Configurations_Test.php
@@ -17,12 +17,33 @@ class WC_Stripe_Payment_Method_Configurations_Test extends WC_Mock_Stripe_API_Un
 	 * @return void
 	 */
 	public function test_get_parent_configuration_id() {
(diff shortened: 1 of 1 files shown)


wjrosa (approved):
Thanks for adding these tests! It looks all good for me 👍

annemirasol (approved):
LGTM!

Side note: when the team has the bandwidth, we should probably move this to setup and teardown (or even somewhere higher up), as it seems to be a common need. That this test doesn't fail when run with other tests likely means a prior test was affecting it.

daledupreez:
> That this test doesn't fail when run with other tests likely means a prior test was affecting it.

Yes, the test was effectively relying on the `testmode` setting being set to `'yes'` before this change. So depending on the previous tests and what state they left things in, this test could pass or fail. 😬

PR #3563, opened 2024-10-30
Check if taxes are enabled in ECE tax compatibility logic

annemirasol:
## Changes proposed in this Pull Request:
In #3493, we hide express checkout if the cart/product and tax setup lead us to display and charge different amounts.

It hides ECE if: a) the product/cart is not shipped, and b) taxes are based on billing or shipping address. Because of how WC tax options are stored, this logic should actually have a third check: c) taxes are enabled

## Testing instructions
1. Go to WooCommerce > Settings > General (tab), check `Enable tax rates and calculations`, and save.
2. Go to WooCommerce > Settings > Tax (tab), set `Calculate tax based on` to `Customer billing address`, and save.
3. Go to the product page of a virtual product, and verify that express checkout is not displaybed.
4. Go back to WooCommerce > Settings > General (tab), **uncheck** `Enable tax rates and calculations`, and save.
5. Refresh the product page of the virtual product.
   - In develop: express checkout is still unavailable.
   - In this branch: express checkout is available.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [x] Added changelog entry **in both** `changelog.txt` and `readme.txt` (or does not apply)
-   [x] Tested on mobile (or does not apply)

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4420, opened 2025-06-18
Checking if taxes are enabled when verifying if ECE should be displayed

wjrosa:
Fixes STRIPE-538

This is a minor fix for an edge case when a store enabled taxes but disabled them afterwards. The value of `woocommerce_tax_based_on` would still exist, making the express checkout buttons hidden when falling into the tax-based logic for virtual products.

## Testing instructions

- Check out and build the `develop` branch on your test environment
- Connect your Stripe account
- Enable taxes for the store
- Set the taxes to be based on the customer's billing address
- Enable the express payment methods (Google Pay, Apple Pay, Link)
- Now, disable taxes
- As a shopper, add a virtual product to your cart (make sure the product is set to "taxable")
- The express checkout buttons are not displayed
- Checkout and build this branch instead (`fix/check-if-taxes-are-enabled-when-checking-for-ece-display`)
- Confirm the buttons are now displayed

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -719,11 +719,12 @@ private function should_hide_ece_based_on_tax_setup() {
 			return false;
 		}
 
+		$taxes_enabled           = 'yes' === get_option( 'woocommerce_calc_taxes', 'no' );
```
annemirasol: We should use `wc_tax_enabled()` instead ([code](https://woocommerce.github.io/code-reference/files/woocommerce-includes-wc-conditional-functions.html#function_wc_tax_enabled)).

annemirasol (changes requested):
I believe we fixed this before in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3563, and somehow lost it during refactors. Thanks for catching and fixing.

Added a suggestion to use the WooCommerce-provided function.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Good call! Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4420/commits/ba4cc7e0b5cea7928efda50df98098f6fa00758f

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -176,48 +193,63 @@ public function provide_test_hides_ece_if_cannot_compute_taxes() {
...
 				'is pay for order' => false,
+				'taxes enabled'    => false,
 				'tax based on'     => 'shipping',
 				'filter value'     => null,
 				'expected'         => $show,
 			],
 			'Do not hide if taxes are not based on customer billing or shipping address.' => [
 				'cart contents'    => [ 'virtual_taxable' ],
 				'is pay for order' => false,
+				'taxes enabled'    => false,
```
annemirasol: Same here -- set to `true`.

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -176,48 +193,63 @@ public function provide_test_hides_ece_if_cannot_compute_taxes() {
...
 				'tax based on'     => 'base',
 				'filter value'     => null,
 				'expected'         => $show,
 			],
 			'Do not hide if cart requires shipping.' => [
 				'cart contents'    => [ 'shippable_taxable' ],
 				'is pay for order' => false,
+				'taxes enabled'    => false,
```
annemirasol: Same here -- set to `true`.

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -176,48 +193,63 @@ public function provide_test_hides_ece_if_cannot_compute_taxes() {
...
 				'tax based on'     => 'billing',
 				'filter value'     => null,
 				'expected'         => $show,
 			],
 			'Do not hide if cart has virtual product and tax is based on shipping address.' => [
 				'cart contents'    => [ 'virtual_taxable', 'virtual_nontaxable' ],
 				'is pay for order' => false,
+				'taxes enabled'    => false,
```
annemirasol: Same here.

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -176,48 +193,63 @@ public function provide_test_hides_ece_if_cannot_compute_taxes() {
...
+				'is pay for order' => false,
+				'taxes enabled'    => false,
+				'tax based on'     => 'billing',
+				'filter value'     => null,
+				'expected'         => $show,
+			],
 			'Do not hide if cart has virtual product and tax is based on billing address, but filter forces to show.' => [
 				'cart contents'    => [ 'virtual_taxable', 'virtual_nontaxable' ],
 				'is pay for order' => false,
+				'taxes enabled'    => false,
```
annemirasol: I think we should set this to `true` for the test to accurately check for what it is describing.

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -176,48 +193,63 @@ public function provide_test_hides_ece_if_cannot_compute_taxes() {
...
 				'tax based on'     => 'billing',
 				'filter value'     => false,
 				'expected'         => $show,
 			],
 			'Do not hide if Pay for Order page.'     => [
 				'cart contents'    => [ 'virtual_taxable' ],
 				'is pay for order' => true,
+				'taxes enabled'    => false,
```
annemirasol: Same here.

annemirasol (changes requested):
I think we should set `$taxes_enabled` to `true` in the unit tests, except for the two cases that specifically test for the "taxes are disabled" scenario. That way, we know the behavior is due to the scenario it is testing, and not only because taxes are enabled/or disabled, e.g. ECE should always be available in the Pay for Order page.

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
wjrosa: Thanks, Anne! Updated all the test cases in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4420/commits/03cef44b0232bd72bf77c9bc327045e66968c539

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4420/commits/03cef44b0232bd72bf77c9bc327045e66968c539

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4420/commits/03cef44b0232bd72bf77c9bc327045e66968c539

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4420/commits/03cef44b0232bd72bf77c9bc327045e66968c539

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4420/commits/03cef44b0232bd72bf77c9bc327045e66968c539

annemirasol (approved):
LGTM, thanks for fixing! 

Just realized I was the one who removed the check when I added per-product tax logic 🙈

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -719,11 +719,12 @@ private function should_hide_ece_based_on_tax_setup() {
...
 
+		$taxes_enabled           = wc_tax_enabled();
 		$is_taxable              = $this->is_product_or_cart_taxable();
 		$needs_shipping          = $this->product_or_cart_needs_shipping();
 		$is_tax_based_on_billing = 'billing' === get_option( 'woocommerce_tax_based_on' );
 
-		if ( $is_taxable && $is_tax_based_on_billing && ! $needs_shipping ) {
+		if ( $taxes_enabled && $is_taxable && $is_tax_based_on_billing && ! $needs_shipping ) {
 			return true;
 		}
```
daledupreez: Not at all blocking, but it might be a worthwhile follow-up to make these checks inline rather than computing all the values and then making a decision, as it means we may end up doing unnecessary work for these cases.

I do realise this is a bit less reable,
```php
		// Hide express checkout when we have the following situation:
		//  - Taxes are enabled
		//  - The current product or cart is taxable
		//  - The product or cart does not need shipping (e.g. a virtual product)
		//  - Taxes are based on the user's billing address
		if (
			wc_tax_enabled() &&
			$this->is_product_or_cart_taxable() &&
			'billing' === get_option( 'woocommerce_tax_based_on' ) &&
			! $this->product_or_cart_needs_shipping()
		) {
```

daledupreez (approved):
Changes LGTM and test as described. Good catch on this.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Alright! Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4420/commits/eb049e618b272b1c34299848348e964109749c75

PR #4423, opened 2025-06-19
Iterate on enabling express checkout for virtual taxable products

daledupreez:
Relates to #4417

This PR implements some improvements/iterations that I proposed for the underlying work in #4417.

## Testing instructions

* Follow the testing for #4417, with the following exceptions:
  - By default, we should not enable express checkout without adding a filter
  - Add a filter to enable the new option, and verify that express checkout is shown in checkout. You can use code like the following:
```php
add_filter( 'wc_stripe_show_express_checkout_taxable_virtual_products_with_geolocation', '__return_true' );
```
 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


readme.txt
```diff
@@ -112,7 +112,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 
 = 9.6.0 - xxxx-xx-xx =
 
-* Tweak - Allow the display of express payment methods when default customer location is set to "Geolocate" or "Geolocate (with page caching support)".
+* Update - Allow the display of express checkout when default customer location is set to "Geolocate" or "Geolocate (with page caching support)"
```
annemirasol: Should we rephrase this as "Introduce a filter to allow displaying...", as we're no longer allowing it by default?

annemirasol (commented):
We already have the filter `wc_stripe_should_hide_express_checkout_button_based_on_tax_setup`. Stores with this setup (taxes based on billing + geolocate) that are okay with potential tax inaccuracies can just use that filter to return false (i.e. do not hide because of tax setup).

Maybe we should just keep things as is, and just close the issue #4170, as the filter already provides them a path forward?

(I also added a comment about the changelog -- if we're not changing the default behavior, we should just say we're adding a filter. If we're moving forward with this.)

daledupreez:
> We already have the filter wc_stripe_should_hide_express_checkout_button_based_on_tax_setup. Stores with this setup (taxes based on billing + geolocate) that are okay with potential tax inaccuracies can just use that filter to return false (i.e. do not hide because of tax setup).
>
> Maybe we should just keep things as is, and just close the issue https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4170, as the filter already provides them a path forward?

That would work for me. What do you think, @wjrosa?

wjrosa:
> > We already have the filter wc_stripe_should_hide_express_checkout_button_based_on_tax_setup. Stores with this setup (taxes based on billing + geolocate) that are okay with potential tax inaccuracies can just use that filter to return false (i.e. do not hide because of tax setup).
> > Maybe we should just keep things as is, and just close the issue #4170, as the filter already provides them a path forward?
> 
> That would work for me. What do you think, @wjrosa?

I agree with Anne 👍 (sorry for not being clearer, that's what I meant with my 👍 reaction)

daledupreez:
@annemirasol and @wjrosa, I took another look and it feels like the `wc_stripe_should_hide_express_checkout_button_based_on_tax_setup` filter is too broad -- it applies across the board and doesn't give any details as to what context we are in. That could be problematic for merchants who sell multiple types of product, while this new filter expressly makes it possible to allow express checkout for this specific use case.

annemirasol:
> @annemirasol and @wjrosa, I took another look and it feels like the `wc_stripe_should_hide_express_checkout_button_based_on_tax_setup` filter is too broad -- it applies across the board and doesn't give any details as to what context we are in. That could be problematic for merchants who sell multiple types of product, while this new filter expressly makes it possible to allow express checkout for this specific use case.

My line of thinking:
- We only hide express checkout for virtual products.
- The merchant already knows what their setup is for tax and default customer location. It is not dynamic.
- If the store's tax is based on `billing`, use `geolocate`, and `__return_true` for `wc_stripe_show_express_checkout_taxable_virtual_products_with_geolocation` (force display), it is effectively the same as `__return_false` for the broader filter. Both will just force display express checkout for virtual products.

daledupreez:
Thanks for clarifying, @annemirasol. That makes more sense to me. I am going to close this PR, as I don't think it really adds any value. If anything, the thing we need to be able to do is make this configurable via a setting rather than a filter. But that should happen via a separate PR anyway.

PR #4430, opened 2025-06-24
Fix page crash when klarna not supported in the country

Mayisha:
Instead of throwing an error when Klarna is not supported in the account country, return an empty array as a fallback for the supported currency list.

## Testing instructions
- Code review as the issue is difficult to constantly reproduce.

I encountered the issue with following steps-
- Connect to a Stripe account based in India (or any country that does not support Klarna)
- Make changes in the settings page.
- Connect to a Stripe account based in the US.
- At this step, if the account cache is outdated (still has data for the account based in India), the page crashes (empty settings page)


client/utils/use-payment-method-currencies.js
```diff
@@ -212,10 +212,7 @@ const getKlarnaCurrencies = () => {
 		return [ 'EUR', 'SEK', 'PLN', 'CHF', 'CZK', 'DKK', 'GBP', 'NOK' ];
 	}
 
-	// Throw an error if the country is not recognized.
-	throw new Error(
-		`Unable to determine Klarna currencies for: ${ accountCountry }`
-	);
+	return [];
```
wjrosa, line 221: We should consider adding a `console.error` (disabling the ' no-console` eslint rule) here for posterity. In case a merchant reaches out in the future. I know we should avoid it, but if a merchant falls into the default return, we will have no clue.

wjrosa (approved):
Changes look good! Left an optional (but appreciated) suggestion.

malithsen (approved):
I wasn't able to replicate the issue on `develop` but the change is sensible and straightforward. Let's merge this after removing the extra changelog entry.

readme.txt
```diff
@@ -150,5 +150,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Tweak - Remove Payment Method Configurations fallback cache
 * Tweak - Update deprecation notice message to specify that legacy checkout experience has been deprecated since version 9.6.0
 * Update - Remove legacy checkout checkbox from settings
+* Update - Improve checks in voucher purchase flow
```
malithsen:
```suggestion
```

Let's remove this irrelevant changelog entry

PR #4432, opened 2025-06-24
Adding a lock for payment processing requests

### LINKED ISSUE #4149: Woocommerce Checkout Doesn't Progress to Thank You Page after Successful Charge
**Describe the bug**
Yesterday (2025-03-25) we had 4 customers who experienced a problem with the checkout. (For reference we received 50-ish orders successfully that day). They click the 'Pay Now' button on the checkout but the page just stays as it is. However, Stripe processes the charge and returns a charge successful message to the webhook (confirmed in the plugin api log: `2025-03-25T12:47:28+00:00 Debug Webhook received: charge.succeeded`). 

**To Reproduce**
1. Enter payment details on checkout and click 'Pay Now'
2. The checkout does not progress
3. User re-attempts checkout payment
4. User is charged double

**Expected behavior**
When payment is confirmed via the API the order status should progress and the checkout page should progress to the Thank You / Order Confirmed page.

**Screenshots**
Image

 - WordPress Version: 6.7.2
 - WooCommerce Version: 9.7.1
 - Stripe Plugin Version: 9.3.1
 - Browser [e.g. chrome, safari] and Version: iPhone Safari, Mac Safari, PC / Desktop
 - Any other plugins installed: WC Order Status Manager (although the orders only go through default statuses i.e. 'pending-payment' and 'processing')
 - PHP: 8.3

**Additional context**
As I mentioned, this affected 4 customers, in slightly different ways. One customer paid twice but it generated two shipments (i.e. the process progressed on the server) but the customer didn't see the Order Confirmation screen the first time. Two customers experienced the same, but the shipment wasn't created (the server didn't process the order). The final customer attempted to pay 3 times but didn't see any Order Confirmation screen. 

The orders vary in payment method: Apple Pay, Google Pay, credit cards, debit cards.

In all these cases the Order Notes show a successful 'payment intent created', but only the ones that progressed also show a 'charge complete' note, however the API log on the server shows the webhook message was received (and there are no delivery errors in the webhook console in Stripe).

wjrosa:
Partially fixes STRIPE-371
Partially fixes #4149

As part of our efforts to avoid duplicate charges, I am introducing a new lock system for the payment processing requests (not specific to order objects). This new lock (saved as a database cache item as `wc_stripe_lock_process_payment_request_<request-identifier>`) is set when the processing of the payment starts, and it is removed at the end of the flow.

To build the `request-identifier` part, I am using MD5 on top of the customer ID, product ID, and quantities. To be able to have that information consistently, I had to move and abstract some code around.

## Testing instructions

- Checkout to this branch on your test environment (`fix/adding-a-lock-for-the-payment-processing-requests`)
- Connect your Stripe account
- Open a checkout page in 2 different tabs and fill in all the required information
- On both tabs, try to place the order at the same time (with less than a minute of difference)
- You should see the following error on your second tab:
Screenshot 2025-06-24 at 17 40 31

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-customer.php
```diff
@@ -20,7 +20,7 @@ class WC_Stripe_Customer {
 	 */
 	const STRIPE_PAYMENT_METHODS = [
 		WC_Stripe_UPE_Payment_Method_CC::STRIPE_ID,
-		WC_Stripe_UPE_Payment_Method_LINK::STRIPE_ID,
+		WC_Stripe_UPE_Payment_Method_Link::STRIPE_ID,
```
wjrosa, line 23: Unrelated fix.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1032,6 +1032,18 @@ private function process_payment_with_payment_method( int $order_id ) {
 
 			$this->validate_selected_payment_method_type( $payment_information, $order->get_billing_country() );
 
+			if ( $this->is_payment_process_request_locked( $order, $payment_information['customer'] ) ) {
+				// If the request is already being processed, return an error.
+				return [
+					'result'   => 'failure',
+					'redirect' => '',
+					'message'  => __( 'Your payment is already being processed. Please wait.', 'woocommerce-gateway-stripe' ),
```
wjrosa, line 1041: Open to suggestions on how to improve the error message 👀 .

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1175,6 +1187,7 @@ private function process_payment_with_payment_method( int $order_id ) {
 			}
 
 			$this->unlock_order_payment( $order );
+			$this->unlock_payment_process_request( $order );
```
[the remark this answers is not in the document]
wjrosa: We don't have a `$payment_information` variable with the customer key. The customer is then retrieved from the order meta (which is available at this point).

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1032,6 +1032,18 @@ private function process_payment_with_payment_method( int $order_id ) {
...
+				// If the request is already being processed, return an error.
+				return [
+					'result'   => 'failure',
+					'redirect' => '',
+					'message'  => __( 'Your payment is already being processed. Please wait.', 'woocommerce-gateway-stripe' ),
+				];
+			}
+
+			// Lock payment process request to prevent concurrent requests.
+			$this->lock_payment_process_request( $order, $payment_information['customer'] );
```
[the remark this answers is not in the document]
wjrosa: That's a valid point. But the lock needs the customer information, which does not exist before this try/catch block. Moving the lock after the try/catch block would significantly reduce its usefulness (since there's a lot of processing happening inside of it).

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -1462,45 +1462,8 @@ public function generate_create_intent_request( $order, $prepared_source ) {
...
-					'product_description' => $product_description, // Up to 26 characters long describing the product.
-					'unit_cost'           => $unit_cost, // Cost of the product, in cents, as a non-negative integer.
-					'quantity'            => $quantity, // The number of items of this type sold, as a non-negative integer.
-					'tax_amount'          => $tax_amount, // The amount of tax this item had added to it, in cents, as a non-negative integer.
-					'discount_amount'     => $discount_amount, // The amount an item was discounted—if there was a sale,for example, as a non-negative integer.
-				];
-			},
-			$order_items
-		);
+		$currency          = $order->get_currency();
```
malithsen: nit: This top-level variable seems unnecessary since it's now only used in a single place. I'd suggest just using `$order->get_currency()` within `shipping_amount`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1032,6 +1032,18 @@ private function process_payment_with_payment_method( int $order_id ) {
 
 			$this->validate_selected_payment_method_type( $payment_information, $order->get_billing_country() );
 
+			if ( $this->is_payment_process_request_locked( $order, $payment_information['customer'] ) ) {
+				// If the request is already being processed, return an error.
```
malithsen, line 1037: nit: It might be worth adding a log entry as well with `WC_Stripe_Logger::info`

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
malithsen: @wjrosa I think Copilot's suggestion is about releasing the lock if there's an exception within the order processing flow. If there's a failure, our unlock code will not be reached and the order will stay locked for a minute. We can improve it by adding a 

```
 } finally {
      // Always unlock regardless of success/failure
      $this->unlock_payment_process_request( $order );
  }
```

malithsen (commented):
Thanks for introducing these checks. Left some comments, but the approach looks good to me.

I initially had some trouble wrapping my head around these new lock and the existing `lock_order_payment`.

I see that they address different double-payment problems at different layers.
Existing `locker_order_payment` is to prevent webhook race conditions and relies on a meta field. Whereas this new lock mechanism is based on the order content to prevent duplicate request submissions.

annemirasol:
I have a high-level question.

Have we considered just reusing the existing order payment locking, i.e. `$this->lock_order_payment( $order )` (called in the same method, just before the call to `$this->process_payment_intent_for_order()`)? It looks like we can just move the lock earlier and add the bail-if-locked logic.

Right now, the lock prevents the webhook handler from double processing the order, but maybe it can also prevent itself from doing the same.

wjrosa:
> I have a high-level question.
> 
> Have we considered just reusing the existing order payment locking, i.e. `$this->lock_order_payment( $order )` (called in the same method, just before the call to `$this->process_payment_intent_for_order()`)? It looks like we can just move the lock earlier and add the bail-if-locked logic.
> 
> Right now, the lock prevents the webhook handler from double processing the order, but maybe it can also prevent itself from doing the same.

That's a good question! And it would be much better if we could reuse the existing lock. But I tried it here (code is [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4450/files)) and it does not work for some unknown reason. I always end up with the order being processed twice:
Screenshot 2025-06-30 at 17 39 47 😢

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Makes sense! Updated it here https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4432/commits/02fcc7d9efd6bcd1405a798dbe274040db72ae4c

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: I agree. Added it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4432/commits/174581b7b02af63856d0bdb34d2f18192bd97ea0

wjrosa:
> @wjrosa I think Copilot's suggestion is about releasing the lock if there's an exception within the order processing flow. If there's a failure, our unlock code will not be reached and the order will stay locked for a minute. We can improve it by adding a

Gotcha, @malithsen . Added the `finally` clause in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4432/commits/208337c43bf5722b348de316ca63857fc6cd2827

malithsen (approved):

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1175,6 +1188,7 @@ private function process_payment_with_payment_method( int $order_id ) {
 			}
 
 			$this->unlock_order_payment( $order );
+			$this->unlock_payment_process_request( $order );
```
annemirasol, line 1191: I believe this is redundant, since we already do this in the `finally` block.

annemirasol (commented):
I am getting a different error message. Is that expected?

Screenshot 2025-07-01 at 3 51 26 PM

annemirasol:
> But I tried it here (code is [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4450/files)) and it does not work for some unknown reason.

That's weird. It works well for me 🤷‍♀️ 

>  see that they address different double-payment problems at different layers.
Existing locker_order_payment is to prevent webhook race conditions and relies on a meta field. Whereas this new lock mechanism is based on the order content to prevent duplicate request submissions.

If we conceptually think of them as different things, then I'm okay with not reusing the existing locking mechanism.

wjrosa:
> > But I tried it here (code is [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4450/files)) and it does not work for some unknown reason.
> 
> That's weird. It works well for me 🤷‍♀️
> 
> > see that they address different double-payment problems at different layers.
> > Existing locker_order_payment is to prevent webhook race conditions and relies on a meta field. Whereas this new lock mechanism is based on the order content to prevent duplicate request submissions.
> 
> If we conceptually think of them as different things, then I'm okay with not reusing the existing locking mechanism.

Yeah, very weird. I removed my local environment entirely and tried again; it worked 😕 . So, I went ahead and created a new PR with only that version: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4462. I agree it is better to reuse the feature; otherwise, we may end up with too many different locks with the same purpose (or a very similar one).

PR #4434, opened 2025-06-24
Upload e2e setup logs when e2e setup fails

daledupreez:

I have have noticed some issues recently where the setup step for e2e tests fail, but it's not at all clear why. Most of the relevant output is pushed to our e2e setup log, and it would be really helpful if we had that available to understand why the setup is failing in GitHub.

## Testing instructions

This is a tricky one to test interactively, because it should only trigger when the e2e test setup step fails in GitHub. Inspection should be sufficient, as the step should do nothing otherwise, and if it breaks, I'll just revert.

---

### Changelog entry

<details>

#### Comment 
This PR only changes a GitHub workflow, and doesn't really touch any aspect of the code itself.
</details>

**Post merge**


.github/workflows/e2e-tests.yml
```diff
@@ -80,6 +80,15 @@ jobs:
...
         run: npm run test:e2e-setup
 
+      - name: Upload e2e setup logs
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: ${{ matrix.checkout }}-WP_latest-WC_latest-setup-logs
+          path: tests/e2e/e2e-setup.log
+          if-no-files-found: ignore
+          retention-days: 14
```
wjrosa: I would reduce the retention to ~4 days to avoid consuming unnecessary resources.

wjrosa (approved):
Change looks good! Left a non-blocking comment

.github/workflows/e2e-tests.yml
daledupreez: I think that's a good idea, but I've only dropped it to 7 days so we have time to investigate. Also, this will only upload a file when the setup steps/script fail, and that file is plaintext only. So I don't think it's a huge additional thing to include.

daledupreez:
I am going to get this shipped, as I am having e2e issues while testing a dependabot patch in #4428.

PR #4435, opened 2025-06-25
Throwing exception when saved method cannot be found

wjrosa:
See https://wordpress.org/support/topic/internal-server-error-with-woocommerce-store-api-after-disabling-legacy-checkout/

We got an edge case on [this thread](https://wordpress.org/support/topic/internal-server-error-with-woocommerce-store-api-after-disabling-legacy-checkout/) in which the merchant got a fatal error when processing a payment with a saved method:
```
Uncaught Error: Call to a member function get_token() on null in D:\projects\mydev\wp-content\plugins\woocommerce-gateway-stripe\includes\payment-methods\class-wc-stripe-upe-payment-gateway.php on line 1296
```

Although it is hard to reproduce (and we didn't hear back from the merchant in the thread), this PR makes the code safer by throwing an exception with a specific error message.

## Testing instructions

Code review should be enough. If you prefer, you can change the code a bit (like resetting the value of $_POST) to force the exception when paying using a saved method in the new checkout experience.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -1303,7 +1303,13 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Gateway_Stripe {
 
 			$token = WC_Stripe_Payment_Tokens::get_token_from_request( $_POST );
 			if ( ! $token ) {
-				throw new WC_Stripe_Exception( __( "We're not able to process this payment. The saved payment method could not be found.", 'woocommerce-gateway-stripe' ) );
+				throw new WC_Stripe_Exception(
+					sprintf(
+						/* translators: %s is the order ID */
+						__( "We're not able to process this payment. The saved payment method for order %s could not be found.", 'woocommerce-gateway-stripe' ),
+						$order_id
+					)
+				);
 			}
 
 			$payment_method          = $this->stripe_request( 'payment_methods/' . $token->get_token(), [], null, 'GET' );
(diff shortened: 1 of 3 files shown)


diegocurbelo (approved):
Thanks for fixing this @wjrosa!

**Before:**
Screenshot 2025-06-26 at 15 28 00

**After:**
Screenshot 2025-06-26 at 15 28 17

_Exception message in the log:_

Screenshot 2025-06-26 at 15 29 27

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1301,7 +1301,11 @@ public function process_payment_with_saved_payment_method( $order_id, $can_retry
 				return $this->process_pre_order( $order_id );
 			}
 
-			$token                   = WC_Stripe_Payment_Tokens::get_token_from_request( $_POST );
+			$token = WC_Stripe_Payment_Tokens::get_token_from_request( $_POST );
+			if ( ! $token ) {
+				throw new WC_Stripe_Exception( __( "We're not able to process this payment. The saved payment method could not be found.", 'woocommerce-gateway-stripe' ) );
```
annemirasol: Should we include at least the order id in the message in case the merchant wants to investigate?

annemirasol (approved):
From the forum thread, I think they were manually constructing the payload to Store API, and not setting the `wc-stripe-is-deferred-intent` flag. I was wondering why it was hitting that code path if UPE is enabled.

Anyway, LGTM -- doesn't hurt to be defensive and check that the token exists.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: That's a good idea. Added it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4435/commits/5e5ea221bb80712b8a20d4cbb7d2e8842c564dff

PR #4436, opened 2025-06-25
E2E: Test the basic Affirm payment flow

malithsen:
Fixes STRIPE-530

This adds an e2e test to cover the basic purchase flow with Affirm. Affirm has rate-limiting turned on even for sandbox accounts, and the KYC + payment flow on Affirm's end is quite lengthy. So I've opted to only check if we redirect to Affirm after clicking the "Place Order" button instead of introducing flaky tests that try to automate the entire flow.

## Testing instructions
1. Make sure e2e tests on GH pass.
2. Optionally, run e2es locally with `npm run test:e2e -- affirm`  and make sure they pass.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/tests/e2e/tests/checkout/blocks/affirm.spec.js
+++ b/tests/e2e/tests/checkout/blocks/affirm.spec.js
@@ -1,0 +1,24 @@
+import { test, expect } from '@playwright/test';
+import { admin, payments } from '../../../utils';
+
+const { setupAffirmCheckout } = payments;
+
+test.describe( 'Affirm payment tests @blocks', () => {
+	test.beforeAll( async ( { browser } ) => {
+		await test.step( 'Setup test environment', async () => {
+			// Enable Klarna in admin
+			await admin.togglePaymentMethod( browser, 'Affirm', true );
+		} );
+	} );
+
+	test.describe.configure( { mode: 'parallel' } );
+
+	test( 'customer can pay with Affirm @smoke', async ( { page } ) => {
+		await setupAffirmCheckout( page, 'blocks' );
+		await page.locator( 'text=Place order' ).click();
+		// Since we don't have control over the Affirm payment flow,
+		// verifying the redirect is all we can do consistently without introducing a
+		// flaky test.
+		await expect( page ).toHaveURL( /.*affirm\.com/ );
+	} );
+} );
--- a/tests/e2e/tests/checkout/blocks/affirm.spec.js
+++ b/tests/e2e/tests/checkout/blocks/affirm.spec.js
@@ -6,7 +6,7 @@ const { setupAffirmCheckout } = payments;
 test.describe( 'Affirm payment tests @blocks', () => {
 	test.beforeAll( async ( { browser } ) => {
 		await test.step( 'Setup test environment', async () => {
-			// Enable Klarna in admin
+			// Enable Affirm in admin
 			await admin.togglePaymentMethod( browser, 'Affirm', true );
 		} );
 	} );
--- a/tests/e2e/tests/checkout/blocks/affirm.spec.js
+++ b/tests/e2e/tests/checkout/blocks/affirm.spec.js
@@ -19,6 +19,6 @@ test.describe( 'Affirm payment tests @blocks', () => {
 		// Since we don't have control over the Affirm payment flow,
 		// verifying the redirect is all we can do consistently without introducing a
 		// flaky test.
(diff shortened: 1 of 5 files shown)


tests/e2e/tests/checkout/blocks/affirm.spec.js
```diff
@@ -0,0 +1,24 @@
+import { test, expect } from '@playwright/test';
+import { admin, payments } from '../../../utils';
+
+const { setupAffirmCheckout } = payments;
+
+test.describe( 'Affirm payment tests @blocks', () => {
+	test.beforeAll( async ( { browser } ) => {
+		await test.step( 'Setup test environment', async () => {
+			// Enable Klarna in admin
```
diegocurbelo:
```suggestion
			// Enable Affirm in admin
```

diegocurbelo (approved):
The code looks good, and it tests well (sometimes the affirm site takes longer than 2 secs to load, but I think that adding waitTimeouts would be worse 🤷🏼)

> Affirm has rate-limiting turned on even for sandbox accounts, and the KYC + payment flow on Affirm's end is quite lengthy. So I've opted to only check if we redirect to Affirm after clicking the "Place Order" button instead of introducing flaky tests that try to automate the entire flow.

This sounds reasonable to me... the rate-limiting makes it hard to test that the loaded Affirm page is not an error page (by checking the title, a header, or a link/button, etc).
And, we are using the redirect_url from the Stripe response, so even if we add the check, we would be testing the Stripe API behavior and not the plugin code 🤷🏼

tests/e2e/tests/checkout/shortcode/affirm.spec.js
```diff
@@ -0,0 +1,24 @@
...
+
+	test.describe.configure( { mode: 'parallel' } );
+
+	test( 'customer can pay with Affirm @smoke', async ( { page } ) => {
+		await setupAffirmCheckout( page, 'shortcode' );
+		await page.locator( 'text=Place order' ).click();
+		// Since we don't have control over the Affirm payment flow,
+		// verifying the redirect is all we can do consistently without introducing a
+		// flaky test.
+		await expect( page ).toHaveURL( /.*affirm\.com/ );
```
annemirasol: I'm running it locally, and the tests are failing because of 

```
Expected pattern: /.*affirm\.com/
Received string:  "https://payments.stripe.com/payment_methods/test_payment?frameId=frame-1&origin=%2A&payment_attempt=payatt_3ReKR1E1kLJpo4bM1JaSsovD"
```

Screenshot 2025-06-26 at 1 28 11 PM

Can we also check for that?

annemirasol (changes requested):

tests/e2e/tests/checkout/shortcode/affirm.spec.js
annemirasol: It looks like we get that Stripe page when using a test-only Connect account. When I changed my e2e keys to use a standard account, it gives me the affirm.com sandbox page.

malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4436/commits/c45ceea3ffe7af7ab78dfd1d56a2429e6f121240

I am not certain what makes Stripe choose between `stripe.com` and `affirm.com` for the off-site confirmation. I always seem to get `sandbox.affirm.com` and that seems to be the case on GH action as well but checking for both is reasonable.

malithsen: > It looks like we get that Stripe page when using a test-only Connect account.

Ah nice find

annemirasol: > Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/commit/c45ceea3ffe7af7ab78dfd1d56a2429e6f121240

I think the above only updated the `blocks` test, and not `shortcode`.

malithsen: 🤦 
Thanks! Updated shortcode and tweaked the comment in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4436/commits/f6f1e13ded2df1897ba79df3cbd209aa0e4d5556

annemirasol (approved):
LGTM, my local runs are green! 🚢

PR #4439, opened 2025-06-26
Prevent irrelevant customer update requests during checkout

Mayisha:
Fixes STRIPE-558

We do not save the same/duplicate card twice in WooCommerce token now, but it is possible to have duplicate cards attached to a customer in Stripe. When a customer has duplicate cards attached in Stripe and this customer places an order, multiple irrelevant POST requests to `customer/cus_id` are sent to Stripe for updating the default payment method.

This happens as 'woocommerce_payment_token_set_default' action gets triggered every time the token is updated and saved. The callback of this action sends a customer update request to Stripe 7-8 times due to a recursive loop. [Here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/payment-tokens/class-wc-stripe-payment-tokens.php#L563) we are updating the payment method of the duplicate token that triggers this loop.

- When a duplicate token is found, before updating the payment method ID of the token, we are now checking if the current payment method ID is also valid and exists in Stripe. If yes, we won't update the payment method ID of the token.

## Testing instructions
- Enable the new checkout experience.
- As a shopper, place an order to the store and create an account.
- Go to **My Account > Payment methods** page as the shopper.
- Add a card (`5555555555554444`)
- Add another card (`4242424242424242`)
- Add the first card again (`5555555555554444`)
- Notice that, in the Payment methods page you still have 2 cards and the duplicate card is not added.
- Go to the Stripe dashboard and find the corresponding customer (https://dashboard.stripe.com/test/customers/cus_id). Notice that there are 3 cards attached to the customer.
- Now as the shopper, add a product to your cart and go to the checkout page. Place an order.
- In `develop`
    - Check the logs in **Woocommerce > Status > Logs** and notice that there are a few customers/cus_id requests.
    - Check the logs in Stripe and observe the same.

**Multiple requests**
Screenshot 2025-06-26 at 1 13 44 PM

**Request data**
Screenshot 2025-06-26 at 1 14 06 PM

- In this branch, confirm that there is no update customer request sent to update the default payment method.
- Regression test
    - Changing the default payment method on **My Account > Payment methods** should work, changes should persist on page refresh and reflect in the Stripe dashboard.
    - As a new shopper, if you create an account during checkout and save the card, this very first saved card should be set as default in **My Account > Payment methods**.
    - As a new shopper, if you purchase a subscription, the card used for this purchase should be set as default **My Account > Payment methods**.


includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -338,7 +338,13 @@ public function woocommerce_get_customer_upe_payment_tokens( $tokens, $user_id,
 				}
 			}
 
-			$payment_methods = array_merge( ...$payment_methods );
+			$payment_methods     = array_merge( ...$payment_methods );
+			$payment_method_ids = array_map(
```
diegocurbelo:
```suggestion
			$payment_methods    = array_merge( ...$payment_methods );
			$payment_method_ids = array_map(
```

diegocurbelo (approved):
Nice job investigating and fixing this @Mayisha!

The code looks good, and it tests well! 🚢 

Screenshot_2025-06-26_at_16_24_55

**Regression test**
- ✅  Changing the default payment method on My Account > Payment methods should work, changes should persist on page refresh and reflect in the Stripe dashboard.
- ✅  As a new shopper, if you create an account during checkout and save the card, this very first saved card should be set as default in My Account > Payment methods.
- ✅  As a new shopper, if you purchase a subscription, the card used for this purchase should be set as default My Account > Payment methods.

Screenshot 2025-06-26 at 16 31 50

Screenshot 2025-06-26 at 16 31 25
Screenshot 2025-06-26 at 16 31 36

wjrosa (approved):
Thanks for improving this flow! Looks good to me 👍

PR #4438, opened 2025-06-25
Checking for other BNPL plugins when showing the promotional banner

wjrosa:
Fixes STRIPE-556
Base PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4328

## Changes proposed in this Pull Request:

We should not display the [BNPL promotional banner](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4328) when other BNPL payment method extensions are installed in the store. This PR adds that check to the existing banner condition. It checks whether Affirm or Klarna official extensions are installed and hides the banner.

The check is done using the `is_plugin_active` function ([docs](https://developer.wordpress.org/reference/functions/is_plugin_active/)).

## Testing instructions

- Checkout to this branch on your test environment (`fix/checking-for-other-bnpl-plugins-when-showing-the-promotional-banner`)
- Change the required version for the banners to 9.5 in `client/settings/payment-settings/constants.js#6`
- Build the frontend files
- Connect your Stripe account
- Make sure the legacy checkout experience and BNPLs (Klarna, Affirm, and Afterpay/Clearpay) are disabled
- Go to any Stripe settings page `wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=methods`
- Confirm the BNPL banner is now shown
Screenshot 2025-06-25 at 19 11 41

- Install [Affirm](https://woocommerce.com/products/woocommerce-gateway-affirm/) or [Klarna](https://woocommerce.com/products/klarna-payments/) extensions (only one is enough)
- Confirm the banner is now gone

PS: If you install the extensions using the private repositories, ensure the zip filename matches the extension name (i.e., `woocommerce-gateway-affirm`).

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4441, opened 2025-06-26
Adding a promotional note for BNPLs

wjrosa:
Fixes STRIPE-548
Figma 0fiyiUe18lGU6kGyQNmXxE?node-id=871-22237&t=qYzLLyDitMTNKLt0-0-fi

This PR introduces a new note to be displayed on the WooCommerce admin home, promoting BNPLs (Klarna and Affirm). The note is added when the Stripe extension is enabled, version is above 9.7, and BNPLs are not enabled (within the extension itself or by using the other official extensions).

The logic to check for other extensions was suggested by @diegocurbelo [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4438#discussion_r2169622041). I will make the block shareable on the last PR to land.

Screenshot 2025-06-26 at 17 55 57

## Testing instructions

- Checkout to this branch on your test environment (`add/bnpl-promotional-note`)
- Connect your Stripe account
- Hardcode the minimum required extension version to `9.5` in `/includes/notes/class-wc-stripe-bnpl-promotion-note.php#85`
- Make sure BNPLs are not enabled and the other official extensions are not installed
- Confirm the note is added and is correctly displayed on your inbox (`wp-admin/admin.php?page=wc-admin`)
- Check if it matches the Figma design contents
- Confirm that by clicking "Learn more" you are sent to `https://woocommerce.com/document/stripe/setup-and-configuration/additional-payment-methods/`
- Confirm that by clicking "Dismiss" the note is gone
- You can make additional tests for the conditions by deleting the note in the `wp_wc_admin_notes ` database table and enabling BNPLs and/or the other official extensions (Affirm, Klarna) to confirm the notes are not added

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/notes/class-wc-stripe-bnpl-promotion-note.php
+++ b/includes/notes/class-wc-stripe-bnpl-promotion-note.php
@@ -95,8 +95,9 @@ class WC_Stripe_BNPL_Promotion_Note {
 
 		$has_other_bnpl_plugins_active = false;
 		$available_payment_gateways    = WC()->payment_gateways->payment_gateways;
+		$other_bnpl_gateway_ids        = [ 'affirm', 'klarna_payments' ];
 		foreach ( $available_payment_gateways as $available_payment_gateway ) {
-			if ( ( 'affirm' === $available_payment_gateway->id || 'klarna_payments' === $available_payment_gateway->id ) && 'yes' === $available_payment_gateway->enabled ) {
+			if ( in_array( $available_payment_gateway->id, $other_bnpl_gateway_ids, true ) && 'yes' === $available_payment_gateway->enabled ) {
 				$has_other_bnpl_plugins_active = true;
 				break;
 			}
--- a/includes/notes/class-wc-stripe-bnpl-promotion-note.php
+++ b/includes/notes/class-wc-stripe-bnpl-promotion-note.php
@@ -95,8 +95,8 @@ class WC_Stripe_BNPL_Promotion_Note {
 
 		$has_other_bnpl_plugins_active = false;
 		$available_payment_gateways    = WC()->payment_gateways->payment_gateways;
-		foreach ( $available_payment_gateways as $gateway ) {
-			if ( ( 'affirm' === $gateway->id || 'klarna_payments' === $gateway->id ) && 'yes' === $gateway->enabled ) {
+		foreach ( $available_payment_gateways as $available_payment_gateway ) {
+			if ( ( 'affirm' === $available_payment_gateway->id || 'klarna_payments' === $available_payment_gateway->id ) && 'yes' === $available_payment_gateway->enabled ) {
 				$has_other_bnpl_plugins_active = true;
 				break;
 			}
--- a/includes/notes/class-wc-stripe-bnpl-promotion-note.php
+++ b/includes/notes/class-wc-stripe-bnpl-promotion-note.php
@@ -86,7 +86,7 @@ class WC_Stripe_BNPL_Promotion_Note {
 			return;
 		}
 
-		$available_upe_payment_methods = $gateway->get_upe_available_payment_methods();
+		$available_upe_payment_methods = $gateway->get_upe_enabled_payment_method_ids();
 		foreach ( WC_Stripe_Payment_Methods::BNPL_PAYMENT_METHODS as $bnpl_payment_method ) {
 			if ( in_array( $bnpl_payment_method, $available_upe_payment_methods, true ) ) {
 				return;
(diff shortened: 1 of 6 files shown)


includes/notes/class-wc-stripe-bnpl-promotion-note.php
```diff
@@ -0,0 +1,116 @@
...
+		$available_upe_payment_methods = $gateway->get_upe_available_payment_methods();
+		foreach ( WC_Stripe_Payment_Methods::BNPL_PAYMENT_METHODS as $bnpl_payment_method ) {
+			if ( in_array( $bnpl_payment_method, $available_upe_payment_methods, true ) ) {
+				return;
+			}
+		}
+
+		$has_other_bnpl_plugins_active = false;
+		$available_payment_gateways    = WC()->payment_gateways->payment_gateways;
+		foreach ( $available_payment_gateways as $gateway ) {
```
malithsen: nit: This loop reassigns the `$gateway` variable passed into the `init` function. I'd suggest renaming it to something else, like `$payment_gateway`.

includes/notes/class-wc-stripe-bnpl-promotion-note.php
```diff
@@ -0,0 +1,116 @@
...
+			if ( ( 'affirm' === $gateway->id || 'klarna_payments' === $gateway->id ) && 'yes' === $gateway->enabled ) {
```
malithsen: Should we also be checking for Afterpay? I see that the copy doesn't mention Afterpay but the changelog does.

nit: Extracting these IDs out to a separate variable (eg: `$OTHER_BNPL_GATEWAY_IDS`) might make it more readable.

includes/notes/class-wc-stripe-bnpl-promotion-note.php
```diff
@@ -0,0 +1,116 @@
...
+		 * - Below version 9.7
+		 * - Store has any BNPLs enabled
+		 * - Other BNPL extensions are active
+		 * - Stripe is not enabled
+		 */
+		if ( ! defined( 'WC_STRIPE_VERSION' ) || version_compare( WC_STRIPE_VERSION, '9.7', '<' ) ) {
+			return;
+		}
+
+		$available_upe_payment_methods = $gateway->get_upe_available_payment_methods();
```
diegocurbelo:
```suggestion
		$available_upe_payment_methods = $gateway->get_upe_enabled_payment_method_ids();
```

diegocurbelo (changes requested):
I think we need to check for `get_upe_enabled_payment_method_ids()` instead of `get_upe_available_payment_methods()`... with that change, it tests well. We should also add a test for the `init(...)` method

includes/notes/class-wc-stripe-bnpl-promotion-note.php
wjrosa: Good catch! Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4441/commits/d40a7da19c83694039aedae1d5441c40aaf916a8

includes/notes/class-wc-stripe-bnpl-promotion-note.php
wjrosa: > Should we also be checking for Afterpay? I see that the copy doesn't mention Afterpay but the changelog does.

We are not promoting Afterpay currently. So, I fixed the changelog entry in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4441/commits/3440d45fe1ba1a32b1d829854791410af581535f

> nit: Extracting these IDs out to a separate variable (eg: $OTHER_BNPL_GATEWAY_IDS) might make it more readable.

Good idea! Applied it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4441/commits/36371272ef95e3e837c4bdc3b6dee1d4629b7c89

wjrosa:
> I think we need to check for `get_upe_enabled_payment_method_ids()` instead of `get_upe_available_payment_methods()`... with that change, it tests well. We should also add a test for the `init(...)` method

Thanks, Diego! I have fixed the code by using the correct method you suggested.

As for the tests, I could not adequately test the `init` method:
- I cannot dynamically set the value of `WC_STRIPE_VERSION`. And it is already defined when the tests start (with the value of `9.5.3`). So the tests never pass the first condition. I even tried to use https://docs.phpunit.de/en/10.5/configuration.html#the-const-element
- I cannot make assertions on top of `possibly_add_note`, since it is a static method. That would be our "success" definition

diegocurbelo (approved):
That's fair, @wjrosa. We don't expect to change this code frequently; refactoring it to test that single method might not be worth the effort (and extra code).

malithsen (approved):

PR #4443, opened 2025-06-27
Show payment methods sync status on the UI

Mayisha:
- Added `is_pmc_enabled` flag in the settings response.
    - This adds a quick way for a user to know whether the PMC is enabled for this site on the frontend instead of checking the DB. 
    - May help with triaging missing payment method related reports as we will know whether to check the PMC or the capabilities for this account.
- Hiding the footer in the payment methods list when PMC is enabled.
    -  As PMC is enabled, settings between the store and the Stripe dashboard is synced. So there is no need to send the user to the Stripe dashboard to enable the capability of other payment methods.
- Added PMC or Sync status on the account configuration modal.
- Added PMC or Sync status in Stripe system status.

## Testing instructions
- Go to the Stripe settings page and check the network tab for `<your_site>/wp-json/wc/v3/wc_stripe/settings` request. You should see `is_pmc_enabled` in the response.
- If you have PMC enabled, the footer in the payment method list containing more payment methods` button should be hidden.

Screenshot 2025-06-27 at 10 38 21 AM

- Go to the Settings tab in the Stripe settings page and click `Configure connection` button. You should see the `SYNC` status.

Screenshot 2025-07-04 at 11 35 40 PM

- Go to **WooCommerce > Status > System status** and check the `WooCommerce Stripe Payment Gateway` section. You should see the sync status there as well.

Screenshot 2025-07-04 at 11 11 04 PM


Mayisha:
> Hiding the footer in the payment methods list when PMC is enabled.
     - As PMC is enabled, settings between the store and the Stripe dashboard are synced. So there is no need to send the user to the Stripe dashboard to enable the capability of other payment methods.
  
@diegocurbelo from your expertise on PMC, does this change make sense? Is my assumption correct that user won't need to go to the Stripe dashboard for additional payment methods? 👀

diegocurbelo:
> > Hiding the footer in the payment methods list when PMC is enabled.
> > - As PMC is enabled, settings between the store and the Stripe dashboard are synced. So there is no need to send the user to the Stripe dashboard to enable the capability of other payment methods.
> 
> @diegocurbelo from your expertise on PMC, does this change make sense? Is my assumption correct that user won't need to go to the Stripe dashboard for additional payment methods? 👀

I agree we should remove the footer... 

We can add the PMC sync status to the connection settings, something like `Sync: Enabled`
Screenshot_2025-07-02_at_11_56_22

And @malithsen suggested to also add it to the status page:
Screenshot 2025-07-02 at 12 02 12

annemirasol (changes requested):
LGTM overall!

It looks like the JS tests for `GeneralSettingsSection` is broken. Looks like it needs the new hook to be mocked, and its tests updated now that we will be hiding a section for PMC enabled.

Mayisha:
I have added the sync status on account configuration modal and the system status.

This is ready for review now. cc: @malithsen @annemirasol

includes/class-wc-stripe-status.php
```diff
@@ -113,6 +113,19 @@ public function render_status_report_section() {
 					</mark>
 				</td>
 			</tr>
+			<tr>
+				<td data-export-label="Sync Enabled"><?php esc_html_e( 'Sync Enabled', 'woocommerce-gateway-stripe' ); ?>:</td>
+				<td class="help"><?php echo wc_help_tip( esc_html__( 'Whether the payment methods are synced between Stripe dashboard and the plugin.', 'woocommerce-gateway-stripe' ) ); /* phpcs:ignore WordPress.XSS.EscapeOutput.OutputNotEscaped, WordPress.Security.EscapeOutput.OutputNotEscaped */ ?></td>
+				<td>
+					<?php
+					$is_pmc_enabled = $this->gateway->get_option( 'pmc_enabled', 'no' );
+					$class          = $is_pmc_enabled ? 'yes' : 'no';
```
malithsen: Even if `$is_pmc_enabled` is `no`, it will be truthy and `$class` will be set to `yes`.

We should check for the value of `$is_pmc_enabled` instead.

client/settings/general-settings-section/index.js
```diff
@@ -99,7 +100,7 @@ const GeneralSettingsSection = ( {
 							onCancel={ () => onChangeDisplayOrder( false ) }
 						/>
 					</AccountRefreshingOverlay>
-					{ isUpeEnabled && <SectionFooter /> }
+					{ isUpeEnabled && isPMCEnabled && <SectionFooter /> }
```
malithsen:
```suggestion
					{ isUpeEnabled && ! isPMCEnabled && <SectionFooter /> }
```

Shouldn't we hide the section when `isPMCEnabled` is `True`?

malithsen (changes requested):
Changes look good so far. Commented on a couple of observations.

client/settings/general-settings-section/index.js
Mayisha: Ahh such stupid mistake 🙄 
Thanks Malith.

malithsen (approved):
LGTM. Nice work :shipit:

annemirasol (approved):
LGTM!

Tested:
✅ When PMC enabled, system status report, Configure Connection modal, and `wc/v3/wc_stripe/settings` response are correct; extra section is not shown
✅ When PMC is not enabled, system status report, Configure Connection modal, and `wc/v3/wc_stripe/settings` response are correct; extra section is shown

PR #4446, opened 2025-06-29
Fix Express Checkout buttons not appearing on order-pay pages due to script timing issue

elvismdev:

This PR fixes Express Checkout buttons (Apple Pay, Google Pay, etc.) not appearing on WooCommerce order-pay pages due to a script timing issue where the `wcStripeExpressCheckoutPayForOrderParams` JavaScript variable was undefined.

### Why is this change needed?
Express Checkout buttons were completely missing from order-pay pages (`/checkout/order-pay/{order_id}/`), preventing customers from using Apple Pay, Google Pay, and other express payment methods to complete pending orders.

### What does this change do?
The fix ensures the `wc_stripe_express_checkout` script is registered before attempting to localize it with order data. The root cause was that `wp_localize_script()` was being called on an unregistered script (which fails silently), causing the required JavaScript variable to be missing from the page.

### Other solutions considered:
- Moving the hook to a later priority - rejected because it could interfere with other plugins
- Using a different hook entirely - rejected to maintain consistency with existing architecture
- Adding the localization to the main `scripts()` method - rejected because it would add order-specific logic to a general method

### Trade-offs:
The solution adds a small amount of code duplication for script registration logic, but this is necessary to ensure proper timing while maintaining clean separation of concerns.

### How can this code break?
- If WordPress changes the behavior of `wp_script_is()` or `wp_localize_script()`
- If the asset file path or structure changes
- If the script handle name changes

- Extracted script registration into a reusable private method
- Added defensive checks using `wp_script_is()` before registration
- Maintained all existing functionality and error handling
- Used the same asset loading logic as the existing `scripts()` method

### Prerequisites:
1. Have WooCommerce and Stripe Gateway plugin installed
2. Configure Stripe with test keys
3. Enable Express Checkout in Stripe settings
4. Ensure Apple Pay/Google Pay is configured in test mode

### Test Steps:
1. Create a test product and add it to cart
2. Go to checkout and place an order selecting "Direct bank transfer" or another non-Stripe payment method
3. This creates an order with "Pending Payment" status
4. Navigate to the order-pay page: `/checkout/order-pay/{order_id}/?pay_for_order=true&key={order_key}`
5. **Expected**: Express Checkout buttons (Apple Pay, Google Pay) should be visible on the page
6. Open browser developer console and verify `wcStripeExpressCheckoutPayForOrderParams` is defined and contains order data
7. Test that clicking the Express Checkout buttons initiates the payment flow correctly

#### Before this fix:
- Express Checkout buttons were missing from order-pay pages
- `wcStripeExpressCheckoutPayForOrderParams` was undefined in console
- Only traditional payment forms were available

#### After this fix:
- Express Checkout buttons appear on order-pay pages
- `wcStripeExpressCheckoutPayForOrderParams` contains proper order data
- All express payment methods work correctly

#### Additional verification:
- Verify Express Checkout still works on product pages, cart, and checkout
- Test with both logged-in and guest users
- Confirm no JavaScript errors in console

---

### Changelog entry

<details>

#### Comment 

This is a bug fix that restores expected functionality (Express Checkout buttons on order-pay pages) that should have been working. It doesn't introduce new features or change existing behavior visible to end users - it simply fixes a technical timing issue that prevented the buttons from appearing.

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -585,6 +580,34 @@ public function scripts() {
...
+		}
+
+		// Register the script if not already registered
+		if ( ! wp_script_is( 'wc_stripe_express_checkout', 'registered' ) ) {
+			$this->register_express_checkout_script();
+		}
+
+		$asset_path = WC_STRIPE_PLUGIN_PATH . '/build/express-checkout.asset.php';
+		$version    = WC_STRIPE_VERSION;
+		if ( file_exists( $asset_path ) ) {
```
wjrosa: We could think of creating a new private method to handle this logic, since it is very similar to the use above.

wjrosa (approved):
Thanks for the nice contribution! Code looks and it works well:

Screenshot 2025-07-01 at 16 35 57
Screenshot 2025-07-01 at 16 36 17

Left a comment that you kinda noticed on your PR description. So feel free to ignore it.

Tip: An easier way to test this is by creating a pending order in the wp-admin for your testing customer.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
elvismdev: @wjrosa  Thanks for the feedback! I've refactored the asset loading logic into a private `get_asset_data()` method that both `register_express_checkout_script()` and `scripts()` now use.

wjrosa: Looking great! Thanks for the refactor

malithsen (approved):
Thanks for the PR! 
This looks good to me. E2Es are failing because the PR is made from a fork. I ran them locally, and they are passing.

daledupreez:
@wjrosa and @malithsen, could one of you make sure we add details for this fix to our changelog and readme?

PR #4452, opened 2025-07-01
Removing the payment method customization feature

wjrosa:
Fixes STRIPE-489
See p1748852469180899-slack-C055WHLA98D?thread_ts=1748576969.440509&cid=C055WHLA98D

We are removing the individual customization of the payment methods feature entirely. Originally, this issue was intended to remove it only when the Optimized Checkout is enabled, but as @souravdebnath1986 mentioned in Slack, we need to remove it also for UPE to keep us in compliance with payment method rules.

Besides removing the "Customize" button in settings (and related frontend code), I am also deleting related unit tests (for both frontend and the backend), and deprecating the public backend methods.

## Testing instructions

- Code review
- Check if the tests are still passing (frontend and backend)
- Checkout and build this branch on your test environment (`update/removes-the-pm-customization-feature`)
- Connect your Stripe account
- Go to the payment methods settings page (`wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=methods`)
- Confirm the "Customize" buttons are gone:
Screenshot 2025-07-01 at 10 57 00

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


malithsen (approved):
LGTM, tested with OC turned on/off, customize button is no longer present

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -644,6 +634,8 @@ private function update_enabled_payment_methods( $payment_method_ids_to_enable,
 	 * @param WP_REST_Request $request Request object.
 	 *
 	 * @return WP_REST_Response
+	 *
+	 * @deprecated 9.6.0 The customization of individual payment methods is now deprecated.
```
Mayisha: We should remove the rest endpoint (`/payment_method_order`) as well as this function.

Mayisha (approved):
Changes look good to me. 👍 

✅ Customize button is not available on the settings page.
✅ The custom title/description that was saved before is not visible on the checkout page. 

I left a suggestion about removing the rest endpoint.

includes/admin/class-wc-rest-stripe-settings-controller.php
wjrosa: Alright. I removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4452/commits/7791f944a0dcff569eaea50c8838b912ad6a740c (`/payment_method`). The `/payment_method_order` endpoint should still be used by UPE

Mayisha: My bad. I mistakenly wrote `/payment_method_order` but I meant `/payment_method` endpoint 🤦‍♀️

PR #4424, opened 2025-06-19
ECE: add support for custom fields in classic checkout

annemirasol:
Towards STRIPE-205
Towards #2969

## Changes proposed in this Pull Request:
This PR adds a way for stores to process custom checkout field data, for Express Checkout orders on classic checkout.

Custom classic checkout fields can be added by ([docs](https://developer.woocommerce.com/docs/code-snippets/customising-checkout-fields/)):
   - Using a plugin like Checkout Field Editor for WooCommerce, -or-
   - Adding them manually via actions and filters:
      1. Displaying the fields via `woocommerce_checkout_fields`, `woocommerce_billing_fields` and `woocommerce_shipping_fields`.
      2. Performing custom data validation via `woocommerce_after_checkout_validation`.
      3. Saving the data (as order metadata) via `woocommerce_checkout_update_order_meta`.
      4. Optionally, displaying the data in the Thank You page and wp-admin, via `woocommerce_order_details_after_order_table`.
   
This PR focuses on the second case, **manually added custom fields**.

### Background

Express Checkout uses Store API to create orders, even when the store is on classic checkout. Store API does not use or invoke the filters and actions listed above, which means it is unaware of custom fields added via `woocommerce_[checkout|billing|shipping]_fields`, and will not run any code hooked into `woocommerce_after_checkout_validation` or `woocommerce_checkout_update_order_meta`, etc. 

This PR bridges the custom checkout field gap by:
1. Determining the list of custom checkout fields, and extracting the data from the checkout form and passing it to Store API, using its mechanism for extensions data, i.e. `extensions`
2. Providing actions that stores can hook into, with signatures similar to `woocommerce_after_checkout_validation` and `woocommerce_checkout_update_order_meta`.
   - Ideally, the same functions that are hooked into the classic actions can be used for the Express Checkout actions. For example: 

    ```
    // Existing hook to save custom checkout data to order
    add_action( 'woocommerce_checkout_update_order_meta', 'my_function_save_custom_checkout_data', 10, 2 );

    // Add this for express checkout:
    add_action( 'wc_stripe_express_checkout_update_order_meta', 'my_function_save_custom_checkout_data', 10, 2 );
    ```

## Testing instructions
1. Install Code Snippet or a similar plugin, and add custom classic checkout fields using the code in https://gist.github.com/annemirasol/6960e65519197c621b547cad38bfa128.
2. Add a product and go to classic checkout.
3. Verify that your custom checkout fields are displayed.
5. Go back to your code snippet, and add these lines:

[...]

PR #4394, opened 2025-06-04
ECE: add support for extensions data

annemirasol:
Towards STRIPE-205
Towards #2969

## Changes proposed in this Pull Request:

WooCommerce Store API provides a way for third-party extensions to pass custom checkout data, via `CheckoutExtensionData` ([here](https://github.com/woocommerce/woocommerce/blob/2410c591abeb4872182145672d78ac64674f8a24/plugins/woocommerce/client/blocks/assets/js/base/context/hooks/use-checkout-extension-data.ts#L11)).

When extension data is available, we should include it when crafting the payload for express checkout. This will add support for ECE and custom checkout fields, for the following setup:
   - store is on block checkout
   - store is using a checkout editor plugin that relies on `CheckoutExtensionData` to process custom data, e.g. Checkout Field Editor for WooCommerce

Screenshot 2025-06-18 at 12 36 42 PM

## Testing instructions

1. Install and activate [ThemeHigh's Checkout Field Editor for WooCommerce](https://wordpress.org/plugins/woo-checkout-field-editor-pro/).
2. Navigate to WooCommerce > Checkout Form > Block Checkout Fields (tab) > Additional Information (tab).
3. Add a custom field, like this:

Screenshot 2025-06-18 at 12 44 05 PM

5. As a shopper, add a product and go to block checkout.
6. You should see the new field in the checkout form. (Note: unless you edit the location via the Block Editor, it will be at the very bottom of the checkout page)
7. Fill out the custom field, and use express checkout to complete your purchase.
8. In the Order Confirmation page, verify that the custom field data appears.

Screenshot 2025-06-18 at 12 51 44 PM

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [x] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4390, opened 2025-06-03
ECE: Add support for custom fields in block checkout

annemirasol:
Partially fixes STRIPE-205
Partially fixes #2969 

## Changes proposed in this Pull Request:
This PR adds support for ECE and custom checkout fields, limited to: 
- block checkout, and
- fields added via Additional Checkout Fields API (i.e. `woocommerce_register_additional_checkout_field()`).

In subsequent PRs, we will add support for
- custom fields added by third-party plugins that use `extensions` (block checkout only) (#4394)
- custom fields for classic checkout

## Testing instructions
 1. Add custom checkout fields to the block checkout:
    - Install [Code Snippets](https://wordpress.org/plugins/code-snippets/) or a similar plugin, and add the snippet provided in the comment below.
2. As a shopper, add a product to cart and go to block checkout.
3. Notice the additional fields in the checkout form. Fill out these fields.
4. Use express checkout to complete your purchase.
5. Verify in the Order Confirmation page that the fields have been persisted.
   - You may also check the `wp_wc_orders_meta`.
   - The `meta_key` will be prefixed with `_wc_billing/`, `_wc_shipping/` or `_wc_other/`.
6. Try using ECE in the product or cart pages.
   - Verify there are no unexpected failures. See known limitations about required custom fields below.
   - Verify it avoids using stale session data: since the custom fields are not presented to the shopper inside these pages, there should be no custom data stored for these orders.

Known limitations:
- When marking a custom field as "required" in `woocommerce_register_additional_checkout_field()`, express checkout will fail when the data is not provided. This invariably means ECE is not usable outside of the block checkout page.
- When validation fails, the error message from the server is displayed but the input field is not highlighted in red

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

Issue #2969, opened 2024-03-06
Custom checkout fields not getting included in checkout form data submitted by GPay/Apple Pay


tarunparswani26:
**Describe the bug**
Encountered an issue with our Smart Coupons plugin affecting custom fields on the checkout page. When utilizing GPay/Apple Pay, the Checkout form data does not include information from these custom fields.

Upon inspecting the getOrderData function ([ref](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/assets/js/stripe-payment-request.js#L92)), it appears that the function generates a checkout form data with predefined, hardcoded fields. Unfortunately, there is no mechanism to extend or add additional data, similar to WooCommerce checkout on the PHP side.

Also,  incorporating additional data through JS code in the AJAX call is not feasible due to the use of iframes for rendering GPay/Apple Pay payment buttons.

Do let me know if any further details are needed for this.

thuautp:
also reported in 8099290-zen

PR #4455, opened 2025-07-01
ECE: custom checkout field support improvements

annemirasol:
Towards STRIPE-205
Towards #2969

In #4390, #4394, and #4424, we added express checkout support for custom checkout fields, for block checkout and classic checkout. This PR iterates on these changes, and introduces the following improvements:

1. Wrap the feature in a default-off feature flag, for a more controlled rollout (classic checkout only).
2. Move code to its own class, for better organization and to avoid cluttering the main express checkout class.
3. Use [FormData API](https://developer.mozilla.org/en-US/docs/Web/API/FormData) to extract checkout form data, instead of manually searching form elements by name and extracting their value.
4. Add previously missing support for multiselect fields (classic checkout).
5. Add previously missing required fields validation (classic checkout).
6. Use Store API's `RouteException` for [consistency](https://github.com/woocommerce/woocommerce/blob/d825abfef66ba5e3d7214f65e684a6fd37399625/plugins/woocommerce/src/StoreApi/Legacy.php#L64) and better error code control, i.e. return a `400` instead of `500` for invalid data (classic checkout).
7. Fix logic for preventing use of stale Cart/Checkout Store data in block checkout.

**Implementation Details**
1. Why are we adding a feature flag for classic checkout?
   - We previously ignored custom fields for ECE orders. If the store had "required" custom classic checkout fields, ECE orders went through even with those fields empty.
   - With these changes, ECE orders will now return a validation error if a required custom field is empty. Among other things, this means express checkout + required custom fields + product/cart pages are no longer compatible.
   - Because of this, we want to give the merchant a chance to evaluate their setup, e.g. Is this field really required? Do we want express checkout in product/cart pages?, and opt-in instead of silently rolling out these behavior changes.
   - Why classic checkout only? Custom fields registered for block checkout already behave this way. Required fields are enforced, even for express checkout orders.

## Testing instructions
We need to smoke test the refactors, test multiselect field support and required fields validation.

1. Install Code Snippets or similar plugin, and add the snippet in https://gist.github.com/annemirasol/2e71ceb3a8bca472e4abfdbccb97e51e to add a custom multiselect field for classic checkout.
2. Add a product to cart and go to _classic_ checkout. Verify the custom checkout field is present.

Screenshot 2025-07-02 at 9 18 09 AM

3. Go back to your snippet and add the following lines. This enables the newly-introduced feature flag, and hooks into the ECE action provided for processing custom classic checkout fields (introduced in #4424)

```
// Enable ECE support for custom checkout fields
add_filter( 'wc_stripe_express_checkout_enable_classic_checkout_custom_fields', '__return_true' );

// Hook into ECE action for processing custom classic checkout fields
add_action( 'wc_stripe_express_checkout_update_order_meta', 'custom_checkout_fields_save_to_order', 10, 2 );
```
4. Go back to classic checkout. Without filling out the new field, try to complete the purchase using Google Pay/Apple Pay/Link.
5. Verify validation fails because a required field is missing.

Screenshot 2025-07-02 at 9 29 39 AM

6. Fill out the field and complete the purchase.
7. Verify the custom field is persisted and displayed in wp-admin or order confirmation page. 

Screenshot 2025-07-02 at 9 31 19 AM

Optional:
- Want to smoke-test test various field types for classic checkout? Here's a handy snippet: https://gist.github.com/annemirasol/a73b869534fd1c91225ca7326ce41c68
- Want to smoke-test test custom fields and block checkout? Here's a handy snippet: https://gist.github.com/annemirasol/aa33c143ef999761eb5f057a427bc1eb

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
```diff
@@ -0,0 +1,275 @@
...
+
+			return $classic_custom_checkout_fields;
+		}
+
+		if ( is_cart() || is_product() ) {
+			return array_merge(
+				$this->get_custom_checkout_fields( 'block' ),
+				$this->get_custom_checkout_fields( 'classic' )
+			);
+		}
```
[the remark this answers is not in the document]
annemirasol: We need the list of custom checkout fields in product and cart pages to pass "empty" custom fields, to avoid possibly stale session data from being used.

readme.txt
```diff
@@ -164,7 +164,9 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Dev - Add e2e tests to cover Affirm purchase flow
 * Dev - Improve e2e tests of some of the LPMs
 * Fix - Add safety check when checking error object
+* Update - Include extension data from block checkout when submitting an express checkout order
```
malithsen:
```suggestion

```
This entry is a duplicate

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
```diff
@@ -0,0 +1,275 @@
...
+	private function get_sanitized_field_value( $value, $type ) {
+		if ( '' === $value ) {
+			return '';
+		}
+
+		switch ( $type ) {
+			case 'checkbox':
+				return empty( $value ) ? $value : 1;
+			case 'multiselect':
+				return implode( ', ', wc_clean( $value ) );
```
malithsen, line 157: nit: This converts the data structure from an array to a comma-separated string. Is that necessary, especially since we are inside a sanitization function?

malithsen (approved):
Nice work!!

Thanks for the clear test instructions and the test snippets 💯 

Image

readme.txt
annemirasol: Removed after merge.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -114,149 +111,10 @@ public function init() {
...
-	 * we pass custom data as extended data, i.e. under extensions.
-	 *
-	 * @param WP_REST_Request $request The request object.
-	 * @return array Custom checkout data.
-	 */
-	private function get_custom_checkout_data_from_request( $request ) {
-		$extensions = $request->get_param( 'extensions' );
-		if ( empty( $extensions ) ) {
-			return [];
+		if ( apply_filters( 'wc_stripe_ece_enable_custom_checkout_fields_support', false ) ) {
```
Mayisha:
```suggestion
                 // Document the filter details here
                 $enable_custom_checkout_fields_support = apply_filters( 'wc_stripe_ece_enable_custom_checkout_fields_support', false );
		if ( $enable_custom_checkout_fields_support ) {
```

I think, it would be better if we document the filter here.

Mayisha (approved):
Thanks @annemirasol for the improvements and the detailed testing instructions.

This is working well. 

✅ I get a validation error when I try to pay with GPay without filling up the custom field.

Screenshot 2025-07-07 at 2 50 53 PM

✅ Checkout succeeds when I fill up the custom field. Order confirmation page has the custom field details.

Screenshot 2025-07-07 at 2 52 10 PM


I have a suggestion about documenting the filter, but that's not a blocker.

PR #4462, opened 2025-07-03
Moving existing order lock and bailing out when it is active

### LINKED ISSUE #4149: Woocommerce Checkout Doesn't Progress to Thank You Page after Successful Charge

wjrosa:
Partially fixes [STRIPE-371](https://linear.app/a8c/issue/STRIPE-371)
Partially fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4149
Suggested in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4432#issuecomment-3014056513
POC PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4450

This PR applies @annemirasol's suggestion [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4432#issuecomment-3014056513) to reuse the existing order lock functionality instead of creating another (which I tried [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4432)).

This is part of our efforts to avoid duplicate charges.

## Testing instructions

- Checkout to this branch on your test environment (`add/moving-existing-order-lock-and-bailing-out-when-active`)
- Connect your Stripe account
- Open a checkout page in 2 different tabs and fill in all the required information
- On both tabs, try to place the order at the same time (with less than a minute of difference)
- You should see the following error on your second tab:
Screenshot 2025-06-24 at 17 40 31

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1186,6 +1193,9 @@ private function process_payment_with_payment_method( int $order_id ) {
 			);
 		} catch ( WC_Stripe_Exception $e ) {
 			return $this->handle_process_payment_error( $e, $order );
+		} finally {
+			// Ensure the order is unlocked in case of an exception.
+			$this->unlock_order_payment( $order );
```
Mayisha, line 1197: To handle the case of an exception, we should call the `unlock` function in the `catch` block instead of `finally`. And also keep the previous `unlock` call in place.  

If we unlock it in finally it might unlock an order which was not locked in the current request.

I haven't been able to reproduce this scenario successfully but I am thinking about this flow-
- Request 1 calls `process_payment_with_payment_method` and adds a lock.
- Request 2 calls `process_payment_with_payment_method` and comes to the newly added condition.

```
if ( $this->is_order_payment_locked( $order ) ) {
				// If the request is already being processed, return an error.
				return [
					'result'   => 'failure',
					'redirect' => '',
					'message'  => __( 'Your payment is already being processed. Please wait.', 'woocommerce-gateway-stripe' ),
				];
			}
```

- Request 2 returns because of this condition but unlocks the order as the unlock call is in finally block.
- A webhook comes and  processes the same order which is now unlocked.

wjrosa: I think it makes sense conceptually, so I moved the unlock call in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4462/commits/df3087af70026918b977766e892d2afce06c906e

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1033,6 +1033,18 @@ private function process_payment_with_payment_method( int $order_id ) {
...
+				return [
+					'result'   => 'failure',
+					'redirect' => '',
+					'message'  => __( 'Your payment is already being processed. Please wait.', 'woocommerce-gateway-stripe' ),
+				];
+			}
+
+			// Lock the order before we create and confirm the payment/setup intents to prevent Stripe sending the success webhook before this request is completed.
+			$this->lock_order_payment( $order );
+
```
annemirasol: I like how this is readable. Slight concern that it adds an extra "check" step, redundant because checking for an existing lock is already performed inside `lock_order_payment( $order )`. I think we want locking to be ideally with minimal operations as possible, to reduce the surface for race conditions.

Maybe something like this? 
```
// Attempt to acquire lock, bail if already locked
$is_already_locked = $this->lock_order_payment( $order );
if ( $is_already_locked ) {
   return [...];
}
```

What do you think?

annemirasol (commented):
I also noticed that we don't display the informative error message for **shortcode** checkout. This should probably be fixed in a separate PR -- IIRC we always output this generic error message for shortcode checkout for server-side errors.

Screenshot 2025-07-08 at 10 33 20 AM

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Makes total sense! Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4462/commits/dfe242f626e760d281de8f57344000290c1984a6

annemirasol (approved):
LGTM! Thanks for working on this :shipit:

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -1785,6 +1775,38 @@ public function unlock_order_payment( $order ) {
...
+	 * @return bool
+	 */
+	protected function is_order_payment_locked( $order ) {
+		$existing_lock = $this->get_order_existing_lock( $order );
+		if ( $existing_lock ) {
+			$parts      = explode( '|', $existing_lock ); // Format is: "{expiry_timestamp}" or "{expiry_timestamp}|{pi_xxxx}" if an intent is passed.
+			$expiration = (int) $parts[0];
+
+			// If the lock is still active, return true.
+			if ( time() <= $expiration ) {
```
Mayisha, line 1801: Just wanted to confirm, is the intent-related check intentionally removed?

Mayisha (approved):
Thank you for the improvement. Looks good to me. 👍 

I have the same observation as Anne for the shortcode checkout notice and a quick question about the change of logic in the locking method.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Not really, it got lost during the change in the implementation. Thanks for flagging. But I attempted to add it back to see how it would behave, and it does not work correctly due to the lack of payment intent at the beginning of the payment flow. I removed it for good now, along with the remaining references.

PR #4466, opened 2025-07-07
Add hook to allow for custom webhook post-processing

daledupreez:
Towards STRIPE-564

This PR proposes the addition of a new `wc_stripe_webhook_received ` action that is triggered just after we process an inbound webhook from Stripe. This will make it easier for merchants and developers to include more details in WooCommerce based on the data from Stripe. The initial use case for this is to include more data for charges to allow for easier fraud detection while processing orders.

## Testing instructions

- Checkout to this branch
- Connect your Stripe account
- Make sure your store is publicly accessible and has webhooks configured
- Add the following code snippet:
```php
add_action( 'wc_stripe_webhook_received', function( $notification_type, $notification_data ) {
	WC_Stripe_Logger::info( 'TEST: wc_stripe_webhook_received action ' . $notification_type . ' with data: ' . print_r( $notification_data, true ) );
}, 10, 2 );
```
- As a shopper, purchase any product
- As a merchant, check your latest Stripe logs
- Confirm you can see the test log:
Screenshot 2025-07-18 at 11 53 50

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1354,6 +1354,17 @@ public function process_webhook( $request_body ) {
...
+		/**
+		 * Fires after a webhook has been processed, but before we respond to Stripe.
+		 * This allows for custom processing of the webhook after it has been processed.
+		 *
+		 * @since 9.7.0
+		 *
+		 * @param string $webhook_type The type of webhook that was processed.
+		 * @param object $notification The webhook data sent from Stripe.
+		 */
+		do_action( 'wc_stripe_webhook_processed', (string) $notification->type, $notification );
```
daledupreez: It feels like we should also include an order reference where possible.

diegocurbelo: `payment_intent.succeeded` and `payment_intent.amount_capturable_updated` events [might be processed async](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/6d4b809b308cbe5c5e11b23d6778283575b9bd83/includes/class-wc-stripe-webhook-handler.php#L1075)

Without a major refactor, what we can do is exclude those events from this action, and do something similar to what was done for the `do_action( 'wc_gateway_stripe_process_payment', $charge, $order )` call, which is called in both cases with the same params (the charge and the order):

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/6d4b809b308cbe5c5e11b23d6778283575b9bd83/includes/class-wc-stripe-webhook-handler.php#L1078

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/6d4b809b308cbe5c5e11b23d6778283575b9bd83/includes/class-wc-stripe-webhook-handler.php#L1267

Specifically for the request in [STRIPE-564](https://linear.app/a8c/issue/STRIPE-564), the existing action might give them the info they want (it's called with the [latest charge](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/6d4b809b308cbe5c5e11b23d6778283575b9bd83/includes/class-wc-stripe-webhook-handler.php#L1258))

malithsen: We might want to wrap the hook in a `try/catch` (and log the error on failure) so that it doesn't interrupt rest of the flow if something fails within the hook.

wjrosa: > It feels like we should also include an order reference where possible.

I attempted to add the `$order` to the action, but some events have a different way to retrieve the order information. We would have to either:
- Move the order retrieval logic for each event type to the `process_webhook` and pass it as an additional parameter for every method handling an event inside of it
- Move the action call inside every method handling an event (duplicating it multiple times)

So, unless that's necessary for some reason, it's not worth the refactoring.

> Without a major refactor, what we can do is exclude those events from this action, and do something similar to what was done for the do_action( 'wc_gateway_stripe_process_payment', $charge, $order ) call, which is called in both cases with the same params (the charge and the order):

Do you mean something like [this](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4466/commits/aa7989c704b5afbf959df3d9c10ff63629a23929), Diego?

wjrosa: > We might want to wrap the hook in a try/catch (and log the error on failure) so that it doesn't interrupt rest of the flow if something fails within the hook.

Agreed! Added it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4466/commits/5866fb2bb22a8c437e8de904fa7f59fc2534fd86

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1096,6 +1096,21 @@ public function process_payment_intent( $notification ) {
...
+						 * This allows for custom processing of the webhook after it has been processed.
+						 *
+						 * @since 9.8.0
+						 *
+						 * @param string $webhook_type The type of webhook that was processed.
+						 * @param object $notification The webhook data sent from Stripe.
+						 */
+						do_action( 'wc_stripe_webhook_processed', (string) $notification->type, $notification );
+					} catch ( Exception $e ) {
+						WC_Stripe_Logger::error( 'Error in wc_stripe_webhook_processed action: ' . $e->getMessage() );
```
diegocurbelo: We can now add the full error to the log context:
```suggestion
						WC_Stripe_Logger::error( 'Error in wc_stripe_webhook_processed action: ' . $e->getMessage(), [ 'error' => $e ] );
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1395,6 +1409,22 @@ public function process_webhook( $request_body ) {
 				$this->process_setup_intent( $notification );
 
 		}
+
+		// These events might be processed async. Skip the action trigger for them here. The trigger will be called inside the specific methods.
+		if ( 'payment_intent.succeeded' === $notification->type || 'payment_intent.amount_capturable_updated' === $notification->type ) {
+			return;
+		}
```
diegocurbelo, line 1460: This seems to indicate that `wc_stripe_webhook_processed` would be called from one of the async webhook processing methods when the event type is `payment_intent.succeeded` or `payment_intent.amount_capturable_updated`.

Then, we'd need to call it from either `process_deferred_webhook()`  (line 1278) or `handle_deferred_payment_intent_succeeded()` (line 1324), right?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1395,6 +1409,22 @@ public function process_webhook( $request_body ) {
...
+
+		/**
+		 * Fires after a webhook has been processed, but before we respond to Stripe.
+		 * This allows for custom processing of the webhook after it has been processed.
```
diegocurbelo: We should rephrase this: for `payment_intent.succeeded` and `payment_intent.amount_capturable_updated`, this action might be executed before the event is processed.

When using deferred handling, we are still responding with a 200 status code to Stripe immediately (in `check_for_webhook()`).

diegocurbelo (changes requested):

includes/class-wc-stripe-webhook-handler.php
wjrosa: I did duplicate the call for the action in `process_payment_intent` (line 1110). I took a look at moving it to `process_deferred_webhook`, and the problem there is that we don't have the full webhook event contents to pass. I could refactor the job scheduling method, but I am not sure if it is OK to pass that much data (compared with the current version). Meaning: why do we just pass `order_id` and `intent_id` instead of the whole notification?

includes/class-wc-stripe-webhook-handler.php
wjrosa: Gotcha. I will keep it that way for now, as I am not using the deferred handling yet.

daledupreez: Might it make sense to have a third argument to the filter that indicates whether the inbound data has been processed yet? It feels like that might be clearer rather than having "missing" events.

includes/class-wc-stripe-webhook-handler.php
daledupreez: I had a different thought about this -- the logic for resolving the order for each event should remain closer to the processing code, as the logic does vary across event types. Rather than returning the order to the caller (and thus changing all method signatures and code paths), might it make sense to create a private instance variable (e.g. `$this->resolved_order`) that each code path populates, and we then include the value of that variable in the filter call? It does introduce some statefulness to the webhook class, but it feels like it would be a relatively small code change for a major boost in developer utility.

includes/class-wc-stripe-webhook-handler.php
diegocurbelo: > why do we just pass order_id and intent_id instead of the whole notification?
 
Looking at the code, it seems to be because for those event types the notification contains the `payment_intent`, and as we are processing them asynchronously, we get the (maybe updated) `intent` from Stripe (using `get_intent_from_order( $order )`) , and then the actual `charge`.

>I could refactor the job scheduling method, but I am not sure if it is OK to pass that much data (compared with the current version)

I think that passing the full notification and calling the new action at the end of `process_deferred_webhook( $webhook_type, $additional_data )` makes the most sense.

We can also look into why we need to process those two events asynchronously; maybe with the order locking we have in place, it's no longer needed, or we can improve the order locking so we don't need to handle these events async.

includes/class-wc-stripe-webhook-handler.php
wjrosa: That's a good idea. Added it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4466/commits/a63cbe1654fabdedd0ee2d70ea6f1ad29a1ba251

includes/class-wc-stripe-webhook-handler.php
wjrosa: You mean basically pass that parameter always as `true`, except when triggering the action inside the deferred events?

includes/class-wc-stripe-webhook-handler.php
wjrosa: Gotcha! I updated the code to pass the notification object in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4466/commits/cd17df2f3e36a71dd919f5dfc2ff23f888c6a6cd

> We can also look into why we need to process those two events asynchronously; maybe with the order locking we have in place, it's no longer needed, or we can improve the order locking so we don't need to handle these events async.

I think that is worth tackling on another issue/PR 👍

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1262,6 +1328,17 @@ public function process_deferred_webhook( $webhook_type, $additional_data ) {
...
+					 * Fires after a webhook has been processed, but before we respond to Stripe.
+					 * This allows for custom processing of the webhook after it has been processed.
+					 *
+					 * @since 9.8.0
+					 *
+					 * @param string $webhook_type The type of webhook that was processed.
+					 * @param object $notification The webhook data sent from Stripe.
+					 * @param WC_Order $order The order being processed by the webhook.
+					 */
+					do_action( 'wc_stripe_webhook_processed', (string) $webhook_type, $notification, $this->resolved_order );
```
diegocurbelo: We should move this to after the `switch` to make it explicit that we expect it to be called for all async events... so we don't forget to call it when handling a new event type

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -49,6 +49,13 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	 */
 	protected $deferred_webhook_action = 'wc_stripe_deferred_webhook';
 
+	/**
+	 * The order object being processed.
+	 *
+	 * @var WC_Order
+	 */
+	protected $resolved_order;
```
diegocurbelo: I'm not a fan of this; setting this Order attribute everywhere to be used only when the hook `wc_stripe_webhook_processed` is registered seems weird to me. It's the kind of thing we might forget when adding code to handle a new `event_type`,

Also, not all `events` have an Order associated (ie, `account.updated`), as we add more features, we might have more of those..

We are already passing the "notification event" received from Stripe to the `wc_stripe_webhook_processed` hook. If needed, the order can be obtained from the notification payload object: (ie: `WC_Stripe_Helper::get_order_by_charge_id( $notification->data->object->charge )`)

diegocurbelo (changes requested):

includes/class-wc-stripe-webhook-handler.php
daledupreez: > We are already passing the "notification event" received from Stripe to the wc_stripe_webhook_processed hook. If needed, the order can be obtained from the notification payload object: (ie: `WC_Stripe_Helper::get_order_by_charge_id( $notification->data->object->charge )`)

My concern here is that the logic across all the code paths for various webhooks to identify the relevant order is not really consistent. So expecting users to know which code to call is likely to fail for some event types. So in this case, I'd rather have us mostly right in a way we control and not push the expectation of "working it out" onto users, where their code may break with upgrades due to new events or similar.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Good point. Moved it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4466/commits/c442208065845153f985e56f2ed5cf2c0605ff50

includes/class-wc-stripe-webhook-handler.php
wjrosa: I am also not a big fan of this approach for the same reasons, but I understand Dale's point of view, too. It is like a "bad code" that is useful for merchants. I am not sure what the best course of action is here (keep it or remove it).

diegocurbelo: My thought was that we are adding this hook for _"merchants and developers to include more details in WooCommerce based on the data from Stripe"_, and to get the data from the event notification, they need to know the payload type/attributes; but, I agree that knowing which specific methods to use to get the Order from the Stripe webhook is a different thing, and merchants might be familiar with Stripe events but not with the gateway code.

In that case, we can refactor our get order logic to use a new `get_webook_order(...)` or `get_order_from_webhook(...)` function that encapsulates how to get the order related/linked to the notification event (if there is one), so we have that code in one place and not scattered throughout the file.

daledupreez: > In that case, we can refactor our get order logic to use a new get_webook_order(...) or get_order_from_webhook(...) function that encapsulates how to get the order related/linked to the notification event (if there is one), so we have that code in one place and not scattered throughout the file.

That's probably the right starting point, to be honest. @wjrosa, do you want to tackle that? Or should I pick that up in a new PR?

wjrosa: I'd appreciate it if you could handle that, Dale 🙏 . Should I keep the `$resolved_order` logic here or revert it meanwhile?

diegocurbelo (approved):
We decided to implement a unified `get_webook_order(...)` or `get_order_from_webhook(...)` in a new PR, so meanwhile I'm approving this one.

wjrosa:
@daledupreez should I go ahead and merge this? Since I see you are working on the other PR, and Diego approved this one (and I tested it myself when developing it).

daledupreez:
@wjrosa, if you don't mind waiting, can I review it one more time on Monday and make sure I am up to speed on the current state? That should also make it easier for me to update my follow-up PR once this change is merged.

wjrosa:
@daledupreez sure! No worries

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
+* Add - Adds a new action (`wc_stripe_webhook_processed`) to allow hooking into the webhook processing flow
```
daledupreez: Suggestion: it may be worth specifying a more user-oriented goal in this entry.
```suggestion
* Add - Adds a new action (`wc_stripe_webhook_processed`) to allow additional actions to be taken for webhook notifications from Stripe
```

includes/class-wc-stripe-webhook-handler.php
daledupreez: Refactor underway in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4546

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1240,8 +1302,9 @@ protected function defer_webhook_processing( $webhook_notification, $additional_
 	 *
 	 * @param string $webhook_type    The webhook event name/type.
 	 * @param array  $additional_data Additional data passed to the scheduled job.
+	 * @param stdClass $notification  The webhook notification payload.
 	 */
-	public function process_deferred_webhook( $webhook_type, $additional_data ) {
+	public function process_deferred_webhook( $webhook_type, $additional_data, $notification ) {
```
daledupreez: I think we should explicitly default `$notification` to `null` to handle cases where the data was not included, e.g. from just before an upgrade.
```suggestion
	public function process_deferred_webhook( $webhook_type, $additional_data, $notification = null ) {
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1269,6 +1335,17 @@ public function process_deferred_webhook( $webhook_type, $additional_data ) {
...
+			/**
+			 * Fires after a webhook has been processed, but before we respond to Stripe.
+			 * This allows for custom processing of the webhook after it has been processed.
+			 *
+			 * @since 9.8.0
+			 *
+			 * @param string $webhook_type The type of webhook that was processed.
+			 * @param object $notification The webhook data sent from Stripe.
+			 * @param WC_Order $order The order being processed by the webhook.
+			 */
```
daledupreez: Nit: we should only have one primary PHPDoc section for the hook so we don't conflict. We can pick which one should be primary, but we should avoid the duplication.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1397,6 +1474,23 @@ public function process_webhook( $request_body ) {
 				$this->process_setup_intent( $notification );
 
 		}
+
+		// These events might be processed async. Skip the action trigger for them here. The trigger will be called inside the specific methods.
+		if ( 'payment_intent.succeeded' === $notification->type || 'payment_intent.amount_capturable_updated' === $notification->type ) {
```
daledupreez, line 1458: It feels like the code handling the event should tell us whether the event was (or was not) processed successfully. Having special code separate from the handling feels like it could be brittle. But that can happen in a follow-up.

At a philosophical and naming perspective, might it make sense to call the action `wc_stripe_webhook_received` so there isn't an implication that we processed the webhook? In various cases, we handle the webhook without taking any action due to the state of our data, but this is getting deeper into the weeds.

includes/class-wc-stripe-webhook-handler.php
daledupreez: Noting that we now call the action after processing the event async, so the distinction is somewhat moot.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1397,6 +1474,23 @@ public function process_webhook( $request_body ) {
...
+		 * Fires after a webhook has been processed, but before we respond to Stripe.
+		 * This allows for custom processing of the webhook after it has been processed.
+		 *
+		 * @since 9.8.0
+		 *
+		 * @param string $webhook_type The type of webhook that was processed.
+		 * @param object $notification The webhook data sent from Stripe.
+		 * @param WC_Order $order The order being processed by the webhook.
+		 */
+		do_action( 'wc_stripe_webhook_processed', (string) $notification->type, $notification, $this->resolved_order );
```
daledupreez: It feels like this call needs to be wrapped in a `try/catch` to prevent unintended (and fairly messy) breakages because our webhook response to Stripe gets broken.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1269,6 +1335,17 @@ public function process_deferred_webhook( $webhook_type, $additional_data ) {
...
+			 * This allows for custom processing of the webhook after it has been processed.
+			 *
+			 * @since 9.8.0
+			 *
+			 * @param string $webhook_type The type of webhook that was processed.
+			 * @param object $notification The webhook data sent from Stripe.
+			 * @param WC_Order $order The order being processed by the webhook.
+			 */
+			do_action( 'wc_stripe_webhook_processed', (string) $webhook_type, $notification, $this->resolved_order );
 		} catch ( Exception $e ) {
```
daledupreez, line 1326: We should catch `Throwable` to be more robust against bad code. 😬  
```suggestion
		} catch ( Throwable $e ) {
```

daledupreez: Updated in 6f832701806f6dfa2f640a6b354063414def16f8.

includes/class-wc-stripe-webhook-handler.php
daledupreez: Added in 6f832701806f6dfa2f640a6b354063414def16f8.

daledupreez: And then centralised in 479c7b2c4bf58d5a92efe9b0e31a77e836191226.

includes/class-wc-stripe-webhook-handler.php
daledupreez: And centralised in 479c7b2c4bf58d5a92efe9b0e31a77e836191226.

includes/class-wc-stripe-webhook-handler.php
daledupreez: I ended up addressing this by centralising the trigger logic in 479c7b2c4bf58d5a92efe9b0e31a77e836191226.

daledupreez (commented):
@wjrosa, I ended up noticing some subtle but important things that I addressed directly, so I think I'd like your eyes on this before we ship:
 * I centralised the logic to trigger the action in a helper function, as we were only wrapping it in a `try/catch` in one location, and we were catching `Exception` instead of `Throwable`, and we had slightly different logging as a result. Everything is now consistent.
 * I added explicit initialisation of `$this->resolved_order` to `null`, and documented that the argument may be `null` when the webhook is not related to an order. That is really important for developers to be aware of.

I have some other comments that I think are less critical, but are worth a look.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Yeah, it makes more sense to me as well. Renamed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4466/commits/0b78f6f98706e79c77b30a0c5dad464069d72ecb

wjrosa (approved):

malithsen (approved):

wjrosa:
Thanks for the review, @malithsen!

@daledupreez should we merge this now?

daledupreez:
Apologies for the missed ping, @wjrosa! I think we should get this merged -- I'll take care of that today if we get a clean test run.

daledupreez:
Merging as the e2e test failure doesn't look related.

PR #4468, opened 2025-07-07
Improve plugin logger

diegocurbelo:

This PR improves the plugin logger to adhere to the [WooCommerce logging best practices](https://developer.woocommerce.com/docs/best-practices/data-management/logging/#best-practices), particularly these:
- Rather than using the WC_Logger's `log()` method directly, it's better to use one of the wrapper methods that's specific to the log level. E.g. `nfo()` or `error()`.
- Avoid outputting structured data in the message string. Put it in a key in the context array instead. The logger will handle converting it to JSON and making it legible in the log viewer.
- If you need to include a stack trace, let the logger generate it for you.

Currently, logging in the plugin is performed through the generic `WC_Stripe_Logger::log(...)`, which has two major problems:
1. All entries are logged as debug (including errors)
2. It does not support structured additional data (`context`)... we generally append `print_r( $request, true ) )` to the message

This PR adds shortcut methods for each log severity level, along with the corresponding log level and additional data, and updates the Webhook Handler class to utilize them (as an example). 

New code should use these new shortcut methods, while existing code still using `WC_Stripe_Logger::log(...)` will be phased out gradually.

**BEFORE**
Screenshot 2025-07-07 at 17 10 00

Screenshot 2025-07-07 at 17 12 19

**AFTER**
Screenshot 2025-07-07 at 12 24 30

Screenshot_2025-07-07_at_16_16_04

## Testing instructions

1. Make sure the webhooks are properly configured
4. In the plugin settings, enable CashApp
5. Add a product to the cart, and go to the `Checkout page`
6. Select `CashApp` as the payment method, and click `Place Order` (no need to complete it, loading the mdoal is enough)
7. Open the plugin logs (`WooCommerce > Status > Logs`, and select `woocommerce-gateway-stripe`)
8. Check that the webhook entries were logged:
    Screenshot 2025-07-07 at 16 56 46
9. Click `Additional context` and check that the additional structured data is present: plugin version, api vesion, and the received event:
    Screenshot 2025-07-07 at 16 57 01
10. Change the webhook secret ('test_webhook_secret') in the settings to an invalid value (using the Dev Tools or WP-CLI)
    `wp option patch update woocommerce_stripe_settings test_webhook_secret 'whsec_INVALID'`
11. Close the modal from Step 4, and click `Place Order` again (or repeat Steps 3 and 4 if needed)
12. Go to the plugin logs again and check that an error entry is present:
    Screenshot 2025-07-07 at 16 15 40
13. Click `Additional context` and check that the additional structured data is present (plugin version, api vesion, request headers, event):
    Screenshot 2025-07-07 at 16 16 04

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo:
Currently, the `log()` method includes this filter to allow ignoring a particular log while still allowing other logs to get through:
```
if ( ! apply_filters( 'wc_stripe_logging', true, $message ) ) {
    return;
}
```
But, I'm not sure if we should include that in the new shortcut/wrapper methods... [Woo has `woocommerce_logger_log_message`](https://developer.woocommerce.com/docs/best-practices/data-management/logging/#turning-off-noisy-logs) that accomplishes the same thing:
```
function my_ignored_logs( $message, $level, $context, $handler ) {
	if ( false !== strpos( $message, 'Look, a squirrel!' ) ) {
		return null;
	}
	
	return $message;
}
add_filter( 'woocommerce_logger_log_message', 'my_ignored_logs', 10, 4 );
```

includes/class-wc-stripe-logger.php
```diff
@@ -52,13 +68,159 @@ public static function log( $message, $start_time = null, $end_time = null ) {
...
+	 * Creates a log entry of type emergency.
+	 *
+	 * @since 9.7.0
+	 *
+	 * @param string $message Message to send to the log file.
+	 * @param array $context Additional context to add to the log.
+	 *
+	 * @return void
+	 */
+	public static function emergency( $message, $context = [] ) {
```
wjrosa, line 91: Instead of creating a specific method for each level and calling the logger method, we could implement the logic in `__call`. And declare all the possible methods in the class PHP doc with the `@method` annotation.

wjrosa (approved):
Nice improvement! Code is good (left an optional suggestion) and works as expected:

> 6. Check that the webhook entries were logged:

Screenshot 2025-07-08 at 11 03 04

> 7. Click Additional context and check that the additional structured data is present: plugin version, api vesion, and the received event:

Screenshot 2025-07-08 at 11 04 50

> 10. Go to the plugin logs again and check that an error entry is present:

Screenshot 2025-07-08 at 11 08 00

> 11. Click Additional context and check that the additional structured data is present (plugin version, api vesion, request headers, event):

Screenshot 2025-07-08 at 11 08 35

includes/class-wc-stripe-logger.php
diegocurbelo: I tried using `__callStatic()` and adding `@method static void info( string $message, array $context = [] ) Creates a log entry of type info` to the class doc, but the autocomplete/navigation don't work in Cursor (it does in PHP Storm 🤷🏼):

I don't have a strong opinion on this... pinging @woocommerce/quark for other opinions

For context the `__callStatic()` would be:
```php
public static function __callStatic( $method, $params ) {
	if ( ! in_array( $method, self::LOG_METHODS ) ) {
		throw new Exception( 'The log severity level ' . $method . ' is not supported' );
	}

	if ( empty( $params[0] ) ) {
		throw new Exception( 'The log message is required' );
	}

	if ( ! self::can_log() ) {
		return;
	}

	if ( empty( self::$logger ) ) {
		self::$logger = wc_get_logger();
	}

	$context = array_merge( self::LOG_CONTEXT, $params[1] ?? [] );
	if ( in_array( $method, [ 'emergency', 'alert', 'critical', 'error' ] ) ) {
		$context['backtrace'] = true;
	}

	call_user_func_array( [ self::$logger, $method ], [ $params[0], $context ] );
}
```

daledupreez: My personal preference is to implement the functions directly. While it's more verbose, it's explicit, and means it's really easy to follow exactly what is happening, regardless of IDE, text editor, or other tool. I find the indirection is more of a hurdle in the long term than helpful, especially in the context of trying to read and follow the code.

malithsen (approved):
Nice work, new logs look great.

includes/class-wc-stripe-logger.php
```diff
@@ -52,13 +68,159 @@ public static function log( $message, $start_time = null, $end_time = null ) {
...
+		if ( ! self::can_log() ) {
+			return;
+		}
```
daledupreez: I don't like the idea that we skip logging critical messages when the high-level logging is disabled. This way, we will generate silent failures that never get logged. In particular, it feels like `emergency()` and `critical()` should always get logged. I think it _might_ also be worth logging `alert()` in all cases too.

diegocurbelo: How about updating the _Debug mode_ message to something like `Save debug information to the Log`, and have it only affect the debug logging (warning, notice, info, and debug)? With error and above always logged.

Image

daledupreez: I think that's probably the right direction, though we need to be careful of too many errors littering the logs. But still, if we're getting errors that are log-worthy, I think we should log them.

Rather than a checkbox, might it make sense to use a select control that is something like the following?

[Error] **Log level**
         The level of logging that we will record in WooCommerce > Status > Logs. (We will always log errors and more serious items.)

And then have the options be `Error`, `Info`, and `Debug`.

daledupreez: Given the above, maybe that makes sense as a follow-up PR?

diegocurbelo: I like the idea, and I agree that we should implement it in a follow-up PR.

It might be confusing to only provide a select for the severity in the Debug mode section; we are using the WC Logger (and its level threshold), so when if the WC Log Level Threshold is set to Emergency (the highest) only those will end up in the log file regardless of the selected option in our plugin.
We can add an explanation, show the WC Log Threshold level value, and add a warning if it's above the selected level (nbothing will be logged) for example.

Screenshot 2025-07-10 at 18 04 54

PR #4474, opened 2025-07-09
Implementing product type constants

wjrosa:

Since WC 9.7 is our current minimum version, we can implement the new [constants for product types](https://github.com/woocommerce/woocommerce/blob/release/9.7/plugins/woocommerce/src/Enums/ProductType.php).

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (approved):

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -92,7 +95,7 @@ public function ajax_add_to_cart() {
 		// First empty the cart to prevent wrong calculation.
 		WC()->cart->empty_cart();
 
-		if ( ( 'variable' === $product_type || 'variable-subscription' === $product_type ) && isset( $_POST['attributes'] ) ) {
+		if ( ( ProductType::VARIABLE === $product_type || 'variable-subscription' === $product_type ) && isset( $_POST['attributes'] ) ) {
```
daledupreez, line 98: Very much non-blocking personal preference (which may be worth discussing as a team): I prefer using fully-qualified references inline as much as possible, mostly because it's not always easy to understand what the class is when viewing isolated sections of the code, especially via diffs.
```suggestion
		if ( ( Automattic\WooCommerce\Enums\ProductType::VARIABLE === $product_type || 'variable-subscription' === $product_type ) && isset( $_POST['attributes'] ) ) {
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.7.0 - xxxx-xx-xx =
+* Dev - Makes use of the new product type constants in the whole codebase
```
daledupreez: I think it may be worth clarifying the details a bit for the changelog entry, just so it's clearer what we're doing.
```suggestion
* Dev - Use product type constants that were added in WooCommerce 9.7
```

daledupreez (approved):
Changes look good to me! I have one suggestion for the changelog entry for clarity, and a non-blocking comment that is driven by a personal style preference so doesn't need action.

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
wjrosa: Gotcha! To add to this, I found a page with a nice discussion about it: https://www.sitepoint.com/community/t/using-namespace-prefix-versus-use-statement/351357/11

PR #4477, opened 2025-07-09
Fix No Such Customer error when adding a payment method

malithsen:
Fixes STRIPE-570

When a merchant switches Stripe accounts (e.g., moving from one Stripe account to another, or switching between test/live
environments), existing customer IDs stored in the db become invalid for the new account. The plugin was incorrectly
handling this scenario by passing `$is_retry = true` to the `update_customer()` method, which prevented the automatic customer recreation logic from executing.

- Removes the unused `$is_retry` parameter from the `update_or_create_customer()` method
- Ensures that when `update_customer()` encounters a "No such customer" error, it triggers the customer recreation logic.

## Testing instructions
1. As a customer, log in and navigate to **My Account → Payment Methods**
2. Add a new payment method
3. Verify the payment method is saved successfully
4. As a merchant, go to **WooCommerce → Settings → Payments → Stripe**
5. Disconnect the current Stripe account
6. Connect a different Stripe account (or reconnect with different test/live keys)
7. As the same customer from step 1, navigate to **My Account → Payment Methods → Add Payment Method**
8. Try to add another payment method
9. In **develop:** The request fails with "No such customer" error.
10. In **this branch:** The payment method should be added successfully.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -12,6 +12,7 @@
 * Tweak - Use the Database Cache for the Stripe Account Data
 * Update - Update filter names to use the wc_stripe_* prefix
 * Add - Show payment methods sync status on the UI
+* Fix - No such customer error when creating payment methods
 
 = 9.6.0 - 2025-07-07 =
 * Fix - Register Express Checkout script before use to restore buttons on “order-pay” pages
--- a/changelog.txt
+++ b/changelog.txt
@@ -12,7 +12,7 @@
 * Tweak - Use the Database Cache for the Stripe Account Data
 * Update - Update filter names to use the wc_stripe_* prefix
 * Add - Show payment methods sync status on the UI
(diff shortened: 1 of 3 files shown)


wjrosa (approved):
It is strange that we had that value hardcoded 🤔 . Anyway, working great!

Screenshot 2025-07-09 at 17 18 17

annemirasol (approved):
LGTM! 🚢 Was able to reproduce the issue and verify the fix.

Documenting something I observed, but out of scope for this PR -- after a Stripe account switch, old saved payment methods/tokens are no longer valid but are still shown for the user. When the user gets recreated, these invalid tokens disappear. EDIT: I will create an issue for this.

PR #4480, opened 2025-07-10
Fix wrong price formatting in Apple/Google Pay

### LINKED ISSUE #4479: Wrong price formatting in Apple/Google Pay
### Description
When the store currency is a 2-decimal currency but the number of decimals is set as `0` in the WooCommerce settings, Apple/Google Pay displays the wrong amount on the modal.

We transform the price [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/client/express-checkout/transformers/wc-to-stripe.js#L19) to express checkout format. The decimal value sent from [the backend](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/payment-methods/class-wc-stripe-express-checkout-helper.php#L326) is considering the WooCommerce decimal. This is causing the issue in the calculation when `0` is set in the setting for a 2 decimal currency.

### Steps to reproduce
- Go to **WooCommerce > Settings > General** and scroll to **Currency options**
- Set `USD` as the store currency
- Set `0` as the **Number of decimals**
- Enable express checkout in your Stripe settings page
- As a shopper, add a product priced $18 to your cart and go to the checkout page.
- Notice that, Google/Apple Pay is not present
- Go to your cart and increase the number of products to 5 so that the cart total becomes $90.
- Notice that,  Google/Apple Pay is now present.
- Click the express checkout button and notice that the displayed price is wrong. 

### Screenshots

**Order Total**
Image

**Price on Google Pay modal**
Image

### Expected behavior
- The displayed and charged price should be correct.

### Reports
1. https://wordpress.org/support/topic/issue-with-price-formatting-and-apple-pay-after-recent-stripe-plugin-update/
2. 9977884-zd-a8c

Mayisha:
Fixes STRIPE-573
Fixes #4479 

- Returning Stripe's expected currency decimal value in `currency_decimals` instead of what's saved in the WooCommerce settings.

## Testing instructions

- Go to **WooCommerce > Settings > General** and scroll to Currency options
- Set `USD` as the store currency
- Set `0` as the **Number of decimals**
- Enable express checkout in your Stripe settings page
- As a shopper, add a product priced $18 to your cart and go to the checkout page.
    - In `develop`, notice that Google/Apple Pay is not present
    - In this branch the Google/Apple Pay button should be present
- Go to your cart and increase the number of products to 5 so that the cart total becomes $90.
    - In `develop`, notice that, Google/Apple Pay is now present on the checkout page. But when you click the express checkout button, notice that the displayed price is wrong.
    - In this branch, the Google/Apple pay button should be present and display the correct amount.
- Now set the store currency to `JPY` and confirm that the price is correct on Google/Apple Pay and Link modals for JPY.

### DIFF
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -306,7 +306,7 @@ class WC_Stripe_Helper {
 	 *
 	 * @return array $currencies
 	 */
-	private static function three_decimal_currencies() {
+	public static function three_decimal_currencies() {
 		return [
 			'bhd', // Bahraini Dinar
 			'jod', // Jordanian Dinar
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -315,6 +315,22 @@ class WC_Stripe_Express_Checkout_Helper {
 	}
 
 	/**
+	 * Get the number of decimals supported by Stripe for the currency.
+	 *
+	 * @return int
+	 */
+	public static function get_stripe_currency_decimals() {
+		$currency = get_woocommerce_currency();
+		if ( in_array( $currency, WC_Stripe_Helper::no_decimal_currencies(), true ) ) {
+			return 0;
+		} elseif ( in_array( $currency, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
+			return 3;
+		} else {
+			return 2;
+		}
+	}
+
+	/**
 	 * JS params data used by cart and checkout pages.
 	 *
 	 * @param array $data
@@ -323,7 +339,7 @@ class WC_Stripe_Express_Checkout_Helper {
 		$data = [
 			'url'                     => wc_get_checkout_url(),
 			'currency_code'           => strtolower( get_woocommerce_currency() ),
-			'currency_decimals'       => wc_get_price_decimals(),
+			'currency_decimals'       => $this->get_stripe_currency_decimals(),
 			'country_code'            => substr( get_option( 'woocommerce_default_country' ), 0, 2 ),
 			'needs_shipping'          => 'no',
 			'needs_payer_phone'       => 'required' === get_option( 'woocommerce_checkout_phone_field', 'required' ),
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -320,7 +320,7 @@ class WC_Stripe_Express_Checkout_Helper {
 	 * @return int
 	 */
 	public static function get_stripe_currency_decimals() {
-		$currency = get_woocommerce_currency();
+		$currency = strtolower( get_woocommerce_currency() );
 		if ( in_array( $currency, WC_Stripe_Helper::no_decimal_currencies(), true ) ) {
 			return 0;
 		} elseif ( in_array( $currency, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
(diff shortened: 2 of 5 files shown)


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -314,6 +314,22 @@ public function get_allowed_shipping_countries() {
 		return array_keys( $allowed_shipping_countries );
 	}
 
+	/**
+	 * Get the number of decimals supported by Stripe for the currency.
+	 *
+	 * @return int
+	 */
+	public static function get_stripe_currency_decimals() {
```
wjrosa, line 322: May be worth it adding a simple unit test for this method.

includes/class-wc-stripe-helper.php
```diff
@@ -306,7 +306,7 @@ public static function no_decimal_currencies() {
 	 *
 	 * @return array $currencies
 	 */
-	private static function three_decimal_currencies() {
+	public static function three_decimal_currencies() {
```
wjrosa, line 309: Now that this method is public, might be good to unit test it as well.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -314,6 +314,22 @@ public function get_allowed_shipping_countries() {
...
+		$currency = get_woocommerce_currency();
+		if ( in_array( $currency, WC_Stripe_Helper::no_decimal_currencies(), true ) ) {
+			return 0;
+		} elseif ( in_array( $currency, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
+			return 3;
+		} else {
```
wjrosa: Totally optional: I am not a fan of `else`, usually it is not needed (like here). I would consider removing it

wjrosa (approved):
Left some optional comments. It works as expected, but when I attempt to add a product slightly cheaper than the $18 amount (the mocked beanie product with discount) I get the error below in console and the buttons are not present:

Screenshot 2025-07-14 at 10 42 03

Not directly related I guess, but might be worth fixing it if that's something simple to do here.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -314,6 +314,22 @@ public function get_allowed_shipping_countries() {
...
 	}
 
+	/**
+	 * Get the number of decimals supported by Stripe for the currency.
+	 *
+	 * @return int
+	 */
+	public static function get_stripe_currency_decimals() {
+		$currency = get_woocommerce_currency();
+		if ( in_array( $currency, WC_Stripe_Helper::no_decimal_currencies(), true ) ) {
```
malithsen, line 324: Should we convert `$currency` to lowercase to be safe?

In my test store `get_woocommerce_currency()` returns the currency in uppercase, but currencies in our helper functions are in lowercase.

includes/class-wc-stripe-helper.php
Mayisha: I will skip adding test for this one as this function has no logic and only returns an array.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
Mayisha: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4480/commits/ec499636306b9b8c4bc7807a2e1042386e858576

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
Mayisha: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4480/commits/0b9ad13e9966f4be01012e4ced95955bd377d542

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
Mayisha: Removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4480/commits/c7ffa71c19130d5675225a9ada5ee757eff5039e

malithsen (approved):
Thanks Mayisha, this looks good to me 🚢

PR #4257, opened 2025-04-24
Fix edge case for webhook comparisons

daledupreez:
Related to:
 * STRIPE-359
 * https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4116
 * https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4250

## Changes proposed in this Pull Request:

This PR is a follow-up to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4250 and addresses an [edge case flagged by](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4250#pullrequestreview-2790542675) @Mayisha: the base URL of the form `https://example.com/` was not being handled correctly. As I [discovered](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4250#issuecomment-2827961922), the issue was due to us calling `untrailingslashit()` on both URLs, which would cause the path for URLs of the form above to be `''` as opposed to `'/'` for the case with a query string.

`untrailingslashit()` will only mangle URLs and result in mismatched URL paths, so I think we should remove its use altogether.

## Testing instructions

Inspection and verifying the unit test should be sufficient for this one.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [N/A] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [N/A] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [N/A] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [N/A] Included this PR in the Release Thread scope (or does not apply)

PR #4484, opened 2025-07-10
Show number of pending webhooks in the warning section of the Account status

### LINKED ISSUE #1225: Consider displaying the number of pending webhooks
Building on https://github.com/woocommerce/woocommerce-gateway-stripe/issues/1200, we could expose the [`pending_webhooks`](https://stripe.com/docs/api/events/object#event_object-pending_webhooks) count that Stripe includes with each webhook request, as a potentially useful enhancement.

My initial assumption was that this was a count of events that had failed or are otherwise yet to be delivered, but I haven't seen this number below 1, so we should make sure we understand and can communicate it properly before displaying it.

diegocurbelo:
Fixes STRIPE-27
Fixes #1225

This PR adds the number of pending webhooks to the warning message displayed when a webhook processing error occurs

It also relates to STRIPE-372, as it detects unexpected configuration problems, and with the new automatic webhook configuration during account connection (and the `Reconfigure webhooks` button in the connection modal it can be fixed from the UI. 

**NORMAL / NO WEBHOOK ERRORS**
Screenshot 2025-07-10 at 17 18 19

**WEBHOOK ERRORS**
image-2

## Testing instructions

0. Make sure the webhooks are properly configured
1. Change the webhook secret ('test_webhook_secret') in the settings to an invalid value (using the Dev Tools or WP-CLI)
    `wp option patch update woocommerce_stripe_settings test_webhook_secret 'whsec_INVALID'`
2. In the plugin settings, enable CashApp
3. Add a product to the cart, and go to the `Checkout page`
4. Select `CashApp` as the payment method, and click `Place Order` (no need to complete it, loading the modal is enough to trigger a couple of webhooks)
5. In the plugin settings, go to the `Settings` tab, and check that the warning message is displayed, and it includes the pending webhooks:
    Screenshot 2025-07-10 at 18 20 41

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -100,6 +100,8 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 
 		WC_Stripe_Logger::debug( 'Webhook received: ' . $event_type );
 
+		WC_Stripe_Webhook_State::set_pending_webhooks_count( $event->pending_webhooks );
+
 		// Validate it to make sure it is legit.
 		$request_headers   = array_change_key_case( $this->get_request_headers(), CASE_UPPER );
 		$validation_result = $this->validate_request( $request_headers, $request_body );
--- a/includes/class-wc-stripe-webhook-state.php
+++ b/includes/class-wc-stripe-webhook-state.php
@@ -272,9 +272,11 @@ class WC_Stripe_Webhook_State {
 
 		$date_format = 'Y-m-d H:i:s e';
 
+		$meesage = '';
+
 		switch ( $code ) {
 			case 1: // Case 1 (Nominal case): Most recent = success
-				return sprintf(
+				$meesage = sprintf(
 					$test_mode ?
 						/* translators: 1) date and time of last webhook received, e.g. 2020-06-28 10:30:50 UTC */
 						__( 'The most recent test webhook, timestamped %s, was processed successfully.', 'woocommerce-gateway-stripe' ) :
@@ -282,8 +284,9 @@ class WC_Stripe_Webhook_State {
 						__( 'The most recent live webhook, timestamped %s, was processed successfully.', 'woocommerce-gateway-stripe' ),
 					gmdate( $date_format, $last_success_at )
 				);
+				break;
 			case 2: // Case 2: No webhooks received yet
-				return sprintf(
+				$meesage = sprintf(
 					$test_mode ?
 						/* translators: 1) date and time webhook monitoring began, e.g. 2020-06-28 10:30:50 UTC */
 						__( 'No test webhooks have been received since monitoring began at %s.', 'woocommerce-gateway-stripe' ) :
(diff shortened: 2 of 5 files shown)


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -100,6 +100,8 @@ public function check_for_webhook() {
 
 		WC_Stripe_Logger::debug( 'Webhook received: ' . $event_type );
 
+		WC_Stripe_Webhook_State::set_pending_webhooks_count( $event->pending_webhooks );
```
annemirasol, line 101: I wonder if we should be using `$event->pending_webhooks` here, instead of manually incrementing the current count by one.

In my setup, I had three webhook endpoints registered (from old setups), so I got "3 pending webhooks" when I really only had 1, for `payment_intent.requires_action`.

Screenshot 2025-07-10 at 11 21 43 PM

diegocurbelo: We can't currently keep track locally, it would always be 1 (the current event) as we only return an error when [the signature doesn't match and the account has duplicate webhooks setup for the site](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/abed6cd0feb94eb1e3833a16ecdffe1e18fe40d0/includes/class-wc-stripe-webhook-handler.php#L127).

Also, Stripe retries failed attempts, so we'd need to keep track of every failed event to prevent double-counting.

The short term objective is to add additional information to help detect (or confirm) unexpected problems; for example, when the Stripe account has multiple webhook configurations pointing to the same store, similar to the [issue solved here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4257).

[STRIPE-372](https://linear.app/a8c/issue/STRIPE-372) proposes further improvements, including returning errors in additional cases (not only duplicate webhook config) and automatically attempting to re-configure the webhook integration.

includes/class-wc-stripe-webhook-state.php
```diff
@@ -269,34 +298,40 @@ public static function get_webhook_status_message() {
 						 * translators: 1) date and time of last failed webhook e.g. 2020-06-28 10:30:50 UTC
 						 * translators: 2) reason webhook failed
 						 * translators: 3) date and time of last successful webhook e.g. 2020-05-28 10:30:50 UTC
+						 * translators: 4) number of pending webhooks
 						 */
-						__( 'Warning: The most recent test webhook, received at %1$s, could not be processed. Reason: %2$s. (The last test webhook to process successfully was timestamped %3$s.)', 'woocommerce-gateway-stripe' ) :
+						__( 'Warning: The most recent test webhook, received at %1$s, could not be processed. Reason: %2$s. (The last test webhook to process successfully was timestamped %3$s, and there are approximately %4$d webhooks pending)', 'woocommerce-gateway-stripe' ) :
```
daledupreez: As written, this approach results in a big block of hard-to-parse text, and has edge cases with weird wording for the warning, e.g. `and there are approximately 1 webhooks pending`. I think it may be clearer to split the wording into a new separate sentence that uses `_n()` and is only shown when `$pending_webhooks > 0`.

Also as a much smaller nit, we are missing punctuation before the closing `)` in the new message.
```suggestion
						__( 'Warning: The most recent test webhook, received at %1$s, could not be processed. Reason: %2$s. (The last test webhook to process successfully was timestamped %3$s.)', 'woocommerce-gateway-stripe' ) :
```

And then later on we have code that adds something like:
```php
					. ( $pending_webhooks < 1 ? '' :
						' ' . _n(
							'There is at least %d webhook pending.',
							'There are approximately %d webhooks pending.',
							$pending_webhooks
						);

**Review by daledupreez — COMMENTED** [review]
I haven't tested yet, but I have some feedback based on inspection. (Though I am fairly confident the implementation is solid.)

File: includes/class-wc-stripe-webhook-handler.php (hunk 1 above)
Comment (annemirasol) [reply includes/class-wc-stripe-webhook-handler.php:101]: Gotcha. I did not think about retries messing up manual counting!

**Review by annemirasol — APPROVED** [review]
Behaves as described, code LGTM!

File: includes/class-wc-stripe-webhook-state.php [hunk 3]
```diff
@@ -240,30 +268,35 @@ public static function get_webhook_status_message() {
...
+		$pending_webhooks    = self::get_pending_webhooks_count();
 
 		$date_format = 'Y-m-d H:i:s e';
 
+		$meesage = '';
+
 		switch ( $code ) {
 			case 1: // Case 1 (Nominal case): Most recent = success
-				return sprintf(
+				$meesage = sprintf(
```
Mayisha:
```suggestion
				$message = sprintf(
```
Typo.

includes/class-wc-stripe-webhook-state.php
```diff
@@ -279,10 +312,12 @@ public static function get_webhook_status_message() {
 						__( 'Warning: The most recent live webhook, received at %1$s, could not be processed. Reason: %2$s. (The last live webhook to process successfully was timestamped %3$s.)', 'woocommerce-gateway-stripe' ),
 					gmdate( $date_format, $last_failure_at ),
 					$last_error,
-					gmdate( $date_format, $last_success_at )
+					gmdate( $date_format, $last_success_at ),
+					$pending_webhooks,
```
Mayisha: Is this needed anymore?

includes/class-wc-stripe-webhook-state.php
```diff
@@ -296,9 +331,25 @@ public static function get_webhook_status_message() {
 						__( 'Warning: The most recent live webhook, received at %1$s, could not be processed. Reason: %2$s. (No live webhooks have been processed successfully since monitoring began at %3$s.)', 'woocommerce-gateway-stripe' ),
 					gmdate( $date_format, $last_failure_at ),
 					$last_error,
-					gmdate( $date_format, $monitoring_began_at )
+					gmdate( $date_format, $monitoring_began_at ),
+					$pending_webhooks,
```
Mayisha: Same as the previous comment. This `$pending_webhooks` seems unnecessary.

Mayisha (approved):
Looks good to me. Found two little issues, one typo and one unnecessary argument that need to be fixed.

Screenshot 2025-07-15 at 3 06 44 PM

PR #4485, opened 2025-07-11
Fix wc_gateway_stripe_process_payment action not called for sync payments

diegocurbelo:
Fixes STRIPE-576

This PR deprecates the actions `wc_gateway_stripe_process_payment`, `wc_gateway_stripe_process_redirect_payment` and `wc_gateway_stripe_process_webhook_payment` and introduces a new one `wc_gateway_stripe_process_payment_charge` called from inside the `process_response()` method, that is called for all payments (synchronous and asynchronous)

## Testing instructions

1. Add the following snippet:
    ```
	add_action( 'wc_gateway_stripe_process_payment_charge', 'custom_info', 10, 2);
	function custom_info( $charge, $order ) {
		$logger = wc_get_logger();
		$logger->info( wc_print_r( $charge, true ), array( 'source' => 'ks-stripe' ) );
	}
    ```
2. Add a product to the cart, and go to the checkout
3. Select Credit Card as the payment method and use the card 4242424242424242
4. Click `Place Order`
5. Open the plugin logs (WooCommerce > Status > Logs, and select `test-stripe`
6. Check that the action was called:
    Screenshot 2025-07-11 at 15 37 29

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,7 +18,7 @@
 * Fix - No such customer error when creating a payment method with a new Stripe account
 * Fix - Validate create customer payload against required billing fields before sending to Stripe
 * Update - Enhanced logging system with support for all log levels and improved context handling
-* Fix - Fixes wc_gateway_stripe_process_payment action not being called for synchronous payments
+* Update - Depreacte `wc_gateway_stripe_process_payment` action in favor of `wc_gateway_stripe_process_payment_charge`
 
 = 9.6.0 - 2025-07-07 =
 * Fix - Register Express Checkout script before use to restore buttons on “order-pay” pages
--- a/changelog.txt
+++ b/changelog.txt
@@ -18,7 +18,7 @@
 * Fix - No such customer error when creating a payment method with a new Stripe account
 * Fix - Validate create customer payload against required billing fields before sending to Stripe
 * Update - Enhanced logging system with support for all log levels and improved context handling
-* Update - Depreacte `wc_gateway_stripe_process_payment` action in favor of `wc_gateway_stripe_process_payment_charge`
+* Update - Depreacte `wc_gateway_stripe_process_payment` action in favour of `wc_gateway_stripe_process_payment_charge`
 
 = 9.6.0 - 2025-07-07 =
 * Fix - Register Express Checkout script before use to restore buttons on “order-pay” pages
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@
 * Update - Update filter names to use the wc_stripe_* prefix
 * Add - Show payment methods sync status on the UI
 * Fix - No such customer error when creating a payment method with a new Stripe account
+* Fix - Fixes wc_gateway_stripe_process_payment action not being called for synchronous payments
 
 = 9.6.0 - 2025-07-07 =
 * Fix - Register Express Checkout script before use to restore buttons on “order-pay” pages
--- a/includes/compat/trait-wc-stripe-subscriptions.php
+++ b/includes/compat/trait-wc-stripe-subscriptions.php
@@ -571,6 +571,14 @@ trait WC_Stripe_Subscriptions_Trait {
 				do_action( 'wc_gateway_stripe_process_payment_subscription_charge_attempt_delayed', $response, $renewal_order );
 			} else {
 				// The charge was successfully captured
+				do_action_deprecated(
+					'wc_gateway_stripe_process_payment',
+					[ $response, $renewal_order ],
+					'9.7.0',
+					'wc_gateway_stripe_process_payment_charge',
+					'The wc_gateway_stripe_process_payment action is deprecated. Use wc_gateway_stripe_process_payment_charge instead.'
+				);
+
 				// Use the last charge within the intent or the full response body in case of SEPA.
 				$latest_charge = $this->get_latest_charge_from_intent( $response );
 				$this->process_response( ( ! empty( $latest_charge ) ) ? $latest_charge : $response, $renewal_order );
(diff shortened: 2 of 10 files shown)


tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
```diff
@@ -590,7 +601,7 @@ public function provide_test_process_payment_intent() {
 				'order locked'                   => false,
 				'payment type'                   => WC_Stripe_Payment_Methods::BOLETO,
 				'order status final'             => false,
-				'expected status'                => OrderStatus::PENDING,
+				'expected status'                => OrderStatus::PROCESSING,
```
diegocurbelo, line 605: The `expected_status` was wrong; it should be `processing` when the intent succeeds.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -571,8 +571,6 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
 				do_action( 'wc_gateway_stripe_process_payment_subscription_charge_attempt_delayed', $response, $renewal_order );
 			} else {
 				// The charge was successfully captured
-				do_action( 'wc_gateway_stripe_process_payment', $response, $renewal_order );
```
annemirasol, line 574: For a renewal flow (wp-admin > Subscriptions > Process renewal), `$response` here is a `payment_intent` object. 

With the proposed changes (moving the action to `process_payment()`), `$response` becomes a charge object. I am worried it will break custom stuff that assume they will get the payment intent object.

annemirasol (commented):
Added a concern about payment intent vs charge for the subscriptions hook.

Lesser concern: If we are deprecating `wc_gateway_stripe_process_[redirect|webhook]_payment`, should we also deprecate their error counterparts, `wc_gateway_stripe_process_[redirect|webhook]_payment_error`?

includes/compat/trait-wc-stripe-subscriptions.php
diegocurbelo: Nice catch 👏🏼 ... I originally moved it inside the `process_response()` function to avoid forgetting to call the action manually before each call to `process_response()`.

A better solution would be to deprecate all `wc_gateway_stripe_process_payment` calls, and add a new filter `wc_gateway_stripe_process_payment_charge` inside `process_response()`, explicitly adding in the doc section that it expects a charge.

changelog.txt
```diff
@@ -18,6 +18,7 @@
 * Fix - No such customer error when creating a payment method with a new Stripe account
 * Fix - Validate create customer payload against required billing fields before sending to Stripe
 * Update - Enhanced logging system with support for all log levels and improved context handling
+* Update - Depreacte `wc_gateway_stripe_process_payment` action in favour of `wc_gateway_stripe_process_payment_charge`
```
```suggestion
* Update - Depreacte `wc_gateway_stripe_process_[redirect|webhook]_payment` action in favour of `wc_gateway_stripe_process_payment_charge`
```

This way or separately, I think we should log all the deprecated webhooks in the changelog.

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
```diff
@@ -1459,6 +1459,45 @@ public function test_process_response_updates_order_by_charge_status() {
 		$this->assertMatchesRegularExpression( '/Payment processing failed./', $exception->getLocalizedMessage() );
 	}
 
+	/**
+	 * Test that the wc_gateway_stripe_process_payment_charge action is triggered when process_response() is called for synchronous payment paths.
+	 */
+	public function test_process_response_triggers_wc_gateway_stripe_process_payment_action() {
```
Mayisha:
```suggestion
	public function test_process_response_triggers_wc_gateway_stripe_process_payment_charge_action() {
```

Mayisha (approved):
The new hook is working perfectly 👍 

I added the following snippet from the PR description and got the log file after I placed an order withna card.

```
add_action( 'wc_gateway_stripe_process_payment_charge', 'custom_info', 10, 2);
function custom_info( $charge, $order ) {
	$logger = wc_get_logger();
	$logger->info( wc_print_r( $charge, true ), array( 'source' => 'ks-stripe' ) );
}
```

Screenshot 2025-07-15 at 3 30 24 PM


❓ I added the following snippet for the deprecated action. It was triggered when I placed an order via CashApp but not when I placed an order via Card. Is this expected?

```
add_action( 'wc_gateway_stripe_process_payment', 'custom_info2', 10, 2);
function custom_info2( $charge, $order ) {
	$logger = wc_get_logger();
	$logger->info( wc_print_r( $charge, true ), array( 'source' => 'pm-stripe' ) );
}
```


> [!Note]
We should add the new action and the deprecated ones in the [wiki](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Action-and-Filter-Hooks).

diegocurbelo:
> ❓ I added the following snippet for the deprecated action. It was triggered when I placed an order via CashApp but not when I placed an order via Card. Is this expected?

Yes, the `wc_gateway_stripe_process_payment` action is not being called for sync payments. We also found that there are two other actions, `wc_gateway_stripe_process_redirect_payment` and `wc_gateway_stripe_process_webhook_payment`, that serve the same purpose, and one of these calls receives a payment intent, not the charge (thanks @annemirasol 🙇🏼‍♂️ ).

So, this PR deprecates them and introduces a new action `wc_gateway_stripe_process_payment_charge` that is called for all payments with consistent params.

annemirasol (approved):
LGTM! 🚢 

Smoke-tested the new action, and a couple of the deprecated ones.

PR #4486, opened 2025-07-11
Improve Stripe API Connector logging

diegocurbelo:

This PR updated the Stripe API connector (which generates more than 90% of the log entries) to utilize the new logging methods, incorporating additional context for the request and response payloads.

It intentionally keeps the same severity (debug) for all entries, except error logs that were changed to `error` severity. Making the entries easier to read (and search) by showing just one line per log (with an expandable context), and as a result, error entries are now easily identifiable:

image

Screenshot 2025-07-15 at 14 35 27

Screenshot 2025-07-15 at 15 05 12

## Testing instructions

The change is straightforward, and a code review is sufficient.

_To manually test it:_
1. Go to WP-Admin and then to the plugin settings, and enable/disable some payment methods and save changes.
2. Go to the Store, add a product to the cart, and place the order using any payment method.
3. Replace the `secret_key` with an invalid value
    `wp option patch update woocommerce_stripe_settings test_secret_key 'sk_test_INVALID`
4. Reload the settings page
5. Go to WooCommerce > Status > Logs and select `woocommerce-gateway-stripe`

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-api.php
```diff
@@ -227,34 +227,33 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
...
-					],
-					true
-				)
+			// Stripe redacts API keys in the response.
+			WC_Stripe_Logger::error(
+				"Stripe API error: {$method} {$api}",
+				[
+					'request'         => $request,
+					'idempotency_key' => $idempotency_key,
+					'response'        => ! empty( $response['body'] ) ? json_decode( $response['body'] ) : $response,
```
diegocurbelo: We can also write this in a shorter and "more readable" way:
```suggestion
					'response'        => json_decode( $response['body'] ?? '' ) ?? $response,
```

includes/class-wc-stripe-api.php
```diff
@@ -214,7 +214,7 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
 		$request = apply_filters( 'wc_stripe_request_body', $request, $api );
 
 		// Log the request after the filters have been applied.
-		WC_Stripe_Logger::log( "{$api} request: " . print_r( $request, true ) );
+		WC_Stripe_Logger::debug( "Stripe API request: {$method} {$api}", [ 'request' => $request ] );
```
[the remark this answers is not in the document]
diegocurbelo: We are including it in case of error, but it adds no value when the call succeeds.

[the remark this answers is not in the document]
diegocurbelo: Makes sense, we should handle this in the `WC_Stripe_Logger` itself; I created a follow-up issue STRIPE-627

includes/class-wc-stripe-api.php
```diff
@@ -227,34 +227,33 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
 		);
 
 		$response_headers = wp_remote_retrieve_headers( $response );
-		// Log the stripe version in the response headers, if present.
-		if ( isset( $response_headers['stripe-version'] ) ) {
```
wjrosa, line 253: Just to confirm this removal was intentional (I imagine this information is not that useful anymore).

wjrosa (approved):
Nice job! Looks great and works as intended ✅ . Left a double-checking comment

includes/class-wc-stripe-api.php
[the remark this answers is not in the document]
daledupreez: Might it be worth logging this with a key of `request_body` instead of just `request`? It's minor, but it more accurately reflects what we are logging.
```suggestion
		// Log the request after the filters have been applied.
		WC_Stripe_Logger::debug( "Stripe API request: {$method} {$api}", [ 'request_body' => $request ] );
```

includes/class-wc-stripe-api.php
```diff
@@ -363,15 +376,14 @@ public static function request_with_level3_data( $request, $api, $level3_data, $
...
-				. print_r( $order->get_items(), true ) . PHP_EOL
-				. print_r( 'Order shipping amount: ', true ) . PHP_EOL
-				. print_r( $order->get_shipping_total(), true ) . PHP_EOL
-				. print_r( 'Order currency: ', true ) . PHP_EOL
-				. print_r( $order->get_currency(), true )
+			WC_Stripe_Logger::error(
+				'Level3 data sum incorrect',
+				[
+					'error'         => $result->error,
+					'order_line_items'      => $order->get_items(),
```
daledupreez: Nit: the alignment here is slightly off.
```suggestion
					'error'                 => $result->error,
					'order_line_items'      => $order->get_items(),
```

includes/class-wc-stripe-api.php
```diff
@@ -227,34 +227,33 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
...
+					'response'        => $response,
+				]
 			);
 
 			throw new WC_Stripe_Exception( print_r( $response, true ), __( 'There was a problem connecting to the Stripe API endpoint.', 'woocommerce-gateway-stripe' ) );
 		}
 
+		$response_body = json_decode( $response['body'] );
+
+		WC_Stripe_Logger::debug( "Stripe API response: {$method} {$api}", [ 'response' => $response_body ] );
```
daledupreez, line 269: Following on from @wjrosa's [comment above](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4486/files#r2220208873), should we not log the response headers here as well?
```suggestion
		WC_Stripe_Logger::debug(
			"Stripe API response: {$method} {$api}",
			[
				'response_body'    => $response_body,
				'response_headers' => $response_headers,
			]
		);
```

includes/class-wc-stripe-api.php
diegocurbelo: Correct, this log entry is redundant; we always send the `stripe-version` in the headers (and was added to the webhook handling with the webhook autoconfiguration)

daledupreez (approved):
The changes look good to me and test well. I have some minor, non-blocking feedback.

includes/class-wc-stripe-api.php
diegocurbelo: The headers do not contain any helpful information (even for errors); that's why it was not logged before. But I don't have a strong opinion about that, except that it adds up to the log size 🤷🏼

daledupreez: That's fair -- if they don't include useful data, then we should not log them, as they are effectively noise.

PR #4491, opened 2025-07-14
Fix rate limit for failed add payment method attempts

diegocurbelo:
Fixes STRIPE-582

This PR fixes the rate-limiting check when adding a new payment method.

## Testing instructions

1. Go to the `My account` page in your `Store`
2. Click the `Payment methods` option in the sidebar
3. Click `Add payment method`
4. Select `Credit / Debit Card` and try to add the card `4000 0000 0000 0002`
5. Check that the error message `Your card was declined.` is displayed:
    Screenshot 2025-07-14 at 15 23 44
6. Re-try to add the same card within 10 seconds (click the button again)
7. Check that the error message is now `You cannot add a new payment method so soon after the previous one.`: 
    Screenshot 2025-07-14 at 15 23 58

**Test the custom delay filter:**
1. Add the following filter:
    ```
    add_filter( 'wc_stripe_add_payment_method_on_error_rate_limit_delay', 'wc_stripe_custom_test', 10, 2 );
    function wc_stripe_custom_test( $rate_limit_delay, $exception ) {
        return 1;
    }
    ```
2. Try to add the same declined card as before (`4000 0000 0000 0002`)
3. Check that the delay is now 1 second

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-intent-controller.php
+++ b/includes/class-wc-stripe-intent-controller.php
@@ -1116,9 +1116,10 @@ class WC_Stripe_Intent_Controller {
 	 * @throws Exception If the AJAX request is missing the required data or if there's an error creating and confirming the setup intent.
 	 */
 	public function create_and_confirm_setup_intent_ajax() {
+		$wc_add_payment_method_rate_limit_id = 'add_payment_method_' . get_current_user_id();
+
 		try {
 			// similar rate limiter is present in WC Core, but it's executed on page submission (and not on AJAX calls).
-			$wc_add_payment_method_rate_limit_id = 'add_payment_method_' . get_current_user_id();
 			if ( WC_Rate_Limiter::retried_too_soon( $wc_add_payment_method_rate_limit_id ) ) {
 				throw new WC_Stripe_Exception( 'Failed to save payment method.', __( 'You cannot add a new payment method so soon after the previous one.', 'woocommerce-gateway-stripe' ) );
 			}
@@ -1173,6 +1174,17 @@ class WC_Stripe_Intent_Controller {
 			);
 		} catch ( WC_Stripe_Exception $e ) {
 			WC_Stripe_Logger::log( 'Failed to create and confirm setup intent. ' . $e->getMessage() );
+
+			/**
+			 * Filter the rate limit delay when adding a payment method.
+			 *
+			 * @since 9.7.0
+			 *
+			 * @param int $rate_limit_delay The rate limit delay in seconds.
+			 */
+			$rate_limit_delay = apply_filters( 'wc_stripe_add_payment_method_rate_limit_delay', 10 );
+
+			WC_Rate_Limiter::set_rate_limit( $wc_add_payment_method_rate_limit_id, $rate_limit_delay );
 
 			// Send back error so it can be displayed to the customer.
 			wp_send_json_error(
--- a/includes/class-wc-stripe-intent-controller.php
+++ b/includes/class-wc-stripe-intent-controller.php
@@ -1176,13 +1176,14 @@ class WC_Stripe_Intent_Controller {
 			WC_Stripe_Logger::log( 'Failed to create and confirm setup intent. ' . $e->getMessage() );
 
 			/**
-			 * Filter the rate limit delay when adding a payment method.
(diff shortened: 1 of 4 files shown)


includes/class-wc-stripe-intent-controller.php
```diff
@@ -1174,6 +1175,17 @@ public function create_and_confirm_setup_intent_ajax() {
...
+			/**
+			 * Filter the rate limit delay when adding a payment method.
+			 *
+			 * @since 9.7.0
+			 *
+			 * @param int $rate_limit_delay The rate limit delay in seconds.
+			 */
+			$rate_limit_delay = apply_filters( 'wc_stripe_add_payment_method_rate_limit_delay', 10 );
+
+			WC_Rate_Limiter::set_rate_limit( $wc_add_payment_method_rate_limit_id, $rate_limit_delay );
```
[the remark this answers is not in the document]
diegocurbelo: In this case, we want to enforce the rate-limiter for all errors, but adding the exception to the filter params is a great idea, added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4491/commits/de670c4868077d0b123cd56a5968c680e934413d

includes/class-wc-stripe-intent-controller.php
```diff
@@ -1116,9 +1116,10 @@ public function create_and_confirm_setup_intent( $payment_information ) {
 	 * @throws Exception If the AJAX request is missing the required data or if there's an error creating and confirming the setup intent.
 	 */
 	public function create_and_confirm_setup_intent_ajax() {
```
wjrosa, line 1118: I am not sure if it is possible, but it might be worth it considering adding a unit test specific to rate limiting here.

diegocurbelo: I agree 100%, but we don't have any tests for ajax calls that use `wp_send_json_error`; I tried using `ob_start()` and `ob_get_clean()`, but it didn't work. We might need a refactoring to be able to test those methods.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -1174,6 +1175,18 @@ public function create_and_confirm_setup_intent_ajax() {
...
 
+			/**
+			 * Filter the rate limit delay after a failure adding a payment method.
+			 *
+			 * @since 9.7.0
+			 *
+			 * @param int $rate_limit_delay The rate limit delay in seconds.
+			 * @param WC_Stripe_Exception $e The exception that occurred.
+			 */
+			$rate_limit_delay = apply_filters( 'wc_stripe_add_payment_method_on_error_rate_limit_delay', 10, $e );
```
diegocurbelo, line 1186: I used 10 seconds here as the default used when success (20 seconds) felt too long, but we may want to change it to something lower, like 5 seconds.

includes/class-wc-stripe-intent-controller.php
wjrosa: Gotcha! No worries then

diegocurbelo: I found a workaround for the `wp_send_json_error` problem, using the filter `wp_die_ajax_handler` to avoid the `wp_die`; added tests in [639d3cd](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4491/commits/639d3cdf0207b8c8a38d7509564b2c9ec7b53fda)

wjrosa (approved):
Great fix! Works as expected:

> 5. Check that the error message Your card was declined. is displayed:

Screenshot 2025-07-14 at 17 46 37

> 7. Check that the error message is now You cannot add a new payment method so soon after the previous one.:

Screenshot 2025-07-14 at 17 46 45

> 3. Check that the delay is now 1 second

"Your card was declined." was shown when repeating the initial testing steps within 10 seconds 👍

includes/class-wc-stripe-intent-controller.php
wjrosa: Awesome! Thanks for adding those

Mayisha (approved):
The rate limiting is working perfectly 👍 

Screenshot 2025-07-15 at 5 11 49 PM

PR #4492, opened 2025-07-14
Disabling BNPLs when other official plugins are active

wjrosa:
Fixes STRIPE-431

To prevent Affirm and Klarna from being displayed alongside other official plugins, we are now removing them when the other plugins are active. Instead of checking for any of the plugins, we are disabling each corresponding method.

I am also adding some new helper methods specific to this logic to reduce duplicate code.

## Testing instructions

- Checkout to the `develop` branch on your test environment
- Connect a US Stripe account
- Install [Affirm](https://woocommerce.com/products/woocommerce-gateway-affirm/) or [Klarna](https://woocommerce.com/products/klarna-payments/) extensions (only one is enough)
- Confirm you can see Klarna and Affirm in the payment methods settings page (`wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=methods`)
- Enable both
- As a shopper, add at least 50 USD in products to your cart
- Go to the checkout page and confirm both methods are available
- Now checkout to this branch instead (`update/disable-bnpls-when-other-plugins-are-active`)
- As a shopper, confirm Klarna and Affirm are now gone from the payment methods settings page (depending on the official extension you installed, or both if you installed both)
- As a merchant, go to the checkout page and confirm that the same happened

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/components/payment-method-unavailable-due-conflict-pill/index.js
+++ b/client/components/payment-method-unavailable-due-conflict-pill/index.js
@@ -59,10 +59,7 @@ const PaymentMethodUnavailableDueConflictPill = ( { id, label } ) => {
 	) {
 		return (
 			<StyledPill>
-				{ __(
-					'Has plugin conflict',
-					'woocommerce-gateway-stripe'
-				) }
+				{ __( 'Has plugin conflict', 'woocommerce-gateway-stripe' ) }
 				<Popover
 					BaseComponent={ IconComponent }
 					content={ interpolateComponents( {
--- a/client/components/payment-method-unavailable-due-conflict-pill/index.js
+++ b/client/components/payment-method-unavailable-due-conflict-pill/index.js
@@ -1,0 +1,99 @@
+/* global wc_stripe_settings_params */
+import { __, sprintf } from '@wordpress/i18n';
+import React from 'react';
+import styled from '@emotion/styled';
+import interpolateComponents from 'interpolate-components';
+import { Icon, info } from '@wordpress/icons';
+import Popover from 'wcstripe/components/popover';
+import {
+	PAYMENT_METHOD_AFFIRM,
+	PAYMENT_METHOD_KLARNA,
+} from 'wcstripe/stripe-utils/constants';
+
+const StyledPill = styled.span`
+	display: inline-flex;
+	align-items: center;
+	gap: 4px;
+	padding: 4px 8px;
+	border: 1px solid #fcf9e8;
+	border-radius: 2px;
+	background-color: #fcf9e8;
+	color: #674600;
+	font-size: 12px;
+	font-weight: 400;
+	line-height: 16px;
+	width: fit-content;
+`;
+
+const StyledLink = styled.a`
+	&:focus,
+	&:visited {
+		box-shadow: none;
+	}
+`;
+
+const IconWrapper = styled.span`
+	height: 16px;
+	cursor: pointer;
+`;
+
+const AlertIcon = styled( Icon )`
+	fill: #674600;
+`;
+
+const IconComponent = ( { children, ...props } ) => (
+	<IconWrapper { ...props }>
+		<AlertIcon icon={ info } size="16" />
+		{ children }
+	</IconWrapper>
+);
+
+const PaymentMethodUnavailableDueConflictPill = ( { id, label } ) => {
+	if (
+		( id === PAYMENT_METHOD_AFFIRM &&
+			// eslint-disable-next-line camelcase
+			wc_stripe_settings_params.has_affirm_gateway_plugin ) ||
+		( id === PAYMENT_METHOD_KLARNA &&
+			// eslint-disable-next-line camelcase
+			wc_stripe_settings_params.has_klarna_gateway_plugin )
+	) {
+		return (
+			<StyledPill>
+				{ __(
+					'Unavailable due conflict',
+					'woocommerce-gateway-stripe'
+				) }
+				<Popover
+					BaseComponent={ IconComponent }
+					content={ interpolateComponents( {
+						mixedString: sprintf(
+							/* translators: $1: a payment method name */
+							__(
+								'%1$s is unavailable due to another official plugin being active.',
+								'woocommerce-gateway-stripe'
+							),
+							label
+						),
+						components: {
+							currencySettingsLink: (
+								<StyledLink
+									href="/wp-admin/admin.php?page=wc-settings&tab=general"
+									target="_blank"
+									rel="noreferrer"
+									onClick={ ( ev ) => {
+										// Stop propagation is necessary so it doesn't trigger the tooltip click event.
+										ev.stopPropagation();
+									} }
+								/>
+							),
+						},
+					} ) }
+				/>
+			</StyledPill>
+		);
+	}
+
+	return null;
+};
+
+export default PaymentMethodUnavailableDueConflictPill;
--- a/client/components/payment-method-unavailable-due-conflict-pill/index.js
+++ b/client/components/payment-method-unavailable-due-conflict-pill/index.js
@@ -60,7 +60,7 @@ const PaymentMethodUnavailableDueConflictPill = ( { id, label } ) => {
 		return (
 			<StyledPill>
 				{ __(
-					'Unavailable due conflict',
+					'Has plugin conflict',
 					'woocommerce-gateway-stripe'
 				) }
 				<Popover
--- a/includes/admin/class-wc-stripe-settings-controller.php
+++ b/includes/admin/class-wc-stripe-settings-controller.php
@@ -278,8 +278,8 @@ class WC_Stripe_Settings_Controller {
 			'is_oc_available'                       => WC_Stripe_Feature_Flags::is_oc_available(),
 			'oauth_nonce'                           => wp_create_nonce( 'wc_stripe_get_oauth_urls' ),
 			'is_sepa_tokens_enabled'                => 'yes' === $this->gateway->get_option( 'sepa_tokens_for_other_methods', 'no' ),
-			'has_affirm_gateway_plugin'             => WC_Stripe_Helper::has_gateway_plugin_active( 'affirm' ),
-			'has_klarna_gateway_plugin'             => WC_Stripe_Helper::has_gateway_plugin_active( 'klarna_payments' ),
+			'has_affirm_gateway_plugin'             => WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_AFFIRM ),
+			'has_klarna_gateway_plugin'             => WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_KLARNA ),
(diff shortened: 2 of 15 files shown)


includes/class-wc-stripe-helper.php
```diff
@@ -1851,4 +1851,20 @@ public static function get_refund_reason_description( $refund_reason_key ) {
 				return __( 'Unknown reason', 'woocommerce-gateway-stripe' );
 		}
 	}
+
+	/**
+	 * Checks if there are other Buy Now Pay Later plugins active.
+	 *
+	 * @return bool
+	 */
+	public static function has_other_bnpl_plugins_active() {
```
wjrosa, line 1874: Since we need this logic in multiple places, I moved it to a new helper function.

includes/notes/class-wc-stripe-bnpl-promotion-note.php
```diff
@@ -93,16 +93,7 @@ public static function init( WC_Stripe_Payment_Gateway $gateway ) {
...
-		$available_payment_gateways    = WC()->payment_gateways->payment_gateways;
-		$other_bnpl_gateway_ids        = [ 'affirm', 'klarna_payments' ];
-		foreach ( $available_payment_gateways as $available_payment_gateway ) {
-			if ( in_array( $available_payment_gateway->id, $other_bnpl_gateway_ids, true ) && 'yes' === $available_payment_gateway->enabled ) {
-				$has_other_bnpl_plugins_active = true;
-				break;
-			}
-		}
-		if ( $has_other_bnpl_plugins_active ) {
+		if ( WC_Stripe_Helper::has_other_bnpl_plugins_active() ) {
```
wjrosa, line 96: Making use of the new helper function.

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -287,7 +278,7 @@ public function admin_scripts( $hook_suffix ) {
 			'is_oc_available'                       => WC_Stripe_Feature_Flags::is_oc_available(),
 			'oauth_nonce'                           => wp_create_nonce( 'wc_stripe_get_oauth_urls' ),
 			'is_sepa_tokens_enabled'                => 'yes' === $this->gateway->get_option( 'sepa_tokens_for_other_methods', 'no' ),
-			'has_other_bnpl_plugins'                => $has_other_bnpl_plugins_active,
+			'has_other_bnpl_plugins'                => WC_Stripe_Helper::has_other_bnpl_plugins_active(),
```
wjrosa, line 283: Making use of the new helper function.

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -287,7 +278,9 @@ public function admin_scripts( $hook_suffix ) {
 			'is_oc_available'                       => WC_Stripe_Feature_Flags::is_oc_available(),
 			'oauth_nonce'                           => wp_create_nonce( 'wc_stripe_get_oauth_urls' ),
 			'is_sepa_tokens_enabled'                => 'yes' === $this->gateway->get_option( 'sepa_tokens_for_other_methods', 'no' ),
-			'has_other_bnpl_plugins'                => $has_other_bnpl_plugins_active,
+			'has_affirm_gateway_plugin'             => WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_AFFIRM ),
+			'has_klarna_gateway_plugin'             => WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_KLARNA ),
+			'has_other_bnpl_plugins'                => WC_Stripe_Helper::has_other_bnpl_plugins_active(),
```
wjrosa, line 283: Another option here would be to reuse the return of both `WC_Stripe_Helper::has_gateway_plugin_active`. Not sure if it is worth it.

diegocurbelo (changes requested):
After installing and enabling the `WooCommerce Gateway Affirm` plugin:

> As a shopper, confirm Klarna and Affirm are now gone from the payment methods settings page (depending on the official extension you installed, or both if you installed both)

This works as indicated; the Stripe payment method is no longer shown in the payment method list in the settings.

> As a merchant, go to the checkout page and confirm that the same happened

But it's still visible in the checkout. 

Image

--

Additionally, removing the Affirm/Klarna methods from the payment method list in the settings without providing any feedback may be confusing.

Maybe we can set it as disabled and add a yellow pill, similar to the one we use for "Requires activation"? 

Or show a notice or message explaining why Affirm/Klarna are not available in the list, or group them at the bottom under an 'Unavailable Payment Methods' group?

Mayisha:
> Additionally, removing the Affirm/Klarna methods from the payment method list in the settings without providing any feedback may be confusing.
Maybe we can set it as disabled and add a yellow pill, similar to the one we use for "Requires activation"?
Or show a notice or message explaining why Affirm/Klarna are not available in the list, or group them at the bottom under an 'Unavailable Payment Methods' group?

**+1** to this. Completely removing the payment methods could be confusing for the user. I like the idea of showing it disabled, preferably with a tooltip explaining the reason for it being disabled.

tests/phpunit/WC_Stripe_Helper_Test.php
```diff
@@ -190,408 +190,104 @@ public function test_get_legacy_enabled_payment_method_ids() {
...
 	 *
-	 * @param string $status              The order status to return.
-	 * @param bool   $success             Whether the order should be found.
+	 * @param array $payment_gateways The available payment gateways.
+	 * @param bool  $expected         The expected result.
+	 * @dataProvider provide_test_has_other_bnpl_plugins_active
 	 * @return void
-	 * @dataProvider provide_test_get_order_by_intent_id
 	 */
-	public function test_get_order_by_intent_id( $status, $success ) {
```
annemirasol: Did we mean to remove all these other tests on purpose? They don't look related to the PR to me -- checking that they weren't erased by accident.

includes/class-wc-stripe-helper.php
```diff
@@ -1851,4 +1865,35 @@ public static function get_refund_reason_description( $refund_reason_key ) {
...
+	/**
+	 * Checks if a given payment gateway plugin is active.
+	 *
+	 * @param string $plugin_id
+	 * @return bool
+	 */
+	public static function has_gateway_plugin_active( $plugin_id ) {
+		$available_payment_gateways = WC()->payment_gateways->payment_gateways ?? [];
+		foreach ( $available_payment_gateways as $available_payment_gateway ) {
+			if ( $plugin_id === $available_payment_gateway->id && 'yes' === $available_payment_gateway->enabled ) {
```
malithsen, line 1893: Not requesting a change here. There's a [is_plugin_active](https://developer.wordpress.org/reference/functions/is_plugin_active/) that we can use but it requires the plugin path. Since this is a simple check, checking the status like you do here is good.

wjrosa:
Thanks for the reviews, folks!

> Maybe we can set it as disabled and add a yellow pill, similar to the one we use for "Requires activation"?

Makes sense. I went for this option in my latest commits. Preview here:
Screenshot 2025-07-16 at 13 46 28

I have also fixed the methods availability in both the shortcode and block checkout.

tests/phpunit/WC_Stripe_Helper_Test.php
wjrosa: Oops! Sorry 🤦 . I removed them by mistake when isolating the new tests to make development faster. Added them back now

includes/class-wc-stripe-helper.php
wjrosa: Yes, I tried to use it before, but I couldn't make it work properly. If I remember correctly, I had issues trying to match the path. I can try again on another PR and consider deprecating this method later.

client/components/payment-method-unavailable-due-conflict-pill/index.js
```diff
@@ -0,0 +1,99 @@
...
+			// eslint-disable-next-line camelcase
+			wc_stripe_settings_params.has_affirm_gateway_plugin ) ||
+		( id === PAYMENT_METHOD_KLARNA &&
+			// eslint-disable-next-line camelcase
+			wc_stripe_settings_params.has_klarna_gateway_plugin )
+	) {
+		return (
+			<StyledPill>
+				{ __(
+					'Unavailable due conflict',
```
annemirasol:
```suggestion
					'Unavailable due to conflict',
```

annemirasol: Or if you want it short, maybe "Has plugin conflict"?

annemirasol (approved):
This looks good and works as expected. Left a suggestion about the wording in the pill.

Tested:

When official plugin is installed and active:
✅ Klarna unavailable in settings, and not shown in block/classic checkout
✅ Affirm unavailable in settings, and not shown in block/classic checkout

When official plugin is not enabled:
✅ Klarna available in settings, and shown in block/classic checkout if enabled
✅ Affirm available in settings, and shown in block/classic checkout if enabled

Noting our earlier discussion where in a future iteration, we may also disable Klarna/Affirm in the merchant's PMC. Right now, they are disabled for the store when a conflict exists, but they remain enabled in the PMC, even after a settings update.

annemirasol:
> Noting our earlier discussion where in a future iteration, we may also disable Klarna/Affirm in the merchant's PMC. Right now, they are disabled for the store when a conflict exists, but they remain enabled in the PMC, even after a settings update.

Actually, I want to double-check that I am understanding the earlier discussion correctly. This is the flow I'm wondering about:
1. Enable Klarna.
2. Install the other plugin and enable it.
3. Notice how Klarna is now unavailable and disabled in Stripe settings.
4. Enable or disable some other payment method, and click "Save changes".
5. Notice in your Stripe dashboard that Klarna is still enabled.

We are okay with this flow, and will update things for OC, is that correct?

client/components/payment-method-unavailable-due-conflict-pill/index.js
wjrosa: I think the short version is better. I will go with that instead 👍

wjrosa:
> > Noting our earlier discussion where in a future iteration, we may also disable Klarna/Affirm in the merchant's PMC. Right now, they are disabled for the store when a conflict exists, but they remain enabled in the PMC, even after a settings update.
> 
> Actually, I want to double-check that I am understanding the earlier discussion correctly. This is the flow I'm wondering about:
> 
> 1. Enable Klarna.
> 2. Install the other plugin and enable it.
> 3. Notice how Klarna is now unavailable and disabled in Stripe settings.
> 4. Enable or disable some other payment method, and click "Save changes".
> 5. Notice in your Stripe dashboard that Klarna is still enabled.
> 
> We are okay with this flow, and will update things for OC, is that correct?

Thanks for the review, Anne! Yes, that's my understanding as well

malithsen (approved):
Image

Looks good!

✅ Tested settings page with Klarna/Affirm official plugins active/inactive.
✅ Tested checkout pages (Blocks/Classic) with official plugins active/inactive.

PR #4495, opened 2025-07-15
Introducing a banner to promote OC

wjrosa:
Fixes STRIPE-546
Figma nFI8jCCAQnd19PNwScqVU6-fi
P2 changing the "Activate now" to "Learn more" pcUrMV-b1m-p2#comment-33344

We are introducing a new promotional banner surface to promote the Optimized Checkout. This banner will appear for merchants using the new checkout experience who don't have OC enabled, and only after we launch 9.8.

Screenshot 2025-07-15 at 16 06 03

## Testing instructions

- Checkout and build this branch on your test environment (`add/oc-promotional-banner`)
- As a merchant, make sure the PMC Sync is enabled, and both the Optimized Checkout feature flag (`_wcstripe_feature_oc` flag or the filter `wc_stripe_is_optimized_checkout_available`) and the setting itself are enabled.
- Hardcode the current extension version to 9.6 in `client/settings/payment-settings/promotional-banner/get-promotional-banner-type.js#69`

### Rendering

- Confirm the banner is not shown yet
- Disable the Optimized Checkout feature
- Confirm the new banner is displayed in the payment methods page (`wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=methods`)

### Visuals

- Confirm the visuals match the Figma file and the preview above

### Dismissing

- Click the "Dismiss" link
- Confirm the banner is gone

### Learn more link

- Click the "Learn more" link
- Confirm it sends you to: https://woocommerce.com/document/stripe/admin-experience/optimized-checkout-suite/

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -257,6 +257,10 @@ public function admin_scripts( $hook_suffix ) {
 			// Show the BNPL promotional banner only if no BNPL payment methods are enabled.
 			&& ! array_intersect( WC_Stripe_Payment_Methods::BNPL_PAYMENT_METHODS, $enabled_payment_methods );
 
+		$show_oc_promotion_banner = get_option( 'wc_stripe_show_oc_promotion_banner', 'yes' ) === 'yes'
+			// Show the OC promotional banner only if OC is disabled
+			&& ! $this->get_gateway()->is_oc_enabled();
+
```
Mayisha: Instead of doing the version comparison in the frontend, we can use `version_compare` here. That way, we don't need to write a new function for the version comparison on the frontend.

On another note, if this PR is shipped in version `9.8.0`, we don't need the version compare at all.

Mayisha (commented):
Thanks for adding the banner. The display logic of the banner is working as described. I have one suggestion about the version comparison.

✅ Banner is hidden when the plugin version does not match
✅ Banner is hidden when OC is enabled
✅ Banner is displayed when the plugin version matches the expected version and OC is disabled
✅ Clicking `Dismiss` removed the banner permanently
🟡 The visuals have minor differences
- There is a cross icon in the banner which is missing here. My opinion on this is, we don't need the icon as we already have a `Dismiss` button.
- The title `Increase conversions with Stripe's Optimized Checkout Suite` has a slightly larger font in the design.

wjrosa:
Thanks for the review once again, Mayisha!

> There is a cross icon in the banner which is missing here. My opinion on this is, we don't need the icon as we already have a Dismiss button.

Yes, we don't need the cross icon 👍 (confirmed with the design team).

> The title Increase conversions with Stripe's Optimized Checkout Suite has a slightly larger font in the design.

Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4495/commits/401574bffbcb8c0859e56c08075be7a645da02bf

diegocurbelo (changes requested):
It tests very well; the display conditions work as expected, it's only shown when OC feature flag is enabled, the plugin version is newer than the target version and OC is not enabled.

Adding to the differences @Mayisha mentioned, the [Figma banner](https://www.figma.com/design/nFI8jCCAQnd19PNwScqVU6/Stripe-Payments--External-?node-id=2066-7132&t=BWUKkVx6LGhpkuHI-0) has the Learn more link as part of the main text, and a CTA ('Activate Now'):

Image

Image

wjrosa:
Thanks for the reviews! I chatted with @orcungogus, and he agreed on removing the cross icon to close the banner and to keep the banner shown below the tabs, just like all the other banners.

For the other inconsistencies, I have fixed them in my latest commits:
- Moved the learn more inside the description paragraph
- Replaced the CTA with "Activate now" (and the proper logic for that with unit tests)

Here's an updated preview:
Screenshot 2025-07-17 at 16 16 30

includes/admin/class-wc-stripe-rest-oc-setting-toggle-controller.php
```diff
@@ -0,0 +1,98 @@
...
+	 */
+	public function set_setting( WP_REST_Request $request ) {
+		$is_oc_enabled = $request->get_param( 'is_oc_enabled' );
+
+		if ( null === $is_oc_enabled ) {
+			return new WP_REST_Response( [ 'result' => 'bad_request' ], 400 );
+		}
+
+		$settings                               = WC_Stripe_Helper::get_stripe_settings();
+		$settings['optimized_checkout_element'] = $is_oc_enabled ? 'yes' : 'disabled';
```
Mayisha: Is there a reason why we want to set it as `disabled` instead of `no`?

tests/phpunit/Admin/WC_Stripe_REST_OC_Setting_Toggle_Controller_Test.php
```diff
@@ -0,0 +1,142 @@
...
+	 *
+	 * @see UPE_Test_Utils::reload_payment_gateways()
+	 */
+	public static function set_up_before_class() {
+		parent::set_up_before_class();
+
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-stripe-rest-oc-setting-toggle-controller.php';
+
+		// All tests assume UPE is enabled.
+		update_option( '_wcstripe_feature_upe', 'yes' );
```
Mayisha: Nit: we may not need to enable UPE anymore as UPE is the default now.

tests/phpunit/Admin/WC_Stripe_REST_OC_Setting_Toggle_Controller_Test.php
```diff
@@ -0,0 +1,142 @@
...
+
+		// Set the user so that we can pass the authentication.
+		wp_set_current_user( 1 );
+	}
+
+	/**
+	 * Deregister WooCommerce Blocks REST routes to prevent _doing_it_wrong() notices
+	 * after calls to rest_do_request().
+	 */
+	public function deregister_wc_blocks_rest_api() {
```
Mayisha: QQ: why is it needed?

Mayisha (approved):
Looks good to me 👍 

Screenshot 2025-07-18 at 2 08 36 PM

Have a couple of minor comments/questions and the same suggestion about version check that I [mentioned](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4495#discussion_r2209431019) before. As we are shipping it in version 9.8.0, I think the version check can be skipped.

includes/admin/class-wc-stripe-rest-oc-setting-toggle-controller.php
wjrosa: The only reason was that I have used the UPE endpoint logic as a reference to build this, but it could be any value (we check for "yes"). I have updated it to "no" in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4495/commits/1486defbf42f599bc03f5eaa34ad1a75a5230cca

tests/phpunit/Admin/WC_Stripe_REST_OC_Setting_Toggle_Controller_Test.php
wjrosa: You're right! Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4495/commits/1f748af33ae2539a55a60159720fb825a444b5e6

tests/phpunit/Admin/WC_Stripe_REST_OC_Setting_Toggle_Controller_Test.php
wjrosa: I don't think this is needed anymore. Thanks for flagging it! It was here because I copied the default setting controller test class (`tests/phpunit/Admin/WC_REST_Stripe_Settings_Controller_Test.php`). I ran the whole test suite without it, and I didn't see any errors or warnings. So I removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4495/commits/45078999565e9227730fe9f22c4899c46f2d99af

includes/admin/class-wc-stripe-settings-controller.php
wjrosa: > On another note, if this PR is shipped in version 9.8.0, we don't need the version compare at all.

That's a good point. I have removed all version checks in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4495/commits/df77ac413e5f1f29e34d709bbe4c0c9630543512

client/settings/payment-settings/promotional-banner/oc-promotion-banner.js
```diff
@@ -0,0 +1,141 @@
...
+
+	const handleBannerDismiss = () => {
+		apiFetch( {
+			path: '/wc/v3/wc_stripe/settings/notice',
+			method: 'POST',
+			data: { wc_stripe_show_oc_promotion_banner: 'no' },
+		} ).finally( () => {
+			setShowPromotionalBanner( false );
+		} );
+		window.location.reload();
```
diegocurbelo:
```suggestion
```

We don't need to reload the page after dismissing the notice, and if we do, we should do it inside the `finally` block; otherwise, the page might reload with the old value

https://github.com/user-attachments/assets/38222822-aea9-45bb-9307-76d954d2257a

diegocurbelo (changes requested):
Thanks for the changes @wjrosa, the new CTA button works as expected.

I left a suggestion to prevent the dismiss button from refreshing the page too soon.

includes/admin/class-wc-stripe-rest-oc-setting-toggle-controller.php
```diff
@@ -0,0 +1,98 @@
...
+		);
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base,
+			[
+				'methods'             => WP_REST_Server::EDITABLE,
+				'callback'            => [ $this, 'set_setting' ],
+				'permission_callback' => [ $this, 'check_permission' ],
+				'args'                => [
+					'is_upe_enabled' => [
```
diegocurbelo: The arg to validate should be `is_oc_enabled`

```suggestion
					'is_oc_enabled' => [
```

diegocurbelo (changes requested):
The banner is always shown in the Settings tab (but it works ok in the Payment Methods tab), even after dismissing it and navigating back (or reloading the page)

https://github.com/user-attachments/assets/ed0f4e17-38c1-4291-aaa6-491b5307da12

wjrosa:
> The banner is always shown in the Settings tab (but it works ok in the Payment Methods tab), even after dismissing it and navigating back (or reloading the page)

Sorry about that, Diego! I did some refactoring in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4495/commits/71730e6e94908c8857e5baf5aeeeef7eb7452851, moving the banner state definition up to share the same value in both tabs. This should address the bug

diegocurbelo (approved):
>Sorry about that, Diego! I did some refactoring in https://github.com/woocommerce/woocommerce-gateway-stripe/commit/71730e6e94908c8857e5baf5aeeeef7eb7452851, moving the banner state definition up to share the same value in both tabs. This should address the bug

Thank you @wjrosa, it works correctly now 👍🏼 

LGTM 🚢

PR #4497, opened 2025-07-15
Add rate-limit/cooldown to Stripe API requests after 401 errors

diegocurbelo:
Fixes STRIPE-586

When the UPE Gateway is instantiated, every available payment method is instantiated, and each one reads the account country to see if the PM can be enabled or not.

This is usually not a problem because we have a static cache and a database cache in place, so we only make one API call. However, when the Stripe keys are invalid and the account cache expires, each attempt to read the account country makes a Stripe API call, which means ~25 requests every time the UPE Gateway is instantiated.

This PR prevents Stripe API calls after several consecutive 401 (Unauthorized) responses (5) by detecting invalid API keys, storing a flag, and blocking further requests for 2 hours or until the keys are updated.

This improves performance and user feedback by avoiding repeated failed attempts and indicating a disconnected state in the UI.

NOTE: We should not be displaying the `Invalid API key...` error message to customers; I have created a follow-up issue STRIPE-632 for this.

## Testing instructions

1. Navigate to `WP-Admin > WooCommerce > Settings > Payments`, and select the `Stripe` option from the list
2. Go to the `Settings` tab and make sure the account is correctly connected:
    Screenshot 2025-07-21 at 15 49 40
3. Replace the API keys with an invalid value:
4. Go to the Store, go to any product page, and attempt to place an order using Google Pay; you should get this error:
    Screenshot 2025-07-21 at 15 54 57
5. Retry the purchase **two** more times (each attempt makes two calls to Stripe, so it will go over the limit (5))
5. Go back to `WP-Admin` and into the plugin `Settings` tab and make sure the account connection error message is displayed:
    Screenshot 2025-07-21 at 16 41 19
6. Check that an error entry was logged to the plugin logs
    Go to`WooCommerce > Status > Los` and then click `woocommerce-gateway-stripe`:
    Screenshot 2025-07-21 at 16 36 46
7. Click `Configure connection` and make sure the Account status is `Disconnected`
    Screenshot 2025-05-16 at 13 16 59
8. Click the `Create or connect a test account` button, a re-connect your account
9. Check that after returning to the WP-Admin dashboard, the top notice is no longer present, and the account status section has been restored:
    Screenshot 2025-05-16 at 13 25 43

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-api.php
```diff
@@ -16,6 +16,27 @@ class WC_Stripe_API {
...
+	 * @var string
+	 */
+	const INVALID_API_KEYS_CACHE_KEY = 'invalid_api_keys_rate_limit';
+
+	/**
+	 * The Invalid API Keys consecutive 401 error count threshold.
+	 *
+	 * @var int
+	 */
+	const INVALID_API_KEYS_COUNT_THRESHOLD = 5;
```
diegocurbelo: This seems like a reasonable number. 
We might want to increase this to a much higher number to cover the account cache refresh edge case (if the Account data cache expires, instantiating the UPE gateway will trigger ~25 API requests in a row) but I'm not sure it's worth it, each page refresh will trigger another ~25 requests.

includes/class-wc-stripe-api.php
```diff
@@ -16,6 +16,27 @@ class WC_Stripe_API {
...
+
+	/**
+	 * The Invalid API Keys delay in seconds enforced for Stripe API calls after the threshold is reached.
+	 *
+	 * @var int
+	 */
+	const INVALID_API_KEYS_DELAY_IN_SECONDS = 2 * HOUR_IN_SECONDS;
```
diegocurbelo: I used the same expiration as the account cache... we don't expect invalid keys to become valid again; this is more of a false positive prevention fallback (discussed for 9.5.1)

includes/class-wc-stripe-api.php
```diff
@@ -16,6 +16,27 @@ class WC_Stripe_API {
 	const ENDPOINT           = 'https://api.stripe.com/v1/';
 	const STRIPE_API_VERSION = '2024-06-20';
 
+	/**
+	 * The Invalid API Keys cache key.
+	 *
+	 * @var string
+	 */
+	const INVALID_API_KEYS_CACHE_KEY = 'invalid_api_keys_rate_limit';
```
daledupreez: Here and below for the other new `const`s: I think we should specify visibility explicitly. It looks like the cache key and threshold should be `public`, while delay can be `private` or `protected`.

daledupreez: I think it would be worth reflecting that the value is a count in the name. It's not obvious to me what data we are caching based on the constant name. Maybe something more like `INVALID_API_RESPONSE_COUNT_CACHE_KEY`?

includes/class-wc-stripe-api.php
daledupreez: That's fair. I would also think it may be worth explicitly calling this a `CACHE_TIMEOUT` rather than a `DELAY`.
```suggestion
	const INVALID_API_KEYS_CACHE_TIMEOUT_IN_SECONDS = 2 * HOUR_IN_SECONDS;
```

includes/class-wc-stripe-api.php
```diff
@@ -265,6 +286,20 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
...
 	public static function retrieve( $api ) {
+		// If keep count of consecutive 401 errors, and it exceeds INVALID_API_KEYS_COUNT_THRESHOLD,
+		// we return null until the cache expires (INVALID_API_KEYS_DELAY_IN_SECONDS) or the keys are updated.
+		$invalid_api_keys_count = WC_Stripe_Database_Cache::get( self::INVALID_API_KEYS_CACHE_KEY );
+		if ( ! empty( $invalid_api_keys_count ) && self::INVALID_API_KEYS_COUNT_THRESHOLD <= $invalid_api_keys_count ) {
+			// We skip logging the error here because when there is no Account cache,
+			// the instantiation of the UPE gateway triggers a call to this method for
+			// every available payment method. This would result in excessive log entries
+			// which is not useful.
+			// We only log the error when the count exceeds the threshold.
```
daledupreez: Nit: it feels like we could be more accurate with this comment.
```suggestion
			// We only log the error when the count exceeds the threshold for the first time.
```

includes/class-wc-stripe-api.php
```diff
@@ -281,7 +316,27 @@ public static function retrieve( $api ) {
...
+				// We need to invalidate the Account Data cache here, so that the UI shows the "Connect to Stripe" button.
+				WC_Stripe_Database_Cache::delete( WC_Stripe_Account::ACCOUNT_CACHE_KEY );
+			}
+
 			return null; // The UI expects this empty response in case of invalid API keys.
+
+		} elseif ( null !== $invalid_api_keys_count ) {
+			// We only track consecutive 401 errors, so we delete the cache when we get a valid response.
+			WC_Stripe_Database_Cache::delete( self::INVALID_API_KEYS_CACHE_KEY );
 		}
```
daledupreez: I am not sure we even need the `elseif`. The code above will always `return null;`, so we can handle this via a code comment. (Non-blocking.)
```suggestion
		}

		// We got a valid, non-401 response, so clear the invalid API key count if it is present.
		if ( null !== $invalid_api_keys_count ) {
			WC_Stripe_Database_Cache::delete( self::INVALID_API_KEYS_CACHE_KEY );
		}
```

daledupreez: That said, what about cases where we never reach Stripe at all due to networking issues or a local error? Should we not check the HTTP response codes before clearing the cache?

tests/phpunit/WC_Stripe_API_Test.php
```diff
@@ -117,6 +127,65 @@ public function test_retrieve_makes_api_call_when_api_keys_are_valid() {
...
+
+		remove_all_filters( 'pre_http_request' );
+		WC_Stripe_Database_Cache::delete( WC_Stripe_API::INVALID_API_KEYS_CACHE_KEY );
+	}
+
+	/**
+	 * Test WC_Stripe_API::retrieve() resets the invalid API key count on successful response.
+	 */
+	public function test_retrieve_resets_invalid_api_key_count_on_successful_response() {
+		$full_cache_key = 'wcstripe_cache_' . ( WC_Stripe_Mode::is_test() ? 'test_' : 'live_' ) . WC_Stripe_API::INVALID_API_KEYS_CACHE_KEY;
```
daledupreez: It looks like we're not using this. We should remove it.
```suggestion
```

daledupreez (commented):

includes/class-wc-stripe-api.php
diegocurbelo: Good idea... fixed in [c144129](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4497/commits/c144129360e40c4be5ce9c26dab917ef511396f4)

diegocurbelo: Good point, renamed them to `INVALID_API_KEY_ERROR_COUNT_CACHE_KEY`, `INVALID_API_KEY_ERROR_COUNT_CACHE_TIMEOUT `, and `INVALID_API_KEY_ERROR_COUNT_THRESHOLD ` in [dd3ea04](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4497/commits/dd3ea04e4543deb0f2d6c308e4e6f687c8c69bc2)

includes/class-wc-stripe-api.php
```diff
@@ -281,7 +317,29 @@ public static function retrieve( $api ) {
...
+			}
+
 			return null; // The UI expects this empty response in case of invalid API keys.
+
+		}
+
+		// We got a valid, non-401 response, so clear the invalid API key count if it is present.
+		if ( null !== $invalid_api_key_error_count ) {
+			WC_Stripe_Database_Cache::delete( self::INVALID_API_KEY_ERROR_COUNT_CACHE_KEY );
 		}
```
daledupreez, line 343: Not blocking, but do we want to confirm we got a 2xx response code before we clear this cache? I don't think it's a big thing either way, but clearing the cache for any non-401 response feels like it may be the wrong approach in some situations, e.g. 500 or 403 responses.

Mayisha (approved):
Works as described 👍

daledupreez (approved):
The changes look good to me and test well.

I have one non-blocking comment that we should think about, but I am not sure if we need to take action on it.

diegocurbelo:
The `WC_Stripe_Subscriptions_Helper_Test::test_is_subscription_payment_method_detached` was failing because of the 401 rate limiter. The underlying problem is that in several tests, we are not properly mocking the Stripe API calls (either by mocking methods or by using pre_http_request).

That fix is outside the scope of this PR, and I'll fix it in a new PR. For now, I updated the `tearDown` methods to reset the rate-limit check (INVALID_API_KEY_ERROR_COUNT_CACHE_KEY) with a TODO comment:

```
// The tests in this file do not mock ALL the calls to the Stripe API, and as we use mocked API keys, they trigger the 401 rate-limiter,
// this is not a problem for these tests as they don't depend on the responses.
//
// TODO: Remove this once we've mocked all calls to the Stripe API (either using the pre_http_request filter, or by using a mocked WC_Stripe_API class).
WC_Stripe_Database_Cache::delete( WC_Stripe_API::INVALID_API_KEY_ERROR_COUNT_CACHE_KEY );
```

PR #4451, opened 2025-07-01
ECE: Fix checkout fails for countries without states

Mayisha:
Fixes STRIPE-422
Fixes #4283

For some countries, Google Pay/Apple Pay returns empty value in the `state` field. This causes validation error in Store API during checkout. This PR fixes the issue for countries without state fields. 

> [!NOTE] 
> **Countries without states** (also noted in `includes/constants/class-wc-stripe-payment-request-button-states.php`
AO (Angola), BD (Bangladesh), BG (Bulgaria), BJ (Benin), BO (Bolivia), DO (Dominican Republic), DZ (Algeria), GH (Ghana), GT (Guatemala), HU (Hungary), KE (Kenya), LA (Laos), LR (Liberia), MD (Moldova), NA (Namibia), NP (Nepal), PK (Pakistan), PY (Paraguay), RO (Romania), TZ (Tanzania), UG (Uganda), UM (United States Minor Outlying Islands), ZA (South Africa), ZM (Zambia)

## Changes proposed in this Pull Request:
- Used the `woocommerce_get_country_locale` filter to make the state field optional for the countries that do not have states.
    - This is the same approach as WooPayments ([PR](https://github.com/Automattic/woocommerce-payments/pull/10770))
    - I have used the existing state mapping from `includes/constants/class-wc-stripe-payment-request-button-states.php`. The state field is marked as optional only for the countries without states from this file.
- Added a new nonce validation for security.
    - The filter will modify the data only if the request is in express checkout context and have the correct nonce.

## Testing instructions
- Enable Google Pay.
- Join this Google group to access various test data for Google Pay: https://groups.google.com/g/googlepay-test-mode-stub-data
- As a shopper, add a product to your cart and go to the checkout page.
- Click Google Pay and select `Romanian User` from the shipping address list.

Screenshot 2025-07-01 at 3 35 22 PM

- Click the Pay button.
- In `develop`, the checkout should fail with an error.

Screenshot 2025-07-01 at 3 35 46 PM

- In this branch, payment and checkout should be successful.

PR #4501, opened 2025-07-17
[ECE] Fix state mapping for Lithuania

Mayisha:
Reported in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4451#issuecomment-3078801734

The mapping for Lithuania was missing in `includes/constants/class-wc-stripe-payment-request-button-states.php`. The state field is not required for Lithuania in Google/Apple Pay. 

## Testing instructions
- Enable express checkout in your Stripe settings page.
- In your Google account, add a new address based in Lithuania but leave the `Province` field empty (it is optional)
- Click Google Pay and select the address in Lithuania as the shipping address.
- Click the `Pay` button.
- In `develop`, the checkout will fail with the following error.

Screenshot 2025-07-17 at 11 38 38 AM

- In this branch, checkout should be successful. Confirm that the shipping address is correct in your order details page.


annemirasol (approved):
LGTM, works well. 🚀 I was able to repro in develop, and verify the fix in this branch.

wjrosa (approved):
Change is clean and works as expected! 👍

PR #4502, opened 2025-07-17
Fix unnecessary payment method creation when checkout has validation errors

Mayisha:
Fixes STRIPE-589

- Before calling `stripe.createPaymentMethod` checking the page for any validation error.

## Testing instructions
- Enable card in your Stripe settings
- As a shopper, open your browser in private/incognito mode and open the network tab of your developer console.
- Add a product to your cart and go to the block checkout page.
- Without filling in the email or billing details enter card details and click the place order button.
- In `develop`, 
    - Notice that a request is sent to Stripe to create a payment method but the checkout does not go through as there are missing required fields.
- In this branch, 
    - Confirm that no request is sent to Stripe to create a payment method.
    - Fill in all the required fields and click checkout again. Confirm that a request is sent to Stripe to create a payment method, this time and the checkout succeeds.
- Do the same test for the shortcode checkout.

### DIFF
--- a/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
+++ b/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
@@ -194,11 +194,11 @@ const PaymentProcessor = ( {
 						};
 					}
 
-					const hasError = document.querySelector(
+					const hasValidationError = document.querySelector(
 						'.wc-block-components-validation-error'
 					);
 					// Return if there is a validation error on the checkout fields.
-					if ( hasError ) {
+					if ( hasValidationError ) {
 						return;
 					}
 
--- a/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
+++ b/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
@@ -194,6 +194,14 @@ const PaymentProcessor = ( {
 						};
 					}
 
+					const hasError = document.querySelector(
+						'.wc-block-components-validation-error'
+					);
+					// Return if there is a validation error on the checkout fields.
+					if ( hasError ) {
+						return;
+					}
+
 					// BLIK is a special case which is not handled through the Stripe element.
 					if ( ! ( isPaymentElementComplete || isBlikSelected ) ) {
 						return {
--- a/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
+++ b/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
@@ -3,6 +3,7 @@
  */
 import { getPaymentMethods } from '@woocommerce/blocks-registry';
 import { __ } from '@wordpress/i18n';
+import { select } from '@wordpress/data';
 import {
 	PaymentElement,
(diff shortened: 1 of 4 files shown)


client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -194,6 +194,14 @@ const PaymentProcessor = ( {
 						};
 					}
 
+					const hasValidationError = document.querySelector(
```
Mayisha: In block checkout, the validation check happens on the frontend. Checking for `.wc-block-components-validation-error` class gives us the result.

client/classic/upe/deferred-intent.js
```diff
@@ -41,6 +41,22 @@ jQuery( function ( $ ) {
 	} );
 
 	$( 'form.checkout' ).on( generateCheckoutEventNames(), function () {
+		const $form = $( 'form.checkout' );
+		// Lose focus for all fields to trigger validation
+		$form
+			.find( '.input-text, select, input:checkbox' )
+			.trigger( 'validate' )
+			.trigger( 'blur' );
```
Mayisha, line 49: In shortcode checkout, the validation check is done after the `/checkout` request is sent. `createPaymentMethod` is called before the `/checkout` request.
So we have to explicitly trigger the validation check here as we want to prevent `createPaymentMethod` call based on the validation result.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
annemirasol: This will not cover fields that do not have an inline error message, e.g. Terms and Conditions checkbox (enabled via the editor). I suggest we use WooCommerce's [Validation Store](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/client/blocks/docs/third-party-developers/extensibility/data-store/validation.md). It would be used like this:

```
const { validationStore } = window.wc?.wcBlocksData ?? {};
if ( validationStore ) {
  const store = select( validationStore );
  const hasValidationErrors = store.hasValidationErrors();
  if ( hasValidationErrors ) {
	  return;
  }
}
```


Screenshot 2025-07-17 at 4 45 12 PM
Screenshot 2025-07-17 at 4 47 39 PM

annemirasol (changes requested):

Mayisha:
Thank you for the suggestion @annemirasol. I have addressed this in [b18bab0](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4502/commits/b18bab09ec02fa8b03bc4db5669a7ec3e24e04e9)

diegocurbelo (approved):
Thanks for fixing this @Mayisha!. 

LGTM, the changes look good and it tests well 🚢

annemirasol (approved):
LGTM!

✅ No `POST /v1/payment_methods` API call observed when there are validation errors, even for custom fields (block/classic checkout)
✅ Can proceed with purchase without problems when there are no validation errors

PR #4506, opened 2025-07-18
Prevent multiple save appearance AJAX calls on Block Checkout

malithsen:
Fixes STRIPE-588

This PR fixes an issue where multiple save appearance AJAX calls were being made on the Block Checkout page.

The `initializeUPEAppearance` function checks if appearance data exists in the global variable (`getStripeServerData().blocksAppearance`). When the data is absent, it generates new appearance settings and sends an AJAX request to save them. However, the global variable wasn't updated until the page was refreshed, causing subsequent calls to the function (triggered by React re-renders) to see the data as still empty and repeat the save process.

This PR:
- Immediately update the global variable to prevent subsequent calls from seeing empty data.
- Adds a flag (`isSavingAppearance`) to prevent multiple simultaneous AJAX saves.

## Testing instructions

1. Remove appearance transients (I do this directly on the db by deleting the results of `select * from wp_options where option_name like '%appearance%'`)
2. Open dev tools -> Network tab on your browser.
3. Add a product to your cart and go to the Block checkout page.
4. Without filling any shipping/billing fields, click the "Place order" button.
5. In **develop**: there'll be multiple `wc_stripe_save_apperance` calls.
6. In **this branch**: there should only be a single AJAX call to save the appearance data.
7. (Optional test) Compare the styling data of the payment element between this branch and **develop**, they should remain unchanged. 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (approved):
Thanks for the fix. It's working as described. 🎉 

✅ One request is made to update the appearance when the corresponding transient is not set
✅ There are no unnecessary subsequent requests.

wjrosa (approved):
Change looks good to me! Thanks for working on this one 👍

PR #4507, opened 2025-07-18
Adding the subscription detached check bulk action

wjrosa:
Fixes STRIPE-501

This PR introduces a new bulk action available in the subscription listing screen to check if the selected subscriptions contain detached payment methods, which would prevent renewals.

## Testing instructions

- Checkout to this branch on your test environment (`add/subscription-detached-bulk-action`)
- Connect your Stripe account
- Add some subscriptions (create or purchase them as a shopper)
- Hardcode the return of `is_subscription_payment_method_detached` to `true` (`includes/compat/class-wc-stripe-subscriptions-helper.php`)
- Go to the subscriptions listing screen
- Select at least one subscription
- Confirm you can see the new bulk action in the dropdown:
Screenshot 2025-07-18 at 19 25 55

- Select the new action and click "Apply"
- Confirm you can see a notice about the detached subscriptions:
Screenshot 2025-07-18 at 19 26 40

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -215,6 +211,20 @@ public static function is_manual_renewal_enabled() {
 		return false;
 	}
 
+	/**
+	 * Extracts data from a subscription object for detached subscriptions.
+	 *
+	 * @param $subscription WC_Subscription The subscription object to extract data from.
+	 * @return array
+	 */
+	public static function get_detached_data_from_subscription( $subscription ) {
```
Mayisha: Optional suggestion,
```suggestion
	public static function get_detached_payment_data_from_subscription( $subscription ) {
```

includes/admin/class-wc-stripe-subscription-detached-bulk-action.php
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	/**
+	 * Add custom bulk action to check for detachment of subscriptions' payment methods.
+	 *
+	 * @param $bulk_actions array An associative array of actions which can be performed on the subscription post type.
+	 * @return array
+	 */
+	public function subscriptions_bulk_actions( $bulk_actions ) {
+		$bulk_actions['check-for-payment-method-detachment'] = __( 'Check for payment method detachment', 'woocommerce-gateway-stripe' );
```
Mayisha, line 31: This action name is perfectly aligned with the feature. But I want to suggest another alternative. We can also call it `Payment method health`.  This optional and just my preference.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -512,6 +513,49 @@ public function subscription_check_detachment() {
...
+				if ( ! $subscription instanceof WC_Subscription ) {
+					continue;
+				}
+				$subscriptions[] = WC_Stripe_Subscriptions_Helper::get_detached_data_from_subscription( $subscription );
+			}
+			$detached_messages = WC_Stripe_Subscriptions_Helper::build_subscriptions_detached_messages( $subscriptions );
+
+			$notice_content = '';
+			if ( empty( $detached_messages ) ) {
+				$notice_content .= '<p>' . esc_html__( 'No detached subscriptions found.', 'woocommerce-gateway-stripe' ) . '</p>';
```
Mayisha: I do not get this notice when I select and check subscriptions that have attached payment methods.

Mayisha (commented):
Thanks @wjrosa for working on this. I have added a couple of optional comments. Also I noticed the following behavior which I don't this is expected-

**Case 1**
- Select one or two subscriptions that have attached payment methods and run the action. 
- There is no notice displayed. But as per the code there should be a notice with the message `No detached subscriptions found.`

**Case 2**
- Select a few subscriptions including a few detached one (or select all) and run the action.
- The notice is displayed correctly

Screenshot 2025-07-21 at 12 36 03 AM

- Refresh the page. Notice is still present (is this expected? There is no way to remove the notice at this point) 
- Now from the subscriptions list select one or two subscriptions that has attached payment methods.
- Run the action. Notice that the previous notice is still displayed.
- If you choose a different set/subset of detached subscription and run the action, the notice updates correctly with the new selected subscriptions details.

includes/admin/class-wc-stripe-subscription-detached-bulk-action.php
wjrosa: Hum, I think "Payment method health" may sound like we would do more than checking if it is detached, feels like a broader feature. I think I will stick to the current label for now 👍

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: I have fixed it on my latest commits 👍

wjrosa:
Thanks for the review, Mayisha!

I have fixed both issues in my latest commits:
- The notice is now dismissible
- The "no detached subscriptions found" feedback is correctly shown

Mayisha (changes requested):
@wjrosa now I see `No detached subscriptions found` at the beginning of the error notice always. 😬 

Screenshot 2025-07-22 at 1 41 08 PM

wjrosa:
> @wjrosa now I see `No detached subscriptions found` at the beginning of the error notice always. 😬
> 
> Screenshot 2025-07-22 at 1 41 08 PM

Oops! Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4507/commits/8cb204503b56b8d34f033752430b7aeaf0d25a08

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -593,6 +636,13 @@ public function hide_notices() {
 				case 'subscriptions':
 					update_option( 'wc_stripe_show_subscriptions_notice', 'no' );
 					break;
+				case 'subscription_detached_bulk_action':
+					update_option( 'wc_stripe_show_subscription_detached_bulk_action_notice', 'no' );
+					break;
+			}
+			// Redirect back to the current page without the query param to hide the notice to avoid issues.
+			if ( isset( $_SERVER['REQUEST_URI'] ) ) {
+				wp_safe_redirect( remove_query_arg( [ 'wc-stripe-hide-notice', '_wc_stripe_notice_nonce' ], esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) ) ) );
```
annemirasol: I wonder if we should limit this redirect clause to just for `subscription_detached_bulk_action`, just to be safe that we're not affecting expected behavior for other notices?

annemirasol (approved):
This works as described, code LGTM!

Tested:
- Bulk action displays selected subs with detached PMs
- Hide notice works as expected

Optional suggestion about the extra redirect after hiding a notice.

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Yeah, that's a valid concern! Moved it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4507/commits/f6c45e310d4b5178955a10b0eb990bfe12c9bff5

Mayisha (approved):
Thanks for the updates @wjrosa. LGTM 👍 

✅ Message shows correctly when there is no subscription with detached payment in the selected list.

Screenshot 2025-07-23 at 1 35 28 PM

✅ Message shows correct number and details of subscriptions when there are subscriptions with detached payment in the selected list. 

Screenshot 2025-07-23 at 1 38 10 PM

✅ Correct message is displayed when the selection is changed and the action is run again.

Screenshot 2025-07-23 at 1 39 07 PM

✅  Notice can be dismissed.

PR #4514, opened 2025-07-22
Making OC available by default

wjrosa:
Fixes STRIPE-636

For version 9.8.0, we want to make the [Optimized Checkout](https://woocommerce.com/document/stripe/admin-experience/optimized-checkout-suite/) feature available to all merchants by default. This PR changes the availability method to ignore the feature flag, thereby avoiding the need to create a migration script later.

## Testing instructions

- Checkout to this branch on your test environment (`add/making-oc-available-by-default`)
- Connect your Stripe account
- Go to the Stripe settings page (`wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings`)

### New store

- Confirm the Optimized Checkout setting is visible

### Existing store

- Confirm the Optimized Checkout setting is visible

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-feature-flags.php
+++ b/includes/class-wc-stripe-feature-flags.php
@@ -24,7 +24,7 @@ class WC_Stripe_Feature_Flags {
 		'_wcstripe_feature_upe'                => 'yes',
 		self::ECE_FEATURE_FLAG_NAME            => 'yes',
 		self::AMAZON_PAY_FEATURE_FLAG_NAME     => 'no',
-		self::OC_FEATURE_FLAG_NAME             => 'yes',
+		self::OC_FEATURE_FLAG_NAME             => 'no',
 		self::LPM_ACH_FEATURE_FLAG_NAME        => 'yes',
 		self::LPM_ACSS_FEATURE_FLAG_NAME       => 'yes',
 		self::LPM_BACS_FEATURE_FLAG_NAME       => 'yes',
--- a/includes/class-wc-stripe-feature-flags.php
+++ b/includes/class-wc-stripe-feature-flags.php
@@ -7,12 +7,20 @@ class WC_Stripe_Feature_Flags {
 	const UPE_CHECKOUT_FEATURE_ATTRIBUTE_NAME = 'upe_checkout_experience_enabled';
 	const ECE_FEATURE_FLAG_NAME               = '_wcstripe_feature_ece';
 	const AMAZON_PAY_FEATURE_FLAG_NAME        = '_wcstripe_feature_amazon_pay';
-	const OC_FEATURE_FLAG_NAME                = '_wcstripe_feature_oc';
 	const LPM_ACH_FEATURE_FLAG_NAME           = '_wcstripe_feature_lpm_ach';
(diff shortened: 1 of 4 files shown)


annemirasol (commented):
Should we also clean up `class-wc-stripe-feature-flags.php`, or at least update OC-related stuff in there?

wjrosa:
> Should we also clean up `class-wc-stripe-feature-flags.php`, or at least update OC-related stuff in there?

What I think we can do for now is deprecate the public constant with the feature flag definition. I did that now in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4514/commits/8420235d503e53ae5dee28069de83ff830f8ae6f. On the next version, we can then safely remove it

annemirasol (approved):
LGTM! Tested that the feature is available even when the feature flag is off.

includes/class-wc-stripe-feature-flags.php
```diff
@@ -174,13 +182,20 @@ public static function are_apms_deprecated() {
...
+		 * @param string $pmc_enabled The value of the 'pmc_enabled' setting.
+		 */
 		return apply_filters(
 			'wc_stripe_is_optimized_checkout_available',
-			'yes' === $default_value && 'yes' === $pmc_enabled,
-			$default_value,
+			'yes' === $pmc_enabled,
+			'yes',
 			$pmc_enabled
 		);
```
daledupreez: In looking at the unit test code, I thought we should add a test case for the case where PMC is disabled, but the filter returns true. However, I then realised that the filter currently overrides the situation where PMC is disabled. Which I think is something we should never allow. I think we need to change this code to avoid calling the filter when PMC is disabled.

I am thinking we need something like the following:
```php
	$pmc_enabled = $stripe_settings['pmc_enabled'] ?? 'no';
	if ( $pmc_enabled !== 'yes' ) {
		return false;
	}
```

wjrosa: That's a good point, Dale. I have updated the code accordingly in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4514/commits/f5e77338c4e3e0f087e96d2baaae6726521b9f32

includes/class-wc-stripe-feature-flags.php
```diff
@@ -174,13 +182,24 @@ public static function are_apms_deprecated() {
...
+		 * @since 9.6.0
+		 * @deprecated This filter will be removed in version 9.9.0. No replacement will be provided as the Optimized Checkout feature will be permanently enabled.
+		 * @param string $default_value The default value for the feature flag.
+		 * @param string $pmc_enabled The value of the 'pmc_enabled' setting.
+		 */
 		return apply_filters(
 			'wc_stripe_is_optimized_checkout_available',
-			'yes' === $default_value && 'yes' === $pmc_enabled,
-			$default_value,
+			'yes' === $pmc_enabled,
```
daledupreez: Given that we would have returned early above if `$pmc_enabled` is not `'yes'`, I think it would be clearer to default this to `true` instead.
```suggestion
			true,
```

daledupreez (approved):
Thanks for all the changes! This is looking good!

PR #4515, opened 2025-07-22
Deactivates Affirm or Klarna in PMC when other extensions are active

wjrosa:
Fixes STRIPE-628

In https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4492, we filtered out Affirm and Klarna when other official plugins are active. Here, I am forcing the deactivation (under the same rules) directly on the merchant's Payment Method Configuration. This is needed to avoid these payment methods from showing up when the Optimized Checkout is enabled.

## Testing instructions

- Checkout to this branch on your test environment (`update/disable-bnpls-in-pmc-when-other-plugins-are-active`)
- Connect a US Stripe account
- Go to the Stripe dashboard and enable both Klarna and Affirm in the main PMC (the one with "WooCommerce Inc. configuration")
- Install the [Affirm](https://woocommerce.com/products/woocommerce-gateway-affirm/) or [Klarna](https://woocommerce.com/products/klarna-payments/) extensions (only one is enough)
- Go to the store's wp-admin
- Revisit the Stripe dashboard and confirm the related payment method was deactivated
- Enable the Optimized Checkout (you can use the `wc_stripe_is_optimized_checkout_available` filter and enable the setting)
- Confirm the related payment method is not available in the checkout pages

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -877,10 +877,10 @@ class WC_Stripe {
 
 		$enabled_payment_methods       = $gateway->get_upe_enabled_payment_method_ids();
 		$payment_method_ids_to_disable = [];
-		if ( in_array( WC_Stripe_Payment_Methods::AFFIRM, $enabled_payment_methods, true ) && $has_affirm_plugin_active ) {
+		if ( $has_affirm_plugin_active && in_array( WC_Stripe_Payment_Methods::AFFIRM, $enabled_payment_methods, true ) ) {
 			$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::AFFIRM;
 		}
-		if ( in_array( WC_Stripe_Payment_Methods::KLARNA, $enabled_payment_methods, true ) && $has_klarna_plugin_active ) {
+		if ( $has_klarna_plugin_active && in_array( WC_Stripe_Payment_Methods::KLARNA, $enabled_payment_methods, true ) ) {
 			$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::KLARNA;
 		}
 
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -874,6 +874,9 @@ class WC_Stripe {
 		}
 
 		$gateway = $this->get_main_stripe_gateway();
+		if ( ! is_a( $gateway, 'WC_Stripe_UPE_Payment_Gateway' ) ) {
+			return;
+		}
 
 		$enabled_payment_methods       = $gateway->get_upe_enabled_payment_method_ids();
 		$payment_method_ids_to_disable = [];
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -281,6 +281,9 @@ class WC_Stripe {
 		add_action( 'init', [ $this, 'initialize_status_page' ], 15 );
 
 		add_action( 'init', [ $this, 'initialize_apple_pay_registration' ] );
+
+		// Check if other official plugins are active for Klarna or Affirm and deactivate those BNPLs if so.
+		add_action( 'init', [ $this, 'maybe_deactivate_bnpls' ] );
 	}
 
 	/**
@@ -856,5 +859,40 @@ class WC_Stripe {
 
 		$wcstripe_status = new WC_Stripe_Status( self::get_main_stripe_gateway(), $this->account );
 		$wcstripe_status->init_hooks();
+	}
+
+	/**
+	 * Maybe deactivate Affirm or Klarna payment methods if other official plugins are active.
+	 *
+	 * @return void
+	 */
+	public function maybe_deactivate_bnpls() {
+		$has_affirm_plugin_active = WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_AFFIRM );
+		$has_klarna_plugin_active = WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_KLARNA );
+		if ( ! $has_affirm_plugin_active && ! $has_klarna_plugin_active ) {
+			return;
+		}
+
+		$gateway = $this->get_main_stripe_gateway();
+
+		$enabled_payment_methods       = $gateway->get_upe_enabled_payment_method_ids();
+		$payment_method_ids_to_disable = [];
+		if ( in_array( WC_Stripe_Payment_Methods::AFFIRM, $enabled_payment_methods, true ) && $has_affirm_plugin_active ) {
+			$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::AFFIRM;
+		}
+		if ( in_array( WC_Stripe_Payment_Methods::KLARNA, $enabled_payment_methods, true ) && $has_klarna_plugin_active ) {
+			$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::KLARNA;
+		}
+
+		if ( [] === $payment_method_ids_to_disable ) {
+			return;
+		}
+
+		$gateway->update_enabled_payment_methods(
+			array_diff(
+				$enabled_payment_methods,
+				$payment_method_ids_to_disable
+			)
+		);
 	}
 }
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -3354,7 +3354,7 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Gateway_Stripe {
 				&& WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_KLARNA ) ) {
 				$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::KLARNA;
 			}
-			if ( count( $payment_method_ids_to_disable ) > 0 ) {
+			if ( $payment_method_ids_to_disable !== [] ) {
 				$this->update_enabled_payment_methods(
 					array_diff(
 						$enabled_payment_methods,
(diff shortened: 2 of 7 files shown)


Mayisha (approved):
Tested the following flow-

- Enabled Klarna, Affirm and a few more methods.
- Installed the official Affirm plugin.
- Refreshed the settings page.
- Affirm got disabled in the plugin settings page as well as in Stripe dashboard. Klarna and other methods remained same as before.
- On checkout page with split PE, Klarna and other methods are present and Affirm is absent.

Screenshot 2025-07-23 at 5 14 22 PM

- ⚠️  On checkout page with split PE, other methods are present, Affirm is absent but Klarna is also absent for me. Is it because of my location?

Screenshot 2025-07-23 at 5 15 01 PM

Approving this as the issue/question I mentioned above is unrelated to the changes of this PR.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3331,4 +3334,32 @@ public function is_payment_request_enabled() {
...
 			in_array( WC_Stripe_Payment_Methods::GOOGLE_PAY, $enabled_payment_method_ids, true );
 	}
+
+	/**
+	 * Maybe deactivate Affirm or Klarna payment methods if other official plugins are active.
+	 *
+	 * @return void
+	 */
+	private function maybe_deactivate_bnpls() {
+		if ( WC_Stripe_Helper::has_other_bnpl_plugins_active() ) {
```
daledupreez: Non-blocking: why not flip this into an early return?
```suggestion
		if ( ! WC_Stripe_Helper::has_other_bnpl_plugins_active() ) {
			return;
		}
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3331,4 +3334,32 @@ public function is_payment_request_enabled() {
...
+			$enabled_payment_methods       = $this->get_upe_enabled_payment_method_ids();
+			$payment_method_ids_to_disable = [];
+			if ( in_array( WC_Stripe_Payment_Methods::AFFIRM, $enabled_payment_methods, true )
+				&& WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_AFFIRM ) ) {
+				$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::AFFIRM;
+			}
```
daledupreez: Should we also require that the relevant payment gateway is enabled before we actively make the API call to Stripe?  Given that we're doing this in the constructor, we might trip over a number of tricky race conditions if that's part of the behaviour. But maybe it would be worth hooking into a much later hook to handle this cleanup anyway?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3331,4 +3334,32 @@ public function is_payment_request_enabled() {
...
+			if ( in_array( WC_Stripe_Payment_Methods::KLARNA, $enabled_payment_methods, true )
+				&& WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_KLARNA ) ) {
+				$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::KLARNA;
+			}
+			if ( count( $payment_method_ids_to_disable ) > 0 ) {
```
daledupreez: Non-blocking, but we could also use a simple comparison here.
```suggestion
			if ( $payment_method_ids_to_disable !== [] ) {
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: > Should we also require that the relevant payment gateway is enabled before we actively make the API call to Stripe?

`WC_Stripe_Helper::has_gateway_plugin_active` already checks if the gateway is enabled 👍 

> But maybe it would be worth hooking into a much later hook to handle this cleanup anyway?

Any suggestions? I could not find a better place to implement this call 🤔

wjrosa:
> ⚠️ On checkout page with split PE, other methods are present, Affirm is absent but Klarna is also absent for me. Is it because of my location?

Thanks for the review, @Mayisha! Klarna will be present when the Stripe account and shopper account country match, as well as the currency (domestic payment). Maybe that's the issue here for you?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: > > Should we also require that the relevant payment gateway is enabled before we actively make the API call to Stripe?
>
> `WC_Stripe_Helper::has_gateway_plugin_active` already checks if the gateway is enabled 👍

Excellent!

> > But maybe it would be worth hooking into a much later hook to handle this cleanup anyway?
>
> Any suggestions? I could not find a better place to implement this call 🤔

Maybe we could hook into `wp_loaded` so we're running after most hooks? Ref: https://developer.wordpress.org/reference/hooks/wp_loaded/

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3331,4 +3334,28 @@ public function is_payment_request_enabled() {
...
 			in_array( WC_Stripe_Payment_Methods::GOOGLE_PAY, $enabled_payment_method_ids, true );
 	}
+
+	/**
+	 * Maybe deactivate Affirm or Klarna payment methods if other official plugins are active.
+	 *
+	 * @return void
+	 */
+	private function maybe_deactivate_bnpls() {
+		if ( WC_Stripe_Helper::has_other_bnpl_plugins_active() ) {
```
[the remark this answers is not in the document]
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4515/commits/011bb2a0513dda6f5ec2ab3014fa8319d0b91364

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Good suggestion, but I went ahead and used `init` instead to follow the pattern we have in the main `WC_Stripe` class (moved the method there). Per the documentation above, it is close to `wp_loaded` (occurring just before).

includes/class-wc-stripe.php
```diff
@@ -857,4 +860,39 @@ public function initialize_status_page() {
...
+		$gateway = $this->get_main_stripe_gateway();
+
+		$enabled_payment_methods       = $gateway->get_upe_enabled_payment_method_ids();
+		$payment_method_ids_to_disable = [];
+		if ( in_array( WC_Stripe_Payment_Methods::AFFIRM, $enabled_payment_methods, true ) && $has_affirm_plugin_active ) {
+			$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::AFFIRM;
+		}
+		if ( in_array( WC_Stripe_Payment_Methods::KLARNA, $enabled_payment_methods, true ) && $has_klarna_plugin_active ) {
+			$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::KLARNA;
+		}
```
daledupreez: Non-blocking nit: I would prefer to check the pre-computed value first in each case.
```suggestion
		if ( $has_affirm_plugin_active && in_array( WC_Stripe_Payment_Methods::AFFIRM, $enabled_payment_methods, true ) ) {
			$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::AFFIRM;
		}
		if ( $has_klarna_plugin_active && in_array( WC_Stripe_Payment_Methods::KLARNA, $enabled_payment_methods, true ) ) {
			$payment_method_ids_to_disable[] = WC_Stripe_Payment_Methods::KLARNA;
		}
```

tests/phpunit/WC_Stripe_Test.php
```diff
@@ -33,214 +25,128 @@ public function test_constants_defined() {
...
+			->getMock();
 
-		// Turn UPE on first.
-		$stripe_settings['upe_checkout_experience_enabled'] = 'yes';
-		WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
+		$wc_stripe->method( 'get_main_stripe_gateway' )
+			->willReturn( $upe_payment_gateway );
 
-		$this->mock_payment_method_configurations( [ WC_Stripe_Payment_Methods::SEPA_DEBIT, WC_Stripe_Payment_Methods::IDEAL ] );
+		$wc_stripe->maybe_deactivate_bnpls();
```
daledupreez, line 65: Should we not verify that we performed the correct actions and that the correct payment methods were disabled? This test feels incomplete.

includes/class-wc-stripe.php
```diff
@@ -857,4 +860,39 @@ public function initialize_status_page() {
...
+	 * @return void
+	 */
+	public function maybe_deactivate_bnpls() {
+		$has_affirm_plugin_active = WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_AFFIRM );
+		$has_klarna_plugin_active = WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_KLARNA );
+		if ( ! $has_affirm_plugin_active && ! $has_klarna_plugin_active ) {
+			return;
+		}
+
+		$gateway = $this->get_main_stripe_gateway();
```
daledupreez, line 876: Should we verify that we have the UPE gateway before we proceed? Users can technically still use the old gateway.

daledupreez (approved):
This is working as expected, but I have some questions that may drive some minor improvements.

Apologies for the delay in getting this tested and reviewed!

includes/class-wc-stripe.php
wjrosa: Added a check in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4515/commits/814ff4bf5febc179434ea7bbccca85adde498a8c

tests/phpunit/WC_Stripe_Test.php
wjrosa: We are not unit-testing the `update_enabled_payment_methods` method (but `maybe_deactivate_bnpls`), but I can check that here if you think it is better.

wjrosa: Also, I tried to create the tests, but I don't see value on them in practice because:
- We need to mock the Stripe API response when retrieving the active methods to decide whether we will deactivate some (or cache the configuration)
- Upon deactivating, the cache is cleared by the PMC class
- We then retrieve the methods again to check if those were properly disabled. At this point we have to mock the response again or cache the configuration

This will result in a test that evaluates a cached configuration or a mocked response, which does not make sense.
You can see a draft of the work here https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4539

PR #4517, opened 2025-07-22
Fix failing unit tests on WC 10.0.*

malithsen:
Fixes STRIPE-631

- `test_process_payment_returns_valid_response` and `test_if_order_has_subscription_payment_method_will_be_saved ` these two tests are failing because `wc_get_order` within the tests returns a fresh instance of the order, while the actual functions have a `$order->save()` call which mutates the metadata. https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4517/commits/3420be3be9790f89dee66a8c70f0fa7e8d956f3e fixes it by only comparing the order IDs, following an existing test pattern in the file. I am not entirely sure why this is erroring out only in WC v10, it's possible something changed in `wc_get_order`, but I couldn't pinpoint the exact change in core.
- `WC_Stripe_Co_Branded_CC_Compatibility_Test::test_is_wc_supported` is failing because we're comparing versions as strings. https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4517/commits/3a5f8265c987815f20ba9e4630f79f06098bb640 fixes it by using `version_compare`
- WC=L-1 and WP=L-2 are not a valid combination for the current versions. I've excluded them from the test matrix.

## Testing instructions
- Make sure the tests pass on GitHub Actions.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


.github/workflows/php-tests.yml
```diff
@@ -71,6 +71,8 @@ jobs:
         exclude: # incompatible versions of PHP, WordPress, and WooCommerce
           - woocommerce_support_policy: L
             wordpress_support_policy: L-2
+          - woocommerce_support_policy: L-1
+            wordpress_support_policy: L-2
```
malithsen, line 75: This combination cannot be installed

```
Installing WooCommerce (9.9.5)
...
Warning: The package could not be installed. "Your WordPress version is 6.6.2, however the uploaded plugin requires 6.7."
```
[logs](https://github.com/woocommerce/woocommerce-gateway-stripe/actions/runs/16453189349/job/46503498750#step:5:188)

tests/phpunit/WC_Stripe_Co_Branded_CC_Compatibility_Test.php
```diff
@@ -18,6 +18,6 @@ class WC_Stripe_Co_Branded_CC_Compatibility_Test extends WP_UnitTestCase {
 	 */
 	public function test_is_wc_supported() {
```
wjrosa, line 19: Another option here would be just to remove the test and update the main method, as the extension would not support versions below that anyway.

wjrosa (approved):
Changes look good to me! Thanks for fixing this as well. I left an alternative approach (totally optional, and your idea might be safer).

tests/phpunit/WC_Stripe_Co_Branded_CC_Compatibility_Test.php
daledupreez: I think it's worth keeping the test, as this failure seems to be driven by string comparisons in the test logic rather than a gap in the underlying code. So this feels like a legitimate bug in the test that we should fix.

daledupreez (approved):
Changes LGTM! I am going to merge this to get our tests running cleanly again.

As a quick note, I have also removed the three required actions from our branch protections for the `develop` branch, all of which use WP L-2 and WC L-1.

PR #4518, opened 2025-07-22
Fix setup intent webhook fatal

malithsen:
Fixes STRIPE-637

When processing the `setup_intent.setup_failed` webhook, OXXO, Multibanco and Boleto payment methods have a call to the `wc_stripe_allowed_payment_processing_statuses` hook to add an additional on-hold status. Prior to v9.7.0 `wc_stripe_allowed_payment_processing_statuses` hook never existed so the on-hold status was never added. In 9.7.0, we introduced `wc_stripe_allowed_payment_processing_statuses` (deprecating `wc_gateway_stripe_allowed_payment_processing_statuses`). With this change, we began to see fatal errors because the payment methods expected the hook to be called with the `$order` object as the second parameter. 

This PR fixes the fatals errors by passing the in `$order` parameter. It also removes the `wc_stripe_allowed_payment_processing_statuses` usage in OXXO, Multibanco and Boleto as it never worked prior to v9.7.0

## Testing instructions

- As a merchant, go to WC -> Setttings -> Subscriptions. Make sure "Allow $0 initial checkout without a payment method." setting is not enabled.
- As a shopper, purchase a free trial subscription.
- As a merchant go to WC -> Status -> Logs ->  `fatal errors` 
- In `develop` notice the `Uncaught ArgumentCountError` fatal error.
- In this branch, no fatal error should be logged.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -13,7 +13,7 @@
 * Dev - Fix failing test cases associated with WooCommerce 10.0.x
 * Fix - Prevent multiple save appearance AJAX calls on Block Checkout
 * Fix - Fix required field error message and PHP warning for custom checkout fields that don't have a label
-* Fix - Fix fatal when processing Boleto setup intents via webhooks
+* Fix - Fix fatal when processing setup intents for free subscriptions via webhooks
 
 = 9.7.0 - 2025-07-21 =
 * Update - Removes BNPL payment methods (Klarna and Affirm) when other official plugins are active
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,6 +9,7 @@
 * Add - Add state mapping for Lithuania in express checkout
 * Tweak - Use wp_ajax prefix for its built-in security for Add Payment Method action
 * Dev - Fix WooCommerce version fetching in GitHub workflows
+* Dev - Fix failing test cases associated with WooCommerce 10.0.x
 * Fix - Fix fatal when processing Boleto setup intents via webhooks
 
 = 9.7.0 - 2025-07-21 =
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
+* Add - Adds a new bulk action option to the subscriptions listing screen to check for detached payment methods
 * Dev - Use product type constants that were added in WooCommerce 9.7
 * Dev - Removes the inclusion of the deprecated WC_Stripe_Order class
 * Add - Introduces a new banner to promote the Optimized Checkout feature in the Stripe settings page for versions 9.8 and above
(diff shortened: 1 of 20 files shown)


includes/payment-methods/class-wc-gateway-stripe-boleto.php
```diff
@@ -106,22 +104,6 @@ public function update_unique_settings( WP_REST_Request $request ) {
...
-	 *
-	 * @return mixed
-	 */
-	public function add_allowed_payment_processing_statuses( $allowed_statuses, $order ) {
-		if ( $this->stripe_id === $order->get_meta( '_stripe_upe_payment_type' ) && ! in_array( OrderStatus::ON_HOLD, $allowed_statuses ) ) {
-			$allowed_statuses[] = OrderStatus::ON_HOLD;
-		}
-
-		return $allowed_statuses;
-	}
```
diegocurbelo: This function is public; we should keep it in case some merchants are using it for some custom code they might have implemented.

Removing the `add_filter` from the `__construct` is enough to keep the 9.6 behavior.

diegocurbelo (changes requested):
Thanks for fixing this @malithsen... 

The fix (adding the order) makes sense, and it tests well using the filter from the voucher constructors for example.

To maintain the same behavior as 9.6, removing the `add_filter` call makes sense. 
However, we should retain the `add_allowed_payment_processing_statuses()` functions, as they are defined as `public`, and merchants may be using them in custom code or snippets.

includes/payment-methods/class-wc-gateway-stripe-boleto.php
malithsen: Added the public methods back in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4518/commits/062b47fc993487d8e2d10e23cea1a6f3e113c667

malithsen:
Diego pointed out that `wc_stripe_allowed_payment_processing_statuses` is called for Boleto and other payment methods in other places, such as `process_payment_intent` and when handling exceptions in `process_payment`.  So adding back the filter back in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4518/commits/635746e7f9dc6862311dd9d650198e1664a75078.

changelog.txt
```diff
@@ -13,6 +13,7 @@
 * Dev - Fix failing test cases associated with WooCommerce 10.0.x
 * Fix - Prevent multiple save appearance AJAX calls on Block Checkout
 * Fix - Fix required field error message and PHP warning for custom checkout fields that don't have a label
+* Fix - Fix fatal when processing Boleto setup intents via webhooks
```
daledupreez: I think we should update the changelog to reflect that this could affect OXXO, Multibanco, and Boleto.
```suggestion
* Fix - Fix fatal when processing Boleto, OXXO, or Multibanco setup intents via webhooks
```

daledupreez (commented):
The change as written seems pretty simple, and should just work. I'd be OK shipping the change after we tweak the changelog entry.

changelog.txt
diegocurbelo: It affects any recurring payment method when purchasing a free subscription
```suggestion
* Fix - Fix fatal when processing setup intents for free subscriptions via webhooks
```

malithsen:
I've updated the PR description with steps to reproduce the error. h/t Diego!

diegocurbelo (approved):
The change looks good.

It's not going to change pre 9.7.0 behavior for order statuses, as only Boleto, OXXO and Multibanco add that filter; and none of them support the flow for the `setup_intent.succeeded` ($0 sub with attached payment method)

daledupreez (approved):
:shipit: 🚀

PR #4527, opened 2025-07-24
Temporarily disabling PHP Unit code coverage report workflow

wjrosa:
See p1753363142624709-slack-C055WHLA98D

We are temporarily disabling the GitHub action workflow that reports the code coverage % for PHP Unit tests to reduce noise. We will re-enable it again once we aim to add more unit tests and increase our current coverage.

## Testing instructions

Code review should be enough.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
+* Dev - Temporarily disabling the PHP Unit tests code coverage report Github Action to reduce noise
 * Add - Adds a new bulk action option to the subscriptions listing screen to check for detached payment methods
 * Dev - Use product type constants that were added in WooCommerce 9.7
 * Dev - Removes the inclusion of the deprecated WC_Stripe_Order class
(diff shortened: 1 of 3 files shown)


changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
+* Dev - Temporarily disabling the PHP Unit tests code coverage report Github Action to reduce noise
```
daledupreez, line 4: Nit: I think this is a moment where "temporarily" is almost certainly going to be inaccurate. 🙃 Given that we don't know when we will re-enable them, it's probably best to be clearer about the action we're taking without committing to re-enable them.
```suggestion
* Dev - Disable the PHP Unit tests code coverage report Github Action to reduce noise
```

daledupreez (commented):
This could work, but based on [this GitHub doc](https://docs.github.com/en/actions/how-tos/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow), I went ahead and disabled the workflow from within GitHub [here](https://github.com/woocommerce/woocommerce-gateway-stripe/actions/workflows/php-code-coverage.yml). So this PR may be unnecessary. 🙃

wjrosa:
Thanks, Dale! I will go ahead and close this then

PR #4529, opened 2025-07-24
Fix booking multiple slots with ECE

wjrosa:
Fixes STRIPE-623

This PR fixes issues with the Bookings extension when trying to add multiple slots to the cart. Currently, it is not possible to add more than one slot when the express checkout payment methods are enabled. This is because of the way we handle buttons on the product page. Also, some slots are not properly freed from cart when purchase using the product page.

## Testing instructions

- Checkout and build this branch on your test environment (`fix/booking-multiple-slots-with-ece`)
- Connect your Stripe account
- Enable express payment methods (Google Pay, Apple Pay, Link)
- Install the [Bookings extension](https://woocommerce.com/products/woocommerce-bookings/)
- Add a booking product

### Adding multiple slots

- As a shopper, attempt to add multiple days to your cart
- Go to your cart and confirm all of them are there

### Product page purchase after adding slots to the cart

- As a shopper, add multiple days to your cart
- Go to the bookings product page again
- Complete a purchase there using an express payment method
- Confirm the days you added previously to your cart are free again

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
@@ -143,7 +143,7 @@ class WC_Stripe_Express_Checkout_Ajax_Handler {
 		if ( $booking_id ) {
 			// When a bookable product is added to the cart, a 'booking' is created with status 'in-cart'.
 			// This status is used to prevent the booking from being booked by another customer
-			// and should be removed when the cart is emptied for ECE purposes.
+			// and should be removed when the cart is emptied for express checkout purposes.
 			do_action( 'wc-booking-remove-inactive-cart', $booking_id ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
 		}
 
--- a/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
@@ -103,8 +103,10 @@ class WC_Stripe_Express_Checkout_Ajax_Handler {
 		// When a bookable product is added to the cart, a 'booking' is created with status 'in-cart'.
 		// This status is used to prevent the booking from being booked by another customer
 		// and should be removed when the cart is emptied for ECE purposes.
-		foreach ( $booking_ids as $booking_id ) {
-			do_action( 'wc-booking-remove-inactive-cart', $booking_id ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
+		if ( has_action( 'wc-booking-remove-inactive-cart' ) ) { // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
+			foreach ( $booking_ids as $booking_id ) {
+				do_action( 'wc-booking-remove-inactive-cart', $booking_id ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
+			}
 		}
 
 		if ( ( ProductType::VARIABLE === $product_type || 'variable-subscription' === $product_type ) && isset( $_POST['attributes'] ) ) {
--- a/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
@@ -92,8 +92,20 @@ class WC_Stripe_Express_Checkout_Ajax_Handler {
 		$product      = wc_get_product( $product_id );
 		$product_type = $product->get_type();
 
+		$booking_ids = [];
+		if ( 'booking' === $product_type ) {
+			$booking_ids = $this->express_checkout_helper->get_booking_ids_from_cart();
+		}
+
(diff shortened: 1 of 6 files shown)


includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -92,9 +92,21 @@ public function ajax_add_to_cart() {
...
+
 		// First empty the cart to prevent wrong calculation.
 		WC()->cart->empty_cart();
 
+		// When a bookable product is added to the cart, a 'booking' is created with status 'in-cart'.
+		// This status is used to prevent the booking from being booked by another customer
+		// and should be removed when the cart is emptied for ECE purposes.
+		foreach ( $booking_ids as $booking_id ) {
+			do_action( 'wc-booking-remove-inactive-cart', $booking_id ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
+		}
```
daledupreez: Is it worth running through the loop when that action has hooks attached?
```suggestion
		if ( has_action( 'wc-booking-remove-inactive-cart' ) ) { // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
			foreach ( $booking_ids as $booking_id ) {
				do_action( 'wc-booking-remove-inactive-cart', $booking_id ); // phpcs:ignore WordPress.NamingConventions.ValidHookName.UseUnderscores
			}
		}
```

diegocurbelo (approved):
The changes look good, and it tests well 👏🏼 

Multiple bookings can now be added to the cart. And using the Express Payment Button from the product page, shows the current booking product total, and clears the cart (correctly freeing the reserved time slots from those bookings in the cart)

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -135,9 +141,9 @@ public function ajax_clear_cart() {
 		WC()->cart->empty_cart();
 
 		if ( $booking_id ) {
-			// When a bookable product is added to the cart, a 'booking' is create with status 'in-cart'.
+			// When a bookable product is added to the cart, a 'booking' is created with status 'in-cart'.
 			// This status is used to prevent the booking from being booked by another customer
-			// and should be removed when the cart is emptied for PRB purposes.
+			// and should be removed when the cart is emptied for ECE purposes.
```
daledupreez: Non-blocking nit: I think we should use the full term rather than `ECE`.
```suggestion
			// and should be removed when the cart is emptied for express checkout purposes.
```

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1669,18 +1669,38 @@ public function cart_prices_include_tax() {
...
+	 * Used to remove the booking from WC Bookings in-cart status.
+	 *
+	 * @return array
+	 */
+	public function get_booking_ids_from_cart() {
+		$cart        = WC()->cart->get_cart();
+		$booking_ids = [];
+
+		foreach ( $cart as $item ) {
+			if ( isset( $item['booking']['_booking_id'] ) ) {
```
daledupreez: Might `! empty()` be a more robust check here? Are there cases where the value could be empty?
```suggestion
			if ( ! empty( $item['booking']['_booking_id'] ) ) {
```

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -671,14 +671,47 @@ public function provide_test_get_stripe_currency_decimals() {
...
+		];
+
+		$helper = new WC_Stripe_Express_Checkout_Helper();
+		$actual = $helper->get_booking_ids_from_cart();
+
+		// Clean up.
+		WC()->session->cleanup_sessions();
+		WC()->cart->empty_cart();
+
+		$this->assertSame( [ $product->get_id() ], $actual );
```
daledupreez: Might it be worth running some more robust checks that include more than one booking product, and maybe some test cases that include a mix of product types? It might also be good to make sure we deduplicate the same booking ID appearing twice. I don't know if that is a realistic situation, but our `array_unique()` call would deduplicate that.

daledupreez (approved):
This is testing well and as described. I have some minor feedback that isn't blocking.

Separately, should we have a warning or similar to indicate that using an express checkout from the product page will remove all bookings currently in the cart? That is not an obvious expectation for me as a shopper, and I would find that frustrating, especially if I've spent time picking multiple slots and then choose to pay after making my final selection. Not something we should address in this PR, but it feels like the UX could be improved.

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
wjrosa: Alright. Added more test cases in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4529/commits/915f345e70d2190bd71a62a5530a14a23fdaf089

PR #4023, opened 2025-03-06
Adding the Single Payment Element feature to the block checkout

wjrosa:
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/3889

## Changes proposed in this Pull Request:

We are adding support for the new Single Payment Element feature on the block checkout. Basically, with this PR, all the payment methods configured on the Stripe dashboard will show up automatically on the block checkout.

This feature is being handled by the feature flag introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3988. Since all methods are displayed on a single container, I decided to name it "Stripe" when selecting the payment method (this may be changed later).

## Testing instructions

- Checkout this branch on your test environment (`add/spe-block-checkout-implementation3`)
- Connect your Stripe account
- On your Stripe dashboard, enable multiple payment methods
- Copy your child payment method configuration ID from here: `https://dashboard.stripe.com/settings/payment_methods` (labeled "Your configuration")
- Insert the ID above on `client/blocks/upe/upe-deferred-intent-creation/payment-elements.js:139`. This is temporary until the parent configuration, labeled WooCommerce Inc., supports all methods by default
- Build the frontend files (`npm run build:webpack`)
- Enable the SPE feature flag (`_wcstripe_feature_spe`). You can do it by either hardcoding the return value of `is_spe_available` to `true` or by running `npm run wp option update _wcstripe_feature_spe 'yes'`
- As a merchant, disable the legacy checkout experience in settings (`wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings`)
- Enable the Single Payment Element feature
- I recommend setting your store currency to EUR so you can have more methods available
- As a shopper, add any product to your cart
- Go to the block checkout page
- Confirm you can see all the payment methods rendered inside the SPE container:
Screenshot 2025-03-07 at 08 37 49

- Confirm you can complete the payment with any of the methods available

Limitations: The payment method instructions (on the top of the container) are still shown incorrectly. The saving of methods will not work as expected.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

[...]

PR #4530, opened 2025-07-25
Fix free trial subscription purchases with payment methods that require redirection

malithsen:
Fixes STRIPE-638

Remove Optimized Checkout related `return_url` handling to fix free-trial subscription purchases. It's not clear to me why these checks were needed originally, so I would appreciate additional regression checks with OC enabled.

## Testing instructions
1. As a merchant, make sure the Subscriptions plugin is installed and active.
2. Go to WC -> Settings -> Subscriptions. Make sure "Allow $0 initial checkout without a payment method." setting is not enabled.
3. Create a free trial subscription.
4. As a shopper, try to purchase the subscription using iDeal or Bancontact.
5. In **develop**  there will be an error at checkout.
6. In this branch, you should be able to purchase the subscription. 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-intent-controller.php
+++ b/includes/class-wc-stripe-intent-controller.php
@@ -1007,8 +1007,8 @@ class WC_Stripe_Intent_Controller {
 
 		$request = $this->maybe_add_mandate_options( $request, $payment_information['selected_payment_type'] );
 
-		// Does not set the return URL if Single Payment Element is enabled or if the request needs redirection.
-		if ( $this->get_upe_gateway()->is_oc_enabled() || $this->request_needs_redirection( $payment_method_types ) ) {
+		// Does not set the return URL if the request needs redirection.
+		if ( $this->request_needs_redirection( $payment_method_types ) ) {
 			$request['return_url'] = $payment_information['return_url'];
 		}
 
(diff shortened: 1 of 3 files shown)


includes/class-wc-stripe-intent-controller.php
```diff
@@ -1097,8 +1097,8 @@ public function create_and_confirm_setup_intent( $payment_information ) {
 			$request['confirm'] = 'false';
 		}
 
-		// Removes the return URL if Single Payment Element is not enabled or if the request doesn't need redirection.
-		if ( ( ! $this->get_upe_gateway()->is_oc_enabled() || ! $this->request_needs_redirection( $request['payment_method_types'] ) ) ) {
+		// Removes the return URL if the request doesn't need redirection.
+		if ( ( ! $this->request_needs_redirection( $request['payment_method_types'] ) ) ) {
```
wjrosa:
```suggestion
		if ( ! $this->request_needs_redirection( $request['payment_method_types'] ) ) {
```

wjrosa (approved):
Code is good and works as expected!

> It's not clear to me why these checks were needed originally, so I would appreciate additional regression checks with OC enabled.

I have introduced that change previously [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4023).  I performed some regression tests, and OC appears to be working correctly with your change (tested multiple payment methods). Maybe some other change we did recently removed the need for that condition 🤔

diegocurbelo (approved):
The code change looks good, and it tests well 🚢

malithsen:
> Maybe some other change we did recently removed the need for that condition 

That makes sense. Thanks for the regression tests!

PR #4535, opened 2025-07-28
Turn off Amazon Pay in PMC if feature flag is off

annemirasol:

Amazon Pay is still unreleased, behind a default-off feature flag. However, when it is enabled directly via PMC, and the store is on Optimized Checkout, Amazon Pay shows up as a payment method.

While I couldn't find a way to exclude a specific payment method for dynamic payment elements, a workaround this PR is proposing is to turn off Amazon Pay in the PMC on Stripe settings update.

_Note: we could also choose to disable Amazon Pay in the parent PMC, until we are ready to release. If we go that way, then this PR is not needed._

## Testing instructions
1. Enable Optimized Checkout for your store.
2. Via the Stripe dashboard, enable Amazon Pay.
3. As a shopper, notice that Amazon Pay appears as a payment method, even though it is still behind a feature flag.
4. Go to Stripe settings > Payment methods, and click "Save changes".
5. Verify that Amazon Pay is now disabled in the Stripe dashboard.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -282,8 +282,9 @@ class WC_Stripe {
 
 		add_action( 'init', [ $this, 'initialize_apple_pay_registration' ] );
 
-		// Check if other official plugins are active for Klarna or Affirm and deactivate those BNPLs if so.
-		add_action( 'init', [ $this, 'maybe_deactivate_bnpls' ] );
+		// Check for payment methods that should be deactivated, e.g. unreleased,
+		// BNPLs when official plugins are active, etc.
+		add_action( 'init', [ $this, 'maybe_deactivate_payment_methods' ] );
 	}
 
 	/**
@@ -390,8 +391,8 @@ class WC_Stripe {
 	public function update_prb_location_settings() {
 		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
 		$prb_locations   = isset( $stripe_settings['payment_request_button_locations'] )
-			? $stripe_settings['payment_request_button_locations']
-			: [];
+		? $stripe_settings['payment_request_button_locations']
+		: [];
 		if ( ! empty( $stripe_settings ) && empty( $prb_locations ) ) {
 			global $post;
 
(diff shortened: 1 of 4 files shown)


diegocurbelo (approved):
The code changes make sense, and tests well; it requires saving the Settings to disable the PM, but manually updating the PMC could be considered an edge case.

I think that disabling it at the parent PMC level (as you mentioned in the note) is the better and more consistent solution.

wjrosa (approved):
Looks good to me!

Similar to what Diego said above, another possibility would be to make something like I just did with BNPLs: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4515

annemirasol:
> another possibility would be to make something like I just did with BNPLs: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4515

@wjrosa I updated the PR to disable Amazon Pay on init, as suggested. I merged it with your code to avoid calling "PMC update" more than once. 

**Requesting a re-review as quite a bit has changed.**

--- 
**Updated test instructions**
1. Install and enable Affirm and Klarna official plugins.
2. Keep Amazon Pay feature flag off.
   - Using the Stripe Dev Tools plugin, -or-
   - `wp option delete _wcstripe_feature_amazon_pay`
3. Via Stripe dashboard, enable any combination of Affirm, Klarna and Amazon Pay.
4. Clear PMC cache.
   - Using the Stripe Dev Tools plugin, -or-
   - Visiting the wp-admin Stripe settings page should also trigger a PMC cache refresh
5. Visit any page to trigger `init`.
6. Refresh your Stripe dashboard and verify that the payment methods are no longer enabled.

readme.txt
```diff
@@ -127,5 +127,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Fix - Require credit cards to be enabled before Apple Pay and Google Pay can be enabled in PMC
 * Fix - Free trial subscription orders with payment methods that require redirection (eg: iDeal, Bancontact)
 * Tweak - Update checkout error message for invalid API key to be more generic and user-friendly
+* Tweak - Disable Amazon Pay in the PMC if it is still behind a feature flag
```
wjrosa:
```suggestion
* Tweak - Disable Amazon Pay in the merchant's Payment Method Configuration object if it is still behind a feature flag
```

includes/class-wc-stripe.php
```diff
@@ -862,23 +863,54 @@ public function initialize_status_page() {
...
 		$gateway = $this->get_main_stripe_gateway();
 		if ( ! is_a( $gateway, 'WC_Stripe_UPE_Payment_Gateway' ) ) {
 			return;
 		}
 
+		$payment_method_ids_to_disable = [];
 		$enabled_payment_methods       = $gateway->get_upe_enabled_payment_method_ids();
+
+		// Check for BNPLs that should be deactivated.
+		$payment_method_ids_to_disable = array_merge(
```
wjrosa, line 881: We could remove the first `array_merge` but I think it is better to read this way.

wjrosa (approved):
That's a nice refactor, Anne! I think it makes much more sense this way. And it opens the possibility of including more methods there when needed ✅

includes/class-wc-stripe.php
annemirasol: Thought of that too, I agree.

PR #4531, opened 2025-07-25
Fix 'Invalid Configuration' error when disabling card

annemirasol:
## Changes proposed in this Pull Request:
When you attempt to disable **Credit card / debit card** as a payment method while Apple Pay and Google Pay are enabled, you will get the following error in your logs:

```
Error: Invalid configuration. Apple Pay and Google Pay require Card to be enabled.: https://dashboard.stripe.com/acct_...
``` 

This PR enforces the card requirement before including Apple Pay and Google Pay in the PMC update.

## Testing instructions
1. Ensure your test site is PMC-enabled: `wp option patch update woocommerce_stripe_settings pmc_enabled 'yes'`
2. Enable cards, Apple Pay/Google Pay, and another payment method like CashApp.
3. Refresh the settings page.
4. Uncheck cards. Notice that the express checkout section gets hidden as cards are a requirement.
5. Click "Save changes."
   - In `develop`, you get the error described in your logs. On page refresh, you will see cards and Apple Pay/Google Pay are still enabled.
   - In this branch, no log errors and your changes should persist.

Note: a separate issue I noticed is that when PMC updates fail, we do not get any indication in the settings page. We get the same "Settings saved" success message, but on page refresh, the changes were not persisted. Documented in STRIPE-643

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4540, opened 2025-07-30
Blocking disabling of cards when OC is enabled

wjrosa:
Fixes STRIPE-644

Currently, the credit card payment method is required to display the single payment element when the Optimized Checkout is enabled. Since cards are no longer needed for the extension to work, that would break OC.

This PR follows a similar pattern to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4531, requiring cards to be enabled when OC is active.

Follow-up issues:
- STRIPE-645 - Block the checkbox and add a new pill to inform merchants of the limitation. This will go before 9.8
- STRIPE-646 - Make OC work without the need for the credit card payment method to be enabled. This will take time, probably after 9.8

## Testing instructions

- Checkout to this branch on your test environment (`fix/block-disabling-of-cards-when-oc-is-enabled`)
- Connect your Stripe account
- Make OC available (you can use the `wc_stripe_is_optimized_checkout_available` filter for that)
- Enable OC
- Attempt to disable the credit card payment method
- Save your settings
- Refresh the page
- Confirm credit card is still enabled

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
-* Fix - Require credit cards to be enabled when the Optimized Checkout is active
+* Fix - Force the card payment method to be enabled when the Optimized Checkout is enabled in the merchant's Payment Method Configuration
 * Update - Deactivates Affirm or Klarna when other official plugins are active in merchant's Payment Method Configuration
 * Fix - Fixes issues related to booking multiple slots with express checkout payment methods enabled
 * Fix - Update the Optimized Checkout promotional inbox note to link to the relevant section in the Stripe settings page
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
+* Fix - Require credit cards to be enabled when the Optimized Checkout is active
 * Update - Deactivates Affirm or Klarna when other official plugins are active in merchant's Payment Method Configuration
 * Fix - Fixes issues related to booking multiple slots with express checkout payment methods enabled
 * Fix - Update the Optimized Checkout promotional inbox note to link to the relevant section in the Stripe settings page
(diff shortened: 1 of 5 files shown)


malithsen (approved):
This change looks good to me. During testing, I noticed that if the card option is already disabled, we still allow merchants to enable OC. This action enables cards in the background, but it’s not clear to the user that cards were activated as a result of enabling OC, since they are on a different page. I suggest that we either disable OC when the card option is disabled or clearly indicate that enabling OC will also enable cards.
All good if you want to tackle that in STRIPE-645 or in a separate PR.

wjrosa:
> This change looks good to me. During testing, I noticed that if the card option is already disabled, we still allow merchants to enable OC. This action enables cards in the background, but it’s not clear to the user that cards were activated as a result of enabling OC, since they are on a different page. I suggest that we either disable OC when the card option is disabled or clearly indicate that enabling OC will also enable cards. All good if you want to tackle that in [STRIPE-645](https://linear.app/a8c/issue/STRIPE-645) or in a separate PR.

That's a good catch, Malith. I did refactor the whole implementation in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4540/commits/44b54bcf8a502d8a324a73688d7f63661540e2ce to follow a similar pattern we had in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4515 and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4535. I think it is safer this way and addresses the issue you found. Can you please review it again?

## Test instructions

- Checkout to the `develop` branch
- Disable credit cards
- Checkout to this branch
- Refresh any store page
- Confirm credit card is enabled again

diegocurbelo (approved):
The changes look good, and it tests as indicated.

My only concern is that we are not actually preventing users from disabling Credit Cards, we are just ignoring the selection when saving settings, which might be confusing. 
I see we have another PR to disable the checkbox and add a warning pill #4541; if we decide not to ship that one, then we should disable the checkbox here.

wjrosa:
Thanks for the review, @diegocurbelo! That PR also needs to go before the next release; it is an integral part of the update.

malithsen (approved):
Thanks for making the changes. Looks good, tests as expected.

PR #4541, opened 2025-07-30
Add pill to inform about card requirement when OC is enabled

wjrosa:
Fixes STRIPE-645
Base PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4540

This PR adds a new pill to notify merchants when the Optimized Checkout is enabled that the payment method cannot be disabled. Currently, it is required for OC to work correctly.

Preview:
Screenshot 2025-07-30 at 12 13 31

## Testing instructions

- Checkout to this branch on your test environment (`add/pill-to-inform-about-card-requirement-when-oc-is-enabled`)
- Connect your Stripe account
- Make OC available (you can use the `wc_stripe_is_optimized_checkout_available` filter for that)
- Enable OC
- Go to the payment methods list screen
- Check if you can see the new pill on top of the credit card method (compare with the preview above)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/components/payment-method-required-for-oc-pill/index.js
+++ b/client/components/payment-method-required-for-oc-pill/index.js
@@ -1,4 +1,3 @@
-/* global wc_stripe_settings_params */
 import { __, sprintf } from '@wordpress/i18n';
 import React from 'react';
 import styled from '@emotion/styled';
@@ -6,6 +5,7 @@ import interpolateComponents from 'interpolate-components';
 import { Icon, info } from '@wordpress/icons';
 import Popover from 'wcstripe/components/popover';
 import { PAYMENT_METHOD_CARD } from 'wcstripe/stripe-utils/constants';
+import { useIsOCEnabled } from 'wcstripe/data';
 
 const StyledPill = styled.span`
 	display: inline-flex;
@@ -46,11 +46,8 @@ const IconComponent = ( { children, ...props } ) => (
 );
 
 const PaymentMethodRequiredForOCPill = ( { id, label } ) => {
-	if (
-		id === PAYMENT_METHOD_CARD &&
-		// eslint-disable-next-line camelcase
-		wc_stripe_settings_params.is_oc_enabled
-	) {
+	const [ isOCEnabled ] = useIsOCEnabled();
+	if ( id === PAYMENT_METHOD_CARD && isOCEnabled ) {
 		return (
 			<StyledPill>
 				{ __( 'Required', 'woocommerce-gateway-stripe' ) }
--- a/client/components/payment-method-required-for-oc-pill/index.js
+++ b/client/components/payment-method-required-for-oc-pill/index.js
@@ -5,7 +5,10 @@ import styled from '@emotion/styled';
 import interpolateComponents from 'interpolate-components';
 import { Icon, info } from '@wordpress/icons';
 import Popover from 'wcstripe/components/popover';
-import { PAYMENT_METHOD_CARD } from 'wcstripe/stripe-utils/constants';
+import {
+	PAYMENT_METHOD_AFFIRM,
+	PAYMENT_METHOD_KLARNA,
+} from 'wcstripe/stripe-utils/constants';
 
 const StyledPill = styled.span`
 	display: inline-flex;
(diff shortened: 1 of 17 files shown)


client/components/payment-method-unavailable-due-conflict-pill/index.js
```diff
@@ -48,25 +45,22 @@ const IconComponent = ( { children, ...props } ) => (
 	</IconWrapper>
 );
 
-const PaymentMethodUnavailableDueConflictPill = ( { id, label } ) => {
+const PaymentMethodRequiredForOCPill = ( { id, label } ) => {
```
diegocurbelo: I think the changes in this file were intended for the new file `client/components/payment-method-required-for-oc-pill/index.js`

client/components/payment-method-required-for-oc-pill/index.js
```diff
@@ -0,0 +1,96 @@
...
+`;
+
+const IconComponent = ( { children, ...props } ) => (
+	<IconWrapper { ...props }>
+		<AlertIcon icon={ info } size="16" />
+		{ children }
+	</IconWrapper>
+);
+
+const PaymentMethodUnavailableDueConflictPill = ( { id, label } ) => {
```
diegocurbelo: I think this file got mixed with `client/components/payment-method-unavailable-due-conflict-pill/index.js`, the changes in that file should be moved here.

diegocurbelo (changes requested):
It seems that the contents of the existing `client/components/payment-method-unavailable-due-conflict-pill/index.js` and the new `client/payment-method-required-for-oc-pill/index.js` got swapped, which breaks the Affirm/Klarna not available due conflict pill.

Also, the disabled checkbox styles might be confusing. It looks the same when the payment method is `enabled and can't be disabled` as when they are `disabled but can't be enabled`, even with the new `Required` pill, they might incorrectly think that Cards is not enabled.
I don't think we have an "active but disabled" style for the checkbox, so maybe we can leave it enabled but make it not interactable?

Screenshot_2025-07-31_at_10_46_33

wjrosa:
Thanks for the review again, Diego!

> It seems that the contents of the existing client/components/payment-method-unavailable-due-conflict-pill/index.js and the new client/payment-method-required-for-oc-pill/index.js got swapped, which breaks the Affirm/Klarna not available due conflict pill.

Oops! Sorry about that 🤦 . I should not code during meetings. I have fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4541/commits/a3cde9319277ea3fa27469762aa499c35cd56edb

> I don't think we have an "active but disabled" style for the checkbox, so maybe we can leave it enabled but make it not interactable?

I tried to create a custom style for it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4541/commits/a2d8f685a11302d676eef1dc721b391fa0586c2c. Preview:
Screenshot 2025-07-31 at 11 50 51

But I think we should not diverge from the current standards, so I reverted it, and now it looks like a regular checked checkbox (but not interactive as you suggested).

client/settings/general-settings-section/payment-method-checkbox.js
```diff
@@ -44,9 +45,17 @@ const PaymentMethodCheckbox = ( {
...
 	const [ , setIsStripeEnabled ] = useIsStripeEnabled();
 	const { isUpeEnabled } = useContext( UpeToggleContext );
+	let checked;
+	if ( disabled ) {
+		checked = false;
+	} else if ( disabledButChecked ) {
+		checked = true;
+	} else {
+		checked = enabledPaymentMethods.includes( id );
+	}
```
diegocurbelo: [nit] This is more of personal preference... feel free to ignore it!
```suggestion
	const checked = !disabled && ( disabledButChecked || enabledPaymentMethods.includes( id ) );
```

diegocurbelo (approved):
Thanks for the change @wjrosa, it looks great! 🚢 

I like your custom "dimmed" style, but I'm also fine using the normal "checked" style, both make it clear that the payment method is enabled.

tests/phpunit/WC_Stripe_Test.php
```diff
@@ -6,6 +6,7 @@
 use WC_Stripe_Helper;
 use WC_Stripe_Payment_Methods;
 use WC_Stripe_UPE_Payment_Gateway;
+use WooCommerce\Stripe\Tests\Helpers\OC_Test_Helper;
```
malithsen: Is this import required?

malithsen (changes requested):
Nice work. I noticed that a refresh is needed for the checkbox to be disabled on the payment method list. Can we fix that?

https://github.com/user-attachments/assets/48d32f54-484b-4588-acb2-c6894462b046

malithsen (approved):
Actually, I'll change my previous review to approve this as the immediate update of the state is not critical for this particular checkbox but it would still be nice to fix it.

tests/phpunit/WC_Stripe_Test.php
wjrosa: Not really. Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4541/commits/4b5ddd3a168710e196c2b98ab69f565c42e9e797

wjrosa:
Thanks, Malith! I have removed the need for a refresh in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4541/commits/531d806118bfe8626c0fe3b9b300a28ccb886727

PR #4542, opened 2025-07-30
Only clear customer cache when an action has been performed

lipemat:

If we skip updating a payment token or get an error response from stripe, we don't want to clear the customer cache as nothing has changed.

Prevents endless loops of clearing the cache and then reloading the customer data.

Discovered when using the same stripe test card twice on the same customer. Different expiration dates, but the same card number return the same fingerprint and are detected by Woo as a duplicate. 

Endless loop of:
1. Retrieve tokens from stripe.
2. Store tokens in transient.
3. Add payment method is triggered for the second token with the same fingerprint.
4. Cache is cleared, clearing the transient from step 2.
5. Duplicate payment method is detected.
6. Update is skipped due to token existing from list of stripe tokens.
7. Update is skipped.
8. Repeats on every request.

## Testing instructions

1. As a customer, add a test payment method.
2. As the same customer, update the payment method using credit card form in My Account settings. Use the same card but a different expiration date.
3. Go to the My Account -> Payment Methods are on the site.
4. Refresh the page.
5. Verify the Stripe API was not called on the second load.

---

-   [x] Does not apply to mobile.

### Changelog entry

<details>

<summary>Only clear the customer cache when the data has been updated.</summary>

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -21,6 +21,7 @@
 * Fix - Free trial subscription orders with payment methods that require redirection (eg: iDeal, Bancontact)
 * Tweak - Update checkout error message for invalid API key to be more generic and user-friendly
 * Tweak - Disable Amazon Pay in the merchant's Payment Method Configuration object if it is still behind a feature flag
+* Fix - Only clear customer cache when an action has been performed
 
 = 9.7.1 - 2025-07-28 =
 * Add - Add state mapping for Lithuania in express checkout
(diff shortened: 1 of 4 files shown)


diegocurbelo (approved):
Thanks for contributing this PR @lipemat 👏🏼 

The code changes look good, and it tests well; the E2E tests fail during setup for this PR, but I tested them locally, and they all pass.

wjrosa (approved):
Code is good and works as expected! Thank you for the contribution 🙏

PR #4060, opened 2025-03-14
Track SPE enablement

malithsen:
Related to #3895 

## Changes proposed in this Pull Request:

This introduces the following two Tracks events that get emitted when enabling and disabling the SPE feature via settings.

- `wcadmin_wcstripe_spe_enabled`
- `wcadmin_wcstripe_spe_disabled`

## Testing instructions

1. Enable SPE feature flag by returning `true` [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/7d11977d63d06388313656030511e894fda253ed/includes/class-wc-stripe-feature-flags.php#L158).
2. As a merchant, enable SPE feature via WooCommerce -> Settings -> Payments -> Stripe -> Settings -> Advanced settings
3. In about 5 minutes, check MC Tracks Live events page for the presence of a `wcadmin_wcstripe_spe_enabled` event.
4. As a merchant, disable SPE feature.
5.  In about 5 minutes, check MC Tracks Live events page for the presence of a `wcadmin_wcstripe_spe_disabled` event.

CleanShot 2025-03-14 at 01 56 35

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4544, opened 2025-07-31
Tracking Optimized Checkout enablement (from the Promotional Banner)

wjrosa:
Fixes STRIPE-649

To complement https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4060, this PR will add tracking for the OC enablement events happening trought the OC toggle endpoint (used currently for the [OC promotional banner](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4495)).

## Testing instructions

- Checkout and build this branch on your test environment (`add/track-optimized-checkout-enablement`)
- Connect your Stripe account
- Make sure the Optimized Checkout is disabled
- Confirm you can see the promotional banner for OC
- Click the "Activate now" button
- In about 5 minutes, check MC Tracks Live events page for the presence of a `wcstripe_oc_enabled` event

PS: If you tested the promotional banner before and dismissed it, you may need to delete the `wc_stripe_show_oc_promotion_banner` option from your database

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/admin/class-wc-stripe-rest-oc-setting-toggle-controller.php
```diff
@@ -88,11 +88,20 @@ public function set_setting( WP_REST_Request $request ) {
...
+		$settings['optimized_checkout_element'] = $value;
 
 		WC_Stripe_Helper::update_main_stripe_settings( $settings );
 
+		if ( $value !== $current_value ) {
+			wc_admin_record_tracks_event(
+				$is_oc_enabled ? 'wcstripe_oc_enabled' : 'wcstripe_oc_disabled',
+				[ 'test_mode' => WC_Stripe_Mode::is_test() ? 1 : 0 ]
+			);
+		}
```
malithsen: Do we want to distinguish between the merchants who enable OC via the banner vs the checkbox? If so, we'd want to introduce another property to the event.

Also, AFAICS, there's no way to disable OC via the banner, so the presence of `wcstripe_oc_disabled` might be redundant at the moment, but I think it's still good to have it in case we end up using `set_setting` later on to disable it.

malithsen (approved):
LGTM, minor suggestion in the comments.

diegocurbelo (approved):
LGTM 🚢

includes/admin/class-wc-stripe-rest-oc-setting-toggle-controller.php
wjrosa: > Do we want to distinguish between the merchants who enable OC via the banner vs the checkbox? If so, we'd want to introduce another property to the event.

That's a good idea! Added the source identifier in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4544/commits/3226fcfba48ee52ff8a4c5127b3e953b0aa9a94f (based on the current calling rest endpoint).

> Also, AFAICS, there's no way to disable OC via the banner, so the presence of wcstripe_oc_disabled might be redundant at the moment, but I think it's still good to have it in case we end up using set_setting later on to disable it.

That was my reasoning as well 👍

PR #3553, opened 2024-10-28
Do not load ECE button if total amount is 0

Mayisha:
## Changes proposed in this Pull Request:
- Not loading the ECE button when your cart total amount is `0`. Because ECE expects the amount to be greater than `0`.

- Fixed the following console error.
``` Uncaught Typeerror: Can not read properties of undefined (reading 'wcSettings')```

## Testing instructions
- Create a free product.
- Add the product to your cart and go to the cart/checkout page.
- In `develop` branch, you should see the following console error.

Screenshot 2024-10-28 at 8 33 14 PM

- In this branch, you should not see any console errors.

**Note**: the branch name is misleading as I initially targeted a different fix. The trial subscription will be fixed in a separate issue as it needs some more investigation (It's still under spike in WooPayments as well).

PR #4557, opened 2025-08-04
Allow purchase of free trials with ECE

wjrosa:
Fixes STRIPE-240

Our current implementation of the Express Checkout Element does not support free trials for subscriptions. I am adding support for that here.

We are basically checking for free trials on the backend, providing this informantion to the frontend, and changing two payment element parameters based on that:
- `mode`: setting it to `subscription` ([docs](https://docs.stripe.com/js/elements_object/create_without_intent#stripe_elements_no_intent-options-mode))
- `paymentMethodCreation`:  setting it to `manual` ([docs](https://docs.stripe.com/js/elements_object/create_without_intent#stripe_elements_no_intent-options-paymentMethodCreation))

## Testing instructions

- Checkout and build this branch on your test environment (`add/allow-ece-purchases-for-free-trials`)
- Connect your Stripe account
- Enable some express payment methods
- Create a subscription product with a free trial period
- As a shopper, access the store using a public-accessible URL
- Add this subscription to your cart
- Go to any checkout page
- Confirm express checkout buttons are shown
- As a merchant, confirm you can renew the subscription in the WP-Admin

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.9.0 - xxxx-xx-xx =
+* Add - Allow the purchase of free trials using the Express Payment methods when the product does not require shipping
```
Mayisha, line 4: QQ: why is it supported only for the products that do not require shipping? Is there any catch with the shippable products?

client/blocks/express-checkout/index.js
```diff
@@ -83,10 +83,6 @@ const expressCheckoutElement = ( expressPaymentMethod, api ) => {
 	);
 	const edit = getEditorElement( expressPaymentMethod );
 	const canMakePayment = ( { cart } ) => {
-		if ( parseFloat( cart.cartTotals.total_price ) === 0.0 ) {
```
Mayisha, line 86: We should check something like this here as well-

```
const hasFreeTrial = getExpressCheckoutData( 'has_free_trial' );
```

We should keep the amount check for other free products and remove the check for free subscription only. 

I added a simple free product to my cart and I see this error on my block checkout page.

Screenshot 2025-09-08 at 2 56 41 PM

client/express-checkout/utils/check-payment-method-availability.js
```diff
@@ -29,7 +30,9 @@ export const checkPaymentMethodIsAvailable = memoize(
 				<Elements
 					stripe={ api.loadStripe() }
 					options={ {
-						mode: 'payment',
+						mode: getExpressCheckoutData( 'has_free_trial' )
+							? 'subscription'
+							: 'payment',
 						...( isManualPaymentMethodCreation( paymentMethod ) && {
```
Mayisha:
```suggestion
						...( isManualPaymentMethodCreation( paymentMethod, hasFreeTrial ) && {
```

Shouldn't we send the `hasFreeTrial` param here to set `paymentMethodCreation` as `manual` for free trial subs?

Mayisha (commented):
Thanks for adding this long awaiting feature @wjrosa 🎉 

I can see the ECE buttons for free trial subscriptions now on product, block cart and block checkout page. I could successfully purchase a free trial subscription using block checkout. However, I do not see the ECE button on classic cart and classic checkout pages. Could you check if this is the same for you?

malithsen (commented):
Nice work! I think Mayisha's review captured everything pretty well. This works well on the Blocks checkout. I am not certain how much of an effort it would be to support shortcode, but if we want to ship the current work first, let's update the changelog entry to reflect that.

changelog.txt
wjrosa: I am honestly not sure. The original comment was included by you [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3553). I just kept the limitation 😅 . However, maybe it is not needed at all (I can't imagine any reasons why it would be). Original code:
```php
// Don't show if the total price is 0.
		// ToDo: support free trials. Free trials should be supported if the product does not require shipping.
		if ( ( ! ( $this->is_pay_for_order_page() || $this->is_product() ) && 0.0 === (float) WC()->cart->get_total( false ) )
			|| ( $this->is_product() && 0.0 === (float) $this->get_product()->get_price() )
		) {
			return false;
		}
```

client/blocks/express-checkout/index.js
wjrosa: Good catch! Included the additional check in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4557/commits/0bd5184294b6b6fe400d264c803dfffe18282c4f

client/express-checkout/utils/check-payment-method-availability.js
wjrosa: Strangely, everything was working fine without the missing parameters like this one. We may not need some of those similar code blocks. Anyway, to keep it consistent, I have added it to all calls of this method in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4557/commits/80d57e3b235f40c2970e0998c16b72559d844e02

wjrosa:
Thanks for the review, folks! The shortcode checkout should be fixed by https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4557/commits/ea853ca0eaf3546fab5b38d87f22a2a24154384c

changelog.txt
Mayisha: Let me try to remember the reason 🤔

Mayisha: Okay, so I don't exactly remember the reason but I remembered following what we have in WooPayments. I found this [epic](https://github.com/Automattic/woocommerce-payments/issues/9983) where item 3 says the expected behavior is ECE would be hidden. This epic was created later but there were something in WooPayments that I followed at that time 🤔 
Not a blocker for this PR, but we should assess separately if support for shippable products should/can be added.

wjrosa: Thanks for the additional context! Yeah, I think we can tackle this specific rule later. I created STRIPE-715 for that

Mayisha (approved):
Thanks for the updates.

✅ ECE button visible on free trial subscription product page.
✅ ECE button visible on block cart and checkout pages with a free trial subscription in cart.
✅ ECE button visible on shortcode cart and checkout pages with a free trial subscription in cart.
✅ Can purchase a free trial subscription with Google Pay on block checkout page.
✅ Can purchase a free trial subscription with Google Pay on shortcode checkout page.
✅ Subscription renewal is successful.

malithsen (approved):
Nice work :shipit: 

Code looks good and tests as described

PR #4559, opened 2025-08-05
Removing the ability to edit the OC container title

wjrosa:
Towards STRIPE-656

As requested by the Stripe team, we are removing here the ability to customize the Optimized Checkout element title (shown above the payment methods on the checkout page).

## Testing instructions

- Checkout and build this branch on your test environment (`update/remove-the-ability-to-edit-oc-title`)
- Connect your Stripe account

### Setting

- Confirm that the input field to set the custom title is gone:
Screenshot 2025-08-05 at 17 16 36

### Block checkout

- Confirm the container title is always shown as "Stripe"
Screenshot 2025-08-05 at 17 18 14

### Classic/shortcode

- Confirm the same:
Screenshot 2025-08-05 at 17 17 36

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
-* Update - Removes the ability to change the title for the Optimized Checkout payment element, and use "Stripe" as the title
+* Update - Removes the ability to change the title for the Optimized Checkout payment element, as it is now set to "Stripe" by default
 * Add - Adds the current setting value for the Optimized Checkout to the Stripe System Status Report data
 * Add - A new pill to the payment methods page to indicate the credit card requirement when the Optimized Checkout feature is enabled
 * Add - Tracks the toggle of the Optimized Checkout feature in the promotional banner
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
+* Update - Removes the ability to change the title for the Optimized Checkout payment element, as it is now set to "Stripe" by default
 * Add - Adds the current setting value for the Optimized Checkout to the Stripe System Status Report data
 * Add - A new pill to the payment methods page to indicate the credit card requirement when the Optimized Checkout feature is enabled
 * Add - Tracks the toggle of the Optimized Checkout feature in the promotional banner
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
-* Update - Removes the ability to change the title for the Optimized Checkout payment element, as it is now set to "Stripe" by default
+* Update - Removes the ability to change the title for the Optimized Checkout payment element, and use "Stripe" as the title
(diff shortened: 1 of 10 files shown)


Mayisha (approved):
LGTM.

✅ Settings page does not offer the ability to set a custom title anymore.
✅ Checkout page (block & shortcode) always shows `Stripe` as the title.
✅ Previously set title is ignored.

client/classic/upe/index.js
```diff
@@ -315,11 +316,10 @@ jQuery( function ( $ ) {
 			// Change the payment method container title when the Optimized Checkout is enabled
 			if (
 				getStripeServerData()?.isOCEnabled &&
-				getStripeServerData()?.OCTitle &&
 				$( 'input#payment_method_stripe' ).is( ':checked' )
 			) {
 				$( 'label[for=payment_method_stripe]' ).text(
-					getStripeServerData()?.OCTitle
+					__( 'Stripe', 'woocommerce-gateway-stripe' )
```
daledupreez: Given that this is a brand, I don't think we should translate this string. If we end up expanding the value to something like `Stripe Optimized Checkout`, then I think we should translate.

includes/payment-methods/class-wc-stripe-upe-payment-method-cc.php
```diff
@@ -166,7 +166,7 @@ private function get_optimized_checkout_title( $payment_details = false ) {
 
 		// Block checkout and pay for order (checkout) page.
 		if ( ( has_block( 'woocommerce/checkout' ) || ! empty( $_GET['pay_for_order'] ) ) && ! is_wc_endpoint_url( 'order-received' ) ) { // phpcs:ignore WordPress.Security.NonceVerification
-			return $this->oc_title;
+			return __( 'Stripe', 'woocommerce-gateway-stripe' );
```
daledupreez: As above, I think this should not be translated.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.8.0 - xxxx-xx-xx =
+* Update - Removes the ability to change the title for the Optimized Checkout payment element, as it is now set to "Stripe" by default
```
daledupreez, line 4: It may be simpler to specify what string we are showing rather than talking about a default.
```suggestion
* Update - Removes the ability to change the title for the Optimized Checkout payment element, and use "Stripe" as the title
```

daledupreez (approved):
This is testing well, but I have some minor non-blocking comments.

I also want to flag that we're still discussing the eventual title, so I think we should change the `Fixes` keyword to `Towards` for the linked issue. We can close the issue once we've made a final decision on the title we want to use.

client/classic/upe/index.js
wjrosa: Removed the translation call in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4559/commits/69903826f1dccd7e61e3b9d92bb198bd3946e808

includes/payment-methods/class-wc-stripe-upe-payment-method-cc.php
wjrosa: Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4559/commits/69903826f1dccd7e61e3b9d92bb198bd3946e808

PR #4560, opened 2025-08-06
Reduce number of payment_methods API calls

daledupreez:
Fixes #<github_issue_id>

This PR tackles multiple pieces of code to try and reduce the number of API calls we make to the Stripe `payment_methods` API, which we may want to split up into smaller PRs:
 * Explicitly check for responses where the customer doesn't exist, and cache an empty result for that case
 * Refactor the code so we are fetching all payment methods for a customer, caching that result, while allowing for filtering on specific payment method types _after_ we fetch the result.
   - The vast majority (if not all) customers should have fewer than 100 saved payment methods, so this should result in 1 API call any time we have a cache miss, and at most 1 API call per request where we look for payment methods.
 * Refactor some loops that were fetching payment methods for each payment method until we found ones that we care about.

### Next steps and concerns

* **SEPA calls** - our logic for SEPA uses an undocumented `expand` URL parameter ([Stripe API docs for the API](https://docs.stripe.com/api/payment_methods/customer_list?lang=curl&api-version=2024-06-20)) -- I am not sure how this would (or would not) work across other payment methods
* **What can/should we include in 9.8.0?** - I _think_ we should add the "no customer" check for 9.8.0, as that should be an improvement that prevents nasty problems. I think we should seriously consider shipping only the `get_payment_methods()` API and the changes to callers that I have implemented in this change, but leave `get_payment_method()` untouched for now.
  - I am really open to feedback on what should (and should not) be included in 9.8.0!
* We may want to consider shifting to the database cache rather than transients for this data, but I haven't even looked at that yet.

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -723,17 +723,20 @@ class WC_Stripe_Customer {
 		$all_payment_methods = get_transient( $cache_key );
 
 		if ( false === $all_payment_methods || ! is_array( $all_payment_methods ) ) {
-			$all_payment_methods  = [];
+			$all_payment_methods    = [];
+			$last_payment_method_id = null;
 
-			while ( true ) {
-				$response = WC_Stripe_API::request(
-					[
-						'customer' => $this->get_id(),
-						'limit'    => 100,
-					],
-					'payment_methods',
-					'GET'
-				);
+			do {
+				$request_params = [
+					'customer' => $this->get_id(),
+					'limit'    => 100,
+				];
+
+				if ( $last_payment_method_id ) {
+					$request_params['starting_after'] = $last_payment_method_id;
+				}
+
+				$response = WC_Stripe_API::request( $request_params, 'payment_methods', 'GET' );
 
 				if ( ! empty( $response->error ) ) {
 					if (
@@ -754,10 +757,16 @@ class WC_Stripe_Customer {
 
 				$all_payment_methods = array_merge( $all_payment_methods, $response->data );
 
-				if ( ! isset( $response->has_more ) || ! $response->has_more ) {
-					break;
+				// Reset the last payment method ID so we can paginate correctly.
+				$last_payment_method_id = null;
+				if ( isset( $response->has_more ) && $response->has_more ) {
+					$last_payment_method = end( $response->data );
+
+					if ( $last_payment_method && ! empty( $last_payment_method->id ) ) {
+						$last_payment_method_id = $last_payment_method->id;
+					}
 				}
-			}
+			} while ( null !== $last_payment_method_id );
 
 			// Always cache the result without any filters applied.
 			set_transient( $cache_key, $all_payment_methods, DAY_IN_SECONDS );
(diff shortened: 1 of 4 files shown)


includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1016,69 +1016,69 @@ public function maybe_render_subscription_payment_method( $payment_method_to_dis
...
+				// Legacy handling for Stripe Card objects. ref: https://docs.stripe.com/api/cards/object
+				if ( isset( $source->object ) && WC_Stripe_Payment_Methods::CARD === $source->object ) {
+					/* translators: 1) card brand 2) last 4 digits */
+					$payment_method_to_display = sprintf( __( 'Via %1$s card ending in %2$s', 'woocommerce-gateway-stripe' ), ( isset( $source->brand ) ? wc_get_credit_card_type_label( $source->brand ) : __( 'N/A', 'woocommerce-gateway-stripe' ) ), $source->last4 );
+					break;
+				}
 
-					// Legacy handling for Stripe Card objects. ref: https://docs.stripe.com/api/cards/object
-					if ( isset( $source->object ) && WC_Stripe_Payment_Methods::CARD === $source->object ) {
+				switch ( $source->type ) {
```
wjrosa, line 1034: I think we have the opportunity to move this to a new method. `maybe_render_subscription_payment_method` is exceptionally long.

wjrosa:
I think it is worth creating specific issues for your "Next steps and concerns", so we can tackle those separately.

daledupreez:
> I think it is worth creating specific issues for your "Next steps and concerns", so we can tackle those separately.

@wjrosa, part of what I am looking for in terms of feedback is finding the right scope across the issues so we tackle the work appropriately. I've already split out the basic check for a missing customer in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4561, and I'll look into adding a new `get_all_payment_methods()` API without refactoring the existing  `get_payment_methods()` API, as I think we need to fix the two loops in the code that currently make multiple API calls.

includes/class-wc-stripe-customer.php
```diff
@@ -704,32 +704,90 @@ public function get_payment_methods( $payment_method_type ) {
...
+
+				if ( ! empty( $response->error ) ) {
+					if (
+						isset( $response->error->code )
+						&& isset( $response->error->param )
+						&& 'customer' === $response->error->param
+						&& 'resource_missing' === $response->error->code
+					) {
+						// If the customer doesn't exist, cache an empty array.
+						set_transient( $cache_key, [], DAY_IN_SECONDS );
```
malithsen, line 749: Should we use a smaller TTL for this negative cache?

daledupreez:
I am going to close this PR, as I have since split out the various implementation aspects into the following smaller PRs:
 * https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4561
 * https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4567
 * https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4568

PR #4567, opened 2025-08-08
Add method to fetch and cache all saved payment methods for a customer

daledupreez:
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4560

This change introduces a new `get_all_payment_methods()` method to the `WC_Stripe_Customer` class as a way to perform fewer overall API calls when trying to load all saved payment methods for a customer. The basic approach mimics the implementation in the existing `get_payment_methods()` method, but removes the query parameter to limit the accepted payment method types. Due to Stripe's API limiting the result page size to 100 rows, the new logic will fetch in a loop. The new code also fetches all the data and caches it before applying any filtering. That way the cached data can be re-used as needed, even if filters are re-applied. It's worth noting that we're using the same underlying mechanism for the cache as for specific payment methods, including the method responsible for clearing the customer cache.

The PR also updates the logic in `WC_Stripe_Payment_Tokens->woocommerce_get_customer_upe_payment_tokens()` to use the new method to fetch all saved tokens for a user across the set of currently enabled payment methods. The key improvement here is that we will now make at most 1 request per customer (or 2, if they have >100 saved payment methods!), and then cache the result. Previously, we would make an API call per enabled payment method, and we'd make more API calls when Optimized Checkout was enabled, as we weren't filtering on active payment methods. The code also makes an important code change to use a more recently introduced function to identify which payment methods are currently active, which also simplifies the code for identifying the active payment methods.

From an implementation perspective, it's worth noting that the new code always includes `expand[]=data.sepa_debit.generated_from.charge&expand[]=data.sepa_debit.generated_from.setup_attempt` as URL parameters, just in case we have any SEPA debits created from other payment methods like Bancontact or iDEAL.

### Basic setup

* Create a store with a location in Europe and using EUR as the currency
* Connect to a Stripe test account
* Ensure you install and activate WooCommerce Subscriptions and a tool like [Debug Bar](https://wordpress.org/plugins/debug-bar/)  that allows you to see API calls made from the server
* Ensure you have a subscription product available
* Ensure that you enable iDEAL or Bancontact as payment methods, keep SEPA disabled, and that the `Enable SEPA Direct Debit tokens for other methods` setting for the Stripe gateway is enabled

### Steps

* Log in to the site as a user (or ensure your site will auto-create a user for you during checkout)
* Work through the flow to purchase a subscription product
* When you get to checkout, pick either iDEAL or Bancontact as your payment method
* Place your order, and confirm the purchase via the Stripe UI
* After confirming the payment, verify that you're taken to the order summary page
* Navigate to _My Account_ and then to _Payment methods_
* Verify that your subscription payment method is listed as a SEPA payment method
* Now clear all transients for your site, which can be done as follows in a dev environment:
```bash
npm run wp transient delete -- --all
```
* Refresh the page, and verify that only one API call was made to `https://api.stripe.com/v1/payment_methods` (though it will have additional URL parameters).
* Refresh the page again, and verify that we don't make another outbound HTTP call
  - For extra credit, you can also check out `develop`, clear transients and repeat the steps to see how many API requests we make 😬 

* Now add a simple product to your cart and navigate to checkout
* Use a card to pay for the product, and save your card details
* Navigate to _My Account_ -> _Payment Methods_
* Verify that your saved card is listed
* Clear transients and refresh the page
* Verify there is only one `payment_methods` API call
* Refresh the page, and verify that we don't trigger another API call

* Now add a product to your cart and navigate to checkout.
* Clear transients again, and reload checkout.
* Verify that you only see one API call to the Stripe `payment_methods` API
* Reload checkout, and verify that you don't see any calls to the `payment_methods` API

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -940,6 +940,7 @@ class WC_Stripe_Customer {
 		foreach ( self::STRIPE_PAYMENT_METHODS as $payment_method_type ) {
 			delete_transient( self::PAYMENT_METHODS_TRANSIENT_KEY . $payment_method_type . $this->get_id() );
 		}
+		delete_transient( self::PAYMENT_METHODS_TRANSIENT_KEY . '__all_' . $this->get_id() );
 		// Clear cache for the specific payment method if provided.
 		if ( $payment_method_id ) {
 			WC_Stripe_Database_Cache::delete( 'payment_method_for_source_' . $payment_method_id );
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -775,8 +775,7 @@ class WC_Stripe_Customer {
 					$request_params['starting_after'] = $last_payment_method_id;
 				}
 
-				// TODO: work out how we can add the SEPA `expand[]` parameters to the request. Will that work when all types are being returned?
-				$response = WC_Stripe_API::request( $request_params, 'payment_methods', 'GET' );
+				$response = WC_Stripe_API::request( $request_params, 'payment_methods?expand[]=data.sepa_debit.generated_from.charge&expand[]=data.sepa_debit.generated_from.setup_attempt', 'GET' );
 
 				if ( ! empty( $response->error ) ) {
 					if (
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -31,6 +31,11 @@ class WC_Stripe_Customer {
 	];
 
 	/**
+	 * The maximum value for the `limit` argument in the Stripe payment_methods API.
+	 */
+	protected const PAYMENT_METHODS_API_LIMIT = 100;
+
+	/**
 	 * Stripe customer ID
 	 *
 	 * @var string
@@ -712,7 +717,7 @@ class WC_Stripe_Customer {
 				[
 					'customer' => $this->get_id(),
 					'type'     => $payment_method_type,
-					'limit'    => 100, // Maximum allowed value.
+					'limit'    => self::PAYMENT_METHODS_API_LIMIT,
 				],
 				'payment_methods' . $params,
 				'GET'
@@ -739,6 +744,95 @@ class WC_Stripe_Customer {
 		}
 
 		return empty( $payment_methods ) ? [] : $payment_methods;
+	}
+
+	/**
+	 * Get all payment methods for a customer.
+	 *
+	 * @param string[] $payment_method_types The payment method types to look for using Stripe method IDs. If the array is empty, it implies all payment method types.
+	 * @param int      $limit                The maximum number of payment methods to return. If the value is -1, no limit is applied.
+	 * @return array
+	 */
+	public function get_all_payment_methods( array $payment_method_types = [], int $limit = -1 ) {
+		if ( ! $this->get_id() ) {
+			return [];
+		}
+
+		$cache_key = self::PAYMENT_METHODS_TRANSIENT_KEY . '__all_' . $this->get_id();
+		$all_payment_methods = get_transient( $cache_key );
+
+		if ( false === $all_payment_methods || ! is_array( $all_payment_methods ) ) {
+			$all_payment_methods    = [];
+			$last_payment_method_id = null;
+
+			do {
+				$request_params = [
+					'customer' => $this->get_id(),
+					'limit'    => self::PAYMENT_METHODS_API_LIMIT,
+				];
+
+				if ( $last_payment_method_id ) {
+					$request_params['starting_after'] = $last_payment_method_id;
+				}
+
+				// TODO: work out how we can add the SEPA `expand[]` parameters to the request. Will that work when all types are being returned?
+				$response = WC_Stripe_API::request( $request_params, 'payment_methods', 'GET' );
+
+				if ( ! empty( $response->error ) ) {
+					if (
+						isset( $response->error->code )
+						&& isset( $response->error->param )
+						&& 'customer' === $response->error->param
+						&& 'resource_missing' === $response->error->code
+					) {
+						// If the customer doesn't exist, cache an empty array.
+						set_transient( $cache_key, [], DAY_IN_SECONDS );
+					}
+					return [];
+				}
+
+				if ( ! is_array( $response->data ) || [] === $response->data ) {
+					break;
+				}
+
+				$all_payment_methods = array_merge( $all_payment_methods, $response->data );
+
+				// Reset the last payment method ID so we can paginate correctly.
+				$last_payment_method_id = null;
+				if ( isset( $response->has_more ) && $response->has_more ) {
+					$last_payment_method = end( $response->data );
+
+					if ( $last_payment_method && ! empty( $last_payment_method->id ) ) {
+						$last_payment_method_id = $last_payment_method->id;
+					}
+				}
+			} while ( null !== $last_payment_method_id );
+
+			// Always cache the result without any filters applied.
+			set_transient( $cache_key, $all_payment_methods, DAY_IN_SECONDS );
+		}
+
+		// If there are no payment methods, no need to apply any filters below.
+		if ( [] === $all_payment_methods ) {
+			return $all_payment_methods;
+		}
+
+		// Only apply the limit and type filters after fetching and caching all payment methods.
+		$filtered_payment_methods = $all_payment_methods;
+		if ( [] !== $payment_method_types ) {
+			$filtered_payment_methods = array_filter(
+				$filtered_payment_methods,
+				function ( $payment_method ) use ( $payment_method_types ) {
+					return in_array( $payment_method->type, $payment_method_types, true );
+				}
+			);
+		}
+
+		if ( $limit > 0 ) {
+			return array_slice( $filtered_payment_methods, 0, $limit );
+		}
+
+		return $filtered_payment_methods;
 	}
 
 	/**
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -781,8 +781,7 @@ class WC_Stripe_Customer {
 
 				if ( ! empty( $response->error ) ) {
 					if (
-						isset( $response->error->code )
-						&& isset( $response->error->param )
+						isset( $response->error->param, $response->error->code )
 						&& 'customer' === $response->error->param
 						&& 'resource_missing' === $response->error->code
 					) {
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -32,6 +32,8 @@ class WC_Stripe_Customer {
 
 	/**
 	 * The maximum value for the `limit` argument in the Stripe payment_methods API.
+	 *
+	 * @see https://docs.stripe.com/api/payment_methods/customer_list#list_customer_payment_methods-limit
 	 */
 	protected const PAYMENT_METHODS_API_LIMIT = 100;
 
--- a/includes/payment-tokens/class-wc-stripe-payment-tokens.php
+++ b/includes/payment-tokens/class-wc-stripe-payment-tokens.php
@@ -301,19 +301,20 @@ class WC_Stripe_Payment_Tokens {
 
 			// Retrieve the payment methods for the enabled reusable gateways.
 			$payment_methods = [];
+
+			$reusable_payment_method_types = array_keys( self::UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD );
+			$active_payment_method_types = [];
+
 			if ( $gateway->is_oc_enabled() ) {
-				// For OC, get all available payment method types
-				foreach ( self::UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD as $payment_method_type => $reausable_gateway_id ) {
+				// For Optimized Checkout, get all available payment method types.
+				foreach ( $reusable_payment_method_types as $payment_method_type ) {
 					$payment_method_instance = WC_Stripe_UPE_Payment_Gateway::get_payment_method_instance( $payment_method_type );
 					if ( $payment_method_instance ) {
-						$retrieved_methods = $customer->get_payment_methods( $payment_method_type );
-						if ( ! empty( $retrieved_methods ) ) {
-							$payment_methods[] = $retrieved_methods;
-						}
+						$active_payment_method_types[] = $payment_method_type;
 					}
 				}
 			} else {
-				foreach ( self::UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD as $payment_method_type => $reausable_gateway_id ) {
+				foreach ( $reusable_payment_method_types as $payment_method_type ) {
 					// The payment method type doesn't match the ones we use. Nothing to do here.
 					if ( ! isset( $gateway->payment_methods[ $payment_method_type ] ) ) {
 						continue;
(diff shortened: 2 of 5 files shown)


includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -301,42 +301,33 @@ public function woocommerce_get_customer_upe_payment_tokens( $tokens, $user_id,
...
-				} elseif ( ! $gateway->payment_methods[ WC_Stripe_UPE_Payment_Method_Sepa::STRIPE_ID ]->is_enabled()
-						&& ( $gateway->payment_methods[ WC_Stripe_UPE_Payment_Method_Ideal::STRIPE_ID ]->is_enabled()
-							|| $gateway->payment_methods[ WC_Stripe_UPE_Payment_Method_Bancontact::STRIPE_ID ]->is_enabled() ) ) {
-
-						$payment_methods[] = $customer->get_payment_methods( WC_Stripe_UPE_Payment_Method_Sepa::STRIPE_ID );
-				}
-			}
+			$payment_methods = $customer->get_all_payment_methods( $active_payment_method_types );
 
 			$payment_methods    = array_merge( ...$payment_methods );
```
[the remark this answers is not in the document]
daledupreez: Reinstated in a refactored way in b733cd3fee5a460379db08cf7d46c163669becf6.

includes/class-wc-stripe-customer.php
```diff
@@ -741,6 +746,94 @@ public function get_payment_methods( $payment_method_type ) {
...
+				$request_params = [
+					'customer' => $this->get_id(),
+					'limit'    => self::PAYMENT_METHODS_API_LIMIT,
+				];
+
+				if ( $last_payment_method_id ) {
+					$request_params['starting_after'] = $last_payment_method_id;
+				}
+
+				$response = WC_Stripe_API::request( $request_params, 'payment_methods?expand[]=data.sepa_debit.generated_from.charge&expand[]=data.sepa_debit.generated_from.setup_attempt', 'GET' );
```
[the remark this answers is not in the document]
daledupreez: I would love to, but that should be tackled in a follow-up PR.

includes/class-wc-stripe-customer.php
```diff
@@ -741,6 +746,94 @@ public function get_payment_methods( $payment_method_type ) {
...
+	 * @param string[] $payment_method_types The payment method types to look for using Stripe method IDs. If the array is empty, it implies all payment method types.
+	 * @param int      $limit                The maximum number of payment methods to return. If the value is -1, no limit is applied.
+	 * @return array
+	 */
+	public function get_all_payment_methods( array $payment_method_types = [], int $limit = -1 ) {
+		if ( ! $this->get_id() ) {
+			return [];
+		}
+
+		$cache_key = self::PAYMENT_METHODS_TRANSIENT_KEY . '__all_' . $this->get_id();
```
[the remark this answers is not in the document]
daledupreez: I think our long-term plan here should just be to use one cache, as most users don't have a large number of saved payment methods, and I think most of our use cases revolve around all saved payment methods rather than saved payment methods for only one particular type of payment.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -301,44 +301,48 @@ public function woocommerce_get_customer_upe_payment_tokens( $tokens, $user_id,
...
+				foreach ( $reusable_payment_method_types as $payment_method_type ) {
 					$payment_method_instance = WC_Stripe_UPE_Payment_Gateway::get_payment_method_instance( $payment_method_type );
 					if ( $payment_method_instance ) {
-						$retrieved_methods = $customer->get_payment_methods( $payment_method_type );
-						if ( ! empty( $retrieved_methods ) ) {
-							$payment_methods[] = $retrieved_methods;
-						}
+						$active_payment_method_types[] = $payment_method_type;
 					}
 				}
```
daledupreez: @wjrosa, can you comment on why we're fetching all possible saved payment methods when Optimized Checkout is enabled? If we have a set of enabled payment methods, I am not sure how we would suddenly end up with additional payment methods enabled. Should we not stick to the same logic as for the non-Optimized Checkout case?

wjrosa: I think this was here due to the first implementation, in which `$gateway->payment_methods` had only the card payment method (before the PMC project, only the card method was shown in the settings screen as well). I did some tests now, and we are safe to remove this condition without regressions. Good catch!

daledupreez: I've refactored the logic to rely on the payment method configuration regardless of Optimized Checkout configuration: 6c9eb2f3e1d522b0aa70c74a458f4b1e6618f6a4

includes/class-wc-stripe-customer.php
```diff
@@ -30,6 +30,11 @@ class WC_Stripe_Customer {
 		WC_Stripe_UPE_Payment_Method_Becs_Debit::STRIPE_ID,
 	];
 
+	/**
+	 * The maximum value for the `limit` argument in the Stripe payment_methods API.
+	 */
+	protected const PAYMENT_METHODS_API_LIMIT = 100;
```
wjrosa, line 38: Nice!

includes/class-wc-stripe-customer.php
```diff
@@ -741,6 +746,94 @@ public function get_payment_methods( $payment_method_type ) {
...
+	 * @param int      $limit                The maximum number of payment methods to return. If the value is -1, no limit is applied.
+	 * @return array
+	 */
+	public function get_all_payment_methods( array $payment_method_types = [], int $limit = -1 ) {
+		if ( ! $this->get_id() ) {
+			return [];
+		}
+
+		$cache_key = self::PAYMENT_METHODS_TRANSIENT_KEY . '__all_' . $this->get_id();
+		$all_payment_methods = get_transient( $cache_key );
```
wjrosa, line 764: Shouldn't we use the `WC_Stripe_Database_Cache` here?

includes/class-wc-stripe-customer.php
```diff
@@ -741,6 +746,94 @@ public function get_payment_methods( $payment_method_type ) {
...
+					$request_params['starting_after'] = $last_payment_method_id;
+				}
+
+				$response = WC_Stripe_API::request( $request_params, 'payment_methods?expand[]=data.sepa_debit.generated_from.charge&expand[]=data.sepa_debit.generated_from.setup_attempt', 'GET' );
+
+				if ( ! empty( $response->error ) ) {
+					if (
+						isset( $response->error->code )
+						&& isset( $response->error->param )
+						&& 'customer' === $response->error->param
```
wjrosa, line 785:
```suggestion
						&& 'customer' === ( $response->error->param ?? null )
```
I would code it like this to reduce the number of lines in the condition, but your way might be easier to read

includes/class-wc-stripe-customer.php
```diff
@@ -741,6 +746,94 @@ public function get_payment_methods( $payment_method_type ) {
...
+						&& 'resource_missing' === $response->error->code
+					) {
+						// If the customer doesn't exist, cache an empty array.
+						set_transient( $cache_key, [], DAY_IN_SECONDS );
```
wjrosa, line 789: Same question about the `WC_Stripe_Database_Cache`

includes/class-wc-stripe-customer.php
```diff
@@ -741,6 +746,94 @@ public function get_payment_methods( $payment_method_type ) {
...
+					}
+					return [];
+				}
+
+				if ( ! is_array( $response->data ) || [] === $response->data ) {
```
wjrosa, line 794: Since this is getting lengthy, I would consider creating new private methods to handle different parts of the logic.

includes/class-wc-stripe-customer.php
```diff
@@ -741,6 +746,94 @@ public function get_payment_methods( $payment_method_type ) {
...
+				}
+
+				$all_payment_methods = array_merge( $all_payment_methods, $response->data );
+
+				// Reset the last payment method ID so we can paginate correctly.
+				$last_payment_method_id = null;
+				if ( isset( $response->has_more ) && $response->has_more ) {
+					$last_payment_method = end( $response->data );
+
+					if ( $last_payment_method && ! empty( $last_payment_method->id ) ) {
```
wjrosa, line 805: If we just set `$last_payment_method_id` when `->id` is not empty, I wonder if we will even fall into the `while` condition 🤔

diegocurbelo:
This looks good @daledupreez!

My only concern is the use of transients; given the problems we had with the PMC cache, maybe we should use the DB cache instead.

And as you mentioned in Slack, we should also implement a scheduled task to remove expired items from the cache, since we use the `customer_id` as part of the cache key. But I don't think this is a blocker, the `wp_options` table has an index by `option_name`, and for both transients and the database cache, we can provide a snippet to clear expired items if needed.

We need to improve the database cache to fully support this use case... currently it stores the `ttl` and `updated` timestamp inside the option_value, which means reading them all, checking if they are expired, and deleting those that are. Transients, for example, use an additional option entry for the expiration timestamp, so it's possible to get the expired entries via a SQL query. Maybe the most flexible solution is to have both (saving an additional expiration entry automatically in `write_to_cache`)

includes/class-wc-stripe-customer.php
daledupreez: We could, but I was wary of adding new code and adding new data to the database cache before we have a cleanup mechanism (which I created an issue for earlier today in https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4569).

For large stores with regular customers, using the database cache would result in a big jump in the options size, which could easily drive other performance issues on sites.

includes/class-wc-stripe-customer.php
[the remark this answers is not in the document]
daledupreez: Yes, this is messy. But I think we need to clean it up in a separate PR, as it requires broader testing of all code that calls the Stripe APIs.

includes/class-wc-stripe-customer.php
daledupreez: I've shortened this slightly in aaa7ba7e08bd0d2b59bbdfb2cf4e7ac20498d071 to use the multi-variable form of `isset()`, and also specified the order of the parameters in the order that we going to check them. (Edited to pick up link.)

includes/class-wc-stripe-customer.php
```diff
@@ -741,6 +746,94 @@ public function get_payment_methods( $payment_method_type ) {
...
+				if ( ! is_array( $response->data ) || [] === $response->data ) {
+					break;
+				}
+
+				$all_payment_methods = array_merge( $all_payment_methods, $response->data );
+
+				// Reset the last payment method ID so we can paginate correctly.
+				$last_payment_method_id = null;
+				if ( isset( $response->has_more ) && $response->has_more ) {
+					$last_payment_method = end( $response->data );
```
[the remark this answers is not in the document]
daledupreez: `$response->data` is not going to be re-used, so I think this is safe. I prefer not assuming anything about array keys to be more defensive.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -301,44 +301,48 @@ public function woocommerce_get_customer_upe_payment_tokens( $tokens, $user_id,
 
 			// Retrieve the payment methods for the enabled reusable gateways.
 			$payment_methods = [];
+
+			$reusable_payment_method_types = array_keys( self::UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD );
```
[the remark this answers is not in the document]
daledupreez: That feels unnecessary to me, as it doesn't make it much easier to read, and we're clearly relying on the constant as the source of truth. So this feels like a marginal benefit, at least for now.

includes/class-wc-stripe-customer.php
daledupreez: I think we're OK, but I'll share my reasoning here (which may make sense as an inline comment).
* We start by (re)setting `$last_payment_method_id = null` to make sure we're acting on the current response.
* We then check if the current response has the `has_more` flag set to a truthy value (though it should be `true` per [the API docs](https://docs.stripe.com/api/payment_methods/customer_list)).
  - If `has_more` is falsey, we don't do anything, so `$last_payment_method_id` stays null and we fall out of the `do/while` loop.
  - If `has_more` is truthy, we then find the last payment method element in `$response->data` using `end()` so we're agnostic to the array key structure.
    * If the last payment method has an `id` (which it should in all cases), we set `$last_payment_method_id` to that value. This should trigger another API request, as we have a non-null value for the `while` condition.
    * If the last payment method doesn't have an `id` (which shouldn't happen), we leave `$last_payment_method_id` as `null`, and we fail the `while` condition. I think this is OK because we didn't have the data we needed to make further API calls anyway.

includes/class-wc-stripe-customer.php
daledupreez: I also want to pull @diegocurbelo into this thread as he flagged the persistent cache usage as well in his review.

I've implemented calling `WC_Stripe_Database_Cache` in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4570, mostly so we can merge that change into this branch if we decide to go in that direction.

includes/class-wc-stripe-customer.php
wjrosa: Makes sense! Thanks for the explanation!

wjrosa (approved):

daledupreez:
> My only concern is the use of transients; given the problems we had with the PMC cache, maybe we should use the DB cache instead.

@diegocurbelo, I agree that this could be problematic, but we're still reducing the number of API calls considerably with this implementation, _even if transients are not working_. I am also concerned about causing the options data size to balloon, even if we're explicitly flagging these items with `autoload = false` in our caching layer.

So I went with the more conservative approach of continuing to use a transient and only changing the API we call.

includes/class-wc-stripe-customer.php
diegocurbelo: > For large stores with regular customers, using the database cache would result in a big jump in the options size, which could easily drive other performance issues on sites.

Expired transients are not automatically removed from the DB; that requires custom code (calling `delete_expired_transients ()`) or using a plugin... and each transient adds two items to the options table (one with the data, and one with the expiration), the DB cache uses one so it would be half the amount of entries than using transients.

I can create a quick PR that adds an additional options entry for the expiration timestamp in the private `write_to_cache()` method, and then add a new public method `delete_expired()` to `WC_Stripe_Database_Cache` similar to [this one](https://developer.wordpress.org/reference/functions/delete_expired_transients/). And schedule it to run every 24 hours (filterable).

This would generate the same number of options entries as using transients, and provide a way for custom code to remove expired cache items, what do you think @daledupreez?

diegocurbelo:
> > My only concern is the use of transients; given the problems we had with the PMC cache, maybe we should use the DB cache instead.
> 
> @diegocurbelo, I agree that this could be problematic, but we're still reducing the number of API calls considerably with this implementation, _even if transients are not working_. I am also concerned about causing the options data size to balloon, even if we're explicitly flagging these items with `autoload = false` in our caching layer.
> 
> So I went with the more conservative approach of continuing to use a transient and only changing the API we call.

Sounds good, we can migrate it to use the DB cache in the next release, after we implement a cache cleanup solution.

includes/class-wc-stripe-customer.php
daledupreez: For now, I am concerned about timelines and stability, and what changes should (and should not) be included in 9.8.0. Maybe it makes sense to have a quick sync with the team to work things out?

malithsen (approved):
Code looks good to me and tests as expected. Only one API call is made to fetch the payment methods and it's cached on subsequent page refreshes.

Image

PR #4568, opened 2025-08-08
Refactor WC_Stripe_Subscriptions_Trait->maybe_render_subscription_payment_method()

daledupreez:

This PR refactors the code in `WC_Stripe_Subscriptions_Trait->maybe_render_subscription_payment_method()`. There are a few aspects to the refactor that are worth capturing in more detail:
* We now fetch only the specific payment method saved with the current subscription via the internal `WC_Stripe_API::get_payment_method()` API so we can handle older `src_` IDs.
* We use a static variable cache to avoid unnecessary API calls within the same request, as I saw two API requests triggered when viewing the Subscription details page
* The logic to translate the payment method details to a string has been extracted into a helper function that is simpler than what we had.
* I've also expanded the unit tests to cover more payment method types

### Open questions

* ~I am not sure how/if we need to handle older `src_*` sources. I don't know how Stripe handles those via the `payment_methods` APIs.~ As per [this suggestion](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4568#discussion_r2281595941) from @Mayisha, I've switched to calling `WC_Stripe_API::get_payment_method()` and then check the customer ID afterwards.
* I don't know if it's worth using a transient for the payment method caching, or whether we should just cache the payment method without the customer ID dimension. I think both _should be_ unique.

## Testing instructions

* Set up a store with WooCommerce Subscriptions installed and Stripe is connected
* Ensure you have multiple payment methods that support recurring payments enabled
* Ensure that you have a plugin like [Debug Bar](https://wordpress.org/plugins/debug-bar/) installed or that you have logging enabled for the Stripe plugin
* While logged in, purchase a subscription
* While running on `develop`, navigate to _My account_ > _Subscriptions_ > Your just-purchased subscription (or _My Subscription_ if you only have one subscription)
* Note what is displayed for the "Payment" row (see the screenshot below)
* Check your logs and verify that you see multiple API calls to the Stripe `/v1/payment_methods` API
* Switch to this branch -- `git checkout refactor/subscriptions-trait-payment-method-rendering`
* Reload the subscription page
* Verify that you have the same value shown for the "Payment" element
* Verify that only one API call was made to the Stripe `/v1/payment_methods/:payment_method_id` endpoint
* If you have an older `src_*` payment method ID, it would be good to test that, too!

Screenshot 2025-08-15 at 12 46 09

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/compat/trait-wc-stripe-subscriptions.php
+++ b/includes/compat/trait-wc-stripe-subscriptions.php
@@ -1068,63 +1068,63 @@ trait WC_Stripe_Subscriptions_Trait {
 	/**
 	 * Helper function to get the descriptive text for a payment method or source.
 	 *
-	 * @param object $source The payment method or source object.
+	 * @param object $payment_method The payment method or source object.
 	 * @return string The descriptive text for the payment method or source.
 	 */
-	protected function get_payment_method_to_display_for_payment_method( object $source ): string {
+	protected function get_payment_method_to_display_for_payment_method( object $payment_method ): string {
 		// Legacy handling for Stripe Card objects. ref: https://docs.stripe.com/api/cards/object
-		if ( isset( $source->object ) && WC_Stripe_Payment_Methods::CARD === $source->object ) {
+		if ( isset( $payment_method->object ) && WC_Stripe_Payment_Methods::CARD === $payment_method->object ) {
 			return sprintf(
 				/* translators: 1) card brand 2) last 4 digits */
 				__( 'Via %1$s card ending in %2$s', 'woocommerce-gateway-stripe' ),
-				( isset( $source->brand ) ? wc_get_credit_card_type_label( $source->brand ) : __( 'N/A', 'woocommerce-gateway-stripe' ) ),
-				$source->last4
+				( isset( $payment_method->brand ) ? wc_get_credit_card_type_label( $payment_method->brand ) : __( 'N/A', 'woocommerce-gateway-stripe' ) ),
+				$payment_method->last4
 			);
 		}
 
-		switch ( $source->type ) {
+		switch ( $payment_method->type ) {
 			case WC_Stripe_Payment_Methods::CARD:
 				return sprintf(
 					/* translators: 1) card brand 2) last 4 digits */
 					__( 'Via %1$s card ending in %2$s', 'woocommerce-gateway-stripe' ),
-					( isset( $source->card->brand ) ? wc_get_credit_card_type_label( $source->card->brand ) : __( 'N/A', 'woocommerce-gateway-stripe' ) ),
-					$source->card->last4
+					( isset( $payment_method->card->brand ) ? wc_get_credit_card_type_label( $payment_method->card->brand ) : __( 'N/A', 'woocommerce-gateway-stripe' ) ),
+					$payment_method->card->last4
 				);
 			case WC_Stripe_Payment_Methods::SEPA_DEBIT:
 				/* translators: 1) last 4 digits of SEPA Direct Debit */
-				return sprintf( __( 'Via SEPA Direct Debit ending in %1$s', 'woocommerce-gateway-stripe' ), $source->sepa_debit->last4 );
+				return sprintf( __( 'Via SEPA Direct Debit ending in %1$s', 'woocommerce-gateway-stripe' ), $payment_method->sepa_debit->last4 );
 			case WC_Stripe_Payment_Methods::CASHAPP_PAY:
 				/* translators: 1) Cash App Cashtag */
-				return sprintf( __( 'Via Cash App Pay (%1$s)', 'woocommerce-gateway-stripe' ), $source->cashapp->cashtag );
+				return sprintf( __( 'Via Cash App Pay (%1$s)', 'woocommerce-gateway-stripe' ), $payment_method->cashapp->cashtag );
 			case WC_Stripe_Payment_Methods::LINK:
 				/* translators: 1) email address associated with the Stripe Link payment method */
-				return sprintf( __( 'Via Stripe Link (%1$s)', 'woocommerce-gateway-stripe' ), $source->link->email );
+				return sprintf( __( 'Via Stripe Link (%1$s)', 'woocommerce-gateway-stripe' ), $payment_method->link->email );
 			case WC_Stripe_Payment_Methods::ACH:
 				return sprintf(
 					/* translators: 1) account type (checking, savings), 2) last 4 digits of account. */
 					__( 'Via %1$s Account ending in %2$s', 'woocommerce-gateway-stripe' ),
-					ucfirst( $source->us_bank_account->account_type ),
-					$source->us_bank_account->last4
+					ucfirst( $payment_method->us_bank_account->account_type ),
+					$payment_method->us_bank_account->last4
 				);
 			case WC_Stripe_Payment_Methods::BECS_DEBIT:
 				return sprintf(
 					/* translators: last 4 digits of account. */
 					__( 'BECS Direct Debit ending in %s', 'woocommerce-gateway-stripe' ),
-					$source->au_becs_debit->last4
+					$payment_method->au_becs_debit->last4
 				);
 			case WC_Stripe_Payment_Methods::ACSS_DEBIT:
 				return sprintf(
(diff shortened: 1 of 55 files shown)


includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1013,80 +1013,85 @@ public function maybe_render_subscription_payment_method( $payment_method_to_dis
 
 		$stripe_customer->set_id( $stripe_customer_id );
 
-		$payment_method_to_display = __( 'N/A', 'woocommerce-gateway-stripe' );
-
 		try {
-			// Retrieve all possible payment methods for subscriptions.
-			foreach ( WC_Stripe_Customer::STRIPE_PAYMENT_METHODS as $payment_method_type ) {
```
daledupreez, line 1020: If we have the Stripe source ID or payment method ID, why are we not just calling Stripe to get the details for that item? This code feels really inefficient.

daledupreez: I ended up fetching the customer-specific payment method instead of querying for saved payment methods and then finding the right one.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1011,82 +1010,124 @@ public function maybe_render_subscription_payment_method( $payment_method_to_dis
...
+
+		$cached_payment_methods[ $stripe_customer_id ][ $payment_method_id ] = $saved_payment_method;
+
+		return $saved_payment_method;
+	}
+
+	/**
+	 * Helper function to get the descriptive text for a payment method or source.
+	 *
+	 * @param object $source The payment method or source object.
```
Mayisha: Optional: My preference would be to call it a `payment_method` in all the new methods.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1011,82 +1010,124 @@ public function maybe_render_subscription_payment_method( $payment_method_to_dis
...
+			return null;
+		}
+
+		if ( isset( $cached_payment_methods[ $stripe_customer_id ][ $payment_method_id ] ) ) {
+			return $cached_payment_methods[ $stripe_customer_id ][ $payment_method_id ];
+		}
+
+		$encoded_customer_id       = rawurlencode( $stripe_customer_id );
+		$encoded_payment_method_id = rawurlencode( $payment_method_id );
+		$api_path                  = "customers/{$encoded_customer_id}/payment_methods/{$encoded_payment_method_id}";
```
Mayisha: As an alternate approach which would cover this related to the open question as well,

> I am not sure how/if we need to handle older src_* sources. I don't know how Stripe handles those via the payment_methods APIs.

we can retrieve the payment method or source using the [get_payment_method method](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-api.php#L468) and check if `payment_method->customer` matches the `$stripe_customer_id`.

AFAIK, `/payment_methods/src_xyz` fails so `customers/customer_abc/payment_methods/src_xyz` might also fail.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1011,82 +1010,124 @@ public function maybe_render_subscription_payment_method( $payment_method_to_dis
...
 			wc_add_notice( $e->getLocalizedMessage(), 'error' );
 			WC_Stripe_Logger::log( 'Error: ' . $e->getMessage() );
 		}
 
-		return $payment_method_to_display;
+		return __( 'N/A', 'woocommerce-gateway-stripe' );
+	}
+
+	/**
+	 * Helper function to get and temporarily cache the payment method details for a customer and payment method ID.
```
Mayisha: Optional: we can move this helper method to [WC_Stripe_Subscriptions_Helper](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/refactor/subscriptions-trait-payment-method-rendering/includes/compat/class-wc-stripe-subscriptions-helper.php).

Mayisha (commented):
Functionally it is working as described. I see only one request to `customers/cus_abc/payment_methods/pm_xyz` in my logs 🎉

includes/compat/trait-wc-stripe-subscriptions.php
daledupreez: Good call. Updated in 9e6ebdd4ba832320f7fa1300a2f0a520c02ad431

includes/compat/trait-wc-stripe-subscriptions.php
daledupreez: Good call. I've switched that over in 3aecb6ca1767ae3c7ecc98b4283f4a515fe0027d.

includes/compat/trait-wc-stripe-subscriptions.php
daledupreez: I think this is also a good suggestion. Done in 292087167749a823d343c767c40f623d25ea7cfb.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1011,82 +1010,124 @@ public function maybe_render_subscription_payment_method( $payment_method_to_dis
...
+	 */
+	private function get_subscription_payment_method_details( string $stripe_customer_id, string $payment_method_id ): ?object {
+		static $cached_payment_methods = [];
+
+		if ( empty( $stripe_customer_id ) || empty( $payment_method_id ) ) {
+			return null;
+		}
+
+		if ( isset( $cached_payment_methods[ $stripe_customer_id ][ $payment_method_id ] ) ) {
+			return $cached_payment_methods[ $stripe_customer_id ][ $payment_method_id ];
```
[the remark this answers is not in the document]
daledupreez: For this use case, I think it's OK to use a static variable, due to us using the customer ID and payment method ID in the path, but I'd be happy to switch to a transient.

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
```diff
@@ -581,61 +581,177 @@ public function test_get_balance_transaction_id_from_charge() {
...
 	 * @see WC_Stripe_Subscriptions_Trait::maybe_render_subscription_payment_method()
+	 * @dataProvider provide_test_render_subscription_payment_method_cases
 	 */
-	public function test_render_subscription_payment_method() {
+	public function test_render_subscription_payment_method( string $payment_method_type, array $payment_method_fields, string $expected_result, ?array $additional_fields = null ) {
 		$mock_subscription = WC_Helper_Order::create_order(); // We can use an order as a subscription.
 		$mock_subscription->set_payment_method( 'stripe' );
 
-		$mock_subscription->update_meta_data( '_stripe_source_id', 'src_mock' );
+		$id_suffix = isset( $payment_method_fields['last4'] ) ? $payment_method_fields['last4'] : (string) rand( 10000, 99999 );
```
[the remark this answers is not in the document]
daledupreez: I've switched to a counter in 7522668a9de43426e765f37ead5eabc7d70e1b65

diegocurbelo (approved):
The code changes look good, and it tests well 🚢 

Screenshot 2025-08-18 at 11 26 23

>I don't know if it's worth using a transient for the payment method caching, or whether we should just cache the payment method without the customer ID dimension. I think both should be unique.

I like the transient idea, I think is consistent with how we are caching other things.. but we cal also improve this later, after we add a auto-cleanup mechanism to the DB cache.

Mayisha (approved):
Thanks Dale. The changes look good 👍 

**Automatic renewal**
✅ Payment method is displayed correctly on the shopper's subscription page.

Screenshot 2025-08-19 at 12 44 08 PM

✅ Only one request is sent to fetch the payment method `payment_methods/pm_xxx`

Screenshot 2025-08-19 at 12 45 22 PM


**Manual renewal**

✅ Payment method is displayed correctly on the shopper's subscription page.

Screenshot 2025-08-19 at 12 44 21 PM

✅ No request is sent to fetch payment methods.

daledupreez:
The e2e failure is unrelated. Merging.

PR #4571, opened 2025-08-11
Add get_icon_url() to payment method classes

diegocurbelo:
Fixes STRIPE-664

WooCommerce 10.1.0 in [Blocks: Add Payment Method Icons Block #57181](https://github.com/woocommerce/woocommerce/pull/57181), started using `$gateway->get_icon_url()` to get the gateway icon:
```
if ( is_callable( array( $gateway, 'get_icon_url' ) ) ){
	$icon_url = $gateway->get_icon_url();
}
```

For most gateways, this is not a problem; if the gateway does not implement the method, it doesn't show any icon.

But in the case of the Stripe Gateway, `$gateway` is a payment method instance that inherits from `WC_Stripe_UPE_Payment_Method`, which doesn't have a `get_icon_url()` method, but it has a `__cal()`  [implementation](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/e0d8375b7fdc870cd6d9ce9f47a5cc95bd516986/includes/payment-methods/class-wc-stripe-upe-payment-method.php#L152), it makes `is_callable()` to always return `true`.

This PR adds a `get_icon_url()` that calls the existing `get_icon()` method that already returns the icon_url.

NOTE: Should we add the individual icons in this PR? or add them as part of 9.9?, it feels that we should also refactor get_icon so we dont have the icons URLs hardrcoded in different palces.

## Testing instructions
0. Update WooCommerce to `10.1.0-rc.4` (manually or using the [beta tester plugin](https://woocommerce.com/products/woocommerce-beta-tester/))
1. Go to `WP-Admin > Pages` and edit the `Checkout` page
2. In the `development` branch, that will trigger an uncaught error:
    Screenshot 2025-08-11 at 15 23 06
3. With this branch (`fix/undefined-method-get-icon-url`), it loads without triggering any errors.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-upe-payment-method.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-method.php
@@ -236,6 +236,16 @@ abstract class WC_Stripe_UPE_Payment_Method extends WC_Payment_Gateway {
 	}
 
 	/**
+	 * Gets the payment method's icon url.
+	 * Each payment method should override this method to return the specific icon url.
+	 *
+	 * @return string The icon url.
+	 */
+	public function get_icon_url() {
+		return '';
+	}
+
+	/**
 	 * Returns boolean dependent on whether payment method
 	 * can be used at checkout
 	 *
--- a/includes/payment-methods/class-wc-stripe-upe-payment-method.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-method.php
@@ -237,11 +237,12 @@ abstract class WC_Stripe_UPE_Payment_Method extends WC_Payment_Gateway {
 
 	/**
 	 * Gets the payment method's icon url.
+	 * Each payment method should override this method to return the icon url.
 	 *
 	 * @return string The icon url.
 	 */
 	public function get_icon_url() {
-		return $this->get_icon();
+		return '';
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -235,6 +235,15 @@ public function get_icon() {
 		return apply_filters( 'woocommerce_gateway_icon', isset( $icons[ $this->get_id() ] ) ? $icons[ $this->get_id() ] : '', $this->id );
 	}
 
+	/**
+	 * Gets the payment method's icon url.
+	 *
+	 * @return string The icon url.
+	 */
+	public function get_icon_url() {
+		return $this->get_icon();
```
daledupreez: It looks like `get_icon()` returns a string that contains an `img` element with the URL in the `src` attribute. So this may need some tweaking.

diegocurbelo: You are right, I reverted it to the original implementation of a base return '', and then each class extending the method in [e8ed281](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4571/commits/e8ed2810ae003a8a0caeb0e76c4899495ff3a083)

wjrosa (approved):
Change is simple and works as expected! ✅

malithsen (approved):
Thank you for explaining the root cause. Code looks good to me and should prevent the fatal. As you noted, we might want to return a proper icon in this function but we can tackle that in a later PR.

PR #4489, opened 2025-07-14
Fix Add Payment Method page and required Stripe customer field validation

annemirasol:
Towards STRIPE-562
## Changes proposed in this Pull Request:

h/t @kaushikasomaiya for finding the breakage, and the deep dive

Iterates on https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4464, where we introduced required field validation before creating a Stripe customer.
- Handle requests from Add Payment Method page, where only the customer's email address is available.
- Bail early when there is no logged in user triggering the Add Payment Method request

## Testing instructions
1. Fix: As a logged in user, make sure you can add a payment method via My Account > Payment methods.
2. Smoke test: as a shopper, make sure you can go through checkout without any problems.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4576, opened 2025-08-12
Fix address validation for manually entered orders

daledupreez:
Fixes STRIPE-659

The underlying driver for this PR is that the original code that made customer validation stricter, https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4464, _always_ required all fields. We previously shipped https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4489 to relax those restrictions when users are adding payment methods. This PR refactors some of the code from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4489 to be a bit more generic so we can identify specific contexts in which we expect a smaller set of customer details to be available. The PR also relaxes customer validation for the "Pay for order" page, making it possible for manually entered orders to be handled with limited customer information.

At a technical level, the changes in this PR are as follows:
 * Change a number of functions to accept a nullable `$current_context` string parameter instead of a `$is_add_payment_method_page` boolean parameter.
   - The changes include a normalization step for all public method so callers of the previous API get the same behaviour.
      * This was only released in 9.7.1, but I still want to avoid unnecessary breakages
 * Update `WC_Stripe_Customer->get_create_customer_required_fields()` to require only an email address when we are on the pay for order page in addition to the case for adding a payment method.
 * Updating the two callers to specify the appropriate context string.
 

## Testing instructions

* Run a current version of the plugin (e.g. from 9.7.1, 9.8.0, or `develop)
* Ensure Stripe is connected in test mode and you have at least one product
* Navigate to _WooCommerce_ -> _Orders_ and then click on the "Add order" button
* Add some line items to the order
* For the billing details, enter only the user's email address
* Save the order
* Once the order is saved, find the "Customer payment page" link in the _General_ section of the order details and copy the link
* In an incognito window or different browser window, navigate to that link
* Try to make payment
* You should get a validation error
* Now switch to this branch, and re-submit the payment
* The payment should now be processed and you should be redirected to the order received page

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -16,7 +16,7 @@ class WC_Stripe_Customer {
 	public const CUSTOMER_CONTEXT_ADD_PAYMENT_METHOD = 'add_payment_method';
 
 	/**
-	 * Constant for the customer context when paying for an order via the "Pay for Order" page..
+	 * Constant for the customer context when paying for an order via the "Pay for Order" page.
 	 */
 	public const CUSTOMER_CONTEXT_PAY_FOR_ORDER = 'pay_for_order';
 
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -209,19 +209,19 @@ class WC_Stripe_Customer {
 	/**
 	 * Validate that we have valid data before we try to create a customer.
 	 *
-	 * @param array $create_customer_request
-	 * @param bool  $is_add_payment_method_page
+	 * @param array       $create_customer_request The base data to build the customer request.
+	 * @param null|string $current_context         Flag to indicate whether we are in a context where limited details are permitted.
 	 *
 	 * @throws WC_Stripe_Exception
 	 */
-	private function validate_create_customer_request( $create_customer_request, $is_add_payment_method_page = false ) {
+	private function validate_create_customer_request( $create_customer_request, ?string $current_context = null ) {
 		/**
 		 * Filters the required customer fields when creating a customer in Stripe.
 		 *
 		 * @since 9.7.0
-		 * @param array $required_fields The required customer fields as derived from the required billing fields in checkout.
+		 * @param array $required_fields The required customer fields as derived from the required billing fields in checkout. In some contexts, like adding a payment method, we allow minimal details to be provided.
 		 */
-		$required_fields = apply_filters( 'wc_stripe_create_customer_required_fields', $this->get_create_customer_required_fields( $is_add_payment_method_page ) );
+		$required_fields = apply_filters( 'wc_stripe_create_customer_required_fields', $this->get_create_customer_required_fields( $current_context ) );
 
 		foreach ( $required_fields as $field => $field_requirements ) {
 			if ( true === $field_requirements ) {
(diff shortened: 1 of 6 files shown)


includes/class-wc-stripe-customer.php
```diff
@@ -258,13 +258,17 @@ private function validate_create_customer_request( $create_customer_request, $is
...
+	 * @param string|null $current_context The context we are creating the customer in. We specifically care about 'pay_for_order' and 'add_payment_method', where minimal details are available.
 	 *
 	 * @return array
 	 */
-	private function get_create_customer_required_fields( $is_add_payment_method_page = false ) {
-		// If we are on the add payment method page, we need to check just for the email field.
-		if ( $is_add_payment_method_page ) {
+	private function get_create_customer_required_fields( ?string $current_context = null ) {
+		// If we are on the add payment method page or the pay for order page, we need to check just for the email field.
+		$contexts_with_minimal_details = [
```
wjrosa: I would move these strings to constants.

includes/class-wc-stripe-customer.php
```diff
@@ -523,19 +530,45 @@ public function update_customer( $args = [], $is_retry = false ) {
...
+	private function normalize_current_context( $current_context ): ?string {
+		if ( null === $current_context ) {
+			return null;
+		}
+
+		if ( is_bool( $current_context ) ) {
+			return $current_context ? 'add_payment_method' : null;
+		}
+
+		if ( is_string( $current_context ) ) {
```
wjrosa, line 578: Another option here would be to use `in_array` with the possible values (making it more strict)

wjrosa (approved):
Code is good and works as expected! Left some optional suggestions

Screenshot 2025-08-12 at 16 41 21

includes/class-wc-stripe-customer.php
daledupreez: In this case, the normalisation is mostly to handle transforming `true/false` to `'add_payment_method'/null`, rather than validating the data otherwise. I would also prefer to make it clearer that other values may be allowed in future. I'd prefer not to add the validation, but I don't feel super-strongly about it.

includes/class-wc-stripe-customer.php
daledupreez: Done in 3953d1a721d311b7e04033946230c9f6420d64c8.

includes/class-wc-stripe-customer.php
```diff
@@ -10,6 +10,24 @@
  */
 class WC_Stripe_Customer {
 
+	/**
+	 * Constant for the customer context when adding a payment method.
+	 */
+	public const CUSTOMER_CONTEXT_ADD_PAYMENT_METHOD = 'add_payment_method';
+
+	/**
+	 * Constant for the customer context when paying for an order via the "Pay for Order" page..
```
malithsen:
```suggestion
	 * Constant for the customer context when paying for an order via the "Pay for Order" page.
```

malithsen (approved):
Looks good and tests well

includes/class-wc-stripe-customer.php
daledupreez: Good catch! Fixed in 6628536490060d36b4a887e06e9f034ce3e34382.

daledupreez:
I've reviewed the E2E failures, and it looks like the Default failures are already being addressed in #4592, and the issues with the Optimized Checkout tests seem to be unrelated to this PR.

freedjack:
I have been seeing this issue on a wp/woocommerce/civicrm instance. I have created a small plugin that solves the address issue for us and may be of use to someone while this issue is being resolved.
https://github.com/freedjack/woocommerce-stripe-address-fix

PR #4581, opened 2025-08-12
Remove connection type check from PMC sync

annemirasol:
Fixes the issue where PMC sync does not get enabled for merchants who oAuth'ed before `connection_type` was introduced.

- We will remove the connection type requirement for PMC migration, and just rely on [this PMC check](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/e53c477e378b491dce8522c572724c4a9a000874/includes/class-wc-stripe-payment-method-configurations.php#L137) to disable incompatible merchants.
- We also introduce code to handle merchants who migrate after BNPL default-on was launched.

## Testing instructions
1. While still on `develop`, in wp-admin:
   - Connect your Stripe account.
   - `wp option patch delete woocommerce_stripe_settings test_connection_type` to simulate an oAuthed merchant with no connection type meta.
   - `wp option patch delete woocommerce_stripe_settings pmc_enabled` to simulate a merchant that has not yet been migrated.
   - `wp option delete wcstripe_cache_test_payment_method_configuration` to clear the cache.
   - Enable Cards and Cash App.
   - Disable Klarna and Affirm.
   - Disable express checkout methods.
2. Go to your Stripe dashboard PMC:
    - Enable Cards, Klarna, Affirm, Google Pay, Apple Pay and Link.
    - Disable Cash App.
4. Checkout this branch.
5. Verify that merchants with no `connection_type` will still get PMC sync'ing enabled.
   - Refresh wp-admin.
   - `wp option get woocommerce_stripe_settings` and verify `pmc_enabled` is `yes`
6. Verify in wp-admin that Cards, Cash App, Klarna, and Affirm are enabled, but express checkout methods are still disabled.
7. Verify that PMC sync gets disabled for merchants with incompatible PMCs.
   - Go to `includes/class-wc-stripe-payment-method-configurations.php` and edit the const `TEST_MODE_CONFIGURATION_PARENT_ID` to some random string.
   - `wp option patch delete woocommerce_stripe_settings test_connection_type`
   - `wp option patch delete woocommerce_stripe_settings pmc_enabled`
   - `wp option delete wcstripe_cache_test_payment_method_configuration`
   - Refresh wp-admin.
   - `wp option get woocommerce_stripe_settings` and verify `pmc_enabled` is `no`

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -376,6 +376,7 @@ class WC_Stripe_Payment_Method_Configurations {
 				}
 			}
 
+			WC_Stripe_Logger::error( 'Switching to Stripe-hosted payment method configuration', [ 'pmc_id' => $merchant_payment_method_configuration->id, 'enabled_payment_methods' => $enabled_payment_methods ] );
 			self::update_payment_method_configuration(
 				$enabled_payment_methods,
 				$available_payment_method_ids
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -303,7 +303,13 @@ class WC_Stripe_Payment_Method_Configurations {
 	 * @return bool
 	 */
 	public static function is_enabled() {
-		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
+		$stripe_settings     = WC_Stripe_Helper::get_stripe_settings();
+		$connection_type_key = WC_Stripe_Mode::is_test() ? 'test_connection_type' : 'connection_type';
+
+		// If the account is known to be something other than a Connect OAuth account, we can't use the payment method configurations API.
+		if ( isset( $stripe_settings[ $connection_type_key ] ) && 'connect' !== $stripe_settings[ $connection_type_key ] ) {
+			return false;
+		}
 
 		// If we have the pmc_enabled flag, and it is set to no, we should not use the payment method configurations API.
 		// We only disable the PMC if the flag is set to no explicitly, an empty value means the migration has not been attempted yet.
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -303,13 +303,7 @@ class WC_Stripe_Payment_Method_Configurations {
 	 * @return bool
 	 */
 	public static function is_enabled() {
-		$stripe_settings     = WC_Stripe_Helper::get_stripe_settings();
-		$connection_type_key = WC_Stripe_Mode::is_test() ? 'test_connection_type' : 'connection_type';
-
-		// If the account is known to be something other than a Connect OAuth account, we can't use the payment method configurations API.
-		if ( isset( $stripe_settings[ $connection_type_key ] ) && 'connect' !== $stripe_settings[ $connection_type_key ] ) {
-			return false;
-		}
+		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
 
 		// If we have the pmc_enabled flag, and it is set to no, we should not use the payment method configurations API.
 		// We only disable the PMC if the flag is set to no explicitly, an empty value means the migration has not been attempted yet.
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -376,7 +376,15 @@ class WC_Stripe_Payment_Method_Configurations {
 				}
 			}
 
-			WC_Stripe_Logger::error( 'Switching to Stripe-hosted payment method configuration', [ 'pmc_id' => $merchant_payment_method_configuration->id, 'enabled_payment_methods' => $enabled_payment_methods ] );
+			WC_Stripe_Logger::error(
+				'Switching to Stripe-hosted payment method configuration',
+				[
+					'pmc_id'                       => $merchant_payment_method_configuration->id,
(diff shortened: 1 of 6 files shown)


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -303,13 +303,7 @@ function ( $id ) use ( $is_test_mode ) {
 	 * @return bool
 	 */
 	public static function is_enabled() {
-		$stripe_settings     = WC_Stripe_Helper::get_stripe_settings();
-		$connection_type_key = WC_Stripe_Mode::is_test() ? 'test_connection_type' : 'connection_type';
-
-		// If the account is not a Connect OAuth account, we can't use the payment method configurations API.
-		if ( ! isset( $stripe_settings[ $connection_type_key ] ) || 'connect' !== $stripe_settings[ $connection_type_key ] ) {
-			return false;
-		}
```
daledupreez, line 312: I think we should still check for known invalid values.
```suggestion
		$connection_type_key = WC_Stripe_Mode::is_test() ? 'test_connection_type' : 'connection_type';

		// If the account is known to be something other than a Connect OAuth account, we can't use the payment method configurations API.
		if ( isset( $stripe_settings[ $connection_type_key ] ) && 'connect' !== $stripe_settings[ $connection_type_key ] ) {
			return false;
		}

Comment (annemirasol) [reply includes/class-wc-stripe-payment-method-configurations.php:312]: It won't let me commit changes to deleted lines, but suggestion pushed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4581/commits/5b0cf289cbdbfb4acb7e47ae5d0e8780d478a58a

Comment (annemirasol) [reply includes/class-wc-stripe-payment-method-configurations.php:312]: During the team call, I remember @diegocurbelo saying we should just remove `connection_type` checks. Was that just for the other PR, or should we remove it from this as well?

Comment (diegocurbelo) [reply includes/class-wc-stripe-payment-method-configurations.php:312]: I think we should remove the `connection_type` check in this PR

**Review by daledupreez — COMMENTED** [review]
I am continuing to test this, as I hit one situation where the back end got stuck in some kind of infinite loop on two requests. I am trying to reproduce.
I did also have a successful test run, but I want to check the loop in more depth.

File: includes/class-wc-stripe-payment-method-configurations.php [hunk 2]
```diff
@@ -362,16 +356,24 @@ public static function maybe_migrate_payment_methods_from_db_to_pmc( $force_migr
...
+				if (
+					! in_array( $payment_method_id, WC_Stripe_Payment_Methods::EXPRESS_PAYMENT_METHODS, true ) &&
+					! in_array( $payment_method_id, $enabled_payment_methods, true ) &&
+					isset( $payment_method->display_preference->value ) && 'on' === $payment_method->display_preference->value
+				) {
+					$enabled_payment_methods[] = $payment_method_id;
+				}
 			}
 
 			self::update_payment_method_configuration(
```
daledupreez, line 388: Not blocking: might it be worth logging an error to note what the merged configuration looks like? (I considered a `warning`, but it looks like that is gated behind the logging flag, so would be less useful.)
```suggestion
			WC_Stripe_Logger::error( 'Switching to Stripe-hosted payment method configuration', [ 'pmc_id' => $merchant_payment_method_configuration->id, 'enabled_payment_methods' => $enabled_payment_methods ] );
			self::update_payment_method_configuration(
```

daledupreez (approved):
I changed my testing approach and did the following, and I have not yet been able to reproduce the issue I saw from my local testing through a few attempts:
* Set up a site running the 9.8.0 released version of the plugin
* Create a test connection, and delete the options and sub-options in the test instructions
* Update the plugin to a version built from this branch
* Verify the end result is as expected

As one possible thought, might we want to actively remove the `pmc_enabled` flag during upgrade to 9.8.1 or the first version that includes this change? It feels like that may allow us to (re)check sites that had the flag set to `'no'` in previous versions. But maybe that should be a follow-up PR, if we decide it's worth it.

I think this is good to ship for the original scope.

includes/class-wc-stripe-payment-method-configurations.php
diegocurbelo: We should also log `available_payment_method_ids`:

```suggestion
			WC_Stripe_Logger::error(
				'Switching to Stripe-hosted payment method configuration',
				[
					'pmc_id'                       => $merchant_payment_method_configuration->id,
					'enabled_payment_methods'      => $enabled_payment_methods,
					'available_payment_method_ids' => $available_payment_method_ids,
				]
			);
			self::update_payment_method_configuration(
```

diegocurbelo (approved):
The code changes look good, and it tests well.

 I tested with a new install/onboarding:
- The default PMC payment methods are enabled
- `pmc_enabled` is set to yes
- `connection_type` is not modified/added

Also tested updating from a `9.4.1` site (with only Cards, CashApp, and Link enabled) to `9.8.1`:
- The PMC was updated to enable CashApp and disable Apple Pay and Google Pay.
    - Cards, CashApp, and Link were still enabled
    - Affirm and Klarna were also enabled


> As one possible thought, might we want to actively remove the pmc_enabled flag during upgrade to 9.8.1 or the first version that includes this change? It feels like that may allow us to (re)check sites that had the flag set to 'no' in previous versions. But maybe that should be a follow-up PR, if we decide it's worth it.

Correct, when the site has `pmc_enabled` set to `no`, this change has no effect, and we should remove the flag on update.

But it would only affect sites that have already attempted the migration and it failed for another reason; in other words, the store is connected via OAuth, but it couldn't read the inherited PMC.

PR #4582, opened 2025-08-13
Add PHP test exclusion: WooCommerce 9.9 requires WordPress >= 6.7

daledupreez:

With WooCommerce 10.1 being released yesterday, WooCommerce 9.9 is now the L-2 release, but it's not compatible with WordPress 6.6, which is currently the WordPress L-2 release. We need to drop the test cases that try to test that combination, and will need to look at this again when the next WordPress release comes out. Note that I have removed the WordPress L-2 option altogether, as it caused weird issues when excluding all combinations that include it.

## Testing instructions

Verify that the automated `PHP tests` workflows all pass, and we don't try to run tests for L-2/L-2 combinations of WordPress and WooCommerce.

---

### Changelog entry

<details>

#### Comment 
This change only impacts our GitHub automated unit tests, and doesn't impact any code in the plugin itself.

</details>

**Post merge**

### DIFF
--- a/.github/workflows/php-tests.yml
+++ b/.github/workflows/php-tests.yml
@@ -40,7 +40,7 @@ jobs:
       max-parallel: 16
       matrix:
         woocommerce_support_policy: [ 'L', 'L-1', 'L-2' ]
-        wordpress_support_policy:   [ 'L', 'L-1', 'L-2' ]
+        wordpress_support_policy:   [ 'L', 'L-1' ]
         php_support_policy:         [ 'L', 'L-1', 'L-2', 'L-3', '7.4' ]
         include:
           # WooCommerce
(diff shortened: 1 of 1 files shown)


daledupreez:
For the record, I have also made the checks for the combination of WordPress L-2 and WooCommerce L-2 PHP tests not required for merges to `develop`.

wjrosa (approved):
Thanks for excluding those incompatible versions! Change looks good to me 👍 

PS: I see this failure, but I believe it will go away once this is merged:
Screenshot 2025-08-13 at 08 59 13

malithsen:
> `Stable (PHP=, WP=L-2, WC=)`

I'm not sure why the PHP and WC versions are not defined. I wonder if we exclude by the exact version (WC=9.9, WP=6.6) will prevent this.

daledupreez:
I think we should drop `L-2` for WordPress altogether, at least for the time being. I think the include/exclude approach is not clean for that case.

daledupreez:
OK, it looks like the PHP tests are now clean, and we don't have that weird failure stemming from the empty versions. 😅

wjrosa (approved):

malithsen (approved):
Thanks for fixing this Dale!

PR #4593, opened 2025-08-14
Prevent PMC migration to run when the plugin is not connected to Stripe

diegocurbelo:
Fixes #<github_issue_id>

When instantiating the main Gateway class `WC_Stripe_UPE_Payment_Gateway`, the constructor lists all available payment methods (`get_upe_enabled_payment_method_ids()`), and that method attempts to migrate the local configuration to Stripe's PMC without checking if there is a connected account first.

This PR prevents the PMC migration from running when the store is not connected to Stripe.
It adds an `not is_connected()` bail out to the PMC `is_enabled()`

As the call originates from the `WC_Stripe_UPE_Payment_Gateway` constructor, we can't use the existing `WC_Stripe::get_instance()->connect->is_connected()`, so we extract the `is_connected` logic to the `WC_Stripe_Helper` so we can also call it from the `maybe_migrate_payment_methods_from_db_to_pmc()`.

## Testing instructions

1. Check out the `release/9.8.1` branch
2. Remove the plugin settings:
    `wp option delete woocommerce_stripe_settings`
3. Remove the Account cache:
    `wp option delete wcstripe_test_payment_method_configuration_cache`
4. Remove the API rate-limit count cache: 
    `wp option delete wcstripe_cache_live_invalid_api_key_error_count`
    `wp option delete wcstripe_cache_test_invalid_api_key_error_count`
5. Go to the plugin settings
6. Go to `WP Admin > WooCommerce > Status > Logs` and click `woocommerce-gateway-stripe` from the list
7. Check that there were multiple GETs to the payment method configuration API (which resulted in 401 errors):
    Screenshot 2025-08-14 at 22 46 31
8. Delete the log entry by clicking `Delete permanently`
    Screenshot 2025-08-15 at 00 50 48
9. Checkout this branch (`fix/prevent-pmc-migration-when-not-connected`)
10. Remove the plugin settings:
    `wp option delete woocommerce_stripe_settings`
11. Remove the Account cache:
    `wp option delete wcstripe_test_payment_method_configuration_cache`
12. Remove the API rate-limit count cache: 
    `wp option delete wcstripe_cache_live_invalid_api_key_error_count`
    `wp option delete wcstripe_cache_test_invalid_api_key_error_count`
13. Go to the plugin settings
14. Go to `WP Admin > WooCommerce > Status > Logs` and click `woocommerce-gateway-stripe` from the list
15. Check that there were no attempts to fetch the payment method configurations

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -1992,7 +1992,7 @@ class WC_Stripe_Helper {
 	 */
 	public static function is_connected( $mode = null ) {
 		// If the mode is not provided, we'll check the current mode.
-		if ( is_null( $mode ) ) {
+		if ( null === $mode ) {
 			$mode = WC_Stripe_Mode::is_test() ? 'test' : 'live';
 		}
 
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -1983,4 +1983,24 @@ class WC_Stripe_Helper {
 
 		throw new Exception( __( "We're not able to process this request. Please try again later.", 'woocommerce-gateway-stripe' ) );
 	}
+
+	/**
+	 * Determines if the store is connected to Stripe.
+	 *
+	 * @param string $mode Optional. The mode to check. 'live' or 'test' - if not provided, the currently enabled mode will be checked.
+	 * @return bool True if connected, false otherwise.
+	 */
+	public static function is_connected( $mode = null ) {
+		// If the mode is not provided, we'll check the current mode.
+		if ( is_null( $mode ) ) {
+			$mode = WC_Stripe_Mode::is_test() ? 'test' : 'live';
+		}
+
+		$options = self::get_stripe_settings();
+		if ( 'test' === $mode ) {
+			return isset( $options['test_publishable_key'], $options['test_secret_key'] ) && trim( $options['test_publishable_key'] ) && trim( $options['test_secret_key'] );
+		} else {
+			return isset( $options['publishable_key'], $options['secret_key'] ) && trim( $options['publishable_key'] ) && trim( $options['secret_key'] );
+		}
+	}
 }
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -303,6 +303,11 @@ class WC_Stripe_Payment_Method_Configurations {
 	 * @return bool
 	 */
 	public static function is_enabled() {
+		// Bail if account is not connected.
+		if ( ! WC_Stripe_Helper::is_connected() ) {
+			return false;
+		}
+
 		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
 
 		// If we have the pmc_enabled flag, and it is set to no, we should not use the payment method configurations API.
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -296,9 +296,7 @@ class WC_Stripe_Payment_Method_Configurations {
 
 	/**
 	 * Check if the payment method configurations API can be used to store enabled payment methods.
-	 * This requires the Stripe account to be connected to our platform ('connection_type' option to be 'connect').
-	 *
-	 * This is temporary until we finish the re-authentication campaign.
+	 * This requires the Stripe account to be connected to Stripe.
 	 *
 	 * @return bool
 	 */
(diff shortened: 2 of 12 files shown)


includes/class-wc-stripe-helper.php
```diff
@@ -1983,4 +1983,24 @@ public static function validate_intent_for_order( $order, $intent, ?string $sele
...
 		throw new Exception( __( "We're not able to process this request. Please try again later.", 'woocommerce-gateway-stripe' ) );
 	}
+
+	/**
+	 * Determines if the store is connected to Stripe.
+	 *
+	 * @param string $mode Optional. The mode to check. 'live' or 'test' - if not provided, the currently enabled mode will be checked.
+	 * @return bool True if connected, false otherwise.
+	 */
+	public static function is_connected( $mode = null ) {
```
daledupreez, line 1993: Non-blocking nit: for new code like this, does it make sense to add explicit types?
```suggestion
	public static function is_connected( ?string $mode = null ) {
```
On second glance, we're calling the function directly from the existing `is_connected()` function, which doesn't have strict types, so maybe it's not worth adding that for now.

includes/class-wc-stripe-helper.php
```diff
@@ -1983,4 +1983,24 @@ public static function validate_intent_for_order( $order, $intent, ?string $sele
...
+		// If the mode is not provided, we'll check the current mode.
+		if ( is_null( $mode ) ) {
```
daledupreez: Also non-blocking nit: is it worth doing a direct `null ===`check rather than a function call?
```suggestion
		if ( null === $mode ) {
```

includes/class-wc-stripe-helper.php
daledupreez, line 1993: Also, from a naming perspective, is it worth calling this function something more like `have_stripe_keys()`? The "connected" wording implies that we have a connection to Stripe, but in this case we only know that we have saved keys, not whether they actually work.

includes/class-wc-stripe-helper.php
```diff
@@ -1983,4 +1983,24 @@ public static function validate_intent_for_order( $order, $intent, ?string $sele
...
+			$mode = WC_Stripe_Mode::is_test() ? 'test' : 'live';
+		}
+
+		$options = self::get_stripe_settings();
+		if ( 'test' === $mode ) {
+			return isset( $options['test_publishable_key'], $options['test_secret_key'] ) && trim( $options['test_publishable_key'] ) && trim( $options['test_secret_key'] );
+		} else {
+			return isset( $options['publishable_key'], $options['secret_key'] ) && trim( $options['publishable_key'] ) && trim( $options['secret_key'] );
+		}
```
daledupreez, line 2004: Is it worth validating that the shape of the keys matches what we expect? e.g. start with `sk_{$mode}_`.

daledupreez (approved):
Changes LGTM and test well! I have some minor comments and questions, but nothing that blocks us shipping this as-is.

As a side note, I tested with `develop` and then this branch, but the behaviour was the same.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -303,6 +303,11 @@ function ( $id ) use ( $is_test_mode ) {
 	 * @return bool
 	 */
 	public static function is_enabled() {
+		// Bail if account is not connected.
```
Mayisha, line 304: Nit: Let's update the method docblock as we have updated the logic for `is_enabled`.

Mayisha (approved):
LGTM. Functionally, it is working as described.

- In the `develop` branch, I see some failed PMC requests in the logs.

Screenshot 2025-08-15 at 3 25 48 PM

- In this branch, there is no attempt to get/update the PMC when a Stripe account is not connected.

includes/class-wc-stripe-helper.php
diegocurbelo: We can, but I think that should be done in a different PR, as it needs some research.
For example, some merchants are using restricted keys (that start with `rk_`), so if we are going to restrict the prefixes, we need to make sure we have the complete list of possible prefixes.

includes/class-wc-stripe-helper.php
diegocurbelo: I agree that we only check that the keys are defined in the plugin settings, but I think that instead of renaming the method, we should add a check if we are connected (in a future PR so we can test it thoroughly to avoid incidents as this is a critical path of most flows). For example, we can use the 401 rate limiter to detect the connections tate without making any extra calls 🤷🏼

includes/class-wc-stripe-payment-method-configurations.php
diegocurbelo: Docblock updated in [6ff99b4](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4593/commits/6ff99b42e0efb0456dee6b6f0339859437c99b4b)

PR #4599, opened 2025-08-18
Fix Stripe customer creation with no email

malithsen:
Fixes STRIPE-673

This PR fixes an edge case where logged-in users without an email address in their profile cannot complete checkout, even when they enter a valid email in the checkout form. For logged-in users, we use their profile email when creating the Stripe customer. Although it is not allowed via the UI, it is possible to have user accounts with empty emails created through custom scripts. In v9.7.0, we introduced validation to prevent Stripe customer creation with empty values which surfaced this issue.

This PR addresses this issue by falling back to using the billing email address if the user's email is empty for a logged-in user.

### Setup
Create a test user without an email address using this SQL:
 ```sql
 INSERT INTO wp_users (ID, user_login, user_pass, user_nicename, user_email, user_url,
user_registered, user_activation_key, user_status, display_name)
 VALUES (NULL, 'nomail', MD5('password'), 'nomail', '', '', '2024-01-01 00:00:00', '', '0', 'No
Email User');
```

#### Test the issue on Develop

1. Log in as the user created above (`username: nomail`, `password: password`)
2. Add a product to the cart and proceed to checkout.
3. Fill in all required fields including a valid email address.
4. Attempt to complete the order.
5. Expected bug: You'll see an error "Missing required customer field: email"

#### Test the fix on this branch

1. Apply the changes from this PR.
2. Repeat steps 1-4 from above.
3. Order should be placed successfully.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -13,6 +13,7 @@
 * Tweak - Small improvements to e2e tests
 * Fix - Prevent the PMC migration to run when the plugin is not connected to Stripe
 * Fix - Fixes a fatal error in the OC inbox note when the new checkout is disabled
+* Fix - Allow checkout for logged-in users without an email in their account when a billing email is provided
 
 = 9.8.0 - 2025-08-11 =
 * Add - Adds the current setting value for the Optimized Checkout to the Stripe System Status Report data
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,13 +11,13 @@
 * Tweak - Small improvements to e2e tests
 * Fix - Fix unnecessary Stripe API calls when rendering subscription details
 * Add - Adds a new action (`wc_stripe_webhook_received`) to allow additional actions to be taken for webhook notifications from Stripe
+* Fix - Allow checkout for logged-in users without an email in their account when a billing email is provided
 
 = 9.8.1 - 2025-08-15 =
 * Fix - Remove connection type requirement from PMC sync migration attempt
 * Fix - Relax customer validation that was preventing payments from the pay for order page
 * Fix - Prevent the PMC migration to run when the plugin is not connected to Stripe
 * Fix - Fixes a fatal error in the OC inbox note when the new checkout is disabled
-* Fix - Allow checkout for logged-in users without an email in their account when a billing email is provided
 
 = 9.8.0 - 2025-08-11 =
 * Add - Adds the current setting value for the Optimized Checkout to the Stripe System Status Report data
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by Mayisha — APPROVED** [review]
Thanks for the fix Malith. Looks good to me 👍 

**develop**
✅ I can reproduce the issue in `develop` following the instructions in the PR description. Checkout fails with the following error-

Screenshot 2025-08-19 at 1 30 05 PM

**fix/stripe-customer-creation-with-no-email**

✅ Checkout succeeds without any errors.
✅ Stripe customer is created using the billing email.

Screenshot 2025-08-19 at 1 33 11 PM

✅ Stripe transaction has the billing email in the payment details section.

Screenshot 2025-08-19 at 1 32 53 PM

File: includes/class-wc-stripe-customer.php [hunk 1]
```diff
@@ -169,11 +171,16 @@ protected function generate_customer_request( $args = [] ) {
 				$billing_last_name = get_user_meta( $user->ID, 'last_name', true );
 			}
 
+			// If the user email is not set, use the billing email.
+			if ( empty( $email ) ) {
```
daledupreez, line 175: Nit: can we co-locate the code that initialises the email and this check?
```suggestion
			$email = $user->user_email;

			// If the user email is not set, use the billing email.
			if ( empty( $email ) ) {
```

changelog.txt
```diff
@@ -17,6 +17,7 @@
 * Fix - Relax customer validation that was preventing payments from the pay for order page
 * Fix - Prevent the PMC migration to run when the plugin is not connected to Stripe
 * Fix - Fixes a fatal error in the OC inbox note when the new checkout is disabled
+* Fix - Allow checkout for logged-in users without an email in their account when a billing email is provided
```
daledupreez: I merged from latest and this location is no longer correct -- the changelog entry needs to be move to 9.9.0.

daledupreez (approved):
The change looks good to me, and tested well. I have two minor notes about shifting the changelog and the email variable initialisation, but neither requires another round of review.

Neatly done!

includes/class-wc-stripe-customer.php
malithsen: Good suggestion, updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4599/commits/8407c06303013e52b442e7638ad11fe5cadd5893

changelog.txt
malithsen: Thanks, moved in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4599/commits/8675b135e210918da8a06e50536a09512dc92549

PR #4607, opened 2025-08-20
Fix frontend unit tests for Node 20

wjrosa:
Fixes STRIPE-454
Base PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4264

As part of the Node 20 upgrade project, this PR extracts most of the required frontend unit tests fixes from the source PR (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4264). The changes here are compatible with both versions (16 and 20).

What this changes:

- Adds the `await` clause in front of every `userEvent.click` call
- Mocks the `createSelector` method when mocking the `@wordpress/data` library
- Correctly awaits `getByRole` calls
- Apply some lint fixes (tests only)

## Testing instructions

- Code review
- Check if the frontend tests and lint validations are still passing

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,7 +2,6 @@
 
 = 9.9.0 - xxxx-xx-xx =
 * Dev - Update Javascript unit tests for compatibility with Node 20
-* Dev - Multiple updates to the frontend unit tests in preparation for the Node 20 upgrade
 * Dev - Replaces some payment method instantiation logic for the Optimized Checkout with calls to the `get_payment_method_instance` method
 * Dev - Multiple lint fixes in preparation for the Node 20 upgrade
 * Dev - Introduces a new helper method to identify Stripe orders
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.9.0 - xxxx-xx-xx =
+* Dev - Multiple updates to the frontend unit tests in preparation for the Node 20 upgrade
 * Dev - Replaces some payment method instantiation logic for the Optimized Checkout with calls to the `get_payment_method_instance` method
 * Dev - Multiple lint fixes in preparation for the Node 20 upgrade
 * Dev - Introduces a new helper method to identify Stripe orders
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.9.0 - xxxx-xx-xx =
+* Dev - Update Javascript unit tests for compatibility with Node 20
 * Dev - Multiple updates to the frontend unit tests in preparation for the Node 20 upgrade
 * Dev - Replaces some payment method instantiation logic for the Optimized Checkout with calls to the `get_payment_method_instance` method
 * Dev - Multiple lint fixes in preparation for the Node 20 upgrade
--- a/client/components/popover/test/index.test.js
+++ b/client/components/popover/test/index.test.js
@@ -40,7 +40,7 @@ describe( 'Popover', () => {
 
 		expect( screen.queryByText( 'Popover Content' ) ).toBeInTheDocument();
 
-		act( async () => {
+		await act( async () => {
 			await userEvent.click( screen.getByTestId( 'base-component' ) );
 		} );
 
--- a/client/components/popover/test/index.test.js
+++ b/client/components/popover/test/index.test.js
@@ -1,4 +1,5 @@
-import { render, screen, act, fireEvent } from '@testing-library/react';
+import { render, screen, act } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import Popover from '..';
 
 const DummyBaseComponent = ( { children, ...props } ) => (
@@ -21,7 +22,7 @@ describe( 'Popover', () => {
 		).not.toBeInTheDocument();
 	} );
 
-	it( 'toggle the visibility on click', () => {
+	it( 'toggle the visibility on click', async () => {
 		render(
 			<Popover
 				BaseComponent={ DummyBaseComponent }
@@ -33,14 +34,14 @@ describe( 'Popover', () => {
 			screen.queryByText( 'Popover Content' )
 		).not.toBeInTheDocument();
 
-		act( () => {
-			fireEvent.click( screen.getByTestId( 'base-component' ) );
+		act( async () => {
+			await userEvent.click( screen.getByTestId( 'base-component' ) );
 		} );
 
 		expect( screen.queryByText( 'Popover Content' ) ).toBeInTheDocument();
 
-		act( () => {
-			fireEvent.click( screen.getByTestId( 'base-component' ) );
+		act( async () => {
+			await userEvent.click( screen.getByTestId( 'base-component' ) );
 		} );
 
 		expect(
--- a/client/components/popover/test/index.test.js
+++ b/client/components/popover/test/index.test.js
@@ -1,5 +1,4 @@
-import { render, screen, act } from '@testing-library/react';
-import userEvent from '@testing-library/user-event';
+import { render, screen, act, fireEvent } from '@testing-library/react';
 import Popover from '..';
 
 const DummyBaseComponent = ( { children, ...props } ) => (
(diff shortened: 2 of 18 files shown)


client/settings/advanced-settings-section/__tests__/optimized-checkout-feature.test.js
```diff
@@ -43,26 +40,20 @@ describe( 'Optimized Checkout Element feature setting', () => {
...
+		useIsOCEnabled.mockReturnValue( [ false, mockSetIsOCEnabled ] );
 
-		userEvent.click( checkbox );
-
-		jest.runAllTimers();
+		render( <OptimizedCheckoutFeature /> );
 
-		expect( checkbox ).toBeDisabled();
-		expect( checkbox ).not.toBeChecked();
+		expect( mockSetIsOCEnabled ).toHaveBeenCalledWith( false );
```
daledupreez: Is the render calling a state setter? That feels pretty weird to me.

daledupreez (commented):
@wjrosa, this seems OK at a glance, but I have some questions for you:
 * Why are we switching from `userEvent` to `fireEvent`? Per [the Testing Library docs for `userEvent`](https://testing-library.com/docs/user-event/intro), we should prefer `userEvent` over `fireEvent` as it is a more realistic reflection of user interactions.
   - I am OK if this is meant to be temporary or a workaround, but I think we should clarify why we're doing this. It's possible we did it in the original PR too, in which case we should understand why we did it there!
 * How can/should we test that these changes still pass in node 20? It would be good to clarify how we can confirm that is true. (But I don't think it's blocking, but I'd rather make sure we confirm it is working so we can avoid later cleanup.)

client/settings/advanced-settings-section/__tests__/optimized-checkout-feature.test.js
wjrosa: Hum, how would you do it instead?

wjrosa:
Hey Dale!

> Why are we switching from userEvent to fireEvent? Per [the Testing Library docs for userEvent](https://testing-library.com/docs/user-event/intro), we should prefer userEvent over fireEvent as it is a more realistic reflection of user interactions.

That's just because it works with Node 20, while userEvent does not (at least with our changes [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4264)). I saw it in the [docs for the latest versions](https://jestjs.io/docs/tutorial-react) and decided to give it a try. I will do some testing with the latest versions of testing library to see if that works for us.


> How can/should we test that these changes still pass in node 20? It would be good to clarify how we can confirm that is true. (But I don't think it's blocking, but I'd rather make sure we confirm it is working so we can avoid later cleanup.)

The changes I made here are exactly the same as I did in the [main PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4264). You can see that tests are passing there now. You can compare both.

client/settings/advanced-settings-section/__tests__/optimized-checkout-feature.test.js
daledupreez: My confusion stems from the fact that the previous test was waiting for the user to click on a button before checking for a state change. Simply rendering a component should generally not result in a state change, which was why I was confused.

Looking at this more closely, I now realise that we have a far more subtle situation: `useIsOCEnabled()` only looks at the underlying setting and not whether that setting is being overridden by UPE being disabled, and we're using an effect to update the client-side state when we detect that UPE is disabled.

It feels to me like it would be cleaner if the `useIsOCEnabled()` logic took into account whether UPE was enabed/disabled so our client-side state more directly represents the server-side state.

But I think that it something we should tackle in other PRs!

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.9.0 - xxxx-xx-xx =
+* Dev - Multiple updates to the frontend unit tests in preparation for the Node 20 upgrade
```
daledupreez: Non-blocking nit: I think we could simplify the description, but I don't feel strongly about it.
```suggestion
* Dev - Update Javascript unit tests for compatibility with Node 20
```

daledupreez (approved):
I think it's worth explaining _why_ you're using `fireEvent` instead of `userEvent` in the PR description, even if it's just "it continues to work". 😁

That said, I think we should eventually aim to switch back to `userEvent` at some point after we've completed the upgrade to Node 20, as I think the idea of testing user interactions is going to help us write better tests in the long run. However, I think that the big obstacle here is [the 14.0 release](https://github.com/testing-library/user-event/releases/tag/v14.0.0) of `@testing-library/user-event`, as that was when the core API changed from synchronous to asynchronous. Still, let's take that on after we have node 20 running.

wjrosa:
> I think it's worth explaining why you're using fireEvent instead of userEvent in the PR description, even if it's just "it continues to work". 😁

I conducted some tests and was able to resolve the issue by simply adding the `await` clause in front of all `userEvent.click` calls. So I reverted that replacement in both here and in the [base PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4264). Thanks for flagging that!

daledupreez (approved):
Thanks for the updates! They look good to me!

Another thing we might be able to is split out a PR to update `@testing-library/user-event` to `14.0.0` (or some similarly early 14.x release), as that was released quite a while ago. We may not be able to switch to a current version until we upgrade React, but we might be able to reduce our work by tackling some of the work in an interim PR.

client/stripe-utils/__tests__/cash-app-limit-notice-handler.test.js
```diff
@@ -1,7 +1,9 @@
 import { render, screen } from '@testing-library/react';
-import { maybeShowCashAppLimitNotice } from 'wcstripe/stripe-utils/cash-app-limit-notice-handler';
+import {
+	CASH_APP_NOTICE_AMOUNT_THRESHOLD,
+	maybeShowCashAppLimitNotice,
+} from 'wcstripe/stripe-utils/cash-app-limit-notice-handler';
 import { callWhenElementIsAvailable } from 'wcstripe/blocks/upe/call-when-element-is-available';
-import { CASH_APP_NOTICE_AMOUNT_THRESHOLD } from 'wcstripe/data/constants';
```
Mayisha, line 4: Isn't `CASH_APP_NOTICE_AMOUNT_THRESHOLD` still part of the constant file? 🤔

client/settings/advanced-settings-section/__tests__/optimized-checkout-feature.test.js
```diff
@@ -90,6 +92,9 @@ describe( 'Optimized Checkout Element feature setting', () => {
 		expect( setLayoutMock ).not.toHaveBeenCalled();
 
 		userEvent.click( screen.getByLabelText( 'Tabs' ) );
```
Mayisha:
```suggestion
		await userEvent.click( screen.getByLabelText( 'Tabs' ) );
```

client/components/popover/test/index.test.js
```diff
@@ -34,14 +34,14 @@ describe( 'Popover', () => {
 			screen.queryByText( 'Popover Content' )
 		).not.toBeInTheDocument();
 
-		act( () => {
-			userEvent.click( screen.getByTestId( 'base-component' ) );
+		act( async () => {
```
Mayisha:
```suggestion
		await act( async () => {
```

client/components/popover/test/index.test.js
```diff
@@ -34,14 +34,14 @@ describe( 'Popover', () => {
...
+			await userEvent.click( screen.getByTestId( 'base-component' ) );
 		} );
 
 		expect( screen.queryByText( 'Popover Content' ) ).toBeInTheDocument();
 
-		act( () => {
-			userEvent.click( screen.getByTestId( 'base-component' ) );
+		act( async () => {
```
Mayisha:
```suggestion
		await act( async () => {
```

client/components/popover/test/index.test.js
Mayisha: Adding `await` here fixes the following console error.
Screenshot 2025-08-25 at 2 24 19 PM

client/components/popover/test/index.test.js
Mayisha: Adding `await` here fixes the following console error.

Screenshot 2025-08-25 at 2 24 19 PM

Mayisha (approved):
Changes are straightforward and look good to me. 👍 
I have suggested a few additional missing `await`.
+1 to the suggestion of updating `@testing-library/user-event` in a follow-up PR.

PR #4621, opened 2025-08-26
Prettier upgrade

wjrosa:
Fixes STRIPE-449
Base PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4264

As part of the Node 20 project, this PR upgrades the Prettier package to `npm:wp-prettier@^3.0.3`, as well as its related libraries. To follow the upgrade, I am also making all the required linting changes. No actual code changes were made (just formatting).

## Testing instructions

- Confirm the frontend unit tests here are passing
- Checkout and build this branch on your test environment (`update/prettier-upgrade`)
- Perform some basic smoke-testing and confirm that no regression was introduced

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


package.json
```diff
@@ -101,6 +101,7 @@
     "@stripe/stripe-js": "^1.36.0",
     "@woocommerce/settings": "^1.0.0",
     "@wordpress/html-entities": "^3.6.1",
+    "eslint-plugin-prettier": "^5.5.4",
```
daledupreez: Should this not be a devDependency?

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.9.0 - xxxx-xx-xx =
+* Update - Upgrades the Prettier package and related libraries
```
daledupreez: Two thoughts:
1. This feels like a `Dev` task, as the impact on merchants and shoppers should be non-existent
2. We are also updating the formatting for a lot of files, so it may be worth noting that
```suggestion
* Dev - Upgrade the Prettier package and related libraries, and apply updated formatting rules
```

daledupreez (approved):
LGTM, and I confirmed that the default and e2e tests are working with my patch to fix the e2e tests.

As a side note, might it be worth updating `@wordpress/prettier-config` as part of this PR too? Per the [package changelog](https://github.com/WordPress/gutenberg/blob/trunk/packages/prettier-config/CHANGELOG.md), it looks like we could update to 3.15.x while we are on node 16. Based on my testing (which was running `npm i @wordpress/prettier-config@^3.15.0` and then running `npm run format:js`), this doesn't generate any changes, so may not be worth doing.

package.json
wjrosa: Absolutely. Moved it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4621/commits/2fd162077fa71568dc506370111e21332e8cc869

wjrosa:
Thanks for another review!

> As a side note, might it be worth updating @wordpress/prettier-config as part of this PR too? Per the [package changelog](https://github.com/WordPress/gutenberg/blob/trunk/packages/prettier-config/CHANGELOG.md), it looks like we could update to 3.15.x while we are on node 16. Based on my testing (which was running npm i @wordpress/prettier-config@^3.15.0 and then running npm run format:js), this doesn't generate any changes, so may not be worth doing.

I don't have a strong opinion on this one, but I updated the package to that version. I don't see any issues with it, and it will be nice to update some additional packages (as we don't do it very often).

annemirasol (approved):
LGTM!
- `npm run test:js` is passing
- smoke tested on Stripe settings page

PR #4626, opened 2025-08-26
Simplifying instructions logic for OC

wjrosa:
Fixes STRIPE-689

This PR simplifies the logic related to displaying payment method instructions when Optimized Checkout is enabled. The goal is to remove as much custom code as possible, keeping the standard flow where it applies. I am also taking the opportunity to remove the deprecated `get_testing_instructions_for_smart_checkout`.

## Testing instructions

- Checkout and build this branch on your test environment (`dev/simplifying-instructions-logic-for-oc`)
- Connect your Stripe account
- Enable test mode
- Enable some payment methods with instructions (card and SEPA, for example)
- Go to the checkout page
- Confirm that by clicking the payment methods, the instructions are correctly shown

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-method-blik.php
```diff
@@ -53,10 +53,16 @@ public function get_retrievable_type() {
 	/**
 	 * Returns testing instructions to be printed at checkout in test mode.
 	 *
-	 * @param bool $show_optimized_checkout_instruction Whether this is being called through the Optimized Checkout instructions method. Used to avoid an infinite loop call.
 	 * @return string
 	 */
 	public function get_testing_instructions( $show_optimized_checkout_instruction = false ) {
+		if ( $show_optimized_checkout_instruction ) {
```
daledupreez: To be explicit, should we check for any non-false value?
```suggestion
		if ( false !== $show_optimized_checkout_instruction ) {
```

includes/payment-methods/class-wc-stripe-upe-payment-method-oc.php
```diff
@@ -103,21 +103,25 @@ public function requires_automatic_capture() {
...
-			/* translators: 1) HTML strong open tag 2) HTML strong closing tag 3) HTML anchor open tag 2) HTML anchor closing tag */
-			esc_html__( '%1$sTest mode:%2$s use the test VISA card 4242424242424242 with any expiry date and CVC. Other payment methods may redirect to a Stripe test page to authorize payment. More test card numbers are listed %3$shere%4$s.', 'woocommerce-gateway-stripe' ),
-			'<strong>',
-			'</strong>',
-			'<a href="https://docs.stripe.com/testing" target="_blank">',
-			'</a>'
-		);
+		$instructions          = '';
+		$base_instruction_html = '<div id="wc-stripe-payment-method-instructions-%s" class="wc-stripe-payment-method-instruction" style="display: none;">%s</div>';
+		foreach ( WC_Stripe_UPE_Payment_Gateway::UPE_AVAILABLE_METHODS as $payment_method_class ) {
```
daledupreez: Question: why are we using `WC_Stripe_UPE_Payment_Gateway::UPE_AVAILABLE_METHODS` instead of `WC_Stripe_Payment_Method_Configurations::get_upe_enabled_payment_method_ids()`? We don't need to generate instructions for disabled payment methods - it only generates unnecessary HTML in the page in addition to the additional back end processing.

daledupreez (approved):
This is working as described, but I think we should only look at enabled payment methods and not all UPE-enabled payment methods. That can happen in this PR or in a follow-up.

Mayisha (approved):
Looks good to me.

Testing instructions are correctly displayed for-
✅ Split upe checkout - classic
✅ Split upe checkout - block
✅ Optimized checkout - classic
✅ Optimized checkout - block

includes/payment-methods/class-wc-stripe-upe-payment-method-oc.php
wjrosa: Good idea! Implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4626/commits/c14ccd6f386777899e689d4bc4d1d7130a87cedf

PR #3819, opened 2025-01-31
Add setting to allow other methods to be saved as SEPA tokens

wjrosa:
Fixes #3786

## Changes proposed in this Pull Request:

This PR adds a new setting to allow merchants to control whether [Bancontact](https://docs.stripe.com/billing/subscriptions/bancontact) and [iDEAL](https://docs.stripe.com/payments/ideal/set-up-payment) payment methods can be saved for later purchases as SEPA Direct Debit tokens. Some merchants would avoid this behavior since the SEPA method takes [too long to be identified (~8 business days)](https://docs.stripe.com/payments/sepa-debit).

## Testing instructions

- Checkout and build this branch on your test environment (`add/sepa-tokens-for-other-methods-setting`)
- Connect your Stripe account
- Set your store currency to EUR
- Enable Bancontact and iDEAL
- Confirm you are able to see the new setting and it is enabled by default:
Screenshot 2025-01-31 at 16 19 12

- Confirm you can disable/enable the setting and that by reloading the page the change is persisted
- As a shopper, with the setting enabled, try to add a new Bancontact method
- Confirm the method is saved
Screenshot 2025-01-31 at 16 22 39

- Disable the setting and confirm you are not able to save the method anymore
- Repeat the test steps for iDEAL

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [x] Added changelog entry **in both** `changelog.txt` and `readme.txt` (or does not apply)
-   [ ] Tested on mobile (or does not apply)

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4633, opened 2025-08-28
Splitting the SEPA tokens for other methods setting

wjrosa:
Fixes STRIPE-693
P2 peNR48-1Jq-p2

Following the discussion in peNR48-1Jq-p2, in this PR, I am splitting the "Enable SEPA Tokens for other methods" setting into separate settings for both iDEAL and Bancontact. I am also updating some of the copy based on the suggestion.

This setting was first introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3819. It allows shoppers to purchase using iDEAL or Bancontact, and saving the payment method as a SEPA Direct Debit token. You can read more about this Stripe behavior in the following links:
- https://docs.stripe.com/billing/subscriptions/bancontact
- https://docs.stripe.com/payments/ideal/set-up-payment

Settings preview:
Screenshot 2025-08-28 at 15 13 58

## Testing instructions

- Checkout and build this branch on your test environment (`update/split-sepa-tokens-for-other-methods-setting`)
- Connect your Stripe account
- Set your store currency to EUR
- Enable Bancontact and iDEAL
- Confirm you can see both settings and they are enabled by default
- Confirm you can disable/enable the settings and that by reloading the page, the change is persisted
- As a shopper, with the first setting enabled, try to add a new iDEAL method
- Confirm the method is saved
- Disable the setting and confirm you are not able to save the method anymore
- Repeat the test steps for Bancontact

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


annemirasol (commented):
Question about the naming for the iDEAL setting. I'm seeing that we're using `is_sepa_tokens_for_other_methods_enabled`, while we're using a specific name for Bancontact (`is_sepa_tokens_for_bancontact_enabled`). Why are we not renaming it to be iDEAL-specific?

wjrosa:
> Question about the naming for the iDEAL setting. I'm seeing that we're using `is_sepa_tokens_for_other_methods_enabled`, while we're using a specific name for Bancontact (`is_sepa_tokens_for_bancontact_enabled`). Why are we not renaming it to be iDEAL-specific?

Good question, Anne. The only reason is to avoid the need for a new migration script (to convert the existing option into the new ones). That was the simplest solution. But I am happy to go the other way if you believe it is better.

diegocurbelo (commented):
The changes look good, but I agree that renaming `...for_other_methods` to `...for_ideal` would make the code easier to follow/understand long term, even if we need an initial migration.

wjrosa:
Alright, folks! I renamed all references to the old setting name, updated the implementation, deprecated public methods on the backend, and added a migration script in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4633/commits/2e9b1d96651dd3866866ab9ac8c714ffb821a516

includes/admin/stripe-settings.php
```diff
@@ -226,10 +226,18 @@
 			'desc_tip'    => true,
 		],
 		'sepa_tokens_for_other_methods'       => [
```
annemirasol: Should we rename this as well? It is still `sepa_tokens_for_other_methods`

client/settings/payments-and-transactions-section/index.js
```diff
@@ -99,11 +104,23 @@ const PaymentsAndTransactionsSection = () => {
 					checked={ isSepaTokensForOtherMethodsEnabled }
 					onChange={ setIsSepaTokensForOtherMethodsEnabled }
```
annemirasol: Should we rename these as well?

client/settings/payments-and-transactions-section/index.js
```diff
@@ -19,6 +19,7 @@ import {
 	useEnabledPaymentMethodIds,
 	useIsShortAccountStatementEnabled,
 	useSepaTokensForOtherMethods,
```
annemirasol: I think we need to rename this too -- we renamed the settings hook in `client/data/settings/hooks.js`.

annemirasol (commented):
Thanks for the renames and adding a migration script. Found a few more renames we might have missed.

wjrosa:
Oops! Thanks for the heads up! Replaced all the remaining occurrences in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4633/commits/11b54b8f4d539fada64eade07fadfac75ff8b7d0

includes/migrations/class-sepa-tokens-for-other-methods-settings-update.php
```diff
@@ -0,0 +1,50 @@
...
+		$stripe_gateway = $this->get_gateway();
+
+		// If the new settings are already set, skip migration.
+		$valid_values = [ 'yes', 'no' ];
+		if ( in_array( $stripe_gateway->get_option( 'sepa_tokens_for_ideal' ), $valid_values, true )
+			|| in_array( $stripe_gateway->get_option( 'sepa_tokens_for_bancontact' ), $valid_values, true ) ) {
+			return;
+		}
+
+		$value = $stripe_gateway->get_option( 'sepa_tokens_for_other_methods', 'no' );
```
diegocurbelo, line 39: [nit] We can't remove the `sepa_tokens_for_other_methods`, as we want to keep the current behavior in case of a rollback/downgrade... but maybe we should create a follow-up issue (or add a TODO here) to remove it in 10.2 (L-2) so we don't keep unused options in the settings

diegocurbelo (approved):
Thanks for the changes... it looks good and tests well 🚢

annemirasol (approved):
LGTM!

includes/migrations/class-sepa-tokens-for-other-methods-settings-update.php
wjrosa: Good idea. Added it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4633/commits/ff32196943340d1a4abbd041c0f324a1e82bc105

PR #4644, opened 2025-09-03
Fix multiple PHP warnings

daledupreez:

This PR fixes three code locations that can generate warnings in at least some cases. They definitely feel like edge cases, but we can make our code more robust regardless. The three fixes are as follows:
 * In the code responsible for building data for Javascript execution, we are now protecting against the case where no publishable keys are present
   - It's worth noting that the fallback value for this case is `''` rather than null, as the client code uses `''` as the fallback value at present (based on [this code](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/671721f51a73bfdc5605b74e0543472fa0f797b7/client/data/account-keys/hooks.js#L84))
 * In the code responsible for finding custom checkout fields, we detect when there aren't any checkout fields to process
 * In the code responsible for getting the statement descriptor, we now handle the case where we don't have the expected account data

## Testing instructions

In general, inspection should be sufficient, as the code changes are pretty small and add explicit fallback values for array accesses.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-rest-stripe-account-controller.php
+++ b/includes/admin/class-wc-rest-stripe-account-controller.php
@@ -116,7 +116,7 @@ class WC_REST_Stripe_Account_Controller extends WC_Stripe_REST_Base_Controller {
 		// Use statement descriptor from settings, falling back to Stripe account statement descriptor if needed.
 		$statement_descriptor = WC_Stripe_Helper::clean_statement_descriptor( $this->gateway->get_option( 'statement_descriptor' ) );
 		if ( empty( $statement_descriptor ) ) {
-			$statement_descriptor = $account['settings']['payments']['statement_descriptor'];
+			$statement_descriptor = $account['settings']['payments']['statement_descriptor'] ?? null;
 		}
 		if ( empty( $statement_descriptor ) ) {
 			$statement_descriptor = null;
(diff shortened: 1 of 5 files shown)


includes/admin/class-wc-rest-stripe-account-controller.php
```diff
@@ -116,7 +116,7 @@ public function get_account_summary() {
 		// Use statement descriptor from settings, falling back to Stripe account statement descriptor if needed.
 		$statement_descriptor = WC_Stripe_Helper::clean_statement_descriptor( $this->gateway->get_option( 'statement_descriptor' ) );
 		if ( empty( $statement_descriptor ) ) {
-			$statement_descriptor = $account['settings']['payments']['statement_descriptor'];
+			$statement_descriptor = $account['settings']['payments']['statement_descriptor'] ?? null;
 		}
 		if ( empty( $statement_descriptor ) ) {
```
wjrosa, line 121: With the change above, maybe this IF is not needed anymore (but I am OK with leaving it here for safety)

wjrosa (approved):
Changes look good to me! Left an optional suggestion

diegocurbelo (approved):
LGTM

includes/admin/class-wc-rest-stripe-account-controller.php
daledupreez: I think I'd rather keep it in place to protect against the value in the setting being `''` or some similar empty value.

PR #4647, opened 2025-09-03
Adding debug metadata to payment intents to investigate BNPLs issues

wjrosa:
Slack p1756327430065689-slack-C07NTGXP12N

To help with the issue related to BNPLs not showing up for eligible merchants, I am adding new metadata keys to the payment intent object. This will allow us to check some of the merchants' transactions and confirm their setup on the Stripe side, avoiding the need to request a system status report.

It would have been better to add this data to the Stripe account object, but we don't make any changes there. Another idea was to include new headers, but Stripe does not currently support it.

## Testing instructions

- Checkout and build this branch on your test environment (`dev/bnpl-metadata-debug-information`)
- Connect your Stripe account
- Go to any checkout page and complete the purchase
- Go to your Stripe dashboard, inspect the new transaction and confirm the new metadata keys are there (`is_legacy_checkout_enabled`, and `is_oc_enabled`)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

cc @daledupreez


Mayisha (approved):
Code works as described and I can see the metadata added to my Stripe transaction.

Though it does not solve the problem directly, I see no harm in adding a couple of metadata if that helps.

malithsen (approved):
LGTM! I'd prefer `true|false` instead of `1|0`,  but that's not a blocker

Image

wjrosa:
We send the value as a boolean, but somehow it gets converted to an integer on the Stripe side. I can dig into this a bit later, but merging it as-is for now.

daledupreez:
@wjrosa, I am just taking a look now. Should we also record the settings sync status? It feels like that would be very useful, especially when trying to understand which sites are (or are not) synchronising their payment methods.

wjrosa:
> @wjrosa, I am just taking a look now. Should we also record the settings sync status? It feels like that would be very useful, especially when trying to understand which sites are (or are not) synchronising their payment methods.

It makes sense. Do you mean something like this? https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4650

daledupreez:
Yes! Let's continue the conversation in that PR, @wjrosa!

PR #4639, opened 2025-09-01
Updating Webpack packages

wjrosa:
Fixes STRIPE-698

## Changes proposed in this Pull Request:

As part of the Node upgrade project, this PR upgrades only the webpack-related packages to match the main PR. Webpack packages are used to build the frontend files.

## Testing instructions

- Checkout and attempt to build this branch on your test environment (`dev/upgrading-webpack-packages`)
- Confirm everything works fine

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4650, opened 2025-09-05
Adding the PMC enabled info to BNPL debug metadata

wjrosa:
See https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4647#issuecomment-3257925517

This PR just adds the `pmc_enabled` metadata to payment intents. The goal is to make it easier to debug when BNPLs are not showing up, allowing us to look directly in the Stripe dashboard instead of relying on the system status report.

## Testing instructions

- Checkout and build this branch on your test environment (`dev/add-pmc-enabled-info-to-bnpl-debug-metada`)
- Connect your Stripe account
- Go to any checkout page and complete the purchase
- Go to your Stripe dashboard, inspect the new transaction and confirm the new metadata key is there (`pmc_enabled`)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -2190,7 +2190,7 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Gateway_Stripe {
 			[
 				'is_legacy_checkout_enabled' => ! WC_Stripe_Feature_Flags::is_upe_checkout_enabled(),
 				'is_oc_enabled'              => $this->is_oc_enabled(),
-				'pmc_enabled'                => 'yes' === $this->get_option( 'pmc_enabled' ),
+				'pmc_enabled'                => $this->get_option( 'pmc_enabled' ),
 			]
 		);
 	}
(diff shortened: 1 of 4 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2190,6 +2190,7 @@ public function add_bnpl_debug_metadata( $metadata, $order ) {
 			[
 				'is_legacy_checkout_enabled' => ! WC_Stripe_Feature_Flags::is_upe_checkout_enabled(),
 				'is_oc_enabled'              => $this->is_oc_enabled(),
+				'pmc_enabled'                => 'yes' === $this->get_option( 'pmc_enabled' ),
```
daledupreez: Do we want to capture more possible states for this value? It feels like it would help to know if the value is disabled vs non initialised, as an example.

wjrosa: That's a good point. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4650/commits/9b71b6eb8447c23979da927587b1d6c157581738

malithsen (approved):
LGTM, value of `pmc_enabled` shows up in meta data section

Image

As an aside, `npm run start` was failing for me on this branch. But that seems to be coming from the changes in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4639. Tested this branch by reverting the webpack changes locally.

Mayisha (approved):
Looks good to me. I can see the `pmc_enabled` attribute in the metadata 👍 

Screenshot 2025-09-08 at 2 30 04 PM


#### Optional suggestion
We can use `yes`/`no` for the `is_legacy_checkout_enabled` and `is_oc_enabled` also similar to the pmc flag, instead of `0`/`1`

PR #4658, opened 2025-09-09
Adding support for subscriptions with Klarna

wjrosa:
Fixes STRIPE-704

As part of our efforts to address issues that reduce the number of merchants displaying Klarna as a payment method option, this PR enables its usage for subscription products. It also allows shoppers to add it as a saved payment method.

## Testing instructions

- Checkout to this branch on your test environment (`add/subscription-support-for-klarna`)
- Connect a US Stripe account
- Enable Klarna as a payment method
- Create a subscription product (price above 50 USD)
- As a shopper, add this subscription to your cart
- Go to the checkout pages
- Confirm Klarna is now available, and that you can complete a purchase
- Confirm you can add a Klarna payment method using the My Account page

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
Note: I am fixing the changelog entries in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4661

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,108 @@
...
+		return sprintf( __( 'Klarna (%s)', 'woocommerce-gateway-stripe' ), $this->format_dob( $dob ) );
+	}
+
+	/**
+	 * Sets the Klarna token's date of birth.
+	 *
+	 * @param object $dob The date of birth.
+	 */
+	public function set_dob( $dob ) {
+		$this->set_prop( 'dob', $dob );
```
daledupreez, line 49: Should we set this as an object or should we normalise it before saving? It feels like using `formatted_dob` as a saved field would be easier to work with in the long term.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,108 @@
...
+			if ( null === $method_dob && null === $this->get_dob() ) {
+				return true;
+			}
+
+			$formated_method_dob = $this->format_dob( $method_dob );
+			$formated_dob        = $this->format_dob( $this->get_dob() );
+			if ( $formated_method_dob === $formated_dob ) {
+				return true;
+			}
+		}
```
daledupreez: High level comment: it feels like this might be easier to reason about if we use early returns for cases where there are not matches. (See the suggestion for my general thinking.)

Implementation comments:
* We are not comparing the user ID. Is that intentional? If we don't have a DOB, we have no other way of distinguishing Klarna tokens, which concerns me.
* For the first check, should we not be checking `$payment_method->get_type()`?
```suggestion
		if ( WC_Stripe_Payment_Methods::KLARNA !== $payment_method->get_type() ) {
			return false;
		}
		// Maybe user ID check?

		$payment_method_dob = $payment_method->klarna->dob ?? null;
		$this_dob            = $this->get_dob();
		if ( $this_dob === null ) {
			return null === $payment_method_dob;
		}
	
		if ( null === $payment_method_dob ) {
			return false;
		}

		$formatted_method_dob = $this->format_dob( $payment_method_dob );
		$formatted_dob        = $this->format_dob( $this_dob );

		return $formatted_method_dob === $formatted_dob;
```

daledupreez (changes requested):
Overall, this seems to be working for me, with one exception: we haven't added code to provide the payment method name in the Payment Methods screen:
Screenshot 2025-09-11 at 14 32 22

I think we should add a case for Klarna to `WC_Stripe_Payment_Tokens->get_account_saved_payment_methods_list_item()`.

As a side note, it also looks like the logic in that function could generate exceptions if any other payment gateway uses the same type value for tokens. I think we should also add a guard for non-Stripe classes to that code so we don't assume token classes from other payment gateways match our implementations. I'll take this on in a separate diff.

daledupreez:
As a quick follow-up, I created https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4666 to implement the check/guard I flagged in my review above.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: Good idea. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/cea090365067ce403433d21539d86008f1970ac8. I just decided the property name as `dob`. Let me know if you think we should actually use `formatted_dob` instead

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,107 @@
...
+	 * @param  string $deprecated Deprecated since WooCommerce 3.0
+	 * @return string The name of the token to display
+	 */
+	public function get_display_name( $deprecated = '' ) {
+		if ( empty( $this->get_dob() ) ) {
+			return __( 'Klarna', 'woocommerce-gateway-stripe' );
+		}
+
+		// Translators: %s is the customer's date of birth.
+		return sprintf( __( 'Klarna (%s)', 'woocommerce-gateway-stripe' ), $this->get_dob() );
```
malithsen: Is it necessary to show the complete DOB? Should we mask it? I am wondering if this would raise PII concerns.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	/**
+	 * Checks if the payment method token is equal a provided payment method.
+	 *
+	 * @inheritDoc
+	 */
+	public function is_equal_payment_method( $payment_method ): bool {
+		if ( WC_Stripe_Payment_Methods::KLARNA !== $this->get_type() ) {
+			return false;
```
malithsen, line 82: Is there a special step I need to do to make Klarna return `dob` here? In my testing, Klarna never asks for a dob, so this is always a `null` for me.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: Good points, Dale. I have updated the comparison logic in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/f8b76e3e364c938ac9598b01755136895a7197a6, since I could not auto-apply your suggestion here.

The payment token owner check is not needed because we retrieve the current user's tokens before performing the check:
```php
	public static function get_duplicate_token( $payment_method, $user_id, $gateway_id ) {
		// Using the base method instead of `WC_Payment_Tokens::get_customer_tokens` to avoid recursive calls to hooked filters and actions
		$tokens = WC_Payment_Tokens::get_tokens(
			[
				'user_id'    => $user_id,
				'gateway_id' => $gateway_id,
				'limit'      => 100,
			]
		);
		foreach ( $tokens as $token ) {
			/**
			 * Token object.
			 *
			 * @var WC_Payment_Token_CashApp|WC_Stripe_Payment_Token_CC|WC_Payment_Token_Link|WC_Payment_Token_SEPA|WC_Payment_Token_ACH|WC_Payment_Token_ACSS|WC_Payment_Token_Becs_Debit $token
			 */
			if ( $token->is_equal_payment_method( $payment_method ) ) {
				return $token;
			}
		}
		return null;
	}
```

I drafted [this other PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4667) to improve the other tokens later, following the same idea here.

malithsen (commented):
Nice work. This works as described. Left a couple of questions in the comments

wjrosa:
> I think we should add a case for Klarna to WC_Stripe_Payment_Tokens->get_account_saved_payment_methods_list_item().

Oops, I forgot about that. Added it now in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/7cb148bd21d01ec3693fcc57538f5ba359d679ac

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: Not really. Thanks for flagging this! I have masked the data in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/110185c58c5b74863627c5768fc547cc5842c751

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: It is the same for me. I just saw this property in the docs. Perhaps it is only returned in production. I will ask Stripe about it. Edit: asked them in p1757617023277309-slack-C9976E5MJ

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -37,54 +37,74 @@ class WC_Payment_Token_Klarna extends WC_Payment_Token implements WC_Stripe_Paym
...
 	public function get_dob() {
 		return $this->get_prop( 'dob' );
 	}
 
+	/**
+	 * Formats and sets the Klarna token's date of birth.
+	 *
+	 * @param object $dob The raw date of birth object (from Stripe's API).
+	 */
+	public function set_formated_dob( $dob ) {
```
daledupreez: Nit: here and elsewhere. `formatted` should have 2 `t`s. It's not just A8c. 😁

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -37,54 +37,74 @@ class WC_Payment_Token_Klarna extends WC_Payment_Token implements WC_Stripe_Paym
...
-			if ( null === $method_dob && null === $this->get_dob() ) {
-				return true;
-			}
-
-			$formated_method_dob = $this->format_dob( $method_dob );
-			$formated_dob        = $this->format_dob( $this->get_dob() );
-			if ( $formated_method_dob === $formated_dob ) {
-				return true;
-			}
+		if ( WC_Stripe_Payment_Methods::KLARNA !== $this->get_type() ) {
```
daledupreez: Should we not be checking `$payment_method->get_type()` here? As written, this condition will never be false.
```suggestion
		if ( WC_Stripe_Payment_Methods::KLARNA !== $payment_method->get_type() ) {
```

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: Oops. I trusted my IDE, but it did not correct the spelling 😅. Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/51717d79c7f538a8cfa2a1303b534a7776fa1f84

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	/**
+	 * Formats and sets the Klarna token's date of birth.
+	 *
+	 * @param object $dob The raw date of birth object (from Stripe's API).
+	 */
+	public function set_formatted_dob( $dob ) {
+		$this->set_dob( $this->format_dob( $dob ) );
+	}
```
daledupreez: I think we need to tighten up the APIs and possibly naming, as it feels like it will be too easy for someone to save an object in the `dob` data unintentionally. While we have PHPDoc to help people, it feels like we're making bad actions too easy.

I think it might be clearer if we explicitly specify that `set_dob()` only accepts a string argument, and if we change the name of `set_formatted_dob()` to something like `set_dob_object( object $dob )` so it's easier to understand which method to call based on what type of object you have. And if you do it wrong, you'll get an exception rather than unexpected data. (We should still guard against bad data, but that's different from making it easy to end up with bad data!)
```suggestion
	/**
	 * Sets the Klarna token's date of birth.
	 *
	 * @param string $dob The formatted date of birth object (YYYY-mm-dd).
	 */
	public function set_dob( string $dob ) {
		$this->set_prop( 'dob', $dob );
	}

	/**
	 * Fetches Klarna token's date of birth (formatted to Y-m-d).
	 *
	 * @return string The Klarna token's date of birth.
	 */
	public function get_dob(): string {
		$dob = $this->get_prop( 'dob' );
		if ( is_string( $dob ) ) {
			return $dob;
		}
		return '';
	}

	/**
	 * Sets the Klarna token's date of birth based on the object returned from the Stripe payment method API.
	 *
	 * @param object $dob The raw `dob` object from Stripe.
	 * @see https://docs.stripe.com/api/payment_methods/object#payment_method_object-klarna-dob
	 */
	public function set_dob_from_object( object $dob ) {
		$this->set_dob( $this->format_dob( $dob ) );
	}
```

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,128 @@
...
+	 * Checks if the payment method token is equal a provided payment method.
+	 *
+	 * @inheritDoc
+	 */
+	public function is_equal_payment_method( $payment_method ): bool {
+		if ( WC_Stripe_Payment_Methods::KLARNA !== $this->get_type() ) {
+			return false;
+		}
+
+		$payment_method_dob = $payment_method->klarna->dob ?? null;
```
daledupreez, line 85: Question: what happens if `dob` is returned as `{}`? Or what happens if only some of the fields are null?

It feels like our logic should guard against that, as the Stripe docs only mention that all three fields are nullable integers.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,128 @@
...
+		return 'woocommerce_payment_token_klarna_get_';
+	}
+
+	/**
+	 * Formats the date of birth for display.
+	 *
+	 * @param $dob object The date of birth object.
+	 * @return string The formatted date of birth.
+	 */
+	protected function format_dob( $dob ) {
```
daledupreez: Nit: While this is a protected method, we may as well explicitly type this argument.
```suggestion
	protected function format_dob( object $dob ) {
```

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Method_Test.php
```diff
@@ -165,6 +167,21 @@ class WC_Stripe_UPE_Payment_Method_Test extends WC_Mock_Stripe_API_Unit_Test_Cas
...
+	/**
+	 * Base template for Stripe's Klarna
+	 */
+	const MOCK_KLARNA_PAYMENT_METHOD_TEMPLATE = [
+		'id'                                  => 'pm_mock_payment_method_id',
+		'type'                                => WC_Stripe_Payment_Methods::KLARNA,
+		WC_Stripe_Payment_Methods::KLARNA => [
+			'dob' => [
+				'day'   => 1,
+				'month' => 1,
```
daledupreez: Thought: it may make sense to set this to 2 or some value other than 1, just to make sure we never mix up the day and month.
```suggestion
				'month' => 2,
```

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Method_Test.php
```diff
@@ -888,6 +908,14 @@ public function test_create_payment_token_for_user() {
 					$this->assertTrue( WC_Payment_Token_Becs_Debit::class === get_class( $token ) );
 					$this->assertSame( $token->get_last4(), $becs_debit_payment_method_mock->{WC_Stripe_UPE_Payment_Method_Becs_Debit::STRIPE_ID}->last4 );
 					break;
+				case WC_Stripe_UPE_Payment_Method_Klarna::STRIPE_ID:
+					$klarna_payment_method_mock = $this->array_to_object( self::MOCK_KLARNA_PAYMENT_METHOD_TEMPLATE );
+					$token                      = $payment_method->create_payment_token_for_user( $user_id, $klarna_payment_method_mock );
+					$token_dob_parts            = explode( '-', $token->get_dob() );
+					$this->assertTrue( WC_Payment_Token_Klarna::class === get_class( $token ) );
+					$this->assertSame( (int) $token_dob_parts[1], $klarna_payment_method_mock->{WC_Stripe_UPE_Payment_Method_Klarna::STRIPE_ID}->dob->month );
```
daledupreez, line 917: It feels like we should test the year matches as well.
```suggestion
					$this->assertSame( (int) $token_dob_parts[0], $klarna_payment_method_mock->{WC_Stripe_UPE_Payment_Method_Klarna::STRIPE_ID}->dob->year );
					$this->assertSame( (int) $token_dob_parts[1], $klarna_payment_method_mock->{WC_Stripe_UPE_Payment_Method_Klarna::STRIPE_ID}->dob->month );
```

tests/phpunit/PaymentTokens/WC_Payment_Token_Klarna_Test.php
```diff
@@ -0,0 +1,86 @@
...
+		$this->assertSame( 'Klarna (xxxx-02-01)', $this->token->get_display_name() );
+	}
+
+	/**
+	 * Tests for the DOB getters and setters.
+	 *
+	 * @return void
+	 */
+	public function test_getters_setters() {
+		$this->token->set_formated_dob(
```
daledupreez: Typo: this should be `set_formatted_dob()`.

I have fixed in 24d132d036b686ca9aeabddc189397a679cae1b3.

tests/phpunit/PaymentTokens/WC_Payment_Token_Klarna_Test.php
```diff
@@ -0,0 +1,86 @@
...
+		$this->token->set_dob( '2000-02-01' );
+		$this->assertSame( 'Klarna (xxxx-02-01)', $this->token->get_display_name() );
+	}
+
+	/**
+	 * Tests for the DOB getters and setters.
+	 *
+	 * @return void
+	 */
+	public function test_getters_setters() {
```
daledupreez, line 46: This might benefit from some additional test cases using values > 10 for day and month. But not blocking.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,128 @@
...
+	 * Returns the masked date of birth for display.
+	 *
+	 * @return string
+	 */
+	public function get_masked_dob() {
+		if ( empty( $this->get_dob() ) ) {
+			return '';
+		}
+
+		return 'xxxx' . gmdate( '-m-d', strtotime( $this->get_dob() ) );
```
daledupreez: I think we should be super-careful about how we approach masking this value. I think it might be better to either display only the year, or just drop displaying this value altogether. I honestly don't know how much value there is in showing this value to users.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,128 @@
...
+		$payment_method_dob = $payment_method->klarna->dob ?? null;
+		if ( empty( $this->get_dob() ) ) {
+			return null === $payment_method_dob;
+		}
+
+		if ( null === $payment_method_dob ) {
+			return false;
+		}
+
+		return $this->format_dob( $payment_method_dob ) === $this->get_dob();
```
daledupreez, line 94: Should we also check the user ID for the token against the payment method's owner/customer?

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: Oops! Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/a42448167043cc1dd2b1aa16259c855340fc38d0 (sorry, I did not see your code block suggesiton)

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: Good question. I have included extra validations and changed how we build the date string in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/9c0cf76959a3fe08bf65c29f8e72d0b34c172711

tests/phpunit/PaymentTokens/WC_Payment_Token_Klarna_Test.php
wjrosa: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/6834cd324b61ef4611aee7512a3f42efdc965ee9

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: I agree. I don't see why a shopper would have multiple Klarna tokens with different dates of birth. Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658/commits/8725369c0b7e528b5d4957dafa61ac541e746878

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
wjrosa: This comparison method is used only when searching for duplicate tokens. And we just retrieve tokens from the current user. There's no need to make this comparison, as it would be redundant

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
daledupreez: Looking at the test failures, `$payment_method` is a basic object, so this access actually needs to be `$payment_method->type`. I fixed that in aa39d909a31781352bfe92ec3f971054834a2b27.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,120 @@
...
+	 * @param $dob object The date of birth object.
+	 * @return string The formatted date of birth.
+	 */
+	protected function format_dob( object $dob ) {
+		$dob_parts = [
+			$dob->year ?? null,
+			isset( $dob->month ) ? str_pad( $dob->month, 2, '0', STR_PAD_LEFT ) : null,
+			isset( $dob->day ) ? str_pad( $dob->day, 2, '0', STR_PAD_LEFT ) : null,
+		];
+		return implode( '-', array_filter( $dob_parts ) );
```
daledupreez: This implementation feels like it will break down for edge cases. As an example, if we get one `dob` with only a day value of `10` and another dob object with only a month of `10`, the resulting strings for both objects would be `10`.

I think it would be more robust to guard against all empty values first, and then build a string with placeholders for unknown values, as that should make eventual string comparisons super-clear.
 
```suggestion
		if ( empty( $dob->year ) && empty( $dob->month ) && empty( $dob->day ) ) {
			// Either return null or ''. Empty string probably make more sense.
			return '';
		}
		$dob_parts = [
			$dob->year ?? 'YYYY',
			$dob->month ? str_pad( $dob->month, 2, '0', STR_PAD_LEFT ) : 'MM',
			$dob->day ? str_pad( $dob->day, 2, '0', STR_PAD_LEFT ) : 'DDDD',
		];
		return implode( '-', array_filter( $dob_parts ) );
```

readme.txt
```diff
@@ -111,6 +111,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 9.10.0 - xxxx-xx-xx =
+* Add - Adds support for Klarna subscriptions
```
daledupreez: It may be clearer to merchants if we rephrase this slightly to make it super clear what the impact is.
```suggestion
* Add - Allow Klarna to be used for recurring payments and subscriptions
```

tests/phpunit/PaymentTokens/WC_Payment_Token_Klarna_Test.php
```diff
@@ -0,0 +1,95 @@
...
+
+	/**
+	 * Tests for `is_equal_payment_method()`.
+	 *
+	 * @return void
+	 */
+	public function test_is_equal_payment_method() {
+		$payment_method = (object) [
+			'id'     => 'pm_123',
+			'type'   => WC_Stripe_Payment_Methods::KLARNA,
```
daledupreez, line 74: I added the `type` here in aa39d909a31781352bfe92ec3f971054834a2b27, as standard payment methods always include that field, and we're depending on the field in `is_equal_payment_method()`.

daledupreez (changes requested):
While testing as a logged-in user, I triggered a fatal error in `WC_Stripe_Payment_Tokens->add_token_to_user()` as there isn't a case to handle Klarna. 😬

wjrosa:
> While testing as a logged-in user, I triggered a fatal error in WC_Stripe_Payment_Tokens->add_token_to_user() as there isn't a case to handle Klarna. 😬

I could not trigger the fatal error myself. For some reason, I could add new methods without any issues. Anyway, I have included Klarna to that method.

malithsen (approved):
I [wasn't able](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4658#discussion_r2341807205) to test the end-to-end flow with a dob, but the code changes look good to me and work as described. Thanks for working on this!

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -110,10 +110,14 @@ protected function get_hook_prefix() {
 	 * @return string The formatted date of birth.
 	 */
 	protected function format_dob( object $dob ) {
+		if ( empty( $dob->year ) && empty( $dob->month ) && empty( $dob->day ) ) {
+			// Either return null or ''. Empty string probably make more sense.
```
daledupreez: We can remove this comment. 😁 
```suggestion
```

daledupreez (approved):
I retested only the fatal I had yesterday, where I confirmed the flow that caused it on the code at that time, and then confirmed that it is now working with the current code.

So everything is LGTM! Thanks for your patience in getting everything worked out, @wjrosa.

#### Details for fatal reproduction

In my testing, I triggered the fatal error as follows:
* Log in as a shopper with Klarna not available as a saved payment
* Purchase a subscription with Klarna as the payment method
* Navigate to My Account -> Payment methods
  - This triggered the fatal error for me

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
daledupreez: I went ahead and removed this in 36da9da4f19d27171c62eaf6ece595ed0fc16165.

includes/payment-tokens/class-wc-stripe-klarna-payment-token.php
```diff
@@ -0,0 +1,123 @@
...
+ * @package WooCommerce_Stripe
+ * @since 10.0.0
+ */
+
+// phpcs:disable WordPress.Files.FileName
+
+// Exit if accessed directly
+defined( 'ABSPATH' ) || exit;
+
+class WC_Payment_Token_Klarna extends WC_Payment_Token implements WC_Stripe_Payment_Method_Comparison_Interface {
```
daledupreez, line 16: I missed this before merging: I think we need to use a Stripe-specific class name here, as it feels like a collision could occur way too easily. Is there a reason you picked a generic name, @wjrosa?
```suggestion
class WC_Stripe_Klarna_Payment_Token extends WC_Payment_Token implements WC_Stripe_Payment_Method_Comparison_Interface
```

PR #4660, opened 2025-09-09
Fix fatal error caused by type mismatch

malithsen:
Slack: p1757448726822629-slack-C055WHLA98D

This fixes the following fatal error

```
PHP Fatal error:  Uncaught TypeError: WC_Stripe_Webhook_Handler::run_webhook_received_action(): Argument #3 ($order) must be of type ?WC_Order, false given
```

`resolved_order` is given a `false` value, this doesn't match the type definition of the function parameter which causes the fatal error. `run_webhook_received_action` uses the `$this->resolved_order` directly instead of using the passed parameter, so in this PR we remove the parameter altogether.

## Testing instructions
Code inspection should be sufficient

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1470,6 +1470,9 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	 */
 	private function run_webhook_received_action( string $webhook_type, object $notification ): void {
 		try {
+			// Ensure we have an order or null.
+			$resolved_order = $this->resolved_order ? $this->resolved_order : null;
+
 			/**
 			 * Fires after a webhook has been processed, but before we respond to Stripe.
 			 * This allows for custom processing of the webhook after it has been processed.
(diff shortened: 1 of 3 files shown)


daledupreez (approved):
This looks good to me. I pushed up a small change in 4608e5269198c392debc4024dcd7e79c3a65e2ba that ensures the order argument we pass to the `wc_stripe_webhook_received` action is either null or the order object.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1459,18 +1459,20 @@ public function process_webhook( $request_body ) {
...
 	 *
 	 * @param string $webhook_type The type of webhook that was processed.
 	 * @param object $notification The webhook data sent from Stripe.
-	 * @param WC_Order|null $order The order being processed by the webhook.
 	 */
-	private function run_webhook_received_action( string $webhook_type, object $notification, ?WC_Order $order = null ): void {
+	private function run_webhook_received_action( string $webhook_type, object $notification ): void {
 		try {
+			// Ensure we have an order or null.
+			$resolved_order = $this->resolved_order ? $this->resolved_order : null;
```
```suggestion
			$resolved_order = $this->resolved_order ?? null;
```

wjrosa (approved):
Thanks for fixing this issue, Malith! It looks good to me. I left an optional suggestion (which I think is related to Dale's update)

includes/class-wc-stripe-webhook-handler.php
daledupreez: @wjrosa, the `??` operator only falls through to the right hand value when the initial value is null, which means that `false` would be left as-is.

However, my tag-on PR in #4662 reverses this change after fixing the logic bug that causes the problem, so I think the point is moot either way. 😁

wjrosa: Oh right. Feel free to ignore this then!

malithsen:
e2e failures don't seem related to this change. Merging

PR #4662, opened 2025-09-10
Fix webhook handler action fatals

daledupreez:

This PR addresses the root cause of the type mismatch being addressed in #4660, which occurs because we were setting `$this->resolved_order` to a value before checking it when handling `charge.refund` webhooks. The specific issue occurred when we can't identify the order for the refund, as we hit a situation where we set `$this->resolved_order` and then return. The PR simply changes the order of operations so we return early when we don't find an order.

As a follow-up, this PR also removes the normalisation logic I pushed into #4660 in 4608e5269198c392debc4024dcd7e79c3a65e2ba, as we should now have no situations where we have `false` as a value.

The PR also tweaks the logging slightly:
 * When we fail to identify the order based on the refund ID, we now call the `debug()` method instead of the deprecated `log()` method on the `WC_Stripe_Logger` class
 * When we also fail to identify the order based on the charge ID, we now call the `warning()` method and include both the refund ID and the charge ID in the log message.

## Testing instructions

This is a particularly tricky one to test, and the change is really meant to build on top of #4660. I _think_ inspection should be sufficient.

---

### Changelog entry

<details>

#### Comment 
This PR is intended to be folded into #4660, so that PR should contain our changelog.
</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1470,9 +1470,6 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	 */
 	private function run_webhook_received_action( string $webhook_type, object $notification ): void {
 		try {
-			// Ensure we have an order or null.
-			$resolved_order = $this->resolved_order ? $this->resolved_order : null;
-
 			/**
 			 * Fires after a webhook has been processed, but before we respond to Stripe.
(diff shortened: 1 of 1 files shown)


Mayisha (approved):
I could not reproduce original error. 
However, the changes here look good and safe to me. Moving the assignment of `$this->resolved_order` is logical 👍

malithsen (approved):
Thanks Dale! This change makes sense and should address the root cause.

PR #4663, opened 2025-09-10
Fix webpack live reload

wjrosa:

Our webpack live reload feature (`npm run start`) broke after we upgraded webpack to 5 in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4639. Here in this PR, I am using the solution found in https://github.com/statianzo/webpack-livereload-plugin/issues/63 to fix it. 

We are basically using an alternative version of the livereload plugin compatible with Webpack 5, avoiding the need to upgrade the `@wordpress/scripts` package (that's the package with the livereload dependency).

## Testing instructions

- Checkout to this branch on your test environment (`dev/fix-webpack-live-reload`)
- Run `npm i`
- Run `npm run start` 
- Confirm it works successfully (some warnings are expected until we upgrade Node)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 9.10.0 - xxxx-xx-xx =
+* Dev - Fix live reload issue with Webpack 5
 * Add - Adds support for the Romanian Leu (RON) currency when paying with Klarna
 * Update - Splits the "Enable SEPA for other methods" setting into two separate settings for Bancontact and iDEAL
 * Dev - Upgrades the Webpack-related packages
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "woocommerce-gateway-stripe",
-      "version": "9.8.1",
+      "version": "9.9.0",
       "hasInstallScript": true,
       "license": "GPL-3.0",
       "dependencies": {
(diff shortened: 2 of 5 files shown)


malithsen (approved):
Thanks for the fix Wesley, can confirm live reloading now works!

> avoiding the need to upgrade the @wordpress/scripts package (that's the package with the livereload dependency).

Out of curiosity, are we planning to upgrade `@wordpress/scripts` after the node upgrade?

wjrosa:
Thanks, Malith!

> Out of curiosity, are we planning to upgrade @wordpress/scripts after the node upgrade?

Yes, we are. My initial idea was to create a patch for it before the Node upgrade, but it is not worth it (the process is a bit complicated, and it will be needed just for a short while).

Mayisha:
> Yes, we are. My initial idea was to create a patch for it before the Node upgrade, 

My opinion is the same, I would prefer upgrading the `@wordpress/scripts`.

> it will be needed just for a short while

Just to be sure, @wjrosa what are you referring here? Do you mean the `livereload-webpack-plugin` will be removed when we upgrade the `@wordpress/scripts`?

wjrosa:
Yes, that's correct Mayisha. This change should be needed just for a very small amount of time

Mayisha (approved):
Thanks @wjrosa. Livereload is working now 👍

PR #4669, opened 2025-09-12
Showing connection modal when oauth is needed

wjrosa:
Fixes STRIPE-549
Figma nFI8jCCAQnd19PNwScqVU6-fi

As requested by our design team, this PR changes how the Account Status section appears for merchants who did not connect via OAuth like below:
Screenshot 2025-09-19 at 16 48 01

It also adds a new notice on top of the settings page:
Screenshot 2025-09-19 at 18 06 03

The goal is to incentivise more merchants to connect via OAuth to improve security.

## Testing instructions

- Checkout to this branch on your test environment (`update/showing-connection-modal-when-oath-is-needed`)
- Set up your store to behave as if it did not connect via OAuth by removing the connection_type (or test_connection_type attr from the settings:
```
wp option patch delete woocommerce_stripe_settings test_connection_type
```
- Confirm you can see the different versions of the connect status section and the new notice on top, and that both look like the preview above.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (commented):
The code looks good and it works as described in the PR.... but I think the UX might be confusing.

When the merchant is using manual/shared keys the gateway can still be used for purchases but settings are not available... maybe we can load the onboarding screen by default, but also keep the "Payment Methods" and "Settings" tabs? 
Or change the `Connect an account" buttons to the same "Re-authenticate" button we use in the current CTA... and maybe the tittle, currently is "Get started with Stripe" (both both onboarding and re-connections)

**BEFORE**
Screenshot 2025-09-15 at 6 30 33 PM

**AFTER**
Screenshot 2025-09-15 at 6 29 15 PM


--

Also, there is a small error in the testing instructions:

>Set up your store to behave as if it did not connect via OAuth. You can do it by either:
-- Removing your keys directly in the database after connecting

This actually disconnects the plugin (the gateway can't be used for purchases), to simulate a non-connected merchant, we need to remove the `connection_type` (or `test_connection_type` attr from the settings:
```
wp option patch delete woocommerce_stripe_settings test_connection_type
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -41,11 +41,12 @@ public function __construct() {
 	 * @since 1.0.0
 	 * @version 4.0.0
 	 */
-	public function add_admin_notice( $slug, $class, $message, $dismissible = false ) {
+	public function add_admin_notice( $slug, $class, $message, $dismissible = false, $icon_img_url = '' ) {
```
daledupreez: Is it worth making this argument type explicit?
```suggestion
	public function add_admin_notice( $slug, $class, $message, $dismissible = false, string $icon_img_url = '' ) {
```

wjrosa:
Thanks for the review, Diego!

> .... but I think the UX might be confusing.

Yeah, I agree! @orcungogus could you please advise here?

Another suggestion would be to update the existing "Account details" block instead. Changing the "Configure connection" button to "Re-connect" and some other parts of it:

Screenshot 2025-09-17 at 10 40 43

> Also, there is a small error in the testing instructions:

Oops. Thanks for pointing that out

orcungogus:
> When the merchant is using manual/shared keys the gateway can still be used for purchases but settings are not available...

So, these merchants are still authenticated (with the old method) and they can still make use of the _Payment methods_ and _Settings_ tabs, correct? If so I like Wesley’s suggestion of taking over the account details CTA, just use "Reauthenticate" for that as well. In addition to an inline notice shown there, we should also show an alert notice on both _Payment methods_ and _Settings_ tabs, what do you think?

Stripe - Reconnect Notice - Payment Methods
Stripe - Reconnect Notice - Settings

You can find these comps [here in Figma](https://www.figma.com/design/nFI8jCCAQnd19PNwScqVU6/Stripe-Payments--External-?node-id=2348-10772&t=kLcLsumeKpMHuQW6-4).

wjrosa:
Thanks, Orcun! I think it looks better 👍 . I have updated the implementation in my latest commits to follow the new designs. And also updated the PR description with the previews. I did not include the "dismiss" link, as we already have the button on the top right.

client/settings/account-details/__tests__/index.test.js
```diff
@@ -89,4 +93,29 @@ describe( 'AccountDetails', () => {
...
+					},
+					payouts_enabled: true,
+					charges_enabled: true,
+				},
+			},
+		} );
+		render( <AccountDetails /> );
+
+		expect(
+			screen.queryByText( /reconnect your Stripe account/i )
```
daledupreez: Is it worth using the longer text to be clearer as to what we're looking for? That also means we should also get clearer signals when there are failures.

client/settings/account-details/reconnect-notice.js
```diff
@@ -0,0 +1,35 @@
...
+			fill: #8a2424;
+		}
+	}
+`;
+
+export const ReconnectNotice = () => {
+	return (
+		<NoticeContainer>
+			<span data-testid="help">
+				<Icon icon={ info } size="18" />
```
daledupreez: I believe the design calls for the `caution` icon rather than `info`.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -73,7 +74,7 @@ public function admin_notices() {
 		}
 
 		foreach ( (array) $this->notices as $notice_key => $notice ) {
-			echo '<div class="' . esc_attr( $notice['class'] ) . '" style="position:relative;">';
+			echo '<div class="' . esc_attr( $notice['class'] ) . '" style="position:relative; overflow: auto;">';
```
daledupreez: Given that this will impact all notices, I think we need some explicit documentation for why its needed, and what we should look for in other notices.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -393,6 +411,29 @@ public function stripe_check_environment() {
...
+				$stripe_connect = woocommerce_gateway_stripe()->connect;
+
+				// Check each mode only if it has keys
+				$needs_live_oauth = $has_live_keys && ! $stripe_connect->is_connected_via_oauth( 'live' );
+				$needs_test_oauth = $has_test_keys && ! $stripe_connect->is_connected_via_oauth( 'test' );
+
+				$oauth_required = $needs_live_oauth || $needs_test_oauth;
+				if ( $oauth_required ) {
+					$message = __( 'Please reconnect to continue using Stripe and avoid disruptions on your store.', 'woocommerce-gateway-stripe' );
+					$link    = esc_url( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings&highlight=account-details' ) );
```
daledupreez, line 436: As a minor note, the `highlight` option doesn't seem to support the account details section. Not sure if that's something you want to tackle in this PR, or simply remove the query parameter for now.

assets/images/info.svg
daledupreez, line 1: It looks like we're using the `info` icon from `@wordpress/icons`, so I don't believe we need this file.

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -76,13 +127,30 @@ public function test_no_notices_are_shown_when_stripe_is_not_enabled() {
 	}
 
 	/**
+	 * Test that the correct notices are shown in all scenarios.
```
daledupreez, line 130: Thanks for all the PHPDocs!

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -109,107 +189,27 @@ public function test_correct_stripe_notices_are_shown_in_all_scenarios( $options
...
-		$this->assertMatchesRegularExpression( '/Your customers cannot use Stripe on checkout/', $notices->notices['keys']['message'] );
-	}
-
-	public function options_to_notices_map() {
+	/**
+	 * Data provider for `test_correct_stripe_notices_are_shown_in_all_scenarios`.
+	 *
+	 * @return array
+	 */
+	public function options_to_notices_map(): array {
```
daledupreez, line 206: Good call on co-locating the data provider and test runner functions.

daledupreez (commented):
Everything is working as expected, but we are not using the correct icon, and we're adding an unused SVG file. I am also concerned about the additional style rule being applied to all notices.

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: I have removed the icon following the latest suggestion from Orcun above. So, this parameter is no longer needed.

client/settings/account-details/__tests__/index.test.js
wjrosa: Sure. I just did it that way due to how the previous tests were written + Copilot suggestion. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4669/commits/b16227d1fb37415c83e2e32f53ed78cc3c295bdf

client/settings/account-details/reconnect-notice.js
wjrosa: I don't see the `caution` icon as part of the package 🤔  (as `alert` as it was named in Figma):
Screenshot 2025-09-23 at 14 26 23

`info` was the closest I could find. I can export the icon as SVG and use it instead if that's better

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: I decided to keep the `overflow` style only when there are actions to render. And also added a comment about it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4669/commits/b66a54db8b563cfc76cc838ab4a67e8fbb922c58

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Oops, I forgot to implement that 🤦  . Added it now in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4669/commits/39483773089edae6e49da5fbf42f91d988ec0152

assets/images/info.svg
wjrosa: I can use it in the above if you think that's better. I will wait for this decision before removing it

client/settings/account-details/reconnect-notice.js
daledupreez: I think this is a versioning/dependency issue -- the `caution` icon was added in late 2024 as part of https://github.com/WordPress/gutenberg/pull/66555.

diegocurbelo (approved):
Thanks for the changes @wjrosa. It works as expected!

client/settings/account-details/reconnect-notice.js
wjrosa: Nice find! I have upgraded the package and replaced the icon in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4669/commits/0b9e97ba0625d0ecf33eec01e4fc504461525993

wjrosa: Preview:
Screenshot 2025-09-23 at 17 47 08

assets/images/info.svg
wjrosa: Deleted it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4669/commits/48270ef11e79416dda97072de0d9accb2bcfb7ad

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -73,7 +74,14 @@ public function admin_notices() {
 		}
 
 		foreach ( (array) $this->notices as $notice_key => $notice ) {
-			echo '<div class="' . esc_attr( $notice['class'] ) . '" style="position:relative;">';
+			$has_actions = count( $notice['actions'] ) > 0;
```
daledupreez: Just to be safe and prevent warnings, I think we should make sure that `$notice['actions']` is defined and countable.
```suggestion
			$has_actions = count( $notice['actions'] ?? [] ) > 0;
			// Or even something like this:
			$notice_actions = $notice['actions'] ?? [];
			$has_actions   = is_array( $notice_actions ) && [] !== $notice_actions;
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -73,7 +74,14 @@ public function admin_notices() {
...
+			$div_style   = 'position:relative;';
```
daledupreez: Super-tiny nit: I think it would be easier to read if we swapped these two lines so the unconditional set of `$div_style` occurs before we start running any logic related to `$has_actions`.
```suggestion
			$div_style   = 'position:relative;';
			$has_actions = count( $notice['actions'] ) > 0;
```

package.json
```diff
@@ -44,7 +44,7 @@
     "@wordpress/eslint-plugin": "^17.13.0",
     "@wordpress/hooks": "3.2.0",
     "@wordpress/i18n": "^4.2.1",
-    "@wordpress/icons": "^5.0.1",
+    "@wordpress/icons": "10.31.0",
```
daledupreez: Do we want this to be a `^`-prefixed dependency? Or should we keep it as an explicit dependency?
```suggestion
    "@wordpress/icons": "^10.31.0",
```

client/settings/payment-settings/account-details-section.js
```diff
@@ -93,15 +94,32 @@ const AccountSettingsDropdownMenu = ( {
...
+			} );
+		}
+	}, [ headingRef ] );
 
 	return (
 		<Card className="account-details">
 			<CardHeader>
 				<HeaderDetails>
-					<h4>
+					<h4 ref={ headingRef }>
```
daledupreez, line 122: My only minor nit is that when I click on the notice CTA from the main WordPress dashboard, the auto-scroll moves me to a weird scroll location. I am not sure that we can improve this much.

Screenshot 2025-09-24 at 15 28 06

daledupreez (approved):
Changes are looking good to me and testing well. My only issue was with the auto-scroll when coming from the main wp-admin dashboard when using Chrome.

package.json
wjrosa: Let's set the prefixed version 👍

client/settings/payment-settings/account-details-section.js
wjrosa: Yeah. If I set the ref to the button, the settings bar covers it:
Screenshot 2025-09-24 at 11 26 17

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Alright. Refactored it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4669/commits/d56e1ea6699e0f2f5b2f3b82a6c5e91eab30776b

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Swapped them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4669/commits/9ea55bc0a27e8650fb6ae20d13726b16b6064f5e

PR #4670, opened 2025-09-12
Upgrading node to v20

wjrosa:
Fixes STRIPE-446
Original PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4264

This is the main PR upgrading the extension Node version to 20. It is basically a recreation of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4264. The purpose is to keep the git history here cleaner, making it easier to review.

After upgrading many necessary packages, the ones we are upgrading here require Node 20, so they must go along with the main version change.

## Testing instructions

- Checkout to this branch on your test environment (`dev/node-20-main-upgrade`)
- Confirm you can successfully build the frontend files for release: `npm run build:webpack && npm run build:release`
- Perform some basic smoke-testing on the frontend
- Confirm the frontend unit tests are still working: `npm run test:js`

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


bin/docker-setup.sh
```diff
@@ -103,7 +103,7 @@ cli wp option set woocommerce_allow_tracking "no"
 cli wp option set woocommerce_coming_soon "no"
 
 echo "Importing WooCommerce shop pages..."
-cli wp --user=admin wc tool run install_pages
+cli wp wc --user=admin tool run install_pages
```
daledupreez: This feels like it is unrelated to the node version, so could be worth dropping from this PR.

package.json
```diff
@@ -29,7 +28,7 @@
     "@typescript-eslint/parser": "^6.21.0",
     "@woocommerce/currency": "^3.2.0",
     "@woocommerce/data": "^1.4.0",
-    "@woocommerce/dependency-extraction-webpack-plugin": "^1.4.0",
+    "@woocommerce/dependency-extraction-webpack-plugin": "^3.1.0",
```
daledupreez, line 32: It looks like version 4.0.0 was released in June per [the npm page for the package](https://www.npmjs.com/package/@woocommerce/dependency-extraction-webpack-plugin) -- we should probably aim for that version rather than 3.x.

tests/e2e/bin/setup.sh
```diff
@@ -116,12 +116,12 @@ redirect_output cli wp option set woocommerce_coming_soon "no"
...
+redirect_output cli wp wc --user=${ADMIN_USER} tool run install_pages
 
 echo " - Configuring Shipping and Taxes"
-redirect_output cli wp --user=${ADMIN_USER} wc shipping_zone create --name="Everywhere" --order=1
-redirect_output cli wp --user=${ADMIN_USER} wc shipping_zone_method create 1 --method_id="flat_rate"
-redirect_output cli wp --user=${ADMIN_USER} wc shipping_zone_method create 1 --method_id="free_shipping"
+redirect_output cli wp wc shipping_zone create --name="Everywhere" --order=1 --user=${ADMIN_USER}
+redirect_output cli wp wc shipping_zone_method create 1 --method_id="flat_rate" --user=${ADMIN_USER}
+redirect_output cli wp wc shipping_zone_method create 1 --method_id="free_shipping" --user=${ADMIN_USER}
 redirect_output cli wp option update --format=json woocommerce_flat_rate_1_settings '{"title":"Flat rate","tax_status":"taxable","cost":"10"}'
```
daledupreez: These changes feel unrelated to the node upgrade. I think we should drop them from the PR.

tests/e2e/utils/playwright-setup.js
```diff
@@ -254,12 +254,12 @@ export const setupWoo = async () => {
...
-		`wp --user=${ ADMIN_USER } wc tool run install_pages`,
+		`wp wc --user=${ ADMIN_USER } tool run install_pages`,
 		'wp plugin install wordpress-importer --activate',
 		'wp import wp-content/plugins/woocommerce/sample-data/sample_products.xml --authors=skip',
-		`wp --user=${ ADMIN_USER } wc shipping_zone create --name="Everywhere" --order=1`,
-		`wp --user=${ ADMIN_USER } wc shipping_zone_method create 1 --method_id="flat_rate"`,
-		`wp --user=${ ADMIN_USER } wc shipping_zone_method create 1 --method_id="free_shipping"`,
+		`wp wc shipping_zone create --name="Everywhere" --order=1 --user=${ ADMIN_USER }`,
+		`wp wc shipping_zone_method create 1 --method_id="flat_rate" --user=${ ADMIN_USER }`,
+		`wp wc shipping_zone_method create 1 --method_id="free_shipping" --user=${ ADMIN_USER }`,
```
daledupreez: As above, these don't feel like node-related changes, so I think we should drop them from this PR.

package.json
```diff
@@ -172,8 +175,8 @@
     "changelog": "node bin/changelog.js"
   },
   "engines": {
-    "node": ">=16.17.0",
-    "npm": ">=8.15.0"
+    "node": ">=20.10.0",
```
diegocurbelo: Why not use the same version as in the `.nvmrc` file?

diegocurbelo (approved):
Thanks for working on this @wjrosa... Node 20 woohoo!  👏🏼 😄 👏🏼 

The changes look good, and it works as expected:
- all checks pass
- tests run locally
- npm start and npm run build work (both after update and with a clean install)
- woorelease simulate generates a valid zip package

Mayisha (approved):
Great work @wjrosa 👍 

✅ up, start and build commands run successfully.
✅ Tests are passing locally.
✅ Installed the zip created with the `build` command and installed it in a JN site with no errors.
✅ Payment element and ECE loads correctly on block and classic pages for both OC enabled/disabled.
✅ Placed an order with card.

package.json
wjrosa: I tried to update the version here to match the original version in .nvmrc (v20.19.1). But my system picked 20.18.1 automatically. That broke `npm install` for me. So, I decided to do the inverse: updated .npmrc version to match the node version in package.json instead.

daledupreez: @wjrosa, I think that some of the WooCommerce dependencies need 20.18 or newer. I can't recall the specific version, but I am confident it was newer than 20.10.

daledupreez:
Sorry, I took a quick look at the changes earlier, and just realised that my comments didn't get published. None of them are blocking!

package.json
daledupreez: This can definitely be handled in a follow-up.

bin/docker-setup.sh
wjrosa: Makes sense! Reverted it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4670/commits/9938e922a8591418cc160f468a260a6fb25e8a5f

tests/e2e/bin/setup.sh
wjrosa: Reverted in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4670/commits/9938e922a8591418cc160f468a260a6fb25e8a5f

tests/e2e/utils/playwright-setup.js
wjrosa: Reverted in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4670/commits/9938e922a8591418cc160f468a260a6fb25e8a5f

package.json
wjrosa: I tried to upgrade to the 4.0.0 version, but it breaks our build. Searching for a solution, all I could find was the suggestion to revert the version. So it is best to stick to this version for now.

PR #4681, opened 2025-09-17
Implement cache prefetch for account data

daledupreez:
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4637

This PR builds upon the initial cache prefetch work in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4637 to support prefetching account data. Most of the framework for this was in the previous PR, so this PR is really adding some configuration to set up prefetching for account data 10 seconds before cache expiry, and adding a flag to the code that fetches the account data to bypass the cache.

**I do have a major concern for this PR**: I don't think it's necessary to prefetch the account data, as I believe the account GETs are really fast, so the number of simultaneous GETs after a cache expiry is going to be relatively small and this logic really complicates a relatively simple thing. A particular aspect of that is that we read the account cache multiple times in many requests, so there needs to be additional in-memory guards in the prefetch logic to avoid lots of prefetch queue entries and logged items.

## Testing instructions

* Identify a site where Action Scheduler is working well -- my local dev instance was not a good candidate!
* Also ensure that you have two or more products in the store
* Check out this branch locally - `git checkout add/cache-prefetch-for-account-data`
* Modify the `ACCOUNT_CACHE_EXPIRATION` constant in `includes/class-wc-stripe-account.php` from `2 * HOUR_IN_SECONDS` to `2 * MINUTE_IN_SECONDS`
* Run a build from this branch, and get this version running on your site
* Ensure your site is connected to Stripe and has logging enabled
* Make sure you clear out the existing option/cache as it will last for 2 hours by default - `wp option delete wcstripe_cache_test_account_data` (assuming you have a test connection -- change `_test_` to `_live_` if you have a production/live connection)
* Navigate to your shop, and navigate between your products as quickly as you can for the next 2-3 minutes.
* After you know you have spent at least 2 minutes since loading the payment method listing page, reload the Stripe logs
* Find the initial call to `account`
* Scroll down until ~1m 50s later, and see if you can see a `Enqueued cache prefetch` entry, followed by a few `Cache prefetch already pending` entries, and then a successful `account` API call and a `Successfully prefetched cache key` entry.
  - There may be one or two `Unable to enqueue cache prefetch: Action Scheduler is not initialized or available` entries as well -- in these cases, I think it's OK to simply skip the refetch logic if we can't actually queue up the relevant deferred actions.
 - There should not be a huge number of `Cache prefetch already pending` entries - the account data is fetched a _lot_, so we don't want super-verbose logging. 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (approved):
The code looks good, and tests well!

>I do have a major concern for this PR: I don't think it's necessary to prefetch the account data, as I believe the account GETs are really fast, so the number of simultaneous GETs after a cache expiry is going to be relatively small and this logic really complicates a relatively simple thing. 

I agree with this, but I also like how the logic is encapsulated in the prefetch class... maybe we can make this off by default, and add a filter to choose what to prefetch?

Or maybe we can add a 3rd tab: `Payment Methods`, `Settings`, `Advanced`, in which we can add advanced customization options, like an input for a custom PMC, cache customization (and behind the scenes it jsut adds the corresponding filters)

>A particular aspect of that is that we read the account cache multiple times in many requests, so there needs to be additional in-memory guards in the prefetch logic to avoid lots of prefetch queue entries and logged items.

One of the reasons we read the account cache so many times (and why the TTL is not longer) is that we use it to get the account capabilities (available PMs, PMs that require additional info, etc). If the store is using OCS then the calls should a lot less.

malithsen (approved):

daledupreez:
> > I do have a major concern for this PR: I don't think it's necessary to prefetch the account data, as I believe the account GETs are really fast, so the number of simultaneous GETs after a cache expiry is going to be relatively small and this logic really complicates a relatively simple thing.
> 
> I agree with this, but I also like how the logic is encapsulated in the prefetch class... maybe we can make this off by default, and add a filter to choose what to prefetch?

https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4680 is specifically intended to allow for opting out of prefetch for specific, supported keys. I plan on picking that up after merging this change, so that may be the right place to introduce default-on vs default-off prefetching as well. I will do that tomorrow.

PR #4686, opened 2025-09-19
Upgrading @wordpress/scripts and related packages

wjrosa:
Related https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4483

## Changes proposed in this Pull Request:

In this PR, I am upgrading the `@wordpress/scripts` package version to `^30.24.0`, along with all related dependencies (`jest`, `postcss`, and `postcss-loader`). I am also including `jest-environment-jsdom`, as it is required in newer versions of Jest.

## Testing instructions

- Checkout to this branch on your test environment (`dev/upgrading-wordpress-scripts-and-related-deps`)
- Install NPM packages (`npm i`)
- Confirm you are able to build the frontend files (`npm run build`) without errors
- Confirm tests still passes (`npm run test:js`)

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #3693, opened 2025-01-07
Make the new Stripe Express Checkout Element enabled by default

Mayisha:
## Changes proposed in this Pull Request:
Now `is_stripe_ece_enabled` returns true unless `_wcstripe_feature_ece` is specifically set to `no`.

## Testing instructions
Code review should be enough.

PR #4687, opened 2025-09-19
Removing frontend code related to PRBs from the checkout page

wjrosa:
Towards STRIPE-724
Depends on https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4710

As part of the removal of the legacy checkout, we are also removing code related to Payment Request Buttons (PRBs), which were deprecated along with it.

Another motivation to remove it now is to make our frontend code compatible with the latest version of `@wordpress/scripts`. A PR upgrading it can be seen [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4686).

Here I am removing only the frontend files. The PR removing the backend part can be seen [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4688).

## Testing instructions

- Checkout and build this branch on your test environment (`update/removing-prbs-frontend-code`)
- Connect your Stripe account
- Enable express checkout methods
- As a shopper, access the store using a public accessible URL
- Add a product to your cart
- Confirm express checkout buttons are still shown in cart, checkout, and product pages
- Confirm you can complete a purchase with them

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (changes requested):
@wjrosa We can not remove the files under `client/entrypoints/payment-request-settings/*`. 
These are the codes for the customization page for the express checkout. As the admin settings were exactly the same for PRB and ECE (except for the buttons preview), we did not create new duplicate code for ECE.

Screenshot 2025-09-30 at 9 40 50 PM

Screenshot 2025-09-30 at 9 41 35 PM

We should transfer these settings related code under `client/entrypoints/express-checkout-settings/*` and load the correct assets.

wjrosa:
Thanks for the heads up, Mayisha! I started to rename the files, but I got into a rabbit hole. Then I decided to revert this part to handle it in another PR later.

wjrosa:
Okay, I gave it another try, and now it works. I renamed all the frontend files 👍 . I will leave some stuff, like the global variable, for another time.

includes/admin/class-wc-stripe-payment-requests-controller.php
```diff
@@ -19,7 +19,7 @@ public function __construct() {
 	 */
 	public function admin_scripts() {
 		// Webpack generates an assets file containing a dependencies array for our built JS file.
-		$script_asset_path = WC_STRIPE_PLUGIN_PATH . '/build/payment-requests-settings.asset.php';
+		$script_asset_path = WC_STRIPE_PLUGIN_PATH . '/build/express-checkout-settings.asset.php';
```
Mayisha, line 22: Note to remember, we should rename this controller when we remove the backend side of the code.

Mayisha (approved):
Looks good to me.

✅ Settings page is loaded and the changes made to the express checkout settings persist on save.
✅ Express checkout buttons are loading correctly on all the pages,
- Product
- Shortcode cart
- Shortcode checkout
- Block card
- Block checkout
✅ Appearance of the buttons follow the saved settings.

Screenshot 2025-10-02 at 1 08 51 PM

✅ Can pay using Google Pay without any errors. 

Screenshot 2025-10-02 at 1 06 13 PM

daledupreez:
@wjrosa, I will continue testing this later today, but how should we test the case where someone previously had payment request buttons enabled in an earlier version and then upgrade to this version? Do we have a PR that ensures that path is handled correctly? In my mind, that is the main test path for this, as express checkout has been the default for some time.

client/blocks/upe/index.js
```diff
@@ -59,25 +58,19 @@ Object.entries( paymentMethodsConfig )
...
-	if ( getBlocksConfiguration()?.isPaymentRequestEnabled ) {
-		registerExpressPaymentMethod( expressCheckoutElementApplePay( api ) );
-		registerExpressPaymentMethod( expressCheckoutElementGooglePay( api ) );
-	}
-	if ( getBlocksConfiguration()?.isLinkEnabled ) {
-		registerExpressPaymentMethod( expressCheckoutElementStripeLink( api ) );
-	}
-} else {
-	// Register Stripe Payment Request.
-	// TODO: We can remove this once we're sure everyone on the new checkout (UPE) has been migrated to ECE.
```
daledupreez, line 79: Related to my previous comment about upgrades: is it possible to have UPE + payment request buttons?

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.0.0 - xxxx-xx-xx =
+* Dev - Removes frontend code related to Payment Request Buttons in the checkout page
```
daledupreez: I think we should flag this as an Update so we make it super-clear in the release notes that it is changing.
```suggestion
* Update - Removes frontend code related to Payment Request Buttons in the checkout page
```

daledupreez (commented):
I am going to test some more tomorrow, but I think we need to ship the back end changes before we ship this change, as we could get really weird behaviour in cases where the back end still enables the feature and/or tries to queue up the removed JS files.

client/blocks/upe/index.js
wjrosa: That's mostly impossible. We load either ECE or PRBs. See this code (`includes/payment-methods/class-wc-stripe-payment-request.php#88`):
```php
		// Check if ECE feature flag is enabled.
		if ( WC_Stripe_Feature_Flags::is_upe_checkout_enabled() && WC_Stripe_Feature_Flags::is_stripe_ece_enabled() ) {
			return;
		}
```

`WC_Stripe_Feature_Flags::is_upe_checkout_enabled()` is always `true` now, since we hardcoded the value. Same for `WC_Stripe_Feature_Flags::is_stripe_ece_enabled()`, since the feature flag is enabled (done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3693, in January).

The only way for this to behave differently would be if the merchant manually set the feature flag value to `no`. What we can do here to block this is ignore the feature flag (hardcode in code like we did for UPE).

wjrosa: I went ahead and created https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4710 to handle this specific issue. I will hold this PR until it is merged

daledupreez (approved):
The changes look good to me and are working well. My only note is that I think we should  merge #4710 first or soon after this change, as setting the `_wcstripe_feature_ece` option to `'no'` successfully broke this PR. 😁

PR #4613, opened 2025-08-21
Introducing a Stripe order helper class

wjrosa:
See peNR48-1Jc-p2

## Changes proposed in this Pull Request:

This PR introduces a new helper class to handle order-specific data within the extension. This is related to the previous attempt to add an order model class, which we reverted due to incompatibility. 

Instead of a model, this new class will retrieve and update `WC_Order` data (from existing instances), so we should not encounter instantiation issues.

In this first part, I am adding some querying methods and some flags. In the [second part](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4614), I will add most of the getters and setters.

## Testing instructions

- Code review
- Check if the tests are still passing
- Perform some basic smoke testing to confirm no regression was introduced

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4693, opened 2025-09-23
Handling source id / refund id / intent ids metas in the order helper class

wjrosa:
Base PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4613
P2 peNR48-1Jc-p2

As part of the Order Helper class project, I am including some additional meta keys to be handled by the new class and implementing the new methods across our codebase.

The metas handled here are: `_stripe_source_id`, `_stripe_refund_id`, `_stripe_intent_id`, and `_stripe_setup_intent`.

## Testing instructions

- Code review
- Check if the tests are still passing
- Perform some basic smoke testing to confirm no regression was introduced

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -1700,7 +1702,7 @@ public function save_intent_to_order( $order, $intent ) {
 				$order->update_meta_data( '_stripe_mandate_id', $charge->payment_method_details->acss_debit->mandate );
 			}
 		} elseif ( 'setup_intent' === $intent->object ) {
-			$order->update_meta_data( '_stripe_setup_intent', $intent->id );
+			$order_helper->update_stripe_setup_intent( $order );
```
malithsen: We should pass the intent id here

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -1918,7 +1919,7 @@ public function setup_intent( $order, $prepared_source ) {
 		if ( is_wp_error( $setup_intent ) ) {
 			WC_Stripe_Logger::log( "Unable to create SetupIntent for Order #$order_id: " . print_r( $setup_intent, true ) );
 		} elseif ( WC_Stripe_Intent_Status::REQUIRES_ACTION === $setup_intent->status ) {
-			$order->update_meta_data( '_stripe_setup_intent', $setup_intent->id );
+			WC_Stripe_Order_Helper::get_instance()->update_stripe_setup_intent( $setup_intent->id );
```
malithsen: We should pass the order as the first param

includes/class-wc-stripe-helper.php
```diff
@@ -1319,7 +1318,7 @@ public static function add_payment_intent_to_order( $payment_intent_id, $order )
 			)
 		);
 
-		$order->update_meta_data( '_stripe_intent_id', $payment_intent_id );
+		$order_helper->update_stripe_intent( $order );
```
malithsen: We should pass the intent id as the second parameter

includes/class-wc-stripe-order-helper.php
```diff
@@ -267,6 +260,186 @@ public function delete_stripe_net( ?WC_Order $order = null ) {
...
+	 * @param WC_Order|null $order
+	 * @param string $source_id
+	 * @return false|void
+	 */
+	public function update_stripe_source( ?WC_Order $order = null, string $source_id = '' ) {
+		if ( is_null( $order ) ) {
+			return false;
+		}
+
+		$order->update_meta_data( self::META_STRIPE_SOURCE_ID, $source_id );
```
malithsen, line 293: I wonder if we should immediately save the meta after the update. It would prevent us from forgetting to update it in the caller, but on the other hand, it could be slightly more performant to do it in the caller along with other meta updates.

malithsen (changes requested):
Thanks for these changes. Looks good overall. Requesting changes due to missing parameters in some of the `update*` functions.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Oops, sorry for the silly mistake. Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4693/commits/8a6f634f1a3a93450d8e8d5792c6fd2a4a5e1a54

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4693/commits/8a6f634f1a3a93450d8e8d5792c6fd2a4a5e1a54

includes/class-wc-stripe-helper.php
wjrosa: Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4693/commits/8a6f634f1a3a93450d8e8d5792c6fd2a4a5e1a54

includes/class-wc-stripe-order-helper.php
wjrosa: That's a good point. I tend to worry more about the performance of this change. I prefer to keep it this way, and we could bring this up for discussion later if that's OK.

malithsen (approved):
Thanks Wesley

This looks good to me

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -679,7 +681,7 @@ public function delete_renewal_meta( $renewal_order ) {
 	 */
 	public function update_failing_payment_method( $subscription, $renewal_order ) {
 		$subscription->update_meta_data( '_stripe_customer_id', $renewal_order->get_meta( '_stripe_customer_id', true ) );
-		$subscription->update_meta_data( '_stripe_source_id', $renewal_order->get_meta( '_stripe_source_id', true ) );
+		$subscription->update_meta_data( '_stripe_source_id', WC_Stripe_Order_Helper::get_instance()->get_stripe_source( $renewal_order ) );
```
annemirasol, line 684: I see there's an `update_stripe_source()` method in the order helper. Why are we not using that here?

annemirasol (approved):
LGTM! Smoke tested on simple purchases, subscriptions and renewals.

Left a question above.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: I just wanted to keep subscriptions out of this change. I left all the meta update/get methods from subscriptions unchanged. The reasons are to reduce the effort required to review this PR and leave room for us to decide later whether we want to create a subscription-specific class (if that makes sense at all).

PR #4696, opened 2025-09-24
Enable settings synchronization in more situations

daledupreez:

This PR updates our logic for settings synchronization to enable the feature in more situations. Prior to this change, we only enabled settings synchronization in situations where the returned payment method configurations (PMCs) from the `/v1/payment_method_configurations` API included a configuration that was linked to a WooCommerce PMC via the `parent` property. If no such PMC was returned, we then disabled settings synchronization.

With this change, we are now enabling settings synchronization for some additional situations. This PR updates our algorithm as follows:
 * Check for a PMC that is linked to a WooCommerce parent PMC.
   - If we find a WooCommerce child PMC, use that PMC. [done]
 * Filter out all PMCs that are inactive (i.e. `active === false`), or have a non-empty `parent` value, i.e. those that are linked to other Stripe Connect platforms.
   - If we have no remaining PMCs, disable settings synchronization. [done]
 * Check the `woocommerce_stripe_pmc_fallback_id` option.
   - If the option has a non-empty value, and we have a matching PMC ID in the list, use that PMC. [done]
   - If we didn't find the corresponding PMC in the list, clear the option.
 * If we only have one remaining PMC, use that PMC and update the `woocommerce_stripe_pmc_fallback_id` option. [done]
 * Identify all remaining PMCs that have the `default` field set to `true`.
   - If we have only one such PMC, use that PMC and update the `woocommerce_stripe_pmc_fallback_id` option.

Note that the use of the option for the fallback PMC is needed to prevent situations where we trigger an unexpected change to the PMCs returned from the API, such as adding a new PMC in the Stripe dashboard. We want merchants to keep the same configuration in use as much as possible.

## Testing instructions

The unit tests try to cover a _lot_ of possible situations! But it's also worth noting that I haven't looked into additional ways of setting up a site such that we would trigger these edge cases. I am 100% open to suggestions on how to mimic the various cases, or even a subset of them.
 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -129,13 +129,150 @@ private static function get_payment_method_configuration_from_stripe() {
...
+	 * @return array {
+	 *     @type object|null $pmc    The fallback Payment Method Configuration.
+	 *     @type string      $reason The reason for using the fallback Payment Method Configuration.
+	 * }
+	 */
+	private static function get_fallback_payment_method_configuration( array $payment_method_configurations ): array {
+		$active_non_child_payment_method_configurations = array_filter(
+			$payment_method_configurations,
+			function ( $configuration ) {
+				if ( $configuration->parent ?? null ) {
```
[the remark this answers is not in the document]
daledupreez: The null coalesce is here to prevent warnings when `parent` is undefined. No plans to change.

tests/phpunit/WC_Stripe_Payment_Method_Configurations_Test.php
```diff
@@ -11,6 +11,87 @@
...
+		'livemode' => true,
+		'active'   => true,
+		'default'  => false,
+	];
+
+	private const MOCK_TEST_NON_CHILD_DEFAULT_PMC = [
+		'id'       => 'pmc_id_test_non_child_default',
+		'parent'   => null,
+		'name'     => 'Test Non Child Default PMC',
+		'livemode' => true,
```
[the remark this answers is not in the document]
daledupreez: Good catch! Fixed in e26bd7c2359ec0ea0076860b7195b4a379484f46.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -129,13 +129,150 @@ private static function get_payment_method_configuration_from_stripe() {
 			// The API returns data for the corresponding mode of the api keys used, so we'll get either test or live PMCs, but never both.
 			if ( $configuration->parent && ( self::LIVE_MODE_CONFIGURATION_PARENT_ID === $configuration->parent || self::TEST_MODE_CONFIGURATION_PARENT_ID === $configuration->parent ) ) {
```
malithsen, line 133: This is unrelated to the changes made in this PR. Right now, we don't really care if the PMC is active or not as long as it is connected to a WC PMC. Is that intentional? What would happen if this PMC is not active?

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -129,13 +129,150 @@ private static function get_payment_method_configuration_from_stripe() {
...
+			WC_Stripe_Logger::debug(
+				'Updating Payment Method Configuration fallback',
+				[
+					'pmc_id'   => $fallback_pmc->id,
+					'reason'   => $fallback_reason,
+					'name'     => $fallback_pmc->name ?? null,
+					'livemode' => $fallback_pmc->livemode ?? null,
+				]
+			);
+			update_option( 'woocommerce_stripe_pmc_fallback_id', $fallback_pmc->id );
```
malithsen: Should we store this under separate keys for live and test mode? Although, given this is a last-resort path, the value added might be low relative to the complexity introduced.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -129,13 +129,150 @@ private static function get_payment_method_configuration_from_stripe() {
...
+			}
+		);
+
+		$fallback_pmc_id = get_option( 'woocommerce_stripe_pmc_fallback_id' );
+		$fallback_pmc    = null;
+
+		if ( [] === $active_non_child_payment_method_configurations ) {
+			if ( $fallback_pmc_id ) {
+				WC_Stripe_Logger::debug(
+					'No Payment Method Configurations returned from Stripe; deleting fallback ID option',
```
malithsen: nit: This could be somewhat misleading when we see it in logs with no other context.

```suggestion
					'No eligible Payment Method Configurations returned from Stripe; deleting fallback ID option',
```

malithsen (approved):
Code makes sense and looks good overall. I had a couple of questions and a minor, non-blocking suggestion.

includes/class-wc-stripe-payment-method-configurations.php
daledupreez: Good question. I _think_ we should limit these checks to active PMCs as a baseline. I can tackle that in a follow-up PR.

includes/class-wc-stripe-payment-method-configurations.php
daledupreez: Good call. Fixed in 24a7dcbac26ca20df4866b8e1a6b95c6f4e2be60.

includes/class-wc-stripe-payment-method-configurations.php
daledupreez: I've split the option keys in b54d1d421a3bc72c3d72ce288ccc09ef9d6fc02a. I think the use of dedicated options will be more robust in the long term, as the intent and expectations are explicit.

malithsen (approved):
The delta between the latest changes and the last review looks good to me. Thanks for the updates!

diegocurbelo (approved):
The code looks good, and tests well!

[nit] I think that in a future PR we could add a mark to the sync status (or maybe show `Limited` in yellow instead):
Image

PR #4705, opened 2025-09-29
Handling more Stripe metas with the order helper class

wjrosa:
Base PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4613
P2 peNR48-1Jc-p2

As part of the Order Helper class project, I am including some additional meta keys to be handled by the new class and implementing the new methods across our codebase.

The metas handled here are: `_stripe_customer_id`, `_stripe_card_id`, `_stripe_upe_payment_type`, `_stripe_upe_waiting_for_redirect`, and `_stripe_upe_redirect_processed`.

## Testing instructions

- Code review
- Check if the tests are still passing
- Perform some basic smoke testing to confirm no regression was introduced

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-order-helper.php
```diff
@@ -440,6 +475,202 @@ public function update_stripe_setup_intent( ?WC_Order $order = null, string $int
...
+	 * @param WC_Order|null $order
+	 * @return false|void
+	 */
+	public function delete_stripe_customer( ?WC_Order $order = null ) {
+		if ( is_null( $order ) ) {
+			return false;
+		}
+
+		$order->delete_meta_data( self::META_STRIPE_CUSTOMER_ID );
+	}
```
daledupreez: From a naming perspective, I think it would be clearer to call these functions `*_stripe_customer_id(`, as we are manipulating the customer ID. A Stripe customer object is distinct from its ID, and I think it would be clearer if our method naming kept that distinction.
```suggestion
	 * Gets the Stripe customer ID for an order.
	 *
	 * @since 10.0.0
	 *
	 * @param WC_Order|null $order
	 * @return false|string|null
	 */
	public function get_stripe_customer_id( ?WC_Order $order = null ) {
		if ( is_null( $order ) ) {
			return false;
		}

		return $order->get_meta( self::META_STRIPE_CUSTOMER_ID, true );
	}

	/**
	 * Updates the Stripe customer ID for an order.
	 *
	 * @since 10.0.0
	 *
	 * @param WC_Order|null $order
	 * @param string $customer_id
	 * @return false|void
	 */
	public function update_stripe_customer_id( ?WC_Order $order = null, string $customer_id = '' ) {
		if ( is_null( $order ) ) {
			return false;
		}

		$order->update_meta_data( self::META_STRIPE_CUSTOMER_ID, $customer_id );
	}

	/**
	 * Deletes the Stripe customer ID for an order.
	 *
	 * @since 10.0.0
	 *
	 * @param WC_Order|null $order
	 * @return false|void
	 */
	public function delete_stripe_customer_id( ?WC_Order $order = null ) {
		if ( is_null( $order ) ) {
			return false;
		}

		$order->delete_meta_data( self::META_STRIPE_CUSTOMER_ID );
	}
```

includes/class-wc-stripe-order-helper.php
```diff
@@ -440,6 +475,202 @@ public function update_stripe_setup_intent( ?WC_Order $order = null, string $int
...
+	/**
+	 * Gets the Stripe customer for order.
+	 *
+	 * @since 10.0.0
+	 *
+	 * @param WC_Order|null $order
+	 * @return false|string|null
+	 */
+	public function get_stripe_customer( ?WC_Order $order = null ) {
+		if ( is_null( $order ) ) {
```
daledupreez: Given the type for the argument, I think we can simplify these null checks to use a direct equality check rather than calling `is_null()`.
```suggestion
		if ( null === $order ) {
```

includes/class-wc-stripe-order-helper.php
```diff
@@ -440,6 +475,202 @@ public function update_stripe_setup_intent( ?WC_Order $order = null, string $int
...
+
+	/**
+	 * Gets the Stripe card for order.
+	 *
+	 * @since 10.0.0
+	 *
+	 * @param WC_Order|null $order
+	 * @return false|string|null
+	 */
+	public function get_stripe_card( ?WC_Order $order = null ) {
```
daledupreez: I think we should also use `*_stripe_card_id(` for these to be super-clear on what we are returning.

includes/class-wc-stripe-order-helper.php
```diff
@@ -440,6 +475,202 @@ public function update_stripe_setup_intent( ?WC_Order $order = null, string $int
...
+	/**
+	 * Updates the Stripe UPE payment type for order.
+	 *
+	 * @since 10.0.0
+	 *
+	 * @param WC_Order|null $order
+	 * @param string $payment_type
+	 * @return false|void
+	 */
+	public function update_stripe_upe_payment_type( ?WC_Order $order = null, string $payment_type = '' ) {
```
daledupreez, line 532: Question on the types: might we expect instances of `WC_Order_Refund` to be supplied to any of these helper methods? If so, we might want to loosen the types to accept `?WC_Abstract_Order`.

includes/class-wc-stripe-order-helper.php
```diff
@@ -440,6 +475,202 @@ public function update_stripe_setup_intent( ?WC_Order $order = null, string $int
...
+	 * @param bool $redirect_processed
+	 * @return false|void
+	 */
+	public function update_stripe_upe_redirect_processed( ?WC_Order $order = null, bool $redirect_processed = false ) {
+		if ( is_null( $order ) ) {
+			return false;
+		}
+
+		$order->update_meta_data( self::META_STRIPE_UPE_REDIRECT_PROCESSED, $redirect_processed );
+	}
```
daledupreez, line 604: Not blocking, but I am wondering whether it might be worth creating some internal/private helper functions so we have less duplicated/boilerplate code. Something like the following is what I was thinking, but I realise that it adds one more layer of indirection.
```suggestion
	public function update_stripe_upe_redirect_processed( ?WC_Order $order = null, bool $redirect_processed = false ) {
		return $this->_update_order_meta( $order, self::META_STRIPE_UPE_REDIRECT_PROCESSED, $redirect_processed );
	}

	private function _update_order_meta( ?WC_Order $order,  string $meta_key, $new_value ) {
		if ( null === $order ) {
			return false;
		}
		
		$order->update_meta_data( $meta_key, $new_value );
	}
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -680,8 +680,9 @@ public function delete_renewal_meta( $renewal_order ) {
 	 * @return void
 	 */
 	public function update_failing_payment_method( $subscription, $renewal_order ) {
-		$subscription->update_meta_data( '_stripe_customer_id', $renewal_order->get_meta( '_stripe_customer_id', true ) );
-		$subscription->update_meta_data( '_stripe_source_id', WC_Stripe_Order_Helper::get_instance()->get_stripe_source( $renewal_order ) );
+		$order_helper = WC_Stripe_Order_Helper::get_instance();
+		$subscription->update_meta_data( '_stripe_customer_id', $order_helper->get_stripe_customer( $renewal_order ) );
+		$subscription->update_meta_data( '_stripe_source_id', $order_helper->get_stripe_source( $renewal_order ) );
```
daledupreez: Not blocking: should we use a constant for these strings? Or are you planning to tackle them in a future change?

daledupreez (commented):
I have some feedback, based mostly on the naming, but there are one or two other areas that may drive changes. I'll dive into testing after confirming we don't plan to make any changes.

includes/class-wc-stripe-order-helper.php
wjrosa: Makes sense 👍 . I will also change the previous methods to follow the same pattern when needed.

includes/class-wc-stripe-order-helper.php
wjrosa: Updated all checks in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4705/commits/10125f0c244606e5201c5c9546a8a8c280e8cd70

includes/class-wc-stripe-order-helper.php
wjrosa: Renamed those in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4705/commits/c77fab6d53c3597ec83da517266f6a2af93345a4

includes/class-wc-stripe-order-helper.php
wjrosa: I actually like the idea! Implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4705/commits/aff3db9c4f49931d8c6c2a7d5c4b578a6b47b177

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: My idea was to keep `WC_Subscription` away from the changes for now, until we decide if we need another helper for them or implement the same helper. Still, that would happen in another PR.

includes/class-wc-stripe-order-helper.php
wjrosa: Looking at the code, I don't think so. I don't see any references to `WC_Order_Refund` in the flows where this method is called. But happy to make this change if you think it is safer 👍

includes/class-wc-stripe-order-helper.php
wjrosa: PR to rename the merged methods here: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4716

includes/class-wc-stripe-order-helper.php
daledupreez: From my side, I was mostly flagging it as a possible concern. I don't know the full codebase well enough to know what objects we expect, so having the strict type _might_ be a subtle compatibility issue. Using `WC_Abstract_Order` would be less likely to break, even if we never see a refund or similar, while the slightly more restrictive type would throw an exception for an unexpected object.

Maybe this is something we should tackle in a separate, dedicated follow-up if we decide to make any changes.

includes/class-wc-stripe-order-helper.php
```diff
@@ -433,11 +444,187 @@ public function get_stripe_setup_intent( ?WC_Order $order = null ) {
...
+	 * @param WC_Order|null $order
+	 * @return false|void
+	 */
+	public function delete_stripe_card_id( ?WC_Order $order = null ) {
+		if ( null === $order ) {
+			return false;
+		}
+
+		$order->delete_meta_data( self::META_STRIPE_CARD_ID );
+	}
```
daledupreez, line 505: Not blocking at all, but we can also consider implementing central `delete_order_meta()` and `get_order_meta()` helper functions much like the `update_order_meta()` helper you implemented. That may be worth doing in one or more follow-up PRs as it should be a relatively quick to review if we're only refactoring the implementation and making no other changes.

includes/class-wc-stripe-order-helper.php
```diff
@@ -480,6 +667,7 @@ public function add_payment_intent_to_order( string $payment_intent_id, WC_Order
 	 * @return void
 	 */
 	public function set_payment_awaiting_action( WC_Order $order, bool $save = true ): void {
+		$this->update_order_meta( $order, self::META_STRIPE_PAYMENT_AWAITING_ACTION, true );
 		$order->update_meta_data( self::META_STRIPE_PAYMENT_AWAITING_ACTION, wc_bool_to_string( true ) );
```
daledupreez: Should we make only one update here? It looks like we're performing two redundant upates.
```suggestion
		$this->update_order_meta( $order, self::META_STRIPE_PAYMENT_AWAITING_ACTION, wc_bool_to_string( true ) );
```

includes/class-wc-stripe-order-helper.php
```diff
@@ -829,4 +1017,20 @@ protected function is_order_refund_locked( WC_Order $order ): bool {
 
 		return false;
 	}
+
+	/**
+	 * Updates order meta data.
```
daledupreez: Not blocking: It may be worth being explicit as to why this is a helper function.
```suggestion
	 * Helper function to update order meta data. The goal of the function is to reduce boilerplate in the helper due to `null` checks everywhere.
```

daledupreez (approved):
The changes LGTM, and the tests are clean, so I think we should 🚢 .

malithsen (approved):
These updates look good to me, and smoke tests didn't reveal any issues.

includes/class-wc-stripe-order-helper.php
wjrosa: Good idea. I went ahead and already implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4705/commits/86e9d72322200650a1173afe5b32b5cdf3e14939

PR #4710, opened 2025-10-02
Hardcoding ECE feature flag to be always enabled

wjrosa:
Fixes STRIPE-733
See https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4687#discussion_r2399736177

In this PR, I am hardcoding the Express Checkout feature flag to be always enabled. This will force ECE for all merchants, even those that, for some reason, changed the default flag value, making it safe to remove all code related to Payment Request Buttons.

## Testing instructions

Code review is enough.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.0.0 - xxxx-xx-xx =
+* Update - Hardcodes the Express Checkout feature flag to be always enabled
 * Dev - Expands the Stripe Order Helper class to handle source ID, refund ID, intent ID, and setup intent ID metas
 * Update - Removing the `wc_stripe_is_upe_checkout_enabled` filter, as Legacy Checkout is no longer supported
 * Dev - Upgrades `jest` to version 29.7.0, `@wordpress/scripts` to 26.19.0, and adds `axios`(version 1.12.2) to the JavaScript development dependencies
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 10.0.0 - xxxx-xx-xx =
-* Update - Hardcodes the Express Checkout feature flag to be always enabled
+* Update - Disable Payment Request Buttons and ensure Express Checkout is used when express checkout buttons are enabled
 * Dev - Expands the Stripe Order Helper class to handle source ID, refund ID, intent ID, and setup intent ID metas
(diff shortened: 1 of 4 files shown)


Mayisha:
@wjrosa this feature flag check becomes unnecessary as we are removing the PRB frontend and backend code/files completely. IMO we can remove the `is_stripe_ece_enabled` function and the corresponding calls to it. With PRB gone, ECE would be the only available option so we don't need the feature flag check anymore.

wjrosa:
I agree. However, that was to address @daledupreez's concerns in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4687#discussion_r2399736177. It is still possible for a merchant to manually change the feature flag value, as far as I am aware. Let me know if that's a wrong assumption (I hope so).

Mayisha:
> I agree. However, that was to address @daledupreez's concerns in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4687#discussion_r2399736177. It is still possible for a merchant to manually change the feature flag value, as far as I am aware. Let me know if that's a wrong assumption (I hope so).

cc @daledupreez for clarification. I think the feature flag should be deleted. Also ECE has been default on for UPE since the beginning of this year

daledupreez:
I saw at least one merchant with the express checkout feature flag disabled during my support rotation in September - they were still using PRBs. 😬 I didn't know that was still possible, but it clearly was.

With that in mind, I think it's probably safest to stick with @wjrosa's approach for the short term.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.0.0 - xxxx-xx-xx =
+* Update - Hardcodes the Express Checkout feature flag to be always enabled
```
daledupreez: I think we should rephrase this to make it clearer what that means for merchants. Maybe something like the following:
```suggestion
* Update - Disable Payment Request Buttons and ensure Express Checkout is used when express checkout buttons are enabled
```

daledupreez (approved):
Other than the changelog entry wording, this LGTM.

Mayisha (approved):
Code looks good to me. 👍 

However, I still think removing the `is_stripe_ece_enabled` completey vs returning `true` from this function would behave the same if we remove all the PRB related files (as there is no fallback and ECE is the only option available) 🤔

PR #4712, opened 2025-10-03
Return false for undefined methods in __call() instead of throwing error

malithsen:
This PR fixes a fatal error that occurs when third-party plugins use `is_callable()` to check for the existence of methods on Stripe UPE payment method objects.

Slack: p1759503846980199-slack-C055WHLA98D

The `WC_Stripe_UPE_Payment_Method` class implements the `__call()` magic method to forward method calls to the main UPE gateway instance. However, when a class has `__call()`, PHP's `is_callable()` returns `true` for any method name, even if the method doesn't exist. But, calling the method then throws a fatal error because the method doesn't exist in either the payment method class or the main gateway.

In this PR, instead of throwing an error for undefined methods, the `__call()` method now returns `false`. 

### How can this code break?
- If our code relies on catching the thrown error from `__call()` for undefined methods, these changes could potentially break it. But from a cursory look, we don't seem to do that, and we'd want to refactor those paths anyway, as it's sort of an anti-pattern.
 

## Testing instructions

**Setup**

1. Install a code snippets plugin (eg: https://wordpress.org/plugins/code-snippets/)
2. Add the following snippet
```
$paymentMethods = WC()->payment_gateways()->get_available_payment_gateways();

foreach ($paymentMethods as $paymentMethod) {

	$callable = is_callable([$paymentMethod, 'getWalletButtonEnabled']) && $paymentMethod->getWalletButtonEnabled();
}
```

**Test dev env**
1. Add `define( 'WP_DEBUG', true );` to your `wp-config.php`
2. Try to save the snippet, it should fail. In the site logs, there should be a fatal error.
3. Set `WP_DEBUG` to `false` and set `define('WP_ENVIRONMENT_TYPE', 'staging');` and repeat step 2. It should fail again with the same fatal error.

**Test prod**
1. Remove all instances of `WP_DEBUG` and `WP_ENVIRONMENT_TYPE` in `wp-config.php` from the previous tests.
2. Saving the snippet should be successful.
3. In `woocommerce-gateway-stripe` log, an entry like the following should be logged

CleanShot 2025-10-06 at 15 33 58@2x
 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -154,10 +154,9 @@ public function __call( $method, $arguments ) {
 
 		if ( in_array( $method, get_class_methods( $upe_gateway_instance ) ) ) {
 			return call_user_func_array( [ $upe_gateway_instance, $method ], $arguments );
-		} else {
-			$message = method_exists( $upe_gateway_instance, $method ) ? 'Call to private method ' : 'Call to undefined method ';
-			throw new \Error( $message . get_class( $this ) . '::' . $method );
 		}
+
+		return false;
```
diegocurbelo, line 170: I think we should log an error if the method can't be called (either because it's private or does not exist).

And, I think we should still throw an error if the call is to a private method.

diegocurbelo (approved):
Thanks for quickly investigating and fixing this @malithsen!

I personally don't like the "silent failure" approach. If someone uses an invalid method, they might not even realize it; for example, `is_enabled_at_checkout()` is a valid method, and false is a valid return value for it; calling `enabled_at_checkout()` (no `is_`) by mistake is going to be hard to debug 🤷🏼

I'm okay with this as a temporary solution for the fatal errors, as once we fully deprecate the legacy gateway, we'll have only one gateway, and the `__call` method would be redundant.

malithsen:
>  If someone uses an invalid method, they might not even realize it; for example, is_enabled_at_checkout() is a valid method, and false is a valid return value for it; calling enabled_at_checkout() (no is_) by mistake is going to be hard to debug 🤷🏼

@diegocurbelo 
Ah that's a good point. What do you think about using WP_DEBUG to determine the failure mode. In prod `return false` and fail silently. In dev (if WP_DEBUG present), `throw Error` like we used to.

daledupreez:
> What do you think about using WP_DEBUG to determine the failure mode. In prod return false and fail silently. In dev (if WP_DEBUG present), throw Error like we used to.

That seems like a really good idea. Might we want to check `wp_get_environment_type()` as well for more situations where we might want to be more aggressive about logging, e.g. for staging environments.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
malithsen: Changed the approach in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4712/commits/57a780899245e24aec305073feac5fe764b09cf6

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -154,10 +154,20 @@ public function __call( $method, $arguments ) {
...
+		// In production, fail gracefully to prevent third-party plugin compatibility issues.
+		$is_dev_environment = ( defined( 'WP_DEBUG' ) && WP_DEBUG )
+			|| in_array( wp_get_environment_type(), [ 'development', 'staging', 'local' ], true );
+
+		if ( $is_dev_environment ) {
 			$message = method_exists( $upe_gateway_instance, $method ) ? 'Call to private method ' : 'Call to undefined method ';
 			throw new \Error( $message . get_class( $this ) . '::' . $method );
 		}
+
+		WC_Stripe_Logger::warning( 'Call to undefined method ' . get_class( $this ) . '::' . $method );
```
daledupreez: If logging is disabled for the Stripe plugin, we won't record this entry. I think we should upgrade it to an error so we always log the entry.
```suggestion
		WC_Stripe_Logger::error( 'Call to undefined method ' . get_class( $this ) . '::' . $method );
```

daledupreez (approved):
The change is LGTM and tests as described. I have one comment about switching from a warning to an error, but that's not blocking.

diegocurbelo (approved):
Thanks for the changes... it tests as expected!

LGTM

PR #4717, opened 2025-10-08
Track the reconnect notice and button clicks

wjrosa:
Fixes STRIPE-740

To complement https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4669, this PR will add tracking for the "Reconnect to Stripe" button clicks (for both the admin notice and the main button in the settings screen).

## Testing instructions

- Checkout and build this branch on your test environment (`add/tracking-account-reconnect-notice-click`)
- Connect your Stripe account
- Set up your store to behave as if it did not connect via OAuth by removing the connection_type (or test_connection_type attr from the settings:
```
wp option patch delete woocommerce_stripe_settings test_connection_type
```
- Click the "Reconnect to Stripe" button (either from the admin notice or the main button in the Stripe settings screen)
- In about 5 minutes, check MC Tracks Live events page for the presence of a `wcstripe_reconnect_button_click` event

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


malithsen:
@wjrosa ~I might need some additional help testing this. I tried deleting the `test_connection_type` key/value pair from the settings option. That forced me to reconnect via the initial connection screen (No "Reconnect to Stripe" button)~

~I also tried hard-coding `oauthConnected = false` but I couldn't get `account-details-section.js` component to get called at all. I might be looking at the wrong views. Do you mind sharing a screenshot?~


EDIT: 
nvm, I was making changes on my local site while looking at a separate JN site 🤦

diegocurbelo (approved):
This is working as indicated, my only nitpick is that we are not actually tracking the click in the admin notice button, but the page load it triggers (&highlight=account-details), so page refreshes will count towards it 🤷🏼 

Image

Also, maybe we should also track the `Re-authenticate` button to get the full picture:

Image

malithsen (approved):
Code looks good, agreeing with Diego's feedback.

wjrosa:
Thanks for the review, folks!

> This is working as indicated, my only nitpick is that we are not actually tracking the click in the admin notice button, but the page load it triggers (&highlight=account-details), so page refreshes will count towards it 🤷🏼

That's a good point. I refactored the code in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4717/commits/f348c3552b6b2be33783bf91e84d7774bbf78c24 to actually track the click instead of the page scroll. I had to use raw JS since I could not import the existing Tracks component there.

> Also, maybe we should also track the Re-authenticate button to get the full picture:

Added tracking for that in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4717/commits/79025c6c0ba29a31ac5ade397409e57eb8d81029

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -432,10 +433,13 @@ public function stripe_check_environment() {
...
-					$link    = esc_url( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings&highlight=account-details' ) );
-					$actions = [
-						'<a href="' . $link . '" style="display: block; padding: 10px; color: #007cba; border: #007cba solid 1px; float: left; text-decoration: none; margin-bottom: 12px; border-radius: 2px;">'
+					$message  = __( 'Please reconnect to continue using Stripe and avoid disruptions on your store.', 'woocommerce-gateway-stripe' );
+					$style    = 'display: block; padding: 10px; color: #007cba; border: #007cba solid 1px; float: left; text-decoration: none; margin-bottom: 12px; border-radius: 2px;';
+					$onclick  = 'window.wcTracks = window.wcTracks || {}; window.wcTracks.recordEvent = window.wcTracks.recordEvent || function() { };';
+					$onclick .= 'window.wcTracks.recordEvent( \'wcstripe_reconnect_button_click\', { source: \'admin_notice\', mode: ' . ( $testmode ? '\'test\'' : '\'live\'' ) . ' } );';
+					$onclick .= 'window.location = \'admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings&highlight=account-details\'; return false;';
```
malithsen: Not a fan of this series of inline JS, but I also don't see a significantly better approach compared to this.

malithsen (approved):
Changes look good to me

includes/admin/class-wc-stripe-admin-notices.php
daledupreez: Can we not add a JS snippet that wraps all of this logic? It feels like manually generating the onclick handling here is going to be really messy to get right in the long run. Also, is it possible for us to keep the `href` and simply let the onclick trigger the tracking?

malithsen: > This is working as indicated, my only nitpick is that we are not actually tracking the click in the admin notice button, but the page load it triggers (&highlight=account-details), so page refreshes will count towards it 🤷🏼 
~ Diego

Another option to address the original issue that this in-line JS is trying to solve is that we pass in a URL fragment here, and in the frontend record the event if the fragment is present. Once the event is recorded, remove the fragment so that page refreshes don't result in duplicate events.

wjrosa: Good point, folks. I refactored it again in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4717/commits/b18cc9c83b7931778e6c1f2d99d920d85130503d, using Malith's suggestion above (thanks!)

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -432,10 +432,11 @@ public function stripe_check_environment() {
 
 				$oauth_required = $needs_live_oauth || $needs_test_oauth;
 				if ( $oauth_required ) {
-					$message = __( 'Please reconnect to continue using Stripe and avoid disruptions on your store.', 'woocommerce-gateway-stripe' );
-					$link    = esc_url( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings&highlight=account-details' ) );
-					$actions = [
-						'<a href="' . $link . '" style="display: block; padding: 10px; color: #007cba; border: #007cba solid 1px; float: left; text-decoration: none; margin-bottom: 12px; border-radius: 2px;">'
+					$message  = __( 'Please reconnect to continue using Stripe and avoid disruptions on your store.', 'woocommerce-gateway-stripe' );
+					$link     = esc_url( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings&highlight=account-details' ) );
```
diegocurbelo: The code in `client/settings/payment-settings/account-details-section.js` is expecting a `track` url param:

```suggestion
					$link     = esc_url( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings&highlight=account-details&track=true' ) );
```

client/settings/payment-settings/account-details-section.js
```diff
@@ -113,7 +130,19 @@ const AccountDetailsSection = ( { setModalType, setKeepModalContent } ) => {
...
+
+	useEffect( () => {
+		const { track } = getQuery();
+		if ( track ) {
+			trackReconnectEvent( 'admin_notice', isTestMode ? 'test' : 'live' );
+
+			// Remove track param from URL.
+			const url = new URL( window.location.href );
+			url.searchParams.delete( 'track' );
+			window.location.replace( url.href );
```
diegocurbelo: `window.location.replace( url.href );` causes a page reload, so clicking the`Reconnect to Stripe` button in the notice causes two consecutive reloads.

We should use `replaceState`:
```suggestion
			history.replaceState( null, '', url.href );
```

diegocurbelo (approved):
Thanks for the fixes @wjrosa. It's working as expected now!

daledupreez (commented):
I removed the reconnection logic yesterday in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4730. I think we need to take a step back to assess how and when we show the reconnection notices due to some of the issues we saw in the 10.0.0 release.

wjrosa:
I removed all code related to the admin notice yesterday. I kept only the code tracking the existing buttons and links. Still, do you want to put this on hold anyway?

client/settings/payment-settings/__tests__/account-details-section.test.js
```diff
@@ -287,5 +292,41 @@ describe( 'AccountDetailsSection', () => {
...
+					},
+				},
+			} );
+
+			render(
+				<AccountDetailsSection setModalType={ setModalTypeMock } />
+			);
+
+			const editKeysButton = screen.getByRole( 'button', {
+				name: /Reconnect to Stripe/i,
```
daledupreez: @wjrosa, I missed the removal of the other code yesterday, but I believe the JS tests are failing because we're no longer changing this button text after my PR. Apologies for the misunderstanding on my part!

wjrosa: No worries! That was the issue with the tests. Fixed them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4717/commits/b8f280584ca07c35892ae86b2b575002e0da0d3a

PR #4718, opened 2025-10-08
Deprecate and remove Legacy Checkout instances

wjrosa:
Fixes STRIPE-524

This PR removes some unnecessary instantiations of the legacy checkout, as UPE is now the only option available. I am also deprecating the main legacy gateway class to make it easier to remove other references to it later.

## Testing instructions

- Checkout to this branch on your test environment (`update/deprecate-and-remove-legacy-checkout-instances`)
- Rebuild your test store (or create a new one using something like JN)
- Install the extension
- Connect your Stripe account
- Confirm the Stripe settings page is working fine
- Check your settings database key and confirm `upe_checkout_experience_enabled` is set to `yes`

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe.php
```diff
@@ -817,17 +817,11 @@ public function register_routes() {
 	 * @return WC_Stripe_Payment_Gateway
 	 */
 	public function get_main_stripe_gateway() {
-		if ( ! is_null( $this->stripe_gateway ) ) {
+		if ( ! is_null( $this->stripe_gateway ) && is_a( $this->stripe_gateway, 'WC_Stripe_UPE_Payment_Gateway' ) ) {
```
diegocurbelo: In which case would `$this->stripe_gateway` be set, but not an instance of `WC_Stripe_UPE_Payment_Gateway`?

I feel we should log this happening as a `warning` (or maybe `error` 🤷🏼)

diegocurbelo (commented):
This looks good. I could not find any other instantiation of the Legacy gateway.

I think we should also update the test suite accordingly. It's still being used in these 4 test files:
- WC_Stripe_Payment_Gateway_Test.php
- WC_REST_Stripe_Settings_Controller_Test.php
- WC_Stripe_Settings_Controller_Test.php
- WC_Stripe_Express_Checkout_Helper_Test.php

includes/class-wc-stripe.php
wjrosa: Good point. But with the latest changes, this is not even needed. I reverted it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4718/commits/f3776ed76ff43166ae1d9cf0e6162f3702955615. I don't see how this condition would not be true anymore.

wjrosa:
Thanks for the review, Diego! I removed all references from the unit tests in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4718/commits/df553ae7863b77194c9259e096e6785222b42a18

includes/class-wc-stripe.php
```diff
@@ -821,13 +821,7 @@ public function get_main_stripe_gateway() {
...
-		if ( WC_Stripe_Feature_Flags::is_upe_preview_enabled() && WC_Stripe_Feature_Flags::is_upe_checkout_enabled() ) {
-			$this->stripe_gateway = new WC_Stripe_UPE_Payment_Gateway();
-
-			return $this->stripe_gateway;
-		}
-
-		$this->stripe_gateway = new WC_Gateway_Stripe();
+		$this->stripe_gateway = new WC_Stripe_UPE_Payment_Gateway();
 
 		return $this->stripe_gateway;
```
daledupreez: Not blocking, but it might be easier to flip the condition here to simplify the code.
```suggestion
		if ( ! $this->stripe_gateway ) {
			$this->stripe_gateway = new WC_Stripe_UPE_Payment_Gateway();
		}

		return $this->stripe_gateway;
```

tests/phpunit/Admin/WC_REST_Stripe_Settings_Controller_GB_Test.php
```diff
@@ -29,7 +26,7 @@ class WC_REST_Stripe_Settings_Controller_GB_Test extends WC_Mock_Stripe_API_Unit
 	/**
 	 * Gateway instance that the controller uses.
 	 *
-	 * @var WC_Gateway_Stripe
+	 * @var WC_Stripe_UPE_Payment_Gateway
```
daledupreez, line 29: Non-blocking style note: I prefer using the fully-qualified class name so it's easier to see exactly which class is in use, especially in diffs. I think we should discuss this as a team before making changes to any code, though.
```suggestion
	 * @var \WC_Stripe_UPE_Payment_Gateway
```

tests/phpunit/Admin/WC_REST_Stripe_Settings_Controller_GB_Test.php
```diff
@@ -60,7 +57,7 @@ public static function set_up_before_class() {
 		$upe_helper->enable_upe();
 		$upe_helper->reload_payment_gateways();
 
-		self::$gateway = WC()->payment_gateways()->payment_gateways()[ WC_Gateway_Stripe::ID ];
+		self::$gateway = WC()->payment_gateways()->payment_gateways()[ WC_Stripe_UPE_Payment_Gateway::ID ];
```
daledupreez, line 60: Not blocking: we still have plenty of references to `WC_Gateway_Stripe::ID` in the main code. But we can tackle them in other PRs.

daledupreez (approved):
Changes LGTM! I have some feedback/commentary, but none of it requires immediate action.

includes/class-wc-stripe.php
wjrosa: I agree. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4718/commits/a1c55f11a25b305c1373f0244640c71c2ef6af3a

tests/phpunit/Admin/WC_REST_Stripe_Settings_Controller_GB_Test.php
wjrosa: I went ahead and replaced all of them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4718/commits/fa8b22910181b7b8b8d3a40cf2974d36c9fc64c3

Issue #4713, opened 2025-10-04
Per-site payment method settings affecting all domains


bayramuygr:
Per-site method visibility not isolated.

Expected: Ability to hide/show methods per site.

Example:
Hide iDEAL on the .at site (Austria).
Hide EPS on the .nl site (Netherlands).

Actual: If we disable EPS on the .nl site, it also becomes disabled on the .at site, and vice versa for iDEAL.

PS: This working correctly on v9.3.2 (we verified twice and still use that version on a few sites).

Environment: seperate WP setups, not multisite.

daledupreez:
Thanks for reporting this, @bayramuygr.

The main changes involved here started in 9.5, where we now store the payment configuration in Stripe and synchronise it to the site so we can make Optimized Checkout available. However, this means that we are using a single payment method configuration across your sites connected to the same Stripe account, which is clearly not ideal for your use case. We've been thinking about ways of allowing for explicit payment method selection in the UI, and this would be a key use case for that.

You should be able to temporarily force the configuration to run locally by setting the `pmc_enabled` flag within the plugin settings to `'no'`, but that value would likely be removed during your next upgrade, so it would be a short-term fix at best.

daledupreez:
@bayramuygr, as a quick update, 10.0 includes a new `wc_stripe_preselect_payment_method_configuration` filter to make it possible for different sets of payment methods to be supported across different sites. That filter was added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4720.

The main complication here is that we don't have a good way to manage multiple payment method configurations within our admin UI, so all management will need to happen via the Stripe dashboard, which doesn't clearly document which payment methods are (or are not) supported by the plugin. However, if this is a critical business need, it should be possible to offer different sets of payment methods on different sites.

In the longer run, we plan to offer richer management of payment method configurations within the settings UI, but it may be some time before we will be able to build that.

daledupreez:
We're still trying to work out how best to handle and manage multiple payment method configurations for the same site. It may take us some time to get that mapped out from a UX perspective, but we do have the `wc_stripe_preselect_payment_method_configuration` filter available as a stopgap in the meantime.

PR #4720, opened 2025-10-09
Add filter to allow for manual payment method configuration (PMC) selection

daledupreez:
Towards STRIPE-735
Towards #4713

While I've kicked off a broader discussion for how we might want to allow for PMC selection (see pc4etw-1WV-p2 ), this PR tackles a much smaller piece of the puzzle: for more advanced setups, can we allow merchants to specify which PMC they want to use? The PR explores doing that by implementing a new `wc_stripe_preselect_payment_method_configuration` filter. If merchants hook into the filter and specify a valid PMC identifier of the form `pmc_*`, we will use that configuration by default, including caching the PMC. If we fail to fetch the PMC, we log an error and continue with the default logic.

### How might this break?

Merchants or developers could specify a PMC ID that doesn't exist, or is not valid in the current live/test environment. In those situations, we'll fall back on the default logic, which could be unexpected for the merchant.

Merchants may also create a PMC that adds payment methods we don't support in the plugin, especially if they're working within the Stripe dashboard to create a new PMC from scratch. However, that's already a possible issue with merchants making changes from the Stripe dashboard.

## Testing instructions

* Run the code from this branch on a test site
* Connect your site to Stripe using an account that you can use to access the Stripe dashboard -- a throwaway test account won't work that well for this use case, but you can use test or live mode for the connection.
* Ensure that logging is enabled for the Stripe plugin
* Make some changes to the enabled payment methods within the Stripe settings page -- make a note of which payment methods are enabled and disabled
* Navigate directly to the Stripe dashboard and navigate to _Settings_ -> _Payments_ -> _Payment methods_. Make sure you're in test/live mode to match your site's mode.
* Create a new payment method configuration within the dashboard, and enable some payment methods supported by the plugin, or even some we don't support if you want
* Copy the identifier for the PMC
* Add the following code snippet to your site to force selection of your new PMC with the `$is_test_mode` check set up to match your site configuration:
```php
function test_select_custom_stripe_pmc( $pmc_id, $is_test_mode ) {
	if ( $is_test_mode ) {
		return 'pmc_YOUR_PMC';
	}
	return $pmc_id;
}
add_filter( 'wc_stripe_preselect_payment_method_configuration', 'test_select_custom_stripe_pmc', 10, 2 );
```
* Reload the payment method settings page to force a reload of the payment configuration
* Verify that the PMC is your new PMC based on the payment methods
* Check the Stripe logs to verify that we're making an API call to `payment_method_configurations/pmc_YOUR_PMC` after the filter is added

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -120,7 +120,16 @@ class WC_Stripe_Payment_Method_Configurations {
 	 * @return object|null
 	 */
 	private static function get_payment_method_configuration_from_stripe() {
-		$preselected_pmc_id = apply_filters( 'wc_stripe_preselect_payment_method_configuration', null );
+		$is_test_mode       = WC_Stripe_Mode::is_test();
+		/**
+		 * Allows merchants to specify the ID of a Payment Method Configuration to use. This makes it possible for
+		 * merchants to create configurations for specific sites, e.g. when they operate sites in different countries
+		 * with different local payment methods.
+		 *
+		 * @param string|null $preselected_pmc_id The ID of the Payment Method Configuration to use. Null by default, but a string value may be returned.
+		 * @param bool        $is_test_mode       Whether the site is in test mode.
+		 */
+		$preselected_pmc_id = apply_filters( 'wc_stripe_preselect_payment_method_configuration', null, $is_test_mode );
 
 		if ( is_string( $preselected_pmc_id ) && str_starts_with( $preselected_pmc_id, 'pmc_' ) ) {
 			$configuration = WC_Stripe_API::retrieve( 'payment_method_configurations/' . $preselected_pmc_id );
@@ -149,7 +158,6 @@ class WC_Stripe_Payment_Method_Configurations {
 		$result         = WC_Stripe_API::get_instance()->get_payment_method_configurations();
 		$configurations = $result->data ?? [];
 
-		$is_test_mode     = WC_Stripe_Mode::is_test();
 		$fallback_pmc_key = $is_test_mode ? 'woocommerce_stripe_pmc_fallback_id_test' : 'woocommerce_stripe_pmc_fallback_id_live';
 
 		// When connecting to the WooCommerce Platform account a new payment method configuration is created for the merchant.
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -121,6 +121,7 @@ class WC_Stripe_Payment_Method_Configurations {
 	 */
 	private static function get_payment_method_configuration_from_stripe() {
 		$is_test_mode       = WC_Stripe_Mode::is_test();
+
 		/**
 		 * Allows merchants to specify the ID of a Payment Method Configuration to use. This makes it possible for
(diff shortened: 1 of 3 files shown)


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -120,10 +120,44 @@ private static function set_payment_method_configuration_cache( $configuration )
...
+				return $configuration;
+			}
+			// If the preselected Payment Method Configuration is not found, we continue with the default logic below.
+		}
+
 		$result         = WC_Stripe_API::get_instance()->get_payment_method_configurations();
 		$configurations = $result->data ?? [];
 
-		$is_test_mode     = WC_Stripe_Mode::is_test();
 		$fallback_pmc_key = $is_test_mode ? 'woocommerce_stripe_pmc_fallback_id_test' : 'woocommerce_stripe_pmc_fallback_id_live';
```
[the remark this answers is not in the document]
daledupreez: It is also used just after the location where it was added, so no plans to change this.

diegocurbelo (approved):
The change looks good and it tests well!.

It correctly uses the default one if the preselected PMC is invalid (or can't be ready by the current API key)

Image

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -120,10 +120,45 @@ private static function set_payment_method_configuration_cache( $configuration )
...
+		 * Allows merchants to specify the ID of a Payment Method Configuration to use. This makes it possible for
+		 * merchants to create configurations for specific sites, e.g. when they operate sites in different countries
+		 * with different local payment methods.
+		 *
+		 * @param string|null $preselected_pmc_id The ID of the Payment Method Configuration to use. Null by default, but a string value may be returned.
+		 * @param bool        $is_test_mode       Whether the site is in test mode.
+		 */
+		$preselected_pmc_id = apply_filters( 'wc_stripe_preselect_payment_method_configuration', null, $is_test_mode );
+
+		if ( is_string( $preselected_pmc_id ) && str_starts_with( $preselected_pmc_id, 'pmc_' ) ) {
```
malithsen, line 135: We still support php 7.4. `str_starts_with` was introduced in php 8, so this will likely cause a fatal for those on 7.4. 

On a separate note, we do use this function in unit tests, but that won't affect real merchants.

malithsen (changes requested):
Looks good and tests well. I think we'd want to replace the `str_starts_with` as it conflicts with our declared minimum PHP version (7.4).

includes/class-wc-stripe-payment-method-configurations.php
daledupreez: @malithsen, WordPress Core has a polyfill for `str_starts_with()` and has had one since WordPress 5.9.0, at least per https://github.com/WordPress/wordpress-develop/commit/1046682e4f2453c55f836b3fe69f529daea0babc. So we should be able to use the function without any issues, even if sites are still running PHP 7.4

malithsen: ah TIL!

malithsen (approved):

PR #4721, opened 2025-10-10
Initial exploration for Agentic Commerce support

daledupreez:
Related to:
* https://github.com/woocommerce/woocommerce/pull/61271
* https://github.com/Automattic/woocommerce-payments/pull/11076

This PR explores implementing the necessary logic in the Stripe gateway to support the Agentic Commerce protocol ([agenticcommerce.dev](https://agenticcommerce.dev); [GitHub](https://github.com/agentic-commerce-protocol/agentic-commerce-protocol)), building on work from @ebinnion and @htdat in https://github.com/woocommerce/woocommerce/pull/61271.

### Gaps

We don't have Agentic Commerce support enabled for our Stripe accounts, so we're not able to fully test this end to end at the moment, but the code does attempt to call the Stripe `payment_intents` API with a shared payment token (and fails after multiple internal retries).

Beyond that, theare are also the following gaps:
 * I've made an assumption that we'll get full details for the payment method back from the payment intent creation, but we may need another API call to get those details
 * I haven't added an option to control the behaviour in addition to the feature flag -- I only have one enabled flag, and we likely need to offer a toggle as well as the feature flag, so we might need separate available and enabled flags.
 * I haven't tested "normal" checkout at all - that probably needs some cleanup too
 * I haven't yet explored what happens with various special types of products that have different flows, like subscriptions or pre-orders, nor have I looked at whether we can or should support any kind of off-session payments.

### Implementation approach

At a high level, I've created a sub-gateway specifically for this payment method, using optimized checkout as a model for something similar, where we don't know until later on which payment method is actually being used, and as a way to control the payment approach being available at all.

I'd be more than happy to take any and all feedback or suggestions, as I definitely consider this a proof of concept rather than a solid implementation.

I've genrally tried to re-use code within the plugin, but have made some changes to areas of the code that were expecting a payment method to be present before purchase, and added some special cases where we don't want actions

### Setup

You'll need a site with the following:
* WooCommerce latest
* The Stripe plugin built from this branch
* Enable the Agentic Commerce feature flag for WooCommerce Core: `wp option update woocommerce_feature_agentic_checkout_enabled 'yes'`
* Enable the Agentic Commerce feature flag for Stripe: `wp option update _wcstripe_feature_shared_payment_token 'yes'`
* Ensure you have one or more products on the site
* It will also help if you have logging enabled for the Stripe gateway

### Actual testing

* Run the following bash script to make an API call to the checkout session API with the following adjustments:
  - Modify `TEST_API_BASE` to your test site's host and port
  - Modify the `id` value to reference a product ID that exists on your site
```bash
TEST_API_BASE='http://localhost:8888/wp-json/wc/agentic/v1/checkout_sessions';
SESSION_ID=$(curl -X POST $TEST_API_BASE \
    -H "Content-Type: application/json" \
    -d '{
      "items": [{"id": "12", "quantity": 2}],
      "buyer": {
        "first_name": "Hello",
        "last_name": "Test",
        "email": "hello.test@example.com"
      },
      "fulfillment_address": {
        "line_one": "1234 Chat Road",
        "line_two": "Apt 101",
        "city": "San Francisco",
        "state": "CA",
        "country": "US",
        "postal_code": "94131"
      }
    }' | jq | tee create_response.json | jq -r .id) && echo "SESSION_ID: $SESSION_ID"
```
* Run the following bash script to try and check out:
```bash
curl -X POST "$TEST_API_BASE/{$SESSION_ID}/complete" \
    -H "Content-Type: application/json" \
    -d '{
    "buyer": {
      "first_name": "John",
      "last_name": "Smith",
      "email": "johnsmith@example.com",
      "phone_number": "15552003434"
    },
    "payment_data": {
      "token": "spt_card_visa",
      "provider": "stripe",
      "billing_address": {
        "name": "John Smith",
        "line_one": "1234 Chat Road,",
        "line_two": "",
        "city": "San Francisco",
        "state": "CA",
        "country": "US",
        "postal_code": "94131"
      }
    }
  }' | jq | tee complete_response.json | jq
```

The checkout step is not actually working at the moment, and works through multiple retries, so it can take quite a while to return. You can check the Stripe debug logs to see what data is being sent as part of the requests to Stripe.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/constants/class-wc-stripe-payment-methods.php
```diff
@@ -9,31 +9,32 @@
...
+	const KLARNA               = 'klarna';
+	const MULTIBANCO           = 'multibanco';
+	const OXXO                 = 'oxxo';
+	const P24                  = 'p24';
+	const SEPA                 = 'sepa';
+	const SEPA_DEBIT           = 'sepa_debit';
+	const SOFORT               = 'sofort';
+	const WECHAT_PAY           = 'wechat_pay';
+	const OC                   = 'card'; // This is a special case for the Optimized Checkout
+	const SHARED_PAYMENT_TOKEN = 'shared_payment_token';
```
htdat, line 39: IMO, SharedPaymentToken is a new type of token, not a new type of payment method so adding it to this list seems not proper. 

Based on my experience with WooPayments, `spt` is equivalent to other existing Stripe payment methods: [`pm`, `card` (in the other context, legacy Stripe), and `src`](https://github.com/Automattic/woocommerce-payments/blob/3463ea98431595a2e9dbfd2eaddec8b6603d5f39/includes/core/server/request/class-create-intention.php#L58).

htdat:
> I've made an assumption that we'll get full details for the payment method back from the payment intent creation, but we may need another API call to get those details

Yes, this is confirmed in [Stripe docs](https://docs.stripe.com/agentic-commerce/concepts/shared-payment-tokens#use-a-shared-payment-token): `When you confirm a PaymentIntent this way, the payment_method field automatically populates with a new payment method that’s reproduced from the one originally used by the buyer to create the SPT.`.


> I haven't added an option to control the behaviour in addition to the feature flag -- I only have one enabled flag, and we likely need to offer a toggle as well as the feature flag, so we might need separate available and enabled flags.

AFAIK, this is only available in US at the moment. And yeah, we might not display UI to enable this feature. 

> I haven't yet explored what happens with various special types of products that have different flows, like subscriptions or pre-orders, nor have I looked at whether we can or should support any kind of off-session payments.

The current protocol is not telling anything about subscriptions or pre-orders yet, so it's safe to exclude them. 

I also read somewhere that right now ChatGPT only supports a single product for each purchase, so the more completed flow such as one digital product + one physical product is not supported.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2818,6 +2862,43 @@ private function get_selected_payment_method_type_from_request() {
...
+
+		if ( ! class_exists( '\Automattic\WooCommerce\StoreApi\Routes\V1\Agentic\Enums\OrderMetaKey' ) ||
+			! defined( \Automattic\WooCommerce\StoreApi\Routes\V1\Agentic\Enums\OrderMetaKey::class . '::AGENTIC_CHECKOUT_SESSION_ID' )
+		) {
+			return null;
+		}
+
+		if ( empty( $order->get_meta( \Automattic\WooCommerce\StoreApi\Routes\V1\Agentic\Enums\OrderMetaKey::AGENTIC_CHECKOUT_SESSION_ID, true ) ) ) {
+			return null;
+		}
```
htdat, line 2900: Nice defensive here!

includes/constants/class-wc-stripe-payment-methods.php
daledupreez: That's completely fair and was part of why I am looking for feedback from other folks working on the Stripe plugin. The reason I went in this direction was partly to mimic how we approached the `OC` method above (for Optimized Checkout), as Optimized Checkout enables a different approach to processing payment methods that is operating much like a facade for one or more underlying payment methods, but we have to resolve the facade to the underlying payment method later on. In the case of Optimized Checkout, we get that resolution when the user submits their payment method from the UI, which is not quite the same as Shared Payment Tokens where we only get the resolved payment method after creating the payment intent, but is similar in that we need some handle on the broader approach to know that we have an SPT before we switch to 

I'd love feedback from @woocommerce/quark on this area, especially @wjrosa.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2568,19 +2583,41 @@ protected function prepare_payment_information_from_request( WC_Order $order ) {
...
 
+		// Agentic commerce uses a shared payment token rather than a payment method.
+		if ( WC_Stripe_Payment_Methods::SHARED_PAYMENT_TOKEN === $selected_payment_type ) {
+			$shared_payment_token = sanitize_text_field( wp_unslash( $_POST['wc-agentic_commerce-token'] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+
+			$payment_information['shared_payment_granted_token'] = $shared_payment_token;
+		} else {
+			// Override the payment method type with the API value when OC is enabled
+			if ( $this->oc_enabled ) {
+				$selected_payment_type = $payment_method_details->type ?? null;
```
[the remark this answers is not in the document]
daledupreez: Oops. Fixed in b1a5ca1c742775414f2765c64b819242c1ddd9dc.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -619,9 +634,12 @@ private function get_enabled_payment_method_config() {
 		$original_method_ids     = $enabled_payment_methods; // For OC, keep the original methods to control availability
 		$payment_methods         = $this->payment_methods;
 
-		// If the Optimized Checkout is enabled, we need to return just the card payment method + express methods.
-		// All payment methods are rendered inside the card container.
-		if ( $this->oc_enabled ) {
+		if ( $this->shared_payment_token_enabled ) {
+			$enabled_payment_methods[] = WC_Stripe_UPE_Payment_Method_Shared_Payment_Token::STRIPE_ID;
+			$payment_methods[ WC_Stripe_UPE_Payment_Method_Shared_Payment_Token::STRIPE_ID ] = new WC_Stripe_UPE_Payment_Method_Shared_Payment_Token();
+		} else if ( $this->oc_enabled ) {
```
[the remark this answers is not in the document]
daledupreez: Fixed in 1e4abcd1e2a12ddb5628b0aaa81716ea5d5b78fb.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -818,8 +818,12 @@ public function create_and_confirm_payment_intent( $payment_information ) {
 
 		$non_empty_params = [];
 
-		// The payment method is not required if we're using the confirmation token flow.
-		if ( empty( $payment_information['confirmation_token'] ) ) {
+		if ( WC_Stripe_Payment_Methods::SHARED_PAYMENT_TOKEN === ( $payment_information['selected_payment_type'] ?? null ) ) {
+			$required_params[] = 'shared_payment_granted_token';
+
+			$non_empty_params[] = 'shared_payment_granted_token';
+		} else if ( empty( $payment_information['confirmation_token'] ) ) {
```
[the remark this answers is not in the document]
daledupreez: Fixed in 1e4abcd1e2a12ddb5628b0aaa81716ea5d5b78fb.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -884,8 +891,12 @@ public function create_and_confirm_payment_intent( $payment_information ) {
 			$order
 		);
 
-		// Only update the payment_type if we have a reference to the payment type the customer selected.
-		if ( '' !== $selected_payment_type ) {
+		// TODO: Verify this works as expected.
+		// If we're processing a shared payment token, we need to look at the intent that was actually created.
+		if ( WC_Stripe_Payment_Methods::SHARED_PAYMENT_TOKEN === $selected_payment_type ) {
+			$order->update_meta_data( '_stripe_upe_payment_type', $payment_intent->type );
+		} else if ( '' !== $selected_payment_type ) {
```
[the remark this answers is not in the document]
daledupreez: Fixed in 1e4abcd1e2a12ddb5628b0aaa81716ea5d5b78fb.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -1053,18 +1064,20 @@ private function validate_payment_intent_required_params( $required_params, $non
...
 
 		$request = [
 			'shipping' => $payment_information['shipping'],
 		];
 
 		$is_using_confirmation_token = ! empty( $payment_information['confirmation_token'] );
-		if ( $is_using_confirmation_token ) {
+		if ( WC_Stripe_Payment_Methods::SHARED_PAYMENT_TOKEN === $selected_payment_type ) {
+			$request['shared_payment_granted_token'] = $payment_information['shared_payment_granted_token'];
+		} else if ( $is_using_confirmation_token ) {
```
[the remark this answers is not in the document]
daledupreez: Fixed in 1e4abcd1e2a12ddb5628b0aaa81716ea5d5b78fb.

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
```diff
@@ -0,0 +1,181 @@
...
+	 * Returns payment method title
+	 *
+	 * @param stdClass|array|bool $payment_details Optional payment details from charge object.
+	 *
+	 * @return string
+	 */
+	public function get_title( $payment_details = false ) {
+		// Wallet type
+		$wallet_type = null;
+		if ( $payment_details && ! empty( $payment_details->card->wallet->type ) ) {
```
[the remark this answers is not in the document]
daledupreez: Fixed in b4dbe297f147da684ed3a4ef577c5304c7c03d3f.

includes/constants/class-wc-stripe-payment-methods.php
wjrosa: I think it is fine to leave it this way for now, but it is worth adding a `@TODO` on top of it (and on top of `OC` as well), mentioning both should be moved later (to a new constant class).

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2818,6 +2862,43 @@ private function get_selected_payment_method_type_from_request() {
...
 	}
 
+	/**
+	 * Helper function to get a shared payment token from an incoming request for a specific order.
+	 *
+	 * @param WC_Order|null $order The order.
+	 * @return string|null The shared payment token, or null if it is not found.
+	 */
+	private function get_agentic_commerce_payment_token_from_request( ?WC_Order $order = null ): ?string {
+		if ( ! $order ) {
```
wjrosa, line 2875: Maybe we should log some of the checks from this method.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2818,6 +2862,43 @@ private function get_selected_payment_method_type_from_request() {
...
+			return null;
+		}
+
+		if ( ! isset( $_POST['wc-agentic_commerce-token'] ) || ! isset( $_POST['wc-agentic_commerce-provider'] ) ) {
+			return null;
+		}
+
+		if ( 'stripe' !== sanitize_text_field( wp_unslash( $_POST['wc-agentic_commerce-provider'] ) ) ) {
```
wjrosa: We can use the constant here (`WC_Gateway_Stripe::ID`).

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/**
+ * Class WC_Stripe_UPE_Payment_Method_Shared_Payment_Token
+ *
+ * This class represents the Stripe UPE payment method for the agentic Shared Payment Token flow.
+ */
+class WC_Stripe_UPE_Payment_Method_Shared_Payment_Token extends WC_Stripe_UPE_Payment_Method {
+
+	const STRIPE_ID = WC_Stripe_Payment_Methods::SHARED_PAYMENT_TOKEN;
```
wjrosa, line 13: This might be a good candidate to keep the payment method constant value (instead of `WC_Stripe_Payment_Methods`).

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
```diff
@@ -0,0 +1,180 @@
...
+		$is_stripe_enabled = ! empty( $main_settings['enabled'] ) && 'yes' === $main_settings['enabled'];
+
+		$shared_payment_token_enabled = $is_stripe_enabled &&
+			$this->shared_payment_token_enabled &&
+			[] !== $this->get_enabled_agentic_commerce_payment_methods();
+
+		$this->enabled     = $shared_payment_token_enabled ? 'yes' : 'no';
+		$this->id          = WC_Gateway_Stripe::ID; // Force the ID to be the same as the main payment gateway.
+		$this->stripe_id   = self::STRIPE_ID;
+		$this->title       = 'Stripe Agentic Commerce';
```
wjrosa: Just double-checking: we don't need to translate this string, right?

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
```diff
@@ -0,0 +1,180 @@
...
+		);
+	}
+
+	/**
+	 * Get the agentic commerce payment provider identifier.
+	 *
+	 * @return string Payment provider identifier.
+	 */
+	public function get_agentic_commerce_provider() {
+		return 'stripe';
```
wjrosa: Again, it might be nice to use the existing constant (`WC_Gateway_Stripe::ID`).

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
```diff
@@ -0,0 +1,180 @@
...
+
+	/**
+	 * Returns payment method title
+	 *
+	 * @param stdClass|array|bool $payment_details Optional payment details from charge object.
+	 *
+	 * @return string
+	 */
+	public function get_title( $payment_details = false ) {
+		// Wallet type
```
wjrosa, line 92: Not exactly something for this PR, but we should consider refactoring this method. We have some duplicate lines here and in the OC payment method class.

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
```diff
@@ -0,0 +1,180 @@
...
+		if ( $payment_details && ! empty( $payment_details->type ) ) { // Setting title for the order details page / thank you page.
+			$payment_method = WC_Stripe_UPE_Payment_Gateway::get_payment_method_instance( $payment_details->type );
+
+			// Avoid potential recursion by checking instance type. This fixes the title on pay for order confirmation page.
+			return $payment_method instanceof self ? parent::get_title() : $payment_method->get_title();
+		}
+
+		// Block checkout and pay for order (checkout) page.
+		if ( ( has_block( 'woocommerce/checkout' ) || ! empty( $_GET['pay_for_order'] ) ) && ! is_wc_endpoint_url( 'order-received' ) ) { // phpcs:ignore WordPress.Security.NonceVerification
+			return 'Stripe Agentic Commerce';
```
wjrosa: Double-checking again the need for translation here.

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
```diff
@@ -0,0 +1,180 @@
...
+	public function requires_automatic_capture() {
+		return false;
+	}
+
+	/**
+	 * Returns testing credentials to be printed at checkout in test mode.
+	 *
+	 * @return string
+	 */
+	public function get_testing_instructions( $show_optimized_checkout_instruction = false ) {
```
wjrosa, line 163: Another method to refactor later

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
daledupreez: @wjrosa, I am not sure what you mean by this comment. Could you rephrase to help me understand?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: Actually, this is a reference to `'stripe'` as the back end provider, and not the gateway ID. I don't think we should switch to the existing constant, as that could lead to confusion in future.

daledupreez: I switched this over to a new constant in 63bfb69af5b630b5cf2915d8e9011cdc4f01cd22.

includes/constants/class-wc-stripe-payment-methods.php
daledupreez: I've added the TODO in 348d7752649b5eed83a1a45377d3ffc1dd9ea4ae.

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
daledupreez: Good callout. We probably do. Done in 1d1cf284bec1085ab9fa8e83527917be5f183bd0.

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
daledupreez: As above, this is a reference to Stripe as the Agentic Commerce back end. I've created a new constant for this in 63bfb69af5b630b5cf2915d8e9011cdc4f01cd22.

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
daledupreez: Agreed.

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
daledupreez: Translation added in cd79ae617c70c8c7c09cdbe122030c60f1610ba6.

includes/payment-methods/class-wc-stripe-upe-payment-method-shared-payment-token.php
wjrosa: [Here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4721#discussion_r2430046014), I mentioned we should consider moving the constant value later for both the shared payment token and OC. I meant this line above could be a candidate for that. Later, we could change:
```php
	const STRIPE_ID = WC_Stripe_Payment_Methods::SHARED_PAYMENT_TOKEN;
```
to
```php
	const STRIPE_ID = 'shared_payment_token';
```
And use `WC_Stripe_UPE_Payment_Method_Shared_Payment_Token::STRIPE_ID` everywhere

ebinnion:
I'm going to close this PR for now as I think we'll go a different way with ACP.

PR #4724, opened 2025-10-13
Rename Klarna token class to 'WC_Payment_Token_Klarna'

Mayisha:
Fixes STRIPE-742
Fixes #4722

## Changes proposed in this Pull Request:
Reverts https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4671 and renames the class to 'WC_Payment_Token_Klarna'.

Deletion of Payment Tokens is handled via the `WC_Payment_Tokens` Class.
The Method `WC_Payment_Tokens::delete()` uses `WC_Payment_Tokens::get_token_classname()` to append `Klarna` to `WC_Payment_Token_`. This returns `WC_Payment_Token_Klarna` as a class name. 
So it expects and calls `**WC_Payment_Token_Klarna**`, not `WC_Stripe_Klarna_Payment_Token`.

## Testing instructions
- Code review.
- Verify the tests are passing.

PR #4725, opened 2025-10-13
Add filter to handle custom Klarna token name

### LINKED ISSUE #4722: WC_Stripe_Klarna_Payment_Token should be named WC_Payment_Token_Klarna or an error will be thrown when trying to delete the Klarna payment method from the WooCommerce client backend.
### Description

Deletion of Payment Tokens is hadled via the "WC_Payment_Tokens" Class.
The Method "WC_Payment_Tokens::delete()" uses "WC_Payment_Tokens::get_token_classname()" to append "klarna" to "WC_Payment_Token_" this returns "WC_Payment_Token_klarna" as a class name.

Since the class is now called "WC_Stripe_Klarna_Payment_Token" and not "WC_Payment_Token_Klarna" an error ("Uncaught Error: Class "WC_Payment_Token_klarna" not found") accours when trying to delete the payment method from the WooCommerce client backend.

//WC_Payment_Tokens::delete()
public static function delete( $token_id ) {
    $type = self::get_token_type_by_id( $token_id );
        if ( ! empty( $type ) ) {
            $class = self::get_token_classname( $type ); // This returns 'klarna', at least in my case
            $token = new $class( $token_id );  // This throws an error, since 'WC_Payment_Token_klarna' does not exist anymore
            $token->delete();
	}
}

//WC_Payment_Tokens::get_token_classname()
protected static function get_token_classname( $type ) {
    apply_filters( 'woocommerce_payment_token_class', 'WC_Payment_Token_' . $type, $type );
}

### Steps to reproduce

$type   = WC_Payment_Tokens::get_token_type_by_id( $token_id );
$class  = WC_Payment_Tokens::get_token_classname( $type );  // This returns 'klarna', at least in my case.
$token = new $class( $token_id ); // This throws an error, since 'WC_Payment_Token_klarna' does not exist anymore.

Try calling WC_Payment_Tokens::delete( $token_id ) with a valid Klarna token.

daledupreez:
Fixes STRIPE-742
Fixes #4722

Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4724

This PR takes a slightly different approach to @Mayisha's approach in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4724. Rather than renaming the token class, this PR simply ensures that we add a case for the Klarna token classname to our existing logic that hooks into the `woocommerce_payment_token_class` filter. This also means we don't need to deprecate the current class name.

One thing to note is that I've added a check to allow other plugins to override the token class name before us. This may be overly defensive, but I thought it better to be wary on this rather than assume all `klarna` tokens necessarily belong to our plugin.

## Testing instructions

* Verify that the unit tests are passing, both for Klarna and the other token classes implemented in `WC_Stripe_Payment_Tokens->woocommerce_payment_token_class()`
* Code review

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -30,6 +30,7 @@
 * Update - Reduce settings Javascript file size by using smaller image
 * Dev - Update @wordpress/scripts to 30.24.0 and @wordpress/base-styles to 6.7.0
 * Fix - Prevent fatal error when third-party plugins check for non-existent methods in payment method classes
+* Fix - Ensure Klarna payment tokens can be deleted and handled correctly
 
 = 9.9.2 - 2025-09-29 =
 * Fix - BACS instruction text appears twice on the Order Confirmation page
--- a/includes/payment-tokens/class-wc-stripe-payment-tokens.php
+++ b/includes/payment-tokens/class-wc-stripe-payment-tokens.php
@@ -805,7 +805,8 @@ class WC_Stripe_Payment_Tokens {
 	}
 
 	/**
-	 * Filters the payment token class to override the credit card class with the extension's version.
+	 * Filters the payment token class to handle token classes that don't match the default
+	 * WooCommerce payment token class name.
 	 *
 	 * @param string $class Payment token class.
 	 * @param string $type Token type.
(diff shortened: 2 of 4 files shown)


malithsen (approved):
This approach looks solid to me, and matches the existing pattern with ACH, BECS etc

Mayisha (approved):
Thanks @daledupreez. This seems like a better solution 👍

diegocurbelo (approved):
Looks good, all tests pass... and the type and class name are as expected.

Image

PR #4727, opened 2025-10-14
Add manual GitHub workflow for running e2e tests for a specific branch

daledupreez:

This PR is intended to start the work of creating a way to trigger all e2e tests for a specific branch in a standard way, with two primary workflows of interest:
 * We want to verify that our e2e tests are working correctly against a specific version of WooCommerce, especially beta releases
 * We want to verify that all e2e tests are working for release branches

This makes it possible to trigger the necessary tests in a completely client-agnostic way.

## Testing instructions

Testing for this will only be possible after the PR has been merged, as we don't have the workflow in place to trigger it manually yet.

So for now the only thing we can do is inspect the code.

---

### Changelog entry

<details>

#### Comment 
This PR only impacts our GitHub configuration and not any merchant-facing features. I don't think we should include the details in the changelog.
</details>

**Post merge**


wjrosa (approved):
Useful feature, Dale. Thank you. Looks good! 👍

.github/workflows/manual-e2e-tests.yml
```diff
@@ -0,0 +1,40 @@
+
+name: Manual E2E tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
```
diegocurbelo, line 7: I don't think the `branch` input is needed, `workflow_dispatch` always asks for a branch to run from. 
Maybe we can update the `.github/workflows/run-e2e-tests.yml` flow to allow `workflow_dispatch`.

For example, the [Generate ZIP file flow](https://github.com/woocommerce/woocommerce-gateway-stripe/actions/workflows/generate-zip.yml) allows you to choose a branch.

Action code: https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/.github/workflows/generate-zip.yml

daledupreez: I believe that selection actually does something different: it controls which version of the workflow file to use when running the workflow. It then has the side-effect that the default SHA/ref will be the same as that branch (as far as I understand [the docs for `workflow_dispatch` actions](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#workflow_dispatch).
However, that implicit expectation feels a bit shaky to me, as my assumption is that we should generally run the most current version of the workflow and only use the target branch for the code. Similarly, when we're testing adjustments to the workflow, we would want to run the workflow code from a specific branch and may want to specify other target branches to use for the contents of the main codebase.

diegocurbelo: You need to have the workflow in the default branch to be able to trigger it... But when you do, it uses the actual yml file from the branch you select in `Use workflow from`.

I think that having two branch inputs would be confusing 🤔 , but let's merge this so we can test it

diegocurbelo (approved):

.github/workflows/manual-e2e-tests.yml
daledupreez: Thanks for the scepticism. Let's see what this look like.

PR #4737, opened 2025-10-16
Improves the error message shown in checkout when a saved payment method is invalid

wjrosa:
Fixes STRIPE-572

As requested in STRIPE-572, this PR improves the error message shown when a saved payment method becomes invalid. The goal is to reduce confusion among shoppers and merchants. Shoppers may then attempt to use another method when this happens.

Error preview:
Screenshot 2025-10-16 at 18 17 34

## Testing instructions

- Checkout to this branch on your test environment (`fix/improving-error-message-for-detached-methods`)
- Connect your Stripe account
- As a shopper, add a new payment method using the My Account screen
- As a merchant, go to your Stripe dashboard
- Open the shopper details
- Delete the newly added payment method
- As the same shopper, attempt to purchase using the same saved payment method
- Confirm the new error message above is shown

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -21,6 +21,13 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 	use WC_Stripe_Subscriptions_Trait;
 	use WC_Stripe_Pre_Orders_Trait;
 
+	/**
+	 * Error string returned by Stripe when a PaymentMethod is detached.
+	 *
+	 * @var string
+	 */
+	protected const DETACHED_PAYMENT_METHOD_ERROR_STRING = 'The provided PaymentMethod was previously used with a PaymentIntent without Customer attachment';
```
diegocurbelo, line 29: Using the specific message text is not ideal, any small change to the text will break this.
But I don't think we have any other way of differentiating the specific error, and to be honest the messages do not change frequently and even if the message changes this doesn't break (it just reverts to the previous behavior).

diegocurbelo (approved):
The change look ok and tests as described!

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Yeah, I had the same concern. We don't have a specific error code for this, unfortunately. So, I think that's currently the only way to do it.

includes/wc-stripe-functions.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * WooCommerce Stripe Functions
+ *
+ * General functions available on both the front-end and admin.
+ *
+ * @package WooCommerce_Stripe\Functions
+ * @version 10.1.0
+ */
+
+if ( ! function_exists( 'str_contains' ) ) {
```
daledupreez: We don't need our own polyfill for this, as a polyfill for `str_contains()` has been in WordPress since 5.9.0 as per https://github.com/WordPress/wordpress-develop/commit/3cc8f1237a737ccf895682029afacafd7305d420. Given that 5.9.0 was released in early 2022 and our minimum supported version is 6.7, I think we should rely on the WordPress core polyfill.

daledupreez (changes requested):
Over and above removing the polyfill, this PR is not working for me.

After debugging things somewhat extensively, it looks like we're missing code in `WC_Stripe_UPE_Payment_Gateway->process_payment_with_saved_payment_method()` to call `$this->get_payment_intent_error_message()` after we have determined that the error is not retryable and before we call `$this->throw_localized_message()` (where we don't have any intent-specific logic -- see below). I also checked, and we can't simply hand the modified error message to `$this->throw_localized_message()`, as it will get overridden based on the error code. 🙃 

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/7d055cd4cb93a05086e97b05f375a25990d9f58f/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L1429-L1438

I am also having a separate issue where the returned error message is not being displayed in the UI, but that is unrelated to this PR.

includes/wc-stripe-functions.php
wjrosa: Thanks for pointing that out! Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4737/commits/44602b6e1196fe8e78b38fbb336e4ae8420fcdbb

wjrosa:
Thanks for reviewing and testing this, Dale!

> Over and above removing the polyfill, this PR is not working for me.

That's strange. I just tried it again after merging the branch with `develop`, and it is still working fine for me (shortcode and blocks checkout). Can you share some details of your testing so I can try to reproduce the issue?

> After debugging things somewhat extensively, it looks like we're missing code in WC_Stripe_UPE_Payment_Gateway->process_payment_with_saved_payment_method() to call $this->get_payment_intent_error_message() after we have determined that the error is not retryable and before we call $this->throw_localized_message() (where we don't have any intent-specific logic -- see below). I also checked, and we can't simply hand the modified error message to $this->throw_localized_message(), as it will get overridden based on the error code. 🙃

This flow is used only by the legacy checkout, which we don't support anymore. This code is from `process_payment_with_saved_payment_method`, which is now skipped (`includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#895`):

```php
		// Flag for using a deferred intent. To be removed.
		// https://github.com/woocommerce/woocommerce-gateway-stripe/issues/3868
		if ( ! empty( $_POST['wc-stripe-is-deferred-intent'] ) ) { <-- This is always `true` now
			return $this->process_payment_with_deferred_intent( $order_id );
		}

		if ( $this->maybe_change_subscription_payment_method( $order_id ) ) {
			return $this->process_change_subscription_payment_method( $order_id );
		}

		if ( $this->is_using_saved_payment_method() ) {
			return $this->process_payment_with_saved_payment_method( $order_id );
		}
```

daledupreez (approved):
I've re-tested and things are now looking good in both classic and block checkout. Not sure what was going on previously, but we should try to get this shipped!

PR #4747, opened 2025-10-23
Remove redundant secret key management when configuring webhooks

malithsen:

`configure_webhooks` method was managing API key state (setting/restoring) even though all callers already handle this responsibility. This PR simplifies things by removing that logic completely.

## Testing instructions

1. Test onboarding flow
2. Test Webhook reconfiguration.
 - Go to the Stripe settings page.
 - Click Account Details -> Configure connection -> Reconfigure webooks
 - Make no errors are logged
3. Test a payment method that involves webhooks (eg: SEPA)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,7 +7,7 @@
 * Update - Removes unnecessary legacy checkout gateway instantiations and UPE disablement code
 * Dev - Renames previous Order Helper class methods to use the `_id` suffix
 * Dev - Expands the Stripe Order Helper class to handle customer ID, card ID, UPE payment type, and UPE redirect status metas
-* Tweak - Remove redundant secret management logic when configuring webhooks
+* Fix - Remove redundant secret management logic when configuring webhooks
 
 = 10.0.1 - 2025-10-15 =
 * Fix - Remove persistent reconnection notices
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,6 +6,7 @@
 * Update - Removes unnecessary legacy checkout gateway instantiations and UPE disablement code
 * Dev - Renames previous Order Helper class methods to use the `_id` suffix
 * Dev - Expands the Stripe Order Helper class to handle customer ID, card ID, UPE payment type, and UPE redirect status metas
+* Tweak - Remove redundant secret management logic when configuring webhooks
 
 = 10.0.1 - 2025-10-15 =
 * Fix - Remove persistent reconnection notices
--- a/includes/admin/class-wc-rest-stripe-account-keys-controller.php
+++ b/includes/admin/class-wc-rest-stripe-account-keys-controller.php
@@ -405,6 +405,9 @@ class WC_REST_Stripe_Account_Keys_Controller extends WC_Stripe_REST_Base_Control
 			$response = $this->account->configure_webhooks( $environment );
 		} catch ( Exception $e ) {
 			return new WP_REST_Response( [ 'message' => $e->getMessage() ], 400 );
+		} finally {
+			// Ensure we reset the key before we do anything else.
+			WC_Stripe_API::set_secret_key( '' );
 		}
 
 		return new WP_REST_Response(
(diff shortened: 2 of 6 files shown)


includes/admin/class-wc-rest-stripe-account-keys-controller.php
```diff
@@ -399,8 +399,10 @@ public function configure_webhooks( WP_REST_Request $request ) {
...
 		}
 
+		WC_Stripe_API::set_secret_key( $secret );
+
 		try {
-			$response = $this->account->configure_webhooks( $environment, $secret );
+			$response = $this->account->configure_webhooks( $environment );
 		} catch ( Exception $e ) {
 			return new WP_REST_Response( [ 'message' => $e->getMessage() ], 400 );
 		}
```
daledupreez, line 400: Should we add a `finally` clause that ensures we reset the secret key?
```suggestion
		} finally {
			// Ensure we reset the key before we do anything else.
			WC_Stripe_API::set_secret_key( '' );
		}
```

changelog.txt
```diff
@@ -7,6 +7,7 @@
 * Update - Removes unnecessary legacy checkout gateway instantiations and UPE disablement code
 * Dev - Renames previous Order Helper class methods to use the `_id` suffix
 * Dev - Expands the Stripe Order Helper class to handle customer ID, card ID, UPE payment type, and UPE redirect status metas
+* Tweak - Remove redundant secret management logic when configuring webhooks
```
daledupreez: If this does address the issues merchants have been seeing, then I think this might need to be classified as a fix rather than a tweak.
```suggestion
* Fix - Remove redundant secret management logic when configuring webhooks
```

daledupreez (commented):
I have some feedback from inspection; will get to testing later today.

includes/admin/class-wc-rest-stripe-account-keys-controller.php
```diff
@@ -399,8 +399,10 @@ public function configure_webhooks( WP_REST_Request $request ) {
 			$secret = $saved_secret;
 		}
```
diegocurbelo: We should remove this code, and always use the secret from the settings.

This was necessary when the modal allowed manual keys to be entered, but currently, there is no reason to allow a manual secret key (aka shared key) here.

diegocurbelo (approved):
Looks good, I highlighted some code we might want to remove as we no longer support that flow.

includes/admin/class-wc-rest-stripe-account-keys-controller.php
malithsen: Make sense, as it would avoid having stale keys in options. 

Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4747/commits/00a70db15482befc9223e1362e6a58127e2b2cdf (invoked when clicking the reconfigure webhook button) and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4747/commits/c9b0f384e3c8984a0c891880ca292c9cd54266f3 (invoked during onboarding flow)

includes/admin/class-wc-rest-stripe-account-keys-controller.php
malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4747/commits/a4e2e37fac128511393643b056bceaba790a91cd

Mayisha (commented):
Functional test results

✅ Reconnection works.
✅ Configure webhook works.
✅ Reconnection and configure webhook both flows register a new webhook in Stripe dashboard.
✅ Reconnection and configure webhook both flows deletes the duplicate webhooks from the Stripe dashboard.

**Before reconnection** (manually created duplicate webhooks)

Screenshot 2025-10-24 at 9 08 58 PM


**After reconnection** (duplicate webhooks are deleted)

Screenshot 2025-10-24 at 9 12 10 PM

diegocurbelo (approved):

daledupreez (approved):
Testing looks good with disconnect, connect, reconnect, and reconfigure webhooks.

PR #4757, opened 2025-11-03
Improve ECE compatibility with One Page Checkout

malithsen:
Fixes STRIPE-743

This PR improves the compatibility between Stripe Express Checkout (ECE) buttons and the WooCommerce One Page Checkout (OPC) plugin. 

Problem:
When using the OPC plugin with ECE configured to show only on "checkout pages":
  1. ECE buttons don't appear on OPC pages because `is_checkout()` returns `false` (they're technically product pages)
  2. When buttons did appear with both locations enabled, they showed in the product section instead of the checkout section.
  3. ECE buttons displayed original product prices instead of discounted prices after applying coupon codes

Changes proposed:
  - Detect OPC pages using `is_wcopc_checkout()` and treat them as checkout context rather than product context
  - Skip product page restrictions for OPC pages when checkout buttons are enabled
  - Force frontend to use cart totals (which include discounts) instead of product pricing on OPC pages when checkout buttons are enabled

## Testing instructions

Prerequisites

- [WooCommerce One Page Checkout](https://woocommerce.com/products/woocommerce-one-page-checkout/) plugin installed and active

Test Case 1: ECE shows on OPC with checkout location enabled

1. Go to WooCommerce > Settings > Payments > Stripe
2. Under "Express Checkout", set button locations to "Checkout" only (uncheck Product and Cart)
3. Create a product and mark it as One Page Checkout enabled (OPC meta field _wcopc = 'yes')
4. View the product page
5. Expected: Express Checkout buttons (Apple Pay, Google Pay, etc.) appear in the checkout section of the OPC
6. Expected: No ECE buttons appear in the add-to-cart section

Test Case 2: Correct pricing with discount codes on OPC

1. With the same setup as Test Case 1
2. On the OPC product page, apply a coupon code in the checkout section
3. Observe the ECE buttons
4. Open an ECE (eg: Google Pay), the total in the payment modal should have the discount applied

Optional:
Test for regressions with OPC disabled.
 

https://github.com/user-attachments/assets/c0b4c26d-806a-4ec7-bc10-97f1e8e7f342

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,7 +15,7 @@
 * Fix - Increase limit when listing available payment method configurations from the Stripe API
 * Fix - Klarna not processing recurring payments
 * Fix - Fix Express Checkout error with free trial subscription on blocks cart/checkout
-* Fix - Improve ECE compatibility with One Page Checkout
+* Fix - Improve Express Checkout compatibility with One Page Checkout
 
 = 10.0.1 - 2025-10-15 =
 * Fix - Remove persistent reconnection notices
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@
 * Fix - Increase limit when listing available payment method configurations from the Stripe API
 * Fix - Klarna not processing recurring payments
 * Fix - Fix Express Checkout error with free trial subscription on blocks cart/checkout
+* Fix - Improve ECE compatibility with One Page Checkout
 
 = 10.0.1 - 2025-10-15 =
 * Fix - Remove persistent reconnection notices
--- a/includes/payment-methods/class-wc-stripe-express-checkout-element.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-element.php
@@ -233,7 +233,7 @@ class WC_Stripe_Express_Checkout_Element {
 			'is_pay_for_order'           => $this->express_checkout_helper->is_pay_for_order_page(),
 			'has_block'                  => has_block( 'woocommerce/cart' ) || has_block( 'woocommerce/checkout' ),
 			'login_confirmation'         => $this->express_checkout_helper->get_login_confirmation_settings(),
-			'is_product_page'            => $this->express_checkout_helper->is_product(),
+			'is_product_page'            => $this->get_is_product_page_for_ece(),
 			'is_checkout_page'           => $this->express_checkout_helper->is_checkout(),
 			'product'                    => $this->express_checkout_helper->get_product_data(),
 			'is_cart_page'               => is_cart(),
@@ -243,6 +243,33 @@ class WC_Stripe_Express_Checkout_Element {
 			'has_free_trial'             => $this->express_checkout_helper->has_free_trial(),
 		];
 	}
+
+	/**
+	 * Should ECE use product pricing (vs. cart pricing) in the current context.
+	 *
+	 * For One Page Checkout (OPC), when checkout buttons are enabled, always use cart
+	 * context so discounts/coupons are reflected.
+	 *
+	 * @return bool True to use product pricing; false to use cart totals.
+	 */
+	private function get_is_product_page_for_ece() {
+		if ( ! $this->express_checkout_helper->is_product() ) {
+			return false;
+		}
+
+		// OPC renders checkout on product pages; if ECE is shown on checkout, use cart pricing.
+		if (
+			function_exists( 'is_wcopc_checkout' )
+			&& is_wcopc_checkout()
+			&& $this->express_checkout_helper->should_show_ece_on_checkout_page()
+		) {
+			return false;
+		}
+
+		// Otherwise, product context is valid for ECE.
+		return true;
+	}
+
 
 	/**
 	 * Localizes additional parameters necessary for the Pay for Order page.
--- a/includes/payment-methods/class-wc-stripe-express-checkout-element.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-element.php
@@ -252,7 +252,7 @@ class WC_Stripe_Express_Checkout_Element {
 	 *
 	 * @return bool True to use product pricing; false to use cart totals.
 	 */
-	private function is_product_page_for_ece() {
+	public function is_product_page_for_ece() {
 		if ( ! $this->express_checkout_helper->is_product() ) {
 			return false;
 		}
--- a/includes/payment-methods/class-wc-stripe-express-checkout-element.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-element.php
@@ -259,8 +259,7 @@ class WC_Stripe_Express_Checkout_Element {
 
 		// OPC renders checkout on product pages; if ECE is shown on checkout, use cart pricing.
 		if (
-			function_exists( 'is_wcopc_checkout' )
-			&& is_wcopc_checkout()
+			$this->express_checkout_helper->is_one_page_checkout()
 			&& $this->express_checkout_helper->should_show_ece_on_checkout_page()
 		) {
 			return false;
--- a/includes/payment-methods/class-wc-stripe-express-checkout-element.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-element.php
@@ -233,7 +233,7 @@ class WC_Stripe_Express_Checkout_Element {
 			'is_pay_for_order'           => $this->express_checkout_helper->is_pay_for_order_page(),
 			'has_block'                  => has_block( 'woocommerce/cart' ) || has_block( 'woocommerce/checkout' ),
 			'login_confirmation'         => $this->express_checkout_helper->get_login_confirmation_settings(),
-			'is_product_page'            => $this->get_is_product_page_for_ece(),
+			'is_product_page'            => $this->is_product_page_for_ece(),
 			'is_checkout_page'           => $this->express_checkout_helper->is_checkout(),
 			'product'                    => $this->express_checkout_helper->get_product_data(),
 			'is_cart_page'               => is_cart(),
@@ -252,7 +252,7 @@ class WC_Stripe_Express_Checkout_Element {
 	 *
 	 * @return bool True to use product pricing; false to use cart totals.
 	 */
-	private function get_is_product_page_for_ece() {
+	private function is_product_page_for_ece() {
 		if ( ! $this->express_checkout_helper->is_product() ) {
 			return false;
 		}
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -888,7 +888,7 @@ class WC_Stripe_Express_Checkout_Helper {
 	 *
 	 * @return boolean True if on a One Page Checkout page, false otherwise.
 	 */
-	public function is_one_page_checkout() {
+	public function is_one_page_checkout(): bool {
 		return function_exists( 'is_wcopc_checkout' ) && is_wcopc_checkout();
 	}
 
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -682,8 +682,12 @@ class WC_Stripe_Express_Checkout_Helper {
 			return false;
 		}
 
+		// Detect One Page Checkout pages for special handling.
+		$is_one_page_checkout = function_exists( 'is_wcopc_checkout' ) && is_wcopc_checkout();
+
 		// Don't show on checkout if disabled.
-		if ( is_checkout() && ! $this->should_show_ece_on_checkout_page() ) {
+		// One Page Checkout plugin creates checkout functionality on product pages, so we need to check for it.
+		if ( ( is_checkout() || $is_one_page_checkout ) && ! $this->should_show_ece_on_checkout_page() ) {
 			WC_Stripe_Logger::log( 'Stripe Express Checkout buttons display on checkout is disabled. ' );
 			return false;
 		}
@@ -691,8 +695,16 @@ class WC_Stripe_Express_Checkout_Helper {
 		$is_product = $this->is_product();
 
 		// Don't show if product page ECE is disabled.
-		if ( $is_product && ! $this->should_show_ece_on_product_pages() ) {
+		// Skip this check for One Page Checkout pages since they should be treated as checkout pages, not product pages.
+		if ( $is_product && ! $is_one_page_checkout && ! $this->should_show_ece_on_product_pages() ) {
 			WC_Stripe_Logger::log( 'Stripe Express Checkout buttons display on product pages is disabled. ' );
+			return false;
+		}
+
+		// On One Page Checkout pages, if we're in the product section and checkout buttons are enabled,
+		// skip rendering here because the button will render in the checkout section instead.
+		if ( $is_one_page_checkout && $is_product && doing_action( 'woocommerce_after_add_to_cart_form' ) && $this->should_show_ece_on_checkout_page() ) {
+			WC_Stripe_Logger::log( 'Stripe Express Checkout buttons on One Page Checkout will be displayed in checkout section, not product section.' );
 			return false;
 		}
 
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -682,12 +682,9 @@ class WC_Stripe_Express_Checkout_Helper {
 			return false;
 		}
 
-		// Detect One Page Checkout pages for special handling.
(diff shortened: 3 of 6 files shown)


changelog.txt
```diff
@@ -15,6 +15,7 @@
 * Fix - Increase limit when listing available payment method configurations from the Stripe API
 * Fix - Klarna not processing recurring payments
 * Fix - Fix Express Checkout error with free trial subscription on blocks cart/checkout
+* Fix - Improve ECE compatibility with One Page Checkout
```
daledupreez: Nit: I think we should use "Express Checkout" rather than "ECE", as it is clearer to a broader audience.

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -984,4 +985,46 @@ public function provide_test_has_free_trial() {
...
+		$gateway = $this->getMockBuilder( WC_Stripe_UPE_Payment_Gateway::class )
+			->disableOriginalConstructor()
+			->getMock();
+
+		$helper = new WC_Stripe_Express_Checkout_Helper( $gateway );
+
+		// The actual test: Check if should_show_ece_on_checkout_page returns the expected value.
+		$result = $helper->should_show_ece_on_checkout_page();
+
+		$this->assertEquals( in_array( 'checkout', $button_locations, true ), $result );
```
daledupreez: This feels like a really complex set of logic to me. Can we split out the logic so it's a bit more explicit as to what we are testing?
```suggestion
		$should_show_ece = in_array( 'checkout', $button_locations, true );
		$this->assertEquals( $should_show_ece, $result );
```

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -984,4 +985,46 @@ public function provide_test_has_free_trial() {
...
+	 *
+	 * @dataProvider provide_opc_detection_scenarios
+	 *
+	 * @param bool  $is_opc Whether OPC function returns true.
+	 * @param array $button_locations Button location settings.
+	 * @param bool  $should_allow_checkout Expected result for checkout location check.
+	 *
+	 * @return void
+	 */
+	public function test_opc_detection_logic( $is_opc, $button_locations, $should_allow_checkout ) {
```
daledupreez: `$is_opc` is not used in the test implementation, so the test cases are not actually testing what they claim to be testing. We need to fix that in some way. It may be easiest to do that by creating an `is_one_page_checkout()` helper method in `WC_Stripe_Express_Checkout_Helper`, which we can then mock as needed.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -244,6 +244,33 @@ public function javascript_params() {
...
 
+	/**
+	 * Should ECE use product pricing (vs. cart pricing) in the current context.
+	 *
+	 * For One Page Checkout (OPC), when checkout buttons are enabled, always use cart
+	 * context so discounts/coupons are reflected.
+	 *
+	 * @return bool True to use product pricing; false to use cart totals.
+	 */
+	private function get_is_product_page_for_ece() {
```
daledupreez: Can we simplify this name to `is_product_page_for_ece()`? It feels like the leading `get_` is unnecessary and a bit weird to me.
```suggestion
	private function is_product_page_for_ece() {
```

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -233,7 +233,7 @@ public function javascript_params() {
 			'is_pay_for_order'           => $this->express_checkout_helper->is_pay_for_order_page(),
 			'has_block'                  => has_block( 'woocommerce/cart' ) || has_block( 'woocommerce/checkout' ),
 			'login_confirmation'         => $this->express_checkout_helper->get_login_confirmation_settings(),
-			'is_product_page'            => $this->express_checkout_helper->is_product(),
+			'is_product_page'            => $this->get_is_product_page_for_ece(),
```
daledupreez: Given that we're adding logic local to `WC_Stripe_Express_Checkout_Element`, it feels like we should add some tests to cover the new logic.

daledupreez (commented):
Functionally, this is working as described. I also tested the situation where express checkout is enabled in the product page _and_ checkout, and confirmed that we're still showing only the express checkout buttons in the One Page Checkout section and not in the main product section. I also smoke tested loading an ECE modal from the button without One Page Checkout active, and that was clear, as were the related e2e tests -- there were some ACH failures, but those are being worked on in a dedicated PR.

However, I think there are some test-related improvements we can make, or at least discuss before we merge this PR.

changelog.txt
malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/2f9cdf8857b6facb9688bad22c3b9c0b0e71aa89

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
malithsen: This test was updated in follow-up commits

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
malithsen: I made some changes to this test in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/a8f06c79ffbab6e2222d13112c6635ab656887ee but I am getting test [failures](https://github.com/woocommerce/woocommerce-gateway-stripe/actions/runs/19112079348/job/54611555245) in the CI test runner, although they pass locally 🤷 so resorted to removing the failing cases in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/1adddef56ae7cfadf4f885cd61d827b03a2bd986 😞

includes/payment-methods/class-wc-stripe-express-checkout-element.php
malithsen: Makes sense, updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/44d54743bd9cf6fd2a0364f6f03c27045fcd0f77

includes/payment-methods/class-wc-stripe-express-checkout-element.php
malithsen: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/fad1b896e8952a349c871f69a6f04bd687e8fb04

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -683,19 +683,28 @@ public function should_show_express_checkout_button() {
 		}
 
 		// Don't show on checkout if disabled.
-		if ( is_checkout() && ! $this->should_show_ece_on_checkout_page() ) {
+		// One Page Checkout plugin creates checkout functionality on product pages, so we need to check for it.
+		if ( ( is_checkout() || $this->is_one_page_checkout() ) && ! $this->should_show_ece_on_checkout_page() ) {
```
daledupreez: Non-blocking nit: rather than calling `$this->is_one_page_checkout()` multiple times, I think it would be simpler to stash the value in a local variable, much like we're doing for `$is_product` below.
```suggestion
		$is_one_page_checkout = $this->is_one_page_checkout();

		// One Page Checkout plugin creates checkout functionality on product pages, so we need to check for it.
		if ( ( is_checkout() || $is_one_page_checkout ) && ! $this->should_show_ece_on_checkout_page() ) {
```

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -874,6 +883,15 @@ public function should_show_ece_on_checkout_page() {
 		);
 	}
 
+	/**
+	 * Detects if the current page is a One Page Checkout page.
+	 *
+	 * @return boolean True if on a One Page Checkout page, false otherwise.
+	 */
+	public function is_one_page_checkout() {
```
daledupreez: Nit: can we specify the return type explicitly?
```suggestion
	public function is_one_page_checkout(): bool {
```

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Element_Test.php
```diff
@@ -431,4 +431,87 @@ public function test_add_order_attribution_data() {
...
+			->method( 'is_product' )
+			->willReturn( $is_product );
+
+		$helper->expects( $this->any() )
+			->method( 'is_one_page_checkout' )
+			->willReturn( $is_opc );
+
+		$helper->expects( $this->any() )
+			->method( 'should_show_ece_on_checkout_page' )
+			->willReturn( $show_ece_on_checkout );
```
daledupreez: Question: can we simply drop the `expects()` statements, as they are not actually verifying any behaviour?
```suggestion
		$helper->method( 'is_product' )
			->willReturn( $is_product );

		$helper->method( 'is_one_page_checkout' )
			->willReturn( $is_opc );

		$helper->method( 'should_show_ece_on_checkout_page' )
			->willReturn( $show_ece_on_checkout );
```

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -21,6 +21,7 @@
  * @package WooCommerce/Stripe/WC_Stripe_Express_Checkout_Helper
  *
  * WC_Stripe_Express_Checkout_Helper_Test class.
+ * @group helper
```
daledupreez: It feels like this is something we should not include in this PR. Is this intentionally included?

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -984,4 +985,79 @@ public function provide_test_has_free_trial() {
...
+		// Create a mock product.
+		$product = WC_Helper_Product::create_simple_product();
+		$is_product_page = $is_opc || in_array( 'product', $button_locations, true );
+
+		// Mock the methods.
+		$wc_stripe_ece_helper_mock->expects( $this->any() )->method( 'is_one_page_checkout' )->willReturn( $is_opc );
+		$wc_stripe_ece_helper_mock->expects( $this->any() )->method( 'is_product' )->willReturn( $is_product_page );
+		$wc_stripe_ece_helper_mock->expects( $this->any() )->method( 'is_checkout' )->willReturn( false );
+		$wc_stripe_ece_helper_mock->expects( $this->any() )->method( 'allowed_items_in_cart' )->willReturn( true );
+		$wc_stripe_ece_helper_mock->expects( $this->any() )->method( 'get_product' )->willReturn( $is_product_page ? $product : false );
```
daledupreez: As above, can we remove the `expects()` calls that are effectively noise?
```suggestion
		$wc_stripe_ece_helper_mock->method( 'is_one_page_checkout' )->willReturn( $is_opc );
		$wc_stripe_ece_helper_mock->method( 'is_product' )->willReturn( $is_product_page );
		$wc_stripe_ece_helper_mock->method( 'is_checkout' )->willReturn( false );
		$wc_stripe_ece_helper_mock->method( 'allowed_items_in_cart' )->willReturn( true );
		$wc_stripe_ece_helper_mock->method( 'get_product' )->willReturn( $is_product_page ? $product : false );
```

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
```diff
@@ -984,4 +985,79 @@ public function provide_test_has_free_trial() {
...
+
+	/**
+	 * Data provider for OPC detection scenarios.
+	 *
+	 * @return array
+	 */
+	public function provide_opc_detection_scenarios() {
+		return [
+			'OPC with checkout enabled'     => [ true, [ 'checkout' ], true ],
+			'Non-OPC with checkout enabled' => [ false, [ 'checkout' ], true ],
```
daledupreez, line 1052: It really feels like we should work out what is going on with the tests so we can add tests for only the `product` page, as well as both the `product` and `checkout` pages.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
malithsen: Good idea! Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/11a0cc0e93d02b541a20581a203a0ea761b6dacb

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Element_Test.php
malithsen: I don't think it adds any value. We have this pattern everywhere in our tests. I removed them from the files touched in this PR in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/2270db97baf30b4119f1edcedc2d3e17c12a3a93

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
malithsen: Ah thanks. It shouldn't have been committed. Removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/831deac79ec74838e4ece1f8529866ceb6b933e0

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
malithsen: Thanks! Removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/2270db97baf30b4119f1edcedc2d3e17c12a3a93

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
malithsen: Turns out I updated the wrong setting. Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/0134bee28ba96da00e755be76eb2e090d8f02d24

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
malithsen: Thanks! Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/ff597e2d29880ca2564c37968c0d0675258e8c2b

diegocurbelo (approved):
The changes look good and work as described.

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Element_Test.php
```diff
@@ -431,4 +426,84 @@ public function test_add_order_attribution_data() {
...
+	 * @dataProvider provide_test_is_product_page_for_ece
+	 */
+	public function test_is_product_page_for_ece( $is_product, $is_opc, $show_ece_on_checkout, $expected_result ) {
+		$ajax_handler = $this->getMockBuilder( WC_Stripe_Express_Checkout_Ajax_Handler::class )
+			->disableOriginalConstructor()
+			->getMock();
+
+		$helper = $this->getMockBuilder( WC_Stripe_Express_Checkout_Helper::class )
+			->disableOriginalConstructor()
+			->setMethods( [ 'is_product', 'is_one_page_checkout', 'should_show_ece_on_checkout_page' ] )
```
daledupreez: Nit: `setMethods()` was removed in PHPUnit 10, so it will be better to use `onlyMethods()` to avoid future work.
```suggestion
			->onlyMethods( [ 'is_product', 'is_one_page_checkout', 'should_show_ece_on_checkout_page' ] )
```

daledupreez (approved):
Thanks for the updates to the tests! 🚢

tests/phpunit/PaymentMethods/WC_Stripe_Express_Checkout_Helper_Test.php
malithsen: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4757/commits/0134bee28ba96da00e755be76eb2e090d8f02d24

PR #4758, opened 2025-11-03
Prevent retries on unrecoverable errors

malithsen:
Relates STRIPE-739

This PR updates the request retry logic so that we don't retry requests that failed due to payment method declines, as they will continue to fail.

Additional context: https://linear.app/a8c/issue/STRIPE-739/payment-failed-the-provided-paymentmethod-cannot-be-attached#comment-d139b0e7

## Testing instructions
Code review is enough. Linked Linear issue has links to Stripe Platform logs that contain the error codes added here.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -157,7 +157,9 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 			(
 				'payment_intent_mandate_invalid' === $error->code || // Don't retry payments when a 3DS mandate is invalid.
 				'charge_exceeds_transaction_limit' === $error->code || // Don't retry payments when the charge exceeds the transaction limit.
-				'amount_too_small' === $error->code
+				'amount_too_small' === $error->code ||
+				'card_declined' === $error->code ||
+				'payment_method_provider_decline' === $error->code
 			)
 		) {
 			return false;
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -152,26 +152,27 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 	public function is_retryable_error( $error ) {
 		// Note that this check is required since the error type is 'invalid_request_error' which
 		// would otherwise return true.
-		if (
-			isset( $error->code ) &&
-			(
-				'payment_intent_mandate_invalid' === $error->code || // Don't retry payments when a 3DS mandate is invalid.
-				'charge_exceeds_transaction_limit' === $error->code || // Don't retry payments when the charge exceeds the transaction limit.
-				'amount_too_small' === $error->code ||
-				'card_declined' === $error->code ||
-				'payment_method_provider_decline' === $error->code
-			)
-		) {
+		$non_retryable_codes = [
+			'payment_intent_mandate_invalid',     // Don't retry payments when a 3DS mandate is invalid.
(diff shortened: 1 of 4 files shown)


diegocurbelo (approved):
Looks good.

[nit]: I find using `in_array` or a `switch` instead of multiple ORs for the same variable easier to understand, but that can be subjective, so I'm ok with keeping it like this.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -157,7 +157,9 @@ public function is_retryable_error( $error ) {
 			(
 				'payment_intent_mandate_invalid' === $error->code || // Don't retry payments when a 3DS mandate is invalid.
 				'charge_exceeds_transaction_limit' === $error->code || // Don't retry payments when the charge exceeds the transaction limit.
-				'amount_too_small' === $error->code
+				'amount_too_small' === $error->code ||
+				'card_declined' === $error->code ||
+				'payment_method_provider_decline' === $error->code
 			)
```
daledupreez: I agree with Diego's comment here about using `in_array()`, especially now that we're up to 5 candidate values!
```suggestion
			in_array(
				$error->code,
				[
					'payment_intent_mandate_invalid', // Don't retry payments when a 3DS mandate is invalid.
					'charge_exceeds_transaction_limit', // Don't retry payments when the charge exceeds the transaction limit.
					'amount_too_small',
					'card_declined',
					'payment_method_provider_decline',
				],
				true
			)

**Review by daledupreez — APPROVED** [review]
Changes look good to me.

As a minor note, it may be worth describing this PR as `Towards` the linked issue rather than `Relates`.

Also, might it be worth adding a unit test for `is_retryable_error()`? I see we don't have any tests for that method at the moment.

File: includes/abstracts/abstract-wc-stripe-payment-gateway.php (hunk 1 above)
Comment (daledupreez) [reply includes/abstracts/abstract-wc-stripe-payment-gateway.php]: Note: this is also a nit, and not blocking!

Comment (malithsen) [reply includes/abstracts/abstract-wc-stripe-payment-gateway.php]: Refactored this function in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4758/commits/2febcbf36ba4c59214744815f80619559364f5bf

**Comment by malithsen** [comment]
> Also, might it be worth adding a unit test for is_retryable_error()? I see we don't have any tests for that method at the moment.

Added tests in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4758/commits/932ac0f53f99d30143db86c642f9e5dc0aa1423e

File: tests/phpunit/WC_Stripe_Payment_Gateway_Test.php [hunk 2]
```diff
@@ -817,4 +816,59 @@ public function provide_test_payment_icons() {
...
+	 */
+	public function test_non_retryable_error_codes_return_false( $error_code ) {
+		$error       = new \stdClass();
+		$error->code = $error_code;
+		$error->type = 'invalid_request_error';
+
+		$result = $this->gateway->is_retryable_error( $error );
+
+		$this->assertFalse( $result, "Error code '{$error_code}' should not be retryable" );
+	}
```
daledupreez, line 833: Not blocking, but the test case for `invalid_request_error` with some other error code is not clearly handled anywhere. `test_retryable_error_types_return_true()` implicitly covers it, but only by having an empty `$error->code` value.

daledupreez (approved):
Still LGTM -- I didn't retest, but the changes were mostly to do with tests.

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
malithsen: Good point, added an extra test in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4758/commits/04d851577b3d27228eca77f3f7f970cea593853d

PR #4761, opened 2025-11-05
Log the masked secret key when making API calls

diegocurbelo:
Fixes #<github_issue_id>

This PR logs the masked secret key when making API calls (to the`request` and `retrieve` methods).
It uses the same masking as the Stripe dashboard (the first 8 characters, 3 dots, and the last 6 characters).

Screenshot 2025-11-05 at 11 38 43

## Testing instructions

Code review.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-api.php
+++ b/includes/class-wc-stripe-api.php
@@ -235,8 +235,16 @@ class WC_Stripe_API {
 		 */
 		$request = apply_filters( 'wc_stripe_request_body', $request, $api );
 
+		$masked_secret_key = self::get_masked_secret_key();
+
 		// Log the request after the filters have been applied.
-		WC_Stripe_Logger::debug( "Stripe API request: {$method} {$api}", [ 'request' => $request ] );
+		WC_Stripe_Logger::debug(
+			"Stripe API request: {$method} {$api}",
+			[
+				'stripe_api_key' => $masked_secret_key,
+				'request'        => $request,
+			]
+		);
 
 		$response = wp_safe_remote_post(
 			self::ENDPOINT . $api,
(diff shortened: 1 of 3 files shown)


malithsen (approved):
LGTM

includes/class-wc-stripe-api.php
```diff
@@ -632,4 +653,18 @@ private static function get_stripe_request_id( $response ) {
...
+	 * It uses the same pattern as the Stripe dashboard: sk_live_...JLWaeq.
+	 *
+	 * @return string The masked secret key.
+	 */
+	private static function get_masked_secret_key(): string {
+		$key = self::get_secret_key();
+		if ( empty( $key ) ) {
+			return 'secret_key_not_configured';
+		}
+		return substr( $key, 0, 8 ) . '...' . substr( $key, -6 );
```
daledupreez, line 696: Minor non-blocking thought: might it be worth using `***` rather than `...`, as that's more generally used as a way to represent masked or hidden data? If we want to be _really_ fancy we could use `•••` too, but I think it may be more reliable to stick with an ASCII value.
```suggestion
		return substr( $key, 0, 8 ) . '***' . substr( $key, -6 );
```

daledupreez (approved):
The change looks good to me, and I verified that it works for `GET` and `POST` requests. I have one minor comment about the mask formatting, but it's _definitely_ not blocking.

includes/class-wc-stripe-api.php
diegocurbelo: I used `...` to be consistent with how Stripe displays the masked keys in their Logs dashboard, but I don't have a strong opinion about the character used 🤷🏼.

PR #4762, opened 2025-11-05
ECE: Make postal code verification optional for certain countries

malithsen:
Fixes STRIPE-699

This PR fixes a checkout validation issue where express checkout fails for countries that don't require postal codes (eg: Israel). Although postal codes are optional on the express checkout form, WooCommerce's default validation requires them, causing order failures.

This PR extends `modify_country_locale_for_express_checkout` method to mark the postcode field as optional for specific countries. A similar approach was first introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4451 for State validation. For postal codes, I couldn't find a definitive list of countries where they are optional on Google Pay/Apple Pay. So in this PR, I've only listed Israel (which was reported in the Linear issue attached) and introduced a filter that merchants can use to add more countries. 

## Testing instructions

**Setup**
Google Test Card suite doesn't have any Israel addresses and it didn't allow me to add one. So I ended up hard-coding an address in the `ajax_normalize_address` function.

```
	/**
	 * Normalizes address fields in WooCommerce supported format.
	 */
	public function ajax_normalize_address() {
		check_ajax_referer( 'wc-stripe-express-checkout-normalize-address', 'security' );

		$data = filter_input( INPUT_POST, 'data', FILTER_DEFAULT, FILTER_REQUIRE_ARRAY );

		// TEMPORARY TEST CODE - Simulate Israel address without postcode from Google Pay.
		if ( ! empty( $data['billing_address'] ) ) {
			$data['billing_address']['country']  = 'IL';
			$data['billing_address']['city']     = 'Tel Aviv';
			$data['billing_address']['address_1'] = 'Rothschild Blvd 1';
			$data['billing_address']['state']    = '';
			$data['billing_address']['postcode'] = ''; // Empty postcode - this is what causes the issue.
		}
		if ( ! empty( $data['shipping_address'] ) ) {
			$data['shipping_address']['country']  = 'IL';
			$data['shipping_address']['city']     = 'Tel Aviv';
			$data['shipping_address']['address_1'] = 'Rothschild Blvd 1';
			$data['shipping_address']['state']    = '';
			$data['shipping_address']['postcode'] = ''; // Empty postcode - this is what causes the issue.
		}

		// Normalizes billing and shipping state values.
		$normalized_data = $this->express_checkout_helper->normalize_state( $data );
		$normalized_data = $this->express_checkout_helper->fix_address_fields_mapping( $normalized_data );

		wp_send_json( $normalized_data );
	}

```

**Test Google Pay**
1. As a shopper, add a product to your cart and go to the checkout page.
2. Pay with Google Pay.
3. In `develop`, order will fail. On this branch, you should be able to place the order.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
@@ -402,6 +402,7 @@ class WC_Stripe_Express_Checkout_Ajax_Handler {
 	/**
 	 * Modify country locale for express checkout.
 	 * Countries that don't have state fields, make the state field optional.
+	 * Make postcode optional for specific countries during express checkout.
 	 *
 	 * @param array $locale The country locale.
 	 * @return array Modified country locale.
@@ -419,6 +420,20 @@ class WC_Stripe_Express_Checkout_Ajax_Handler {
 			if ( empty( $states ) ) {
 				$locale[ $country_code ]['state']['required'] = false;
 			}
+		}
+
+		// List of countries where postcode is optional in payment providers (Google Pay, Apple Pay).
+		// These countries allow addresses without postal codes, but WooCommerce requires them by default.
+		$countries_with_optional_postcode = apply_filters(
+			'wc_stripe_express_checkout_countries_with_optional_postcode',
+			[
+				'IL', // Israel
+			]
+		);
+
+		// Make postcode optional for countries where payment providers don't require it.
+		foreach ( $countries_with_optional_postcode as $country_code ) {
+			$locale[ $country_code ]['postcode']['required'] = false;
 		}
 
 		return $locale;
--- a/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
@@ -422,7 +422,7 @@ class WC_Stripe_Express_Checkout_Ajax_Handler {
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -421,6 +422,20 @@ public function modify_country_locale_for_express_checkout( $locale ) {
...
+		$countries_with_optional_postcode = apply_filters(
+			'wc_stripe_express_checkout_countries_with_optional_postcode',
+			[
+				'IL', // Israel
+			]
+		);
+
+		// Make postcode optional for countries where payment providers don't require it.
+		foreach ( $countries_with_optional_postcode as $country_code ) {
+			$locale[ $country_code ]['postcode']['required'] = false;
```
daledupreez: Is it worth being defensive to make sure the country code exists before we cause more problems?
```suggestion
			if ( isset( $locale[ $country_code ] ) ) {
				$locale[ $country_code ]['postcode']['required'] = false;
			}
```

malithsen: Good idea, it makes sense to add that check, given that we are allowing merchants to set the country_code in the filter above.

diegocurbelo (approved):
The change looks good, and tests well with the hard-coded empty postcode.

daledupreez (approved):
Thanks for the tweak. 🚢

PR #4765, opened 2025-11-06
Add verbose debug mode for OAuth connect flow

diegocurbelo:
Fixes #<github_issue_id>

This PR adds detailed logging to the OAuth connection flow, especially to the save/update keys related functions.

All the new logging calls are wrapped in `if ( WC_Stripe_Helper::is_verbose_debug_mode_enabled() ) {...}` checks, so it won't have any impact on the plugin's normal usage unless the filter `wc_stripe_is_verbose_debug_mode_enabled` is set.

## Testing instructions

1. Add the following snippet to your store:
    `add_filter( 'wc_stripe_is_verbose_debug_mode_enabled', '__return_true' );`
2. Disconnect the Stripe account (if connected)
3. Navigate to the plugin settings
4. Connect a Stripe account (any mode)
5. Open the plugin logs (`WooCommerce > Status > Logs` and then `woocommerce-gateway-stripe`)
6. Check that the new log entries (`OAuth: *`) are being logged:
    Screenshot 2025-11-06 at 17 16 37
7. Check that no sensitive information is being logged:
    Screenshot 2025-11-06 at 17 18 26

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,6 +16,7 @@
 * Fix - Increase limit when listing available payment method configurations from the Stripe API
 * Fix - Klarna not processing recurring payments
 * Fix - Fix Express Checkout error with free trial subscription on blocks cart/checkout
+* Dev - Add Stripe's masked API key to API request/response logs
 
 = 10.0.1 - 2025-10-15 =
 * Fix - Remove persistent reconnection notices
--- a/changelog.txt
+++ b/changelog.txt
@@ -27,6 +27,7 @@
 * Fix - Detect WooCommerce Subscriptions staging sites when checking if payments can be detached
 * Fix - Fix saved ACH payment methods sending unsupported capture_method parameter causing checkout failures
 * Dev - Add Stripe's masked API key to API request/response logs
+* Dev - Add verbose debug logging mode to the OAuth connect flow
 
 = 10.0.1 - 2025-10-15 =
 * Fix - Remove persistent reconnection notices
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -2046,13 +2046,13 @@ class WC_Stripe_Helper {
 	}
 
 	/**
-	 * Checks if the enhanced debug mode is enabled.
+	 * Checks if the verbose debug mode is enabled.
 	 *
 	 * @return bool True if enabled, false otherwise.
 	 */
-	public static function is_enhanced_debug_mode_enabled(): bool {
+	public static function is_verbose_debug_mode_enabled(): bool {
 		/**
-		 * Filters the flag that decides if the enhanced debug mode is enabled.
+		 * Filters the flag that decides if the verbose debug mode is enabled.
 		 *
 		 * @since 10.1.0
 		 *
@@ -2060,6 +2060,6 @@ class WC_Stripe_Helper {
 		 *
 		 * @return bool True if enabled, false otherwise.
 		*/
-		return apply_filters( 'wc_stripe_enhanced_debug_mode_enabled', false );
+		return apply_filters( 'wc_stripe_is_verbose_debug_mode_enabled', false );
 	}
 }
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -2045,7 +2045,21 @@ class WC_Stripe_Helper {
 		return WC_Stripe_UPE_Payment_Gateway::ID === substr( (string) $order->get_payment_method(), 0, 6 );
 	}
 
-	public static function is_enhanced_debug_mode_enabled() {
+	/**
+	 * Checks if the enhanced debug mode is enabled.
+	 *
+	 * @return bool True if enabled, false otherwise.
+	 */
+	public static function is_enhanced_debug_mode_enabled(): bool {
+		/**
+		 * Filters the flag that decides if the enhanced debug mode is enabled.
+		 *
+		 * @since 10.1.0
+		 *
+		 * @param bool $enabled True if enabled, false otherwise.
+		 *
+		 * @return bool True if enabled, false otherwise.
+		*/
 		return apply_filters( 'wc_stripe_enhanced_debug_mode_enabled', false );
 	}
 }
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -2044,4 +2044,8 @@ class WC_Stripe_Helper {
 	public static function is_stripe_gateway_order( $order ) {
 		return WC_Stripe_UPE_Payment_Gateway::ID === substr( (string) $order->get_payment_method(), 0, 6 );
 	}
+
+	public static function is_enhanced_debug_mode_enabled() {
+		return apply_filters( 'wc_stripe_enhanced_debug_mode_enabled', false );
+	}
 }
--- a/includes/class-wc-stripe-logger.php
+++ b/includes/class-wc-stripe-logger.php
@@ -244,6 +244,10 @@ class WC_Stripe_Logger {
 	 * @return boolean
 	 */
 	public static function can_log(): bool {
+		if ( WC_Stripe_Helper::is_enhanced_debug_mode_enabled() ) {
+			return true;
+		}
+
 		$settings = WC_Stripe_Helper::get_stripe_settings();
 
 		if ( empty( $settings ) || ( isset( $settings['logging'] ) && 'yes' !== $settings['logging'] ) ) {
--- a/includes/class-wc-stripe-logger.php
+++ b/includes/class-wc-stripe-logger.php
@@ -244,7 +244,7 @@ class WC_Stripe_Logger {
 	 * @return boolean
 	 */
 	public static function can_log(): bool {
-		if ( WC_Stripe_Helper::is_enhanced_debug_mode_enabled() ) {
+		if ( WC_Stripe_Helper::is_verbose_debug_mode_enabled() ) {
 			return true;
 		}
 
--- a/includes/connect/class-wc-stripe-connect-api.php
+++ b/includes/connect/class-wc-stripe-connect-api.php
@@ -182,7 +182,7 @@ if ( ! class_exists( 'WC_Stripe_Connect_API' ) ) {
 			if ( WC_Stripe_Helper::is_enhanced_debug_mode_enabled() ) {
 				// Log the request after the filters have been applied.
 				WC_Stripe_Logger::debug(
-					"WCC API request: {$method} {$path}",
+					"OAuth: WCC API request: {$method} {$path}",
 					[
 						'current_stripe_api_key' => WC_Stripe_API::get_masked_secret_key(),
 						'url'                    => $url,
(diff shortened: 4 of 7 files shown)


includes/class-wc-stripe-helper.php
```diff
@@ -2044,4 +2044,8 @@ public static function is_connected( $mode = null ) {
 	public static function is_stripe_gateway_order( $order ) {
 		return WC_Stripe_UPE_Payment_Gateway::ID === substr( (string) $order->get_payment_method(), 0, 6 );
 	}
+
+	public static function is_enhanced_debug_mode_enabled() {
```
wjrosa: The method is missing the description, parameters and return type.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -396,5 +496,78 @@ public function refresh_connection() {
...
+				'_wpnonce',
+				'state',
+				'code',
+				'secretKey',
+				'refreshToken',
+			];
+
+			if ( is_object( $data ) ) {
+				// Handle objects (stdClass) by converting to array, processing, and returning as array
+				return self::redact_sensitive_data( (array) $data );
```
wjrosa: I think we could replace this call with a simple cast:
```
$data = (array) $data;
```

includes/connect/class-wc-stripe-connect.php
```diff
@@ -84,12 +96,38 @@ public function connect_oauth( $state, $code, $type = 'connect', $mode = 'live'
...
 			if ( get_transient( 'wcs_stripe_connect_state_' . $mode ) !== $state ) {
-				return new WP_Error( 'Invalid state received from Stripe server' );
+				if ( WC_Stripe_Helper::is_enhanced_debug_mode_enabled() ) {
+					WC_Stripe_Logger::error(
+						'OAuth: Invalid state received from the WCC server',
+						[
+							'current_stripe_api_key' => WC_Stripe_API::get_masked_secret_key(),
+							'connect_mode'           => $mode,
+							'connect_type'           => $type,
+							'state'                  => self::redact_string( $state ),
```
wjrosa, line 106: Is it necessary to redact `state`, `code`, and `connect_mode`?

diegocurbelo: `connect_mode` no, it's just `test` or `live`.

`state` and `code`, on the other hand, yes:
- The `state` can be used to craft a "custom" return_url associated with another Stripe account; it needs the attacker also to have access to the `_wpnonce` (which we are also redacting) and to trick the merchant to click the link. Still, as a precaution, it's better to redact it.
- `code` in theory is safe, as it has already been used and can only be used once. But again, it's better to redact it.

includes/connect/class-wc-stripe-connect.php
diegocurbelo: Sounds good, added in [82bd658](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4765/commits/82bd6582b95bcea6d0718db13bf2e744e2dd0fbe)

wjrosa (approved):
Change looks good to me and works as expected!

> 6. Check that the new log entries (OAuth: *) are being logged:

Image

> 7. Check that no sensitive information is being logged:

Image

includes/connect/class-wc-stripe-connect-api.php
```diff
@@ -171,18 +171,44 @@ protected function request( $method, $path, $body = [] ) {
...
+
+			if ( WC_Stripe_Helper::is_enhanced_debug_mode_enabled() ) {
+				// Log the request after the filters have been applied.
+				WC_Stripe_Logger::debug(
+					"OAuth: WCC API request: {$method} {$path}",
+					[
+						'current_stripe_api_key' => WC_Stripe_API::get_masked_secret_key(),
+						'url'                    => $url,
+						'headers'                => $headers,
+						'body'                   => WC_Stripe_Connect::redact_sensitive_data( json_decode( $body ) ),
```
daledupreez, line 190: Do we also want to log the value of `$args`? I am not sure if we do, but it might be helpful.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -396,5 +496,82 @@ public function refresh_connection() {
...
+			$len = strlen( $string );
+			if ( $len > 15 ) {
+				return substr( $string, 0, 8 ) . '...' . substr( $string, -6 );
+			}
+			if ( $len > 9 ) {
+				// This applies only to wponces.
+				return substr( $string, 0, 3 ) . '...' . substr( $string, -3 );
+			}
+
+			// This should never be the case, as the shortest strings are 10 chars long (wponces).
```
daledupreez: Nit: typo
```suggestion
			// This should never be the case, as the shortest strings are 10 chars long (wp nonces).
```

includes/class-wc-stripe-helper.php
```diff
@@ -2044,4 +2044,22 @@ public static function is_connected( $mode = null ) {
...
+		 * Filters the flag that decides if the enhanced debug mode is enabled.
+		 *
+		 * @since 10.1.0
+		 *
+		 * @param bool $enabled True if enabled, false otherwise.
+		 *
+		 * @return bool True if enabled, false otherwise.
+		*/
+		return apply_filters( 'wc_stripe_enhanced_debug_mode_enabled', false );
+	}
```
daledupreez, line 2064: From a naming perspective, I think it would be better if we use a more specific name relating to OAuth or connections, as "enhanced debug mode" is pretty generic, and I would think that we'd generally want to continue using specific flags in future cases where we want to enable detailed debugging related to a specific feature or area of the code. So I would vote for `is_connection_debug_mode_enabled()` or `is_oauth_debug_mode_enabled()` for the method name, and then carry that through to the filter name.

changelog.txt
```diff
@@ -16,9 +16,6 @@
 * Fix - Increase limit when listing available payment method configurations from the Stripe API
 * Fix - Klarna not processing recurring payments
 * Fix - Fix Express Checkout error with free trial subscription on blocks cart/checkout
-* Fix - Prevent retrying requests that errored out due to declined payment methods
-* Fix - Detect WooCommerce Subscriptions staging sites when checking if payments can be detached
-* Fix - Fix saved ACH payment methods sending unsupported capture_method parameter causing checkout failures
 * Dev - Add Stripe's masked API key to API request/response logs
```
daledupreez, line 30: This matches the changelog entry for https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4761 -- we should update this to be more accurate.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -146,6 +222,18 @@ public function maybe_handle_redirect() {
 		 * @return stdObject|WP_Error OAuth's response result or WP_Error.
 		 */
 		private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' ) {
+			if ( WC_Stripe_Helper::is_enhanced_debug_mode_enabled() ) {
```
daledupreez: Is it worth stashing this result in a local variable so we're only applying the filters once per method?
```suggestion
			$enhanced_debugging_enabled = WC_Stripe_Helper::is_enhanced_debug_mode_enabled();
			if ( $enhanced_debugging_enabled ) {
```

includes/class-wc-stripe-logger.php
```diff
@@ -244,6 +244,10 @@ public static function debug( $message, $context = [] ) {
 	 * @return boolean
 	 */
 	public static function can_log(): bool {
+		if ( WC_Stripe_Helper::is_enhanced_debug_mode_enabled() ) {
```
daledupreez: Hmm... I now see that we're adding a general enhanced logging flag rather than something specific to OAuth. Might it be worth naming this something related to verbose logging rather than enhanced logging?

daledupreez (commented):
This is working as described, but I think we do need to fix the changelog entry, and I am still not sure if I like a blanket filter to bump up logging across the plugin.

includes/connect/class-wc-stripe-connect-api.php
diegocurbelo: It might be helpful if the args filter (which is deprecated) is used, but it makes the code to redact the data inside args very complex (because the body is string encoded json)

I added `is_args_filtered` to the log to see if the deprecated filter is in use in [0249fcc](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4765/commits/0249fcc51ddfa6381aff4d88511487e0149ec080); what do you think?
Screenshot 2025-11-10 at 13 29 33

includes/class-wc-stripe-logger.php
diegocurbelo: `is_verbose_debug_mode_enabled` is a better fit 👍🏼 

Refactored in [34f649b](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4765/commits/34f649b92a645455eb2d4ef2967d3edf60a46c36)

includes/class-wc-stripe-helper.php
diegocurbelo: Refactored to `is_verbose_debug_mode_enabled` in [34f649b](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4765/commits/34f649b92a645455eb2d4ef2967d3edf60a46c36)

includes/connect/class-wc-stripe-connect.php
diegocurbelo: The methods are not called frequently, but it makes sense 👍🏼 

Changed in [e810053](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4765/commits/e810053b0a00cfc0c684f650932c013ad4d470ac).

changelog.txt
diegocurbelo: Added changelog entry in [8317f10](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4765/commits/8317f1040390bbaa6bf67ff7a81c73a96a8cb2d1):

```
* Dev - Add verbose debug logging mode to the OAuth connect flow
```

includes/connect/class-wc-stripe-connect-api.php
daledupreez: That's fair. Happy to at least know whether some filter exists.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -396,5 +499,82 @@ public function refresh_connection() {
...
+						'/([?&]' . preg_quote( $key, '/' ) . '=)([^&#]*)/i',
+						function ( $matches ) {
+							$value = $matches[2];
+							if ( strlen( $value ) > 0 ) {
+								return $matches[1] . self::redact_string( $value );
+							}
+							return $matches[0];
+						},
+						$data
+					);
```
daledupreez, line 553: Not blocking: Is it worth adding an early return if we don't have a URL or form payload so we're not running multiple unnecessary regular expression matches?
```suggestion
				if ( ! str_contains( $data, '?' ) && ! str_contains( $data, '&' ) ) {
					return $data;
				}
				foreach ( $sensitive_keys as $key ) {
					$data = preg_replace_callback(
						'/([?&]' . preg_quote( $key, '/' ) . '=)([^&#]*)/i',
						function ( $matches ) {
							$value = $matches[2];
							if ( strlen( $value ) > 0 ) {
								return $matches[1] . self::redact_string( $value );
							}
							return $matches[0];
						},
						$data
					);
```

daledupreez (approved):
This is testing well for me! I think this is ready to ship.

includes/connect/class-wc-stripe-connect.php
diegocurbelo: It's also be used to redact the POST body (`application/x-www-form-urlencoded`) in the call to exchange the auth code, which sends only the `code` param (no `?` or `&`)
```
code=a188926...6f4a3b
```

PR #4766, opened 2025-11-06
Fix shipping method integration error when there are more than 9 shipping options

malithsen:
Towards STRIPE-749

When there are 10 or more shipping options in the shipping options array, Stripe SDK throws the following error and the express checkout element never opens

```
Uncaught (in promise) IntegrationError: Invalid value for shippingaddresschange event resolve callback: shippingRates should be an array of size less than 10. 
```

WooPayments gets around this by truncating the shipping options array (https://github.com/Automattic/woocommerce-payments/pull/10491). This PR proposes the same solution to Stripe. While this fix is not ideal, as merchants don't have control over which options are cut off, it prevents ECE from breaking completely.

## Testing instructions
1. Add more than 10 shipping options. You can do this by installing USPS and UPS extensions or simply creating 10 or more local pickup options.
2. Go to a page where Google Pay / Apple Pay is rendered.
3. Click the Google Pay/ Apple Pay button.
4. In `develop`, the payment modal will not open. On this branch, the payment modal should open, and the shipping options list should have 9 options.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -17,6 +17,7 @@
 * Fix - Klarna not processing recurring payments
 * Fix - Fix Express Checkout error with free trial subscription on blocks cart/checkout
 * Fix - Prevent retrying requests that errored out due to declined payment methods
+* Fix - GooglePay/ApplePay fail when there are more than 9 shipping options
 
 = 10.0.1 - 2025-10-15 =
 * Fix - Remove persistent reconnection notices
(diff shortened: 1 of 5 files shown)


diegocurbelo (approved):
The change looks good and makes sense 🚢

daledupreez (approved):
Thanks for fixing this! I confirmed that it was failing for Apple Pay, Google Pay, and Amazon Pay, and then confirmed that all three are working with this patch.

PR #4767, opened 2025-11-07
Fix saved checkout failures with saved ACH payment methods

malithsen:
Fixes STRIPE-796

This PR fixes a bug where saved ACH payment methods fail during WooCommerce Blocks checkout with the error: `capture_method=manual is not supported by payment method type us_bank_account`.

### Root Cause
When using saved ACH payment methods in Blocks checkout, WooCommerce Blocks doesn't set the `isSavedToken` flag for split UPE gateways (it only sets it for the primary `stripe` gateway). This caused the server-side blocks support to miss the saved token detection, resulting in the payment going through the wrong flow (`process_payment_with_saved_payment_method` instead of `process_payment_with_deferred_intent`). The wrong flow tried to set `capture_method=manual`, which ACH doesn't support.

- Updated the saved token detection in `WC_Stripe_Blocks_Support::add_payment_request_order_meta()` to check for the `wc-{gateway_id}-payment-token` field when `isSavedToken` is not set. This ensures saved ACH (and other split UPE gateway) payments are properly routed through the deferred intent flow.

- As a secondary defensive fix, added a check in `WC_Stripe_UPE_Payment_Gateway::process_payment_with_saved_payment_method()` to only set `capture_method` when `$request_details['capture']` exists.

## Testing instructions

1. As a logged in customer, add a product to the cart, go to blocks checkout.
2. Place an order using ACH, saving the payment method.
3. Try to place another order using the saved ACH payment method.
4. In **develop**, clicking place order will enter the site into a loop. In logs, there will be repeated error entries. On this branch, you should be able to place the order without any errors.

**Optional tests**
Place an order using:
- a new card payment method.
- a new ACH payment method
- saved card payment method.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-blocks-support.php
+++ b/includes/class-wc-stripe-blocks-support.php
@@ -459,7 +459,19 @@ final class WC_Stripe_Blocks_Support extends AbstractPaymentMethodType {
 		 * When using UPE on the block checkout and a saved token is being used, we need to set a flag
 		 * to indicate that deferred intent should be used.
 		 */
-		if ( $is_upe && isset( $data['issavedtoken'] ) && $data['issavedtoken'] ) {
+		$is_using_saved_token = isset( $data['issavedtoken'] ) && $data['issavedtoken'];
+
+		// For split UPE gateways (e.g., stripe_us_bank_account), WooCommerce Blocks doesn't set the isSavedToken flag.
+		// Check if a payment token is being used by looking for the wc-{gateway_id}-payment-token field.
+		if ( ! $is_using_saved_token && ! empty( $data['token'] ) ) {
+			// Payment data keys use underscores, not hyphens (e.g., wc-stripe_us_bank_account-payment-token).
+			$token_key = 'wc-' . $context->payment_method . '-payment-token';
+			if ( isset( $data[ $token_key ] ) && ! empty( $data[ $token_key ] ) ) {
+				$is_using_saved_token = true;
+			}
+		}
+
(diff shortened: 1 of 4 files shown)


includes/class-wc-stripe-blocks-support.php
```diff
@@ -459,7 +459,19 @@ public function add_payment_request_order_meta( PaymentContext $context, Payment
...
+		// Check if a payment token is being used by looking for the wc-{gateway_id}-payment-token field.
+		if ( ! $is_using_saved_token && ! empty( $data['token'] ) ) {
+			// Payment data keys use underscores, not hyphens (e.g., wc-stripe_us_bank_account-payment-token).
+			$token_key = 'wc-' . $context->payment_method . '-payment-token';
+			if ( isset( $data[ $token_key ] ) && ! empty( $data[ $token_key ] ) ) {
+				$is_using_saved_token = true;
+			}
+		}
+
+		if ( $is_upe && $is_using_saved_token ) {
```
wjrosa, line 474: Nit: `$is_upe` is now always `true`. We can remove it, but feel free to keep it here now, and we can do it later in a cleaning-up PR.

wjrosa (approved):
Nice fix! Looks good and works as expected:

> On this branch, you should be able to place the order without any errors.

Image

diegocurbelo (approved):
The change looks good, and is working as described.

Issue #4750, opened 2025-10-24
Amazon Pay has mismatched taxes when taxes are based on billing address


daledupreez:
### Description

When a store has taxes enabled and being computed based on billing address and Amazon Pay is enabled, we are hitting a situation where we only receive the shopper's billing address after they have confirmed the payment amount, but that may result in a different tax amount being applied during checkout as a result of the billing address being set. When the amount changes like this, the Amazon Pay payment is declined.

In the initial report, the symptom was the creation of a payment intent that didn't include any taxes for the express checkout, but that was one manifestation driven by the specific tax configurations involved.

### Steps to reproduce

The issue here is partially driven by your store configuration, specifically around taxes for shopper locations and your billing address in Amazon Pay (which will drive which tax location you're in), as well as whether we have a billing address for the current user, either in the current session, or in the customer database. To get the best reproduction, it's best to test from a completely new, logged-out session, as we want to have no billing details set up for the user.

 1. Enable the Amazon Pay feature flag - `wp option update _wcstripe_feature_amazon_pay 'yes'`
 2. Configure your store to include taxes for purchases.
 3. Configure the tax to be driven by the customer billing address.
    * **Note:** this step was missing in my initial report.
 4. Enable Amazon Pay as an express checkout method, and ensure that it is displayed on the product page
 5. Ensure you have a taxable product in your store.
 6. Go to the product page for your product while in a new, logged-out browser session.
 7. Click on the Amazon Pay express checkout button.
 8. Log into the Amazon Pay sandbox or create a new sandbox account.
 9. Pick or create a supported shipping address.
10. Pick a valid card.
11. Note the total includes taxes for the item.
12. Complete the checkout.
13. Note that the payment fails. If you then review the logs, you can see that we're creating a payment intent for the item total plus shipping, but we are sending a different amount for the  taxes.

### Expected behavior

The payment should include the correct taxes, and it should be processed successfully.

### Additional context

wjrosa:
As we discussed earlier, Amazon Pay is working for me with the setup above. I would appreciate if you can share your specific setup information so I can try to reproduce the issue again when you have the chance.

daledupreez:
Thanks for testing, @wjrosa! As we worked out, we also need taxes to be based on the customer billing address. I've updated the steps above to reflect that by adding what is now step 3.

wjrosa:
[...]

PR #4768, opened 2025-11-07
Explore post-confirm price recalculation for Amazon Pay

daledupreez:
Towards STRIPE-757
Towards https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4750

This PR explores adding some special logic for Amazon Pay when the store is configured to calculate taxes based on billing address. The core issue is that we are getting the shopper's address in the `confirm` event, and are then initiating checkout immediately. However, depending on the store's tax setup, this may cause the taxes in checkout to be different from the taxes we sent to Amazon which the user agreed to, as the user's tax rates may change. This results in the payment being declined.

To try and address that problem, this PR does the following:
 * When we receive the `confirm` event for an express checkout method, we check if taxes are based on billing address and we have an Amazon Pay payment.
    - If we are in that situation, we then extract the billing address from the event payload, pass that address through our standard express checkout address normalisation API, and then update the customer's billing address (which may write to the current session or the persistent customer data), the server returns an update cart, and we use the total amount from the returned cart to update the express checkout element.

This is not working at present, though it is updating the amount of the intent when we update the amount.
 

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/api/index.js
+++ b/client/api/index.js
@@ -537,6 +537,19 @@ export default class WCStripeAPI {
 	}
 
 	/**
+	 * Update cart customer data.
+	 *
+	 * @param {Object} customerData Customer data.
+	 * @return {Promise} Promise for the request to the server.
+	 */
+	expressCheckoutUpdateCartCustomer( customerData ) {
+		return this.postToBlocksAPI(
+			'/wc/store/v1/cart/update-customer',
+			customerData
+		);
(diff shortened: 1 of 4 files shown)


daledupreez:
Closing this exploration.

PR #4772, opened 2025-11-07
Fix fatal error related to the safe_style_css filter

wjrosa:

We are getting some fatal errors in Grafana related to the inclusion of the `display` CSS property as a safe style using the `safe_style_css` filter (`includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#290`):
```
PHP Fatal error:  Uncaught TypeError: array_merge(): Argument #1 must be of type array, null given in /wordpress/plugins/woocommerce-gateway-stripe/10.0.1/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php:295
Stack trace:
```

While the root cause is not known (probably due to another filter overriding the default value), the change in this PR is pretty safe, as it just checks if the default style is an array.

## Testing instructions

Code review should be enough.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.1.0 - xxxx-xx-xx =
+* Fix - Fix fatal error when trying to allow the `display` CSS property using the `safe_style_css` filter
 * Update - Renames and migrates all Payment Request Buttons settings to Express Checkout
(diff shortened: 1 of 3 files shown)


diegocurbelo (approved):
Makes sense, according to the [docs](https://developer.wordpress.org/reference/hooks/safe_style_css/) `$styles` is expected to be a `string[]`

malithsen (approved):
Thanks for the fix. LGTM!

PR #4773, opened 2025-11-08
Don't show Amazon Pay when taxes are based on billing address

daledupreez:
Related to STRIPE-757
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4750

This PR ensures that Amazon Pay will not be shown to shoppers when WooCommerce is configured to use shoppers' billing address to compute taxes on orders. The reasons for this are because we only get shoppers' billing address after they have approved a specific amount in Amazon, which may lead to a change in tax computations on our side, and thus a different purchase amount that will be declined.

While we are trying to work around that situation, we still want to offer Amazon Pay to other merchants. So this PR does the following:
 * We add data to the client configuration fields to represent when taxes are enabled and tax calculations are based on the shopper's billing address
 * The express checkout display logic in the client now requires that flag to be falsey for Amazon Pay to be shown (in addition to Amazon Pay being enabled)
 * When taxes are enabled and based on billing address _and_ Amazon Pay is enabled, we now show notices in the main payment method settings page and the Amazon Pay customisation page to tell merchants that Amazon Pay will not be shown to shoppers, and why.

#### Notice in payment method listing screen

Screenshot 2025-11-11 at 18 27 51

#### Notice in Amazon Pay customisation screen

Screenshot 2025-11-11 at 18 28 12

## Testing instructions

* Set up a test store using the code from this branch and connect to Stripe
* For now, ensure that the store is either configured not to compute taxes, or is set to calculate taxes based on the customer shipping address.
  - The overall flag to enable tax handling is on _WooCommerce_ > _Settings_ > _General_ via the "Enable taxes" checkbox
  - The control for which location to us is available from _WooCommerce_ > _Settings_ > _Tax_ via the "Calculate tax based on" dropdown.
* Enable the Amazon Pay feature flag: `wp option update _wcstripe_feature_amazon_pay yes`
* Navigate to the Stripe settings screen
* Enable Amazon Pay as an express checkout method
* Confirm that you don't see a notice regarding Amazon Pay and the billing address.
* Click on the "Customize" link for Amazon Pay
* Confirm that you don't see a notice regarding Amazon Pay and the billing address.
* Navigate to a screen/context where Amazon Pay (and other express checkout methods) should be shown to shoppers.
* Verify that Amazon Pay is displayed. (Keep this tab/window open.)
* Open a new tab/window, and enable tax (via _WooCommerce_ > _Settings_ > _General_ and then "Enable taxes") and change the logic to use the billing address (via _WooCommerce_ > _Settings_ > _Tax_ and then "Calculate tax based on").
* Reload the shopper-facing view.
* Verify that Amazon Pay is not shown to the shopper.
* Reload the tab looking at the customisation screen for Amazon Pay.
* Verify that you see the Amazon Pay notice, and it includes a link to the tax settings page.
* Click on the link so it opens in a new tab, and verify that it takes you to the tax settings page. Close the tab (or keep it open if it will be handy!)
* Disable Amazon Pay.
* Verify that you no longer see the Amazon Pay notice.
* Re-enable Amazon Pay.
* Verify that you see the notice again.
* Reload the shopper-facing view.
* Verify that Amazon Pay is not shown to the shopper.
* Now go back to the main payment method listing.
* Verify that you see the Amazon Pay notice, without a link to update the tax settings.
* Disable Amazon Pay.
* Verify that the notice is no longer shown.
* Re-enable Amazon Pay and verify that the notice is shown again.
* Disable taxes for the store.
* Verify that the notice now longer shows after reloading the page.
* Reload the shopper-facing view.
* Verify that Amazon Pay is shown to the shopper.
* Re-enable taxes.
* Verify that the notice appears.
* Now modify the tax calculation approach for the store, and verify that the notice is no longer shown.
* Reload the shopper-facing view.
* Verify that Amazon Pay is shown to the shopper.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/components/amazon-pay-taxes-billing-address-notice/index.js
+++ b/client/components/amazon-pay-taxes-billing-address-notice/index.js
@@ -1,0 +1,52 @@
+import { getAdminLink } from '@woocommerce/settings';
+import React from 'react';
+import { Notice } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { useAmazonPayEnabledSettings } from 'wcstripe/data';
+
+const AmazonPayTaxesBillingAddressNotice = ( {
+	areTaxesBasedOnBillingAddress = false,
+	noticeStatus = 'error',
+	showUpdateSettingsLink = true,
+} ) => {
+	const [ isAmazonPayEnabled ] = useAmazonPayEnabledSettings();
+
+	if ( ! isAmazonPayEnabled ) {
+		return null;
+	}
+	if ( ! areTaxesBasedOnBillingAddress ) {
+		return null;
+	}
+
+	const actions = showUpdateSettingsLink
+		? [
+				{
+					label: __(
+						'Update tax settings',
+						'woocommerce-gateway-stripe'
+					),
+					url: getAdminLink( 'admin.php?page=wc-settings&tab=tax' ),
+					variant: 'secondary',
+				},
+		  ]
+		: [];
+
+	return (
+		<Notice
+			status={ noticeStatus }
+			isDismissible={ false }
+			actions={ actions }
+		>
+			<p>
+				<strong>
+					{ __(
+						'Amazon Pay does not support taxes based on the billing address. The checkout button will not be visible to shoppers with this setting in effect.',
+						'woocommerce-gateway-stripe'
+					) }
+				</strong>
+			</p>
+		</Notice>
+	);
+};
+
+export default AmazonPayTaxesBillingAddressNotice;
--- a/client/components/amazon-pay-taxes-billing-address-notice/index.js
+++ b/client/components/amazon-pay-taxes-billing-address-notice/index.js
@@ -1,13 +1,14 @@
 import { getAdminLink } from '@woocommerce/settings';
 import React from 'react';
+import interpolateComponents from '@automattic/interpolate-components';
 import { Notice } from '@wordpress/components';
 import { __ } from '@wordpress/i18n';
 import { useAmazonPayEnabledSettings } from 'wcstripe/data';
 
+import './style.scss';
+
 const AmazonPayTaxesBillingAddressNotice = ( {
 	areTaxesBasedOnBillingAddress = false,
-	noticeStatus = 'error',
-	showUpdateSettingsLink = true,
 } ) => {
 	const [ isAmazonPayEnabled ] = useAmazonPayEnabledSettings();
 
@@ -18,33 +19,30 @@ const AmazonPayTaxesBillingAddressNotice = ( {
 		return null;
 	}
 
-	const actions = showUpdateSettingsLink
-		? [
-				{
-					label: __(
-						'Update tax settings',
-						'woocommerce-gateway-stripe'
-					),
-					url: getAdminLink( 'admin.php?page=wc-settings&tab=tax' ),
-					variant: 'secondary',
-				},
-		  ]
-		: [];
+	const actions = [
+		{
+			label: __( 'Update tax settings', 'woocommerce-gateway-stripe' ),
+			url: getAdminLink( 'admin.php?page=wc-settings&tab=tax' ),
+			variant: 'secondary',
+		},
+	];
 
 	return (
 		<Notice
-			status={ noticeStatus }
+			className="wc-stripe-amazon-pay-taxes-billing-address-notice"
+			status="error"
 			isDismissible={ false }
 			actions={ actions }
 		>
-			<p>
-				<strong>
-					{ __(
-						'Amazon Pay does not support taxes based on the billing address. The checkout button will not be visible to shoppers with this setting in effect.',
-						'woocommerce-gateway-stripe'
-					) }
-				</strong>
-			</p>
+			{ interpolateComponents( {
+				mixedString: __(
+					'{{strong}}Amazon Pay does not support taxes based on the billing address.{{/strong}} The checkout button will not be visible to shoppers with this setting in effect.',
+					'woocommerce-gateway-stripe'
+				),
+				components: {
+					strong: <strong />,
+				},
+			} ) }
 		</Notice>
 	);
 };
--- a/client/entrypoints/express-checkout/index.js
+++ b/client/entrypoints/express-checkout/index.js
@@ -219,6 +219,9 @@ jQuery( function ( $ ) {
 					?.is_amazon_pay_enabled;
 			const isLinkEnabled =
 				wc_stripe_express_checkout_params?.stripe?.is_link_enabled; // eslint-disable-line camelcase
+			const areTaxesBasedOnBillingAddress = getExpressCheckoutData(
+				'taxes_based_on_billing'
+			);
 
 			// For each supported express payment type, create their own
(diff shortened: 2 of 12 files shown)


wjrosa (approved):
Nice job! Looks good to me and tests well:

Image

As an alternative approach, we could have added a new "pill" similar to the other payment methods ("Requires tax setting" or a similar wording). And reuse that in the other express methods.

client/settings/payment-methods/index.js
```diff
@@ -48,6 +49,20 @@ const PaymentRequestDescription = () => (
...
 
+const AmazonPayTaxesBasedOnBillingAddressSection = () => {
+	const areTaxesBasedOnBillingAddress =
+		!! wc_stripe_settings_params?.taxes_based_on_billing; // eslint-disable-line camelcase
+
+	return (
+		<SettingsSection>
+			<AmazonPayTaxesBillingAddressNotice
+				areTaxesBasedOnBillingAddress={ areTaxesBasedOnBillingAddress }
+				showUpdateSettingsLink={ false }
```
frosso: Why show the "update tax settings" link only in the Amazon Pay-specific screen? Why not show the link/action in both the "main" screen and in the Amazon Pay one?

client/entrypoints/express-checkout/index.js
```diff
@@ -219,6 +219,9 @@ jQuery( function ( $ ) {
 					?.is_amazon_pay_enabled;
 			const isLinkEnabled =
 				wc_stripe_express_checkout_params?.stripe?.is_link_enabled; // eslint-disable-line camelcase
+			const areTaxesBasedOnBillingAddress = getExpressCheckoutData(
```
frosso, line 222: (minor) Rather than duplicating this check on the front-end and passing another `taxes_based_on_billing` flag from the back-end, what do you think of piggybacking on the pre-existing `isAmazonPayEnabled` flag? In the back-end, the `express_checkout_helper->is_amazon_pay_enabled()` could check on the "taxes based on billing" flag.
This way, both the block-based and shortcode-based scripts won't need to be updated in case the requirements change further.

It makes sense, instead, for the admin pages to consider both options/flags separate.

client/components/amazon-pay-taxes-billing-address-notice/index.js
```diff
@@ -0,0 +1,52 @@
...
+			actions={ actions }
+		>
+			<p>
+				<strong>
+					{ __(
+						'Amazon Pay does not support taxes based on the billing address. The checkout button will not be visible to shoppers with this setting in effect.',
+						'woocommerce-gateway-stripe'
+					) }
+				</strong>
+			</p>
```
frosso: I noticed that the `p` tag is adding some padding. The spacing makes this notice feel inconsistent with the other notices on the app, especially when they're displayed on the same page:

Image
Image
Image
Image

I also wonder about the necessity of the `strong` tag - the background color should already be capturing the attention enough?

client/components/amazon-pay-taxes-billing-address-notice/index.js
```diff
@@ -0,0 +1,52 @@
+import { getAdminLink } from '@woocommerce/settings';
+import React from 'react';
+import { Notice } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { useAmazonPayEnabledSettings } from 'wcstripe/data';
+
+const AmazonPayTaxesBillingAddressNotice = ( {
+	areTaxesBasedOnBillingAddress = false,
+	noticeStatus = 'error',
```
frosso: Is there a scenario where `noticeStatus` should be something different than `error`? What do you think about just passing `status="error"` to the `Notice` component below and removing this prop?

frosso (commented):
Tests well, I like the additions of the tests. My main suggestions are around making the notice more consistent within the UI of the settings and removing a seemingly unused `noticeStatus` prop. 
The eligibility logic for the customer-facing payment component can be controversial :)

client/settings/payment-methods/index.js
daledupreez: I am happy to make this visible in both situations. I wasn't sure that it was worth taking up the real estate in the main payment method list UI. But it's not a lot extra.

client/entrypoints/express-checkout/index.js
daledupreez: Pushing this into the general enabled check feels like it would be a bit hard to untangle and could lead to unexpected behaviour down the road if/when we remove the restriction, as that flag will suddenly change behaviour in a way that may have future code not written to expect the tax situation to be a factor in the behaviour. Pushing it down into that code feels like it makes it more implicit.

client/components/amazon-pay-taxes-billing-address-notice/index.js
daledupreez: Good call. I've removed `noticeStatus` in 69d6cee8c9fb6a7d8a5ad8d2c23d78f08fb9f0bc.

client/components/amazon-pay-taxes-billing-address-notice/index.js
daledupreez: This is really good feedback. I've made the following changes in 69d6cee8c9fb6a7d8a5ad8d2c23d78f08fb9f0bc:
 * Only apply `<strong>` to the first sentence in the notice, as that is really important for folks to know.
 * Removed the wrapping `<p>` from the contents
 * Always show the CTA
 * Add some CSS to ensure we have enough space between the text and the action

daledupreez:
> As an alternative approach, we could have added a new "pill" similar to the other payment methods ("Requires tax setting" or a similar wording). And reuse that in the other express methods.

@wjrosa, good call. I think it may make sense to tackle both:
 * If we get into a situation where Amazon Pay is enabled and taxes are based on billing address, we show the notice.
 * When taxes are based on billing address, and Amazon Pay is not enabled yet, add a pill and disable the checkbox.

I'd like to tackle the better UX in a follow-up PR.

client/entrypoints/express-checkout/index.js
frosso: That's fair. However, consider the scenario where only Amazon Pay is enabled in the settings (Google Pay/Apple Pay & Stripe Link are disabled).
The store's tax is also based on the customer's billing address, so we know that Amazon Pay is never going to be displayed.

In this scenario, the `/wp-content/plugins/woocommerce-gateway-stripe/build/express-checkout.js` (and CSS) will always be queued, parsed, and executed for each customer, despite them not being necessary.

frosso (approved):
Nothing else additional from me, thank you for entertaining my suggestions!

Image

client/entrypoints/express-checkout/index.js
daledupreez: I created a follow-up issue for this: https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4783

daledupreez:
I also created https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4782 to improve the initial UX so we don't enable Amazon Pay when it won't work.

PR #4785, opened 2025-11-12
Allow invalid payment methods to be disabled

### LINKED ISSUE #4711: Cannot disable payment method that is enabled in Stripe dashboard
### Description

After the release of 9.9.2, we have a number of merchants reporting unsupported payment methods being enabled and shown as available in checkout, but they cannot disable the payment method from the admin settings UI because the controls are disabled.

The following forum threads report similar issues:
 * https://wordpress.org/support/topic/how-do-i-disable-blik-payments/ - EUR store, where BLIK is enabled but not modifiable because BLIK only supports PLN
 * https://wordpress.org/support/topic/blik-is-enabled-it-requires-store-currency-to-be-set-to-pln-other-methods/ - multiple merchants are seeing BLIK enabled despite transacting in EUR

### Steps to reproduce

Set up a store and use the Stripe dashboard to enable an unsupported payment method. This new payment method should be returned to the store, and then incorrectly shown as enabled.

### Expected behavior

The expected behaviour is not super clear cut. There are multiple ways we could address this:
* We could filter out unsupported payment methods in the store itself -- this should result in no unexpected changes to the upstream PMC in Stripe
   - We may need some way to indicate this in the settings UI, but that would not be needed in a first pass
* We could make it possible to disable these unsupported payment methods via the settings UI
   - The downside of this approach is that we'd then be modifying the PMC. Given the changes in 9.9.2 to allow more PMCs to be used, the changes here might not be something merchants would want, depending on context.

There are possibly other options that we'd want to consider, too.

### System report

<details>

</details>

### Additional context

These issues started happening with 9.9.2, almost certainly as a result of the changes from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4696 and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4698

daledupreez:
Fixes STRIPE-734
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4711

This PR updates our logic in the payment method list to make it possible for enabled payment methods to be disabled when they should not be available, as we currently don't allow merchants to fix that problem in the UI. We will not allow the payment method to be (re)enabled, but we do make it possible for the user to disable the payment method.

While this is a slightly weird UX, it allows merchants to fix their payment method configuration to align with their store and avoid any notices we might show them when they have a payment method enabled that we don't allow under the current store configuration.

I also added some initial unit tests for the `PaymentMethod` component.

## Testing instructions

* Check the unit tests to check they're testing the two existing cases, and the new case where we want the checkbox to be enabled when the payment method is enabled despite not being permitted.
* Run this branch in a test site.
* Connect to Stripe.
* Enable one or more payment methods that are only permitted in your current store currency.
   - If your store is using USD, you can enable payment methods like Affirm and ACH, which support USD but not EUR.
   - If your store is using EUR, you can enable payment methods like SEPA, Bancontact, and iDEAL, which only support EUR.
* Navigate to _WooCommerce_ > _Settings_ and change your store currency in such a way that at least some of the payment methods above are incompatible with the new currency.
* Open the Stripe payment method list.
* Verify that the previously enabled payment methods show the expected pill about the currency _and_ are enabled.
* Verify that you can disable one or more of the payment methods, and can save the result.
* After disabling each payment method, verify that the checkbox is disabled and does not permit re-enabling the payment method.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Fix - Allow payment methods to be disabled when they are not available
 
 = 10.1.0 - 2025-11-11 =
 * Dev - Remove unused `shouldShowPaymentRequestButton` parameter and calculations from backend
--- a/client/settings/general-settings-section/__tests__/payment-method.test.js
+++ b/client/settings/general-settings-section/__tests__/payment-method.test.js
@@ -1,0 +1,176 @@
+import React from 'react';
+import { screen, render } from '@testing-library/react';
+import icons from '../../../payment-method-icons';
+import UpeToggleContext from '../../upe-toggle/context';
+import PaymentMethod from '../payment-method';
+import PaymentMethodDescription from '../payment-method-description';
+import { useEnabledPaymentMethodIds, useManualCapture } from 'wcstripe/data';
+import {
+	PAYMENT_METHOD_CARD,
+	PAYMENT_METHOD_SEPA,
+	PAYMENT_METHOD_UNAVAILABLE_REASONS,
+} from 'wcstripe/stripe-utils/constants';
+import usePaymentMethodUnavailableReason from 'utils/use-payment-method-unavailable-reason';
+import { getFormattedPaymentMethodDescription } from 'wcstripe/settings/general-settings-section/get-formatted-payment-method-description';
+
+jest.mock( '../payment-method-description' );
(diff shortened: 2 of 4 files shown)


wjrosa (approved):
Change makes sense to me and works as intended ✅

diegocurbelo (approved):
The code change makes sense, and it tests well!

PR #4784, opened 2025-11-11
Remove logic from settings controller for UPE being disabled

daledupreez:
## Changes proposed in this Pull Request:

This PR explores removing some logic from the settings controller that handles the case where UPE is disabled, as UPE can no longer be disabled.

## Testing instructions

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4787, opened 2025-11-12
Removing legacy payment method classes

wjrosa:

In this simple cleanup PR, I am removing all the legacy checkout payment method classes, as we are no longer using them.

## Testing instructions

Code review. Check if the tests are still passing. Perform some basic smoke testing and confirm no regression was introduced.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


malithsen (approved):
Looks good. Are we planning to do a similar cleanup for the frontend code (to remove `is_upe_enabled` distinction) in a separate PR?

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -605,22 +604,6 @@ private function update_enabled_payment_methods( $payment_method_ids_to_enable,
...
-					$gateway->update_option( 'enabled', 'yes' );
-				}
-			}
-
-			return;
-		}
-
 		if ( $this->gateway instanceof WC_Stripe_UPE_Payment_Gateway ) {
 			$this->gateway->update_enabled_payment_methods( $payment_method_ids_to_enable );
 		}
```
daledupreez: Is this condition needed? I don't think we create a non-UPE instance any more.
```suggestion
		$this->gateway->update_enabled_payment_methods( $payment_method_ids_to_enable );
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -186,21 +186,11 @@ public static function display_legacy_deprecation_notice( $plugin_file ) {
 	 *
 	 * @since 4.1.0
 	 * @return array
+	 *
+	 * @deprecated 10.1.0 This method will be removed in a future release.
```
daledupreez: Nit: This should be 10.2.0.
```suggestion
	 * @deprecated 10.2.0 This method will be removed in a future release.
```

includes/class-wc-stripe-helper.php
```diff
@@ -439,73 +439,33 @@ public static function get_settings( $method = null, $setting = null ) {
 	 * List of legacy payment method classes.
 	 *
 	 * @return array
+	 *
+	 * @deprecated 10.1.0 This method will be removed in future versions.
```
daledupreez: Nit: Another version fix.
```suggestion
	 * @deprecated 10.2.0 This method will be removed in future versions.
```

includes/class-wc-stripe-helper.php
```diff
@@ -439,73 +439,33 @@ public static function get_settings( $method = null, $setting = null ) {
...
-		return $payment_method_classes;
+		return [];
 	}
 
 	/**
 	 * List of legacy payment methods.
 	 *
 	 * @return array
+	 *
+	 * @deprecated 10.1.0 This method will be removed in future versions.
```
daledupreez: Nit: Another version fix.
```suggestion
	 * @deprecated 10.2.0 This method will be removed in future versions.
```

includes/class-wc-stripe-helper.php
```diff
@@ -439,73 +439,33 @@ public static function get_settings( $method = null, $setting = null ) {
...
-
 		return self::$stripe_legacy_gateways;
 	}
 
 	/**
 	 * Get legacy payment method by id.
 	 *
 	 * @return object|null
+	 *
+	 * @deprecated 10.1.0 This method will be removed in future versions.
```
daledupreez: Nit: Another version fix.
```suggestion
	 * @deprecated 10.2.0 This method will be removed in future versions.
```

includes/class-wc-stripe-helper.php
```diff
@@ -439,73 +439,33 @@ public static function get_settings( $method = null, $setting = null ) {
...
-				$gateway_id     = array_search( $payment_method_class, $payment_gateway_classes, true );
-				$payment_method = $payment_gateways[ $gateway_id ];
-			} else {
-				$payment_method = new $payment_method_class();
-			}
-
-			self::$stripe_legacy_gateways[ $payment_method->id ] = $payment_method;
-		}
-
 		return self::$stripe_legacy_gateways;
```
daledupreez: Can we just return `[]` here?
```suggestion
		return [];
```

includes/class-wc-stripe-helper.php
```diff
@@ -516,183 +476,33 @@ public static function get_legacy_payment_method( $id ) {
 	 * The ids are mapped to the corresponding equivalent UPE method ids for rendeing on the frontend.
 	 *
 	 * @return array
+	 *
+	 * @deprecated 10.1.0 This method will be removed in future versions.
```
daledupreez: Nit: Another version fix.
```suggestion
	 * @deprecated 10.2.0 This method will be removed in future versions.
```

includes/class-wc-stripe-helper.php
```diff
@@ -516,183 +476,33 @@ public static function get_legacy_payment_method( $id ) {
...
-		return $payment_method_ids;
+		return [];
 	}
 
 	/**
 	 * List of enabled legacy payment methods.
 	 *
 	 * @return array
+	 *
+	 * @deprecated 10.1.0 This method will be removed in future versions.
```
daledupreez: Nit: Another version fix.
```suggestion
	 * @deprecated 10.2.0 This method will be removed in future versions.
```

includes/class-wc-stripe-helper.php
```diff
@@ -516,183 +476,33 @@ public static function get_legacy_payment_method( $id ) {
...
-		return array_merge( $enabled_payment_method_ids, $mapped_enabled_payment_method_ids );
-	}
-
-	/**
-	 * Get settings of individual legacy payment methods.
 	 *
-	 * @return array
-	 *
-	 * @deprecated 9.6.0 The customization of individual payment methods is now deprecated.
+	 * @deprecated 10.1.0 This method will be removed in future versions.
```
daledupreez: Nit: Another version fix.
```suggestion
	 * @deprecated 10.2.0 This method will be removed in future versions.
```

includes/class-wc-stripe.php
```diff
@@ -704,23 +633,6 @@ protected function enable_upe( $settings ) {
 	protected function disable_upe( $settings ) {
```
daledupreez, line 637: Should we not remove this function altogether if UPE can't be disabled?

tests/phpunit/Admin/Migrations/WC_Stripe_Subscriptions_Repairer_Legacy_SEPA_Tokens_Test.php
```diff
@@ -68,7 +67,7 @@ class WC_Stripe_Subscriptions_Repairer_Legacy_SEPA_Tokens_Test extends WP_UnitTe
 	 *
 	 * @var string
 	 */
-	private $legacy_sepa_gateway_id = WC_Gateway_Stripe_Sepa::ID;
+	private $legacy_sepa_gateway_id = 'stripe_sepa';
```
daledupreez: Might it be worth using a new constant for this rather than having `'stripe_sepa'` everywhere?

It also feels like it could be worth pushing just that change into a dedicated PR, as that would be easy to inspect and review, and we could get it shipped with minimal risk. That would also reduce the size of this PR by quite a few files.

daledupreez (commented):
I have some initial comments from inspection, and I will move to test this later this morning. That said, it feels like it might be worth splitting this PR up a bit so we can tackle more specific and focused testing.

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -209,14 +209,13 @@ public function register_routes() {
 	 * @return WP_REST_Response
 	 */
 	public function get_settings() {
-		$is_upe_enabled = WC_Stripe_Feature_Flags::is_upe_checkout_enabled();
 		// When UPE and the payment method configurations API are enabled, fetch the enabled payment methods from the payment method configurations API.
 		// We force a refresh of the enabled payment methods (by passing true) when on the settings page to ensure the latest data.
 		// The available payment methods are also fetched from the payment method configurations API under similar conditions,
 		// but we do not force a refresh for available methods, since calling get_upe_enabled_payment_method_ids first already ensures the list is up to date.
-		$enabled_payment_method_ids   = $is_upe_enabled ? $this->gateway->get_upe_enabled_payment_method_ids( true ) : WC_Stripe_Helper::get_legacy_enabled_payment_method_ids();
```
daledupreez: As an FYI, I am also exploring some related cleanup in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4784. Some of these changes overlap, but I have some additional changes.

Where we have overlap, the changes are the same, so merging/resolving should be pretty simple, but I wanted to make sure we're all aware.

wjrosa: Thanks for letting me know. I also have many other related PRs to open (as I mentioned during our last team meeting). However, I am willing to resolve any conflicts that may arise as well 👍

includes/class-wc-stripe.php
wjrosa: I will remove it in one of my future PRs. I wanted to keep this PR focused on just removing the payment method classes. All the other changes I did here were necessary to avoid breaking the code

wjrosa:
> Looks good. Are we planning to do a similar cleanup for the frontend code (to remove is_upe_enabled distinction) in a separate PR?

@malithsen yes. I have some other branches to push related to code cleaning up. I will open other PRs in the next couple of weeks

> I have some initial comments from inspection, and I will move to test this later this morning. That said, it feels like it might be worth splitting this PR up a bit so we can tackle more specific and focused testing.

@daledupreez I tried to make this PR focused on just removing the payment method classes, to avoid it getting too big or risky. Which parts do you think would be better to move to another PR?

daledupreez:
For splitting the PR, I was thinking it might make sense to change the references to `WC_Gateway_Stripe_Sepa::ID` (which occurs a few times) to `'stripe_sepa'` ([or another constant](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4787#discussion_r2526203533)), as changing the string and what we reference can happen without removing the underlying class. It also touches quite a few files, but is pretty easy to validate from inspection alone, as this should be an implementation detail.

wjrosa:
Oh right. I created the other PR here: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4802

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2176,7 +2176,7 @@ public function payment_scripts() {
 		}
 
 		// Bail if no Stripe payment method is enabled.
-		if ( 'no' === $this->enabled && empty( WC_Stripe_Helper::get_legacy_enabled_payment_methods() ) ) {
```
daledupreez, line 2180: Question: should we be queuing these scripts for _any_ pages when UPE is enabled?

includes/class-wc-stripe-helper.php
```diff
@@ -459,73 +459,33 @@ public static function get_settings( $method = null, $setting = null ) {
...
-		}
-
-		return self::$stripe_legacy_gateways;
+		return [];
 	}
 
 	/**
 	 * Get legacy payment method by id.
 	 *
 	 * @return object|null
```
daledupreez: Nit: Is it worth updating this return type to match the new implementation?
```suggestion
	 * @return null
```

includes/class-wc-stripe-helper.php
```diff
@@ -536,183 +496,33 @@ public static function get_legacy_payment_method( $id ) {
...
-
-	/**
-	 * Get settings of individual legacy payment methods.
 	 *
-	 * @return array
-	 *
-	 * @deprecated 9.6.0 The customization of individual payment methods is now deprecated.
+	 * @deprecated 10.2.0 This method will be removed in future versions.
 	 */
-	public static function get_legacy_individual_payment_method_settings() {
```
daledupreez: Nit: It feels like it would be worth splitting this out into a dedicated PR so we can correctly capture the removal in a changelog entry. I almost missed that these two functions had been removed.

includes/payment-methods/class-wc-gateway-stripe-alipay.php
```diff
@@ -1,311 +0,0 @@
...
-}
-
-/**
- * Class that handles Alipay payment method.
- *
- * @extends WC_Gateway_Stripe
- *
- * @since 4.0.0
- */
-class WC_Gateway_Stripe_Alipay extends WC_Stripe_Payment_Gateway {
```
daledupreez, line 18: These classes were not marked as deprecated. Just to be good citizens, I think we should deprecate them before removing them, as merchants may be referencing them in custom code.

includes/payment-methods/class-wc-stripe-upe-payment-method-alipay.php
```diff
@@ -10,8 +10,6 @@ class WC_Stripe_UPE_Payment_Method_Alipay extends WC_Stripe_UPE_Payment_Method {
 
 	const STRIPE_ID = WC_Stripe_Payment_Methods::ALIPAY;
 
-	const LPM_GATEWAY_CLASS = WC_Gateway_Stripe_Alipay::class;
```
daledupreez, line 18: As with the class removals, I think we should deprecate these constants first, as they are public and could be in use.

includes/class-wc-stripe.php
```diff
@@ -651,38 +612,6 @@ protected function toggle_upe( $settings, $old_settings ) {
 	protected function enable_upe( $settings ) {
 		$settings['upe_checkout_experience_accepted_payments'] = [];
 
-		$payment_gateways = WC_Stripe_Helper::get_legacy_payment_methods();
```
daledupreez, line 666: Can we safely remove this logic? What will happen to merchants upgrading from older versions where UPE was not enabled yet?

tests/phpunit/WC_Stripe_Helper_Test.php
```diff
@@ -799,60 +766,6 @@ public function statement_descriptor_sanitation_provider() {
 		];
 	}
 
-	public function test_turning_on_upe_enables_the_correct_upe_methods_based_on_which_legacy_payment_methods_were_enabled() {
```
daledupreez, line 802: Don't we still need the underlying logic for upgrades?

daledupreez (commented):
While I would _love_ to remove all of this code, I am really worried about unforeseen consequences of removing so much code without looking deeply into some of the nuances of what we're changing and/or removing.

includes/payment-methods/class-wc-stripe-upe-payment-method-alipay.php
wjrosa: Alright. New PR here https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4838

includes/payment-methods/class-wc-gateway-stripe-alipay.php
wjrosa: Good point. New PR is here https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4839

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: I don't think so, according to the remaining conditions. We have checks for product, cart, and other pages. Still, removing this part won't change the current behavior (as `get_legacy_enabled_payment_methods` is always empty)

includes/class-wc-stripe-helper.php
wjrosa: Makes sense. New PR here https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4865

includes/class-wc-stripe.php
daledupreez: I realise that we can. The only possible/supported payment method from older versions is card, so the logic below that defaults to `[ 'card', 'link' ]` should cover this.

tests/phpunit/WC_Stripe_Helper_Test.php
daledupreez: As noted above, the only supported legacy payment method for over a year was `card`, so the tests here don't need to be kept running.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.3.0 - xxxx-xx-xx =
+* Dev - Removes legacy checkout payment method classes
```
daledupreez: Given that we are removing code that developers might be referencing, I think we should flag this as an `Update` as well as adding a section to the Compatibility Notes in readme.txt. We want to make sure people can see the changes quickly before their stores break.
```suggestion
* Update - Remove legacy checkout payment method classes
```

phpstan-baseline.neon
```diff
@@ -2400,6 +2394,12 @@ parameters:
 			count: 1
 			path: includes/admin/class-wc-rest-stripe-orders-controller.php
 
+		-
+			message: '#^Call to an undefined method WC_Gateway_Stripe\:\:update_enabled_payment_methods\(\)\.$#'
```
daledupreez: It looks like we may be missing a guard/definition for the type here -- we "know" this should be the UPE gateway, but the code doesn't reflect that. Maybe something for a follow-up?

includes/class-wc-stripe.php
wjrosa: Only cards were enabled for the legacy checkout. So, this migration part (LPM to UPE) would not occur anyway.

tests/phpunit/WC_Stripe_Helper_Test.php
wjrosa: I don't think so, for the same reasons I mentioned in my previous comment. We don't need to toggle payment methods based on the legacy checkout methods, since only cards were available there.

phpstan-baseline.neon
```diff
@@ -2142,6 +2136,12 @@ parameters:
...
+			identifier: method.notFound
+			count: 1
+			path: includes/admin/class-wc-rest-stripe-settings-controller.php
```
daledupreez, line 2143: It looks like we may be missing a guard or have an old definition for the type here -- we "know" this should be the UPE gateway, but the code doesn't reflect that. Maybe something for a follow-up?

daledupreez (approved):
Other than making the change more prominent in the changelog and readme, I think this change is ready to ship. I got 90% of the way through reviewing last week, but I must have forgotten to submit my comments. 😳

phpstan-baseline.neon
wjrosa: Yeah. I think we need to remove the legacy gateway class after moving some of its methods to the child (UPE) class. This should fix this issue.

PR #4788, opened 2025-11-13
Transform WooCommerce prices correctly for express checkout with blocks

daledupreez:
Fixes STRIPE-808

Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4480

This PR ensures that we correctly compute cart totals for express checkout when we're in a block context and there is a mismatch between the number of decimals expected by Stripe and the number of decimals specified in WooCommerce. We addressed the same issue for classic/shortcode contexts in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4480, but that didn't address the issue for block cart and checkout, where the billing data is injected differently.

From an implementation perspective, this patch refactors the main `transformPrice()` function added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4480 to create a new `transformPriceWithMinorUnits()` function, and ensures that this function is used to compute the correct Stripe-facing price based on the Stripe and WooCommerce decimal configurations.

## Testing instructions

* Set up a store to use USD with `0` decimal places
* Connect the store to Stripe, and enable one or more express checkout options, and ensure they will show up in the cart and checkout
* Set up a product with a price over $1,000 so you exceed the minimum price when the computed price is 1/100 of that value.
* Add your expensive product to your cart
* Run the code from `develop`
* Navigate to the block cart
* Launch the express payment method
   - Depending on which express payment methods are enabled, you may see an API request to `https://api.stripe.com/v1/elements/sessions` which includes a `deferred_intent[amount]` URL parameter that can be used to verify the amount we're sending to Stripe
* Notice that the price is sent as the original value from the server, which is 100 times too small, as it should use cents for the amount, e.g. if the price was `$1,800`, it would be shown as `1800` (i.e. $18), when it should be `180000`.
* Close the express checkout without completing payment
* Navigate to block checkout 
* Launch an express checkout (or check the API call) and confirm the amount is still incorrect
* Close the express checkout without completing payment
* Now run the code from this branch - `fix/incorrect-price-computations-for-block-express-checkout`
  - Note that you need to change branch and rebuild the JS files - `npm run build`
* Navigate to the block cart
  - As above, you may see an API request to `https://api.stripe.com/v1/elements/sessions` which includes a `deferred_intent[amount]` URL parameter that can be used to verify the amount we're sending to Stripe
* Launch an express checkout method, and verify that the price is now correct
* Close the express checkout without completing payment
* Navigate to block checkout
* Launch an express checkout method, and verify that the price is now correct

You can also check that things work if you have WooCommerce set to return 3 digits for USD -- we need to divide by 10 in this case, so $18 should be represented as $18.000 in the UI, and `1800` in the call to Stripe. (You don't need to use the expensive product for this case.)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,7 +4,7 @@
 * Fix - Ensure Amazon Pay, Apple Pay, and Google Pay display settings are managed correctly
 * Dev - Add logging with DNS resolution diagnostics for URL validation issues when calling Stripe API
 * Fix - Allow payment methods to be disabled when they are not available
-* Fix - Ensure correct express checkout prices in block cart and checkout with non-default decimal configuration"
+* Fix - Ensure correct express checkout prices in block cart and checkout with non-default decimal configuration
 
 = 10.1.0 - 2025-11-11 =
 * Dev - Remove unused `shouldShowPaymentRequestButton` parameter and calculations from backend
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,7 @@
 * Fix - Ensure Amazon Pay, Apple Pay, and Google Pay display settings are managed correctly
 * Dev - Add logging with DNS resolution diagnostics for URL validation issues when calling Stripe API
 * Fix - Allow payment methods to be disabled when they are not available
+* Fix - Ensure correct express checkout prices in block cart and checkout with non-default decimal configuration"
 
 = 10.1.0 - 2025-11-11 =
 * Dev - Remove unused `shouldShowPaymentRequestButton` parameter and calculations from backend
--- a/client/blocks/express-checkout/express-checkout-container.js
+++ b/client/blocks/express-checkout/express-checkout-container.js
@@ -7,6 +7,7 @@ import {
 	getPaymentMethodTypesForExpressMethod,
 	isManualPaymentMethodCreation,
 } from 'wcstripe/express-checkout/utils';
+import { transformPriceWithMinorUnits } from 'wcstripe/express-checkout/transformers/wc-to-stripe';
 
 export const ExpressCheckoutContainer = ( props ) => {
 	const { stripe, billing, expressPaymentMethod } = props;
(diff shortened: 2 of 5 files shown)


wjrosa (changes requested):
The code changes look good to me, but I'm not seeing the ECE buttons rendered on this branch when I set the decimal places to `0`. If I revert it to `2`, it works just fine. We may have some constraint on that (maybe in `should_show_express_checkout_button` or something related).

diegocurbelo (commented):
Adding a little bit more context to the problem @wjrosa mentioned: 

Image

|||
| - | - |
| Image | Image |

daledupreez:
Thanks for the reviews, @wjrosa and @diegocurbelo! I've now updated the testing instructions to specify that the test product should have a large price so we exceed Stripe's minimum transaction threshold even when the amount is 100 times too small.

wjrosa (approved):
Works as expected now 👍 . But maybe we should fix the threshold issue (add this check to `canMakePayment` or something)

diegocurbelo (approved):
It tests as expected now!

I think we should merge this and create a follow-up issue for the threshold fix, as the underlying problem is not caused by changes in this PR:

When rendering the product page, there are two calls to sessions, both with the expected amount (1200):
Image

But when rendering the Cart, it makes three extra calls with the displayed price (12); if those fail, then the buttons are not rendered:
Image

Now, when using 3 decimal digits, the first three calls do not fail (but use the wrong amount 12000), and then the "correct" two are made (with 1200) and the buttons re-rendered so everything works as expected:
|||
|-|-|
|Image|Image|

daledupreez:
Good catch, @diegocurbelo!

I've created https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4808 to track that problem. Can you read through the issue to confirm that I've captured the problem correctly? I am particularly interested in knowing whether the issue is for the block or classic cart, or both.

PR #4794, opened 2025-11-13
Making the system status report more secure

wjrosa:
Fixes STRIPE-792

To prevent public disclosure of the merchant's Stripe account email address, I am removing it from the System Status Report data we generate.

## Testing instructions

Code review. OR:

- Checkout to this branch on your test environment (`dev/removing-the-merchant-emails-from-report`)
- Connect your Stripe account
- Go to WooCommerce _> Status and click the "Get system report" button
- Confirm the Stripe account email address is not included in the report

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
-* Dev - Removes the merchant email address from the System Status Report
+* Dev - Remove the merchant email address from the System Status Report
 * Update - Changes labels related to saved payment methods from "cards" to "payment methods"
 * Fix - Ensure Amazon Pay, Apple Pay, and Google Pay display settings are managed correctly
 * Dev - Add logging with DNS resolution diagnostics for URL validation issues when calling Stripe API
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Dev - Removes the merchant email address from the System Status Report
 * Fix - Ensure Amazon Pay, Apple Pay, and Google Pay display settings are managed correctly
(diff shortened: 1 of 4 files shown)


Mayisha:
@wjrosa instead of removing it completely, what do you think about masking the email?
I don't have any specific example in mind but having the account ID and email combination might be useful in the SSR.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Dev - Removes the merchant email address from the System Status Report
```
daledupreez: I have two minor comments:
1. I think this should qualify as a Fix rather than a Dev change. This does have implications for merchants.
2. I think it would be a bit clearer if we use "Remove" instead of "Removes", as that matches the approach for most of our changelog entries.
```suggestion
* Dev - Remove the merchant email address from the System Status Report
```

daledupreez (approved):
I am happy to remove the email address, especially if we have the account ID. I don't feel super strongly about masking the email address, but I think we should prefer not to include it unless it's super-necessary.

wjrosa:
I personally never saw any support request needing the email address information, so I think we can just remove it for now 👍

Mayisha (approved):
> I personally never saw any support request needing the email address information, so I think we can just remove it for now 

Sounds good then 👍 

:shipit:

readme.txt
```diff
@@ -111,6 +111,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.2.0 - xxxx-xx-xx =
+* Dev - Remove unused `shouldShowPaymentRequestButton` parameter and calculations from backend
```
Mayisha: @wjrosa the readme entry is wrong. Please check it before merging.

PR #4803, opened 2025-11-17
Disable express checkout when Amazon Pay is disabled and the only method

### LINKED ISSUE #4783: Prevent express checkout resources from being enqueued unnecessarily when Amazon Pay is disabled
In https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4773, we have implemented some initial code to hide Amazon Pay when the store is configured to calculate taxes based on the user's billing address. While that works, there are some situations that may require  more nuance in the client, as we don't want to send any more data to the client than necessary. The most egregious situation would be for the case where Amazon Pay is the only express checkout method, but has been disabled due to the store's tax configuration, as we would enqueue the resources, but would never render anything.

This was originally flagged by @frosso in [this comment thread](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4773#discussion_r2513572744).

daledupreez:
Fixes STRIPE-807
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4783

Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4773

This PR ensures that when Amazon Pay is the only available express checkout method AND Amazon Pay will be disabled because taxes are enabled on billing address, we won't enqueue any express checkout scripts or resources.

## Testing instructions

* Set up a store using USD, enable taxes, and configure taxes to be based on billing address
* Run the Stripe plugin out of `develop`, and ensure it is connected to Stripe
* Enable Amazon Pay as a feature by setting the feature flag (`wp update option _wcstripe_feature_amazon_pay yes`) or by hooking into the feature filter (`add_filter( 'wc_stripe_is_amazon_pay_available', '__return_true' );`)
* Navigate to the Stripe payment methods screen
* Ensure that Apple Pay/Google Pay and Link are disabled, and Amazon Pay is enabled.
* Click on the "Customize" link for Amazon Pay, and confirm that Amazon Pay is configured to display in the cart, checkout, and product pages.
* For each of the following pages, navigate to the page and verify that the `express-checkout.js` script is loaded, or that the `wc_stripe_express_checkout_params` global variable is defined in the Javascript console:
   - a product page -- add a standard product to your cart so the pages below work
   - block cart
   - classic cart
   - block checkout
   - classic checkout
* Now switch to this branch: `git checkout fix/unnecessary-express-checkout-resources-when-amazon-pay-billing-address-disabled`
* Reload the product page, cart pages, and checkout pages, and verify that the scripts are no longer enqueued/visible

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -735,6 +735,15 @@ class WC_Stripe_Express_Checkout_Helper {
 			}
 		}
 
+		// Check if Amazon Pay is the only enabled method, but not available due to the tax configuration.
+		if ( $this->is_amazon_pay_enabled() &&
+			! ( $this->is_payment_request_enabled() || $this->is_link_enabled() ) &&
+			( wc_tax_enabled() && 'billing' === get_option( 'woocommerce_tax_based_on' ) )
+		) {
+			WC_Stripe_Logger::debug( 'Stripe Express Checkout is hidden due to Amazon Pay being the only enabled method, but not available due to taxes being based on billing address.' );
+			return false;
+		}
+
 		// Hide if cart/product doesn't require shipping and tax is based on billing or shipping address.
 		$hide_based_on_tax          = $this->should_hide_ece_based_on_tax_setup();
 		$hide_based_on_tax_filtered = apply_filters( 'wc_stripe_should_hide_express_checkout_button_based_on_tax_setup', $hide_based_on_tax );
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,6 +9,7 @@
 * Fix - Ensure state and postal code are optional in express checkout for Gulf countries (UAE, Bahrain, Kuwait, Oman, Qatar)
 * Dev - Removes the `_wcstripe_feature_upe` feature flag and the related method from the `WC_Stripe_Feature_Flags` class
 * Dev - Fixes some incorrect subscriptions support implementations for payment methods
+* Fix - Disable express checkout when Amazon Pay is disabled and the only method
 
 = 10.1.0 - 2025-11-11 =
 * Dev - Remove unused `shouldShowPaymentRequestButton` parameter and calculations from backend
--- a/readme.txt
+++ b/readme.txt
@@ -119,5 +119,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Fix - Ensure state and postal code are optional in express checkout for Gulf countries (UAE, Bahrain, Kuwait, Oman, Qatar)
 * Dev - Removes the `_wcstripe_feature_upe` feature flag and the related method from the `WC_Stripe_Feature_Flags` class
 * Dev - Fixes some incorrect subscriptions support implementations for payment methods
+* Fix - Disable express checkout when Amazon Pay is disabled and the only method
 
 [See changelog for full details across versions](https://raw.githubusercontent.com/woocommerce/woocommerce-gateway-stripe/trunk/changelog.txt).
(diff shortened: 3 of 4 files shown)


wjrosa (approved):
Change looks good and works as intended 👍

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -735,6 +735,15 @@ public function should_show_express_checkout_button() {
 			}
 		}
 
+		// Check if Amazon Pay is the only enabled method, but not available due to the tax configuration.
+		if ( $this->is_amazon_pay_enabled() &&
```
Mayisha, line 739: QQ: Is there a reason we are not relying on the `should_hide_ece_based_on_tax_setup` for hiding the button and adding this separate check? 
I see the `should_hide_ece_based_on_tax_setup` function is a bit different than this check as it has two additional checks
- `is_product_or_cart_taxable`
- `product_or_cart_needs_shipping`

Could you help me understand why these 2 checks are not needed for Amazon Pay? I am assuming it's not needed because you added a separate check instead of relying on `should_hide_ece_based_on_tax_setup`.

Mayisha (commented):
I have a QQ for my understanding. 

And I see a different behavior in block vs classic checkout when I have
- Tax based on billing address
- Google/Apple Pay enabled
- Amazon Pay enabled

With the above setup
- Block checkout has Google, Apple, Amazon Pay buttons.

Screenshot 2025-11-18 at 3 49 25 PM

- Classic checkout has Google & Apple Pay buttons.

Screenshot 2025-11-18 at 3 48 28 PM

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
daledupreez: The core issue here is that the output of `should_hide_ece_based_on_tax_setup()` is passed into the `wc_stripe_should_hide_express_checkout_button_based_on_tax_setup` filter, which merchants can override.

For this specific tax configuration, Amazon Pay _will not work_ so we want to ensure that it is not enabled or enqueued, as we can't make it work at present.

daledupreez:
Thanks for the review, @Mayisha -- Amazon Payshould not be enabled in block checkout, so that feels like a separate bug.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -735,6 +735,15 @@ public function should_show_express_checkout_button() {
...
+			! ( $this->is_payment_request_enabled() || $this->is_link_enabled() ) &&
+			( wc_tax_enabled() && 'billing' === get_option( 'woocommerce_tax_based_on' ) )
+		) {
+			WC_Stripe_Logger::debug( 'Stripe Express Checkout is hidden due to Amazon Pay being the only enabled method, but not available due to taxes being based on billing address.' );
+			return false;
+		}
```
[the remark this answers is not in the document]
daledupreez: This is intentional, as `should_hide_ece_based_on_tax_setup()` can be overridden by the `wc_stripe_should_hide_express_checkout_button_based_on_tax_setup`. For this specific case, Amazon Pay won't work, so we should not allow it to be enabled.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
Mayisha: Got it. Thanks for the details.

Mayisha (approved):
> Amazon Payshould not be enabled in block checkout, so that feels like a separate bug.

Considering it a separate bug and to be solved in it's own PR, we should create an issue to track this. 

Changes in this PR look good to me. :shipit:

PR #4805, opened 2025-11-17
Removing the UPE checkout enabled feature flag usages

wjrosa:

In this simple cleanup PR, I am removing all references to the UPE-enabled feature flag, as that is no longer necessary, since UPE is now the only version available.

I am also deprecating `WC_Stripe_UPE_Availability_Note` and `WC_Stripe_REST_UPE_Flag_Toggle_Controller` as they do not make sense anymore.

## Testing instructions

Code review. Check if the tests are still passing. Perform some basic smoke testing and confirm no regression was introduced.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -32,7 +32,6 @@ class WC_Stripe_Admin_Notices {
 		add_action( 'admin_notices', [ $this, 'admin_notices' ] );
 		add_action( 'wp_loaded', [ $this, 'hide_notices' ] );
 		add_action( 'woocommerce_stripe_updated', [ $this, 'stripe_updated' ] );
-		add_action( 'after_plugin_row_woocommerce-gateway-stripe/woocommerce-gateway-stripe.php', [ $this, 'display_legacy_deprecation_notice' ], 10, 1 );
 	}
 
 	/**
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -128,57 +128,7 @@ class WC_Stripe_Admin_Notices {
 	 * @param string $plugin_file Plugin file.
 	 */
 	public static function display_legacy_deprecation_notice( $plugin_file ) {
-		global $wp_list_table;
-		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
-
-		// If Stripe is not enabled, don't show the legacy deprecation notice.
-		if ( ! isset( $stripe_settings['enabled'] ) || 'no' === $stripe_settings['enabled'] ) {
-			return;
-		}
-
-		if ( WC_Stripe_Feature_Flags::is_upe_checkout_enabled() ) {
-			return;
-		}
-
-		if ( is_null( $wp_list_table ) ) {
-			return;
-		}
-
-		$columns_count   = $wp_list_table->get_column_count();
-		$is_active       = is_plugin_active( $plugin_file );
-		$is_active_class = $is_active ? 'active' : 'inactive';
-
-		$setting_link = esc_url( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe&panel=settings' ) );
-		$message      = sprintf(
-			/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag */
-			__( 'WooCommerce Stripe Gateway legacy checkout experience has been deprecated since version 9.6.0. Please %1$smigrate to the new checkout experience%2$s to access more payment methods and avoid disruptions. %3$sLearn more%4$s', 'woocommerce-gateway-stripe' ),
-			'<a href="' . $setting_link . '">',
-			'</a>',
-			'<a href="https://woocommerce.com/document/stripe/admin-experience/legacy-checkout-experience/" target="_blank">',
-			'</a>'
-		);
-
-		?>
-		<tr class='plugin-update-tr <?php echo esc_html( $is_active_class ); ?>' data-id="woocommerce-gateway-stripe-update" data-slug="woocommerce-gateway-stripe" data-plugin='<?php echo esc_html( $plugin_file ); ?>'>
-			<td colspan='<?php echo esc_html( $columns_count ); ?>' class='plugin-update colspanchange'>
-				<div class='notice inline notice-warning notice-alt'>
-					<p>
-						<span style="display: inline-block; vertical-align: text-top;">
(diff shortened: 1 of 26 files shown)


includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
daledupreez, line 69: Should we also remove the logic in the update path?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -248,7 +248,7 @@ public function __construct() {
 		$this->statement_descriptor          = ! empty( $main_settings['statement_descriptor'] ) ? $main_settings['statement_descriptor'] : '';
 
 		// When feature flags are enabled, title shows the count of enabled payment methods in settings page only.
```
daledupreez: We may want to update this comment to match the new logic. I think we now do this whenever we're in the specific page context.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -474,31 +470,27 @@ public function get_account_saved_payment_methods_list_item( $item, $payment_tok
...
+			 * When all conditions are met, we don't want to delete the payment method from Stripe.
+			 * This is to avoid detaching the payment method from the live stripe account on non production environments.
+			 */
+			if (
+				WC_Stripe_Mode::is_live() &&
+				is_admin() &&
+				'production' !== wp_get_environment_type()
+			) {
+				return;
 			}
```
daledupreez: Unrelated note: this code should be calling `WC_Stripe_API::should_detach_payment_method_from_customer()` so we can (re)use that central logic and not have it partially replicated here. I'll create a PR to do just that shortly.

daledupreez (commented):
Thanks for taking this on, @wjrosa!

Based on inspection, this looking really good, with the exception of some legitimate test failures in `WC_Stripe_Inbox_Notes_Test`.

It may also be worth triggering a manual e2e test run that includes the secondary LPMs. I will also aim to do some more manual testing over the next few hours.

wjrosa:
Thanks for the review, Dale! I have fixed tests in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4805/commits/bfa000db8b858dee5eb7f6ff721c3f71dd8048de and also deprecated the `WC_Stripe_UPE_Availability_Note` class, as it does not make more sense to us.

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
wjrosa: I think it makes sense to deprecate the whole class, actually. Did it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4805/commits/0cecb0fe3f604010491d13cd5d6997bab4a60424

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4805/commits/86b8113e0891d325af4cb2629c981ec1f2cdaf4a

includes/payment-tokens/class-wc-stripe-payment-tokens.php
daledupreez: Working on this in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4812

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
```diff
@@ -64,16 +68,12 @@ public function get_flag() {
...
-
-		if ( null === $is_upe_enabled ) {
-			return new WP_REST_Response( [ 'result' => 'bad_request' ], 400 );
-		}
-
 		$settings = WC_Stripe_Helper::get_stripe_settings();
-		$settings[ WC_Stripe_Feature_Flags::UPE_CHECKOUT_FEATURE_ATTRIBUTE_NAME ] = $is_upe_enabled ? 'yes' : 'disabled';
+		$settings[ WC_Stripe_Feature_Flags::UPE_CHECKOUT_FEATURE_ATTRIBUTE_NAME ] = 'yes';
 
 		WC_Stripe_Helper::update_main_stripe_settings( $settings );
```
daledupreez: I don't think we even need to modify the settings, TBH.
```suggestion
```

daledupreez (approved):
This is looking good to me, though I don't know how much I covered with my smoke tests. Still, given that we can no longer disable UPE, I think these changes should be relatively safe.

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
wjrosa: I agree. Applied it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4805/commits/718d6a9d82a3fd6fe1a089725270fad7d8be51c7

includes/payment-methods/class-wc-stripe-payment-request.php
```diff
@@ -68,6 +70,8 @@ class WC_Stripe_Payment_Request {
 	 * @version 4.0.0
 	 */
 	public function __construct() {
+		return;
```
diegocurbelo: This looks a bit odd; I think we should move the `return` below the `add_action` (as it was before), or remove the unreachable code (or the entire constructor).

includes/admin/class-wc-stripe-admin-notices.php
diegocurbelo: `display_legacy_deprecation_notice()` does nothing now; there is no need to register this anymore:

```suggestion
```

diegocurbelo (approved):
The changes look good. I left two optional suggestions.

includes/payment-methods/class-wc-stripe-payment-request.php
wjrosa: You're right! I did not pay much attention to this part. I have updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4805/commits/1dd8cb6e1487a8c709e1a27d60163d4bebd98a8c

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Thanks for the heads up. Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4805/commits/55e59305869b9e89094fed198c1c819a775849a3

PR #4807, opened 2025-11-17
Use the built-in Database Cache for Connect flow data instead of transients

### LINKED ISSUE #4729: Persistent reconnection notice with v10.0.0
### Description

After merchants update to Stripe v10.0.0 merchants are prompted to reconnect their site to their Stripe account. **After this, a couple of different results occur on different sites:**

1. Some merchants see that their site is successfully reconnected to Stripe but the site notice prompting them to reconnect to Stripe still shows and cannot be closed/removed.
2. Some sites see that when they follow the instructions to reconnect their site and Stripe account, they are unable to successfully connect the site to Stripe anymore. They go in a loop and end up in the same place, their account still not successfully reconnecting. One of the merchant says the loop is stuck on this page:

Image

What worked for some merchants who could not reconnect is that they rolled back Stripe to v9.9.2, this solution however did not work for all merchants as at least one merchant says that even after rollbacks, their site is not reconnecting to Stripe - 10358774-zd-a8c.

### Steps to reproduce

There is no clear reproduction steps as merchants said that all they did was update Stripe to v10.0.0 and the notice to Reconnect their Stripe account to their WooCommerce site.

### Expected behavior

When a merchant attempts to reconnect their site to their Stripe account, the reconnection works without issue and the notice prompting them to reconnect does not show anymore.

### System report

+++ System report

<details><summary></summary>

### Additional context

There are several instances of the issue so far:

* Support forum: [https://wordpress.org/support/topic/woocommerce-stripe-gateway-10-0-0-causing-connection-issues/](https://wordpress.org/support/topic/woocommerce-stripe-gateway-10-0-0-causing-connection-issues/) (here a rollback worked for one of the merchants)
* Internal reports:
  * 10358795-zd-a8c
  * 10358633-zd-a8c
  * 10359120-zd-a8c
  * 10359133-zd-a8c
  * 10358774-zd-a8c
  * 10343116-zd-a8c

</details>

diegocurbelo:
Fixes STRIPE-745
Fixes #4729

This PR replaces the transient used during the OAuth Connect with an entry in the built-in Database cache.

## Testing instructions

1. Disconnect your Stripe account
2. Go to the Onboard page (WooCommerce > Payments > Stripe)
3. Click `Create or connect a test account instead`
4. Verify that the connection succeeds. 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,6 +11,7 @@
 * Fix - Ensure state and postal code are optional in express checkout for Gulf countries (UAE, Bahrain, Kuwait, Oman, Qatar)
 * Dev - Removes the `_wcstripe_feature_upe` feature flag and the related method from the `WC_Stripe_Feature_Flags` class
 * Dev - Fixes some incorrect subscriptions support implementations for payment methods
+* Fix - Use the builtin Database Cache for the Connect flow data
 
 = 10.1.0 - 2025-11-11 =
 * Dev - Remove unused `shouldShowPaymentRequestButton` parameter and calculations from backend
--- a/includes/class-wc-stripe-database-cache.php
+++ b/includes/class-wc-stripe-database-cache.php
@@ -329,7 +329,13 @@ class WC_Stripe_Database_Cache {
 	 * @return string The key with the prefix.
 	 */
 	private static function add_key_prefix( string $key, ?string $mode = null ): string {
-		$mode = $mode ?? ( WC_Stripe_Mode::is_test() ? 'test' : 'live' );
+		if ( null === $mode ) {
+			$mode = WC_Stripe_Mode::is_test() ? 'test' : 'live';
+		} elseif ( 'live' !== $mode && 'test' !== $mode ) {
+			// Don't allow other values for $mode
+			$mode = 'test';
+		}
+		// Otherwise $mode is either 'live' or 'test'
 		return self::CACHE_KEY_PREFIX . $mode . '_' . $key;
 	}
 
--- a/includes/class-wc-stripe-database-cache.php
+++ b/includes/class-wc-stripe-database-cache.php
@@ -80,7 +80,21 @@ class WC_Stripe_Database_Cache {
 	 * @return void
 	 */
 	public static function set( $key, $data, $ttl = HOUR_IN_SECONDS ) {
-		$prefixed_key = self::add_key_prefix( $key );
+		self::set_with_mode( $key, $data, $ttl, null );
+	}
+
+	/**
+	 * Stores a value in the cache for the specified mode.
+	 *
+	 * @param string $key The key to store the value under.
+	 * @param mixed  $data The value to store.
+	 * @param int    $ttl  The TTL of the cache. Dafault 1 hour.
+	 * @param string|null $mode The mode to use as prefix for the key. Default is null, which means the current plugin mode.
+	 *
+	 * @return void
+	 */
+	public static function set_with_mode( $key, $data, $ttl = HOUR_IN_SECONDS, ?string $mode = null ) {
+		$prefixed_key = self::add_key_prefix( $key, $mode );
 		self::write_to_cache( $prefixed_key, $data, $ttl );
 	}
 
@@ -94,7 +108,21 @@ class WC_Stripe_Database_Cache {
 	 * @return mixed|null The cache contents. NULL if the cache value is expired or missing.
 	 */
 	public static function get( $key ) {
-		$prefixed_key = self::add_key_prefix( $key );
+		return self::get_with_mode( $key, null );
+	}
+
+	/**
+	 * Gets a value from the cache for the specified mode.
+	 *
+	 * The key is automatically prefixed with "wcstripe_cache_[mode]_".
+	 *
+	 * @param string $key       The key to look for.
+	 * @param string|null $mode The mode to use as prefix for the key. Default is null, which means the current plugin mode.
+	 *
+	 * @return mixed|null The cache contents. NULL if the cache value is expired or missing.
+	 */
+	public static function get_with_mode( $key, ?string $mode = null ) {
+		$prefixed_key = self::add_key_prefix( $key, $mode );
 		$cache_contents = self::get_from_cache( $prefixed_key );
 		if ( is_array( $cache_contents ) && array_key_exists( 'data', $cache_contents ) ) {
 			if ( self::is_expired( $prefixed_key, $cache_contents ) ) {
@@ -119,8 +147,19 @@ class WC_Stripe_Database_Cache {
 	 * @return void
 	 */
 	public static function delete( $key ) {
-		$prefixed_key = self::add_key_prefix( $key );
+		self::delete_with_mode( $key, null );
+	}
 
+	/**
+	 * Deletes a value from the cache for the specified mode.
+	 *
+	 * @param string $key  The key to delete.
+	 * @param string $mode The mode to use as prefix for the key. Default is null, which means the current plugin mode.
+	 *
+	 * @return void
+	 */
+	public static function delete_with_mode( $key, ?string $mode = null ): void {
+		$prefixed_key = self::add_key_prefix( $key, $mode );
 		self::delete_from_cache( $prefixed_key );
 	}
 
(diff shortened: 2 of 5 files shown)


changelog.txt
```diff
@@ -11,6 +11,7 @@
 * Fix - Ensure state and postal code are optional in express checkout for Gulf countries (UAE, Bahrain, Kuwait, Oman, Qatar)
 * Dev - Removes the `_wcstripe_feature_upe` feature flag and the related method from the `WC_Stripe_Feature_Flags` class
 * Dev - Fixes some incorrect subscriptions support implementations for payment methods
+* Fix - Use the builtin Database Cache for the Connect flow data
```
wjrosa:
```suggestion
* Fix - Use the built-in Database Cache for the Connect flow data
```

readme.txt
```diff
@@ -121,5 +121,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Fix - Ensure state and postal code are optional in express checkout for Gulf countries (UAE, Bahrain, Kuwait, Oman, Qatar)
 * Dev - Removes the `_wcstripe_feature_upe` feature flag and the related method from the `WC_Stripe_Feature_Flags` class
 * Dev - Fixes some incorrect subscriptions support implementations for payment methods
+* Fix - Use the builtin Database Cache for the Connect flow data
```
wjrosa:
```suggestion
* Fix - Use the built-in Database Cache for the Connect flow data
```

wjrosa (approved):
Looks good to me! Left a typo fix suggestion

includes/connect/class-wc-stripe-connect.php
```diff
@@ -64,7 +64,7 @@ public function get_oauth_url( $return_url = '', $mode = 'live' ) {
 				return $result;
 			}
 
-			set_transient( 'wcs_stripe_connect_state_' . $mode, $result->state, 6 * HOUR_IN_SECONDS );
+			WC_Stripe_Database_Cache::set( 'wcs_stripe_connect_state_' . $mode, $result->state, 6 * HOUR_IN_SECONDS );
```
daledupreez: I have a concern here around what would happen if the current store mode doesn't match the value of `$mode`, as the cache code appends the current store mode as part of the cache key. But there is no guarantee that `$mode` is the same (or will be the same) as the current store mode (if one is even set yet). I think we need to explicitly set the mode before we write to the cache, rather than including `$mode` in the key.

I ended up testing and it's possible to get into weird states if you do the following:
* Connect a site in test mode - make sure you've completed the connection flow
* Trigger a (re)connection flow, and keep this window/tab open -- it can be for live or test mode
* Note that we create a database cache entries for `wcstripe_cache_test_wcs_stripe_connect_state_live` and `wcstripe_cache_test_wcs_stripe_connect_state_test`
* In a new window, disconnect your site from Stripe, which shifts us back into "live" mode because the `testmode` sub-option is not specified
* Complete the Stripe connection flow in the first window - notice that you end up on the connection screen, as we can't read the database cache due to the mode switching from test to live as a result of the disconnection

While this is a somewhat artificial test, it does feel like we should guard against this in some way. I considered having this code temporarily switch the mode via the `testmode` sub-option, but that could have other indirect effects for other incoming requests or logic. It feels like it may be better to expose something like `WC_Stripe_Database_Cache::set_for_mode()`, `WC_Stripe_Database_Cache::get_for_mode()`, and `WC_Stripe_Database_Cache::delete_for_mode()` APIs that then allow the key prefix to be calculated based on the mode. Maybe something like the following:
```php
	/**
	 * Adds the CACHE_KEY_PREFIX + plugin mode prefix to the key.
	 * Ex: "wcstripe_cache_[mode]_[key].
	 *
	 * @param string       $key  The key to add the prefix to.
	 * @param bool|null $test_mode If we should force test mode (true) or live mode (false). Null implies we should use the current mode. 
	 *
	 * @return string The key with the prefix.
	 */
	private static function add_key_prefix( $key, ?bool $test_mode = null ): string {
		if ( null === $test_mode ) {
			$is_test_mode = WC_Stripe_Mode::is_test();
		} else {
			$is_test=mode = $test_mode;
		}
		$mode = $is_test_mode ? 'test_' : 'live_';
		return self::CACHE_KEY_PREFIX . $mode . $key;
	}

	/**
	 * Gets a value from the cache.
	 *
	 * The key is automatically prefixed with "wcstripe_cache_[mode]_".
	 *
	 * @param string $key The key to look for.
	 *
	 * @return mixed|null The cache contents. NULL if the cache value is expired or missing.
	 */
	public static function get( $key ) {
		return self::get_with_mode( $key, WC_Stripe_Mode::is_test() );
	}

	/**
	 * Gets a value from the cache for a specified mode.
	 *
	 * The key is automatically prefixed with "wcstripe_cache_[mode]_".
	 *
	 * @param string $key     The key to look for.
	 * @param bool  $is_test_mode Whether we are in test mode or not.
	 *
	 * @return mixed|null The cache contents. NULL if the cache value is expired or missing.
	 */
	public static function get_with_mode( $key, bool $is_test_mode ) {
		$prefixed_key = self::add_key_prefix( $key, $is_test_mode );
		$cache_contents = self::get_from_cache( $prefixed_key );
		if ( is_array( $cache_contents ) && array_key_exists( 'data', $cache_contents ) ) {
			if ( self::is_expired( $prefixed_key, $cache_contents ) ) {
				return null;
			}

			self::maybe_trigger_prefetch( $key, $cache_contents );

			return $cache_contents['data'];
		}

		return null;
	}
```

includes/connect/class-wc-stripe-connect.php
```diff
@@ -131,7 +131,7 @@ public function connect_oauth( $state, $code, $type = 'connect', $mode = 'live'
 				return $response;
 			}
 
-			delete_transient( 'wcs_stripe_connect_state_' . $mode );
+			WC_Stripe_Database_Cache::delete( 'wcs_stripe_connect_state_' . $mode );
```
daledupreez: As above, I am concerned that the current mode (in the `testmode` setting) may not match the mode at the time the write occurred.

daledupreez (commented):
Gah! I forgot to submit my comments yesterday! Apologies for the delay, @diegocurbelo!

includes/connect/class-wc-stripe-connect.php
diegocurbelo: I agree it's an edge case, but it makes sense, and it could be helpful to allow setting the mode manually if desired (maybe for caching data that is the same regardless of the mode).

Refactored in [624ef27](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4807/commits/624ef277b853381d8ac6e63eec2973a1aa06cfc0)

includes/class-wc-stripe-database-cache.php
```diff
@@ -284,13 +323,14 @@ private static function maybe_trigger_prefetch( string $key, array $cache_conten
...
+	 * @param string $key       The key to add the prefix to.
+	 * @param string|null $mode The mode to use as prefix for the key. Default is null, which means the current plugin mode.
 	 *
 	 * @return string The key with the prefix.
 	 */
-	private static function add_key_prefix( $key ) {
-		$mode = WC_Stripe_Mode::is_test() ? 'test_' : 'live_';
-		return self::CACHE_KEY_PREFIX . $mode . $key;
+	private static function add_key_prefix( string $key, ?string $mode = null ): string {
+		$mode = $mode ?? ( WC_Stripe_Mode::is_test() ? 'test' : 'live' );
```
daledupreez: As things stand, any string value for `$mode` will be accepted. Might it make sense to validate that we have `'live'`, `'test'`, or `null` at this stage?
```suggestion
		if ( null === $mode ) {
			$mode = WC_Stripe_Mode::is_test() ? 'test' : 'live';
		} elseif ( 'live' !== $mode && 'test' !== $mode ) {
			// Don't allow other values for $mode
			$mode = 'test';
		}
		// Otherwise $mode is either 'live' or 'test'
```

daledupreez (approved):
Other than the validation check, this is LGTM and tests well!

It also feels like using the cache here may mean that we might have a simpler time using my proposed alternative approach for https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4818, where we might want to append the current user name/ID to the cache so we keep the same state for a given user without continually regenerating it.

PR #4811, opened 2025-11-18
Making OCS enabled by default for new installs

wjrosa:
Fixes STRIPE-818

As originally envisioned, with this PR, we are making the Optimized Checkout Suite feature enabled by default for all new installations. This is to make sure merchants are using the latest integration recommended by Stripe.

## Testing instructions

- Checkout and build this branch on a new test environment (`update/making-ocs-enabled-by-default-for-new-installs`)
- Connect your Stripe account
- Confirm the Optimized Checkout feature is enabled in settings:
Screenshot 2025-11-18 at 16 52 41

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Update - Makes the Optimized Checkout Suite feature enabled by default for all new installations
 * Dev - Expands the Stripe Order Helper class to handle mandate ID, Multibanco data, refund status, card brand, charge captured flag, status final flag, and the refund failure reason 
 * Dev - Remove the merchant email address from the System Status Report
 * Dev - Replace the constant reference for the legacy SEPA payment method
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
-* Update - Makes the Optimized Checkout Suite feature enabled by default for all new installations
+* Update - Enable the Optimized Checkout Suite feature for all new installations
 * Dev - Deprecates all the legacy checkout payment method classes
 * Dev - Deprecates all the LPM class constants
 * Dev - Remove all references to the UPE-enabled feature flag
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -370,13 +370,22 @@ class WC_Stripe {
 			// Try to schedule the daily async cleanup of the Stripe database cache.
 			WC_Stripe_Database_Cache::maybe_schedule_daily_async_cleanup();
 
-			// If we have previously disabled settings synchronization, remove the flag after the upgrade,
-			// just to make sure we are still ineligible for settings synchronization.
 			$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
-			if ( isset( $stripe_settings['pmc_enabled'] ) && 'no' === $stripe_settings['pmc_enabled'] ) {
-				unset( $stripe_settings['pmc_enabled'] );
-				WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
-				WC_Stripe_Logger::warning( 'Settings synchronization eligibility will be re-checked after upgrade' );
+			if ( isset( $stripe_settings['pmc_enabled'] ) ) {
+				// If we have previously disabled settings synchronization, remove the flag after the upgrade,
+				// just to make sure we are still ineligible for settings synchronization.
+				if ( 'no' === $stripe_settings['pmc_enabled'] ) {
+					unset( $stripe_settings['pmc_enabled'] );
+					WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
+					WC_Stripe_Logger::warning( 'Settings synchronization eligibility will be re-checked after upgrade' );
+				} else if ( 'yes' === $stripe_settings['pmc_enabled'] ) {
+					// Enable the Optimized Checkout feature by default if not set.
+					// It requires PMC to be enabled.
+					if ( ! isset( $stripe_settings['optimized_checkout_element'] ) ) {
+						$stripe_settings['optimized_checkout_element'] = 'yes';
+						WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
+					}
+				}
 			}
 		}
 	}
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -344,7 +344,7 @@ class WC_Stripe {
 				define( 'WC_STRIPE_INSTALLING', true );
 			}
 
-			// Mark as fresh install if no version is set.
+			// Mark optimized checkout as default on for new installs.
 			if ( false === get_option( 'wc_stripe_version' ) && false === get_option( 'wc_stripe_optimized_checkout_default_on' ) ) {
 				update_option( 'wc_stripe_optimized_checkout_default_on', true );
 			}
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -375,15 +375,13 @@ class WC_Stripe {
 			// Try to schedule the daily async cleanup of the Stripe database cache.
 			WC_Stripe_Database_Cache::maybe_schedule_daily_async_cleanup();
 
+			// If we have previously disabled settings synchronization, remove the flag after the upgrade,
+			// just to make sure we are still ineligible for settings synchronization.
 			$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
-			if ( isset( $stripe_settings['pmc_enabled'] ) ) {
-				// If we have previously disabled settings synchronization, remove the flag after the upgrade,
-				// just to make sure we are still ineligible for settings synchronization.
-				if ( 'no' === $stripe_settings['pmc_enabled'] ) {
-					unset( $stripe_settings['pmc_enabled'] );
-					WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
-					WC_Stripe_Logger::warning( 'Settings synchronization eligibility will be re-checked after upgrade' );
-				}
+			if ( isset( $stripe_settings['pmc_enabled'] ) && 'no' === $stripe_settings['pmc_enabled'] ) {
+				unset( $stripe_settings['pmc_enabled'] );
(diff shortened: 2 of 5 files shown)


includes/connect/class-wc-stripe-connect.php
```diff
@@ -255,6 +255,7 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 			$options['enabled']                         = 'yes';
 			$options['testmode']                        = $is_test ? 'yes' : 'no';
 			$options['upe_checkout_experience_enabled'] = $this->get_upe_checkout_experience_enabled();
+			$options['optimized_checkout_element']      = $this->get_optimized_checkout_element_default_value();
```
daledupreez: I will keep thinking about this tomorrow, but it feels like it might be _really_ good to add some unit tests so we can correctly track what is going on here and what edge cases might exist.

I am concerned that we might have edge cases where merchants upgrade, and then reconnect, and we auto-enable Optimized Checkout just because we hadn't saved any data in that sub-option from before that reconnection attempt.

Might it be better to create some code in `WC_Stripe::install()` and/or hook into the `woocommerce_stripe_updated` action to set/pick the default value for optimized checkout based on whether there is data in the `wc_stripe_version` option from before? That way OCS would be default-on for fresh new installs only, whereas we'd leave it off for any other install.

Mayisha (approved):
It looks good and is working well as per my testing in JN sites. 👍 

- New installations
    - Directly installed this branch build in a fresh JN site.
    - Connected to a test Stripe account.
    - OCS is enabled after account connection.

- Existing account
    - Installed Stripe `10.1.0` in a fresh site.
    - Connected to a test Stripe account.
    - Changed a few settings keeping OCS disabled.
    - Updated the plugin with this branch's build.
    - Refreshed settings page -> OCS disabled.
    - Reconnected to the Stripe account -> OCS disabled.
    - Disconnected the Stripe account, connected again -> OCS disabled.

On the topic of tracking possible edge cases, can we add some logs when we auto enable OCS?

daledupreez (changes requested):
I just confirmed that the current logic will not work as expected for upgrades from versions where Optimized Checkout was not initialised by default, and then the merchant reconnects.

I specifically tested the following:
* Install Stripe 9.5.2 on a fresh site
* Connect to Stripe
* Verify that the `optimized_checkout_element` sub-option is empty -- `wp option pluck woocommerce_stripe_settings optimized_checkout_element`
  - I also confirm that sync is enabled, but that may not be relevant -- `wp option pluck woocommerce_stripe_settings pmc_enabled` should show `yes`
* Upgrade to a Stripe build from this branch
* Verify that the `optimized_checkout_element` sub-option is still empty -- `wp option pluck woocommerce_stripe_settings optimized_checkout_element`
* Reconnect to Stripe
* Verify that the `optimized_checkout_element` sub-option is now set to `yes` -- `wp option pluck woocommerce_stripe_settings optimized_checkout_element`

wjrosa:
Good catch, Dale! Thanks. I have updated the main implementation in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4811/commits/3ee6a77495ab82bb1764ca41f5af2370ab995946. I have also included unit tests specific for this behavior.

includes/class-wc-stripe.php
```diff
@@ -370,13 +370,22 @@ public function install() {
...
-				WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
-				WC_Stripe_Logger::warning( 'Settings synchronization eligibility will be re-checked after upgrade' );
+			if ( isset( $stripe_settings['pmc_enabled'] ) ) {
+				// If we have previously disabled settings synchronization, remove the flag after the upgrade,
+				// just to make sure we are still ineligible for settings synchronization.
+				if ( 'no' === $stripe_settings['pmc_enabled'] ) {
+					unset( $stripe_settings['pmc_enabled'] );
+					WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
+					WC_Stripe_Logger::warning( 'Settings synchronization eligibility will be re-checked after upgrade' );
+				} else if ( 'yes' === $stripe_settings['pmc_enabled'] ) {
```
daledupreez: I think this approach also has issues. If we have a merchant who has settings sync enabled but hasn't had the `optimized_checkout_element` sub-option set, upgrading will now result in Optimized Checkout being enabled. AFAICT, we don't have any logic that's only going to run for a new/fresh install.

I had a fairly different thought about how to approach this much earlier in `install()`:
 * Get current value of `wc_stripe_version` (which we're already doing for the comparison against `WC_STRIPE_VERSION`)
 * If the value is empty, we have a new install, so set a new option (or maybe sub-option) like `wc_stripe_optimized_checkout_default_on` to `yes`
 * Leave all other logic in this code as-is -- we only want to take action when we (re)enable settings sync
 * Then in `WC_Stripe_Payment_Method_Configurations::maybe_migrate_payment_methods_from_db_to_pmc()`, if we execute the code path where we set `pmc_enabled` to `yes`, also set `optimized_checkout_element` to `yes` if it wasn't previously set.

I am planning to do something very similar for enabling Amazon Pay on fresh installs as well. Let me know if it would help for me to put together an initial PR. I am now thinking that it could be worth creating a helper method like `init_on_new_install()` that handles initialising some options for managing default behaviours like this.

wjrosa: I see. I'd appreciate it if you could open another PR then. We can move forward from there 👍

wjrosa: The problem here is that if we run the code in `install` just for new installations (`wc_stripe_version` is `false`), `pmc_enabled` will be undefined, and we will never set `wc_stripe_optimized_checkout_default_on` there

wjrosa: What we can do is set a flag/constant when we identify a fresh install, and proceed with the settings update in `maybe_migrate_payment_methods_from_db_to_pmc`.

wjrosa: That wouldn't work either, because the install and the connection method are called different in different flows 🤔

wjrosa: I did a refactor using a new flag in options to identify new installs. See https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4811/commits/7f2b87d996a26d858b344198084f5793e454717f. Let me know what you think. It worked for me

includes/class-wc-stripe.php
```diff
@@ -344,6 +344,11 @@ public function install() {
 				define( 'WC_STRIPE_INSTALLING', true );
 			}
 
+			// Mark as fresh install if no version is set.
+			if ( false === get_option( 'wc_stripe_version' ) && false === get_option( 'wc_stripe_fresh_install' ) ) {
+				update_option( 'wc_stripe_fresh_install', true );
+			}
```
daledupreez: I think I would prefer a more specific name for the option, as it feels like this generic name could result in unexpected or unintended uses in future. Can we rename this to something explicitly mentioning Optimized Checkout? Given that is the only thing we are using it for, I think it's better to use a more specific option name.
```suggestion
			if ( false === get_option( 'wc_stripe_version' ) && false === get_option( 'wc_stripe_optimized_checkout_default_on' ) ) {
				update_option( 'wc_stripe_optimized_checkout_default_on', true );
			}
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Update - Makes the Optimized Checkout Suite feature enabled by default for all new installations
```
daledupreez: Nit: I think we can simplify the description a bit.
```suggestion
* Update - Enable the Optimized Checkout Suite feature for all new installations
```

daledupreez (commented):
🙌  This is pretty much exactly what I was thinking, @wjrosa! My only minor comment is that we should use a more specific option name to match the specificity of the current use case. I'll re-test once we confirm the option name we want to use.

includes/class-wc-stripe.php
wjrosa: I agree. My initial idea was to keep it generic in case we wanted to use this for Amazon Pay later as well, but since we are seeing if it can be enabled automatically on the Stripe side, we can keep it specific

includes/connect/class-wc-stripe-connect.php
```diff
@@ -267,6 +267,10 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 			$options[ $prefix . 'secret_key' ]          = $secret_key;
 			$options[ $prefix . 'connection_type' ]     = $type;
 			$options['pmc_enabled']                     = 'connect' === $type ? 'yes' : 'no'; // When not connected via Connect OAuth, the PMC is disabled.
+			if ( 'connect' === $type && get_option( 'wc_stripe_optimized_checkout_default_on' ) ) {
+				$options['optimized_checkout_element'] = 'yes';
+				delete_option( 'wc_stripe_optimized_checkout_default_on' );
+			}
```
daledupreez: I think we should delete the option outside of the connect block, as we won't correctly clean up the option.
```suggestion
			$should_default_optimized_checkout_on = get_option( 'wc_stripe_optimized_checkout_default_on' );
			// Clean up the option.
			delete_option( 'wc_stripe_optimized_checkout_default_on' );
			if ( 'connect' === $type && $should_default_optimized_checkout_on ) {
				$options['optimized_checkout_element'] = 'yes';
			}
```

includes/connect/class-wc-stripe-connect.php
```diff
@@ -267,6 +267,10 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 			$options[ $prefix . 'secret_key' ]          = $secret_key;
 			$options[ $prefix . 'connection_type' ]     = $type;
 			$options['pmc_enabled']                     = 'connect' === $type ? 'yes' : 'no'; // When not connected via Connect OAuth, the PMC is disabled.
```
daledupreez, line 269: Thought for a follow-up/separate PR: maybe we should default to `''` or not even set `pmc_enabled` when we don't have ` connect` type.
```suggestion
			$options['pmc_enabled']                     = 'connect' === $type ? 'yes' : ''; // When not connected via Connect OAuth, defer the decision until we try to get a valid PMC.
```

daledupreez (changes requested):
Testing went smoothly, for both a simulated new install, reconnection, and simulated upgrade. However, I did notice an issue we should address in this PR.

daledupreez (approved):
Thanks for the tweak! The update looks good to me based on inspection. 🚀

PR #4813, opened 2025-11-18
Removing ECE enabled feature flag usages

wjrosa:

In this cleaning-up PR, I am removing all usages of the `is_stripe_ece_enabled` feature flag method since this FF has been enabled for a while, and the method is now deprecated.

## Testing instructions

Code review. Check if the tests are still passing. Perform some basic smoke testing with ECE and confirm no regression was introduced.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-amazon-pay-controller.php
+++ b/includes/admin/class-wc-stripe-amazon-pay-controller.php
@@ -41,7 +41,6 @@ class WC_Stripe_Amazon_Pay_Controller {
 		$params          = [
 			'key'                    => WC_Stripe_Mode::is_test() ? $stripe_settings['test_publishable_key'] : $stripe_settings['publishable_key'],
 			'locale'                 => WC_Stripe_Helper::convert_wc_locale_to_stripe_locale( get_locale() ),
-			'is_ece_enabled'         => true,
 			'taxes_based_on_billing' => wc_tax_enabled() && 'billing' === get_option( 'woocommerce_tax_based_on' ),
 		];
 		wp_localize_script(
--- a/includes/admin/class-wc-stripe-amazon-pay-controller.php
+++ b/includes/admin/class-wc-stripe-amazon-pay-controller.php
@@ -41,7 +41,7 @@ class WC_Stripe_Amazon_Pay_Controller {
 		$params          = [
 			'key'            => WC_Stripe_Mode::is_test() ? $stripe_settings['test_publishable_key'] : $stripe_settings['publishable_key'],
 			'locale'         => WC_Stripe_Helper::convert_wc_locale_to_stripe_locale( get_locale() ),
-			'is_ece_enabled' => WC_Stripe_Feature_Flags::is_stripe_ece_enabled(),
+			'is_ece_enabled' => true,
 		];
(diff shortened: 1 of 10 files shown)


includes/admin/class-wc-stripe-amazon-pay-controller.php
```diff
@@ -41,7 +41,7 @@ public function admin_scripts() {
 		$params          = [
 			'key'                    => WC_Stripe_Mode::is_test() ? $stripe_settings['test_publishable_key'] : $stripe_settings['publishable_key'],
 			'locale'                 => WC_Stripe_Helper::convert_wc_locale_to_stripe_locale( get_locale() ),
-			'is_ece_enabled'         => WC_Stripe_Feature_Flags::is_stripe_ece_enabled(),
+			'is_ece_enabled'         => true,
```
diegocurbelo: Should we also remove this? It's only used once in one file: https://github.com/woocommerce/woocommerce-gateway-stripe/blob/2c0ad609a52d2ca40b44562570340044e7720aec/client/entrypoints/express-checkout-settings/express-checkout-settings-section.js#L124-L125

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -528,7 +528,7 @@ public function javascript_params() {
 		$stripe_params['saveAppearanceNonce'] = wp_create_nonce( 'wc_stripe_save_appearance_nonce' );
 
 		// ECE feature flag
-		$stripe_params['isECEEnabled'] = WC_Stripe_Feature_Flags::is_stripe_ece_enabled();
+		$stripe_params['isECEEnabled'] = true;
```
diegocurbelo: This is not used anymore. There is a variable with the same name as this, but it's the one defined in the comment above (the only use of `is_ece_enabled`):
https://github.com/woocommerce/woocommerce-gateway-stripe/blob/2c0ad609a52d2ca40b44562570340044e7720aec/client/entrypoints/express-checkout-settings/express-checkout-settings-section.js#L273-L284

malithsen (approved):
LGTM. We'd want to update the frontend code as well, since we are always returning `true` for this flag now, but that can be done later.

diegocurbelo (approved):
Looks good,  but I think we should also clean-up the frontend, as it only affects one file.

includes/admin/class-wc-stripe-amazon-pay-controller.php
wjrosa: Thanks for pointing this out! Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4813/commits/3ddb09a99be131bc72ff5b988752fc715f9e1c03

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Cool! Remove it as well in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4813/commits/3ddb09a99be131bc72ff5b988752fc715f9e1c03

PR #4816, opened 2025-11-19
Including support for physical free trial subscription products with ECE

wjrosa:
Fixes STRIPE-715

Our current implementation of the Express Checkout Element does not support free-trial subscriptions that require shipping. I am adding support for that here.

We are basically removing the constraint for physical products from the backend. Until now, we could not find reasons why this could break (and was prevented).

## Testing instructions

- Checkout and build this branch on your test environment (`add/include-support-for-physical-free-trials`)
- Connect your Stripe account
- Enable some express payment methods
- Create a physical subscription product with a free trial period
- As a shopper, access the store using a public-accessible URL
- Add this subscription to your cart
- Go to any checkout page
- Confirm express checkout buttons are shown
- As a merchant, confirm you can renew the subscription in the WP-Admin

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (commented):
@wjrosa do you see the following error for a free trial subscription with ECE? I see this error when I click the Google Pay button on both product and checkout pages. Works fine for simple products and subscriptions without a free trial for the same address.

Screenshot 2025-11-20 at 12 27 51 PM

I can continue and pay, but was wondering if this will be an issue in live mode.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Add - Includes support for free trial subscription products that require shipping when using express checkout payment methods
```
daledupreez: Nit: this is pretty verbose, and doesn't feel super-clear to me. How about something like the following?
```suggestion
* Add - Support express checkout for free trial subscription products that require shipping
```

daledupreez (commented):
I am seeing the same issue as @Mayisha when computing shipping options in both Google Pay and Apple Pay.

wjrosa:
I did some deep debugging today and found out the issue is due the subscriptions extension. We have this filter there:
```
		// Remove any subscriptions with a free trial from the initial shipping packages
		add_filter( 'woocommerce_cart_shipping_packages', __CLASS__ . '::set_cart_shipping_packages', -10, 1 );
```
It does what the comment says. If I comment out this filter, it works just fine. That must be why we blocked physical free trials in the first place. I will ask the team about it

wjrosa:
Asked the Chronos team about it in p1764795994520959-slack-C0866MKN6H1

wjrosa:
@Mayisha @daledupreez I decided to remove the filter above in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/commits/860684e370acb68c0ed214ec05887f9d356e12db. Per the discussion in p1764795994520959-slack-C0866MKN6H1, I believe it makes sense. I did some testing, and things are working fine for me. Let me know what you think

Mayisha (commented):
I don't see the previous error notice anymore. 👍 

I have two observations from the functional tests.

- Now I see the shipping options on the Google Pay modal. However, if I select a different shipping option from the modal, it does not persist. In the subscription, the first shipping option is saved and charged during renewal.

Screenshot 2025-12-08 at 2 07 28 PM

Steps to reproduce,
1. Add 3 shipping methods in your store (Express $20, free $0 and flat rate $10)
2. Add a free trial physical subscription to your cart.
3. Try to pay with Google Pay.
4. On the modal, there is the shipping options now and the first shipping method is selected (Express)
5. Change the shipping method and select `free`.
6. Complete payment.
7. As an admin, go to the subscription edit page and check the amount breakdown. You will see express as the shipping method instead of free.
8. Renewal will charge for express shipping method as well.

- In Google Pay modal, the price breakdown is off. 

Screenshot 2025-12-08 at 2 09 23 PM

wjrosa:
Good catch, Mayisha. Fixed the shipping method swapping in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/commits/3ed791ad54f7c8cbd3e955ca57a0e23d775c760b

Mayisha (approved):
Thanks for fixing the issues @wjrosa. 
I don't see the previous issues anymore and I could successfully purchase a physical subscription with a trial period using Google Pay 👍

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1412,6 +1403,11 @@ protected function calculate_shipping( $address = [] ) {
 			}
 		}
 
+		// Remove subscription shipping package filter if there is free trial in the cart to allow the calculation of shipping costs.
+		if ( $this->cart_contains_free_trial() ) {
+			remove_filter( 'woocommerce_cart_shipping_packages', 'WC_Subscriptions_Cart::set_cart_shipping_packages', -10, 1 );
+		}
```
daledupreez, line 1453: Question: what should happen if the cart contains a free shippable trial and a free non-shippable trial?

While that filter currently handles only filtering out free trials, it might change in future. I think it may be worth considering one of the following two approaches to this:
 * Fix this upstream in WooCommerce Subscriptions -- that code should not filter out shippable free trials
 * Change the implementation here so that we run a filter at an earlier priority that identifies any shippable free trials, then runs the filter above, and then runs a later filter that reinstates the shippable free trials

We may need to tackle those things in a follow-up, but I think we should think through the edge cases.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1782,7 +1778,11 @@ public function use_blocks_api() {
 	 * @param array $previous_chosen_methods The previously chosen shipping methods.
 	 */
 	public function maybe_restore_recurring_chosen_shipping_methods( $previous_chosen_methods = [] ) {
-		if ( empty( WC()->cart->recurring_carts ) || ! method_exists( 'WC_Subscriptions_Cart', 'get_recurring_shipping_package_key' ) ) {
+		if (
+			empty( WC()->cart->recurring_carts )
+			|| ! method_exists( 'WC_Subscriptions_Cart', 'get_recurring_shipping_package_key' )
+			|| $this->cart_contains_free_trial()
```
daledupreez: Could you help me to understand why we should _not_ reinstate the chosen shipping method for a free trial? I assume we should do so, so I am a bit confused as to why we are returning early for this case.

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-helper-test.php
```diff
@@ -309,82 +309,6 @@ public function test_hides_ece_if_stripe_gateway_unavailable() {
 		WC()->payment_gateways()->payment_gateways = $original_gateways;
 	}
 
-	/**
-	 * Test should_show_express_checkout_button, free trial logic.
-	 *
-	 * @return void
-	 */
-	public function test_hides_ece_if_free_trial_requires_shipping() {
```
daledupreez, line 310: Should we not invert the test to verify that we show express checkout when a free trial requires shipping?

daledupreez (commented):
This seems to be mostly working as described for the happy path, but I am not seeing the shipping total reflected in the Google Pay modal. I am also not seeing the charge for the shipping. I may be tripping over a webhook configuration issue, but I don't see the charge in the Stripe dashboard.

Screenshot 2025-12-17 at 14 43 50

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-helper-test.php
wjrosa: Good idea. Reverted the test logic in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/commits/7eda5dadf9c32a3f8d8da36ad6a865aee2249632. I also took the opportunity to add the missing return types and PHPDoc in this file.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: The only reason is that this method is causing the bug Mayisha described in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/#pullrequestreview-3550700058.

I am honestly not sure why. So, I decided to skip it strictly for this feature (free trials) to avoid breaking other flows. Maybe @Mayisha can give more context on this, as she introduced it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3429. I understood this part of the method description:

> 	 * When the cart contains multiple subscriptions with different billing periods, customers are able to select different shipping
>	 * methods for each subscription, however, this is not supported when purchasing with Apple Pay and Google Pay as it's
>	 * only concerned about handling the initial purchase.

But I am not sure what this has to do with reverting the shipping method selection 🤔 . I would assume we had to pick a single shipping method or something in that line.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: That's a fair point. I have added a TODO comment in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/commits/bbcb906e82caf21fc87e3206a71167eca4464e16

Let me know if you think this is enough or if we should add more information to it (I thought of adding the link to this comment, but I don't see us doing that much often).

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
Mayisha: From my functional testing, if we do not have the early return for free trial the selected shipping method from the modal does not persist.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1412,6 +1403,12 @@ protected function calculate_shipping( $address = [] ) {
 			}
 		}
 
+		// Remove subscription shipping package filter if there is free trial in the cart to allow the calculation of shipping costs.
+		// TODO: revisit this when WC Subscriptions adds support for free trials with shipping.
```
daledupreez, line 1448: Rather than just having a TODO that is super-local to this code, can we create an issue for WooCommerce Subscriptions (or find an existing one), and then link to it from here? It feels like that would be more actionable for whoever looks at this code in future.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
daledupreez: Maybe we can add a comment to that effect? It feels pretty mysterious otherwise, so having an inline comment to explain the weirdness feels like it would be a good idea.

daledupreez (commented):
Overall, I think this is working as expected, but I tripped over one super-weird UX with Apple Pay, and an inconsistency across Apple Pay and 

##### Apple Pay doesn't show updated total

When I pick a different shipping method, Apple Pay doesn't show me the updated total (which is still 0.00), but shows "Processing".

IMG_4895

##### Shipping methods are and are not free

In the UI, both Apple Pay and Google Pay show the price of the shipping options, but the total for the purchase and the eventual payment both remain free. While that works, it is still pretty confusing for the shopper, as it is not clear what they will actually be charged. I think it may be worth showing the shipping prices as free when there is a free shippable trial in the cart, and we may need a filter to limit the available shipping options for shippable free trials given that shoppers won't be charged for shipping.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
Mayisha: > From my functional testing, if we do not have the early return for free trial the selected shipping method from the modal does not persist.

I would also recommend one of you test and confirm that this is actually the case and not a behavior only in my local for some reason.

wjrosa: > Maybe we can add a comment to that effect? It feels pretty mysterious otherwise, so having an inline comment to explain the weirdness feels like it would be a good idea.

Comment added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/commits/c54b6c5382a8fcbcd915d5b056d8c4dddd153944

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Makes sense. Added WOOSUBS-1485

wjrosa:
> I think it may be worth showing the shipping prices as free when there is a free shippable trial in the cart, and we may need a filter to limit the available shipping options for shippable free trials given that shoppers won't be charged for shipping.

Maybe that can be even more confusing? Since the shopper will be charged for the shipping once the free trial ends. I think it would be better to add a specific notice to the checkout screen or something like that. WDYT?

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1426,6 +1417,12 @@ protected function calculate_shipping( $address = [] ) {
...
+		if ( $this->cart_contains_free_trial() ) {
+			remove_filter( 'woocommerce_cart_shipping_packages', 'WC_Subscriptions_Cart::set_cart_shipping_packages', -10 );
+		}
+
 		$packages = apply_filters( 'woocommerce_cart_shipping_packages', $packages );
 
 		WC()->shipping->calculate_shipping( $packages );
```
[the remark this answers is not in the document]
wjrosa: The priority (Risky 1) is indeed -10.

Addressed the "Risky 2" topic in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/commits/ffdfd0a2d604514e40e917440d734d9493c56faf

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1729,7 +1726,13 @@ public function is_link_enabled() {
 	 * @return void
 	 */
 	public function maybe_restore_recurring_chosen_shipping_methods( $previous_chosen_methods = [] ) {
-		if ( empty( WC()->cart->recurring_carts ) || ! method_exists( 'WC_Subscriptions_Cart', 'get_recurring_shipping_package_key' ) ) {
+		// If there are no recurring carts, or the `WC_Subscriptions_Cart::get_recurring_shipping_package_key` method doesn't exist, or if the cart contains a free trial (which doesn't require shipping), we skip this step.
+		// That's because this logic overrides the selected shipping method with the previously chosen shipping method.
```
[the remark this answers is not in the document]
wjrosa: Decided to just remove that part in parentheses instead (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/commits/53fd8871d4b37b1ef991f428f6d4510988e0f699)

daledupreez (approved):
The changes LGTM overall and test well.

There are some testing suggestions that may be worth incorporating, but I don't see any blockers.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1729,7 +1731,13 @@ public function is_link_enabled() {
...
+		if (
+			empty( WC()->cart->recurring_carts )
+			|| ! method_exists( 'WC_Subscriptions_Cart', 'get_recurring_shipping_package_key' )
+			|| $this->cart_contains_free_trial()
+		) {
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4816/commits/6888fc16b4500218bf0233e14bdfe05e2509a4ee

PR #4818, opened 2025-11-20
Generate the OAuth Connect URL on click instead of page load

### LINKED ISSUE #4800: Immediate invalidation of keys after reconnection
### Description

We have had one merchant report of a really confusing and frustrating situation. After reconnecting to Stripe, they have one or two successful API requests using the newly received key, whereafter Stripe API requests fail due to the secret key having been invalidated. The specific error code that gets returned is `platform_api_key_expired`.

### Steps to reproduce

The merchant that can reproduce this issue triggered a reconnection from the Stripe Settings tab, and could never get the reconnection to "stick".

### Expected behavior

The reconnection should complete and the key should remain valid.

### System report

<details>

</details>

### Additional context

Original report: 
* https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4729
* STRIPE-745
Internal issue: 10365825-zen

diegocurbelo:
Fixes STRIPE-745
Fixes #4800

Based on the feedback received in the issue STRIPE-745, we ruled out the problem being the transient, and the log indicates that multiple calls to `generate_outh_url` are replacing the `stored state` before the Connect flow completes, and the merchant is redirected back.

This PR changes the logic to only generate the connect oauth url, when clicking the connect button instead of pre-generating them on page load.

It also extracts the connect button logic into a new `ConnectButton` component to avoid duplication, except for the re-authenticate logic.

## Testing instructions

1. Add the filter to enable verbose logging:
    `add_filter('wc_stripe_is_verbose_debug_mode_enabled', '__return_true');`
2. Navigate to the plugin settings (`WooCommerce > Settings > Payments > Stripe`) and then into the `Settings` tab
3. Click the `Configure connection` button
4. Check that there are no log entries with the `OAuth` prefix
5. Click the `Create or connect a test account` button and check you are redirected to Stripe to connect your account
6. Check the logs, and this time, there should be entries for the generated URL (only one URL is generated)
    Screenshot 2025-11-21 at 17 21 32
7. Complete the connect flow, and make sure the account is correctly connected.
8. Disconnect your account:
    Screenshot 2025-11-21 at 17 24 13
9. Check that you are redirected to the onboarding page:
    Screenshot 2025-11-21 at 17 25 05
10. Check that there are no new OAuth entries in the log after disconnecting
11. Open a second page of the plugin settings (right-click on the browser tab and then duplicate)
12. Go back to the first settings tab, and click `Create or connect a test account`
13. Check the logs, there should be only one new entry `OAuth: Generated test connect URL`:
    Screenshot 2025-11-21 at 17 29 03
14. In the tab duplicated in step 11, refresh the page
15. Check in the logs that no additional URLs were generated
16. Go back to the Stripe flow tab, and connect your account.
17. Confirm that the account was successfully connected.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/connect/class-wc-stripe-connect.php
```diff
@@ -133,7 +133,7 @@ public function connect_oauth( $state, $code, $type = 'connect', $mode = 'live'
 				return $response;
 			}
 
-			delete_transient( 'wcs_stripe_connect_state_' . $mode );
+			WC_Stripe_Database_Cache::delete( 'wcs_stripe_connect_state_' . $mode );
```
daledupreez: I think we should remove these changes from this PR -- I believe they are from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4807.

daledupreez (commented):
I have a major concern with this implementation: I have tripped over situations where browsers disallow navigation events that occur after an HTTP request that takes too long to return. In my recollection, "too long" was around 1 second, but I can't find a source for the actual timeout value (if any), and whether those rules still apply.

Might it be simpler to cache the URLs for the current user for a few minutes on the server and then re-use the value across multiple page loads as needed? That way we can make multiple requests to the server, but we don't regenerate the URLs every time. And we may not need to rewrite a lot of the current logic as much as in this PR.

daledupreez:
@diegocurbelo, I put together a quick first pass of my thinking in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4834. I haven't tested it _at all_, but wanted to share a concrete exploration.

malithsen:
> I have a major concern with this implementation: I have tripped over situations where browsers disallow navigation events that occur after an HTTP request that takes too long to return. In my recollection, "too long" was around 1 second, but I can't find a source for the actual timeout value (if any), and whether those rules still apply.

@daledupreez @diegocurbelo This might be the [Transient Activation](https://developer.mozilla.org/en-US/docs/Glossary/Transient_activation) restrictions. It mainly applies to `window.open`. It doesn't seem to apply to `window.location.assign()` which is what's being used here. FWIW, I also tested this PR by adding delays manually ( via `setTimeout`) and tested across Chrome/Firefox and didn't notice an issue.

malithsen (approved):

client/settings/payment-settings/promotional-banner/re-connect-account-banner.js
```diff
@@ -14,32 +15,57 @@ import {
...
+			{}
+		);
+		recordEvent( 'wcstripe_reconnect_button_click', {
+			source: 're-connect-account-banner',
+			mode,
+		} );
+
+		setIsLoading( true );
+
+		try {
```
wjrosa, line 39: I just wanted to mention here what I said during our team call today: one way to avoid timeout issues is to add some retry logic. Move the logic inside the block to a helper function and call it a couple of times until we get a valid response. It won't look pretty, but might solve the issue

client/settings/connect-stripe-account/index.js
```diff
@@ -74,80 +63,37 @@ const ConnectStripeAccount = ( { oauthUrl, testOauthUrl } ) => {
...
+					} ) }
+				</TermsOfServiceText>
+				<p className="woocommerce-stripe-auth__help">
+					{ __(
+						'Some payment methods are automatically enabled when you connect your account. Review your Payment Methods settings for details.',
+						'woocommerce-gateway-stripe'
+					) }
+				</p>
+				<ButtonWrapper>
+					<ConnectButton testMode={ false } buttonVariant="primary" />
```
daledupreez: A subtle side-effect of this change is that we will _always_ show the option to connect a live/production account, regardless of the environment we're in, e.g. non-HTTPS and/or local URLs.
Can we replicate the original logic for this? It doesn't feel like a good idea to allow that.

client/settings/payment-settings/promotional-banner/re-connect-account-banner.js
```diff
@@ -14,32 +15,57 @@ import {
...
+				method: 'POST',
+				data: {
+					action: 'wc_stripe_get_oauth_url',
+					mode,
+					nonce: wc_stripe_settings_params.oauth_nonce, // eslint-disable-line camelcase
+				},
 			} );
-			window.location.assign( oauthUrl );
-		} else {
+
```
daledupreez, line 49: Minor nit: is it worth setting the loading state to false here regardless of the result?
```suggestion

			setIsLoading( false );
```

client/settings/stripe-auth-account/connect-button.js
```diff
@@ -0,0 +1,86 @@
...
+			} ) }
+		</InlineNotice>
+	) : (
+		<Button
+			variant={ buttonVariant }
+			onClick={ handleClick }
+			text={ buttonText }
+			disabled={ isLoading }
+			isBusy={ isLoading }
+		/>
```
daledupreez: By using the ` error ? ( <InlineNotice>...</InlineNotice> ) : `<Button />;` construction, when we show the error message, the user can't retry because the button is no longer visible. While errors shouldn't occur in these cases, it feels like we should have a better, more graceful failure mode. Is it worth showing the inline notice _and_ the button?

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -140,50 +140,29 @@ private function is_payments_onboarding_task_completed(): bool {
...
-		$settings      = WC_Stripe_Helper::get_stripe_settings();
-		$has_live_keys = ! empty( $settings['publishable_key'] ) && ! empty( $settings['secret_key'] );
-		$has_test_keys = ! empty( $settings['test_publishable_key'] ) && ! empty( $settings['test_secret_key'] );
-
-		// If no keys at all, we need OAuth URLs for new account setup
-		if ( ! $has_live_keys && ! $has_test_keys ) {
-			return true;
+		$mode = isset( $_POST['mode'] ) ? sanitize_text_field( wp_unslash( $_POST['mode'] ) ) : 'live';
+		if ( ! in_array( $mode, [ 'live', 'test' ], true ) ) {
+			$mode = 'live';
```
daledupreez: Should we not default to test mode when we have a caller behaving poorly? It feels like that should result be less problematic by default for cases where we get a nonce leaked etc.

client/settings/connect-stripe-account/index.js
daledupreez: Might it make sense to show an inline notice when the document location or AJAX URL are not using SSL? And then show the test connection button as primary?

client/settings/connect-stripe-account/index.js
```diff
@@ -74,80 +63,37 @@ const ConnectStripeAccount = ( { oauthUrl, testOauthUrl } ) => {
...
+					<ConnectButton
+						testMode={ true }
+						buttonVariant="secondary"
+					/>
+				</ButtonWrapper>
```
daledupreez, line 121: On my screen, there is no space between the buttons. Not sure if that's a new issue, but we should add some kind of gap/spacing here.

Image

daledupreez (commented):
Functionally, this is generally working, but the experience on test sites is not great. There is also a small layout issue to address.

client/settings/stripe-auth-account/connect-button.js
diegocurbelo: That inline error is used in the reconnect modal, and as we are re-using the button now I added the option to handle the error at the parent level, restoring the previous notice above the buttons:
Screenshot 2025-12-08 at 01 32 55

client/settings/connect-stripe-account/index.js
diegocurbelo: I restored the error notice, so if the store is not using SSL the merchant will get this error... but the button is still available; I can check the URL and not render it if it's not using HTTPS, or maybe the "hiding" the button is more confusing, I don't have a strong opinion either way 🤷🏼 

Screenshot 2025-12-08 at 01 32 55

client/settings/payment-settings/promotional-banner/re-connect-account-banner.js
diegocurbelo: We want to keep the button `disbaled` while we redirect (`isLoading === true`)... to avoid the possibility of additional clicks triggering another request and generating multiple URLs (and triggering the state invalidation bug)

client/settings/connect-stripe-account/index.js
diegocurbelo: Added the same separation as the promotion banners CTA buttons in [2d2a05e](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4818/commits/2d2a05e92e22ca4863c18854db001d802c7c8178)

includes/admin/class-wc-stripe-settings-controller.php
diegocurbelo: Defaulting to `test` makes sense, and the frontend always sends the mode, so it has no impact on the normal flows... changed in [16a3d85](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4818/commits/16a3d85fc40f0d669e37c2c08a67c5cf811440ee)

client/settings/payment-settings/promotional-banner/re-connect-account-banner.js
diegocurbelo: We want the process to be as deterministic as possible, minimizing the chances of generating multiple valid URLs, adding a retry logic means we need to cancel "in-flight" requests... to avoid processing a response arriving after we already sent another request (which would trigger the original error).

client/settings/connect-stripe-account/index.js
```diff
@@ -30,11 +31,19 @@ const TermsOfServiceText = styled.p`
...
+
 const ButtonWrapper = styled.div`
 	align-items: center;
 	display: flex;
 	flex-wrap: wrap;
 
+	> :first-child {
+		margin-right: 8px;
+	}
+
```
daledupreez, line 43: Nit: I think it would be clearer to use `column-gap` or even `gap` instead so it's a bit clearer what we're aiming for. I also think we should aim for a larger gap by default.
```suggestion
	flex-wrap: wrap;
	column-gap: 12px;

```

client/settings/stripe-auth-account/connect-button.js
```diff
@@ -57,22 +64,28 @@ const ConnectButton = ( { testMode, buttonVariant } ) => {
...
-				),
-				components: {
-					br: <br />,
-					Link: (
-						<ExternalLink href="https://woocommerce.com/document/stripe/setup-and-configuration/connecting-to-stripe/" />
+	// If onErrorChange is provided, parent handles error display
+	// Otherwise, show error inline for backward compatibility
+	if ( ! onErrorChange && error ) {
+		return (
+			<InlineNotice isDismissible={ false } status="error">
```
daledupreez: Non-blocking nit: duplicating this notice content feels like it could lead to trouble in the long term due to the content drifting. It may be worth shifting into a component so both locations can re-use exactly the same source code.

client/settings/connect-stripe-account/index.js
daledupreez: I think showing the button as active and then showing an error after clicking it feels like a poor UX.

I think it would be better to disable the button and show a pre-emptive warning/notice when we're running over HTTP. However, the current implementation makes it hard to communicate the type of warning/notice to show. But I still think it would be better to communicate that a live connection is not available before we make the merchant take an action that is guaranteed to fail.

daledupreez (commented):
This is working as described, but I still have a concern with how we're handling the UX for live mode in non-HTTPS contexts.

wjrosa (approved):
Looks good to me! 👍 I second Dale on the UX suggestion

client/settings/stripe-auth-account/connect-button.js
diegocurbelo: Fixed in [d7d0e06](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4818/commits/d7d0e06112a4b07dd2ff6c2f2d6abc924a7650ac)

client/settings/connect-stripe-account/index.js
diegocurbelo: Changed in [d7d0e06](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4818/commits/d7d0e06112a4b07dd2ff6c2f2d6abc924a7650ac)

client/settings/connect-stripe-account/index.js
diegocurbelo: That's fair... changed it to a disabled button + tooltip in [8845656](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4818/commits/88456561b488ccce09e9a7e610d57f1b88435467)

Screenshot 2025-12-08 at 11 28 28

client/settings/connect-stripe-account/index.js
```diff
@@ -58,9 +55,9 @@ const ButtonWrapper = styled.div`
 const ConnectStripeAccount = () => {
 	const [ hasError, setHasError ] = useState( false );
 
-	const handleErrorChange = ( error ) => {
+	const handleErrorChange = useCallback( ( error ) => {
 		setHasError( !! error );
-	};
+	}, [] );
```
daledupreez: Nit: should this not depend on `setHasError`?
```suggestion
	}, [ setHasError ] );
```

client/settings/stripe-auth-account/connect-button.js
```diff
@@ -34,6 +36,10 @@ const ConnectButton = ( { testMode, buttonVariant, onErrorChange } ) => {
 	const handleClick = async () => {
 		setIsLoading( true );
 		setError( null );
+		// Clear parent error when any button is clicked
+		if ( onErrorChange ) {
+			onErrorChange( null );
+		}
```
daledupreez, line 42: Nit: I preferred the use of the effect that was based on the value of `error`. But this also works.

client/settings/connect-stripe-account/index.js
daledupreez: That looks great! 😍

daledupreez (approved):
Thanks for those updates! This looks much better to me!

PR #4820, opened 2025-11-21
Allow filtering the state when "normalize_state" is called for express payments

### LINKED ISSUE #4819: Allow filtering the address on "normalize_address" from express payments
**Context:**

Whenever I mention "Apple Pay", "Google Pay" might also be valid.

I have a plugin that adds the Portuguese districts (states) to WooCommerce.
If the shop owner decides to set shipping zones using those "Portuguese States", the Apple Pay payment modal won't request a Portuguese State, so the shipping state will be sent blank to `wc_ajax_wc_stripe_get_shipping_options`, and there's an error of the address not being valid:

Image

Which I can fix with the [`wc_stripe_payment_request_shipping_posted_values`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/10.1.0/includes/payment-methods/class-wc-stripe-payment-request.php#L1243) filter (since I can easily convert the postcode to the required state in my plugin).

The next step for Apple Pay is to call `wc_stripe_normalize_address`, in which Stripe [makes some corrections to the address](ajax_normalize_address), based on exceptions, in the [`normalize_state`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/10.1.0/includes/payment-methods/class-wc-stripe-express-checkout-helper.php#L1064) and [`fix_address_fields_mapping`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/10.1.0/includes/payment-methods/class-wc-stripe-express-checkout-helper.php#L1273C18-L1273C44) methods.

In these methods, there are no filters that allow 3rd party developers to manipulate the address, which makes the address go back to the original, without the new state that we were able to fix on get_shipping_options.

The payment seems to go through, the Apple Pay dialog is closed, and the final `/wp-json/wc/store/v1/checkout?_locale=site` call is made, which fails with [`woocommerce_rest_invalid_shipping_option`](https://github.com/woocommerce/woocommerce/blob/10.3.5/plugins/woocommerce/src/StoreApi/Utilities/OrderController.php#L665), I suspect because the fact that there's no state makes the chosen shipping method not be valid.

**Request:**

Add filter to both the [`normalize_state`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/10.1.0/includes/payment-methods/class-wc-stripe-express-checkout-helper.php#L1064) and [`fix_address_fields_mapping`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/10.1.0/includes/payment-methods/class-wc-stripe-express-checkout-helper.php#L1273C18-L1273C44) methods to allow 3rd party developers to manipulate the address, similar to [`wc_stripe_payment_request_shipping_posted_values`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/10.1.0/includes/payment-methods/class-wc-stripe-payment-request.php#L1243).

Of course, in `normalize_state`, this needs to be done for `$billing_postcode`, `$billing_state`, `$shipping_postcode`, and `$shipping_state` before calling the `get_normalized_state` method, or else this [will fail immediately](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/10.1.0/includes/payment-methods/class-wc-stripe-express-checkout-helper.php#L1215).

Further down the process, the final `/wp-json/wc/store/v1/checkout?_locale=site` call is made, and it fails with [`woocommerce_rest_invalid_shipping_option`](https://github.com/woocommerce/woocommerce/blob/10.3.5/plugins/woocommerce/src/StoreApi/Utilities/OrderController.php#L665C5-L665C45), probably because the fact that the state is now empty makes the chosen shipping method not valid.

Actually, I can confirm that's the reason, as I've forced a state on normalize_state, after your Hong Kong exceptions, and the checkout went through without any glitch:

Image

This is precisely where we need the filters, so that 3rd party developers can implement exceptions just like you did for Hong Kong.

webdados:
Issue: https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4819

Fixes #4819

See the linked issue.

## Testing instructions

As mentioned in the linked issue, we can filter the address at the choosing shipping method step, but not later in the process. If a similar filter is added here, we can override the state and allow the checkout process to finish.

### Changelog entry

Allow filtering the state when "normalize_state" is called for express payments

### DIFF
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -1127,6 +1127,10 @@ class WC_Stripe_Express_Checkout_Helper {
 			}
 		}
 
+		// Allow 3rd parties to hook into state normalization process.
+		$billing_state  = apply_filters( 'wc_stripe_normalize_billing_state', $billing_state, $billing_country, $data );
+		$shipping_state = apply_filters( 'wc_stripe_normalize_shipping_state', $shipping_state, $shipping_country, $data );
+
 		// Finally we normalize the state value we want to process.
 		if ( $billing_state && $billing_country ) {
 			$data['billing_address']['state'] = $this->get_normalized_state( $billing_state, $billing_country );
--- a/includes/payment-methods/class-wc-stripe-payment-request.php
+++ b/includes/payment-methods/class-wc-stripe-payment-request.php
@@ -1540,6 +1540,10 @@ class WC_Stripe_Payment_Request {
 			}
 		}
 
+		// Allow 3rd parties to hook into state normalization process.
+		$billing_state  = apply_filters( 'wc_stripe_normalize_billing_state', $billing_state, $billing_country, $data );
+		$shipping_state = apply_filters( 'wc_stripe_normalize_shipping_state', $shipping_state, $shipping_country, $data );
+
 		// Finally we normalize the state value we want to process.
 		if ( $billing_state && $billing_country ) {
 			$_POST['billing_state'] = $this->get_normalized_state( $billing_state, $billing_country );


daledupreez (commented):
Thanks for the PR, @webdados!

I was literally responding to our issue when I noticed the PR.

Might it make sense to have the filter live a bit higher up the stack so it's a bit more general? I was going to propose adding a `wc_stripe_express_checkout_normalize_address` filter to `WC_Stripe_Express_Checkout_Ajax_Handler->ajax_normalize_address()` that filters `$normalized_data` and accepts the original `$data` as an argument.

Pushing the filter higher up the stack could make it easier to override any logic, and be less specific.

Would that work for you? Or would the more specific filters be better?

webdados:
That would work just fine, also.

Feel free to ignore/reject the PR. You know your codebase better, and it works for me the same.

webdados:
@daledupreez, are you taking this and implementing the filter as you mentioned?

daledupreez:
Apologies for the missed ping, @webdados! I did implement the general filter in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4837, which was released in 10.2.0 earlier in December.

webdados:
Closed by https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4837

PR #4832, opened 2025-11-26
Use current payment method configuration for Optimized Checkout

### LINKED ISSUE #4831: Optimized Checkout does not work with non-connected accounts
### Description

When a merchant connects to Stripe in such a way that they are not connected to the WooCommerce platform account in Stripe, the merchant won't have a payment method configuration (PMC) that has the WooCommerce PMC listed as the parent PMC. We still allow those merchants to synchronize their payment method list with Stripe. However, the current implementation for Optimized Checkout uses the parent PMC's ID. This breaks when the merchant doesn't have a child PMC linked to that parent.

Additionally, if a merchant uses the `wc_stripe_preselect_payment_method_configuration` filter to specify an alternative PMC that isn't a child PMC, but they are connected and have a child PMC, the child PMC will be used for Optimized Checkout instead of the specified one.

### Steps to reproduce

1. Set up a situation such that you have a Stripe account with settings sync enabled, but the account is not connected to the WooCommerce platform account, and doesn't have a child PMC
  * Alternatively, use the `wc_stripe_preselect_payment_method_configuration` filter to specify a non-child PMC
2. Enable Optimized Checkout
3. Add a product to the cart and navigate to checkout
4. The Optimized Checkout won't load correctly because the payment method configuration can't be found
  * If you specified an alternative PMC for a connected account, you would not get an error, but you should see the payment methods from the child PMC instead of the specified PMC.

### Expected behavior

Optimized Checkout should just work and use the correct PMC.

daledupreez:
Fixes STRIPE-830
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4831

This PR ensures that we use the current payment method configuration (PMC) for Optimized Checkout. This fixes two sets of problems stemming from our current logic that uses the WooCommerce parent PMC ID:
 * Optimized Checkout doesn't work for merchants who are not connected to the WooCommerce platform account, as they don't have a child PMC that can be used
 * If merchants are connected and specifying an alternative, non-child PMC, as can be done using the `wc_stripe_preselect_payment_method_configuration` filter, Optimized Checkout will use the child PMC instead of the specified PMC

## Testing instructions

It's a bit hard to test the case where an account is not connected, but we can verify that Optimized Checkout continues to work and the second case where an alternative PMC is specified.

* Set up a test store and connect to Stripe
* Find the PMCs for your test account, and identify a PMC that doesn't have a parent -- make a note of which payment methods are enabled
  - You can do this via the Stripe dashboard or via the debugging logs when we request all PMCs
* Enable Optimized Checkout
* Implement the following snippet to pick the alternative PMC:
```php
function test_pick_non_child_test_pmc( $pmc_id, $test_mode ) {
	if ( ! $test_mode ) {
		return $pmc_id;
	}
	
	return 'pmc_YourPMC';
}
add_filter( 'wc_stripe_preselect_payment_method_configuration', 'test_pick_non_child_test_pmc', 10, 2 );
```
* You may need to clear the database cache to pull down the new PMC (assuming you're in test mode): `wp option delete wcstripe_cache_test_payment_method_configuration`
* If needed, pick different payment methods in the Stripe settings so the list diverges from the child PMC list
* As a shopper, add a product to your cart
* Navigate to the block checkout
* Note that the list of payment methods shown in Optimized Checkout is from the child PMC, and not your custom PMC
* Now navigate to the classic checkout, and verify the same mismatch
* Check out this branch and build the updated UI code: `git checkout fix/use-of-parent-pmc-for-optimized-checkout && npm run build`
* As a shopper, navigate to block checkout
* Confirm that the list of payment methods in Optimized Checkout matches those from your custom PMC
* Now navigate to classic checkout
* Confirm that you see the same list of payment methods in Optimized Checkout

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -334,6 +334,24 @@ class WC_Stripe_Payment_Method_Configurations {
 	}
 
 	/**
+	 * Get the current payment method configuration ID.
+	 *
+	 * @return string|null The payment method configuration ID when settings sync is enabled and we have a PMC. Null otherwise.
+	 */
+	public static function get_configuration_id(): ?string {
+		if ( ! self::is_enabled() ) {
+			return null;
+		}
+
+		$primary_configuration = self::get_primary_configuration();
+		if ( ! $primary_configuration || empty( $primary_configuration->id ) ) {
+			return null;
+		}
+
+		return (string) $primary_configuration->id;
+	}
+
+	/**
 	 * Get the UPE available payment method IDs.
 	 *
 	 * @return array
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -534,8 +534,6 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Gateway_Stripe {
 		$stripe_params['isOCEnabled'] = $this->oc_enabled;
 		$stripe_params['OCLayout']    = $this->get_option( 'optimized_checkout_layout', self::OPTIMIZED_CHECKOUT_DEFAULT_LAYOUT );
 
-		// Single Payment Element payment method parent configuration ID
-		$stripe_params['paymentMethodConfigurationParentId'] = WC_Stripe_Payment_Method_Configurations::get_parent_configuration_id();
 		if ( $this->oc_enabled ) {
 			$stripe_params['paymentMethodConfigurationId'] = WC_Stripe_Payment_Method_Configurations::get_configuration_id();
 		}
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -536,6 +536,9 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Gateway_Stripe {
 
 		// Single Payment Element payment method parent configuration ID
 		$stripe_params['paymentMethodConfigurationParentId'] = WC_Stripe_Payment_Method_Configurations::get_parent_configuration_id();
+		if ( $this->oc_enabled ) {
+			$stripe_params['paymentMethodConfigurationId'] = WC_Stripe_Payment_Method_Configurations::get_configuration_id();
+		}
 
 		// Checking for other BNPL extensions.
 		$stripe_params['hasAffirmGatewayPlugin'] = WC_Stripe_Helper::has_gateway_plugin_active( WC_Stripe_Helper::OFFICIAL_PLUGIN_ID_AFFIRM );
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -532,9 +532,9 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Gateway_Stripe {
 
 		// Optimized Checkout feature flag + setting.
 		$stripe_params['isOCEnabled'] = $this->oc_enabled;
-		$stripe_params['OCLayout']    = $this->get_option( 'optimized_checkout_layout', self::OPTIMIZED_CHECKOUT_DEFAULT_LAYOUT );
 
 		if ( $this->oc_enabled ) {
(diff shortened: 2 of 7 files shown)


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -333,6 +333,24 @@ public static function get_parent_configuration_id() {
...
+			return null;
+		}
+
+		$primary_configuration = self::get_primary_configuration();
+		if ( ! $primary_configuration || empty( $primary_configuration->id ) ) {
+			return null;
+		}
+
+		return (string) $primary_configuration->id;
+	}
```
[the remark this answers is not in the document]
daledupreez: Added in 7f112472f6d541c761f40ca3ebc2581ec4c47d16.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -536,6 +536,9 @@ public function javascript_params() {
 
 		// Single Payment Element payment method parent configuration ID
 		$stripe_params['paymentMethodConfigurationParentId'] = WC_Stripe_Payment_Method_Configurations::get_parent_configuration_id();
+		if ( $this->oc_enabled ) {
+			$stripe_params['paymentMethodConfigurationId'] = WC_Stripe_Payment_Method_Configurations::get_configuration_id();
+		}
```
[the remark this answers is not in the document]
daledupreez: I am not as sure that this logic needs to be unit tested. Happy to take feedback on that.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -536,6 +536,9 @@ public function javascript_params() {
 
 		// Single Payment Element payment method parent configuration ID
 		$stripe_params['paymentMethodConfigurationParentId'] = WC_Stripe_Payment_Method_Configurations::get_parent_configuration_id();
```
wjrosa: This param (`paymentMethodConfigurationParentId`) is used only when OC is enabled as well. We could either leave both outside the condition below or inside it for consistency (but it does not make a big difference, as it is just ignored when OCS is disabled).

wjrosa (approved):
Code is good and works as intended. Left an optional suggestion 👍

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: My personal preference would be to avoid sending unnecessary data over the wire whenever possible. Should we push both into the conditional?

wjrosa: Yes, this would keep it consistent with your change

wjrosa: We can consider doing the same for:
```php
		$stripe_params['OCLayout']    = $this->get_option( 'optimized_checkout_layout', self::OPTIMIZED_CHECKOUT_DEFAULT_LAYOUT );
```

diegocurbelo: Now that we are sending the actual payment method_configuration_id (primary_configuration), we can remove this since we no longer use it in the checkout.

In other words, if `paymentMethodConfigurationId` is null, it means we don't have a PMC, so rendering the elements with `paymentMethodConfigurationParentId` will result in an error

diegocurbelo (approved):
The change makes sense, and it tests well.

[nit] I think we should remove `paymentMethodConfigurationParentId` from `javascript_params` as we are now rendering the elements directly with the merchant pmc_id.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: Good catch! I will remove it from the back end and the front end logic.

daledupreez: Removed in 1a5a7ce5809a62998f7a9b9f1c22d2806994155c.

daledupreez: As per Diego's [comment below](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4832#discussion_r2569936788), we should not use the parent PMC ID in any situation, so I've removed it in 1a5a7ce5809a62998f7a9b9f1c22d2806994155c. I've then gone ahead and moved `OCLayout` into the condition in 19d5a4e23551c5f4ee695cf557ec3a0a8fa403c6.

PR #4835, opened 2025-11-28
Remove the saved oauth_connect_state immediately after validating it

### LINKED ISSUE #4836: Keys can be revoked after OAuth connection flow due to duplicate requests
### Description

As part of investigating specific issues with https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4729, @diegocurbelo identified a specific problem with the way we're handling the redirects back from Stripe to the store. The issue is that in some situations, we end up making two API calls to Stripe to get the newly issued key. The first API call completes successfully, but the second API call fails because the code we're using to get the key has already been used. When that failure occurs, Stripe immediately revokes the key issued in the first request.

### Steps to reproduce

We're not sure what specific configuration on stores triggers this issue, but the steps are as follows:
1. Go to the Stripe settings for your store
2. Follow the flow to reconnect to Stripe
3. From the Stripe UI, get redirected back to the store
4. Get weird behaviour after being redirected back

### Expected behavior

We should only make one API request, as the second one is guaranteed to fail.

</details>

diegocurbelo:
Fixes STRIPE-831
Fix https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4836

Attempting to exchange an `authorization_code` for a `token` multiple times results in the previously exchanged token to be revoked.

> This authorization code has already been used. All tokens issued with this code have been revoked.

This PR removes the `saved_state` cache entry immediately after validating it, preventing multiple concurrent requests from causing the `access_token` (secret_key) to be revoked.

More info in this issue comment: https://linear.app/a8c/issue/STRIPE-745/persistent-reconnection-notice-with-v1000#comment-61a81388

## Testing instructions

Code review should be enough.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,7 +3,7 @@
 = 10.2.0 - xxxx-xx-xx =
 * Dev - Remove all references to the UPE-enabled feature flag
 * Dev - Removing all usages of the `is_stripe_ece_enabled` feature flag method
-* Dev - Expands the Stripe Order Helper class to handle mandate ID, Multibanco data, refund status, card brand, charge captured flag, status final flag, and the refund failure reason 
+* Dev - Expands the Stripe Order Helper class to handle mandate ID, Multibanco data, refund status, card brand, charge captured flag, status final flag, and the refund failure reason
 * Dev - Remove the merchant email address from the System Status Report
 * Dev - Replace the constant reference for the legacy SEPA payment method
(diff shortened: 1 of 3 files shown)


daledupreez (approved):
The change LGTM! 🚢

malithsen (approved):
Change makes sense. Thanks for the detailed explanation in the Linear issue.

PR #4841, opened 2025-12-02
Exclude order parameter from the request args when creating/updating a Stripe customer

diegocurbelo:
Fixes STRIPE-832

I was unable to to actually reproduce this problem

This PR refactors how the order object is passed when creating or updating Stripe customers. Previously, the order was included in the `$args` array (`$args['order']`), which caused it to be sent to the Stripe API. 

Now, the order is passed as a separate parameter, ensuring it's only used to extract billing details and never included in the API request payload.

--

An alternative solution would have been to remove the `order' entry from the `$args` parameter just before making the call to the `customers` endpoint, but that feels confusing as the `$order` arg is being added in the gateway class, not in this customer class:
```
unset( $args['order'] );
$response = WC_Stripe_API::request( $args, 'customers/' . $this->get_id() );
```
- https://github.com/woocommerce/woocommerce-gateway-stripe/blob/5e73f347de6c0d322041f61c76acc41b06523f43/includes/class-wc-stripe-customer.php#L480
- https://github.com/woocommerce/woocommerce-gateway-stripe/blob/5e73f347de6c0d322041f61c76acc41b06523f43/includes/class-wc-stripe-customer.php#L486
- https://github.com/woocommerce/woocommerce-gateway-stripe/blob/5e73f347de6c0d322041f61c76acc41b06523f43/includes/class-wc-stripe-customer.php#L531

## Testing instructions

**With a new user (wihtout a Stripe customer)**

1. Go to the Store and add a product to the Cart
2. Go to the checkout, fill in billing details, and select Stripe payment method
4. Complete the payment using a test card (e.g., `4242 4242 4242 4242`)
5. Verify that the order was completed successfully
6. Go to the Stripe Dashboard and verify that the customer was correctly created and that the billing information matches

|Stripe | Woo |
|-|-|
| Screenshot 2025-12-04 at 15 53 08 | Screenshot 2025-12-04 at 15 53 19|

**As a registered user: **
1. Go to the `Store > My Account > Subscriptions` and select the Subcription from the list:
2. Remove the saved `customer_id` from the `order_meta` for the Subscriptions order created in the steps above, ex:
    `DELETE FROM `wp_wc_orders_meta` WHERE `meta_key` = '_stripe_customer_id' AND `order_id` = '163';`
3. Click change payment and use any test card (eg., 5555 5555 5555 4444)
4. In `develop` the change fails with the error from the issue:
    Screenshot 2025-12-04 at 16 15 39
5. In this branch (`fix/832-exclude-order-parameter-when-creating-stripe-customers`), the change succeeds:
    Screenshot 2025-12-04 at 16 19 52

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -30,6 +30,7 @@
 * Tweak - Hide Amazon Pay from the standard payments in Optimized Checkout
 * Fix - Always use the current payment method configuration in Optimized Checkout
 * Fix - Fix revoked secret_key error during the OAuth account connection flow
+* Fix - Exclude order parameter from customer creation request arguments
 
 = 10.1.0 - 2025-11-11 =
 * Dev - Remove unused `shouldShowPaymentRequestButton` parameter and calculations from backend
--- a/changelog.txt
+++ b/changelog.txt
@@ -38,7 +38,7 @@
 * Update - Expand Amazon Pay support for all permitted currencies and countries
 * Add - Allow cache prefetch window to be adjusted via the wc_stripe_database_cache_prefetch_window filter
 * Fix - Respect button.radius value of 0 in Express Checkout Element appearance settings
-* Fix - Exclude order parameter from customer creation request arguments
+* Fix - Exclude order parameter from customer request data
 
 = 10.1.0 - 2025-11-11 =
 * Dev - Remove unused `shouldShowPaymentRequestButton` parameter and calculations from backend
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -522,7 +522,7 @@ class WC_Stripe_Customer {
 			throw new WC_Stripe_Exception( 'id_required_to_update_user', __( 'Attempting to update a Stripe customer without a customer ID.', 'woocommerce-gateway-stripe' ) );
 		}
 
-		$args = $this->generate_customer_request( $args );
+		$args = $this->generate_customer_request( $args, $order );
 
 		/**
 		 * Filters the arguments used to update a customer.
@@ -571,7 +571,7 @@ class WC_Stripe_Customer {
 
 			return $this->recreate_customer( $args, $current_context, $order );
 		} else {
-			return $this->update_customer( $args, $order );
+			return $this->update_customer( $args, false, $order );
 		}
 	}
 
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -175,7 +175,7 @@ class WC_Stripe_Customer {
 
 			// If the user email is not set, use the billing email.
 			if ( empty( $email ) ) {
-				$email = $this->get_billing_data_field( 'billing_email', $args );
+				$email = $this->get_billing_data_field( 'billing_email', $order );
 			}
 
 			// translators: %1$s First name, %2$s Second name, %3$s Username.
@@ -191,9 +191,9 @@ class WC_Stripe_Customer {
 				$defaults['name'] = $billing_full_name;
 			}
 		} else {
-			$billing_email      = $this->get_billing_data_field( 'billing_email', $args );
-			$billing_first_name = $this->get_billing_data_field( 'billing_first_name', $args );
-			$billing_last_name  = $this->get_billing_data_field( 'billing_last_name', $args );
+			$billing_email      = $this->get_billing_data_field( 'billing_email', $order );
+			$billing_first_name = $this->get_billing_data_field( 'billing_first_name', $order );
+			$billing_last_name  = $this->get_billing_data_field( 'billing_last_name', $order );
 
 			// translators: %1$s First name, %2$s Second name.
 			$description = sprintf( __( 'Name: %1$s %2$s, Guest', 'woocommerce-gateway-stripe' ), $billing_first_name, $billing_last_name );
@@ -226,7 +226,7 @@ class WC_Stripe_Customer {
 			if ( $user ) {
 				$defaults['address'][ $key ] = get_user_meta( $user->ID, $field, true );
 			} else {
-				$defaults['address'][ $key ] = $this->get_billing_data_field( $field, $args );
+				$defaults['address'][ $key ] = $this->get_billing_data_field( $field, $order );
 			}
 		}
 
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -156,6 +156,12 @@ class WC_Stripe_Customer {
 	 * @return array
 	 */
 	protected function generate_customer_request( $args = [], $order = null ) {
+		// The $order parameter was added in 10.2.0, so check for `$args['order'] for backwards compatibility.
+		if ( null === $order && isset( $args['order'] ) && $args['order'] instanceof WC_Order ) {
+			$order = $args['order'];
+		}
+		// Always ensure the old 'order' key is unset
+		unset( $args['order'] );
 		$user = $this->get_user();
 		if ( $user ) {
 			$billing_first_name = get_user_meta( $user->ID, 'billing_first_name', true );
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -150,10 +150,12 @@ class WC_Stripe_Customer {
 	/**
 	 * Generates the customer request, used for both creating and updating customers.
 	 *
-	 * @param  array $args Additional arguments (optional).
+	 * @param  array        $args  Additional arguments for the request (optional).
+	 * @param WC_Order|null $order The order object (optional). If provided, billing details will be retrieved from the order.
+	 *
 	 * @return array
 	 */
-	protected function generate_customer_request( $args = [] ) {
+	protected function generate_customer_request( $args = [], $order = null ) {
 		$user = $this->get_user();
 		if ( $user ) {
 			$billing_first_name = get_user_meta( $user->ID, 'billing_first_name', true );
@@ -341,12 +343,12 @@ class WC_Stripe_Customer {
 	/**
 	 * Get value of billing data field, either from POST or order object.
 	 *
-	 * @param string $field Field name.
-	 * @param array  $args  Additional arguments (optional).
+	 * @param string        $field Field name.
+	 * @param WC_Order|null $order The order object (optional).
 	 *
 	 * @return string
 	 */
-	private function get_billing_data_field( $field, $args = [] ) {
+	private function get_billing_data_field( $field, $order = null ) {
 		$valid_fields = [
 			'billing_email',
 			'billing_first_name',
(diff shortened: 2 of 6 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2992,8 +2992,7 @@ private function get_customer_id_for_order( WC_Order $order ): string {
 
 		// Pass the order object so we can retrieve billing details
 		// in payment flows where it is not present in the request.
-		$args = [ 'order' => $order ];
-		return $customer->update_or_create_customer( $args, $current_context );
+		return $customer->update_or_create_customer( [], $current_context, $order );
```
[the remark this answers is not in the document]
diegocurbelo: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4841#discussion_r2582724244)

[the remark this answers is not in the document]
diegocurbelo: Added in [314951b](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4841/commits/314951b64616ac1f54c48e15908eff980b537cdd)

diegocurbelo:
> @diegocurbelo I've opened a new pull request, #4844, to work on those changes. Once the pull request is ready, I'll request review from you.

This is a nice feature, but generating them locally with Cursor/Claude is more convenient, easier to interact with when asking for changes/refactoring (to use dataProviders, for example)

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
```diff
@@ -3340,4 +3340,244 @@ public function test_add_bnpl_debug_metadata() {
...
+			$this->assertArrayHasKey( 'address', $request_body, 'Request should include address data.' );
+			$this->assertEquals( $expected_customer_data['address']['line1'], $request_body['address']['line1'] ?? '', 'Billing address line1 should match order.' );
+			if ( ! empty( $expected_customer_data['address']['line2'] ) ) {
+				$this->assertEquals( $expected_customer_data['address']['line2'], $request_body['address']['line2'] ?? '', 'Billing address line2 should match order.' );
+			}
+			$this->assertEquals( $expected_customer_data['address']['city'], $request_body['address']['city'] ?? '', 'Billing city should match order.' );
+			$this->assertEquals( $expected_customer_data['address']['state'], $request_body['address']['state'] ?? '', 'Billing state should match order.' );
+			$this->assertEquals( $expected_customer_data['address']['postal_code'], $request_body['address']['postal_code'] ?? '', 'Billing postal code should match order.' );
+			$this->assertEquals( $expected_customer_data['address']['country'], $request_body['address']['country'] ?? '', 'Billing country should match order.' );
+		}
```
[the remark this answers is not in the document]
diegocurbelo: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4841/commits/a53b06fc78ec32130df6e4a712d37ba5ad3bc1a3

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
```diff
@@ -3340,4 +3340,244 @@ public function test_add_bnpl_debug_metadata() {
...
+		// Use reflection to access the private method.
+		$reflection = new \ReflectionClass( $gateway );
+		$method     = $reflection->getMethod( 'get_customer_id_for_order' );
+		$method->setAccessible( true );
+
+		// Mock get_stripe_customer_id to return empty (no existing customer).
+		$gateway = $this->getMockBuilder( WC_Stripe_UPE_Payment_Gateway::class )
+			->onlyMethods( [ 'get_stripe_customer_id', 'get_user_from_order', 'is_valid_pay_for_order_endpoint' ] )
+			->getMock();
+
```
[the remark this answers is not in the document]
diegocurbelo: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4841/commits/a53b06fc78ec32130df6e4a712d37ba5ad3bc1a3

includes/class-wc-stripe-customer.php
```diff
@@ -366,31 +368,30 @@ private function get_billing_data_field( $field, $args = [] ) {
...
-			}
+			$filter = 'billing_email' === $field ? FILTER_SANITIZE_EMAIL : FILTER_SANITIZE_SPECIAL_CHARS;
+			return filter_var( wp_unslash( $_POST[ $field ] ), $filter ); // phpcs:ignore WordPress.Security.NonceVerification
+		}
 
-			return filter_var( wp_unslash( $_POST[ $field ] ), FILTER_SANITIZE_SPECIAL_CHARS ); // phpcs:ignore WordPress.Security.NonceVerification
-		} elseif ( isset( $args['order'] ) && $args['order'] instanceof WC_Order ) {
+		if ( $order instanceof WC_Order ) {
 			switch ( $field ) {
 				case 'billing_email':
```
wjrosa, line 383: I think there are multiple ways we could improve this `switch`, like just checking if the method exists within `$order` and calling it dynamically, like `$order->{'get_' . $field}();`. But that's not in the context of this PR

includes/class-wc-stripe-customer.php
```diff
@@ -150,10 +150,12 @@ public function set_customer_data( $data ) {
...
 	 * Generates the customer request, used for both creating and updating customers.
 	 *
-	 * @param  array $args Additional arguments (optional).
+	 * @param  array        $args  Additional arguments for the request (optional).
+	 * @param WC_Order|null $order The order object (optional). If provided, billing details will be retrieved from the order.
+	 *
 	 * @return array
 	 */
-	protected function generate_customer_request( $args = [] ) {
+	protected function generate_customer_request( $args = [], $order = null ) {
```
wjrosa, line 158: Since this is a new parameter, I think we should consider adding the type (here and everywhere):
```
?WC_Order $order = null
```

wjrosa (approved):
Fix makes sense to me! Left some optional suggestions

includes/class-wc-stripe-customer.php
daledupreez: While I would love to be able to do that, not having strict types higher up makes that a risky proposition. So I don't think we should tackle that in this PR.

includes/class-wc-stripe-customer.php
```diff
@@ -366,31 +368,30 @@ private function get_billing_data_field( $field, $args = [] ) {
...
-			if ( 'billing_email' === $field ) {
-				return filter_var( wp_unslash( $_POST[ $field ] ), FILTER_SANITIZE_EMAIL ); // phpcs:ignore WordPress.Security.NonceVerification
-			}
+			$filter = 'billing_email' === $field ? FILTER_SANITIZE_EMAIL : FILTER_SANITIZE_SPECIAL_CHARS;
+			return filter_var( wp_unslash( $_POST[ $field ] ), $filter ); // phpcs:ignore WordPress.Security.NonceVerification
+		}
 
-			return filter_var( wp_unslash( $_POST[ $field ] ), FILTER_SANITIZE_SPECIAL_CHARS ); // phpcs:ignore WordPress.Security.NonceVerification
-		} elseif ( isset( $args['order'] ) && $args['order'] instanceof WC_Order ) {
+		if ( $order instanceof WC_Order ) {
```
daledupreez, line 381: Good catch on making this a simple `if` after the return above.

includes/class-wc-stripe-customer.php
daledupreez, line 158: Given that all the calling functions are public, I think we should extract `$order` from `$args` just to be extremely defensive against existing callers of the functions.
```suggestion
	protected function generate_customer_request( $args = [], $order = null ) {
		// The $order parameter was added in 10.2.0, so check for `$args['order'] for backwards compatibility.
		if ( null === $order && isset( $args['order'] ) && $args['order'] instanceof WC_Order ) {
			$order = $args['order'];
		}
		// Always ensure the old 'order' key is unset
		unset( $args['order'] );
```

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
```diff
@@ -3343,4 +3343,246 @@ public function test_add_bnpl_debug_metadata() {
...
+	 * @param bool   $is_guest Whether the order is for a guest user.
+	 * @param string $existing_stripe_customer_id Existing Stripe customer ID for the user (empty for new customer).
+	 * @param string $expected_customer_id Expected Stripe customer ID to be returned.
+	 * @param string $api_url_pattern Pattern to match the API URL.
+	 * @param array  $billing_data Billing data to set on the order (and user meta for logged-in users).
+	 * @param array  $expected_customer_data Expected customer data in the API request.
+	 *
+	 * @return void
+	 */
+	public function test_get_customer_id_for_order_retrieves_billing_details_from_order( $scenario_name, $is_guest, $existing_stripe_customer_id, $expected_customer_id, $api_url_pattern, $billing_data, $expected_customer_data ) {
```
daledupreez: Nit: why not specify explicit types?
```suggestion
	public function test_get_customer_id_for_order_retrieves_billing_details_from_order( string $scenario_name, bool $is_guest, string $existing_stripe_customer_id, string $expected_customer_id, string $api_url_pattern, array $billing_data, array $expected_customer_data ) {
```

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
```diff
@@ -3343,4 +3343,246 @@ public function test_add_bnpl_debug_metadata() {
...
+			// Verify billing details from the order are used in the customer creation/update request.
+			$this->assertEquals( $expected_customer_data['email'], $request_body['email'] ?? '', 'Billing email should match order billing email.' );
+			$this->assertEquals( $expected_customer_data['name'], $request_body['name'] ?? '', 'Billing name should match order billing name.' );
+
+			// Verify address details are present and match the order.
+			$this->assertArrayHasKey( 'address', $request_body, 'Request should include address data.' );
+			$this->assertEquals( $expected_customer_data['address']['line1'], $request_body['address']['line1'] ?? '', 'Billing address line1 should match order.' );
+			if ( ! empty( $expected_customer_data['address']['line2'] ) ) {
+				$this->assertEquals( $expected_customer_data['address']['line2'], $request_body['address']['line2'] ?? '', 'Billing address line2 should match order.' );
+			}
```
daledupreez, line 3508: Nit: If the value is empty (i.e. `''`), should we not verify that the sent data is either empty or not present?

changelog.txt
```diff
@@ -38,6 +38,7 @@
 * Update - Expand Amazon Pay support for all permitted currencies and countries
 * Add - Allow cache prefetch window to be adjusted via the wc_stripe_database_cache_prefetch_window filter
 * Fix - Respect button.radius value of 0 in Express Checkout Element appearance settings
+* Fix - Exclude order parameter from customer creation request arguments
```
daledupreez: Nit: I would argue that we're excluding the order parameter from the data. The impact is to ensure we're not sending the `order` key, so I think it's worth noting that explicitly. Also, this could affect updates, so it may be worth reflecting that in the description too.
```suggestion
* Fix - Exclude order parameter from customer request data
```

includes/class-wc-stripe-customer.php
daledupreez: If we add this code, we should also add an explicit test case for it to make sure that direct calls to the relevant functions don't send the order data to Stripe _and_ use the correct data from the order.

daledupreez (approved):
The changes are working as described for me -- I didn't trigger the issue in `develop`, as the Stripe customer ID was saved somewhere outside the subscription order because I didn't see any other API calls to Stripe.

My only concern is that it may be well worth adding backwards compatibility to check for data in `order` and removing it when present.

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
diegocurbelo: Added the check in [c5179a5](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4841/commits/c5179a539050050aafef65cfa7d55756719de48d)

PR #4840, opened 2025-12-02
Incompatible tax pill + Requires card method pill

wjrosa:
Fixes STRIPE-806 / #4782
Fixes STRIPE-685 / #4618

## Changes proposed in this Pull Request:

In this PR I am introducing two new payment method pills to be shown above methods in the list screen:
- Incompatible tax: shown above the Amazon Pay method when the store taxes are based on the customer's billing address
- Requires card: shown above Apple Pay / Google Pay and Link when the card payment method is disabled

Preview for the incompatible tax pill:
Screenshot 2025-12-02 at 14 34 35

Preview for the "requires card method" pill:
Screenshot 2025-12-02 at 17 11 42

I am also filtering out Amazon Pay when saving settings, and the store has an incompatible tax setup.

## Testing instructions

- Checkout and build this branch on your test environment (`update/remove-amazon-pay-from-list-when-taxes-are-based-on-billing`)

## Incompatible tax pill

- Enable the Amazon Pay feature flag (you can use the `wc_stripe_is_amazon_pay_available` filter)
- Enable taxes for the store
- In the Tax tab, set taxes based on the customer billing address
- Go to the Stripe payment methods list in wp-admin
- Make sure Amazon Pay is disabled
- Confirm the new pill is shown above Amazon Pay

## Requires card method pill

- Disable the credit card payment method
- Go to the Stripe payment methods list in wp-admin
- Make sure Link and Apple Pay / Google Pay are disabled
- Confirm the pill is shown above both

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4845, opened 2025-12-04
Move pills display logic outside the component

wjrosa:
See https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4840#discussion_r2586713439

Following @diegocurbelo 's suggestion in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4840#discussion_r2586713439, I am moving the display logic for the existing payment method pill components outside of them.

## Testing instructions

Code review. Check if the tests are still passing. You can also perform some basic smoke-testing in the payment methods listing screen.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Update - Block Amazon Pay from being enabled when taxes are based on the billing address. Adds new pills to indicate why it is disabled, and another pill to indicate when a express method is disabled due card being disabled
 * Dev - Deprecates all the legacy checkout payment method classes
 * Dev - Deprecates all the LPM class constants
 * Dev - Remove all references to the UPE-enabled feature flag
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
-* Dev - Move the payment method pill components display logic outside of them
+* Dev - Refactor display logic for payment method issue pills
 * Update - Improves the error log for SSL connection missing when trying to render the express checkout buttons
 * Update - Block Amazon Pay from being enabled when taxes are based on the billing address. Adds new pills to indicate why it is disabled, and another pill to indicate when a express method is disabled due card being disabled
 * Update - Enable the Optimized Checkout Suite feature for all new installations
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 10.2.0 - xxxx-xx-xx =
 * Dev - Move the payment method pill components display logic outside of them
 * Update - Block Amazon Pay from being enabled when taxes are based on the billing address. Adds new pills to indicate why it is disabled, and another pill to indicate when a express method is disabled due card being disabled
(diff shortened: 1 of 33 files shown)


client/settings/general-settings-section/payment-method-description.js
```diff
@@ -50,6 +52,16 @@ const PaymentMethodDescription = ( {
...
+	const unavailableReason = usePaymentMethodUnavailableReason( id );
+	const showMissingCurrencyPill =
+		! deprecated &&
+		PAYMENT_METHOD_UNAVAILABLE_REASONS.UNSUPPORTED_CURRENCY ===
+			unavailableReason;
+	const showCapabilityStatusPill = ! deprecated;
+	const showUnavailableDueConflictPill =
+		! deprecated &&
+		PAYMENT_METHOD_UNAVAILABLE_REASONS.OFFICIAL_PLUGIN_CONFLICT ===
+			unavailableReason;
```
daledupreez, line 64: Not blocking, but I am wondering whether this might be easier to handle via a helper function that allows for early returns, given how interrelated all the conditions are. I was thinking something like the following:
```js
	const getIssuePillContent = ( { deprecated, id, label, unavailableReason } ) => {
		if ( deprecated ) {
			return <PaymentMethodDeprecationPill />;
		}

		const pillsToShow = [];

		if ( PAYMENT_METHOD_UNAVAILABLE_REASONS.UNSUPPORTED_CURRENCY === unavailableReason ) {
			pillsToShow.push( <PaymentMethodMissingCurrencyPill id={ id } label={ label } /> );
		}

		pillsToShow.push( <PaymentMethodCapabilityStatusPill id={ id } label={ label } /> );

		if ( PAYMENT_METHOD_UNAVAILABLE_REASONS.OFFICIAL_PLUGIN_CONFLICT === unavailableReason ) {
			pillsToShow.push( <PaymentMethodUnavailableDueConflictPill id={ id } label={ label } /> );
		}

		return (
			<>
				{ ...pillsToShow }
			</>
		);
	}
```

After looking at the code, it might also be simpler if we added the capability pill followed by the other item, as we might be able to use inline conditions rather than an array.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Dev - Move the payment method pill components display logic outside of them
```
daledupreez: This description is not super-clear to me. Might it be worth shifting a bit higher level in terms of the details? "Restructure" could also work.
```suggestion
* Dev - Refactor display logic for payment method issue pills
```

daledupreez (approved):
Changes look good to me and test well. I have one non-blocking suggestion that might be worth considering.

client/settings/general-settings-section/payment-method-description.js
wjrosa: Maybe that's something we could tackle in another PR. Something else I am thinking is: this logic shows a single pill for each method, while a method could be unavailable for multiple reasons. I think we can also refactor the `getPaymentMethodUnavailableReason` function to return a list of reasons

diegocurbelo (approved):
Thanks for refactoring this @wjrosa!

The changes look good and test well.

PR #3184, opened 2024-06-06
Change display order upe methods

Mayisha:
Fixes #2994

This PR implements the ability to order the payment methods on the Stripe settings page when UPE is enabled.

## Changes proposed in this Pull Request:
- Saving the UPE payment method order in an option `stripe_upe_payment_method_order`. If the option is empty but we have the order of legacy payment methods saved, we copy that order to this option to preserve the previous setting.
- Moved the capability check  to the backend.
- To render the payment methods in the saved order on the checkout page, we are reordering them in `reorder_available_payment_gateways` function similar to [the legacy one](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/4f2b0222221d1af815e991ee4f9d3518110d1c13/includes/class-wc-gateway-stripe.php#L658) 

## Testing instructions
- In `develop` branch enable the new checkout experience.
- On the Stripe settings page, note that there is no `Change display order` button at the top of the payment method list.
- Now enable the legacy checkout experience. You should see a `Change display order` button at the top of the payment method list.
- Change display order of a few methods and save changes.
- Now checkout to this (`task/2994-change-display-order-upe`) branch.
- Enable the new checkout experience and refresh the settings page.
- Confirm that the payment method order previously saved in legacy mode still persists and the additional UPE specific payment methods (BNPL, Wallet etc. which are not available in legacy mode) are placed at the end of the list.
- You should see a `Change display order` button at the top of the payment method list as well.
- Now change the display order of a few methods and save changes. If you refresh the page, the new order should persist.
- Enable a few methods randomly.
- As a shopper, add a product to the cart and go to the shortcode checkout page. Confirm that the available payment methods on the checkout page are following the order saved on the settings page.
- This PR does not cover the block checkout, so the order on the block checkout page will not follow the settings. It will be handled in a separate PR.

Issue #2735, opened 2023-10-30
[Settings] Allow changing the display order of payment methods


a-danae:
This issue is blocked by https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2734

### The design

- When clicking on the "Change display order" button, a merchant can drag and drop the payment methods to reorder them: xVAk98kbRKUByVTrhPc98g-fi-616_33462
- There's a dismissible banner on top of the Payment methods section xVAk98kbRKUByVTrhPc98g-fi-603_28309

### Some considerations

- When UPE is enabled, we'll disable reordering. This will be enabled again when split UPE is implemented.
- For non-UPE, the order defined here must be reflected on the checkout page.
- When a merchant upgrades from a previous version and non-UPE is enabled, the order of the payment methods from the WooCommerce -> Settings -> Payments page must persist.
- If possible, the display order of the non-UPE payment methods would be backward-compatible. The order would persist if the merchant rolls back to a previous version.

Mayisha:
This can be closed with PR #2756 and #2808 merged

PR #4846, opened 2025-12-04
Removing the UPE enabled flag references from the frontend + deprecated banners

wjrosa:

In this cleaning-up PR, I am removing all references to the UPE-enabled feature flag, related code, and the promotional banners that were shown only for the legacy checkout.

I am also updating the rendering logic for the order customization notice according to the discussion in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4846#discussion_r2597173375

## Testing instructions

Code review. Check if the tests are still passing. You can also perform some basic smoke-testing on the frontend.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.2.0 - xxxx-xx-xx =
+* Dev - Removes all references to the UPE-enabled feature flag, as well as deprecated promotional banners
```
Mayisha: How about adding two separate entries for the two tasks (remove the ff reference and deprecate banner)?

client/settings/display-order-customization-notice/index.js
```diff
@@ -1,60 +0,0 @@
...
-	if ( isUpeEnabled || ! showNotice ) {
-		return null;
-	}
-
-	return (
-		<NoticeWrapper isDismissible={ true } onRemove={ handleDismissNotice }>
-			<NoticeContent>
-				<Icon icon={ info } size={ 24 } />
-				{ __(
-					"Customize the display order of Stripe payment methods for customers at checkout. This customization occurs within the plugin and won't affect the order in relation to other installed payment providers.",
```
Mayisha, line 52: We have the `Change display order` option in UPE also.  Shouldn't we keep this notice? 👀

client/settings/general-settings-section/section-heading.js
```diff
@@ -90,25 +88,25 @@ const SectionHeading = ( { isChangingDisplayOrder, onChangeDisplayOrder } ) => {
...
-										title: __(
-											'Refresh payment methods',
-											'woocommerce-gateway-stripe'
-										),
-										onClick: refreshAccount,
-									},
-								] }
-							/>
-						) }
+						(
```
Mayisha:
```suggestion
```

Need to remove this bracket. This is adding a bracket around the 3 dot menu.

client/settings/general-settings-section/section-heading.js
```diff
@@ -90,25 +88,25 @@ const SectionHeading = ( { isChangingDisplayOrder, onChangeDisplayOrder } ) => {
...
+								{
+									title: __(
+										'Refresh payment methods',
+										'woocommerce-gateway-stripe'
+									),
+									onClick: refreshAccount,
+								},
+							] }
+						/>
+						)
```
Mayisha:
```suggestion
```

Mayisha (commented):
Left a couple of comments from code review. 

✅ Settings page looks good.
✅ Storefront looks good.
✅ Tested payment with card, CashApp and GPay.
⚠️ I had a GB account connected with store currency GBP and AliPay was offered in the checkout page with the following error. Probably unrelated to this PR but documenting here as discovered during smoke testing.

Screenshot 2025-12-08 at 12 49 56 PM

⚠️ A few JS tests are failing.

client/settings/display-order-customization-notice/index.js
wjrosa: I removed this notice because, in the logic, it is not rendered when UPE is enabled. See: https://github.com/woocommerce/woocommerce-gateway-stripe/blob/eaec13976b2719b16bd267afe82577f3bf9dd5d2/client/settings/display-order-customization-notice/index.js#L43

changelog.txt
wjrosa: Good idea. Updated the entries in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4846/commits/efa864acac1b8bb441cb3692348c53be2896bde6

wjrosa:
Thanks for the review, Mayisha! I addressed all the issues, and fixed the unit tests

client/settings/display-order-customization-notice/index.js
Mayisha: I finally found the history. We implemented the re-ordering and this banner for [legacy at first](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2735). It was because split PE work was in-progress then and we were supposed to implement re-ordering in UPE after split PE is done. We implemented it later for [UPE](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3184) but I missed to remove this check from the banner 😅

wjrosa: Oh right. Thanks for the context. So, in this case, should I just update the condition above? Removing the UPE-enabled check

Mayisha: That would be nice. We were wrongly hiding the notice, actually 😅 
But the notice should be hidden for OCS because reordering is not supported with OCS.

wjrosa: Makes sense. Reverted the removal and updated the logic in my latest commits 👍

Mayisha (approved):
Looks good to me 👍

client/settings/payments-and-transactions-section/__tests__/manual-capture-control.test.js
```diff
@@ -65,15 +38,11 @@ describe( 'ManualCaptureControl', () => {
 		expect( manualCaptureToggleMock ).not.toHaveBeenCalled();
 	} );
 
-	it( 'should toggle the flag when UPE is enabled', async () => {
+	it( 'should toggle the flag when', async () => {
```
malithsen: This looks incomplete, let's update the test name

malithsen (approved):
Smoke tests look good.

client/settings/payments-and-transactions-section/__tests__/manual-capture-control.test.js
wjrosa: Oops! Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4846/commits/15528058f8448f23d52919eb3563f0d1d8096a4a. And improved it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4846/commits/9df2aa3bd57eb42d8bd3b19a619d9a247b255dbc

PR #4848, opened 2025-12-05
Preload customer billing info in Stripe element fields

diegocurbelo:
Fixes STRIPE-838
Fixes #<github_issue_id>

During checkout, billing fields in the Stripe payment elements are hidden, and the values from the WooCommerce form are used, causing a discrepancy with the Change Payment Method and Add Payment Method pages, which require those fields.

This PR adds prefilling of the billing field in the Stripe Payment Element on the Change Payment Method and Add Payment Method pages using the customer's billing information.

| Checkout | add / Change Payment Method |
|-|-|
|Screenshot 2025-12-05 at 14 39 41|Screenshot 2025-12-05 at 14 39 29|
|Screenshot 2025-12-05 at 14 38 19|Screenshot 2025-12-05 at 14 39 14|

## Testing instructions

0. Set the store currency to `EUR` and enable `SEPA`.

### Checkout Page (Regression Test)
1. Add a subscription product to the Cart
2. Go to the Checkout page
3. Fill in billing details
4. Select Stripe as the payment method
5. Select SEPA Direct Debit
6. Verify only the IBAN field is present
    Screenshot 2025-12-05 at 14 38 19

### Change Payment Method Page (Subscription)
1. Navigate to: `My Account → Subscriptions → [subscription created in step above] → Change Payment Method`
2. Select Stripe as the payment method
3. Select SEPA Direct Debit payment method
4. Verify that the fields are correctly preloaded:
    Screenshot 2025-12-05 at 14 39 14

### Test 3: Add Payment Method Page
1. Navigate to: `My Account → Payment Methods → Add Payment Method`
2. Select Stripe as the payment gateway
3. Select SEPA Direct Debit payment method
4. Verify that the fields are correctly preloaded

### Pay for Order Page
1. Navigate to `WP-Admin > WooCommerce > Orders` and create a pending order (or use an existing pending order)
2. Use the `Customer payment page →` link to go to the Pay for order page
3. Select Stripe as the payment method
4. Select SEPA Direct Debit payment method
5. Verify that the fields are correctly preloaded

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/stripe-utils/__tests__/utils.test.js
+++ b/client/stripe-utils/__tests__/utils.test.js
@@ -1,4 +1,4 @@
-import { getFontSizeBase } from '../utils';
+import { getFontSizeBase, getDefaultValues } from '../utils';
 
 describe( 'utils', () => {
 	describe( 'getFontSizeBase', () => {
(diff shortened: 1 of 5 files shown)


client/stripe-utils/utils.js
```diff
@@ -455,10 +455,64 @@ export const getUpeSettings = () => {
...
+	if ( isOrderPay || isChangingPayment || isAddPaymentMethod ) {
+		const billingData = stripeServerData?.customerBillingData;
+
+		if ( billingData && billingData.email ) {
+			// Build address object, only including non-empty values
+			const address = {};
+			const country = billingData.address?.country?.trim();
+			if ( country ) {
+				// Country must be uppercase ISO 3166-1 alpha-2 code for Stripe
+				address.country = country.toUpperCase();
```
wjrosa, line 479: Maybe we should also validate the `country` param has the format we expect

wjrosa (approved):
Nice work! Looks good to me. Left an optional suggestion

client/stripe-utils/__tests__/utils.test.js
```diff
@@ -39,4 +39,208 @@ describe( 'utils', () => {
...
+					customerBillingData: {
+						name: '  John Doe  ',
+						email: '  john@example.com  ',
+						phone: '  +1234567890  ',
+						address: {
+							country: '  us  ',
+							line1: '  123 Main St  ',
+							line2: '', // empty, should be filtered out
+							city: '  New York  ',
+							state: '', // empty, should be filtered out
```
daledupreez: Nit: it feels like it would be worth making one of these fields whitespace only to check that we handle that correctly.
```suggestion
							state: '    ', // only whitespace, should be filtered out
```

client/stripe-utils/utils.js
```diff
@@ -455,10 +455,64 @@ export const getUpeSettings = () => {
...
+	const stripeServerData = getStripeServerData();
+	const isOrderPay = stripeServerData?.isOrderPay;
+	const isChangingPayment = stripeServerData?.isChangingPayment;
+	const isAddPaymentMethod = stripeServerData?.isAddPaymentMethod;
+
+	// On order pay, change payment method, and add payment method pages, use billing data from customer.
+	if ( isOrderPay || isChangingPayment || isAddPaymentMethod ) {
+		const billingData = stripeServerData?.customerBillingData;
+
+		if ( billingData && billingData.email ) {
```
daledupreez: Nit: Should we not validate that the email after running `trim()`? I believe `'  '` will evaluate as truthy.
```suggestion
		if ( billingData && billingData.email?.trim() ) {
```

client/stripe-utils/utils.js
```diff
@@ -455,10 +455,64 @@ export const getUpeSettings = () => {
...
+				address.postal_code = postalCode;
+			}
+
+			return {
+				defaultValues: {
+					billingDetails: {
+						name: billingData.name?.trim() || undefined,
+						email: billingData.email.trim(),
+						phone: billingData.phone?.trim() || undefined,
+						...( Object.keys( address ).length > 0 && { address } ),
```
daledupreez: Nit: I prefer the structure here to be slightly different, as I honestly don't know offhand how `...( false )` should work. It may be worth pulling this out before the return block so it's clearer.
```suggestion
						...( Object.keys( address ).length > 0 ? { address } : {} ),
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -548,6 +548,27 @@ public function javascript_params() {
...
+						'line2'       => $customer->get_billing_address_2(),
+						'city'        => $customer->get_billing_city(),
+						'state'       => $customer->get_billing_state(),
+						'postal_code' => $customer->get_billing_postcode(),
+					],
+				];
+			}
+		}
+
 		if ( parent::is_valid_pay_for_order_endpoint() || $is_change_payment_method ) {
```
daledupreez: Nit: it feels like it could be worth extracting the `is_valid_pay_for_order_endpoint()` result into a local variable.
```suggestion
		$is_pay_for_order = parent::is_valid_pay_for_order_endpoint();

		// Pass billing details from user's customer data for preloading Payment Element fields in Pay for Order, Change Payment Method, and Add Payment Method pages.
		if ( is_wc_endpoint_url( 'add-payment-method' ) || $is_pay_for_order || $is_change_payment_method ) {
			// Get billing details from the current user's customer data instead of the order.
			$customer = WC()->customer;
			if ( $customer ) {
				$stripe_params['customerBillingData'] = [
					'name'    => trim( $customer->get_billing_first_name() . ' ' . $customer->get_billing_last_name() ),
					'email'   => $customer->get_billing_email(),
					'phone'   => $customer->get_billing_phone(),
					'address' => [
						'country'     => $customer->get_billing_country(),
						'line1'       => $customer->get_billing_address_1(),
						'line2'       => $customer->get_billing_address_2(),
						'city'        => $customer->get_billing_city(),
						'state'       => $customer->get_billing_state(),
						'postal_code' => $customer->get_billing_postcode(),
					],
				];
			}
		}

if ( $is_pay_for_order || $is_change_payment_method ) {
```

daledupreez (approved):
This is working as described and should help most shoppers when they get to these screens. I also smoke tested the default data for a few other payment methods, and they generally look good.

client/stripe-utils/utils.js
diegocurbelo: Makes sense, `...false` adds nothing, but the suggestion makes that clearer 👍🏼

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
diegocurbelo: Changed in [59efe28](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4848/commits/59efe28fc79836bdb4f52d57fb7d990933871d8a)

client/stripe-utils/utils.js
diegocurbelo: The billingData comes directly from the server (`stripeServerData?.customerBillingData`), I think we need to either validate it server side (although it might be redundant as it's the WC user billing info), or add the values from `WC()->countries->get_countries()` to `stripeServerData`  so we can properly check the code is valid.
In either case, let's do that in a follow-up PR.

PR #4857, opened 2025-12-09
Enable Amazon Pay for eligible new installs

daledupreez:
Fixes STRIPE-817

Blocked by https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4858

This PR ensures that new installs will set the `wc_stripe_amazon_pay_default_on` option, and our logic to initialise PMCs will now check that flag to see if Amazon Pay should be enabled by default alongside Apple Pay and Google Pay. Those checks take into account the store currency and the Stripe account country so we only enable Amazon Pay where it is available through Stripe.

Note that this PR doesn't work without the changes from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4858, as I uncovered those subtle issues while testing this PR. We will need to merge that PR into `develop` and then merge `develop` into `feature/amazon-pay-beta` for this to be fully testable.

## Testing instructions

As noted above, you'll need the changes from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4858 and the changes from this branch.

* Build the code out of this branch along with the necessary fixes form https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4858 patched or cherry-picked onto this code.
* Create a brand new site with WooCommerce installed - we don't want the Stripe plugin ever to have been installed before
* Install and activate the newly built plugin on the site
* Verify that the `wc_stripe_amazon_pay_default_on` option is now set to `yes`: `wp option get wc_stripe_amazon_pay_default_on`
* Ensure your store is configured to use a currency supported by Amazon Pay
* Connect to Stripe via OAuth using an account that supports Amazon Pay in your store currency
* Confirm that Amazon Pay is enabled
* Now change your store currency and disconnect from Stripe
* Connect to Stripe using an account country and currency combination not supported by Amazon Pay -- the easiest is EUR for a US account, as Amazon Pay only supports USD for US accounts.
* Confirm that Amazon Pay is not enabled.

* Use the built plugin to upgrade a site that already had the Stripe plugin installed and activated - ensure the plugin is also activated after the installation
* Confirm that the `wc_stripe_amazon_pay_default_on` option is _not_ set: `wp option get wc_stripe_amazon_pay_default_on`
* (Re)connect to Stripe, and confirm that Amazon Pay is not defaulted on

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -343,58 +343,60 @@ class WC_Stripe {
 
 		$previous_version = get_option( 'wc_stripe_version' );
 
-		if ( WC_STRIPE_VERSION !== $previous_version ) {
-			do_action( 'woocommerce_stripe_updated' );
+		if ( WC_STRIPE_VERSION === $previous_version ) {
+			return;
+		}
 
-			if ( ! defined( 'WC_STRIPE_INSTALLING' ) ) {
-				define( 'WC_STRIPE_INSTALLING', true );
-			}
+		do_action( 'woocommerce_stripe_updated' );
 
-			$is_new_install = false === $previous_version;
+		if ( ! defined( 'WC_STRIPE_INSTALLING' ) ) {
+			define( 'WC_STRIPE_INSTALLING', true );
+		}
 
-			// Mark optimized checkout as default on for new installs.
-			if ( $is_new_install && false === get_option( 'wc_stripe_optimized_checkout_default_on' ) ) {
-				update_option( 'wc_stripe_optimized_checkout_default_on', true );
-			}
+		$is_new_install = false === $previous_version;
 
-			if ( $is_new_install ) {
-				update_option( 'wc_stripe_amazon_pay_default_on', 'yes' );
-			}
+		// Mark optimized checkout as default on for new installs.
+		if ( $is_new_install && false === get_option( 'wc_stripe_optimized_checkout_default_on' ) ) {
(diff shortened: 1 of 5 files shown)


includes/class-wc-stripe.php
```diff
@@ -337,18 +337,30 @@ public function install() {
...
 		}
 
-		if ( ! defined( 'IFRAME_REQUEST' ) && ( WC_STRIPE_VERSION !== get_option( 'wc_stripe_version' ) ) ) {
+		if ( defined( 'IFRAME_REQUEST' ) ) {
+			return;
+		}
+
+		$previous_version = get_option( 'wc_stripe_version' );
+
+		if ( WC_STRIPE_VERSION !== $previous_version ) {
```
diegocurbelo: [nit]: It feels like this could be an early return too:

```php
if ( WC_STRIPE_VERSION === $previous_version ) {
    return;
}
```

diegocurbelo (approved):
Looks good to me!, I left an optional suggestion.

includes/class-wc-stripe.php
wjrosa: I agree. It is more elegant. Addressed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4857/commits/d1c723d029a841380f88eb1caedc046c336b1358

malithsen (approved):
This looks good to me

PR #4858, opened 2025-12-09
Fix initialization logic after account key updates

daledupreez:
Fixes STRIPE-846

Related to:
 * STRIPE-817
 * https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4857

This PR fixes some subtle bugs in our logic for handling key updates, including the following:
 * When testing on a local/development site, reconfiguring webhooks can trigger an error due to the site being unreachable, which was preventing some actions from running as they should have
 * When testing reconnections, it was possible that the old PMC was getting loaded from the database cache
 * ~When updating the keys, we should remove the previous PMC fallback ID, as it may no longer exist~ - we should not remove the fallback PMC in all cases, as we should retain the fallback PMC ID across reconnections

## Testing instructions

This is a tricky one to test, as the specific failure modes are not all immediately apparent, especially the PMC cache clearing after key updates.

The main change around the order of operations when saving keys can be seen by doing the following:
* Set up a local site to run using the code from `develop`
* Set up a Stripe connection, and enable logging
* Add a debug log statement to the end of `WC_Stripe_Connect->save_stripe_keys()` just before the final `return` statement, e.g. `WC_Stripe_Logger::debug( 'Final steps of save_stripe_keys()' );`
* Ensure the current log is clear (this is optional, but will make it easier to check later)
* Reconnect a Stripe account
* Verify that you do not see your log entry, but you do see an error with the text `OAuth: Account connection failed`, which then contains a `wc_stripe_webhook_error` when expanded.
* Repeat the tests in this branch, and verify that the reconnection proceeds as expected
* The PMC 
* You can also repeat the tests in this branch after running `npm run wp option patch update woocommerce_stripe_settings pmc_enabled "''"`

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/connect/class-wc-stripe-connect.php
```diff
@@ -290,6 +301,8 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 			update_option( 'wc_stripe_' . $prefix . 'oauth_failed_attempts', 0 );
 			update_option( 'wc_stripe_' . $prefix . 'oauth_last_failed_at', '' );
 
+			$this->clear_caches_after_key_update( true );
```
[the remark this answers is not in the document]
daledupreez: Fixed in 5b922147c2d6d4a0710e69658d11f9fedddb1dab.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -311,6 +324,16 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 				$this->unschedule_connection_refresh();
 			}
 
+			// For new installs the legacy gateway gets instantiated because there is no settings in the DB yet,
+			// so we need to instantiate the UPE gateway just for the PMC migration.
+			WC_Stripe::get_instance()->get_main_stripe_gateway();
+
+			// If pmc_enabled is not set (aka new install) or is not 'yes' (aka migration already done) we need to migrate the payment methods from the DB option to Stripe PMC API.
```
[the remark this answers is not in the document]
daledupreez: This was a pre-existing comment, so I am not going to modify it here. I think it's actually accurate, though it is slightly confusing.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -96,6 +96,15 @@ private static function get_payment_method_configuration_from_cache() {
 		return self::$primary_configuration;
 	}
 
+	/**
+	 * Reset the payment method configuration fallback ID.
+	 */
+	public static function reset_payment_method_configuration_fallback_id(): void {
```
malithsen: nit: Should we remove both test and live keys here?

malithsen (approved):
Looks good to me.

Mayisha (approved):
Connect, disconnect, reconnect all flows are working well 👍

includes/class-wc-stripe-payment-method-configurations.php
daledupreez: I just realised that this approach is a bit heavy-handed for reconnections, where we want to keep the same fallback PMC across the reconnection. I think we need to remove the fallback PMC removal from this PR for now.

daledupreez: Removed in ab5a9dd192cf879770d9e92e50b39b692c609237.

Mayisha: > where we want to keep the same fallback PMC across the reconnection.

What is the expectation with fallback PMC if the store is reconnecting to a different Stripe account?

daledupreez: As things stand, the logic will only use the fallback PMC if we got a response from the `/v1/payment_method_configurations` API _and_ the fallback PMC ID exists in that response.

It would definitely be better to clear the ID more proactively, but I think we need a dedicated PR to get that right.

Mayisha (approved):

Mayisha:
Noting one observation from my logs. I reconnected to a different Stripe account and observed multiple requests to PMC.

Screenshot 2025-12-16 at 4 10 10 PM

I can not consistently reproduce this. This happened just once during functional testing (once during my previous round of testing as well)

malithsen (approved):
Looks good, tests well

daledupreez:
Noting that I merged this despite the E2e tests failing -- the errors all seem related to ACH and are occurring for other PRs as well.

PR #4859, opened 2025-12-10
[Feature] Full enablement of Amazon Pay

daledupreez:

This PR maps to the `feature/amazon-pay-beta` feature branch, which we're using to keep the main Amazon Pay enablement code slightly separate from our development as a way to put together early builds for testing without committing the code to the `develop` branch, just in case we end up not being in a position to launch in the 10.3.0 release in January.

## Testing instructions

Testing will be managed via each PR, but the broad approach is as follows:
 * On a new site, install a build from this branch and set up your store
 * Connect a new Stripe account
 * If the Stripe account is from a country that Amazon Pay supports and where the store currency is supported, Amazon Pay should be enabled
 * When reconnecting an eligible account, we should also auto-enable Amazon Pay much like we do for Apple Pay and Google Pay
 * When upgrading from a previous version, Amazon Pay should not be auto-enabled after connecting or reconnecting
 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


malithsen (approved):
This looks good to me. Thanks for refactoring the `install` method, the flow is a lot clearer now.

I see some dead code, especially in the frontend (`isAmazonPayAvailable`), but I suspect it's intentionally left out to be removed in a follow-up release.

includes/class-wc-stripe-feature-flags.php
```diff
@@ -75,15 +75,7 @@ public static function get_option_with_default( $flag ) {
 	 * @return bool
 	 */
```
wjrosa, line 84: Shouldn't we include the `@deprecated` line already so we can remove this method later? And maybe add a call to `wc_deprecated_function`.

tests/phpunit/Admin/Migrations/Migrate_Payment_Methods_From_DB_To_PMC_Test.php
```diff
@@ -49,6 +49,24 @@ public function test_migration_not_executed_when_pmc_enabled_is_yes() {
 			->method( 'update_payment_method_configurations' );
 	}
 
+	public function test_migration_executed_when_pmc_enabled_is_yes_and_force_migration_is_true() {
```
wjrosa, line 58: This test is missing the PHPDoc.

wjrosa (approved):
Looks good to me! Left some optional comments

tests/phpunit/Admin/Migrations/Migrate_Payment_Methods_From_DB_To_PMC_Test.php
daledupreez: Good catch -- added in 6c5e4012cab66a0089019666bbc4bc53bba7f515.

includes/class-wc-stripe.php
```diff
@@ -328,52 +328,66 @@ public function install() {
...
+		if ( $is_new_install && false === get_option( 'wc_stripe_optimized_checkout_default_on' ) ) {
+			update_option( 'wc_stripe_optimized_checkout_default_on', true );
+		}
 
-			// TODO: Remove this when we're reasonably sure most merchants have had their
-			// settings updated like this. ~80% of merchants is a good threshold.
-			// - @reykjalin
-			$this->update_prb_location_settings();
+		if ( $is_new_install ) {
+			update_option( 'wc_stripe_amazon_pay_default_on', 'yes' );
```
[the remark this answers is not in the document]
daledupreez: I've renamed the existing test and updated it to include checks for the `wc_stripe_amazon_pay_default_on` option in 880973010cdf43230ef44d184cc0caf5f3c7afa4.

includes/class-wc-stripe-feature-flags.php
daledupreez: Good call! I've added `@deprecated` to this function and the `AMAZON_PAY_FEATURE_FLAG_NAME` constant and added a call to `wc_deprecated_function()` in 5bb8d4b2cd09c4bb0b69ff2190b83210db144806.

includes/class-wc-stripe-feature-flags.php
```diff
@@ -75,15 +75,7 @@ public static function get_option_with_default( $flag ) {
...
-		$enable_amazon_pay = 'yes' === self::get_option_with_default( self::AMAZON_PAY_FEATURE_FLAG_NAME );
-
-		/**
-		 * Filter to control the availability of the Amazon Pay feature.
-		 *
-		 * @since 10.1.0
-		 * @param bool $enable_amazon_pay Whether Amazon Pay should be enabled.
-		 */
-		return (bool) apply_filters( 'wc_stripe_is_amazon_pay_available', $enable_amazon_pay );
+		return true;
```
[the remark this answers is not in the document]
daledupreez: This was never intended as a way to disable Amazon Pay, so I think we should move forward with the feature. Merchants can disable Amazon Pay via the UI.

includes/class-wc-stripe-feature-flags.php
```diff
@@ -75,15 +75,7 @@ public static function get_option_with_default( $flag ) {
...
 	}
```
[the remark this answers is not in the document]
daledupreez: As per the suggestion in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4859/files#r2765812754, I have deprecated the constant and method in 5bb8d4b2cd09c4bb0b69ff2190b83210db144806.

includes/class-wc-stripe-feature-flags.php
daledupreez: I reverted the `wc_deprecated_function()` call in eb9d6db7faa21a3e776c5776f1d97a75902b4740, as it was causing ~150 unit test failures due to our internal code still calling the function. I've created https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4989 to track fixing that, but I want to make sure that work doesn't block or delay the 10.4.0 release next week.

wjrosa (approved):

PR #4864, opened 2025-12-10
Preserve express checkout button location settings during upgrade

### LINKED ISSUE #4861: Plugin install inverted `Show express checkouts on` settings
### Description

I've only ever displayed the payment buttons on my checkout page. Somehow the settings got changed to product and cart pages instead. I surely did not do that myself. My guess is that this happens during an install of this plugin.

### Steps to reproduce

I have no idea how to reproduce this.

### Expected behavior

A plugin install should not alter my existing settings.

### Screenshots

Image

### System report

I don't think this is relevant.

malithsen:
Fixes STRIPE-848
Fixes #4861

When merchants upgrade from versions prior to 10.1.0, their express checkout button location settings (`payment_request_button_locations`) were being ignored. Instead of copying the merchant's existing settings, the plugin was falling back to filter defaults (`['product', 'cart']`), effectively changing merchant-configured settings without merchant  consent.

**Root cause:** In version 10.1.0, `update_prb_location_settings()` was updated to check `express_checkout_button_locations` instead of `payment_request_button_locations`. However, merchants upgrading from older versions only have `payment_request_button_locations` set, so the check would find no existing value and fall through to the filter defaults.

**The fix:** Before falling back to filter defaults, the method now checks if `payment_request_button_locations` exists in the settings and copies it to `express_checkout_button_locations`. This preserves the original configuration during the upgrade.

## Testing instructions

To replicate the issue:
1.  Set up a test site with WooCommerce and Stripe plugin version 10.0.0 
2. Configure express checkout button locations to only show on the Checkout page.
3. Upgrade to the latest version.
4. Notice that the button locations are changed to Product and Cart in settings.

Test the fix
1. Generate a build of this branch
2. Repeat the previous steps, but instead of step 3, upgrade to the build from this branch.
3. Express button location setting should remain unchanged.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (approved):
The change looks good, and it tests well!

includes/class-wc-stripe.php
```diff
@@ -421,6 +421,14 @@ public function update_prb_location_settings() {
 			? $stripe_settings['express_checkout_button_locations']
 			: [];
 		if ( ! empty( $stripe_settings ) && empty( $prb_locations ) ) {
+			// Use existing payment_request_button_locations if it exists.
+			if ( array_key_exists( 'payment_request_button_locations', $stripe_settings ) ) {
+				$stripe_settings['express_checkout_button_locations'] = $stripe_settings['payment_request_button_locations'];
```
daledupreez, line 426: It feels like we should also unset the old, now-unused setting as part of this migration logic.
```suggestion
				$stripe_settings['express_checkout_button_locations'] = $stripe_settings['payment_request_button_locations'];
				unset( $stripe_settings['payment_request_button_locations'] );
```

daledupreez (approved):
This is working as described for me. I have one non-blocking suggestion, but I think we should take that on.

includes/class-wc-stripe.php
malithsen: Makes sense, Thank you!

PR #4867, opened 2025-12-11
Fix level3 data validation error caused by rounding precision

malithsen:
Fixes STRIPE-850

This PR fixes a rounding precision issue in Level3 data that caused Stripe to reject payments with the error: `Invalid level3: pricing must be consistent. Sum(unit_cost * quantity + tax_amount - discount_amount) + shipping_amount != total charged.`

When real-time shipping carriers (e.g., UPS) return rates with 3+ decimal precision (e.g., $20.685), the `get_stripe_amount()` function was truncating instead of rounding when converting to cents:
  - 20.685 * 100 = 2068.5
  - Previous behavior: `absint(wc_format_decimal(2068.5, 2))` → 2068 (truncated)
  - New behavior: `absint(round(2068.5))` → 2069 (rounded)

This PR proposes the two-decimal currency conversion in `WC_Stripe_Helper::get_stripe_amount()` to use `round()` before `absint()`, to ensure Level3 amounts match the charged amount.

## Testing instructions

1. Add the following snippet to simulate 3-decimal shipping cost
```
 add_filter( 'woocommerce_order_get_shipping_total', function( $total, $order ) {
      return 20.685;
  }, 10, 2 );
```
2. Place an order.
3. Check Stripe logs in WooComerce -> Status -> Logs.
4. In `develop` there will be an `Invalid level3: pricing must be consistent.` entry
5. In this branch, there should not be such errors.

CleanShot 2025-12-11 at 12 46 55@2x

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (approved):
Nice catch @malithsen 

The code looks good, and it tests well

Image

| develop | this branch |
|-|-|
| Image|Image|

Mayisha (commented):
@malithsen I am still seeing the level3 error when I pay with card. Could you please review my test setup and confirm if this is correct? 🤔 
 
I have used the following setup when testing,

- Selected product: Beanie ($18.00)
- Shipping method: Flat rate ($20.68)
- Added the script you shared in the description

```
add_filter( 'woocommerce_order_get_shipping_total', function( $total, $order ) {
      return 20.685;
  }, 10, 2 );
```

✅ If I pay with ECE (Google Pay), I do not get the level3 validation error in this branch.

**Log for payment intent request (Google Pay)**
Screenshot 2025-12-15 at 12 06 50 PM

⚠️ If I pay with card, I still see the error in my logs

**Log for payment intent request (Card)**
Screenshot 2025-12-15 at 12 07 32 PM

**Error log**
Screenshot 2025-12-15 at 12 07 55 PM

malithsen:
> ⚠️ If I pay with card, I still see the error in my logs

@Mayisha Hm, I am not able to replicate that.
Maybe there's another code path we need to adjust? In my tests, `get_stripe_amount` is always called, regardless of whether express checkout or card, so the rounding fix is applied.

Can you share more about your setup? Is OC enabled? Are you on classic/blocks checkout?

Mayisha (approved):
Thanks @malithsen for helping me find the issue in my setup. As you discovered the filter added in the snippet is not working in classic checkout, this was the reason I was still getting the error.

Tested again in block checkout with the filter and confirmed that the fix is working 👍

PR #4870, opened 2025-12-11
Renaming the ECE customization route

wjrosa:

In this simple PR, I am renaming the express checkout customization route to comply with the feature.

## Testing instructions

- Checkout and build this branch on your test environment (`dev/renaming-the-ece-customization-route`)
- Connect your Stripe account
- Go to the Stripe payment methods list screen
- Click the "Customize" link next to Apple Pay / Google Pay
- Confirm you are redirected to the correct screen

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (approved):
Looks good to me 👍

includes/class-wc-stripe.php
```diff
@@ -235,7 +235,7 @@ public function init() {
 			require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-stripe-admin-notices.php';
 			require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-stripe-settings-controller.php';
 
-			if ( isset( $_GET['area'] ) && 'payment_requests' === $_GET['area'] ) {
+			if ( isset( $_GET['area'] ) && 'express_checkout' === $_GET['area'] ) {
```
daledupreez: This approach will break existing deep links to this page. I think we should either allow the old name, as per this suggestion
```suggestion
			if ( isset( $_GET['area'] ) && in_array( $_GET['area'], [ 'express_checkout', 'payment_requests' ], true ) ) {
```

I think it would be better if we redirect users to the new `area` when we receive the old one, though I am less sure of where that code should live. It may be simplest to accept both parameters in this PR for the time being, and handle the full redirect in a dedicated PR.
```php
			if (
				isset( $_GET['page'], $_GET['tab'], $_GET['section'], $_GET['panel'], $_GET['area'] ) &&
				'wc-settings' === $_GET['page'] &&
				'checkout' === $_GET['tab'] &&
				'stripe' === $_GET['section'] &&
				'methods' === $_GET['panel'] &&
				'payment' === $_GET['area']
			) {
				wp_safe_redirect( /* fixed URL */, WP_Http::MOVED_PERMANENTLY );
				exit;
			}
```

daledupreez (commented):
This is working as described, but I think we need to handle the backward compatibility more gracefully.

includes/class-wc-stripe.php
wjrosa: Good idea 👍 . Suggestion applied

daledupreez (approved):
The changes LGTM. 🚀

wjrosa:
Unit tests are failing for reasons unrelated to the changes in this PR. They are passing locally. I will merge the PR now, and check what might be happening later

PR #4871, opened 2025-12-11
Deprecating and replacing PRBs classes

wjrosa:

In this cleaning-up PR, I am deprecating existing backend classes related to PRBs and moving the logic to new, ECE-named classes.

## Testing instructions

- Checkout to this branch on your test environment (`dev/deprecating-and-replace-prbs-classes`)
- Connect your Stripe account
- Enable express checkout methods
- As a shopper, access the store using a public-facing URL
- Add a product to your cart
- Confirm you can complete a payment using the express methods

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/constants/class-wc-stripe-payment-request-button-states.php
diegocurbelo: I think we should replace this with the list from the copied class; we reduce the code duplication, and prevent any future PR from only updating one of the copies:
```php
	const STATES = WC_Stripe_Express_Checkout_Button_States::STATES;
```

includes/admin/class-wc-stripe-express-checkout-controller.php
```diff
@@ -0,0 +1,78 @@
...
+	public function admin_options() {
+		global $hide_save_button;
+		$hide_save_button = true;
+		$return_url       = admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe' );
+		$header          = __( 'Customize express checkouts', 'woocommerce-gateway-stripe' );
+		$return_text     = __( 'Return to Stripe', 'woocommerce-gateway-stripe' );
+
+		WC_Stripe_Helper::render_admin_header( $header, $return_text, $return_url );
+
+		echo '<div class="wrap"><div id="wc-stripe-express-checkout-settings-container"></div></div>';
```
diegocurbelo, line 85: If we use a different Id here, we also need to update it in `client/entrypoints/express-checkout-settings/index.js`, otherwise the Express Checkout settings page won't load:
https://github.com/woocommerce/woocommerce-gateway-stripe/blob/6226ac949db0a477a2f86dcf27a0b24886ba089f/client/entrypoints/express-checkout-settings/index.js#L5-L11

includes/admin/class-wc-stripe-express-checkout-controller.php
```diff
@@ -0,0 +1,78 @@
...
+
+		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
+		$params          = [
+			'key'            => WC_Stripe_Mode::is_test() ? $stripe_settings['test_publishable_key'] : $stripe_settings['publishable_key'],
+			'locale'         => WC_Stripe_Helper::convert_wc_locale_to_stripe_locale( get_locale() ),
+			'is_ece_enabled' => WC_Stripe_Feature_Flags::is_stripe_ece_enabled(),
+		];
+		wp_localize_script(
+			'wc-stripe-express-checkout-settings',
+			'wc_stripe_express_checkout_settings_params',
```
diegocurbelo, line 55: If we use a different object name here, we also need to update the global in `client/entrypoints/express-checkout-settings/express-checkout-preview-component.js` (and the corresponding tests) that depend on it:

Image

diegocurbelo (changes requested):
The `Customize express checkouts` page doesn't work with the new `root element id` and the new `global js param object` in the new  `WC_Stripe_Express_Checkout_Controller`:

Image

We can:
1. Use the old names in the new controller
2. Use the new names in the new controller, update the JS, and update the deprecated controller to also use the new names
3. Use the new names in the new controller and update the JS with extra logic to use the old ones if the new ones are not found

I think option 2 is the cleanest solution.

includes/constants/class-wc-stripe-payment-request-button-states.php
wjrosa: Good idea. Applied it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4871/commits/2ac4fc6351b3de24ba8780d2d25528993a7e2792

includes/admin/class-wc-stripe-express-checkout-controller.php
wjrosa: Oops, forgot about that 🤦 . Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4871/commits/b64356487439d66859b2e7a50dd7a439d181169d

includes/admin/class-wc-stripe-express-checkout-controller.php
wjrosa: You're right. Updated all references in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4871/commits/7d3c6892d556d996f198e976be2056f10cca8b03

includes/constants/class-wc-stripe-payment-request-button-states.php
```diff
@@ -31,6 +29,10 @@
  *    5. The following countries are not listed in WC:
  *        AC (Ascension Island), IC (Canary Islands), EA (Ceuta & Melilla), CP (Clipperton Island), DG (Diego Garcia),
  *        TA (Tristan da Cunha), XK (Kosovo).
+ *
+ * @since 5.1.0
+ *
+ * @deprecated 10.3.0 Moved to includes/wc-stripe-express-checkout-button-states.php
```
malithsen:
```suggestion
 * @deprecated 10.3.0 Moved to includes/constants/wc-stripe-express-checkout-button-states.php
```

malithsen (approved):
Looks good

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.3.0 - xxxx-xx-xx =
+* Dev - Deprecates and replaces PRBs classes with ECE equivalents
```
daledupreez: Nit: I think we should use longer names here rather than the acronyms.
```suggestion
* Dev - Deprecates and replaces Payment Request Button classes with Express Checkout equivalents
```

includes/constants/class-wc-stripe-payment-request-button-states.php
```diff
@@ -31,6 +29,10 @@
  *    5. The following countries are not listed in WC:
  *        AC (Ascension Island), IC (Canary Islands), EA (Ceuta & Melilla), CP (Clipperton Island), DG (Diego Garcia),
  *        TA (Tristan da Cunha), XK (Kosovo).
+ *
+ * @since 5.1.0
+ *
+ * @deprecated 10.3.0 Moved to includes/constants/wc-stripe-express-checkout-button-states.php
  */
```
daledupreez, line 38: To be extra-safe, we could add a `require_once` that includes the new class file. That way any dependent code that loads this file directly will continue to work as it did before.

changelog.txt
daledupreez: Actually I think this change may need to be described as an `Update` due to the deprecation. This is something we should flag as much as possible. I think it should also get an entry in the "Compatibility Notes" section of the main readme.txt file.

daledupreez (approved):
The changes look good to me and test well. I do think we should flag the change as an `Update` rather than as `Dev`, however. It also looks like it would be good to add some PHPDoc or `void` return types to some of the functions you added.

I also see that we need to add some logic for our baseline constants that we create via `define()` to keep PHPStan happy. I'll work on that in a quick PR.

includes/constants/class-wc-stripe-payment-request-button-states.php
wjrosa: Good idea. Addressed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4871/commits/cc1222c2abb6aec6c383bf7497ca38679227bea4

wjrosa:
Thanks for the review and for pointing those things out. I have addressed the changelog entry prefix, PHPDoc update, the missing return types, and also the removal (by mistake) of the Ukraine states Diego added in my latest commits.

PR #4874, opened 2025-12-15
Add PHPStan analysis workflow

daledupreez:

This PR adds a PHPStan analysis workflow to the Stripe plugin, which is based heavily on the work done for the WooCommerce plugin in https://github.com/woocommerce/woocommerce/pull/62064. We want to leverage PHPStan to address clear issues in our code based on static analysis. However, we're using the same approach as the WooCommerce plugin, where we're making it easier to approach things incrementally:
 * We will run PHPStan on every PR as a normal workflow, which should catch any new issues in code we introduce
 * We have set up a baseline in `phpstan-baseline.neon` so PHPStan will be "clean" against our current issues
 * For any PRs that fix issues tracked in the baseline file, we'll need to regenerate the baseline to remove those defunct/resolved problems.

This means we can attack issues incrementally rather than needing a clean baseline immediately. (This follows the approach discussed internally in pdV5qK-10a-p2.) There are definitely some issues that we should fix ASAP, including missing null checks and some PHPDoc parse errors that may be masking issues, but there are plenty of issues that we can resolve more incrementally, e.g. the hundreds of functions that don't document a return type.

There are a few things that are worth flagging/noting:
 * We are using a PHPStan level of 8, which matches the current level used by WooCommerce
    - The initial PR used level 9, but it was reduced to level 8 to avoid unnecessary noise stemming from loose types in WordPress and WooCommerce
 * The PR adds new `phpstan` and `phpstan:baseline` scripts that can be run via `composer` or `npm`:
   - `composer run-script phpstan`
   - `composer run-script phpstan:baseline`
   - `npm run phpstan`
   - `npm run phpstan:baseline`
 * The PR adds the capability for stubs to be added to `php-stubs/` - any PHP file added in this directory will be loaded by PHPStan, but not included in plugin releases etc.

## Testing instructions

The actual workflow testing will only be possible in GitHub _after_ we've merged the PR. If you _really_ want, you can install `act` via HomeBrew (`brew install act`) and run the GitHub workflow from your local Stripe directory as follows:
```
act -W .github/workflows/phpstan.yml
```

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


phpstan-stubs/WCS_Background_Repairer.php
wjrosa, line 1: I would consider renaming the folder to `phpstan-stubs` to make the purpose more clear

wjrosa (approved):
Nice work, Dale! When I joined the team, I proposed adding SonarQube to our workflows. Since we usually prefer open-source solutions, someone suggested PHPStan. But I never got a chance to look into it. So, it is great to see something that contains some of the SonarQube features added 🎉

PS: TIL about Act ❤️

phpstan-stubs/WCS_Background_Repairer.php
daledupreez: Good idea -- I'll do that shortly!

daledupreez: Done in 8f47ac06713df57fb6436846c472eb9ca5dbe2e4 - it did require some minor changes elsewhere to exclude from PHPCS and to include the stubs.

malithsen (approved):
Looks great! :shipit:

PR #4882, opened 2025-12-19
Fix default locations for Amazon Pay

### LINKED ISSUE #4881: Amazon Pay should default to the same locations as other express checkout methods
### Description

While testing Amazon Pay, I noticed that for new installs, the other express checkout methods default to showing in the product page, cart, and checkout, but Amazon Pay defaults to the product page and the cart. We should ensure they match up.

### Steps to reproduce

1. Install WooCommerce and Stripe on a brand new site
2. Ensure that the Amazon Pay feature flag is enabled, or you are using a branch that has this feature flag enabled: `wp option update _wcstripe_feature_amazon_pay yes`
3. Connect to Stripe via the plugin
4. Navigate to the Payment Methods tab in the Stripe settings
5. Click on the Customize option for Apple Pay and Google Pay
6. Verify that they are configured to show up in the product page, cart, and checkout.
7. Navigate to the Payment Methods tab
8. Click on the Customize option for Amazon Pay
9. Notice that Amazon Pay only defaults to showing in the cart and product page, but _not_ checkout.
 
### Expected behavior

Amazon Pay should use the same default locations as the other express checkout options.

### Additional context

It looks like the issue stems from this code in `includes/admin/stripe-settings.php`:
https://github.com/woocommerce/woocommerce-gateway-stripe/blob/2924752103277a5ff5cb256c7f3c25ed78fe8c4c/includes/admin/stripe-settings.php#L294-L295

We're only defaulting on checkout for the other options, and not Amazon Pay.

daledupreez:
Fixes STRIPE-859
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4881

This PR ensures that Amazon Pay picks up the same default locations as the other express checkout methods by always defaulting on the checkout location outside of the `wc_stripe_settings` filter.

Note that this PR doesn't perform any checks when enabling Amazon Pay. We _might_ want to update the logic in `WC_Stripe_Payment_Method_Configurations::maybe_migrate_payment_methods_from_db_to_pmc()` to make the value of `amazon_pay_button_locations` match the value of `express_checkout_button_locations`, but for new installs, we should always be migrating the PMC configuration before the user can modify any settings. After that stage, we should generally leave the settings as-is.

## Testing instructions

* Install WooCommerce on a brand new site that has never had the Stripe plugin installed
* Install the Stripe plugin built from this branch
* Connect to Stripe
* After connecting to Stripe, confirm that Amazon Pay is enabled
* Also confirm that Amazon Pay is configured to display in the product page, cart, and checkout.
* Also check that Apple Pay and Google Pay are still configured to display in the product page, cart, and checkout.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/stripe-settings.php
+++ b/includes/admin/stripe-settings.php
@@ -291,8 +291,9 @@ $stripe_settings = apply_filters(
 	]
 );
 
-// in the new settings, "checkout" is going to be enabled by default (if it is a new WCStripe installation).
+// Ensure express checkout buttons are displayed on the checkout page by default.
 $stripe_settings['express_checkout_button_locations']['default'][] = 'checkout';
+$stripe_settings['amazon_pay_button_locations']['default'][]       = 'checkout';
 
 // no longer needed in the new settings.
 unset( $stripe_settings['express_checkout_button_branded_type'] );
--- a/changelog.txt
+++ b/changelog.txt
@@ -12,6 +12,7 @@
 * Fix - Preserve express checkout button location settings when upgrading from older plugin versions
 * Fix - Fix some initialization bugs for reconnections
 * Fix - Update Ukraine state mapping list
+* Fix - Use same default locations for Amazon Pay express checkout
 
 = 10.2.0 - 2025-12-08 =
 * Dev - Refactor display logic for payment method issue pills
--- a/readme.txt
+++ b/readme.txt
@@ -151,5 +151,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Fix - Preserve express checkout button location settings when upgrading from older plugin versions
 * Fix - Fix some initialization bugs for reconnections
 * Fix - Update Ukraine state mapping list
+* Fix - Use same default locations for Amazon Pay express checkout
 
 [See changelog for full details across versions](https://raw.githubusercontent.com/woocommerce/woocommerce-gateway-stripe/trunk/changelog.txt).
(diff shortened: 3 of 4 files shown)


includes/admin/stripe-settings.php
```diff
@@ -291,8 +291,9 @@
 	]
 );
 
-// in the new settings, "checkout" is going to be enabled by default (if it is a new WCStripe installation).
+// Ensure express checkout buttons are displayed on the checkout page by default.
 $stripe_settings['express_checkout_button_locations']['default'][] = 'checkout';
+$stripe_settings['amazon_pay_button_locations']['default'][]       = 'checkout';
```
Mayisha, line 296:
```suggestion

```

We can remove this code and add `checkout` in the `default` list of `express_checkout_button_locations` (line 192) and `amazon_pay_button_locations` (line 252)

daledupreez: Based on my reading of the code, I believe this separate location is intentional -- those locations are all within the data passed to the `wc_stripe_settings` filter, while this separate location looks like it is forcing the default to be on after the filter has been applied.

Mayisha: Not a blocker, but my understanding is this section was inside `if UPE is enabled` [check before](https://github.com/woocommerce/woocommerce-gateway-stripe/commit/dbb908b92bf430f2564df99d71607221ecd919cc#diff-32613ee37c492be638d26bb11a6311a43606306cfebee187a2316efb7d3f1236?w=1). Now that we have UPE enabled by default, these should be part of the original `$stripe_settings` above.

daledupreez: I didn't check the history for this, so my assumptions were clearly wrong! Is it OK if I add the change as-is in this PR and then tackle the shift in a follow-up?

Mayisha: Yup, as I mentioned before it's a non blocker suggestion 👍

Mayisha (approved):
Looks good to me :shipit:

malithsen (approved):
LGTM. Works as described!

PR #4883, opened 2025-12-19
Redirecting to Stripe settings upon activation

wjrosa:
See: peNR48-23P-p2

As part of the latest hack week, I decided to redirect merchants to the Stripe settings page upon plugin activation. The goal is to make it easier for them to connect their account and set up the extension. This is similar to the behavior found in the WooCommerce plugin.

## Testing instructions

- Checkout and build this branch (`hack/redirect-to-stripe-settings-upon-activation`)
- Install the extension on a new website (i.e., a JN website)
- Activate the extension
- Confirm you are correctly redirected to the Stripe settings screen

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe.php
```diff
@@ -99,6 +99,7 @@ public function __wakeup() {}
 	 */
 	public function __construct() {
 		add_action( 'admin_init', [ $this, 'install' ] );
+		add_action( 'admin_init', [ $this, 'maybe_redirect_to_stripe_settings' ] );
```
Mayisha: Should we add a lower priority to this one so that the previous callback `install` runs first?

includes/class-wc-stripe.php
```diff
@@ -386,6 +387,21 @@ public function install() {
...
+	 * Redirects to the Stripe settings page upon plugin activation if the transient is set,
+	 * and if not activating multiple plugins at once.
+	 *
+	 * @return void
+	 */
+	public function maybe_redirect_to_stripe_settings(): void {
+		if ( get_transient( 'wc_stripe_redirect_to_settings' ) && ! isset( $_GET['activate-multi'] ) ) {
+			delete_transient( 'wc_stripe_redirect_to_settings' );
+
+			wp_redirect( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe' ) );
```
Mayisha:
```suggestion
			wp_safe_redirect( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe' ) );
```
It would be better to use `wp_safe_redirect` here.

Mayisha (commented):
This is a nice enhancement 👍 

I have tested the following scenarios in a JN site-
✅ Activate in a fresh JN site -> redirect works
✅ Deactivate the plugin and activate from the plugins page -> redirect works
✅ Deactivate the plugin, activate from the `view details` modal on the plugins page -> redirect works
⚠️ Deactivated the plugin. Activated Stripe and another plugin using the bulk action -> redirects to the Stripe settings page. Is it expected? From multi check in the code it looks like an unexpected behavior.
⚠️  Activate the latest released version of the plugin. Update the plugin with the build of this PR by uploading -> no redirects. Is it expected?

daledupreez:
Noting that I will review this PR after we resolve Mayisha's open questions above.

includes/class-wc-stripe.php
wjrosa: Yeah, it makes sense. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4883/commits/bc4dedc559b1fa64286926797c5d26d14ef31d6e

wjrosa:
Thanks for the review, Mayisha!

> ⚠️ Deactivated the plugin. Activated Stripe and another plugin using the bulk action -> redirects to the Stripe settings page. Is it expected? From multi check in the code it looks like an unexpected behavior.

That's not expected. I fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4883/commits/6a169a9ccb4de332089ee66fd46377f81f6d0abe

> ⚠️ Activate the latest released version of the plugin. Update the plugin with the build of this PR by uploading -> no redirects. Is it expected?

That's the correct behavior 👍

Mayisha (approved):
Thanks for the confirmation. Code looks good to me and it tests well 👍 

**Note**: The `test_maybe_redirect_to_stripe_settings` test is failing. We need to set `$_GET['activate'] = 'true';` in the test to as the condition has changed now.

includes/class-wc-stripe.php
```diff
@@ -375,6 +376,21 @@ public function install() {
...
+	 *
+	 * @return void
+	 */
+	public function maybe_redirect_to_stripe_settings(): void {
+		if ( get_transient( 'wc_stripe_redirect_to_settings' ) && isset( $_GET['activate'] ) ) {
+			delete_transient( 'wc_stripe_redirect_to_settings' );
+
+			wp_safe_redirect( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe' ) );
+			exit;
+		}
```
daledupreez: It feels like we should clean up the transient any time it is truthy, not just when the `activate` parameter is set. I know we're using a pretty short timeout for the transient, but I think it's good to clean up as well as we can to avoid unexpected weirdness because the transient is still set from a different user action.
```suggestion
		if ( get_transient( 'wc_stripe_redirect_to_settings' ) ) {
			delete_transient( 'wc_stripe_redirect_to_settings' );

			if ( isset( $_GET['activate'] ) ) {
				wp_safe_redirect( admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe' ) );
				exit;
			}
		}
```

tests/phpunit/WC_Stripe_Test.php
```diff
@@ -293,6 +299,45 @@ public function test_update_prb_location_settings_uses_filter_defaults_when_no_e
...
+
+		// Add filter to catch redirects.
+		add_filter( 'wp_redirect', $wp_redirect_filter );
+
+		// Set the transient to trigger redirection.
+		set_transient( 'wc_stripe_redirect_to_settings', true, 30 );
+
+		try {
+			WC_Stripe::get_instance()->maybe_redirect_to_stripe_settings();
+		} catch ( WC_Stripe_Catch_Redirect_Exception $e ) {
```
daledupreez: Side comment: by using the short name for this class, it's no longer visible that this is a test-only exception class, as the `Tests` element is in the namespace.

Not blocking either way, but I almost missed that this is a test-only class as a result.

woocommerce-gateway-stripe.php
```diff
@@ -120,6 +120,17 @@ function add_woocommerce_inbox_variant() {
...
 register_activation_hook( __FILE__, 'add_woocommerce_inbox_variant' );
 
+register_activation_hook( __FILE__, 'set_settings_redirection_transient' );
+
+/**
+ * Set a transient to redirect the user to the settings page upon activation.
+ *
+ * @return void
+ */
+function set_settings_redirection_transient(): void {
```
daledupreez: This file is not namespaced, so I think we need to use a more plugin-specific function name to make collisions with other plugins way less likely.
```suggestion
function wc_stripe_set_settings_redirection_transient(): void {
```

daledupreez (commented):
The change is working as described, but we need to update the unit tests, and I think we  also need to rename the new activation function to avoid collisions.

tests/phpunit/WC_Stripe_Test.php
wjrosa: I see! I decided to remove the custom exception class in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4883/commits/8d4249df7ace0a7ea50856fbea947d8ecfe8809c

tests/phpunit/WC_Stripe_Test.php
```diff
@@ -6,6 +6,7 @@
 use WC_Stripe_Helper;
 use WC_Stripe_Payment_Methods;
 use WC_Stripe_UPE_Payment_Gateway;
+use WooCommerce\Stripe\Tests\Exceptions\WC_Stripe_Catch_Redirect_Exception;
```
daledupreez: We should remove this `use` statement for the test exception.
```suggestion
```

readme.txt
```diff
@@ -140,6 +140,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.4.0 - xxxx-xx-xx =
+* Update - Redirects merchants to the Stripe settings screen upon plugin activation
```
daledupreez: Nit: we should use `Redirect` here.
```suggestion
* Update - Redirect merchants to the Stripe settings screen upon plugin activation
```

daledupreez (approved):
Thanks again for your patience! This is looking excellent. 💟

PR #4887, opened 2025-12-23
Simplify default settings and only call wc_stripe_settings filter once

daledupreez:
Related:
* https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4882

This PR tackles some lingering complexity in the Stripe default settings code that resulted in us calling the `wc_stripe_settings` filter twice with different default values. This PR ensures that we push all the default values into the initial array of data, which then avoids the need for secondary modifications and a second call to the `wc_stripe_settings` filter.

## Testing instructions

I think code inspection should be sufficient for this one.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/admin/stripe-settings.php
```diff
@@ -290,23 +291,3 @@
 		],
 	]
 );
-
-// Ensure express checkout buttons are displayed on the checkout page by default.
-$stripe_settings['express_checkout_button_locations']['default'][] = 'checkout';
-$stripe_settings['amazon_pay_button_locations']['default'][]       = 'checkout';
-
-// no longer needed in the new settings.
-unset( $stripe_settings['express_checkout_button_branded_type'] );
```
wjrosa, line 299: Shouldn't we account for these `unset` calls as well? Directly removing these keys above.

malithsen (approved):
Looks good

includes/admin/stripe-settings.php
daledupreez: These keys are no longer being set by our code. (I just triple-checked!) And while we were previously calling `unset()` after the first call to the `wc_stripe_settings` filter, the items could still be added back after the second call to the filter.

wjrosa: I see. So we will keep them for now, right?

wjrosa (approved):
Still looks good to me 👍 . Just double-checking your response on the `unset` calls.

includes/admin/stripe-settings.php
daledupreez: The values are not present, and we were not removing them after the second filter call, so I think it's OK to ignore any situations where merchants were returning these values, as they were already returning them from the second filter call.

PR #4888, opened 2025-12-23
Update README.md to document usage of PHPStan

daledupreez:
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4874

This PR updates the main README.md to document our usage of PHPStan, and how developers can interact with it. This is heavily modeled on the approach being used for WooCommerce Core at present.

## Testing instructions

Review the instructions, and confirm they look OK.

---

### Changelog entry

<details>

#### Comment 
This PR only modifies our developer-facing documentation, so does not warrant a changelog entry in the plugin.
</details>

**Post merge**


README.md
```diff
@@ -22,6 +22,33 @@ If you run into errors with `npm install` it may be due to node version, try `nv
...
 When running the `composer install/update`, Composer may prompt you for a GitHub OAuth token before it can fetch the `subscriptions` and `pre-orders` extensions from GitHub.
 
+### PHPStan workflow
+
+We are using [PHPStan](https://phpstan.org/) to perform static analysis of our code. Every PR will be analysed to check for issues via a GitHub workflow, which may include the following:
+ * Legitimate issues that you should resolve before merging, like missing null check or type checks.
+ * Error messages that are less clear if they need to be resolved, like array types not specifying a type for their contents.
+ * Error messages that indicate an error doesn't exist or could not be found.
+
+For the latter two situations, you will need to update the PHPStan baseline file, which you can do by running either of the following commands:
```
wjrosa, line 34: It is not very clear to me when we need to update the baseline file. Is it after fixing the issues reported in the workflow?

Also, unrelated to this PR, but do you think it makes sense to add PHPStan to our pre-commit hooks?

daledupreez: > It is not very clear to me when we need to update the baseline file. Is it after fixing the issues reported in the workflow?

To be honest, I am not 100% sure of the decision around what changes we should fix in the PR and which changes should drive us to update the baseline! :-/
 * For cases where we fixed an existing issue in the baseline, we should update the baseline.
 * For cases where PHPStan is flagging a legitimate run-time issue (e.g. that a variable may be null/false), we should fix the issue.
 * For cases where it's less clear that we _need_ to fix the problem, like array return values not being explicitly typed, I think we need to work on what the next step should be, but I don't know what that is or should be for now.
    - In particular, I think we may want to ignore the explicit typing for array values reported from the `missingType.iterableValue` error type (which is ignored in WooCommerce), but I want some sense of what we'd be removing before we start ignoring specific errors.

daledupreez: > Also, unrelated to this PR, but do you think it makes sense to add PHPStan to our pre-commit hooks?

I think that would probably be more helpful in terms of workflow, TBH. I will try to take a look at implementing that tomorrow.

wjrosa: > To be honest, I am not 100% sure of the decision around what changes we should fix in the PR and which changes should drive us to update the baseline! :-/ (...)

Gotcha! I think we can improve this specific part, then, with some of the explanations you shared here

malithsen (approved):

README.md
daledupreez: @wjrosa, let me know what you think of my additional note from 7ce2b6a067d97ec8c5c1630fbcf8e9d27b465cef.

wjrosa: It is good for me. Thanks for adding it. Approved 👍

wjrosa (approved):

PR #4891, opened 2025-12-24
Fix PHPStan handling of plugin constants defined via function calls

daledupreez:

This PR works around [a bug or limitation in PHPStan](https://github.com/phpstan/phpstan/issues/11210) where it does not recognise constants defined using the return value from a function. The issue was resulting in the `WC_STRIPE_PLUGIN_URL` and `WC_STRIPE_PLUGIN_PATH` constants not being found/seen by PHPStan. To work around the issue, I've added conditional definitions for those two constants in `phpstan-stubs/wc_stripe_constants.php` so they get picked up during PHPStan analysis runs. Note that the values are not correct, as we need to use constant values for both.

I have also updated the PHPStan baseline to remove the previously tracked errors stemming from the "missing" constants.

## Testing instructions

Code review should be good enough on this one, and the PHPStan analysis should be clean.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/phpstan-stubs/wc_stripe_constants.php
+++ b/phpstan-stubs/wc_stripe_constants.php
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * Stubs for constants defined via function calls in WooCommerce Stripe that
+ * PHPStan doesn't seem to pick up correctly. As of 2025-12-24, this is not resolved.
+ *
+ * @see https://github.com/phpstan/phpstan/issues/11210
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+if ( ! defined( 'WC_STRIPE_PLUGIN_URL' ) ) {
+	define( 'WC_STRIPE_PLUGIN_URL', 'https://example.com/wp-content/plugins/woocommerce-gateway-stripe' );
+}
+
+if ( ! defined( 'WC_STRIPE_PLUGIN_PATH' ) ) {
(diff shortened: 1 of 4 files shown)


phpstan-stubs/wc_stripe_constants.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+/**
+ * Stubs for constants defined via function calls in WooCommerce Stripe that
+ * PHPStan doesn't seem to pick up correctly. As of 2025-12-24, this is not resolved.
+ *
+ * @see https://github.com/phpstan/phpstan/issues/11210
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
```
malithsen: Is this check needed for PHPStan? It might give the impression that this is a WP runtime file.

malithsen (approved):
LGTM

phpstan-stubs/wc_stripe_constants.php
daledupreez: Hmmm... our default development Docker configuration makes these files available via the plugin directory, and those environments can be exposed on the internet when we use tools like Jurassic Tube. Also, this is not the only directory we expose. 😬 I've opened https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4897 to address that gap, and I have removed the check from this PR in 1e0f5c1b47aa187469ca56459951878574a7b10d.

wjrosa (approved):
Looks good to me 👍

Issue #4855, opened 2025-12-09
Express Checkout Script Dependencies Mistakenly Added to Product Pages


thekendog:
### Description

The main stripe js file, wc-country-select, wc-address-i18n, wc-checkout, wp-i18n, etc. are loaded on all product pages even when Express Checkout on product pages is turned off.

### Steps to reproduce

1. Toggle off express checkout for product pages.

### Expected behavior

These scripts and styles should not be loaded since I have express checkout turned off and I am not on the checkout page.

thekendog:
After looking at things some more, it appears you have to use the "wc_stripe_load_scripts_on_product_page_when_prbs_disabled" filter and return false since it returns true by default.  Why?  Why would you need to do this instead of just using the setting?

malithsen:
Thanks for the report @thekendog 
I can replicate the issue. From a cursory look, I think two issues are contributing to unnecessary script loads. As you pointed out, `wc_stripe_load_scripts_on_product_page_when_prbs_disabled` should not be returning `true` by default. Secondly, I see that `register_express_checkout_script` is being called even when express checkouts is disabled on product pages, so there might be a missing check there.

els2025:
Any update on this bug its been 2 weeks? 
I have even added `add_filter( 'wc_stripe_load_scripts_on_product_page_when_prbs_disabled', '__return_false' );`  on child theme function.php file and disable stripe express checkout on product page but still no change the stripe express checkout files still load on all product page

malithsen:
Thanks for checking in @els2025, and sorry for the delay.

This issue has been triaged, but work on a fix hasn’t started yet. As noted earlier, there appear to be two underlying problems:

1. `wc_stripe_load_scripts_on_product_page_when_prbs_disabled` defaulting to true when it shouldn’t.
2. Express Checkout scripts being registered even when product-page express checkout is disabled.

Because of the second issue, the filter alone isn’t sufficient to fully prevent the scripts from loading, which matches what you’re seeing.

The issue is tracked and prioritized in our backlog, but I don’t have an ETA yet for when the fix will land. We’ll update this thread once an update is available.

Thanks for your patience and for helping surface the problem.

daledupreez:
@Mayisha, can you confirm that #4893 only addressed the addition of scripts when they shouldn't be? I believe the filter handling still needs to be addressed.

Mayisha:
There are two problems mentioned in this issue @daledupreez -

[...]

PR #4893, opened 2025-12-24
Add high-level check for express checkout being enabled

daledupreez:
Towards STRIPE-845
Towards https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4855

This PR fixes one of the issues raised in https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4855 which is express checkout scripts are loaded on the product page even when it's disabled for product page. 
@daledupreez investigated and found that with Amazon Pay also being a possible express checkout option, we need to shift our logic to detect if any express checkout mechanisms are enabled. To be specific this [should_show_ece_on_location](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/fix/express-checkout-script-logic/includes/payment-methods/class-wc-stripe-express-checkout-helper.php#L943) function is returning true as we check both ECE and Amazon Pay in this function.

In this PR, we have the `is_express_checkout_enabled` check early in the flow of `should_show_express_checkout_button` function. 'is_express_checkout_enabled()' checks locations as well as the necessary express checkout methods being available.

Additionally, added check for one page checkout in the 'is_enabled_for_current_context' function so that the one page checkout is also taken into consideration in the flow of 'is_express_checkout_enabled()' fucntion.

## Testing instructions

1. Go to Stripe settings page.
2. Enable Google/Apple Pay. Enable them in the cart and checkout but disable them on the product page. Save the settings.
3. Enable Amzaon Pay temprarily. Enable it in all the pages( cart, checkout and product). Save the settings.
4. Disable Amazon Pay and Save the settings.
(Steps 2-4 are done this way to reproduce the issue. So that we have Google/Apple Pay enabled but turned off on product page while Amazon Pay is disabled but the product page is set as the allowed location for Amazon Pay in database)
5. As a shopper go to any product page and check the network tab. 
6. In `develop` branch, notice that the `express-checkout` file is loaded on this page.
7. In `fix/express-checkout-script-logic` branch, ensure that the script is not loaded.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -671,6 +671,12 @@ public function should_show_express_checkout_button() {
 			return false;
 		}
 
+		// Note that is_express_checkout_enabled() checks locations as well as the necessary express checkout methods being available.
+		if ( ! $this->is_express_checkout_enabled() ) {
+			WC_Stripe_Logger::log( 'No Stripe Express Checkout options are enabled in the current context.' );
```
wjrosa: We should use the new log methods here (probably `info()` or `debug()`).

wjrosa (approved):
Looks good to me! Left an observation

malithsen (approved):
Thanks for the fix Dale, this works as described.

The other issue noted in https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4855 is that the `wc_stripe_load_scripts_on_product_page_when_prbs_disabled` filter has the wrong default value. If we are planning to fix that separately, I'd suggest we use "Towards" prefix instead of "Fixes" in PR description to avoid closing 4855 automatically when this gets merged.

daledupreez:
Thanks for getting this shipped, @Mayisha! 💟

PR #4897, opened 2025-12-29
Fix unnecessary file exposure via Docker

daledupreez:

In [this review comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4891/files#r2646255967),  @malithsen flagged that PHP files in the `phpstan-stubs` subdirectory probably don't need an `ABSPATH` check. I explored further and confirmed that we are unnecessarily exposing multiple subdirectories via our development Docker environment that are probably safer not to expose. This PR ensures that the following subdirectories are **not** mirrored between the development directory and the Docker environment, largely because we do expose the directories via external hosts for certain types of testing, which could expose us to unexpected vulnerabilities or exposure:
* `docker` -- this directory includes Apache and PHP logs, which we shouldn't expose unnecessarily; it also exposes scripts in `docker/bin`, though I am not sure if these would be executable from a web server context.
* `node_modules` -- we have a number of indirect dependencies that probably wouldn't be callable, but best to be safe, especially seeing as there shouldn't be any code we need in this directory
* `phpstan-stubs` -- these files should not be runnable from any context, and are not included in the released plugin, so we should reflect that in the dev environment
* `release_directory` -- we shouldn't push this second "copy" of the build plugin to the Docker container

## Testing instructions

* Check out this branch - `git checkout fix/unnecessary-file-exposure-via-docker`
* Ensure that you (re)start your Docker instance: `npm run down; npm run up`
* Open up a bash shell in the context of your Docker instance: `docker run -it --env-file default.env --rm --volumes-from woocommerce_stripe_wordpress --network container:woocommerce_stripe_wordpress bash`
* Change into the plugin directory: `cd /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/`
* Verify that the following subdirectories in the Docker container are all empty:
  - `docker`
  - `node_modules`
  - `phpstan-stubs`
  - `release`

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa (approved):
Nice find and fix! 👍

Mayisha (approved):
Good improvement. Thanks for fixing this 👍

daledupreez:
Merging as e2e tests are related to ACH, which have been failing this week.

Issue #4902, opened 2025-12-30
PHPStan: Better handling for values returned from wc_get_order()


daledupreez:
The `wc_get_order()` function from WooCommerce is explicitly documented to have a return type of `false|WC_Order|WC_Order_Refund`. We have a _lot_ of code that calls `wc_get_order()` and expects the return value to be either a `WC_Order` instance or `false|WC_Order`. PHPStan flags a large number of errors due to this mismatch, as we have downstream functions that expect `WC_Order` values.

A quick fix for this may be to update our callers to check `! $order instanceof WC_Order` instead of `! $order`. We could also consider adding a wrapper function that enforces the return is a `WC_Order` instance, but I feel that would be risky, as it effectively obfuscates what we're doing to a reader unfamiliar with our code.

#### Current function definition in WooCommerce

https://github.com/woocommerce/woocommerce/blob/454fbe99b4df13a0be22e6fe41b854c900722f30/plugins/woocommerce/includes/wc-order-functions.php#L80-L95

PR #4903, opened 2025-12-30
Fix return type from WC_Stripe_Order_Helper::get_instance()

daledupreez:
Towards STRIPE-875
Towards https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4907

This PR fixes the return type for `WC_Stripe_Order_Helper::get_instance()` and changes it from `?self` to just `self`, as we won't return `null` from this function. However, by declaring the return type as nullable, lots of downstream code was being flagged as problematic by PHPStan, as the return value could technically be null. In addition to fixing the return type, I've also updated the PHPStan baseline to remove all the downstream errors stemming from the possibly-null return.

While this reduces the number of errors by a few hundred overall, it does introduce a number of new errors due to more code being examined by PHPStan.

## Testing instructions

* Verify the changed return type is correct
* Inspect the changes to `phpstan-baseline.neon` to confirm that we're only removing `method.nonObject` errors that have the following `message` format: `'#^Cannot call method some_method\(\) on WC_Stripe_Order_Helper\|null\.$#'`
* Note that there are a large number of new errors in the baseline, including many new ones due to the possible return type from `wc_get_order()` -- see https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4902 -- but the number is lower overall
 * Verify that PHPStan analysis is clean
 * Confirm that tests continue to work, especially e2e tests

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@
 * Fix - Ensure that 'Link' and 'Stripe Link' are not translated
 * Fix - Fix situation where Stripe errors were not translated
 * Dev - Ensure multiple subdirectories are not exposed via Docker container
+* Tweak - Fix return type for WC_Stripe_Order_Helper::get_instance()
 
 = 10.2.0 - 2025-12-08 =
 * Dev - Refactor display logic for payment method issue pills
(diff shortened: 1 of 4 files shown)


wjrosa (approved):
The change makes sense to me :shipit:

Mayisha (approved):
Nice and simple. LGTM :shipit:

PR #4904, opened 2025-12-30
Removing the main PRB class

wjrosa:

In this clean-up PR, I am removing the main Payment Request Button backend class (`WC_Stripe_Payment_Request`). It was previously deprecated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4805.

## Testing instructions

Code review should be enough.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-blocks-support.php
```diff
@@ -194,9 +194,6 @@ public function get_payment_method_data() {
 				'showSaveOption'                  => $this->get_show_save_option(),
 				'isAdmin'                         => is_admin(),
 				'shouldShowExpressCheckoutButton' => $this->should_show_express_checkout_button(),
-				'button'                          => [
-					'customLabel' => $this->payment_request_configuration->get_button_label(),
-				],
```
daledupreez, line 199: Should we leave the data structure in place to prevent Javascript runtime errors?
```suggestion
				'button' => [
					'customLabel' => '',
				],
```

AFAICT, the custom label field has been `''` for some time, so it's extremely unlikely that anyone is using it, so I am thinking we probably **don't** want to keep it around.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.3.0 - xxxx-xx-xx =
+* Update - Removes the main Payment Request Buttons backend class (previously deprecated)
```
daledupreez: I think we should use `WC_Stripe_Payment_Request` in the description to make it easier to work out what changed.
```suggestion
* Update - Remove the main Payment Request Buttons backend class, WC_Stripe_Payment_Request, which was deprecated in 10.2.0
```

includes/class-wc-stripe-blocks-support.php
```diff
@@ -37,13 +29,21 @@ final class WC_Stripe_Blocks_Support extends AbstractPaymentMethodType {
 	/**
 	 * Constructor
 	 *
-	 * @param WC_Stripe_Payment_Request  The Stripe Payment Request configuration used for Payment
-	 *                                   Request buttons.
+	 * @param null                               $payment_request_configuration The Stripe Payment Request configuration used for Payment Request buttons (removed).
+	 * @param WC_Stripe_Express_Checkout_Element $express_checkout_configuration The Stripe Express Checkout configuration used for Express Checkout buttons.
```
daledupreez: The changes to the PHPDoc fixed some existing PHPStan errors, so we need to regenerate the baseline to get those stale error entries removed.

daledupreez (changes requested):
The e2e tests are all failing, and I believe it's because there is still an early exit condition in `woocommerce_gateway_stripe_woocommerce_block_support()` where it's checking for `! class_exists( 'WC_Stripe_Payment_Request' )`, and this is preventing Stripe from being available in that context.

daledupreez:
I am also wondering whether it may be worth deferring this removal for 10.4.0 in February, as the deprecation only occurred in 10.3.0, and many developers and/or agencies may not have been taking on work between early December and early January, in addition to merchants likely wanting stability over December.

includes/class-wc-stripe-blocks-support.php
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4904/commits/327421be4960dd0308fa750dbc0de885332041c0

includes/class-wc-stripe-blocks-support.php
wjrosa: Good call. Manually applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4904/commits/cddfd0887d166948a46973975bffbff2b5976b8b (Github blocked me from applying your suggestion directly)

wjrosa:
> The e2e tests are all failing, and I believe it's because there is still an early exit condition in woocommerce_gateway_stripe_woocommerce_block_support() where it's checking for ! class_exists( 'WC_Stripe_Payment_Request' ), and this is preventing Stripe from being available in that context.

Thanks for the heads up, Dale! I removed the check in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4904/commits/a7ffcaeb5aa5d96a2f3df8cd5a99ba89fe653407

> I am also wondering whether it may be worth deferring this removal for 10.4.0 in February, as the deprecation only occurred in 10.3.0, and many developers and/or agencies may not have been taking on work between early December and early January, in addition to merchants likely wanting stability over December.

Makes sense, and I agree. I have updated the changelog entries and assigned the 10.4.0 milestone here.

daledupreez (approved):
Thanks for getting those changes done so quickly! 🚀

malithsen (approved):
Looks good, thanks for the cleanup!

Issue #4907, opened 2025-12-30
PHPStan: Reduce baseline errors


daledupreez:
This is a general-purpose issue to collate general work to reduce the number of baseline errors in `phpstan-baseline.neon`. There are a number of more specific issues tackling subsets of problems, but having this issue will make it easier to link all the work to a single place, especially small PRs that make things slightly better over time.

PR #4908, opened 2025-12-30
Fix PHPDoc for WC_Stripe_Payment_Gateway->get_intent_from_order()

daledupreez:
Towards STRIPE-875
Towards https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4907

This PR fixes the PHPDoc for `WC_Stripe_Payment_Gateway->get_intent_from_order()` to specify a return type of `object|false` instead of `obect|bool`, as PHPStan was rightfully complaining about `obect` not being a known class. 😁 However, some of the relevant errors were also seeing that `bool` could have a return value of `true`, which is not correct per the code. I also updated the PHPDoc for `WC_Stripe_Payment_Gateway->get_intent()` to return `object|false` instead of `object|bool` to reflect the same reality.

I then updated the PHPStan baseline as well, which reduced our error count by ~20 or so.

## Testing instructions

Code review on this one should be sufficient.

Also confirm that the changes in the baseline errors make sense:
 - We are removing a number of errors stemming from the invalid `object` type, and some errors where the type was being detected as `obect|true` (likely due to `! $value` checks passing)
 - We are adding or bumping some error counts for `property.notFound` errors, as we are now accessing `$object->foo` values

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@
 * Fix - Ensure that 'Link' and 'Stripe Link' are not translated
 * Fix - Fix situation where Stripe errors were not translated
 * Dev - Ensure multiple subdirectories are not exposed via Docker container
+* Tweak - Update PHPDoc for WC_Stripe_Payment_Gateway->get_intent_from_order()
 
 = 10.2.0 - 2025-12-08 =
 * Dev - Refactor display logic for payment method issue pills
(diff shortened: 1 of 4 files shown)


wjrosa (approved):
Looks good to me! 👍

Mayisha (approved):
Looks good to me :shipit: 

Sidenote: Should we track the new `property.notFound` in a separate issue and fix it?

daledupreez:
> Sidenote: Should we track the new property.notFound in a separate issue and fix it?

For now, we have a general issue for reducing all PHPStan errors in https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4907. I haven't created a specific issue for `property.notFound` issues, as that is a broad category of issue, largely stemming from the fact that we don't have type definitions or shapes for the objects returned from the Stripe APIs. I think we have some smaller, more impactful issues we can tackle first, and I have created issues for those.

PR #4909, opened 2025-12-30
Emptying the main legacy checkout backend class

wjrosa:

In this clean-up PR, I am emptying the main legacy checkout backend class (`WC_Gateway_Stripe`), and temporarily making it extends the UPE class for backwards compatibility. It was previously deprecated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4718.

## Testing instructions

Code review. Perform some basic smoke-testing to make sure the current (new) checkout did not break.
Check if the tests are still passing.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3465,4 +3653,495 @@ public function is_payment_request_enabled() {
...
+
+	/**
+	 * Ensures validated field values.
+	 *
+	 * @param string $field_key the form field key.
+	 * @param string $field_value the form field value.
+	 *
+	 * @return string validated field value.
+	 */
+	private function validate_field( string $field_key, string $field_value ): string {
```
Mayisha: We need to update the signature of this method. Some tests in  `WC_REST_Stripe_Settings_Controller_*` test class are failing because of this.

Mayisha (changes requested):
In this branch, I am getting an empty response for the GET settings request. This results in empty payment method list and all disabled/unchecked settings page for me. This is possibly related to the same issue as the test failures.

Nit: You may want to change the mention of `WC_Gateway_Stripe` [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/update/removing-the-main-legacy-checkout-class/assets/js/stripe.js#L963) as well.

includes/class-wc-gateway-stripe.php
```diff
@@ -1,1234 +0,0 @@
...
-}
-
-/**
- * WC_Gateway_Stripe class.
- *
- * @extends WC_Payment_Gateway
- *
- * @deprecated Deprecated in favor of WC_Stripe_UPE_Payment_Gateway. To be removed in a future version.
- */
-class WC_Gateway_Stripe extends WC_Stripe_Payment_Gateway {
```
daledupreez, line 16: I don't think we should fully remove this class in the short term. I agree that we should move all the functionality into the UPE gateway class, but I am concerned that there is a good deal of code out there that may be looking for or interacting with this specific class. Rather than completely removing it in this release, might it be worth keeping this class and making it an empty subclass of the UPE gateway class? That would keep all the logic and constants in the new class, but allow all older usages to continue for a release or two.

wjrosa: Personally, I don't see a reason to keep the class for an additional while, but I did make the changes you requested so we can feel safer about this change: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4909/commits/95afcd9053120563cc9f4467408e2c676d0c16db.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Thanks for the heads up, Mayisha! I have fixed the issue in my latest commits 👍

includes/abstracts/abstract-wc-stripe-payment-gateway-voucher.php
```diff
@@ -11,7 +11,7 @@
  * Abstract class that will be inherited by voucher payment methods.
  * Used by Boleto and OXXO
  *
- * @extends WC_Gateway_Stripe
+ * @extends WC_Stripe_Payment_Gateway
```
daledupreez: Should this not be `WC_Stripe_UPE_Payment_Gateway`?
```suggestion
 * @extends WC_Stripe_UPE_Payment_Gateway
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.3.0 - xxxx-xx-xx =
+* Update - Update the legacy checkout experience gateway class, removing all the logic and temporarily extending the UPE class for backwards compatibility
```
daledupreez: I think we should be specific about the class involved, and try to make the description shorter.
```suggestion
* Update - Move all logic from WC_Gateway_Stripe to WC_Stripe_UPE_Payment_Gateway as part of deprecation
```

includes/class-wc-stripe.php
```diff
@@ -155,7 +155,6 @@ public function init() {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/payment-tokens/class-wc-stripe-klarna-payment-token.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-apple-pay-registration.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-status.php';
-		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-gateway-stripe.php';
```
daledupreez, line 160: I think we should keep loading the existing class. If anyone has logic that declares this as a type, or tries to perform any type checks or `class_exists()` calls to identify the presence of the Stripe plugin, we should ensure the class is still available.

includes/class-wc-gateway-stripe.php
daledupreez: My main concern is that there may be other plugin or snippet code that's referencing or checking for the old class. They _shouldn't_ be doing that, but I'd rather keep the class around and then remove in the next release, just to be safer and give folks more time to switch to the new class.

daledupreez (commented):
Thanks for the additional cleanup. It looks like the PHPStan baseline should be regenerated after merging from develop, and I think we need to keep the class available in memory for one more release.

includes/class-wc-stripe.php
wjrosa: Alright. Reverted the removal in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4909/commits/d317658f0af3f38975cce1899276a1ed4cc6471f

daledupreez (approved):
Thanks for pushing this through! The changes are looking good to me and the tests are clean. 🚀

Mayisha (approved):
Looks good to me. Both the tests and the settings page are working correctly now 👍

PR #4922, opened 2026-01-09
Fix unsupported payment methods on checkout page when optimized checkout is enabled

Mayisha:
Fixes STRIPE-863

We have received reports that, in optimized checkout, unsupported payment methods are sometimes displayed. This PR fixes the issue by leveraging the `excludedPaymentMethodTypes` param in Stripe payment element.

Stripe payment element supports `excludedPaymentMethodTypes` param which is used to exclude any payment method from the element. Before, only Amazon Pay was sent in this list so that we render Amazon Pay only as an express method and not within optimized checkout. Now we are generating a list on the server side which includes Amazon Pay as well as any other unsupported payment methods. From the PMC response we check if there are any payment methods `on` for this PMC which are not supported in the plugin and send this list to exclude from the payment element.

## Testing instructions
- Enable optimized checkout in the Stripe settings page.
- To reproduce the issue, I used a Japanese Stripe account and store currency JPY. You may create/use an account from any country that has a payment method unsupported by this plugin.
- Go to the `/settings/payment_methods` page in the Stripe dashboard of your selected account and create a new payment method configuration. Note the PMC ID in the URL.
- Enable one or more payment methods unsupported by the plugin. In my tests, I enabled Kinbini.
- Now connect to this Stripe account in your store.
- Add the following filter to use the PMC above
```
add_filter( 'wc_stripe_preselect_payment_method_configuration', function ( $value, $is_test_mode ) {
	return <your_pmc_id>;
}, 10, 2 );
```
- As a shopper, add a product to your cart and go to the checkout page. (Make sure the products in the cart are over the minimum amount for the currency)
- In `develop` branch, notice that the unsupported payment method (`Konbini` in my case) is displayed as a payment method (non functioning through)

Screenshot 2026-01-12 at 11 06 35 PM

- In this branch `fix/unsupported-payment-methods-in-ocs`, verify that the unsupported payment method is absent.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -376,6 +376,30 @@ public static function get_upe_available_payment_method_ids() {
 		return $available_payment_method_ids;
 	}
 
+	/**
+	 * Get the enabled payment method IDs in the PMC that are not supported in the plugin.
+	 *
+	 * @return array
+	 */
+	public static function get_unsupported_enabled_payment_method_ids_in_pmc() {
```
malithsen: Should we bail if PMC sync is not enabled? Similar to what other functions are doing in this file

```suggestion
	public static function get_unsupported_enabled_payment_method_ids_in_pmc() {
	    // Bail if the payment method configurations API is not enabled.
            if ( ! self::is_enabled() ) {
	        return [];
            }

```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -629,6 +630,27 @@ public function javascript_params() {
...
+	/**
+	 * Returns the list of payment methods that should be excluded from the Payment Element in optimized checkout.
+	 *
+	 * @return array
+	 */
+	private function get_excluded_payment_method_types() {
+		$unsupported_methods = WC_Stripe_Payment_Method_Configurations::get_unsupported_enabled_payment_method_ids_in_pmc();
+
+		// 'Link', 'Apple Pay', and 'Google Pay' are not supported in the 'excludedPaymentMethodTypes' parameter, so we always allow them.
+		$allowed_methods = [ WC_Stripe_Payment_Methods::LINK, WC_Stripe_Payment_Methods::APPLE_PAY, WC_Stripe_Payment_Methods::GOOGLE_PAY, 'cartes_bancaires' ];
```
malithsen: Just double-checking as `cartes_bancaires` is in the list but absent from the comment above. Should it be in this list? If so, should we define `cartes_bancaires` as a constant in `WC_Stripe_Payment_Methods` similar to other methods?

client/stripe-utils/utils.js
```diff
@@ -534,6 +535,22 @@ export const getDefaultValues = () => {
...
+ * Fallback to exclude Amazon Pay, which should be only shown via Express Checkout, and not within Optimized Checkout.
+ *
+ * @return {Array<string>} Array of payment method types to exclude.
+ */
+export const getExcludedPaymentMethodTypes = () => {
+	const stripeServerData = getStripeServerData();
+	return (
+		stripeServerData?.excludedPaymentMethodTypes || [
+			PAYMENT_METHOD_AMAZON_PAY,
+		]
```
malithsen: Is this fallback necessary as we include Amazon Pay in the backend?

Also, this check likely won't work as intended if `excludedPaymentMethodTypes` is an empty array, as JS considers `[]` as truthy.

`[] || ['amazon_pay']` evaluates to `[]`

malithsen (approved):
Looks good! I left some suggestions in the comments, but I don't believe they are blocking.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: Alternatively, we should document why `cartes_bancaires` is special and included in the list.

changelog.txt
```diff
@@ -33,6 +33,7 @@
 * Fix - Ensure express checkout is enabled in current context before registering the express checkout script
 * Fix - Prevent warnings for Amazon Pay in Express Checkout block
 * Dev - Ensure PHPStan runs when pushing changes
+* Fix - Resolve and issue where unsupported payment methods gets displayed on checkout page when optimized checkout is enabled
```
daledupreez: Two comments on this:
* The `and` should be `an`
* This is really long -- I think we can simplify it
```suggestion
* Fix - Prevent Optimized Checkout from showing unsupported payment methods
```

client/stripe-utils/utils.js
```diff
@@ -534,6 +535,22 @@ export const getDefaultValues = () => {
...
+	);
```
daledupreez: It may be worth addressing @malithsen's [comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4922/files#r2687791968)  on the existing logic by adding some early returns.
```suggestion
	if ( ! Array.isArray( stripeServerData?.excludedPaymentMethodTypes ) ) {
		return [ PAYMENT_METHOD_AMAZON_PAY ];
	}
	if ( stripeServerData.excludedPaymentMethodTypes.includes( PAYMENT_METHOD_AMAZON_PAY ) ) {
		return stripeServerData.excludedPaymentMethodTypes;
	}
	return [ ...stripeServerData.excludedPaymentMethodTypes, PAYMENT_METHOD_AMAZON_PAY ];
```

changelog.txt
```diff
@@ -7,7 +7,7 @@
 * Dev - Ensure PHPStan runs when pushing changes
 * Dev - Add PHPStan stub for WC_Subscription class
 * Dev - Remove the deprecated WC_Stripe_Apple_Pay class
-
+* Fix - Resolve and issue where unsupported payment methods gets displayed on checkout page when optimized checkout is enabled
```
daledupreez: Two comments on this:
* The `and` should be `an`
* This is really long -- I think we can simplify it
```suggestion
* Fix - Prevent Optimized Checkout from showing unsupported payment methods
```

tests/phpunit/WC_Stripe_Payment_Method_Configurations_Test.php
```diff
@@ -870,4 +870,161 @@ public function test_get_configuration_id(
...
+		WC_Stripe_Payment_Method_Configurations::clear_payment_method_configuration_cache();
+
+		$result = WC_Stripe_Payment_Method_Configurations::get_unsupported_enabled_payment_method_ids_in_pmc();
+
+		$this->assertEmpty( $result );
+
+		// Reset settings
+		$settings['pmc_enabled'] = 'yes';
+		WC_Stripe_Helper::update_main_stripe_settings( $settings );
+		WC_Stripe_Payment_Method_Configurations::clear_payment_method_configuration_cache();
```
daledupreez: Minor note for here and elsewhere: is it worth cleaning up before we assert anything? A failed assertion could end up impacting other tests if it prevents us from cleaning up semi-global settings like this one.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -657,6 +658,27 @@ public function javascript_params() {
 		return array_merge( $stripe_params, WC_Stripe_Helper::get_localized_messages() );
 	}
 
+	/**
+	 * Returns the list of payment methods that should be excluded from the Payment Element in optimized checkout.
+	 *
+	 * @return array
+	 */
+	private function get_excluded_payment_method_types() {
```
daledupreez: Nit: Can we add an explicit return type and be more specific about the array contents in the PHPDoc?
```suggestion
	 * @return string[]
	 */
	private function get_excluded_payment_method_types(): array {
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: Also, it may be worth using a multi-line array for these values so it's easier to track, see, and comment on changes to the entries in the array.
```suggestion
		$allowed_methods = [
			WC_Stripe_Payment_Methods::LINK,
			WC_Stripe_Payment_Methods::APPLE_PAY,
			WC_Stripe_Payment_Methods::GOOGLE_PAY,
			// Explanation for cartes_bancaires is in the array
			'cartes_bancaires',
		];
```

includes/class-wc-stripe-payment-method-configurations.php
daledupreez: As for the private method, can we make the PHPDoc more specific and add an explicit return type?
```suggestion
	 * @return string[]
	 */
	public static function get_unsupported_enabled_payment_method_ids_in_pmc(): array {
```

tests/phpunit/WC_Stripe_Payment_Method_Configurations_Test.php
```diff
@@ -870,4 +870,161 @@ public function test_get_configuration_id(
...
+	 */
+	public function test_get_unsupported_enabled_payment_method_ids_in_pmc_when_pmc_disabled() {
+		$settings                = WC_Stripe_Helper::get_stripe_settings();
+		$settings['pmc_enabled'] = 'no';
+		WC_Stripe_Helper::update_main_stripe_settings( $settings );
+		WC_Stripe_Payment_Method_Configurations::clear_payment_method_configuration_cache();
+
+		$result = WC_Stripe_Payment_Method_Configurations::get_unsupported_enabled_payment_method_ids_in_pmc();
+
+		$this->assertEmpty( $result );
```
daledupreez: As a separate comment on not having an explicit `array` return type and expectations, invalid return values of `0`, `''`, `null`, or `false` would all pass this unit test. If we add the explicit `array` return type, those are no longer possibilities.

tests/phpunit/WC_Stripe_Payment_Method_Configurations_Test.php
```diff
@@ -870,4 +870,161 @@ public function test_get_configuration_id(
 			$this->assertEquals( $expected_configuration_id, $configuration_id );
 		}
 	}
+
+		/**
```
daledupreez: Nit: Looks like a minor whitespace issue.
```suggestion
	/**
```

tests/phpunit/WC_Stripe_Payment_Method_Configurations_Test.php
```diff
@@ -870,4 +870,161 @@ public function test_get_configuration_id(
...
+		$property->setValue( null, $mock_api );
+
+		WC_Stripe_Payment_Method_Configurations::clear_payment_method_configuration_cache();
+		$result = WC_Stripe_Payment_Method_Configurations::get_unsupported_enabled_payment_method_ids_in_pmc();
+
+		$this->assertEmpty( $result );
+
+		// Reset API instance
+		$property->setValue( null, null );
+		WC_Stripe_Payment_Method_Configurations::clear_payment_method_configuration_cache();
```
daledupreez: Non-blocking observation: the basic shape of all the unit tests is the same: initialise settings, clear PMC cache, mock API and returned PMC, call API, match results, clean up. That could be written as a data provider and test running to avoid the setup duplication across the test methods.

includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -376,6 +376,30 @@ public static function get_upe_available_payment_method_ids() {
...
+		$unsupported_payment_method_ids        = [];
+		$merchant_payment_method_configuration = self::get_primary_configuration();
+
+		if ( $merchant_payment_method_configuration ) {
+			foreach ( $merchant_payment_method_configuration as $payment_method_id => $payment_method ) {
```
daledupreez: Would casting `$merchant_payment_method_configuration` to an array address the PHPStan iterable error/warning?
```suggestion
			foreach ( (array) $merchant_payment_method_configuration as $payment_method_id => $payment_method ) {
```

daledupreez (commented):
This is working as described, but I have a good amount of feedback on various implementation details.

includes/constants/class-wc-stripe-payment-methods.php
```diff
@@ -102,4 +102,90 @@ class WC_Stripe_Payment_Methods {
...
+		'pix',
+		'promptpay',
+		'revolut_pay',
+		'sunbit',
+		'netbanking',
+		'id_bank_transfer',
+		'demo_pay',
+		'shop_pay',
+		'zip',
+	];
```
Mayisha, line 122: I did not find this list directly in any documentation. This was extracted from this console error-

Image

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -660,6 +661,36 @@ public function javascript_params() {
...
+			$unsupported_methods,
+			function ( $method ) use ( $excludable_methods ) {
+				return in_array( $method, $excludable_methods, true );
+			}
+		);
+
+		// Always exclude amazon_pay (shown via Express Checkout, not in Payment Element)
+		if ( ! in_array( 'amazon_pay', $excluded_methods, true ) ) {
+			$excluded_methods[] = 'amazon_pay';
+		}
```
daledupreez: I think we need to do two things differently here:
1. I think we should maintain a list of known exceptions that are not supported in `excludedPaymentMethodTypes`, as that list is going to change less often than the full set of possible Stripe payment methods, which will continually be changing.
2. Given that new values may be added to the list at any time, I think we need to add a filter to make it possible to add extra values to the list so that merchants can keep this working without needing a patched version of the plugin.
```suggestion
		$non_excludable_methods = WC_Stripe_Payment_Methods::NON_EXCLUDABLE_PAYMENT_METHOD_TYPES;
		
		// TODO: Add filter docs
		$custom_non_excludable_methods = apply_filters( 'wc_stripe_ocs_non_excludable_payment_methods', [] );
		// Make sure we have a non-empty array
		if ( is_array( $custom_non_excludable_methods ) && [] !== $custom_non_excludable_methods ) {
			// Make sure we have string values
			$custom_non_excludable_methods = array_filter( $custom_non_excludable_methods, 'is_string' );
			$non_excludable_methods = array_unique( array_merge( $custom_non_excludable_methods, $non_excludable_methods ) );
		}

		// Some payment methods in the PMC are not supported in the 'excludedPaymentMethodTypes' parameter
		// of the Payment Element (e.g. link, apple_pay, google_pay, cartes_bancaires etc.).
		$excluded_methods = array_filter(
			$unsupported_methods,
			function ( $method ) use ( $non_excludable_methods ) {
				return ! in_array( $method, $non_excludable_methods, true );
			}
		);

		// Always exclude amazon_pay (shown via Express Checkout, not in Payment Element)
		if ( ! in_array( 'amazon_pay', $excluded_methods, true ) ) {
			$excluded_methods[] = 'amazon_pay';
		}
```

daledupreez: Nit: We should be using the constant instead of the `'amazon_pay'` literal.
```suggestion
		// Always exclude Amazon Pay, as it is shown via Express Checkout and not in the standard Payment Element.
		if ( ! in_array( WC_Stripe_Payment_Methods::AMAZON_PAY, $excluded_methods, true ) ) {
			$excluded_methods[] = WC_Stripe_Payment_Methods::AMAZON_PAY;
		}
```

client/stripe-utils/utils.js
```diff
@@ -534,6 +535,24 @@ export const getDefaultValues = () => {
...
+ *
+ * @return {Array<string>} Array of payment method types to exclude.
+ */
+export const getExcludedPaymentMethodTypes = () => {
+	const stripeServerData = getStripeServerData();
+	const excludedTypes = stripeServerData?.excludedPaymentMethodTypes;
+
+	if ( ! Array.isArray( excludedTypes ) ) {
+		return [ PAYMENT_METHOD_AMAZON_PAY ];
+	}
```
daledupreez: Just to be extra defensive, might it be worth checking for an empty array here as well?
```suggestion
	if ( ! Array.isArray( excludedTypes ) || excludedTypes.length === 0 ) {
		return [ PAYMENT_METHOD_AMAZON_PAY ];
	}
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: We do not have any official documentation listing the supported or unsupported values for this field (refer to my [other comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4922/changes#r2768196935)). I initiated a question in p1770298253207149-slack-C07NTGXP12N

Mayisha: @daledupreez as there is no defined list of non_excludable values, I have kept the list of excludable values and added a filter `wc_stripe_ocs_excludable_payment_methods` to extend the list.

Alternatively, I can use the so far discovered list as following and add the `wc_stripe_ocs_non_excludable_payment_methods` filter to be extended by stores if they discover any new unsupported value.

```
$non_excludable_methods = [ 'apple_pay', 'google_pay', 'link', 'cartes_bancaires' ];
```

daledupreez: I think we can infer the currently unsupported types by comparing the accepted arguments to `excluded_payment_method_types` ([docs](https://docs.stripe.com/api/payment_intents/update#update_payment_intent-excluded_payment_method_types)) against the possible elements that are present in the Payment Method Configuration object ([docs](https://docs.stripe.com/api/payment_method_configurations/object)). The full data from both lists are below, but the specific case we care about is when we get a payment method in a PMC, and that payment method can't be specified as an excluded payment method. That leaves us with the following list of non-excludable payment methods as things stand:
* `apple_pay`
* `cartes_bancaires`
* `google_pay`
* `jcb`
* `link`

I think we can use that as our starting point, which adds `jcb` to your proposed list.

| Payment method in PMC | Excludable type | 
|--------|------------|
| acss_debit | acss_debit |
| affirm | affirm |
| afterpay_clearpay | afterpay_clearpay |
| alipay | alipay |
| alma | alma |
| amazon_pay | amazon_pay |
| apple_pay | _**[NOT PRESENT]**_ |
| au_becs_debit | au_becs_debit |
| bacs_debit | bacs_debit |
| bancontact | bancontact |
| billie | billie |
| blik | blik |
| boleto | boleto |
| card | card |
| _**[NOT PRESENT]**_ | card_present |
| cartes_bancaires | _**[NOT PRESENT]**_ |
| cashapp | cashapp |
| crypto | crypto |
| customer_balance | customer_balance |
| eps | eps |
| fpx | fpx |
| giropay | giropay |
| google_pay | _**[NOT PRESENT]**_ |
| grabpay | grabpay |
| ideal | ideal |
| _**[NOT PRESENT]**_ | interac_present |
| jcb | _**[NOT PRESENT]**_ |
| kakao_pay | kakao_pay |
| klarna | klarna |
| konbini | konbini |
| kr_card | kr_card |
| link | _**[NOT PRESENT]**_ |
| mb_way | mb_way |
| mobilepay | mobilepay |
| multibanco | multibanco |
| naver_pay | naver_pay |
| nz_bank_account | nz_bank_account |
| oxxo | oxxo |
| p24 | p24 |
| pay_by_bank | pay_by_bank |
| payco | payco |
| paynow | paynow |
| paypal | paypal |
| paypay | paypay |
| payto | payto |
| pix | pix |
| promptpay | promptpay |
| revolut_pay | revolut_pay |
| samsung_pay | samsung_pay |
| satispay | satispay |
| sepa_debit | sepa_debit |
| sofort | sofort |
| swish | swish |
| twint | twint |
| us_bank_account | us_bank_account |
| wechat_pay | wechat_pay |
| zip | zip |

Mayisha: Thanks for looking into this in depth. I have updated the logic in [999ea04](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4922/commits/999ea0490d4efa7feb62493809f88e596a95b8ae)

includes/constants/class-wc-stripe-payment-methods.php
```diff
@@ -104,88 +104,16 @@ class WC_Stripe_Payment_Methods {
 	];
 
 	/**
-	 * Payment method types that can be excluded via Stripe's 'excludedPaymentMethodTypes' parameter.
-	 * Note: 'Link', 'Apple Pay', and 'Google Pay' are not supported in this parameter, so they are not included.
+	 * Payment method types that can not be excluded via Stripe's 'excludedPaymentMethodTypes' parameter.
+	 * These values are not supported in the 'excludedPaymentMethodTypes' parameter and causes an error when trying to render the Payment Element excluding them.
```
daledupreez, line 108: It may be worth documenting how we built the list as well. Either a link to my comment, or to the two sources should be good enough.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -671,21 +671,33 @@ public function javascript_params() {
 	private function get_excluded_payment_method_types(): array {
 		$unsupported_methods = WC_Stripe_Payment_Method_Configurations::get_unsupported_enabled_payment_method_ids_in_pmc();
 
-		$excludable_methods = WC_Stripe_Payment_Methods::EXCLUDABLE_PAYMENT_METHOD_TYPES;
+		$non_excludable_methods = WC_Stripe_Payment_Methods::NON_EXCLUDABLE_PAYMENT_METHOD_TYPES;
+
+		/**
+		 * Filters the list of payment methods that can not be excluded from the Payment Element in optimized checkout.
```
daledupreez: It may be worth documenting that the filter is only adding to the base list, not filtering the original one.
```suggestion
		 * Filters the list of additional payment methods that can not be excluded from the Payment Element in optimized checkout.
		 * This list will be added to the base list in {@see WC_Stripe_Payment_Methods::NON_EXCLUDABLE_PAYMENT_METHOD_TYPES}.
```

tests/phpunit/WC_Stripe_Payment_Method_Configurations_Test.php
```diff
@@ -870,4 +870,153 @@ public function test_get_configuration_id(
...
+				'expect_empty'          => true,
+			],
+			'No configuration'         => [
+				'settings'              => array_merge( $settings_base, [ 'pmc_enabled' => 'yes' ] ),
+				'mock_api_response'     => (object) [ 'data' => [] ],
+				'expected_contains'     => [],
+				'expected_not_contains' => [],
+				'expect_empty'          => true,
+			],
+			'With unsupported methods' => [
```
daledupreez, line 958: It may be worth adding a test case for the `wc_stripe_ocs_non_excludable_payment_methods` filter to ensure that when the filter adds any payment methods, those payment methods are handled correctly.

And maybe another test case to ensure that we handle duplicates introduced via the filter correctly as well.

daledupreez (approved):
Thanks for sticking with this PR! Everything is looking good to me, including some tests for the filter, which I tested by adding on of the extra payment methods I enabled in Stripe (mostly so I could visually confirm it was getting into the exclude list in the client).

PR #4928, opened 2026-01-13
Trim customer_name before sending to Stripe

### LINKED ISSUE #4927: If the WooCommerce checkout hides the "last name" field, the customer name goes to Stripe with an extra space
### Description

As mentioned in the issue title: if the checkout does not request the last name, the customer name will reach Stripe with an extra space:

Image

Image

### Steps to reproduce

1. Remove the "last name" field from the checkout, using the WooCommerce-provided filters
2. Place a Stripe order
3. Check `customer_name` on the `payment_intent.created` logs on the Stripe dashboard

### Expected behavior

Full name should be trimmed

webdados:
Fixes #4927 

Trim `customer_name` before sending to Stripe, avoiding an unnecessary space at the end of the name if the shop owner did not request the customer's last name on the checkout.

## Testing instructions

As mentioned on #4927 

### Changelog entry

Trim customer_name before sending to Stripe

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.4.0 - xxxx-xx-xx =
+* Update - Trim the `customer_name` field value before sending it to Stripe when creating a payment intent
 * Update - Move all logic from WC_Gateway_Stripe to WC_Stripe_UPE_Payment_Gateway as part of deprecation
 * Update - Remove the main Payment Request Buttons backend class, WC_Stripe_Payment_Request, which was deprecated in 10.2.0
 * Dev - Replace deprecated logger method calls with severity specific methods
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 10.4.0 - xxxx-xx-xx =
-* Update - Trim the `customer_name` field value before sending it to Stripe when creating a payment intent
+* Update - Ensure the `customer_name` metadata sent to Stripe does not have leading or trailing spaces
 * Update - Move all logic from WC_Gateway_Stripe to WC_Stripe_UPE_Payment_Gateway as part of deprecation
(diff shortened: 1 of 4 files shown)


wjrosa:
Hi @webdados ! Thanks for the contribution. Is this ready for review?

webdados:
Yes. Review it. It's just to "trim" calls.

wjrosa:
Alright. Do you have any example names to share that motivated you to open this PR? Asking because `sanitize_text_field` already performs a `trim` (see [here](https://developer.wordpress.org/reference/functions/_sanitize_text_fields/)). But maybe that does not cover some names you saw.

webdados:
See https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4927

image

webdados:
`sanitize_text_field` is being applied to first name and last name, but they're being concatenated with a space between them.

daledupreez:
I also checked other situations where we concatenated first and last names in `WC_Stripe_Customer` and I noticed that those locations already `trim()` the concatenated values, so this change is aligned with those.

daledupreez (approved):
Thanks for the fix, @webdados! This looks good to me!

wjrosa (approved):
Thanks for the explanation! Looks good to me as well 👍 . I will add changelog entries here and merge it

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.4.0 - xxxx-xx-xx =
+* Update - Trim the `customer_name` field value before sending it to Stripe when creating a payment intent
```
daledupreez: @wjrosa, I think it may be better to describe the outcome of the change rather than the technical fix.
```suggestion
* Update - Ensure the `customer_name` metadata sent to Stripe does not have leading or trailing spaces
```

webdados: Agree

PR #4929, opened 2026-01-13
Removing the deferred intent flag and related methods

wjrosa:

In this cleaning-up PR, I am removing the `wc-stripe-is-deferred-intent` flag and all the related flows, since its value is always `1` for a while (since we removed the legacy checkout experience).

The goal is to remove those later.

## Testing instructions

Code review. Perform basic smoke testing in both the shortcode and block checkout as a shopper to confirm nothing breaks. Testing for subscriptions and changes of payment method is appreciated.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
Adding @Mayisha, as well, she must have most of the context to confirm it is OK to remove the flag now.

Mayisha:
> Adding @Mayisha, as well, she must have most of the context to confirm it is OK to remove the flag now.

On top of my head, this looks safe to me as we have this flag set by default for many versions and this is the code path being executed for customized checkout for a while. 
I remembered one case where another plugin used the `/checkout` endpoint without sending the flag ( peNR48-19i-p2 ). It might be worth double checking that removing this flag is safe for this kind of usage.

tests/phpunit/WC_Stripe_Subscription_Initial_Test.php
```diff
@@ -82,6 +83,8 @@ public function tear_down() {
 	 * 5. More assertions are made.
 	 */
 	public function test_initial_intent_parameters() {
+		$this->markTestSkipped( 'Not working at the moment - needs fixing.' );
```
malithsen: Is this intentional? Should we fix this before merge?

tests/phpunit/WC_Stripe_Subscription_Initial_Test.php
```diff
@@ -151,7 +154,7 @@ public function test_initial_intent_parameters() {
 				'filename' => null,
 			];
 
-			// Respond with a successfull intent for confirmations.
+			// Respond with a successfully intent for confirmations.
```
malithsen: nit: typo
```suggestion
			// Respond with a successful intent for confirmations.
```

tests/phpunit/WC_Stripe_Subscription_Initial_Test.php
wjrosa: Yes. That's my plan. I spent some time trying to fix this without success, so I decided to mark it skipped for now and fix it after it is approved.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -947,292 +945,7 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
...
-			return $this->process_payment_with_deferred_intent( $order_id );
-		}
-
-		if ( $this->maybe_change_subscription_payment_method( $order_id ) ) {
-			return $this->process_change_subscription_payment_method( $order_id );
-		}
-
-		if ( $this->is_using_saved_payment_method() ) {
-			return $this->process_payment_with_saved_payment_method( $order_id );
-		}
```
daledupreez, line 970: I am particularly concerned about some of the missing logic in this section. How would a shopper change their subscription payment method in this new code?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1543,6 +1256,8 @@ private function handle_process_payment_error( WC_Stripe_Exception $e, $order )
 	 *
 	 * @param int $order_id   The order ID being processed.
 	 * @param bool $can_retry Should we retry on fail.
+	 *
+	 * @deprecated 10.4.0 This method is deprecated and will be removed in a future release.
 	 */
 	public function process_payment_with_saved_payment_method( $order_id, $can_retry = true ) {
 		try {
```
daledupreez, line 1273: I think we should add calls to `wc_deprecated_function()` for these methods that we're deprecating so that is done and we can be safer in removing the logic later.

wjrosa: I decided to move the deprecation part of this PR to another PR: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4998

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: The handling for switching subscription payment methods happens inside `process_payment_with_deferred_intent` now:
```php
		if ( $this->is_changing_payment_method_for_subscription() ) {
			return $this->process_change_subscription_payment_with_deferred_intent( $order_id );
		}
```

Just tested this specific flow with this branch and it is still working.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -955,292 +953,7 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
...
 
-		// Flag for using a deferred intent. To be removed.
-		// https://github.com/woocommerce/woocommerce-gateway-stripe/issues/3868
-		if ( ! empty( $_POST['wc-stripe-is-deferred-intent'] ) ) {
-			return $this->process_payment_with_deferred_intent( $order_id );
-		}
-
-		if ( $this->maybe_change_subscription_payment_method( $order_id ) ) {
-			return $this->process_change_subscription_payment_method( $order_id );
-		}
```
daledupreez, line 966: I just want to note that I was concerned about us "losing" this logic, but I went ahead and checked, and the downstream `process_payment_with_payment_method()` code that is called from `process_payment_with_deferred_intent()` also includes logic to check for subscription payment method changes. I also confirmed that we currently include the `wc-stripe-is-deferred-intent` form field in the change subscription page, so we do go down that path.

One separate thing I also noticed is that the other code branch in `process_payment_with_deferred_intent()` calls `process_payment_with_confirmation_token()`, which does _not_ handle subscription payment method changes. However, I think Amazon Pay is the only payment method that uses confirmation tokens, and we don't make it available as a replacement payment method due to it being available only for express checkout.

daledupreez (commented):
This is looking good to me, but I think we should take a good look at that failing/skipped test to make sure that it has value. I see that it's storing a `source` against the subscription, but it's not clear to me whether we're intentionally checking for that case or if the test is just old. 😁

tests/phpunit/Compat/WC_Stripe_Subscription_Initial_Test.php
```diff
@@ -0,0 +1,141 @@
...
+				->method( 'create_and_confirm_payment_intent' )
+				->willReturn(
+					(object) [
+						'id'                => 'pi_123abc',
+						'object'            => 'payment_intent',
+						'payment_method'    => (object) [
+							WC_Stripe_Payment_Methods::CARD => (object) [
+								'brand'    => 'visa',
+								'exp_month' => 12,
+								'exp_year'  => 2024,
```
daledupreez: Non-blocking nit: It would be good to set this year to be in the future.
```suggestion
								'exp_year'  => 2034,
```

tests/phpunit/Compat/WC_Stripe_Subscription_Initial_Test.php
```diff
@@ -0,0 +1,141 @@
...
+			return [
+				'headers'  => [],
+				'body'     => json_encode(
+					(object) [
+						'id'              => 'pm_test_123',
+						'object'          => 'payment_method',
+						WC_Stripe_Payment_Methods::CARD => (object) [
+							'brand'    => 'visa',
+							'exp_month' => 12,
+							'exp_year'  => 2024,
```
daledupreez: Nit: expiry year
```suggestion
							'exp_year'  => 2034,
```

tests/phpunit/Compat/WC_Stripe_Subscription_Renewal_Test.php
```diff
@@ -214,7 +214,7 @@ public function test_renewal_successful() {
 			return [
 				'headers'  => [],
 				// Too bad we aren't dynamically setting things 'cus_123abc' when using this file.
-				'body'     => file_get_contents( __DIR__ . '/dummy-data/subscription_renewal_response_success.json' ),
+				'body'     => file_get_contents( __DIR__ . '/DummyData/subscription_renewal_response_success.json' ),
```
daledupreez: Not blocking: I am not sure I like the change in case -- it feels like the new case implies the folder may have PHP code in it.

readme.txt
```diff
@@ -148,6 +148,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.5.0 - xxxx-xx-xx =
+* Update - Remove the main deferred intent flag (`wc-stripe-is-deferred-intent`) and all the code not using it, as it is the default behavior now
```
Mayisha: Nit: Can we simplify the statement? Following is just a suggestion, any other simplified form is also welcome.
 
```suggestion
* Update - Remove the deferred intent flag (`wc-stripe-is-deferred-intent`) and the non-deferred intent code path
```

tests/phpunit/Compat/WC_Stripe_Subscription_Initial_Test.php
```diff
@@ -0,0 +1,141 @@
...
+ * The responses from HTTP requests are mocked using the WP filter `pre_http_request`.
+ *
+ * There are a few methods that need to be mocked in the class WC_Stripe_UPE_Payment_Gateway, which is
+ * why that class is mocked even though the method under test is part of that class.
+ *
+ * @package WooCommerce/Stripe/WC_Stripe_Subscription_Initial
+ *
+ * WC_Stripe_Subscription_Initial_Test
+ */
+class WC_Stripe_Subscription_Initial_Test extends WP_UnitTestCase {
```
daledupreez, line 26: Not blocking: it may be clearer to call this something like `WC_Stripe_Subscription_Purchase_Test` or `WC_Stripe_Subscription_Initial_Purchase_Test` to reflect that we're testing the initial purchase flow.

[Edit] I now see that it's actually a file move, so probably best not to change it as part of this PR.

daledupreez (approved):
Things are looking good to me. I have some minor non-blocking comments that don't need to be folded in.

Mayisha (approved):
Looks good to me.

Let's address the coderabbit suggestion about the test.

Also, we still have some places across the codebase where `wc-stripe-is-deferred-intent` is found. We should clean up those as well.
1. `WC_Stripe_UPE_Payment_Gateway_Test` is still setting this flag in the tests. 
2. Some frontend code is setting this flag to include in the request data ([example](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/update/removing-the-deferred-intent-flag/client/blocks/normalize.js#L64))

tests/phpunit/Compat/WC_Stripe_Subscription_Renewal_Test.php
wjrosa: Reverted it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4929/commits/2df354f4553657d8b38d7ce324407dacf4cbda6b

tests/phpunit/Compat/WC_Stripe_Subscription_Initial_Test.php
```diff
@@ -0,0 +1,141 @@
...
+
+		add_filter( 'pre_http_request', $pre_http_request_response_callback, 10, 3 );
+
+		// Act: call process_subscription_payment().
+		// We need to use `wc_gateway_stripe` here because we mocked this class earlier.
+		$result = $wc_gateway_stripe->process_payment( $order_id );
+
+		// Clean up.
+		remove_filter( 'pre_http_request', [ $this, 'pre_http_request_response_success' ] );
+		WC_Stripe_Helper::delete_main_stripe_settings();
```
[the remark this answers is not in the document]
wjrosa: Suggestion applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4929/commits/d52efe29cced1cdb75e0caefa309b8e5a3bc2e38

PR #4930, opened 2026-01-13
Removing long-deprecated code

wjrosa:

In this cleaning-up PR, I am removing all deprecated code up to 10.1.0. Methods, classes, and include calls for those.

## Testing instructions

Code review. Perform some basic smoke testing to make sure nothing breaks.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-helper.php
```diff
@@ -74,166 +74,6 @@ public static function delete_main_stripe_settings() {
...
-	/**
-	 * Gets the Stripe currency for order.
-	 *
-	 * @since 4.1.0
-	 * @param object $order
-	 * @return string $currency
-	 *
-	 * @deprecated 10.0.0 Use `WC_Stripe_Order_Helper::get_stripe_currency()` instead.
-	 */
-	public static function get_stripe_currency( $order = null ) {
```
malithsen, line 86: Since we weren't emitting deprecation notices via `wc_deprecated_function`, I wonder whether 3rd-party plugins that might be using these helpers are truly aware of the deprecation. 

Maybe it'd be safer to wrap with `wc_deprecated_function` and then remove after one more release?

malithsen (approved):
Thanks for removing this dead code 💯 
This looks good to me.
However, I am concerned that users who might be relying on some of our helper methods are truly aware of the deprecation, as they were not properly emitted.

includes/class-wc-stripe-helper.php
daledupreez: I think this is probably the right approach (even if I would prefer to remove all the code!). I think we need to be a bit conservative to err on the side of keeping stores running, even if third parties have done weird or unsupported things.

wjrosa: Oh right. We can do that. Reverted the removal of them and added the `wc_deprecated_function` calls in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/bf8ea9ecca2dae50d28c593fef26d959025512f5

includes/class-wc-stripe-feature-flags.php
```diff
@@ -35,7 +16,6 @@ class WC_Stripe_Feature_Flags {
 	protected static $feature_flags = [
 		'_wcstripe_feature_upe'                => 'yes',
```
daledupreez: Not necessarily for this PR, but I think we can also remove this feature flag. 😁

includes/class-wc-stripe-helper.php
```diff
@@ -1164,33 +1180,6 @@ private static function should_load_scripts_for_ece_location( $location ) {
...
-	/**
-	 * Adds payment intent id and order note to order if payment intent is not already saved
-	 *
-	 * @param $payment_intent_id
-	 * @param $order
-	 *
-	 * @deprecated 10.0.0 Use WC_Stripe_Order_Helper::add_payment_intent_to_order() instead.
-	 */
-	public static function add_payment_intent_to_order( $payment_intent_id, $order ) {
-		$order_helper  = WC_Stripe_Order_Helper::get_instance();
```
daledupreez, line 1176: I think we should also add the call to `wc_deprecated_function()` for `add_payment_intent_to_order()`.

includes/class-wc-stripe-helper.php
```diff
@@ -1279,25 +1268,6 @@ public static function get_payment_method_from_intent( $intent ) {
...
-	/**
-	 * Returns the payment intent or setup intent ID from a given order object.
-	 *
-	 * @param WC_Order $order The order to fetch the Stripe intent from.
-	 *
-	 * @return string|bool  The intent ID if found, false otherwise.
-	 *
-	 * @deprecated 10.0.0 Use WC_Stripe_Order_Helper::get_intent_id_from_order() instead.
-	 */
-	public static function get_intent_id_from_order( $order ) {
```
daledupreez, line 1291: As above, I think we should also add the call to `wc_deprecated_function()` for `get_intent_id_from_order()`.

includes/class-wc-stripe-helper.php
```diff
@@ -1316,44 +1286,6 @@ public static function get_stripe_gateway_ids() {
...
-	 *
-	 * @deprecated 10.0.0 Use WC_Stripe_Order_Helper::remove_payment_awaiting_action() instead.
-	 */
-	public static function remove_payment_awaiting_action( $order, $save = true ) {
-		$order->delete_meta_data( self::PAYMENT_AWAITING_ACTION_META );
-
-		if ( $save ) {
-			$order->save();
-		}
-	}
```
daledupreez, line 1355: Here and below for most of the other removed methods -- it feels like we should be calling `wc_deprecated_function()` for all of them except `is_wallet_payment_method()`, which can be safely removed.

includes/class-wc-stripe-logger.php
```diff
@@ -29,45 +29,6 @@ class WC_Stripe_Logger {
...
 	public static $logger;
 
-	/**
-	 * Utilize WC logger class
-	 *
-	 * @deprecated 9.7.0 Use the shortcut methods for each log severity level: info(), error(), etc. instead.
-	 *
-	 * @since 4.0.0
-	 */
-	public static function log( $message, $start_time = null, $end_time = null ) {
```
daledupreez, line 45: Are we sure 100% of our callers have been updated? This also feels like it _really_ should call `wc_deprecated_function()` for a release.

phpstan-baseline.neon
```diff
@@ -6582,6 +6312,12 @@ parameters:
 			count: 1
 			path: includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
 
+		-
+			message: '#^Call to an undefined static method WC_Stripe_Logger\:\:log\(\)\.$#'
```
daledupreez: It looks like we still have a caller!

daledupreez (changes requested):
We still have some cleanup to do.

includes/class-wc-stripe-feature-flags.php
wjrosa: I agree. Removed it already in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/2ff2e862547bd0737f7ac1887991a964579f78b5

includes/class-wc-stripe-helper.php
wjrosa: Makes sense. Did it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/94a592ae0e9ecc68f9c0024a7a803587d05167c3

includes/class-wc-stripe-helper.php
wjrosa: Did it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/94a592ae0e9ecc68f9c0024a7a803587d05167c3

includes/class-wc-stripe-helper.php
wjrosa: Did it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/94a592ae0e9ecc68f9c0024a7a803587d05167c3

phpstan-baseline.neon
wjrosa: Good catch! Replaced the call in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/285824bc9a0fa96d331bdc85abdaf02730b4cf81

includes/class-wc-stripe-logger.php
wjrosa: Alright. Did it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/ad00f813b606d4a07dc1e787f540d118df1a6065

includes/class-wc-stripe-helper.php
```diff
@@ -1710,6 +1802,99 @@ public static function has_gateway_plugin_active( $plugin_id, $available_payment
...
+	 * @param string|null   $selected_payment_type The selected payment type, which is generally applicable for updates. If null, we will use the stored payment type for the order.
+	 *
+	 * @throws Exception Throws an exception if the intent is not valid for the order.
+	 *
+	 * @deprecated 10.0.0 Use WC_Stripe_Order_Helper::validate_intent_for_order() instead.
+	 */
+	public static function validate_intent_for_order( $order, $intent, ?string $selected_payment_type = null ): void {
+		wc_deprecated_function( __METHOD__, '10.0.0', 'WC_Stripe_Order_Helper::validate_intent_for_order()' );
+
+		$intent_id = null;
```
daledupreez, line 1821: I just checked the implementation for most of these functions and it looks like they are the same in this code and in `WC_Stripe_Order_Helper`. Is it worth updating all of these methods to simply call the new code where we can do so? The longer we have what is effectively duplicated code, the easier it will be for them to drift apart.

Also, I think we should tackle it in a follow-up rather than in this PR, just to be clear!

phpstan-baseline.neon
```diff
@@ -5827,20 +5623,20 @@ parameters:
 			path: includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
 
 		-
-			message: '#^Parameter \#2 \$product of function wc_format_stock_quantity_for_display expects WC_Product, WC_Product\|false\|null given\.$#'
+			message: '#^Parameter \#2 \$callback of function add_action expects callable\(\)\: mixed, array\{\$this\(WC_Stripe_Express_Checkout_Ajax_Handler\), ''ajax_pay_for_order''\} given\.$#'
```
daledupreez: It looks like `ajax_pay_for_order` was removed, but is still being hooked in somewhere.

daledupreez (commented):
This is looking really good overall! However, it still looks like one of the removed functions is hooked in somewhere. PHPStan for the win!

phpstan-baseline.neon
wjrosa: Nice catch! Removed the reference in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/1a636d12b16860d6b9d642d8a33ed66134b7d441

readme.txt
```diff
@@ -142,6 +142,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.4.0 - xxxx-xx-xx =
+* Update - Remove all deprecated code up to version 10.1.0
```
daledupreez, line 151: I think we should also add some details about what was removed in the Compatibility Notes. Maybe something like the following:
```
 * Remove deprecated feature flags for UPE, ECE, and OCS; remove various helpers including is_sepa_tokens_for_other_methods_enabled() validate_minimum_order_amount(), get_booking_id_from_cart(), get_owner_details(), lock_order_payment(), unlock_order_payment(), lock_order_refund(), unlock_order_refund() and remove AJAX pay for order handler.

daledupreez (approved):
Thanks for fixing that issue! This is looking good to me, though I think we do need some compatibility documentation.

readme.txt
wjrosa: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4930/commits/6d205271e2551f27b421e2c107f6631e717a0e66

PR #4931, opened 2026-01-13
Deprecating unused feature flag methods

wjrosa:

In this PR, I am deprecating some old, unused feature flag methods in the backend and removing all references to them from the frontend.

## Testing instructions

Code review.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed payment method deprecation messaging from the checkout interface.

* **Chores**
  * Cleaned up legacy deprecation logic for alternative payment methods.
  * Simplified internal payment method handling and state management.
  * Marked unused backend methods as deprecated.


includes/class-wc-stripe-feature-flags.php
```diff
@@ -112,6 +114,8 @@ public static function did_merchant_disable_upe() {
 	 * Checks if the APMs are deprecated. Stripe deprecated them on October 29, 2024 (for the legacy checkout).
 	 *
 	 * @return bool Whether the APMs are deprecated.
+	 *
+	 * @deprecated 10.5.0 APMs are deprecated and the legacy checkout no longer exists. This method will be removed in a future release.
 	 */
 	public static function are_apms_deprecated() {
 		return false;
```
daledupreez, line 133: It feels like we should add `wc_deprecated_function()` calls as well.
```suggestion
	wc_deprecated_function( __FUNCTION__, '10.5.0' );

	$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
	return ! empty( $stripe_settings[ self::UPE_CHECKOUT_FEATURE_ATTRIBUTE_NAME ] ) && 'disabled' === $stripe_settings[ self::UPE_CHECKOUT_FEATURE_ATTRIBUTE_NAME ];
}


/**
	 * Checks if the APMs are deprecated. Stripe deprecated them on October 29, 2024 (for the legacy checkout).
	 *
	 * @return bool Whether the APMs are deprecated.
	 *
	 * @deprecated 10.5.0 APMs are deprecated and the legacy checkout no longer exists. This method will be removed in a future release.
	 */
public static function are_apms_deprecated() {
	wc_deprecated_function( __FUNCTION__, '10.5.0' );
	return false;
```

daledupreez (commented):
It looks like the e2e and linting issues are legitimate problems relating to an undefined variable and some other cleanup. I think we need to get those resolved.

client/settings/general-settings-section/payment-method.js
Mayisha: We need to remove this `deprecated` prop.

wjrosa: Good catch. I removed this reference, along with the prop definition in the component, and the pill component itself in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4931/commits/483a562421ebcf1d4fd9cee97def29978281b67f

Mayisha (approved):
LGTM 👍

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -241,7 +241,6 @@ public function admin_scripts( $hook_suffix ) {
 			'is_test_mode'                          => $this->get_gateway()->is_in_test_mode(),
 			'plugin_version'                        => WC_STRIPE_VERSION,
 			'account_country'                       => $this->account->get_account_country(),
-			'are_apms_deprecated'                   => WC_Stripe_Feature_Flags::are_apms_deprecated(),
```
daledupreez, line 244: I am wondering whether it may be safer to keep the flag present for a release or two, just in case there are ecosystem consumers of the data? Given that `undefined` is never going to be truthy, it is probably OK to drop this, however.

```suggestion
			'are_apms_deprecated'                   => false,
			'is_amazon_pay_available'               => WC_Stripe_Feature_Flags::is_amazon_pay_available(),
```

includes/class-wc-stripe-feature-flags.php
```diff
@@ -110,19 +110,26 @@ public static function is_checkout_sessions_available() {
...
-
 	/**
 	 * Checks if the APMs are deprecated. Stripe deprecated them on October 29, 2024 (for the legacy checkout).
 	 *
 	 * @return bool Whether the APMs are deprecated.
+	 *
+	 * @deprecated 10.5.0 APMs are deprecated and the legacy checkout no longer exists. This method will be removed in a future release.
 	 */
 	public static function are_apms_deprecated() {
+		wc_deprecated_function( __FUNCTION__, '10.5.0' );
```
daledupreez: CodeRabbit's suggestion to use `__METHOD__` seems like a good one, both here and in `did_merchant_disable_upe()`.
```suggestion
		wc_deprecated_function( __METHOD__, '10.5.0' );
```

daledupreez (approved):
This change is looking good to me, though it may be worth updating the magic constant for the deprecated methods.

PR #4935, opened 2026-01-14
Add token validation checks to prevent fatal errors when token creation fails

Mayisha:
Fixes STRIPE-889

- Fixed the return and logs in `create_token_from_setup_intent` method. 
- Added checks for null in the caller methods and handled the null case depending on each caller method (either throw an error or add condional block)
- phpstan baseline update.

## Testing instructions
- Code review.
- Checkout to this branch `fix/handle-create-token-from-setup-intent-error` and test the possible scenarios when the `create_token_from_setup_intent` might be called. Some possible scenarios are-
    - Save payment during checkout
    - Add a card in the 'My Account > Payment Methods' page.
    - Checkout with a saved payment method.
    - Purchase a subscription.
    - Change the payment method of a subscription.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,7 +7,7 @@
 * Dev - Ensure PHPStan runs when pushing changes
 * Dev - Add PHPStan stub for WC_Subscription class
 * Dev - Remove the deprecated WC_Stripe_Apple_Pay class
-
+* Fix - Return null from create_token_from_setup_intent on failure and add validation checks to prevent fatal errors when token creation fails
 
 = 10.3.0 - 2026-01-13 =
 * Update - Increase Afterpay/Clearpay maximum transaction amount to 4,000 AUD and 4,000 NZD
--- a/changelog.txt
+++ b/changelog.txt
@@ -10,7 +10,7 @@
 * Dev - Ensure PHPStan runs when pushing changes
 * Dev - Add PHPStan stub for WC_Subscription class
 * Dev - Remove the deprecated WC_Stripe_Apple_Pay class
-* Fix - Return null from create_token_from_setup_intent on failure and add validation checks to prevent fatal errors when token creation fails
+* Fix - Better error handling when token creation fails
 * Fix - Add order validation in Multibanco email instructions to prevent fatal error when order is invalid
 * Fix - Add validation to prevent fatal error when setting default payment token if token doesn't exist
 * Fix - Validate order object before accessing methods in my account orders actions to prevent fatal errors
--- a/includes/class-wc-stripe-intent-controller.php
+++ b/includes/class-wc-stripe-intent-controller.php
@@ -1335,7 +1335,7 @@ class WC_Stripe_Intent_Controller {
 			$token   = $gateway->create_token_from_setup_intent( $setup_intent_id, $subscription->get_user() );
 
 			if ( ! $token ) {
-				throw new WC_Stripe_Exception( 'token_creation_failed', __( "We're not able to process this subscription change payment request payment. Please try again later.", 'woocommerce-gateway-stripe' ) );
+				throw new WC_Stripe_Exception( 'token_creation_failed', __( "We can't process your payment method change at this time. Please try again later.", 'woocommerce-gateway-stripe' ) );
 			}
 
 			$notice = __( 'Payment method updated.', 'woocommerce-gateway-stripe' );
--- a/includes/class-wc-stripe-intent-controller.php
+++ b/includes/class-wc-stripe-intent-controller.php
@@ -1333,7 +1333,12 @@ class WC_Stripe_Intent_Controller {
 
 			$gateway = $this->get_upe_gateway();
 			$token   = $gateway->create_token_from_setup_intent( $setup_intent_id, $subscription->get_user() );
-			$notice  = __( 'Payment method updated.', 'woocommerce-gateway-stripe' );
+
+			if ( ! $token ) {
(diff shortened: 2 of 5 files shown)


wjrosa (approved):
Nice improvement! :shipit:

changelog.txt
```diff
@@ -9,7 +9,7 @@
 * Dev - Ensure PHPStan runs when pushing changes
 * Dev - Add PHPStan stub for WC_Subscription class
 * Dev - Remove the deprecated WC_Stripe_Apple_Pay class
-
+* Fix - Return null from create_token_from_setup_intent on failure and add validation checks to prevent fatal errors when token creation fails
```
daledupreez: This is really specific and verbose. Might it be better to focus on the outcomes rather than the specifics?
```suggestion
* Fix - Better error handling when token creation fails
```

includes/class-wc-stripe-intent-controller.php
```diff
@@ -1333,7 +1333,12 @@ public function confirm_change_payment_from_setup_intent_ajax() {
 
 			$gateway = $this->get_upe_gateway();
 			$token   = $gateway->create_token_from_setup_intent( $setup_intent_id, $subscription->get_user() );
-			$notice  = __( 'Payment method updated.', 'woocommerce-gateway-stripe' );
+
+			if ( ! $token ) {
+				throw new WC_Stripe_Exception( 'token_creation_failed', __( "We're not able to process this subscription change payment request payment. Please try again later.", 'woocommerce-gateway-stripe' ) );
```
daledupreez: This message is confusing to me, as I am not sure what we're wanting to communicate. I think we can simplify the wording a bit and make it a bit more generic.
```suggestion
				throw new WC_Stripe_Exception( 'token_creation_failed', __( "We can't process your payment method change at this time. Please try again later.", 'woocommerce-gateway-stripe' ) );
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1734,6 +1736,9 @@ public function maybe_process_upe_redirect() {
 					try {
 						$setup_intent_id = isset( $_GET['setup_intent'] ) ? wc_clean( wp_unslash( $_GET['setup_intent'] ) ) : '';
 						$token           = $this->create_token_from_setup_intent( $setup_intent_id, wp_get_current_user() );
+						if ( ! $token ) {
+							throw new Exception( __( 'Unable to create token for updating payment method billing information.', 'woocommerce-gateway-stripe' ) );
```
daledupreez: As above, can we simplify this error message? If we want, we could mention a "recurring token", but that may be too complicated.
```suggestion
							throw new Exception( __( 'Unable to create token for the payment method.', 'woocommerce-gateway-stripe' ) );
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2503,6 +2508,7 @@ public function create_token_from_setup_intent( $setup_intent_id, $user ) {
 		try {
 			$setup_intent = $this->stripe_request( 'setup_intents/' . $setup_intent_id . '?&expand[]=latest_attempt' );
 			if ( ! empty( $setup_intent->last_payment_error ) ) {
+				WC_Stripe_Logger::error( 'Error in creating token from setup intent.', [ 'error' => $setup_intent->last_payment_error ] );
```
daledupreez: Can we use a different error message for this log entry? We're going to log the same text in the `catch` block below, so telling them apart won't be easy. Though I see that we can get the Stripe error details in this entry, so it's probably worth keeping.

daledupreez (commented):
I have some feedback from inspection. I'll test tomorrow.

includes/class-wc-stripe-intent-controller.php
Mayisha: I copied the message from other failure cases in the same flow. The suggestion looks good. Updating here.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Changed in [7057566](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4935/commits/7057566a12d7dbf65ff3d8cb456b60002d9dd078)

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1734,6 +1736,9 @@ public function maybe_process_upe_redirect() {
 					try {
 						$setup_intent_id = isset( $_GET['setup_intent'] ) ? wc_clean( wp_unslash( $_GET['setup_intent'] ) ) : '';
 						$token           = $this->create_token_from_setup_intent( $setup_intent_id, wp_get_current_user() );
+						if ( ! $token ) {
+							throw new Exception( __( 'Unable to create token for the payment method.', 'woocommerce-gateway-stripe' ) );
```
daledupreez, line 1740: Minor note: it doesn't look like we are showing any shopper-facing error or notice from the catch block for this `try`. Not blocking for this PR, but maybe worth handling in a follow-up.

daledupreez (approved):
The changes look good to me based on inspection. Thanks for adding the error handling for these cases!

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Added a notice in catch block.

PR #2561, opened 2023-03-02
Support eMandates for recurring payments made in INR

reykjalin:
Supercedes #1981 

## Changes proposed in this Pull Request

The changes in this PR are intended to add support for new 3DS regulations for cards issued in India.

- Migrate all card payments to the Payment Method API from the Source API
- Add necessary information to API requests for subscription payment that allow Stripe to follow Indian 3DS regulations
- Add handling for specific payment intent statuses and errors that can occur during the new Indian 3DS flow

## Known Issues

<details><summary>There are several known issues with the current implementation, expand for details</summary>

- If you save a card before applying the changes included in this PR it will not work with the new Indian 3DS regulations.
    - Manual fix: delete the saved payment method from your account and add it again.
    - Repro steps:
        - Set your store currency to INR
        - `git switch develop`
        - `npm run start:webpack` (you can exit this as soon as it's finished bundling the JS)
        - Buy a regular product with the Indian 3DS mandate test card: `4000003560000008` and make sure to save the card
        - `git switch fix/indian-3ds-recurring-regulation-change`
        - `npm run start:webpack` (you can exit this as soon as it's finished bundling the JS)
        - Buy a subscription using the card you saved earlier
        - Go to **Tools** > **Scheduled Actions** > **Pending**
        - Locate the renewal task for the subscription you just bought
        - Click **Run** to run the renewal payment
        - Notice that the payment is immediately successful in your Stripe dashboard when it should take ~5 minutes to transition from `requires_action` to `succeeded`
    - Suggested fix: Not sure. Transition the card somehow from being a `src_xxx` to `pm_xxx`? Not sure if that's possible though.
- Changing sub payment method from one that requires a mandate (`4000003560000008`) to one that does not (e.g. `4000000000003220`) is not possible.
    - Repro steps:
        - Purchase a subscription with Indian 3DS mandate test card `4000003560000008`
        - Go to **Tools** > **Scheduled Actions** > **Pending**
        - Locate the renewal task for the subscription you just bought
        - Click **Run** to run the renewal payment
        - Wait ~5 minutes to make sure the renewal payment is successful (order will transition from `pending` to `processing`)
        - Go to **My Account** > **Subscriptions**, find the subscription you just purchased, and change the payment method to a non-India 3DS mandate test card, e.g. `4000000000003220`.
        - Go to **Tools** > **Scheduled Actions** > **Pending**
        - Locate the renewal task for the subscription you just bought
        - Click **Run** to run the renewal payment
[...]

PR #4943, opened 2026-01-19
Fix edit permissions for indian subscriptions

wjrosa:
Fixes STRIPE-884

In this PR, I am changing the main logic used to block editing of Indian subscriptions. Instead of just checking for any existing mandate ID meta in the parent order, I am now retrieving the mandate object from the Stripe API and comparing the specific fields that flag an Indian subscription as uneditable.

## Testing instructions

- Checkout and build this branch on your test environment (`fix/edit-permissions-for-indian-subscriptions`)
- Install the subscription extension and add a subscription product
- Connect your Stripe account
- As a customer, purchase the subscription using Stripe Link (or another mandate-based payment method: SEPA, Bancontact, iDEAL, BACS, ACH, etc.)
- Complete the purchase successfully
- As the customer, go to "My Account → Subscriptions" and change the payment method to a regular Stripe credit card
- As a merchant/admin, navigate to WooCommerce → Orders and open the subscription
- Attempt to edit a line item (e.g., change shipping method from "DHL" to "Local Pickup")
- Confirm the pencil icon now appears next to line items, allowing edits

If you want, connect an Indian Stripe account, set your store currency to INR, purchase the subscription again, and this time confirm you are not able to edit it as a merchant.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1242,22 +1242,44 @@ public function update_subscription_payment_method_from_order( $order, $payment_
...
+		}
+
+		$method_details = $mandate->payment_method_details;
+		if ( WC_Stripe_Payment_Methods::CARD !== ( $method_details->type ?? '' ) ) {
+			return $editable;
+		}
+
+		$amount_type     = $method_details->card->amount_type ?? '';
+		$supported_types = $method_details->card->supported_types ?? [];
+		if ( 'fixed' !== $amount_type || empty( $supported_types ) || 'india' !== $supported_types[0] ) {
```
daledupreez: @wjrosa, do we have a doc reference for these fields? I am not seeing them documented in [the relevant API docs](https://docs.stripe.com/api/mandates/object?api-version=2024-06-20#mandate_object-payment_method_details-card) or [the recurring payments guide for India](https://docs.stripe.com/india-recurring-payments), and I am wondering how we know what values to look for.

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
```diff
@@ -844,4 +844,116 @@ public function test_invalid_request_error_with_non_blocked_code_is_retryable()
...
+			WC_Stripe_Order_Helper::get_instance()->update_stripe_mandate_id( $parent_order, $mandate_id );
+			$parent_order->save_meta_data();
+
+			$order->set_parent_id( $parent_order->get_id() );
+			$order->save();
+		}
+
+		// Mock response from Stripe API using request arguments.
+		$mock_request = function ( $preempt, $parsed_args, $url ) {
+			$mandate_id = str_replace( 'https://api.stripe.com/v1/mandates/', '', $url );
```
daledupreez: Nit: it feels like we should ignore all requests for non-mandate URLs.
```suggestion
			if ( ! str_starts_with( $url, 'https://api.stripe.com/v1/mandates/' ) ) {
				return $preempt;
			}
			$mandate_id = str_replace( 'https://api.stripe.com/v1/mandates/', '', $url );
```

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
```diff
@@ -844,4 +844,116 @@ public function test_invalid_request_error_with_non_blocked_code_is_retryable()
...
+					// Card payment method, but not Indian.
+					$payment_method_type = 'card';
+					$amount_type         = '';
+					$supported_types     = [];
+					break;
+				case 'mandate_789':
+					// Indian card payment method.
+					$payment_method_type = 'card';
+					$amount_type         = 'fixed';
+					$supported_types     = [ 'india' ];
```
daledupreez: Is it possible for a card from India to have a value other than `fixed`? If so, it feels like we should test for that.

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
```diff
@@ -844,4 +844,116 @@ public function test_invalid_request_error_with_non_blocked_code_is_retryable()
...
+					break;
+				default:
+					return $preempt;
```
daledupreez: Minor note: might it not be easier to have the mandate data supplied by the data provider? That way scanning the data provider allows you to see exactly what cases we are and are not testing. With just the ID being specified, we need to compare the description in the data provider against the data being specified in the test implementation.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1242,22 +1242,44 @@ public function update_subscription_payment_method_from_order( $order, $payment_
 	}
 
 	/**
-	 * Disables the ability to edit a subscription for orders with mandates.
+	 * Disables the ability to edit a subscription for Indian orders with fixed amount mandates.
 	 *
 	 * @param $editable boolean The current editability of the subscription.
-	 * @param $order WC_Order The order object.
+	 * @param $order \WC_Order The subscription order object.
```
daledupreez: Nit: The PHPDoc here is not formatted correctly.
```suggestion
	 * @param boolean   $editable The current editability of the subscription.
	 * @param \WC_Order $order    The order object.
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1242,22 +1242,44 @@ public function update_subscription_payment_method_from_order( $order, $payment_
...
 	 * @return boolean true if the subscription can be edited, false otherwise.
 	 */
 	public function disable_subscription_edit_for_india( $editable, $order ) {
 		$parent_order = wc_get_order( $order->get_parent_id() );
-		if ( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled()
-			&& $this->is_subscription( $order )
```
daledupreez, line 1279: Is there a reason for removing the `$this->is_subscription()` check? It feels like the method naming and logic are 100% focused on subscriptions, but we are not checking if we have a subscription anywhere.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/20e74c7839185b1cf81b5c4d5d60b7b862536198

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: It was just to keep the code simpler. This check is a bit redundant, since we don't have mandates for non-subscriptions (and we skip orders without the meta). I will put it back just to keep it safer

wjrosa: Added it back in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/cddb922bde4ef2c9b7860186c0f6e2c4245dad8c

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
wjrosa: Good idea. Moved the definition to the provider in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/c0288cc97cdd8c53f25f6feda517ca3cf448e74a

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
wjrosa: Makes sense. Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/ccf11b489b2a7cad9e3f02aa496d49f227a6080a

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: I could not see anything there either. Most of the work done in this PR was guided by the AI suggestions present in the source issue (STRIPE-884). You can take a look there to better understand the reasoning. As for the field values here, you can see the PR adding them [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/2561). Those seem directly related to 3DS payments (not general/regular).

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1242,22 +1242,48 @@ public function update_subscription_payment_method_from_order( $order, $payment_
...
+		if ( empty( $mandate_id ) ) {
+			return $editable;
+		}
+
+		$subscription_id = $order->get_id();
+		$cache_key       = 'mandate_for_subscription_' . $subscription_id;
+
+		$mandate = WC_Stripe_Database_Cache::get( $cache_key );
+		if ( ! $mandate ) {
+			$mandate = WC_Stripe_API::retrieve( 'mandates/' . $mandate_id );
```
Mayisha: We should check for errors in this request.

includes/compat/trait-wc-stripe-subscriptions.php
Mayisha: @wjrosa the mandate response does not have these data. I have tested locally and my subscription remains editable in this branch for a subscription purchased with an Indian card with mandate support.

This is the response I see in my logs for the `/mandate` request-

```
{
    "stripe_version": "10.3.1",
    "stripe_api_version": "2024-06-20",
    "stripe_api_key": "sk_test_..",
    "stripe_request_id": "req_...",
    "response": {
        "id": "mandate_1StnkjSGrDfhP3zSLCr1YLcg",
        "object": "mandate",
        "customer_acceptance": {
            "accepted_at": 1769426068,
            "online": {
                "ip_address": [redacted],
                "user_agent": "WooCommerce Stripe Gateway/10.3.1 (https://woocommerce.com/products/stripe/)"
            },
            "type": "online"
        },
        "livemode": false,
        "multi_use": {},
        "payment_method": "pm_1StnkRSGrDfhP3zSZJaH0lov",
        "payment_method_details": {
            "card": {},
            "type": "card"
        },
        "status": "active",
        "type": "multi_use"
    }
}
```

Mayisha (changes requested):

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1267,22 +1267,33 @@ public function update_subscription_payment_method_from_order( $order, $payment_
...
+		$payment_method_id = WC_Stripe_Order_Helper::get_instance()->get_stripe_source_id( $parent_order );
+		$payment_method    = $this->stripe_request( 'payment_methods/' . $payment_method_id );
+
+		// If the payment method is not a card or the card is not issued in India, allow editing.
+		if ( ! $payment_method || WC_Stripe_Payment_Methods::CARD !== ( $payment_method->type ?? '' ) || 'IN' !== ( $payment_method->card->country ?? '' ) ) {
+			return $editable;
 		}
 
-		return $editable;
+		return false;
```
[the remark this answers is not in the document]
wjrosa: Suggestion applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/bebada73f67d87f84d6693213b00aa6b4f0f1eb6

wjrosa:
Thanks for checking the correct response format, @Mayisha ! I ran some tests here and fixed the main logic by simply checking the payment method's country. I'd appreciate another round of review

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1267,22 +1267,37 @@ public function update_subscription_payment_method_from_order( $order, $payment_
...
+			$payment_method = $this->stripe_request( 'payment_methods/' . $source_id );
+			WC_Stripe_Database_Cache::set( $cache_key, $payment_method, HOUR_IN_SECONDS );
+		}
+
+		// If the payment method is not a card or the card is not issued in India, allow editing.
+		if ( ! $payment_method || 'IN' !== ( $payment_method->card->country ?? '' ) ) {
+			return $editable;
+		}
+
+		return false;
```
daledupreez: Nit: The logic feels inverted here, as I would normally expect the "special" case to be the one we check for and do something different. I have preference for returning `false` from this if block, and then having the default be to return `$editable`.
```suggestion
		// If the payment method is not a card or the card is not issued in India, allow editing.
		if ( $payment_method && 'card' === $payment_method->type && 'IN' === ( $payment_method->card->country ?? '' ) ) {
			return false;
		}

		return $editable;
```

includes/compat/trait-wc-stripe-subscriptions.php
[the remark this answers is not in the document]
daledupreez: This feels like it could be _terrible_ on sites with lots of subscriptions.

Are there some local fields we could use to guess at the billing method and/or billing country so we're not making remote requests to Stripe? It feels like we should be able to look at the local token to at least identify the payment method.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1267,22 +1267,37 @@ public function update_subscription_payment_method_from_order( $order, $payment_
...
 	public function disable_subscription_edit_for_india( $editable, $order ) {
-		$parent_order = wc_get_order( $order->get_parent_id() );
-		if ( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled()
-			&& $this->is_subscription( $order )
-			&& $parent_order
-			&& ! empty( WC_Stripe_Order_Helper::get_instance()->get_stripe_mandate_id( $parent_order ) ) ) {
-			$editable = false;
+		if ( ! WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() || ! $this->is_subscription( $order ) ) {
+			return $editable;
 		}
```
Mayisha, line 1279:
```suggestion
		}
		
	// Bail if subscription does not have a mandate ID
	if ( empty( WC_Stripe_Order_Helper::get_instance()->get_stripe_mandate_id( $parent_order ) ) {
			return $editable
			}
	 	
		
```

We can keep the mandate check here and bail early if the subscription does not have any mandate ID.

includes/compat/trait-wc-stripe-subscriptions.php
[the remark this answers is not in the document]
Mayisha: I would recommend the option 2 here. We can add a check and see if this is the subscription edit page and continue with the rest of the code on on subscription edit page.

With this check added, I think we should be good here and don't need to bother much about the API call. Also On subscription edit page, we already make an API call to get the payment method to check if the payment method is detached ([ref](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/admin/class-wc-stripe-admin-notices.php#L455)) and we cache the result. So in most of the cases, we might end up getting the value in cache also

Mayisha (commented):
This mostly looks good to me and tests well. I would recommend we address the feedback related to the subscription edit page check.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1267,22 +1267,44 @@ public function update_subscription_payment_method_from_order( $order, $payment_
...
+			$payment_method = $this->stripe_request( 'payment_methods/' . $source_id );
+			WC_Stripe_Database_Cache::set( $cache_key, $payment_method, HOUR_IN_SECONDS );
+		}
+
+		// If the payment method is not a card or the card is not issued in India, allow editing.
+		if ( ! $payment_method || 'IN' !== ( $payment_method->card->country ?? '' ) ) {
+			return $editable;
+		}
+
+		return false;
```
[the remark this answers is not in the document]
wjrosa: Fixed previously

includes/compat/trait-wc-stripe-subscriptions.php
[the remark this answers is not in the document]
wjrosa: Good call, folks! I added a check for the current screen in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/7f3f2642e3229def91acef227541c13950714d20

wjrosa:
I am working on fixing the unit tests 👀. Edit: fixed

Mayisha (approved):
Looks good to me. I have tested and verified that subscriptions paid with a normal card is editable but subscriptions paid with Indian 3DS card is not editable in this branch. 👍

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.5.0 - xxxx-xx-xx =
+* Fix - Correctly handle edit permissions for Indian subscriptions
```
daledupreez: Nit: I think it may be worth updating the changelog as we're actually doing two things in this PR:
* We are generally _allowing_ edits for subscriptions with mandates
* We are restricting edit permissions for subscriptions with mandates where the payment method is a card issued in India
```suggestion
* Fix - Allow subscription edits when mandates are present, except for mandates from India
```

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
```diff
@@ -844,4 +845,171 @@ public function test_invalid_request_error_with_non_blocked_code_is_retryable()
...
+				'is subscription'   => true,
+				'edit page'         => true,
+				'has parent order'  => true,
+				'parent mandate ID' => 'mandate_123',
+				'payment method'    => [
+					'id'                     => 'pm_456',
+					'type' => 'card',
+					'card' => [
+						'amount_type'     => '',
+						'supported_types' => [],
```
daledupreez: Nit: Should we specify some other country to make sure that we have the expected data?

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
```diff
@@ -844,4 +845,171 @@ public function test_invalid_request_error_with_non_blocked_code_is_retryable()
...
+			],
+			'missing payment method ID meta'                 => [
+				'is subscription'   => true,
+				'edit page'         => true,
+				'has parent order'  => true,
+				'parent mandate ID' => 'mandate_123',
+				'payment method'    => [],
+				'expected'          => true,
+			],
+			'payment method is not card'                      => [
```
daledupreez: Question, just to be 100% sure: do any of the test cases below cover the cases where subscriptions have  Apple Pay, Google Pay, or Link as the saved payment method? My understanding is that all three come through as `card` payments, but I want to make sure we will get the same data shape for them.

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
```diff
@@ -844,4 +845,171 @@ public function test_invalid_request_error_with_non_blocked_code_is_retryable()
...
+				'expected'          => true,
+			],
+			'not subscriptions edit page'                    => [
+				'is subscription'   => true,
+				'edit page'         => false,
+				'has parent order'  => false,
+				'parent mandate ID' => '',
+				'payment method'    => [],
+				'expected'          => true,
+			],
```
daledupreez: For these two cases, might it be worth setting everything up for a valid subscription _except_ for the subscription edit page? As things stand, any of the various failing checks could be result in the expected result. So it feels like we could be setting ourselves up for subtle test problems.

includes/compat/class-wc-stripe-subscriptions-helper.php
```diff
@@ -322,4 +322,22 @@ public static function get_subscription_payment_method_details( string $stripe_c
...
+	 */
+	public static function is_subscription_edit_page(): bool {
+		$query_params = wp_unslash( $_REQUEST ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+		if ( WC_Stripe_Woo_Compat_Utils::is_custom_orders_table_enabled() ) { // If custom order tables are enabled, we need to check the page query param.
+			return isset( $query_params['page'] ) && 'wc-orders--shop_subscription' === $query_params['page'] && isset( $query_params['id'] );
+		}
+
+		// If custom order tables are not enabled, we need to check the post type and action query params.
+		$is_shop_subscription_post_type = isset( $query_params['post'] ) && 'shop_subscription' === get_post_type( $query_params['post'] );
+		return isset( $query_params['action'] ) && 'edit' === $query_params['action'] && $is_shop_subscription_post_type;
```
daledupreez: Not blocking, and probably a micro-optimisation, but it feels like we should leave the `get_post_type()` lookup until last.
```suggestion
		// If custom order tables are not enabled, we need to check the post type and action query params.
		if ( 'edit' !== ( $query_params['action'] ?? '' ) ) {
			return false;
		}

		return isset( $query_params['post'] ) && 'shop_subscription' === get_post_type( $query_params['post'] );
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1267,21 +1267,52 @@ public function update_subscription_payment_method_from_order( $order, $payment_
 	}
 
 	/**
-	 * Disables the ability to edit a subscription for orders with mandates.
+	 * Disables the ability to edit a subscription for Indian orders.
```
daledupreez: I think we should be more specific in the description.
```suggestion
	 * Disable edits for subscriptions that have a mandate and a card payment method from India.
```

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -770,24 +759,22 @@ public function provide_test_subscription_check_detachment() {
...
+			'post globals'   => [
+				'is edit page'    => true,
+				'subscription'    => $subscription,
 				'theorder global' => null,
-				'request params'  => [
-					'page'   => 'wc-orders--shop_subscription',
-					'id'     => $subscription->get_id(),
-					'post'   => $subscription,
-					'action' => 'edit',
-				],
```
daledupreez, line 790: Question: why did you remove the non-HPOS test cases?

daledupreez (approved):
The changes look good to me and test well, but I do have some questions and minor nits that may drive some minor changes (mostly in the tests).

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
wjrosa: Reverted these changes in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/215d1d1b723fff0a79a8f4fc15369e7a308eacec

These tests were conflicting with other tests I created for the new methods, but that's no longer necessary

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
wjrosa: Actually, that was the initial idea for this specific test case, but it ended up with a different shape. Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/08228839891532bfd2070a74f355bb5db973f07a

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
wjrosa: The difference between ECE and regular card methods is the presence of `wallet->type` inside the `card` property. I think it is OK to keep the current treatment we have, in which Indian ECE methods would fall into the block. Those are actual cards saved inside your Google/Apple Pay/Link accounts. I added a specific test case for Google Pay in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4943/commits/d4bb35b8ba618792e0bdbda436d22c4572919209

tests/phpunit/WC_Stripe_Payment_Gateway_Test.php
wjrosa: Alright. Updated both test cases in my latest commits

PR #4944, opened 2026-01-20
Run all e2e tests on PRs

daledupreez:

This PR ensures that our standard e2e tests that run on every PR run the separate tests for specific payment methods in addition to the standard and optimized checkout e2e tests that we have been running. In particular, that means that we're adding the e2e tests for ACSS, Becs, and BLIK to our standard e2e test runs so we can identify problems with those payment methods far sooner.

At an implementation level, the PR also adds a new `test:e2e-run` command, as the existing `test:e2e` command always specified the `--project=default` argument, which is incompatible with the Optimized Checkout e2e tests. The new command doesn't specify any arguments, and expects callers to provide their own.

## Testing instructions

* Code review
* Verify that the e2e tests are passing, and that the additional `acss`, `becs`, and `blik` e2e tests are running.

---

### Changelog entry

<details>

#### Comment 
This is a workflow automation change that doesn't impact merchants, so I don't think we need a changelog entry.
</details>

**Post merge**

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,6 +163,7 @@
     "test:e2e-down": "./tests/e2e/bin/down.sh",
     "test:e2e-cleanup": "npm run test:e2e-down && ./tests/e2e/bin/cleanup.sh",
     "test:e2e": "./tests/e2e/bin/run-tests.sh --project=default",
+    "test:e2e-run": "./tests/e2e/bin/run-tests.sh",
     "test:e2e-debug": "npm run test:e2e -- --debug",
     "test:e2e-legacy": "./tests/e2e/bin/run-tests.sh --project=legacy",
     "test:e2e-legacy-debug": "npm run test:e2e-legacy -- --debug",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -163,6 +163,7 @@
     "test:e2e-down": "./tests/e2e/bin/down.sh",
     "test:e2e-cleanup": "npm run test:e2e-down && ./tests/e2e/bin/cleanup.sh",
     "test:e2e": "./tests/e2e/bin/run-tests.sh --project=default",
+    "test:e2e-run": "./tests/e2e/bin/run-tests.sh",
```
wjrosa, line 166: If I understood correctly, this command runs all tests, right? If so, what if we call this command `test:e2e-all`?

wjrosa (approved):
Looks good! Left a suggestion

package.json
daledupreez: It's more that it doesn't specify a project so callers can do so. So the E2E workflow is calling `npm run test:e2e-run -- --project=foo` to run a specific project. With that in mind, `run-all` doesn't feel right, because the caller would then use `npm run test:e2e-all -- --project=foo`, which then feels confusing!

wjrosa: I see! We can keep it like this then

Mayisha (approved):
Changes look good to me and the tests are clean 👍

PR #4946, opened 2026-01-20
Agentic Commerce: CSV Feed

RadoslavGeorgiev:
Fixes STRIPE-894

This PR implements a streaming CSV feed generator for Stripe's Agentic Commerce product catalog feature, based on the Product Catalog PRs that were merged into Woo Core recently, supporting large catalogs (100k+ products, 200MB files) with constant memory usage.

### Key Changes:

1. **New: Agentic Commerce CSV Feed Generator** (`includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php`)
   - Implements WooCommerce 10.5.0+ `FeedInterface` for product feed generation
   - Streaming I/O via direct `fputcsv()` writes - handles 100k+ products without memory issues
   - Two-stage storage: writes to temp dir, moves to uploads on finalization
   - Automatic UTF-8 encoding, CSV escaping, and type conversion
   - **Why needed**: Support Stripe's Agentic Commerce feature which requires CSV product catalogs
   - **Trade-off**: Requires WooCommerce 10.5.0+ (feature-gated via version check)

2. **Comprehensive Documentation** (`includes/agentic-commerce/README.md`)
   - 344 lines covering architecture, usage, data format requirements, API reference
   - Stripe specification mapping and integration examples
   - Performance characteristics and error handling patterns

3. **Fixed WooCommerce beta version loading** (`bin/install-wp-tests.sh:181`)
   - ~Changed from alphabetical sorting to version-aware sorting using `sort -V`~
   - ~Now correctly loads WooCommerce 10.5.0-beta instead of 9.x beta versions~
   - ~**Why needed**: jq's alphabetical sort treats "10.5.0-beta.1" as less than "9.9.0-beta.1"~
   - I had this change stuck in the PR for a while and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5005 addressed it in the meantime.

4. **Created PHPStan stub files** (instead of suppressing errors)
   - Added `phpstan-stubs/woocommerce-product-feed.php` - Defines `FeedInterface` from WooCommerce 10.5.0+
   - Added `phpstan-stubs/woocommerce-filesystem-util.php` - Defines `FilesystemUtil` class
   - **Trade-off**: Stubs need maintenance if WooCommerce changes these interfaces, but provides proper type checking

5. **Comprehensive test suite** (`tests/phpunit/WC_Stripe_Agentic_Commerce_Csv_Feed_Test.php`)
   - 21 tests with 40 assertions covering:
     - Normal lifecycle (start → add_entry → end)
     - Data type handling (null, bool, int, float, arrays, objects, strings)
     - CSV escaping (quotes, commas, newlines)
     - UTF-8 encoding (café, 日本語, 🎉)
     - Error conditions (wrong column count, state violations, double start)
     - File management (temp directory, unique filenames, filename format)
     - Cleanup on error scenarios
   - Properly gated with `markTestSkipped()` for WooCommerce < 10.5.0

### Implementation Details:

**Feed Interface Compliance:**
- Follows WooCommerce's `FeedInterface` pattern (similar to `JsonFileFeed` in core)
- Methods: `start()`, `add_entry()`, `end()`, `get_file_path()`, `get_file_url()`
- File naming: `{base-name}-{YYYY-MM-DD}-{hash}.csv` using `wp_hash()`
- Supports `woocommerce_product_feed_time` filter for testability

**Data Format:**
- All entry values must be scalars (`string`, `int`, `float`, `bool`) or `null`
- Booleans converted to `"true"`/`"false"` strings (Stripe spec requirement)
- Null values converted to empty strings
- Arrays/objects must be pre-formatted as strings by caller
- Full validation with descriptive error messages

**Performance:**
- O(1) memory usage - constant regardless of catalog size
- Linear disk I/O - no seeks, just streaming writes

**Security:**
- Files stored in `wp-content/uploads/stripe-agentic-commerce/product-feeds/`
- Automatic `.htaccess` and `index.html` creation via `FilesystemUtil::mkdir_p_not_indexable()`
- Prevents directory listing, files accessible via direct URL only

- **Tests**: 21 PHPUnit tests with 40 assertions covering all critical paths
- **PHPStan**: Level 8 passes with zero errors (type safety enforced)
- **Version gating**: Feature only loads with WooCommerce 10.5.0+ (via `interface_exists()` check)
- **Error handling**: Automatic cleanup on exceptions via destructor
- **Logging**: All operations logged via `WC_Stripe_Logger` for debugging

## Testing instructions

**Prerequisites:**
- WooCommerce 10.5.0-beta.1 or later must be installed. For unit tests, modify the `bin/install-wp-tests.sh` script to install WooCommerce beta.

Not much to manually test at the time. The next PR (mapping products) should provide a much better picture.

---

- [x] Covered with tests (21 PHPUnit tests, 40 assertions)
- [x] Tested on mobile (does not apply - backend only)

### Changelog entry

- [x] Changelog entry added: "Add - Add the base CSV feed for agentic commerce"

**Post merge**

- [ ] Added testing instructions to the Release Testing Instructions wiki page (does not apply - internal infrastructure for unreleased feature)
- [ ] Added the needs docs label (does not apply - comprehensive README included in PR)
- [ ] Included this PR in the Release Thread scope (pending - depends on Agentic Commerce release timeline)


includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+	/**
+	 * Get public URL to finalized feed.
+	 *
+	 * Moves file from temp directory to uploads directory if needed.
+	 *
+	 * @since 10.4.0
+	 * @throws Exception If the feed file cannot be moved to the upload directory.
+	 * @return string|null Public URL, or null if not finalized.
+	 */
+	public function get_file_url(): ?string {
```
ebinnion, line 404: For what it's worth, this feels a bit off and threw off my agent multiple times. I understand that it's benign though.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+		return $this->file_path;
+	}
+
+	/**
+	 * Get public URL to finalized feed.
+	 *
+	 * Moves file from temp directory to uploads directory if needed.
+	 *
+	 * @since 10.4.0
+	 * @throws Exception If the feed file cannot be moved to the upload directory.
```
ebinnion: Can you explain this docblock comment? I'm not sure how this has anything to do with moving the file to the upload directory.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+			throw new Exception(
+				sprintf(
+					/* translators: %s: directory path */
+					__( 'Unable to create feed directory: %s', 'woocommerce-gateway-stripe' ),
+					$directory_path
+				)
+			);
+		}
+
+		$directory_url = $upload_dir['baseurl'] . '/stripe-agentic-commerce/product-feeds/';
```
ebinnion: This URL doesn't seem to match with the path above on line 133.

ebinnion: Perhaps it doesn't matter if we always send the CSV as mentioned elsewhere in the PR. But, it still reads weird. :)

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+	}
+
+	/**
+	 * Clean up temporary files and resources.
+	 *
+	 * Called on errors or object destruction.
+	 *
+	 * @since 10.4.0
+	 */
+	private function cleanup(): void {
```
ebinnion, line 415: Does this actually get called on destruct as the docblock says and in the readme?

ebinnion (approved):
I left a few comments that are non-blocking, especially given that this is scaffolding, e2e tests pass, and a smoke test locally works.

There is actually one failing e2e test, but it should be fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5016.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: Not at all. We don't want to automagically delete the feed when the object is destructed. I removed the misleading comment in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946/commits/4a70b1c2c5cbcbf5d515b43f96e8b122301450a5.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: Claude was following the existing JSON feed as an example. For that, there's a pattern that was established in https://github.com/woocommerce/woocommerce-product-feed-for-openai/pull/34 for VIP sites. We can't generate the file in `wp-content/uploads`, as uploads are treated in a completely separate way there.

As you can see in the description of the product feed PR, the logic is that the file gets generated in a temporary folder, and it can be pushed from there. However, the moment the file needs to have a URL, we need to move it to `uploads`.

The method is implemented here because of the interface that we already have. Right now, we wouldn't provide a URL for the file, but once it comes to an admin overview/status, I believe we will likely use the URL and apply the same logic here.

I've updated the docblock to properly describe the exception.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: See the comment in [the other thread](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946#discussion_r2803359460) for why we have the method in the first place.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: It does matter, it just went under the radar because the URL of the feed is not used yet. Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946/commits/ff1293305328e40e98463baa29a39d0eae9c272a

phpstan-stubs/woocommerce-product-feed.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+namespace Automattic\WooCommerce\Internal\ProductFeed\Feed;
+
+/**
+ * Feed Interface.
+ *
+ * @since 10.5.0
+ */
+interface FeedInterface {
```
daledupreez, line 18: Rather than replicating the code, can we bump the `php-stubs/woocommerce-stubs` Composer dependency to pick up the new stub from 10.5?

tests/phpunit/WC_Stripe_Agentic_Commerce_Csv_Feed_Test.php
```diff
@@ -0,0 +1,468 @@
...
+		$feed->start();
+		$feed->add_entry( [ '1', 'Description with "quotes" and, commas' ] );
+		$feed->add_entry( [ '2', "Line with\nnewline" ] );
+		$feed->end();
+
+		$file_path = $feed->get_file_path();
+		$content   = file_get_contents( $file_path );
+
+		// CSV should properly escape quotes by doubling them.
+		$this->assertStringContainsString( '""quotes""', $content );
```
daledupreez, line 180: We're missing the checks for the comma and the newline.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+
+	/**
+	 * Set CSV column headers.
+	 *
+	 * @since 10.5.0
+	 * @param array $headers CSV column headers.
+	 * @return self
+	 */
+	public function set_columns( array $headers ): self {
+		$this->headers = $headers;
```
daledupreez, line 118: Would it be worth validating the specified header values?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+			// Generate filename using WooCommerce pattern.
+			/**
+			 * Allows the current time to be overridden before a feed is stored.
+			 *
+			 * @param int           $time The current time.
+			 * @param FeedInterface $feed The feed instance.
+			 * @return int The current time.
+			 * @since 10.5.0
+			 */
+			$current_time    = apply_filters( 'woocommerce_product_feed_time', time(), $this );
```
daledupreez: Not blocking, but it feels like this should not be something we replicate in plugin code.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+
+			// Write to file (fputcsv handles escaping automatically).
+			// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_fputcsv
+			if ( false === fputcsv( $this->file_handle, $sanitized_entry, ',', '"', '\\' ) ) {
+				throw new Exception( __( 'Failed to write entry to CSV.', 'woocommerce-gateway-stripe' ) );
+			}
+
+			++$this->entry_count;
+		} catch ( Exception $e ) {
+			WC_Stripe_Logger::error( 'Error adding entry to feed: ' . $e->getMessage() );
```
daledupreez, line 286: General thought: it may be worth considering helper methods for logging so we include more context. As things stand, it is not apparent which feed encountered an error, or what parameters were used, or any other state. If we use a central helper for that, we can include more context by default. That may be something for a follow-up rather than this iteration, though.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+
+			// Handle boolean values - Stripe spec: must be literal "true" or "false".
+			// Without this, PHP's fputcsv would convert true->1 and false->0.
+			if ( is_bool( $value ) ) {
+				$sanitized[] = $value ? 'true' : 'false';
+				continue;
+			}
+
+			// Reject arrays and objects - caller must format these.
+			if ( is_array( $value ) || is_object( $value ) ) {
```
daledupreez: Might it be easier to check if we don't have a scalar?
```suggestion
			if ( ! is_scalar( $value ) ) {
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -0,0 +1,420 @@
...
+	 * @since 10.5.0
+	 * @param array $entry Raw entry data (must contain only scalar values or null).
+	 * @throws Exception If entry contains arrays or objects.
+	 * @return array Sanitized entry data.
+	 */
+	private function sanitize_entry( array $entry ) {
+		$sanitized = [];
+
+		foreach ( $entry as $index => $value ) {
+			// Handle null values - Stripe spec: leave blank for optional fields.
```
daledupreez: Idea: rather than a loop with `continue` statements, it might be worth implementing the body of the loop as a function that uses early returns. That would change this code to something like:
```php
		foreach ( $entry as $index => $value ) {
			$sanitized[] = $this->sanitize_value( $value, $index );
		}
```

Definitely not blocking, BTW.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
daledupreez: It may also make it easier to run with different levels of debugging/verbosity.

includes/agentic-commerce/README.md
```diff
@@ -0,0 +1,344 @@
...
+
+## Error Scenarios
+
+### File System Issues
+```php
+try {
+    $feed->start();
+} catch ( Exception $e ) {
+    // Cannot create temp file or uploads directory
+    WC_Stripe_Logger::log( 'Feed start failed: ' . $e->getMessage() );
```
daledupreez: Nit: we are in the process of deprecating the `log()` method in favour of `info|notice|warning|error`.

daledupreez (approved):
The changes look good to me. I have a bunch of non-blocking feedback and suggestions, but none of it is major.

phpstan-stubs/woocommerce-product-feed.php
RadoslavGeorgiev: I might have done something wrong... While incorporating the Product Feed classes into core for the POS catalog, I put them in the `Internal` namespace, both because I was thinking of only using them for the POS catalog at the time, and because I thought `Internal` was for stuff that non-a12s shouldn't use. I know of pcShBQ-3IO-p2, yes 😓, but it was a few weeks too late.

Because of that, the classes are neither in the 10.5.0 nor 10.5.1 core stubs.

tests/phpunit/WC_Stripe_Agentic_Commerce_Csv_Feed_Test.php
RadoslavGeorgiev: Added those tests in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946/commits/e47273700915da87c466ad4388cdaf39bc81a169

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: Given that all the code will be coming from the integration in later classes, I think we should be good.

I added non-empty checks in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946/commits/be8a196afbeb87d0932e6e1bc9ac91d38b9b1e63, just in case.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: Let's see how debugging goes in future iterations. For now, there shouldn't be too much confusion regarding where the issue comes from, as the particular file is mentioned and it's just the file operations that would trigger logging from within this class.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: Switched in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946/commits/a7568bc1e6408ba18ee6cd1dd129208d57970567

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946/commits/ee4920e6e44c63adc3b9b9a54e4497f50bffc070

includes/agentic-commerce/README.md
RadoslavGeorgiev: Yeah, I had already noticed that and fixed the instances within PHP. The markdown is fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946/commits/1c67904c97d2adbdb761fb8507e7e9338594720f as well.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
RadoslavGeorgiev: We had it in the product feed plugin, and it ended up here. Its purpose was only to allow time to be altered for tests, and the pattern is well established already. However, I removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4946/commits/09d672032c8a5752d725cdfcffb43e5c93c5acf0 as I felt the verification of the file name is not needed anyway.

PR #4959, opened 2026-01-28
Add a check to verify remote_oid exists locally

malithsen:

The pre-push hook was showing a cryptic error when pushing to a branch where the remote has commits that
don't exist locally:

```
fatal: Invalid revision range ad6a3f605c7e4c55480b1ea43f7d9cc309a033b0..822606c33f46c16d3b7f21509264f7895189bd9b
husky - pre-push hook exited with code 128 (error)
error: failed to push some refs to 'github.com:woocommerce/woocommerce-gateway-stripe.git'
```

This happens because the hook tried to run `git diff --name-only "$remote_oid..$local_oid"` when the `remote_oid` isn't in the local repo.

This PR adds a check using `git cat-file -e` to detect when the remote has commits we don't have locally.
When this happens, the hook now fails fast with a more clear message:

```
Error: Remote branch has commits that don't exist locally.
Update your local branch, then try pushing again.
```

## Testing instructions
I think it's sufficient to verify `git cat-file -e` is working as expected.

```
# This should succeed 
git cat-file -e HEAD && echo "exists" || echo "missing"

# This should fail as the SHA doesn't exist
git cat-file -e 0000000000000000000000000000000000000001 2>/dev/null && echo "exists" || echo "missing"
```

For a more comprehensive test:

1. Create a test branch off of this branch and push it.
2. Add a commit directly on GitHub (edit any file via the UI)
3. Locally, create another commit without pulling: `git commit --allow-empty -m "Local"`
4. Try to push: `git push`
5. Verify you see the new error message instead of "Invalid revision range"
6. Delete the remote branch in 1.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -45,10 +45,7 @@ do
 			echo "Update your local branch, then try pushing again."
 			exit 1
 		fi
-	fi
 
-	if [ -n "$range" ]
-	then
 		# Get the changed files.
 		changed_files=$(git diff --name-only "$range")
 		# Get the changed PHP files or directories in the push.
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -36,6 +36,19 @@ do
 
 	if [ -n "$range" ]
 	then
+		# Verify that the remote_oid exists locally. If the remote has commits
+		# we don't have, the push will be rejected by the remote anyway.
+		# Fail fast with a clear message.
+		if ! git cat-file -e "$remote_oid" 2>/dev/null
+		then
+			echo "Error: Remote branch has commits that don't exist locally."
+			echo "Update your local branch, then try pushing again."
+			exit 1
+		fi
+	fi
+
+	if [ -n "$range" ]
+	then
 		# Get the changed files.
 		changed_files=$(git diff --name-only "$range")
 		# Get the changed PHP files or directories in the push.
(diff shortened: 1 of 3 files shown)


wjrosa (approved):
Nice fix! Looks good 👍

.husky/pre-push
```diff
@@ -34,6 +34,19 @@ do
...
+		if ! git cat-file -e "$remote_oid" 2>/dev/null
+		then
+			echo "Error: Remote branch has commits that don't exist locally."
+			echo "Update your local branch, then try pushing again."
+			exit 1
+		fi
+	fi
+
 	if [ -n "$range" ]
 	then
```
daledupreez: Non-blocking nit: We could consolidate the logic into one `if` block. It feels weird to me to have two independent code blocks depending on the same check.
```suggestion
	if [ -n "$range" ]
	then
		# Verify that the remote_oid exists locally. If the remote has commits
		# we don't have, the push will be rejected by the remote anyway.
		# Fail fast with a clear message.
		if ! git cat-file -e "$remote_oid" 2>/dev/null
		then
			echo "Error: Remote branch has commits that don't exist locally."
			echo "Update your local branch, then try pushing again."
			exit 1
		fi

```

daledupreez (approved):
The changes look good to me! Thanks for fixing this!

PR #4960, opened 2026-01-28
Fix block editor incompatibility warning

### LINKED ISSUE #4949: Payment methods show an "Incompatible with block-based checkout" error when editing Checkout page
### Description

Some of the Stripe payment methods show an "Incompatible with block-based checkout" error when editing the Checkout page. All of our payment methods are compatible, so we need to address the issue.

### Steps to reproduce

1. Install the Stripe plugin and connect your account to Stripe using a US account and USD as the currency.
2. Enable all payment methods, including Klarna, Affirm, and Afterpay.
3. Open up our block-based checkout page in the page editor.
4. Select the Payment Options block.
5. Inspect the Block options in the Inspector on the right hand side.
6. See the errors.

### Expected behavior

No warnings should be shown, as the payment methods are compatible with block checkout.

### Screenshots

Image

### Additional context

Originally reported in WOOPMNT-5640
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4913
Related to STRIPE-876

Mayisha:
Fixes STRIPE-913
Fixes #4949 

WooCommerce's block editor compares `globalPaymentMethods`  against `availablePaymentMethods` (JS methods registered via registerPaymentMethod()). If there is any mismatch, then the block editor shows this incompatibility notice.

In `add_gateways` method, we are adding all UPE methods irrespective of whether they are available on the checkout page or not. But in the frontend, we only [send the list ](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L538)that is enabled at checkout.

In this PR, I am filtering out the gateways that are not enabled at checkout from the `add_gateways` method so that both the list aligns.

## Testing instructions
- Connect to an US Stripe account.
- Set store currency to Euro and enable a few Euro supported payment methods like SEPA, iDeal etc.
- Switch store currency to USD enable a few other USD supported payment methods.
- Go to the checkout block editor and select the payment block.
- In `develop` notice that there are 'Incompatible with block-based checkout' beside some methods.
- In this branch, those incompatible methods should be hidden and for the displayed methods there should be no incompatibility notice.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,7 +16,7 @@
 * Fix - Add validation to prevent fatal error when setting default payment token if token doesn't exist
 * Fix - Validate order object before accessing methods in my account orders actions to prevent fatal errors
 * Dev - Use WC_STRIPE_PLUGIN_PATH constant instead of __DIR__ for more reliable file path resolution
-* Dev - Automate release note creation PR
+* Fix - Prevent incompatibility warnings for payment methods in block editor
 
 = 10.3.1 - 2026-01-15 =
 * Fix - Fatal error when using express payment method with certain addresses
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,6 +16,7 @@
 * Fix - Add validation to prevent fatal error when setting default payment token if token doesn't exist
 * Fix - Validate order object before accessing methods in my account orders actions to prevent fatal errors
 * Dev - Use WC_STRIPE_PLUGIN_PATH constant instead of __DIR__ for more reliable file path resolution
+* Dev - Automate release note creation PR
 * Fix - Prevent incompatibility warnings for payment methods in block editor
 
 = 10.3.1 - 2026-01-15 =
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -524,7 +524,7 @@ class WC_Stripe {
 
 					if ( $method instanceof WC_Stripe_UPE_Payment_Method ) {
 						$method_id = $method->get_id();
-						if ( isset( $upe_payment_methods[ $method_id ] ) && is_object( $upe_payment_methods[ $method_id ] ) ) {
+						if ( isset( $upe_payment_methods[ $method_id ] ) && $upe_payment_methods[ $method_id ] instanceof WC_Stripe_UPE_Payment_Method ) {
 							if ( ! $upe_payment_methods[ $method_id ]->is_enabled_at_checkout() ) {
 								return false;
 							}
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -507,16 +507,30 @@ class WC_Stripe {
 
 		$methods = array_merge( $methods, $upe_payment_methods );
 
-		// When we are in an admin context, filter out Link and Amazon Pay, as they are only available as
-		// express checkout methods, and including them in the list results in warnings about block support
+		// When we are in an admin context,
+		// 1. Filter out Link and Amazon Pay, as they are only available as express checkout methods,
+		// and including them in the list results in warnings about block support
 		// when viewing the Express Checkout block in the editor for the cart and checkout pages.
+		// 2. Filter out UPE payment methods that are not enabled at checkout, as they are not available in the checkout block
+		// and including them in the list results in warnings about block support
+		// when viewing the payment methods block in the editor for the cart and checkout pages.
 		if ( is_admin() ) {
 			$methods = array_filter(
 				$methods,
-				function ( $method ) {
+				function ( $method ) use ( $upe_payment_methods ) {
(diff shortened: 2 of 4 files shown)


includes/class-wc-stripe.php
```diff
@@ -507,16 +507,30 @@ public function add_gateways( $methods ) {
...
 						return false;
 					}
+
+					if ( $method instanceof WC_Stripe_UPE_Payment_Method ) {
+						$method_id = $method->get_id();
+						if ( isset( $upe_payment_methods[ $method_id ] ) && $upe_payment_methods[ $method_id ] instanceof WC_Stripe_UPE_Payment_Method ) {
+							if ( ! $upe_payment_methods[ $method_id ]->is_enabled_at_checkout() ) {
+								return false;
+							}
+						}
```
daledupreez: Why do we need to reach into `$upe_payment_methods` if we already have a UPE payment method instance?
```suggestion
						if ( ! $method->is_enabled_at_checkout() ) {
							return false;
						}
```

changelog.txt
```diff
@@ -16,7 +16,7 @@
 * Fix - Add validation to prevent fatal error when setting default payment token if token doesn't exist
 * Fix - Validate order object before accessing methods in my account orders actions to prevent fatal errors
 * Dev - Use WC_STRIPE_PLUGIN_PATH constant instead of __DIR__ for more reliable file path resolution
-* Dev - Automate release note creation PR
```
daledupreez, line 22: I think we should keep the previous entry here.
```suggestion
* Dev - Automate release note creation PR
* Fix - Prevent incompatibility warnings for payment methods in block editor

File: includes/class-wc-stripe.php (hunk 1 above)
Comment (daledupreez) [reply includes/class-wc-stripe.php]: I also found a case where this fails: when Optimized Checkout is enabled, we only show the main gateway, so we get warnings for all other enabled payment methods.
```suggestion
						// When Optimized Checkout is enabled, we only show the main payment gateway, so remove all other UPE methods.
						if ( $is_oc_enabled ) {
							return false;
						}

						// If the payment method is not enabled at checkout, remove it.
						if ( ! $method->is_enabled_at_checkout() ) {
							return false;
						}
```

daledupreez (changes requested):
The fix doesn't handle OCS correctly, but I think I have a working proposal to address that.

includes/class-wc-stripe.php
Mayisha: Addressed both of the comments.

Mayisha:
Fixing tests....

includes/class-wc-stripe.php
```diff
@@ -498,25 +498,44 @@ public function plugin_row_meta( $links, $file ) {
 	 * @version 5.6.0
 	 */
 	public function add_gateways( $methods ) {
-		$main_gateway = $this->get_main_stripe_gateway();
-		$methods[]    = $main_gateway;
+		$main_gateway  = $this->get_main_stripe_gateway();
+		$methods[]     = $main_gateway;
+		$is_oc_enabled = $main_gateway->get_option( 'optimized_checkout_element', 'no' );
```
daledupreez: Nit: the naming makes me think the value will be a boolean rather than `'yes'|'no'`. It may be worth making it a boolean so it matches the name.
```suggestion
		$is_oc_enabled = 'yes' === $main_gateway->get_option( 'optimized_checkout_element', 'no' );
```

daledupreez (commented):
The main code changes are looking good and testing well. We just need to get the PHP tests working, and I think we should be in good shape.

There were also e2e test failures during the first run, but those looked to be due to `apt` connectivity issues during the system setup rather than anything in the tests, but those all passed on a re-run.

tests/phpunit/WC_Stripe_Test.php
```diff
@@ -421,10 +425,12 @@ public function provide_test_add_gateways(): array {
 		$klarna_gateway = $this->getMockBuilder( \WC_Stripe_UPE_Payment_Method_Klarna::class )
 			->disableOriginalConstructor()
 			->getMock();
+		$klarna_gateway->method( 'is_enabled_at_checkout' )->willReturn( true );
 
 		$afterpay_clearpay_gateway = $this->getMockBuilder( \WC_Stripe_UPE_Payment_Method_Afterpay_Clearpay::class )
 			->disableOriginalConstructor()
 			->getMock();
+		$afterpay_clearpay_gateway->method( 'is_enabled_at_checkout' )->willReturn( true );
```
malithsen, line 433: nit: Should we also include a test that returns `is_enabled_at_checkout` as `false`, since that seems to be a path that's not currently covered?

malithsen (approved):
This looks good to me and tests as described.

daledupreez (approved):
The updates look good to me! And thanks for taking on all the test updates, too!

tests/phpunit/WC_Stripe_Test.php
Mayisha: Added ✌️

PR #4964, opened 2026-01-29
Fix credit card input from being cleared in OCS when save payment method checkbox is selected

### LINKED ISSUE #4629: Credit card fields not persisting when optimized checkout suite enabled in classic checkout
### Description

In classic checkout, updating the checkout form via something like the save payment method checkbox clears credit card input when the optimized checkout suite is enabled:

### Steps to reproduce

1. As admin, setup a classic checkout page
2. Ensure optimized checkout suite is enabled via WooCommerce > Settings > Payments > Manage (Stripe) > Settings > Advanced Settings
3. Also ensure payment via saved cards is enabled
4. As a customer, add any non-subscription item to cart and go to (classic) checkout
5. Select Stripe as the payment method and input some credit card information
6. After filling in CC info, select the save payment method checkbox
7. Observe the credit card details have been cleared

### Screenshots

https://github.com/user-attachments/assets/91bc76e0-9a06-4294-a261-46ad6d885ff5

Mayisha:
Fixes STRIPE-690
Fixes #4629 

When the save payment method checkbox is changed, we unmount and remount the payment element. This is done so that we can show the correct list of payments that support saving the method and hide that do not support it. As a side effect of this approach, the payment element gets refreshed when the checkbox is selected. If shopper adds any card info before selecting the checkbox, those info also disappears.

In this PR, I have removed the unmount/re-mount approach and using [elements.update](https://docs.stripe.com/js/elements_object/update) method to update the `setupFutureUsage` parameter of the element only so that Stripe correctly calculates and renders the supported method for saving. 

## Testing instructions
- As an admin, set up a classic checkout page if you already do not have one.
- Enable optimized checkout suite in **WooCommerce > Settings > Payments > Manage (Stripe) > Settings > Advanced Settings**.
- Enable payment via saved cards.
- As a shopper, add a simple product to your cart and go to the classic checkout page.
- Select Stripe as the payment method and input some credit card information
- After filling in credit card info, select the save payment method checkbox
- In `develop`, observe that the credit card details have been cleared and the Stripe payment element is refreshed.
- In this branch, verify that the credit card information still persists and the Stripe element is not refreshed.
    - Also verify that there is no regression related to the list of payment methods displayed based on setup future usage capability. For example, if you have card, CashApp, WeChat and AliPay available on the checkout page initially and you select the save payment method checkbox, it should hide WeChat and AliPay, keeping card and CashApp still on the page. 
    - Verify that there is no other regressions and OCS checkout works with different methods.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -17,6 +17,7 @@
 * Fix - Validate order object before accessing methods in my account orders actions to prevent fatal errors
 * Dev - Use WC_STRIPE_PLUGIN_PATH constant instead of __DIR__ for more reliable file path resolution
 * Dev - Automate release note creation PR
+* Fix - Prevent credit card input from being cleared when toggling save payment method checkbox in classic checkout with Optimized Checkout enabled
 
 = 10.3.1 - 2026-01-15 =
 * Fix - Fatal error when using express payment method with certain addresses
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,7 +20,7 @@
 * Dev - Automate release note creation PR
 * Dev - Introduce a feature flag for the Stripe checkout sessions feature
 * Dev - Improve the pre-push hook
-* Fix - Prevent credit card input from being cleared when toggling save payment method checkbox in classic checkout with Optimized Checkout enabled
+* Fix - Keep payment method details when toggling save payment method checkbox in classic checkout with Optimized Checkout enabled
 * Fix - Better error handling when token creation fails
 * Tweak - Improve PHPDoc for payment token code
 * Tweak - Update PHPDoc for email notification classes
--- a/client/classic/upe/payment-processing.js
+++ b/client/classic/upe/payment-processing.js
@@ -396,6 +396,30 @@ export async function mountStripePaymentElement( api, domElement ) {
 }
 
 /**
+ * Gets the mounted UPE element for a payment method type.
+ *
+ * @param {string} paymentMethodType The payment method type.
+ * @return {Object|null} The UPE element component object or null if not found.
+ */
+export function getMountedUPEComponent( paymentMethodType ) {
+	if ( ! gatewayUPEComponents[ paymentMethodType ] ) {
+		return null;
+	}
+
+	const component = gatewayUPEComponents[ paymentMethodType ];
+	const domElement = document.querySelector(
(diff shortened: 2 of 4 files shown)


client/classic/upe/payment-processing.js
```diff
@@ -395,6 +395,30 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+	const domElement = document.querySelector(
+		`.wc-stripe-upe-element[data-payment-method-type="${ paymentMethodType }"]`
+	);
+
+	// Only return if the Elements object exists and is mounted.
+	if ( component.elements && domElement && domElement.children.length > 0 ) {
+		return component;
+	}
+
+	return null;
```
daledupreez, line 424: Nit: It feels like we could use early returns to do less work here. We are always running `document.querySelector()` even in cases where we may not need to.
```suggestion
	// Only return if the Elements object exists and is mounted.
	if ( ! component.elements ) {
		return null;
	}
	
	const domElement = document.querySelector(
		`.wc-stripe-upe-element[data-payment-method-type="${ paymentMethodType }"]`
	);

	if ( domElement && domElement.children.length > 0 ) {
		return component;
	}

	return null;
```

changelog.txt
```diff
@@ -18,6 +18,7 @@
 * Dev - Use WC_STRIPE_PLUGIN_PATH constant instead of __DIR__ for more reliable file path resolution
 * Dev - Automate release note creation PR
 * Dev - Improve the pre-push hook
+* Fix - Prevent credit card input from being cleared when toggling save payment method checkbox in classic checkout with Optimized Checkout enabled
```
daledupreez: Nit: This is really long. I think we can simplify it a bit. Also, I checked that it affects other payment methods that have input fields, not just cards.
```suggestion
* Fix - Keep payment method details when toggling save payment method checkbox in classic checkout with Optimized Checkout enabled
```

daledupreez (approved):
Nice catch on this one! This is working well for me. 🚀

wjrosa (approved):
Thanks for fixing this issue! Looks good to me 👍

client/classic/upe/payment-processing.js
Mayisha: Done.

PR #4968, opened 2026-01-30
Endpoint to create Checkout Sessions

### LINKED ISSUE #4969: Checkout Sessions: Endpoint to create sessions
We need a new public (shopper-facing) AJAX endpoint that creates Checkout Sessions objects and returns the client secret for them. This secret is required to build the payment element on the frontend (both blocks and shortcode checkout).

wjrosa:
Fixes STRIPE-932
Fixes #4969
Doc https://docs.stripe.com/api/checkout/sessions/create

This PR introduces a new public (shopper-facing) AJAX endpoint that creates Checkout Sessions objects and returns the client secret for them. This secret is required to build the payment element on the frontend (both blocks and shortcode checkout).

The endpoint accepts the `payment_method_type` parameter to build the `payment_method_types` property. It builds the product information using the current cart session.

We are using `custom` as `ui_mode` so we can render the payment element following the existing standards for payment intents on the frontend. We are hardcoding `payment` as the `mode` since saving of payment methods won't be supported for now.

No payment intent data is currently being provided.

## Testing instructions

It is a bit hard to test this new endpoint manually because it is not yet used by the frontend, and it would require adding session information to the cURL request. For now, a code review should be enough. Also, check if the tests are passing.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-checkout-sessions-controller.php
+++ b/includes/class-wc-stripe-checkout-sessions-controller.php
@@ -1,0 +1,78 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * WC_Stripe_Checkout_Sessions_Controller class.
+ */
+class WC_Stripe_Checkout_Sessions_Controller {
+	/**
+	 * Initialize hooks.
+	 *
+	 * @return void
+	 */
+	public function init_hooks() {
+		add_action( 'wc_ajax_wc_stripe_create_checkout_session', [ $this, 'create_checkout_session' ] );
+	}
+
+	public function create_checkout_session() {
+		$is_nonce_valid = check_ajax_referer( 'wc_stripe_create_checkout_session_nonce', false, false );
+		if ( ! $is_nonce_valid ) {
+			throw new Exception( __( "We're not able to process this payment. Please refresh the page and try again.", 'woocommerce-gateway-stripe' ) );
+		}
+
+		if ( ! defined( 'WOOCOMMERCE_CART' ) ) {
+			define( 'WOOCOMMERCE_CART', true );
+		}
+
+		$payment_method_type     = isset( $_POST['payment_method_type'] ) ? wc_clean( wp_unslash( $_POST['payment_method_type'] ) ) : '';
+		$enabled_payment_methods = $payment_method_type ? [ $payment_method_type ] : [];
+
+		WC()->cart->calculate_totals();
+
+		$user     = wp_get_current_user();
+		$customer = new WC_Stripe_Customer( $user->ID );
+		$customer->update_or_create_customer();
+
+		// TODO: fix issue when customer does not have a billing address.
+		// Critical Uncaught WC_Stripe_Exception: missing_required_customer_field: name in includes/class-wc-stripe-customer.php:265
+
+		$currency   = get_woocommerce_currency();
+		$line_items = [];
+		foreach ( WC()->cart->get_cart() as $cart_item ) {
+			$product_name   = $cart_item['data']->get_name();
+			$line_items[] = [
+				'price_data' => [
+					'currency' => strtolower( $currency ),
+					'product_data' => [
+						'name' => $product_name,
+					],
+					'unit_amount' => WC_Stripe_Helper::get_stripe_amount( $cart_item['line_subtotal'] / $cart_item['quantity'], $currency ),
+				],
+				'quantity' => $cart_item['quantity'],
+			];
+		}
+
+		$request = [
+			'ui_mode'              => 'custom',
+			'customer'             => $customer->get_id(),
+			'line_items'           => $line_items,
+			'payment_method_types' => $enabled_payment_methods,
+			'payment_intent_data'  => [],
+			'mode'                 => 'payment',
+			'adaptive_pricing'     => [
+				'enabled' => 'true',
+			],
+		];
+
+		$checkout_session = WC_Stripe_API::request( $request, 'checkout/sessions' );
+
+		if ( ! empty( $checkout_session->error ) ) {
+			throw new Exception( $checkout_session->error->message );
+		}
+
+		wp_send_json( [ 'client_secret' => $checkout_session->client_secret ] );
+	}
+}
--- a/includes/class-wc-stripe-checkout-sessions-controller.php
+++ b/includes/class-wc-stripe-checkout-sessions-controller.php
@@ -13,15 +13,22 @@ class WC_Stripe_Checkout_Sessions_Controller {
 	 *
 	 * @return void
 	 */
-	public function init_hooks() {
+	public function init_hooks(): void {
 		add_action( 'wc_ajax_wc_stripe_create_checkout_session', [ $this, 'create_checkout_session' ] );
 	}
 
-	public function create_checkout_session() {
-		$is_nonce_valid = check_ajax_referer( 'wc_stripe_create_checkout_session_nonce', false, false );
-		if ( ! $is_nonce_valid ) {
-			throw new Exception( __( "We're not able to process this payment. Please refresh the page and try again.", 'woocommerce-gateway-stripe' ) );
-		}
+	/**
+	 * Create a Stripe Checkout Session and return the client secret.
+	 *
+	 * @return void
+	 * @throws WC_Stripe_Exception When unable to create the Checkout Session.
+	 */
+	public function create_checkout_session(): void {
+		// TODO: verify nonce
+		// $is_nonce_valid = check_ajax_referer( 'wc_stripe_create_checkout_session_nonce', false, false );
+		// if ( ! $is_nonce_valid ) {
+		// throw new Exception( __( "We're not able to process this payment. Please refresh the page and try again.", 'woocommerce-gateway-stripe' ) );
+		// }
 
 		if ( ! defined( 'WOOCOMMERCE_CART' ) ) {
 			define( 'WOOCOMMERCE_CART', true );
--- a/includes/class-wc-stripe-checkout-sessions-controller.php
+++ b/includes/class-wc-stripe-checkout-sessions-controller.php
@@ -36,11 +36,12 @@ class WC_Stripe_Checkout_Sessions_Controller {
 		$payment_method_type     = isset( $_POST['payment_method_type'] ) ? wc_clean( wp_unslash( $_POST['payment_method_type'] ) ) : '';
 		$enabled_payment_methods = $payment_method_type ? [ $payment_method_type ] : [];
 
-		$customer = new WC_Stripe_Customer( wp_get_current_user()->ID );
-		$customer->maybe_create_customer();
-
-		// TODO: fix issue when customer does not have a billing address.
-		// Critical Uncaught WC_Stripe_Exception: missing_required_customer_field: name in includes/class-wc-stripe-customer.php:265
+		try {
+			$customer = new WC_Stripe_Customer( wp_get_current_user()->ID );
+			$customer->maybe_create_customer();
+		} catch ( Exception $e ) {
+			throw new Exception( __( 'Unable to create or retrieve Stripe customer.', 'woocommerce-gateway-stripe' ) );
+		}
 
 		if ( is_null( WC()->cart ) || WC()->cart->is_empty() ) {
 			throw new Exception( __( 'Your cart is currently empty.', 'woocommerce-gateway-stripe' ) );
@@ -67,7 +68,7 @@ class WC_Stripe_Checkout_Sessions_Controller {
 			'customer'             => $customer->get_id(),
 			'line_items'           => $line_items,
 			'payment_method_types' => $enabled_payment_methods,
-			'payment_intent_data'  => [],
+			'payment_intent_data'  => [], // @todo Pass additional data if needed.
 			'mode'                 => 'payment',
 			'adaptive_pricing'     => [
 				'enabled' => 'true',
--- a/includes/class-wc-stripe-checkout-sessions-controller.php
+++ b/includes/class-wc-stripe-checkout-sessions-controller.php
@@ -24,11 +24,10 @@ class WC_Stripe_Checkout_Sessions_Controller {
 	 * @throws WC_Stripe_Exception When unable to create the Checkout Session.
 	 */
 	public function create_checkout_session(): void {
-		// TODO: verify nonce
-		// $is_nonce_valid = check_ajax_referer( 'wc_stripe_create_checkout_session_nonce', false, false );
-		// if ( ! $is_nonce_valid ) {
-		// throw new Exception( __( "We're not able to process this payment. Please refresh the page and try again.", 'woocommerce-gateway-stripe' ) );
-		// }
+		$is_nonce_valid = check_ajax_referer( 'wc_stripe_create_checkout_session_nonce', false, false );
+		if ( ! $is_nonce_valid ) {
+			throw new Exception( __( "We're not able to process this payment. Please refresh the page and try again.", 'woocommerce-gateway-stripe' ) );
+		}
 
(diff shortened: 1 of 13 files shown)


tests/phpunit/Helpers/Ajax_Test_Helper.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+class Ajax_Test_Helper {
+	/**
+	 * Initialize hooks.
+	 *
+	 * @return void
+	 */
+	public static function init_hooks(): void {
+		add_filter( 'wp_doing_ajax', '__return_true' );
+		add_filter( 'wp_die_ajax_handler', [ Ajax_Test_Helper::class, 'wp_ajax_halt_handler_filter' ] );
```
Mayisha:
```suggestion
		add_filter( 'wp_die_ajax_handler', [ self::class, 'wp_ajax_halt_handler_filter' ] );
```

Alternatively, we could also use 'self' here.

tests/phpunit/Helpers/Ajax_Test_Helper.php
```diff
@@ -0,0 +1,41 @@
...
+		remove_filter( 'wp_die_ajax_handler', [ Ajax_Test_Helper::class, 'wp_ajax_halt_handler_filter' ] );
+	}
+
+	/**
+	 * Filter to return a custom handler for AJAX requests.
+	 *
+	 * @return callable The custom handler function.
+	 */
+	public static function wp_ajax_halt_handler_filter(): callable {
+		return [ Ajax_Test_Helper::class, 'wp_ajax_print_handler' ];
```
Mayisha: We could use 'self' here as well.

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,90 @@
...
+	/**
+	 * Create a Stripe Checkout Session and return the client secret.
+	 *
+	 * @return void
+	 * @throws WC_Stripe_Exception When unable to create the Checkout Session.
+	 */
+	public function create_checkout_session(): void {
+		$is_nonce_valid = check_ajax_referer( 'wc_stripe_create_checkout_session_nonce', false, false );
+		if ( ! $is_nonce_valid ) {
+			throw new Exception( __( "We're not able to process this payment. Please refresh the page and try again.", 'woocommerce-gateway-stripe' ) );
```
Mayisha: This exception and a couple other exceptions are not being caught. As suggested in my previous comment, we should catch them and use 'wp_send_json_error' for response.

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,90 @@
...
+	 */
+	public function init_hooks(): void {
+		add_action( 'wc_ajax_wc_stripe_create_checkout_session', [ $this, 'create_checkout_session' ] );
+	}
+
+	/**
+	 * Create a Stripe Checkout Session and return the client secret.
+	 *
+	 * @return void
+	 * @throws WC_Stripe_Exception When unable to create the Checkout Session.
```
Mayisha: I think we should not throw error from this method. Instead we should wrap the whole thing in a try catch block, catch the individual thrown errors and return correct error in the response with `wp_send_json_error`. We can add logs to the catch block and log the exact error there for future debugging.

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,90 @@
...
+				'price_data' => [
+					'currency' => strtolower( $currency ),
+					'product_data' => [
+						'name' => $product_name,
+					],
+					'unit_amount' => WC_Stripe_Helper::get_stripe_amount( $cart_item['line_subtotal'] / $cart_item['quantity'], $currency ),
+				],
+				'quantity' => $cart_item['quantity'],
+			];
+		}
```
Mayisha: QQ: We are not considering the tax, shipping cost, discount etc here. Do we not need the actual total amount to create the session correctly? Or does the total amount get updated later with the correct value?

Mayisha: If needed, we can extract the [build_display_items](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/df8fa5ca63cdef0dc51cc5e88966ef1148057a17/includes/payment-methods/class-wc-stripe-express-checkout-helper.php#L1495) method to the helper class and reuse it from there.

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,90 @@
...
+
+		if ( ! defined( 'WOOCOMMERCE_CART' ) ) {
+			define( 'WOOCOMMERCE_CART', true );
+		}
+
+		$payment_method_type     = isset( $_POST['payment_method_type'] ) ? wc_clean( wp_unslash( $_POST['payment_method_type'] ) ) : '';
+		$enabled_payment_methods = $payment_method_type ? [ $payment_method_type ] : [];
+
+		try {
+			$customer = new WC_Stripe_Customer( wp_get_current_user()->ID );
```
Mayisha: QQ: Should we check first if `wp_get_current_user` exists? What happens for a guest checkout?

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,90 @@
...
+
+		if ( ! empty( $checkout_session->error ) ) {
+			throw new Exception( $checkout_session->error->message );
+		}
+
+		if ( empty( $checkout_session->client_secret ) ) {
+			throw new Exception( __( 'Unable to create Stripe Checkout Session.', 'woocommerce-gateway-stripe' ) );
+		}
+
+		wp_send_json( [ 'client_secret' => $checkout_session->client_secret ] );
```
Mayisha: Nit
```suggestion
		wp_send_json_success( [ 'client_secret' => $checkout_session->client_secret ] );
```

Mayisha (commented):
Thanks @wjrosa for adding the endpoint. I have posted a few comments from code review.

tests/phpunit/Helpers/Ajax_Test_Helper.php
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4968/commits/c644710b388e7d9b5808b0549d378c00435bd3d5

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Good idea. Implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4968/commits/4c4c2ec73adc8e5799e248857464c79471dc8e06

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4968/commits/4c4c2ec73adc8e5799e248857464c79471dc8e06

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: `wp_get_current_user()->ID` will be `0` for guest users. But it is present. When this happens, `maybe_create_customer` will create a new customer. It should work properly, but it needs some testing once we have all the consumer code in place. I will add a TODO here.

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Good idea! Did exactly that in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4968/commits/2292ab3bdb43023ec7386b3d21b82400de5fd6c9. Also included unit tests for it

Mayisha (approved):
Looks good to me. 
Confirmed that express checkout (Google Pay) is working correctly on product and checkout pages.

includes/class-wc-stripe-checkout-sessions-controller.php
daledupreez: We need to remove the `@throws` PHPDoc after updating the code.
```suggestion
```

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,100 @@
...
+			if ( ! defined( 'WOOCOMMERCE_CART' ) ) {
+				define( 'WOOCOMMERCE_CART', true );
+			}
+
+			$payment_method_type     = isset( $_POST['payment_method_type'] ) ? wc_clean( wp_unslash( $_POST['payment_method_type'] ) ) : '';
+			$enabled_payment_methods = $payment_method_type ? [ $payment_method_type ] : [];
+
+			// TODO: Test guest checkout flow.
+			try {
+				$customer = new WC_Stripe_Customer( wp_get_current_user()->ID );
```
daledupreez: Not blocking: should this code rely on `WC()->customer` instead of the current user?

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,100 @@
...
+				$customer->maybe_create_customer();
+			} catch ( Exception $e ) {
+				throw new Exception( __( 'Unable to create or retrieve Stripe customer.', 'woocommerce-gateway-stripe' ) );
+			}
+
+			if ( is_null( WC()->cart ) || WC()->cart->is_empty() ) {
```
daledupreez: Nit: Rather than specifically checking for `is_null()`, might it make sense to check for any falsey value?
```suggestion
			if ( ! WC()->cart || WC()->cart->is_empty() ) {
```

includes/class-wc-stripe-checkout-sessions-controller.php
daledupreez: Also, might it make sense to call the local variable `$stripe_customer` so it's clearer what type of customer we're dealing with?

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,100 @@
...
+	 * Create a Stripe Checkout Session and return the client secret.
+	 *
+	 * @return void
+	 * @throws WC_Stripe_Exception When unable to create the Checkout Session.
+	 */
+	public function create_checkout_session(): void {
+		try {
+			$is_nonce_valid = check_ajax_referer( 'wc_stripe_create_checkout_session_nonce', false, false );
+			if ( ! $is_nonce_valid ) {
+				throw new Exception( __( "We're not able to process this payment. Please refresh the page and try again.", 'woocommerce-gateway-stripe' ) );
```
daledupreez: Maybe we should use "this request", as this is not in response to a payment attempt.
```suggestion
				throw new Exception( __( "We're not able to process this request. Please refresh the page and try again.", 'woocommerce-gateway-stripe' ) );
```

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -0,0 +1,100 @@
...
+				'mode'                 => 'payment',
+				'adaptive_pricing'     => [
+					'enabled' => 'true',
+				],
+			];
+
+			$checkout_session = WC_Stripe_API::request( $request, 'checkout/sessions' );
+
+			if ( ! empty( $checkout_session->error ) ) {
+				throw new Exception( $checkout_session->error->message );
```
daledupreez: Nit: I think we should make sure this value is not empty before we throw an exception with an empty message.
```suggestion
				$message = empty( $checkout_session->error->message ) ? __( 'Checkout Sessions API returned an error', 'woocommerce-gateway-stripe' ) : $checkout_session->error->message;
				throw new Exception( $message );
```

phpstan-baseline.neon
```diff
@@ -3162,6 +3162,30 @@ parameters:
...
+			identifier: property.notFound
+			count: 1
+			path: includes/class-wc-stripe-helper.php
+
+		-
+			message: '#^Access to an undefined property WC_Cart\:\:\$tax_total\.$#'
+			identifier: property.notFound
+			count: 1
+			path: includes/class-wc-stripe-helper.php
+
```
daledupreez, line 3188: I think we should aim to call current/documented APIs or getters for these values. AFAICT, it looks like we can do the following:
| Existing property | New accessor | Notes |
|------|-----|----------|
| `WC()->cart->dp` | `wc_get_price_decimals()` | We should probably use a local variable to avoid repeated function calls |
| `WC()->cart->shipping_tax_total` | `WC()->cart->get_shipping_tax()` | |
| `WC()->cart->shipping_total` | `WC()->cart->get_shipping_total()` | |
| `WC()->cart->tax_total` | `WC()->cart->get_fee_tax() + WC()->cart->get_cart_contents_tax()` | |

All of the above are from looking at `WC_Legacy_Cart->__get()`.

Though that should be tackled in a dedicated PR.

tests/phpunit/WC_Stripe_Checkout_Sessions_Controller_Test.php
```diff
@@ -0,0 +1,217 @@
...
+					'headers'  => [ 'Content-Type' => 'application/json' ],
+					'body'     => json_encode( (object) [ 'id' => 'cus_123' ] ),
+				];
+			}
+
+			return [
+				'response' => 200,
+				'headers'  => [ 'Content-Type' => 'application/json' ],
+				'body'     => json_encode( $checkout_session_response ),
+			];
```
daledupreez: Nit: it feels like we should only mock the response when we know we're getting an API call for the expected URL.
```suggestion
			// Mock the customer retrieval response.
			if ( strpos( $url, '/v1/customers' ) !== false ) {
				return [
					'response' => 200,
					'headers'  => [ 'Content-Type' => 'application/json' ],
					'body'     => json_encode( (object) [ 'id' => 'cus_123' ] ),
				];
			}

			if ( 'https://api.stripe.com/v1/checkout/sessions' === $url ) {
				return [
					'response' => 200,
					'headers'  => [ 'Content-Type' => 'application/json' ],
					'body'     => json_encode( $checkout_session_response ),
				];
			}
			
			return $return_value;
```

tests/phpunit/WC_Stripe_Checkout_Sessions_Controller_Test.php
```diff
@@ -0,0 +1,217 @@
...
+			$controller->create_checkout_session();
+			$output = ob_get_clean();
+		} catch ( \Exception $e ) {
+			ob_end_clean();
+			throw $e;
+		}
+
+		// Clean up.
+		remove_filter( 'pre_http_request', $test_request, 10, 3 );
+		Ajax_Test_Helper::remove_hooks();
```
daledupreez: Should we handle this cleanup in a `finally` clause to make sure they get run?
```suggestion
		} finally {
			// Clean up.
			remove_filter( 'pre_http_request', $test_request, 10, 3 );
			Ajax_Test_Helper::remove_hooks();
		}
```

daledupreez (approved):
This is looking good to me based on inspection, though I do have a number of small items that we could choose to address.

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Good ideas. Updated both in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4968/commits/9fb0658c903169850ae7a2866d75f384711313a6

phpstan-baseline.neon
wjrosa: Makes sense to me. Not sure about `WC()->cart->tax_total`. We have multiple similar getters for it, like `get_tax_totals`. This one might need some additional investigation

PR #4269, opened 2025-04-25
Amazon Pay ECE: display custom error message for amount mismatch declines

annemirasol:
Fixes STRIPE-377

## Changes proposed in this Pull Request:

When taxes are calculated based on billing address, and the shopper uses express checkout, we don’t have access to their billing address early enough to ensure accurate tax calculation in the payment modal.

For Amazon Pay, any mismatch between the pre-authorized amount (shown in the payment modal) and the final charged amount results in a transaction decline. This security feature does is not present in Apple Pay or Google Pay.

Rather than disabling Amazon Pay for this common tax configuration, we've opted to allow the decline to occur and instead show a more user-friendly error message, encouraging the shopper to choose a different payment method.

## Testing instructions
1. Turn on the feature flag and enable Amazon Pay as a payment method.
2. In WooCommerce > Settings > General, choose `No location by default` for **Default customer location**.
3. Enable taxes and set tax computation to customer billing address.
   - In WooCommerce > Settings > General, check `Enable tax rates and calculations`,
   - In WooCommerce > Settings > Tax > Tax options, choose `Customer billing address` for **Calculate tax based on**
4. In WooCommerce > Settings > Tax > Standard rates, enter different tax rates for your Amazon Pay shipping address, and the Amazon Pay test card billing address (Wisconsin). See screenshot below for my setup.
5. As a shopper, go to a shippable product page.
6. Clear your site cookies to clear previous orders' billing address information.
7. Use Amazon Pay to purchase the product.
   - In `develop`, the error message is a generic `The payment failed.`
   - In this branch, the error message is `Amazon Pay is not compatible for this order. Please try a different payment method.`
 
Sample tax setup:
Screenshot 2025-04-25 at 4 12 20 PM

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [x] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

The changes in this PR are for a still-unreleased payment method.
</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
[...]

PR #4977, opened 2026-02-03
Bump versions of Stripe JS and Stripe API

Mayisha:
Bumped the Stripe JS and Stripe API versions in this PR as groundwork for adaptive pricing implementation. These common changes will work for both classic and block implementation.

## Testing instructions
- Code review.
- Run `npm install` and `npm start`.
- Smoke test a few scenarios in settings, block and classic checkout pages to verify the version updates does not hamper any existing feature (no regressions)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2180,7 +2180,7 @@ public function payment_scripts() {
 		wp_register_style( 'stripe_styles', plugins_url( 'assets/css/stripe-styles.css', WC_STRIPE_MAIN_FILE ), [], WC_STRIPE_VERSION );
 		wp_enqueue_style( 'stripe_styles' );
 
-		wp_register_script( 'stripe', 'https://js.stripe.com/v3/', '', '3.0', true );
+		wp_register_script( 'stripe', 'https://js.stripe.com/clover/stripe.js', '', '3.0', true );
```
daledupreez: Here and below, do we want the `3.0` version to stick around?

package.json
```diff
@@ -102,8 +102,8 @@
   },
   "dependencies": {
     "@automattic/interpolate-components": "^1.2.1",
-    "@stripe/react-stripe-js": "^2.8.0",
-    "@stripe/stripe-js": "^1.36.0",
+    "@stripe/react-stripe-js": "5.4.1",
+    "@stripe/stripe-js": "8.6.0",
```
daledupreez: As a quick reaction, this involves jumping multiple major versions. Have we reviewed the package history and/or release notes to understand what changes were in those major versions?

package.json
```diff
@@ -102,8 +102,8 @@
   },
   "dependencies": {
     "@automattic/interpolate-components": "^1.2.1",
-    "@stripe/react-stripe-js": "^2.8.0",
-    "@stripe/stripe-js": "^1.36.0",
+    "@stripe/react-stripe-js": "5.4.1",
```
wjrosa: I would consider adding the `^` symbol. But just to follow our current standard (personally, I don't see why, since we are locking the versions anyway).

wjrosa (approved):
Looks good to me! I second Dale's comment on the `3.0` query string

package.json
Mayisha: Added.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
Mayisha: Removed the version.

package.json
Mayisha: Found some inconsistency in the UI. 
I have fixed the payment method loading issue in [77926ed](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977/commits/77926edf3c0cf1216890ee2033e8d08d3edec8b9) but found some other styling issues like missing radio or extra margins etc. 
Working on it.

daledupreez:
I see that most of the e2e tests are failing, with only the Blik tests passing, so it feels like we still have some work to do.

Also, I am wondering whether it may make sense to split this PR into two: one PR to update Stripe.js and another PR to update the API version. As things stand, it is going to be tricky to work out which failures stem from the JS update, which ones stem from the API update, and which stem from both.

Mayisha:
@daledupreez I have split the upgrade in two PRs
- https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977 (this one)
- https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4993 (npm package upgrade)

Apparently the E2E tests are looking good in the other PR and the update in this PR seems problematic 😬

Mayisha:
E2E tests are passing after the fix in [5a598bc](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977/commits/5a598bca594eff72a4e1c7da3a262cc8a19b35a8)

changelog.txt
```diff
@@ -11,6 +11,7 @@
 * Tweak - Update PHPDoc for express checkout classes, block support class, and intent controller
 * Tweak - Update PHPDoc for UPE payment method classes
 * Dev - Fix WC beta version resolution in tests
+* Update - Update Stripe API version
```
daledupreez: Nit: I think we should specify which version we updated to.
```suggestion
* Update - Update Stripe API to 2025-09-3.clover and dynamic Javascript to Clover
```

includes/class-wc-stripe-api.php
```diff
@@ -14,7 +14,7 @@ class WC_Stripe_API {
 	 * Stripe API Endpoint
 	 */
 	const ENDPOINT           = 'https://api.stripe.com/v1/';
-	const STRIPE_API_VERSION = '2024-06-20';
+	const STRIPE_API_VERSION = '2025-09-30.clover';
```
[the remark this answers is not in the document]
daledupreez: > Decline codes are more granular; update error handling logic if relying on specific codes

Have we done a deep dive into the details for the related changes? It feels like some of our error handling could be broken or invalidated if the decline codes change. I am especially concerned because none of these issues will be found by happy path testing.

daledupreez (commented):
@Mayisha, on the surface, this PR looks OK, but I want to confirm that we've investigated the decline code changes that CodeRabbit alluded to. Changes in that area would be really unlikely to show up in most of our tests, but would be hugely problematic in production.

daledupreez:
After a quick look at the [Stripe changelog](https://docs.stripe.com/changelog), it looks like we may need to review the following changes:
 - [New error code for account requirements](https://docs.stripe.com/changelog/clover/2025-09-30/business-type-error-code)
 - [Additional states for account review](https://docs.stripe.com/changelog/clover/2025-09-30/expanded-risk-requirement-descriptions)
 - [Decline code changes for some LPMs, including Amazon Pay](https://docs.stripe.com/changelog/clover/2025-09-30/lpm-descriptive-decline-codes)
 - [New error codes for account verification](https://docs.stripe.com/changelog/basil/2025-03-31/adds-requirement-error-codes)
 - [New error codes for account verification)(https://docs.stripe.com/changelog/acacia/2024-09-30/additional-risk-verification-details-connected-accounts)

Most of these changes look like they're unlikely to affect us in a major way, but I think we need to confirm we are not introducing incompatibilities or issues.

Also, it looks like it should be easier to [block specific card brands when using Checkout Sessions](https://docs.stripe.com/changelog/acacia/2025-02-24/checkout_add_brands_blocked), which may be something we expose in future.

Mayisha:
Thanks @daledupreez for going through the changes. I have 

> [New error code for account requirements](https://docs.stripe.com/changelog/clover/2025-09-30/business-type-error-code)

We currently do not show any [account requirements errors](https://docs.stripe.com/api/accounts/object?api-version=2025-09-30.clover#account_object-requirements-errors) in the plugin. We can explore if we want to show the errors in the settings page for UX improvement.

> [Additional states for account review](https://docs.stripe.com/changelog/clover/2025-09-30/expanded-risk-requirement-descriptions)

This update involves the `requirements.errors` and `requirements.currently_due` fields. We are not using `requirements.error` so we are good here. [The current implementation](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-account.php#L173) only checks whether `requirements.currently_due` has any value, so we are good here as well.

> [Decline code changes for some LPMs, including Amazon Pay](https://docs.stripe.com/changelog/clover/2025-09-30/lpm-descriptive-decline-codes)

I have removed the check for `generic_decline` in [6494947](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977/commits/6494947f526367c3ccf9eabbd91bd50a62826ea8) and am checking if the `decline_code` key is present.

> [New error codes for account verification](https://docs.stripe.com/changelog/basil/2025-03-31/adds-requirement-error-codes)
[New error codes for account verification](https://docs.stripe.com/changelog/acacia/2024-09-30/additional-risk-verification-details-connected-accounts)

These are also related to the `requirements.errors` in the account response which we don't use atm.

Additionally, added two additional intent status checks ('processing' and 'requires_capture') in 'get_existing_compatible_payment_intent' method [5f2cf32](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977/commits/5f2cf3245ebd839ea9e84344410f93915dcfb442) as per this changelog https://docs.stripe.com/changelog/clover/2025-09-30/client-secret-reuse

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2407,8 +2407,7 @@ private function process_payment_intent_for_order( WC_Order $order, array $payme
 	private function get_payment_intent_error_message( $payment_intent ) {
 		if ( isset( $payment_intent->error->payment_intent->payment_method_types[0] ) &&
 			'amazon_pay' === $payment_intent->error->payment_intent->payment_method_types[0] &&
-			isset( $payment_intent->error->decline_code ) &&
-			'generic_decline' === $payment_intent->error->decline_code
+			isset( $payment_intent->error->decline_code )
```
daledupreez: @Mayisha, this change feels risky, as we will now return the compatibility error for _any_ Amazon Pay decline, which is even broader than the original check for the `generic_decline` code.

Could you find a list of what actually changed in the update to Clover? I wasn't able to find a specific set of codes that changed, or what they changed to.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3102,7 +3101,9 @@ private function get_existing_compatible_payment_intent( $order, $payment_method
 		}
 
 		// Check if the status of the intent still allows update.
-		if ( in_array( $intent->status, [ WC_Stripe_Intent_Status::CANCELED, WC_Stripe_Intent_Status::SUCCEEDED ], true ) ) {
+		// Per Stripe API 2025-09-30.clover: do not reuse client secrets for intents in succeeded, canceled,
```
daledupreez: It would be super-helpful to include a link to the relevant docs.

Mayisha: Added.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: @daledupreez, upon further investigation, I concluded that it is safe to delete this check completely and this is the most reasonable solution ([020b564](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977/commits/020b5649ac65d2ceb1065d6c44bc0a9613709df2)).

This if block was added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4269 and the goal was to show a more meaningful failure notice to the shopper as the previous `generic_decline` decline code used to have a generic `The payment failed` message.

Now that, the breaking change in clover https://docs.stripe.com/changelog/clover/2025-09-30/lpm-descriptive-decline-codes is actually adding more descriptive decline codes and messages, we can directly show the message returned by Stripe here. Here is the list of possible decline codes https://docs.stripe.com/declines/codes#lpm-decline-codes.

As I remove the if block here, the code below will call `get_localized_error_message_from_response`. It will then show the message from `$payment_intent->error->message` which now has the descriptive failure reason.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3102,7 +3101,10 @@ private function get_existing_compatible_payment_intent( $order, $payment_method
...
+		// processing, or requires_capture when initializing Elements.
+		// see https://docs.stripe.com/changelog/clover/2025-09-30/client-secret-reuse.
+		if ( in_array( $intent->status, [ WC_Stripe_Intent_Status::CANCELED, WC_Stripe_Intent_Status::SUCCEEDED, WC_Stripe_Intent_Status::PROCESSING, WC_Stripe_Intent_Status::REQUIRES_CAPTURE ], true ) ) {
```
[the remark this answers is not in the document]
daledupreez: @Mayisha, do we have a constant for `requires_action`? I don't think it should be blocking, but it's unlikely that we'll address this later if we don't tackle it now, so probably best to include it in the list, or at least a comment as to why it's not in this list given that it is in the Stripe list.

daledupreez (approved):
The changes look good to me overall. There is one possible gap around preventing secret re-use, but I am not sure if it needs a code change or a comment change.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: Excellent news! Removing the code is definitely the best result here. 🎉

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: On second thought, I now think that the change mentioned in the log is not relevant to this server side code. The doc is exclusively about frontend Stripe.js Elements initialization. https://docs.stripe.com/changelog/clover/2025-09-30/client-secret-reuse mentions initializing `Elements` which is a frontend thing.

I am going to revert this change from this PR as it does not fall under the side effect of updating the Stripe API version.

However, logically the newly added statuses `WC_Stripe_Intent_Status::PROCESSING` and `WC_Stripe_Intent_Status::REQUIRES_CAPTURE` actually make sense here. 

From Claude

> The reason both statuses belong here
 The correct justification is that the Stripe API itself rejects updates/confirms on intents in these states:
> - PROCESSING — the payment is already in-flight (async methods: ACH, SEPA, bank redirects). Stripe will reject any attempt to update or re-confirm it. You must wait for the webhook outcome.
> - REQUIRES_CAPTURE — the intent is already authorized, awaiting a separate capture call. Calling update_and_confirm_payment_intent() on it is semantically wrong and would fail. The correct action is
  capture, not re-confirmation.

 > This is exactly the same reason SUCCEEDED and CANCELED were already excluded — those intents are in a non-updatable terminal state. PROCESSING and REQUIRES_CAPTURE are non-updatable non-terminal states and belong in the same list for the same reason.

 I plan to create a small follow-up to handle this in it's own PR. Let me know your thoughts @daledupreez

daledupreez: That works for me -- this PR is currently something of a blocker for lots of downstream work, while that cleanup can be tackled in a follow-up PR.

Mayisha:
@daledupreez could you please review this comment https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977#discussion_r2848894514 and see if that makes sense?

PR #4879, opened 2025-12-17
POC - Checkout Sessions API

wjrosa:
See: pc4etw-222-p2 / p1765788574224389-slack-C055WHLA98D
Guides: https://docs.stripe.com/checkout/embedded/quickstart?client=react&lang=php / https://docs.stripe.com/payments/payment-element/migration-ewcs?integration-path=one-time#one-time-update-method

## Changes proposed in this Pull Request:

This is a hacky POC for the Checkout Sessions API integration. It uses the `custom` UI option to render the payment methods just like we do currently when the Optimized Checkout feature is enabled.

For this POC, I am disabling the express checkout methods because they are causing conflicts and API signature change errors.

Also, the payment intent object created when we add a checkout session does not contain mandate information.

## Testing instructions

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #4982, opened 2026-02-04
[PoC] Show currency selector after payment methods

daledupreez:
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4879

This PR explores showing the currency selector for adaptive pricing _after_ the payment method listing, rather than within the payment method section of checkout. A UX challenge that we've noticed is that when Stripe is not the first payment method, showing the currency selector above the payment methods results in us mixing interactions: shoppers are picking a payment method from a list,  but we're showing a currency selector in the middle of that list.

The key aspects of this exploration are as follows:
 * When we're rendering the Stripe payment details, we are using an effect to do the following:
   - Create some DOM elements immediately after the payment methods section -- this will need some other approach later to make sure we have the elements available on page load
   - Use `ReactDOM.createPortal()` ([docs](https://18.react.dev/reference/react-dom/createPortal)) to render the currency selector as a child of a separate DOM element
   - Render a hidden sibling fallback element that gets displayed when the currency selector is removed/unmounted -- this provides guidance as to what happened to the element. We'd definitely want this to display on page load.

### Video

https://github.com/user-attachments/assets/e59a110d-97b6-4ac2-81b3-b4daf56fee69

One thing I realised is that our current admin UX makes it hard to enable additional payment methods that don't support the current store currency. However, that can restrict the available payment methods when a shopper picks another currency, as we may have made it hard to enable those payment methods. 

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


daledupreez:
Closing this PR, as it is no longer needed.

PR #4985, opened 2026-02-04
Add an admin notice for merchants potentially affected by the express checkout button location issue

malithsen:
Fixes STRIPE-937

In version 10.1.0, a regression caused express checkout button location settings to be reset during plugin upgrades — removing the checkout page location and keeping only product and cart pages. This was fixed in 10.3.0 (#4864), but merchants who upgraded through 10.1.0–10.2.x may still have incorrect settings without realizing it.

Since we can't distinguish between "the bug set this" and "the merchant intentionally set this," we can't auto-fix. Instead, this PR adds a dismissible admin notice that alerts potentially affected merchants to review their express checkout button location settings.

How it works:

1. During plugin upgrade, if the previous version is in the affected range (≥10.1.0, <10.3.0),
the `wc_stripe_show_ece_location_notice` option is set to 'yes'. Previously dismissed notices are not overwritten.
2. The notice is shown only when all of these conditions are met:
    - The flag option is 'yes' (merchant upgraded through affected versions)
    - Express checkout is enabled
    - Button locations include product and cart but not checkout
    - The notice hasn't been dismissed
3. Clicking dismiss sets the option to 'no', permanently hiding the notice.

#### Prerequisites
- Set up 3 JN sites with plugin versions [10.0.0](https://downloads.wordpress.org/plugin/woocommerce-gateway-stripe.10.0.0.zip), [10.2.0](https://downloads.wordpress.org/plugin/woocommerce-gateway-stripe.10.2.0.zip) and [10.3.1](https://downloads.wordpress.org/plugin/woocommerce-gateway-stripe.10.3.1.zip)
- Onboard to Stripe on all 3. 
- Change Apple Pay/Google Pay locations to cart and product only (uncheck checkout)
- Get a build from this branch

#### Upgrade from 10.2.0
- Upgrade a JN site running Stripe 10.2.0 with a build from this branch
- Go to WC -> Orders (or any WC admin page)
- Make sure there's an admin notice
CleanShot 2026-02-04 at 12 12 51@2x

- Clicking "Review settings" should take you to the express checkout settings page.
- Dismiss the notice. 
- Refresh the page, the notice should not reappear.

#### Upgrade from 10.3.0
- Upgrade a JN site running Stripe 10.3.0 with a build from this branch
- Notice should appear

#### Upgrade from 10.0.0
- Upgrade a JN site running Stripe 10.0.0 with a build from this branch
- Notice should not appear as the bug has not yet been introduced.

#### Upgrade from 10.3.1
- Upgrade a JN site running Stripe 10.3.1 with a build from this branch
- Notice should not appear as the bug has already been fixed.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -625,6 +625,9 @@ class WC_Stripe_Admin_Notices {
 						wp_safe_redirect( remove_query_arg( [ 'wc-stripe-hide-notice', '_wc_stripe_notice_nonce' ], esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) ) ) );
 					}
 					break;
+				case 'ece_location':
+					update_option( 'wc_stripe_show_ece_location_notice', 'no' );
+					break;
 			}
 		}
 	}
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -668,7 +668,7 @@ class WC_Stripe_Admin_Notices {
 		// A bug in these versions reset express checkout button locations during upgrade.
 		$was_affected_version = ! empty( $previous_version )
 			&& version_compare( $previous_version, '10.1.0', '>=' )
-			&& version_compare( $previous_version, '10.3.0', '<' );
+			&& version_compare( $previous_version, '10.4.0', '<' );
 
 		if ( $was_affected_version && 'no' !== get_option( 'wc_stripe_show_ece_location_notice' ) ) {
 			update_option( 'wc_stripe_show_ece_location_notice', 'yes' );
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -407,7 +407,8 @@ class WC_Stripe_Admin_Notices {
 		$has_cart     = in_array( 'cart', $locations, true );
 		$has_checkout = in_array( 'checkout', $locations, true );
 
-		if ( ! $has_product || ! $has_cart || $has_checkout ) {
+		// We only need to show the notice if we have ( product + cart ) but not checkout, so return if we have anything else.
+		if ( ! ( $has_product && $has_cart && ! $has_checkout ) ) {
 			return;
 		}
 
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -390,7 +390,7 @@ class WC_Stripe_Admin_Notices {
 	public function check_express_checkout_location(): void {
 		$show_notice = get_option( 'wc_stripe_show_ece_location_notice' );
 
-		if ( 'no' === $show_notice ) {
+		if ( 'yes' !== $show_notice ) {
 			return;
 		}
 
@@ -664,13 +664,12 @@ class WC_Stripe_Admin_Notices {
 
 		// Set the ECE location notice flag if upgrading from the affected version range (10.1.0–10.2.x).
 		// A bug in these versions reset express checkout button locations during upgrade.
-		// For affected versions, don't overwrite — the flag may have been set by a previous upgrade.
 		$is_affected_version = ! empty( $previous_version )
 			&& version_compare( $previous_version, '10.1.0', '>=' )
 			&& version_compare( $previous_version, '10.3.0', '<' );
 
-		if ( ! $is_affected_version ) {
-			update_option( 'wc_stripe_show_ece_location_notice', 'no' );
+		if ( $is_affected_version && 'no' !== get_option( 'wc_stripe_show_ece_location_notice' ) ) {
+			update_option( 'wc_stripe_show_ece_location_notice', 'yes' );
 		}
 	}
 
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -605,6 +605,17 @@ class WC_Stripe_Admin_Notices {
 		if ( empty( $previous_version ) || version_compare( $previous_version, '4.3.0', 'ge' ) ) {
 			update_option( 'wc_stripe_show_sca_notice', 'no' );
 		}
+
+		// Set the ECE location notice flag if upgrading from the affected version range (10.1.0–10.2.x).
+		// A bug in these versions reset express checkout button locations during upgrade.
+		// For affected versions, don't overwrite — the flag may have been set by a previous upgrade.
+		$is_affected_version = ! empty( $previous_version )
+			&& version_compare( $previous_version, '10.1.0', '>=' )
+			&& version_compare( $previous_version, '10.3.0', '<' );
+
+		if ( ! $is_affected_version ) {
+			update_option( 'wc_stripe_show_ece_location_notice', 'no' );
+		}
 	}
 
 	/**
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -67,6 +67,7 @@ class WC_Stripe_Admin_Notices {
 		$this->payment_methods_check_environment();
 
 		// Check for merchants affected by ECE button location bug.
+		// https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4861
 		$this->check_express_checkout_location();
 
 		// Check for subscriptions detached from the customer.
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -384,8 +384,10 @@ class WC_Stripe_Admin_Notices {
 	 * in versions 10.1.0–10.2.x and displays a notice if so.
 	 *
 	 * @since 10.4.0
+	 *
+	 * @return void
 	 */
-	public function check_express_checkout_location() {
+	public function check_express_checkout_location(): void {
 		$show_notice = get_option( 'wc_stripe_show_ece_location_notice' );
 
 		if ( 'no' === $show_notice ) {
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -416,7 +416,7 @@ class WC_Stripe_Admin_Notices {
 
 		$message = sprintf(
 			/* translators: 1) HTML strong open tag 2) HTML strong closing tag 3) HTML line break tag */
-			__( '%1$sAction Required: Review your express checkout settings.%2$s%3$sA recent update may have unintentionally changed where Apple Pay / Google Pay buttons appear. Currently, they are active on the product and cart pages but not on the checkout page. Please verify your settings to ensure your customers have the best checkout experience.', 'woocommerce-gateway-stripe' ),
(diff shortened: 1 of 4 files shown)


includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -605,6 +662,16 @@ public function stripe_updated() {
 		if ( empty( $previous_version ) || version_compare( $previous_version, '4.3.0', 'ge' ) ) {
 			update_option( 'wc_stripe_show_sca_notice', 'no' );
 		}
+
+		// Set the ECE location notice flag if upgrading from the affected version range (10.1.0–10.2.x).
+		// A bug in these versions reset express checkout button locations during upgrade.
+		$is_affected_version = ! empty( $previous_version )
```
Mayisha: Nit.  Something like `$was_affected_version` or `$is_upgraded_from_affected_version` as the variable name would be better for readability, as we are checking the previous version.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -376,6 +380,56 @@ public function payment_methods_check_environment() {
...
+
+		if ( ! $enabled ) {
+			return;
+		}
+
+		$has_product  = in_array( 'product', $locations, true );
+		$has_cart     = in_array( 'cart', $locations, true );
+		$has_checkout = in_array( 'checkout', $locations, true );
+
+		if ( ! $has_product || ! $has_cart || $has_checkout ) {
```
Mayisha: QQ: If ECE was enabled for cart and checkout (not product) before the affected version, did the affected version not disable it from `checkout`, keeping only `cart` enabled?
Was it only affected if all 3 locations were enabled?

Mayisha (approved):
Working as described. I have one optional suggestion and a question to confirm we are not overlooking any possible scenarios.

Image

includes/admin/class-wc-stripe-admin-notices.php
malithsen: Good point. Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4985/commits/55b7760b502f97efe4fccca9e1743f6caac5e4da

includes/admin/class-wc-stripe-admin-notices.php
malithsen: After the upgrade, locations were reset to the filter defaults (`cart` and `product`), regardless of the merchant's initial configuration.

daledupreez: Nit: This expression needs some parsing to work out what we're checking. It may be worth doing two things:
1. Adding an inline comment to document what we're checking for, and
2. Using an outer `! ( )` condition so we are more explicitly identifying the case that we care about. While the expressions are definitely equivalent, I find the current expression hard to grok!
```suggestion
		// We only need to show the notice if we have ( product + cart ) but not checkout, so return if we have anything else.
		if ( ! ( $has_product && $has_cart && ! $has_checkout ) ) {
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -605,6 +662,16 @@ public function stripe_updated() {
 		if ( empty( $previous_version ) || version_compare( $previous_version, '4.3.0', 'ge' ) ) {
 			update_option( 'wc_stripe_show_sca_notice', 'no' );
 		}
+
+		// Set the ECE location notice flag if upgrading from the affected version range (10.1.0–10.2.x).
+		// A bug in these versions reset express checkout button locations during upgrade.
+		$was_affected_version = ! empty( $previous_version )
+			&& version_compare( $previous_version, '10.1.0', '>=' )
+			&& version_compare( $previous_version, '10.3.0', '<' );
```
daledupreez: @malithsen, what would happen to merchants who followed a normal upgrade procedure and got auto-updated to 10.1.0, then 10.2.0, and then 10.3.0? Would they not have had the locations modified during the initial upgrade to 10.1.0 (or 10.2.0 if they skipped 10.1.0), but then they would have remained untouched during the upgrade to 10.3.0? Given that, I think we need to shift the version comparison to `< 10.4.0`.
```suggestion
		// Set the ECE location notice flag if upgrading from the affected version range (10.1.0–10.2.x).
		// A bug in these versions reset express checkout button locations during upgrade.
		$was_affected_version = ! empty( $previous_version )
			&& version_compare( $previous_version, '10.1.0', '>=' )
			&& version_compare( $previous_version, '10.4.0', '<' );
```

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -873,4 +876,254 @@ public function provide_test_subscription_check_detachment_bulk_action() {
...
+
+		// Remove hooks to prevent side effects during construction.
+		remove_all_actions( 'admin_notices' );
+		remove_all_actions( 'wp_loaded' );
+		remove_all_actions( 'woocommerce_stripe_updated' );
+		$notices = new WC_Stripe_Admin_Notices();
+
+		$notices->stripe_updated();
+
+		$this->assertNotEquals( 'yes', get_option( 'wc_stripe_show_ece_location_notice' ) );
```
daledupreez: Nit: It feels like testing that the option was not set at all would be better than confirming it is not `'yes'`. And ensuring that the value is a specific starting value before the test feels like it would be more robust here and elsewhere.
```suggestion
		update_option( 'wc_stripe_version', '10.0.0' );
		delete_option( 'wc_stripe_show_ece_location_notice' );

		// Remove hooks to prevent side effects during construction.
		remove_all_actions( 'admin_notices' );
		remove_all_actions( 'wp_loaded' );
		remove_all_actions( 'woocommerce_stripe_updated' );
		$notices = new WC_Stripe_Admin_Notices();

		$notices->stripe_updated();

		$this->assertFalse( get_option( 'wc_stripe_show_ece_location_notice' ) );
```

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -873,4 +876,254 @@ public function provide_test_subscription_check_detachment_bulk_action() {
 			],
 		];
 	}
+
+	/**
+	 * Test that upgrading from an affected version (10.1.0–10.2.x) sets the flag to 'yes'.
+	 */
+	public function test_stripe_updated_sets_ece_location_flag_for_affected_versions() {
```
daledupreez: Not blocking, but there is a lot of repetition across the test methods.

daledupreez (commented):
I have some feedback based on inspection -- I'll pick up testing tomorrow.

includes/admin/class-wc-stripe-admin-notices.php
malithsen: ah good catch. Changing to 10.4.0 makes sense. It does increase the false positives a little bit (merchants who upgraded from 10.0.0 -> 10.3.0 and never ran into the bug) could also get the notice if they match the location signature. But in this case, we'd want to optimize for minimizing false negatives rather than false positives.

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
malithsen: Switched to the dataprovider pattern in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4985/commits/e23ac7ddc5e72b4f44c6876ed7adde6cb289aec2 to reduce repetition.

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
```diff
@@ -878,252 +878,154 @@ public function provide_test_subscription_check_detachment_bulk_action() {
...
 	/**
 	 * Test that dismissing the ece_location notice sets the option to 'no'.
 	 */
 	public function test_hide_notices_dismisses_ece_location_notice() {
-		// Set the current user to an admin.
 		$admin_user = self::factory()->user->create( [ 'role' => 'administrator' ] );
 		wp_set_current_user( $admin_user );
 
-		// Simulate the dismiss request.
 		$_GET['wc-stripe-hide-notice']   = 'ece_location';
```
daledupreez, line 1023: It may be worth ensuring that the option is either deleted or set to `'yes'` before running the test.
```suggestion
		delete_option( 'wc_stripe_show_ece_location_notice' );

		$_GET['wc-stripe-hide-notice']   = 'ece_location';
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -376,6 +380,57 @@ public function payment_methods_check_environment() {
...
+		// We only need to show the notice if we have ( product + cart ) but not checkout, so return if we have anything else.
+		if ( ! ( $has_product && $has_cart && ! $has_checkout ) ) {
+			return;
+		}
+
+		$settings_url = admin_url( 'admin.php?page=wc-settings&tab=checkout&section=stripe&panel=methods&area=express_checkout' );
+
+		$message = sprintf(
+			/* translators: 1) HTML strong open tag 2) HTML strong closing tag 3) HTML line break tag */
+			__( '%1$sAction Required: Review your express checkout settings.%2$s%3$sA recent update may have unintentionally changed where Apple Pay / Google Pay buttons appear. Currently, they are active on the product and cart pages but not on the checkout page. Please verify your settings to ensure your customers have the best checkout experience.', 'woocommerce-gateway-stripe' ),
```
daledupreez: Given that we are going to show this notice in a lot of places outside the plugin, we need to include more context in the text.
```suggestion
			__( '%1$sAction Required: Review your Stripe express checkout settings.%2$s%3$sA recent update to the Stripe plugin may have unintentionally changed where Apple Pay and/or Google Pay buttons appear. Currently, they are active on the product and cart pages but not on the checkout page. Please review your express checkout settings to ensure your customers have the best checkout experience.', 'woocommerce-gateway-stripe' ),
```

daledupreez: Also, is it worth mentioning that there was a bug? e.g. "A recent bug in the Stripe plugin may have unintentionally..."

daledupreez (approved):
Functionally, the changes are working well, including test upgrades from 10.3.0. (Side note: please update the testing instructions, as upgrades from 10.3.0 _should_ see the warning.

I do think it would be a good idea to see if @pierorocca can take a look at the wording, though.

includes/admin/class-wc-stripe-admin-notices.php
daledupreez: And is it worth being explicit in terms of what may have changed, i.e. "may have stopped showing Apple Pay and/or Google Pay buttons on the Checkout page".

tests/phpunit/Admin/WC_Stripe_Admin_Notices_Test.php
malithsen: Good point, thanks! I've set it to 'yes' instead of deleting it, so the assertion proves the value was actively changed.

includes/admin/class-wc-stripe-admin-notices.php
pierorocca: Stripe is similar to WooPayments in which Apple Pay and Google Pay are tethered together? If so "and" is more  precise than "and/or".  Regarding "bug", I find it softer using the "update with unintentional" language and it gets across the same message. Otherwise great edits @daledupreez.

malithsen: Thank you both! Stripe is similar to WooPayments, and Apple Pay and Google Pay are both controlled by a single checkbox. I will accept Dale's suggestion, then change "and/or" to "and".

Issue #4906, opened 2025-12-30
PHPStan: Address issues in WC_Stripe_Subscriptions_Trait


daledupreez:
PHPStan reports issues from traits every time they are triggered from classes that use the trait. This means that issues from `WC_Stripe_Subscriptions_Trait` in `includes/compat/trait-wc-stripe-subscriptions.php` are currently reporting over 100 errors for each class using the trait! And we currently have 15 classes using the trait directly, which is a _lot_ of errors.

It is worth noting that a good number of the errors stem from missing stubs for WooCommerce Subscription classes, which is being tracked in https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4901, but any work to reduce the number of errors in the trait will have a large impact on the overall number of errors reported by PHPStan.

Issue #4905, opened 2025-12-30
PHPStan: Add return types


daledupreez:
Our current `phpstan-baseline.neon` file includes over 650 `missingType.return` errors, which are reported when methods or functions don't specify a return type. Where possible, we should aim to specify return types. Ideally, we would like to specify explicit return types, e.g. `some_function(): void`, but in many cases we will need to use the PHPDoc `@return` element to specify return types.

Given the number of cases for this specific error type, we almost certainly need to tackle this in smaller chunks, possibly via a few classes at a time.

PR #4990, opened 2026-02-05
Improve PHPDoc and fix bugs in compatibility classes

daledupreez:
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4955
Towards https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4905
Towards https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4906
Towards STRIPE-873
Towards STRIPE-874

This PR updates the PHPDoc for our compatibility code for WooCommerce Subscriptions and WooCommerce Pre Orders, as well as some code called from the related classes and traits.

As a result of some issues flagged by PHPStan, this PR also addresses some small bugs in `includes/compat/trait-wc-stripe-subscriptions.php`:
 * I am changing the code to use `add_action()` instead of `add_filter()` for the `woocommerce_subscription_validate_payment_meta` action
 * The code in `process_subscription_payment()` consistently returns `void` instead of returning the result from calling itself on retries.
 * I have added a `phpstan-ignore` for the code that hooks `$this->delete_renewal_meta()` into the  `wcs_renewal_order_created` action -- `delete_renewal_meta()` returns a value, but changing that behaviour could break other code, so we are going to ignore that.

All of the above address a _lot_ of PHPStan errors, as the issues were reported for all classes that use the traits.

## Testing instructions

Code review should be sufficient, but pay extra attention to the bug fixes.

Also confirm that we are not adding any PHPStan errors.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -241,7 +241,7 @@ public function is_no_such_token_error( $error ) {
 	 * error and it is no such source.
 	 *
 	 * @since 4.1.0
-	 * @param object $error
+	 * @param object|false $error
 	 */
 	public function is_no_such_source_error( $error ) {
```
Mayisha, line 247: Let's add the return type in this method (and the other two below) while we are fixing the param type.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -357,8 +363,9 @@ public function process_change_subscription_payment_with_deferred_intent( $subsc
 	/**
 	 * Scheduled_subscription_payment function.
```
Mayisha: Nit: Let's add a better description for `scheduled_subscription_payment` and `process_subscription_payment` methods.

Mayisha (approved):
Thanks for the improvements in phpdoc 👍

malithsen (approved):
Looks good

includes/compat/trait-wc-stripe-subscriptions.php
daledupreez: Good call. Done in b11957b9fd6a1d6975eb11a26fe5967fe3b8a489.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
daledupreez: Good call. Added both in f9b4cf3f6b4a05df79d739bada17aaa2b1ae9ff3.

PR #4994, opened 2026-02-05
Adaptive pricing in the block checkout

wjrosa:
Towards STRIPE-945
Towards #4995

Needs to be merged after the following branches
- https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977
- https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4993

- Creates a checkout session on block checkout.
- Renders the currency selector.
- No payment processing implemented yet.

Preview:
Screenshot 2026-02-12 at 13 13 12

## Testing instructions

- Checkout and build this branch on your test environment (`add/checkout-sessions-block-checkout`)
- Connect your Stripe account
- Enable the Optimized Checkout feature
- Added the following snippet to enable checkout sessions on your store:
```
```
- Enable the Adaptive Pricing feature in settings
- Confirm the checkout sessions payment container loads and renders the list of payment methods along with the currency selector
- Confirm that the list of payment methods is automatically updated when the currency is switched

\* If the checkout session ajax fails for you, comment out the [nonce check](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-checkout-sessions-controller.php#L28-L30). (nonce issue will be fixed separately)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (commented):
Observations from a quick functional test

1. The testing instructions are missing.

`develop`

Image

`add/checkout-sessions-block-checkout`

Image

2. I see in the screenshot of the description that Google Pay is visible in the list which is not supposed to be there 🤔

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,85 @@
...
+} from '@stripe/react-stripe-js/checkout';
+import { useRef, useState } from 'react';
+import { __ } from '@wordpress/i18n';
+
+/**
+ * No operation function.
+ *
+ * @return {null} Returns null.
+ */
+const noop = () => null;
```
Mayisha: Is this needed?

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,85 @@
...
+
+const CheckoutForm = ( {
+	components: { LoadingMask },
+	onLoadError = noop,
+} ) => {
+	const checkoutState = useCheckout();
+	const [ , setSelectedPaymentMethodType ] = useState( null );
```
Mayisha: Don't we need the states? (`selectedPaymentMethodType` & `isPaymentElementComplete`)

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,85 @@
...
+	) {
+		const { checkout } = checkoutState;
+		setCheckoutSessionId( checkout.id );
+	}
+
+	return (
+		<>
+			<CurrencySelectorElement />
+			<PaymentElement
+				options={ {
```
Mayisha: Should we use the `getStripeElementOptions` function from [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/add/checkout-sessions-block-checkout/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js#L42) or move it to a common util class?

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Checkout Sessions Container component.
+ *
+ * @param {Object} props Component props.
+ * @return {JSX.Element} The Checkout Sessions Container component.
+ */
+export const CheckoutSessionsContainer = ( props ) => {
+	const { api } = props;
+	const promise = useMemo( async () => {
+		const response = await api.checkoutSessionsCreateSession();
```
Mayisha, line 23: In case we fail to create a checkout session, I think we should fallback to the regular payment element?

client/blocks/checkout-sessions/checkout-form.js
daledupreez: Or, phrased differently, we are only setting both of these local state values in `onSelectedPaymentMethodChange()`, but we don't have any code using the values, which feels unnecessary. If we aren't doing anything with the values, we should remove the related code.

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -0,0 +1,61 @@
+import { StoreNotice } from '@woocommerce/blocks-checkout';
+import { CheckoutProvider } from '@stripe/react-stripe-js/checkout';
+import React, { useMemo, useState } from 'react';
+import CheckoutForm from 'wcstripe/blocks/checkout-sessions/checkout-form';
+import { loadStripe } from 'wcstripe/blocks/load-stripe';
+import { initializeUPEAppearance } from 'wcstripe/stripe-utils';
+import { getFontRulesFromPage } from 'wcstripe/styles/upe';
+
+const stripePromise = loadStripe();
```
daledupreez, line 8: Is this the right moment/timing to load Stripe? It feels like we might over-eagerly load Stripe if this file gets referenced from elsewhere, even if we don't use the component. Would it make sense to trigger the load as part of an effect within the component?

client/blocks/checkout-sessions/checkout-sessions-container.js
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Checkout Sessions Container component.
+ *
+ * @param {Object} props Component props.
+ * @return {JSX.Element} The Checkout Sessions Container component.
+ */
+export const CheckoutSessionsContainer = ( props ) => {
+	const { api } = props;
+	const promise = useMemo( async () => {
```
daledupreez: Nit: I think it would help to identify what the `promise` holds.
```suggestion
	const checkoutSessionPromise = useMemo( async () => {
```

client/blocks/checkout-sessions/hooks.js
```diff
@@ -0,0 +1,73 @@
...
+	onCheckoutSuccess
+) => {
+	useEffect(
+		() =>
+			onCheckoutSuccess(
+				( { processingResponse: { paymentDetails } } ) => {
+					if ( checkoutState.type !== 'success' ) {
+						return {
+							type: 'error',
+							message: 'Checkout is not ready for confirmation.',
```
daledupreez: Nit: I think this message should be translated.
```suggestion
							message: __( 'Checkout is not ready for confirmation.', 'woocommerce-gateway-stripe' ),
```

client/blocks/checkout-sessions/hooks.js
```diff
@@ -0,0 +1,73 @@
...
+						returnUrl: redirect,
+					} );
+					if ( confirmResult?.type === 'error' ) {
+						return {
+							type: 'error',
+							message: confirmResult.error.message,
+						};
+					}
+
+					// If no error, we assume success for now. This return value is never used the `confirm` is success.
```
daledupreez: I am not sure what this comment means -- I think we're missing one or more words, maybe something like the following?
```suggestion
					// If no error, we assume success for now. This return value is never used, as the `confirm` call indicates success.
```

client/blocks/checkout-sessions/checkout-container.js
daledupreez: Secondary note: the nonce checks are failing because the server code is looking for the nonce in the default `_ajax_nonce` or `_wpnonce` parameters, but we're sending the nonce in the `security` parameter.

client/blocks/checkout-sessions/checkout-sessions-container.js
```diff
@@ -0,0 +1,61 @@
...
+		const response = await api.checkoutSessionsCreateSession();
+		const clientSecret = response.data?.client_secret;
+		if ( ! clientSecret ) {
+			throw new Error( 'Missing client secret in response' );
```
daledupreez: We should ensure we display a more shopper-friendly message, as this is just confusing.

Also, we need to be clear about how we intend to handle cases where the Stripe payment method(s) are displayed before the shopper has entered any details in checkout. This API call is currently failing for me because a customer can't be created before I enter any data in the checkout form.

client/blocks/checkout-sessions/checkout-form.js
daledupreez: I think this is why we're missing the recent changes to specify `excludedPaymentMethodTypes`, which would likely be handled by this shared logic. You can see unsupported payment method types in my screenshot below -- we don't support Crypto or Bank transfer.

Screenshot 2026-02-16 at 13 32 37

daledupreez (commented):
Functionally, this is working for me when my shopper has existing details, but not working when we don't have enough data to create a Stripe customer. Things still feel a bit rough, but it's not clear how much of that is expected.

client/blocks/checkout-sessions/checkout-form.js
wjrosa: Not really. Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/11816ab670c056f42c6eb7e944108d1be7729398

client/blocks/checkout-sessions/checkout-form.js
wjrosa: I removed them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/changes/b2009181b49c4c0bd9c3c6e8c6559dbb51e55179. I copied those from the original implementation by mistake. Thanks for the heads-up!

wjrosa:
Thanks for the review, folks! I have added the missing testing instructions in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/dde66cc0da8c761c5f6ef39bef3cead6a37d2144

client/blocks/checkout-sessions/checkout-form.js
wjrosa: > Should we use the getStripeElementOptions function from [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/add/checkout-sessions-block-checkout/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js#L42) or move it to a common util class?Should we use the getStripeElementOptions function from [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/add/checkout-sessions-block-checkout/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js#L42) or move it to a common util class?

Good idea! Moved the function to `client/blocks/utils.js` and imported it in both places (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/8669122d7bf074d0efa1c66a5c081e6c107e06be)

wjrosa: > I think this is why we're missing the recent changes to specify excludedPaymentMethodTypes, which would likely be handled by this shared logic.

Not really. We have 2 types of Stripe component options: element options and provider options. `excludedPaymentMethodTypes` is a provider option. But `CheckoutProvider` does not accept this setting. I will check with Stripe what we can do in this case. See the error below:
Screenshot 2026-02-19 at 10 51 39

wjrosa: Asked them about it in p1771509346517089-slack-C9976E5MJ

wjrosa: I actually saw that, for the checkout sessions flow, we have to set `excluded_payment_method_types` when creating the session on the backend instead ([doc](https://docs.stripe.com/api/checkout/sessions/create#create_checkout_session-excluded_payment_method_types)). So I implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/234c24c64f55c8a231eb0f8644fb2f25e32e8fa4

client/blocks/checkout-sessions/checkout-sessions-container.js
wjrosa: > We should ensure we display a more shopper-friendly message, as this is just confusing.

Makes sense. Made it friendlier in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/d5ec6ae58f9484d89722f5f994a867d0e6ad6813

> Also, we need to be clear about how we intend to handle cases where the Stripe payment method(s) are displayed before the shopper has entered any details in checkout. This API call is currently failing for me because a customer can't be created before I enter any data in the checkout form.

This is a known issue. I created STRIPE-953 now just to fix it

client/blocks/checkout-sessions/checkout-container.js
wjrosa: `loadStripe` here is actually the promise, not the resolved function. Stripe controls when the promise is called/resolved. That's the same implementation we have today for ECE (see `client/blocks/express-checkout/index.js#24`)

client/blocks/checkout-sessions/checkout-container.js
wjrosa: > In case we fail to create a checkout session, I think we should fallback to the regular payment element?

That's a good call. I added the fallback logic in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/233a17f2df33729b1b6d814147f7808ae80b6dbb

client/blocks/checkout-sessions/__tests__/checkout-form.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+		render(
+			<CheckoutForm
+				components={ components }
+				onLoadError={ onLoadError }
+			/>
+		);
+
+		expect( screen.getByText( 'Payment Element' ) ).toBeInTheDocument();
+	} );
```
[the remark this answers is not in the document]
wjrosa: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/92a2606875d6d17024ac0a214421622638993e5d

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,92 @@
+import {
+	CurrencySelectorElement,
+	PaymentElement,
+	useCheckout,
+} from '@stripe/react-stripe-js/checkout';
+import { useRef, useState } from 'react';
+import { __ } from '@wordpress/i18n';
+import { handleDisplayOfPaymentInstructions } from 'wcstripe/optimized-checkout/handle-display-of-payment-instructions';
```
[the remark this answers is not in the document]
wjrosa: > Lines 49–55 call setCheckoutSessionId() directly during render, creating a state-update-in-render violation. Although the guard check prevents an infinite loop, this pattern violates React rules. Move the state update into a useEffect with checkoutState.checkout.id as a dependency.

Changing it as suggested causes an exhaustive deps error.

Applied the other suggestions in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/eb2e0525f95a8af66306bb84371326bad196089c

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,92 @@
...
+	onLoadError,
+	testingInstructions,
+} ) => {
+	const checkoutState = useCheckout();
+	const [ checkoutSessionId, setCheckoutSessionId ] = useState( null );
+	const hasLoadErrorRef = useRef( false );
+	const setHasLoadError = ( event ) => {
+		hasLoadErrorRef.current = true;
+		onLoadError( event );
+	};
```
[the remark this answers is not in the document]
wjrosa: Removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/eb2e0525f95a8af66306bb84371326bad196089c

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,92 @@
...
+			/>
+		);
+	} else if ( checkoutState.type === 'error' ) {
+		return <div>Error: { checkoutState.error.message }</div>;
+	} else if (
+		checkoutState.type === 'success' &&
+		checkoutSessionId !== checkoutState.checkout.id
+	) {
+		const { checkout } = checkoutState;
+		setCheckoutSessionId( checkout.id );
```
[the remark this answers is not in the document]
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/eb2e0525f95a8af66306bb84371326bad196089c

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,92 @@
...
+	}
```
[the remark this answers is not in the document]
wjrosa: This causes an exhaustive deps error.

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,92 @@
...
+
+	return (
+		<>
+			<p
+				className="content"
+				dangerouslySetInnerHTML={ {
+					__html: testingInstructions,
+				} }
+			/>
```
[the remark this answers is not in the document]
wjrosa: Applied in the latest commits.

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -0,0 +1,62 @@
...
+	] = useState( null );
+
+	const providerOptions = {
+		clientSecret: checkoutSessionPromise,
+		adaptivePricing: { allowed: true },
+		elementsOptions: {
+			appearance: initializeUPEAppearance( api, 'true' ),
+			fonts: getFontRulesFromPage(),
+		},
+	};
```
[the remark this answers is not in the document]
wjrosa: Applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/483bbf9a09bb950df6b821940e142c54e8a8fe84

client/blocks/checkout-sessions/hooks.js
```diff
@@ -0,0 +1,77 @@
...
+					const { checkout } = checkoutState;
+					const confirmResult = checkout.confirm( {
+						returnUrl: redirect,
+					} );
+					if ( confirmResult?.type === 'error' ) {
+						return {
+							type: 'error',
+							message: confirmResult.error.message,
+						};
+					}
```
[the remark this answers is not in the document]
wjrosa: Applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/d018119037d039e3bccbab715b449b3ce622f470

client/blocks/checkout-sessions/hooks.js
```diff
@@ -0,0 +1,77 @@
...
+		() =>
+			onCheckoutFail( ( { processingResponse: { paymentDetails } } ) => {
+				return {
+					type: 'failure',
+					message: paymentDetails.errorMessage,
+					messageContext: emitResponse.noticeContexts.PAYMENTS,
+				};
+			} ),
+		[ checkoutState, onCheckoutFail, emitResponse.noticeContexts.PAYMENTS ]
+	);
```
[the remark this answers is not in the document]
wjrosa: Applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/341a4d514646d9302dbd8345d70d180a697ffd93

client/blocks/upe/upe-element.js
```diff
@@ -65,26 +69,30 @@ export const upeElement = ( paymentMethod, api, upeConfig ) => {
...
 			upeConfig.supportsDeferredIntent
-		),
+		);
+	}
+
+	return {
+		name: upeMethods[ paymentMethod ],
+		content: paymentMethodComponent,
+		edit: paymentMethodComponent,
 		savedTokenComponent: <SavedTokenHandler api={ api } />,
```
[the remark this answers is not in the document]
wjrosa: Applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/5eaca818ac99114663297c5bb826fec832edb373

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,77 @@
...
+	const checkoutState = useCheckout();
+	const [ checkoutSessionId, setCheckoutSessionId ] = useState( null );
+	const hasLoadErrorRef = useRef( false );
+	const setHasLoadError = ( event ) => {
+		hasLoadErrorRef.current = true;
+		onLoadError( event );
+	};
+	const onSelectedPaymentMethodChange = ( { value } ) => {
+		handleDisplayOfPaymentInstructions( value.type );
+	};
```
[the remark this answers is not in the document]
wjrosa: This feature works only when OC is enabled, so that would be redundant

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -408,7 +339,7 @@ const PaymentProcessor = ( {
 			<p
 				className="content"
 				dangerouslySetInnerHTML={ {
-					__html: testingInstructionsIfAppropriate,
+					__html: testingInstructions,
 				} }
 			/>
```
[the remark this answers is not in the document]
wjrosa: Fixed it previously

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,77 @@
...
+				className="content"
+				dangerouslySetInnerHTML={ {
+					__html: testingInstructions,
+				} }
+			/>
+			<CurrencySelectorElement />
+			<PaymentElement
+				options={ getStripeElementOptions( true ) }
+				onChange={ onSelectedPaymentMethodChange }
+				onLoadError={ setHasLoadError }
```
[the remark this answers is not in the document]
wjrosa: Applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/a3ad05c2592faaea8f5df5f605f0e7b591a01039

client/blocks/utils.js
```diff
@@ -142,3 +145,73 @@ export const getStripeImageUrl = ( imageName ) => {
...
+		};
+		if ( layout.type === OPTIMIZED_CHECKOUT_DEFAULT_LAYOUT ) {
+			layout.radios = false;
+			layout.spacedAccordionItems = false;
+		}
+		options = {
+			...options,
+			layout,
+		};
+	}
```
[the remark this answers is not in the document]
wjrosa: Applied in ddbbdcf70

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -69,13 +72,14 @@ public function create_checkout_session(): void {
...
+				'customer'                      => $stripe_customer->get_id(),
+				'line_items'                    => $line_items,
+				'payment_method_types'          => $enabled_payment_methods,
+				'excluded_payment_method_types' => $gateway->get_excluded_payment_method_types(),
+				'payment_intent_data'           => [], // @todo Pass additional data if needed.
+				'mode'                          => 'payment',
+				'adaptive_pricing'              => [
 					'enabled' => 'true',
 				],
 			];
```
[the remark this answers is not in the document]
wjrosa: Applied in 8a9e886b1

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -0,0 +1,68 @@
...
+			) }
+			<CheckoutProvider
+				stripe={ stripePromise }
+				options={ providerOptions }
+			>
+				<CheckoutForm
+					onLoadError={ setPaymentProcessorLoadErrorMessage }
+					testingInstructions={ testingInstructions }
+					{ ...props }
+				/>
```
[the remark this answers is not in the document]
wjrosa: Applied in 84938e3ed

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,79 @@
...
+				showSpinner={ true }
+				screenReaderLabel={ __(
+					'Loading payment method…',
+					'woocommerce-gateway-stripe'
+				) }
+			/>
+		);
+	} else if ( checkoutState.type === 'error' ) {
+		setShouldLoadStripeElements( true ); // If there was an error loading the checkout session, we fallback to loading Stripe Elements.
+		return <div>Error: { checkoutState.error.message }</div>;
```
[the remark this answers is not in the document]
wjrosa: Passing `checkoutState` to any `useEffect` dependency list causes an exhaustive deps error.

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -48,6 +45,9 @@ public function create_checkout_session(): void {
 				throw new Exception( __( 'Your cart is currently empty.', 'woocommerce-gateway-stripe' ) );
 			}
 
+			$gateways = WC()->payment_gateways()->payment_gateways();
+			$gateway  = $gateways[ WC_Stripe_UPE_Payment_Gateway::ID ];
```
Mayisha: QQ: I think we use `WC_Stripe::get_instance()->get_main_stripe_gateway()` in all other places to get the gateway. Is this not used here intentionally?

client/blocks/checkout-sessions/checkout-form.js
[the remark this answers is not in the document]
Mayisha: @wjrosa we should consider moving it to useEffect.

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -0,0 +1,87 @@
...
+		);
+	} else if ( checkoutState.type === 'error' ) {
+		setShouldLoadStripeElements( true ); // If there was an error loading the checkout session, we fallback to loading Stripe Elements.
+		return <div>Error: { checkoutState.error?.message }</div>;
+	} else if (
+		checkoutState.type === 'success' &&
+		checkoutSessionId !== checkoutState.checkout?.id
+	) {
+		const { checkout } = checkoutState;
+		setCheckoutSessionId( checkout.id );
```
Mayisha, line 63: I did not find the usage of this checkout session ID. Is this supposed to be used later in the payment processing piece?

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -0,0 +1,55 @@
...
+ * Checkout Sessions Container component.
+ *
+ * @param {Object} props Component props.
+ * @return {JSX.Element} The Checkout Sessions Container component.
+ */
+export const CheckoutContainer = ( props ) => {
+	const { api, setShouldLoadStripeElements } = props;
+	const checkoutSessionPromise = useMemo( async () => {
+		const response = await api.checkoutSessionsCreateSession();
+		const clientSecret = response.data?.client_secret;
```
Mayisha:
```suggestion
		const clientSecret = response?.data?.client_secret;
```
Just in case.

client/blocks/upe/upe-deferred-intent-creation/payment-elements.js
```diff
@@ -182,6 +168,79 @@ const PaymentElements = ( {
...
+	const isAdaptivePricingSupported =
+		stripeServerData?.isAdaptivePricingEnabled;
+
+	const [ errorMessage ] = useState( null );
+	const [ paymentProcessorLoadErrorMessage ] = useState( null );
+	const [ shouldLoadStripeElements, setShouldLoadStripeElements ] = useState(
+		! stripeServerData?.isAdaptivePricingEnabled
+	);
+
+	if ( errorMessage ) {
```
Mayisha, line 220: Seems like it's not going to work. We are not setting the errorMessage in this component so the sate should be always null.

client/blocks/upe/upe-deferred-intent-creation/payment-elements.js
```diff
@@ -182,6 +168,79 @@ const PaymentElements = ( {
...
+				supportsDeferredIntent={ supportsDeferredIntent }
+				LoadingMask={ LoadingMask }
+				{ ...props }
+			/>
+		);
+	}
+
 	return (
 		<>
 			{ paymentProcessorLoadErrorMessage?.error?.message && (
```
Mayisha, line 261: Same as above comment. We are not setting `paymentProcessorLoadErrorMessage` in this component.

Mayisha (commented):
I see the currency selector is rendered properly on the block checkout page. This mostly looks good to me, considering payment processing is not part of these changes.

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Not really. I picked the first occurrence I found. But that way is definitely better. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/46269f286a5b85aa69850139e8dd95ee2b4c312f

client/blocks/checkout-sessions/checkout-form.js
wjrosa: That's right. That's how we will set the post data in the payment setup hook to send it to the backend later

client/blocks/upe/upe-deferred-intent-creation/payment-elements.js
wjrosa: Good catch! Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/b6b8660f80104305d96b74e2a5a158e607bf5956

client/blocks/upe/upe-deferred-intent-creation/payment-elements.js
wjrosa: Also fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/b6b8660f80104305d96b74e2a5a158e607bf5956

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -0,0 +1,58 @@
...
+		api,
+		setPaymentProcessorLoadErrorMessage,
+		setShouldLoadStripeElements,
+	} = props;
+	const checkoutSessionPromise = useMemo( async () => {
+		const response = await api.checkoutSessionsCreateSession();
+		const clientSecret = response?.data?.client_secret;
+		if ( ! clientSecret ) {
+			setShouldLoadStripeElements( true );
+			throw new Error(
```
Mayisha: May be we should log this error and return instead of throwing this error.

Image

includes/class-wc-stripe-checkout-sessions-controller.php
Mayisha: I think these were removed before in this branch. I I see these unused variables again, probably got re-added during a merge.

Mayisha (approved):
Thanks for the updates. The adaptive pricing option is correctly displayed on the block checkout page.

Image

client/blocks/checkout-sessions/checkout-container.js
wjrosa: Alright. Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/e237d1ba7d323dc243f240c52c678eafbda67f3e

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Good catch! Re-removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4994/commits/83b80dfad0a4fd53bece13b6c079608e6068839c

daledupreez (approved):
The changes look good to me. My only minor comment is that the spacing for the "Save payment information" element is bad. We may want to specify additional spacing, because it doesn't look great at the moment. (This is likely a separate, pre-existing issue.)

Image

wjrosa:
Thanks for the review, folks!

@daledupreez added STRIPE-964 to fix that

daledupreez:
And GitHub link for the tracking issue: https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5073

PR #4998, opened 2026-02-05
Deprecate unused non-deferred intent methods

wjrosa:
See https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4929/changes#r2741768964

In this PR, I am deprecating some unused non-deferred intent methods in the `WC_Stripe_UPE_Payment_Gateway` class:
- `process_payment_with_saved_payment_method`
- `validate_upe_checkout_experience_accepted_payments_field`
- `generate_upe_checkout_experience_accepted_payments_html`

We plan to remove these in later versions. We are deprecating them first to be safe (in case some 3rd-party plugins reference them).

## Testing instructions

Code review should be enough.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (approved):
Looks good to me 👍

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1551,8 +1551,12 @@ private function handle_process_payment_error( WC_Stripe_Exception $e, $order )
 	 *
 	 * @param int $order_id   The order ID being processed.
 	 * @param bool $can_retry Should we retry on fail.
+	 *
+	 * @deprecated 10.5.0 This method is deprecated and will be removed in a future release.
 	 */
 	public function process_payment_with_saved_payment_method( $order_id, $can_retry = true ) {
```
daledupreez, line 1557: This method is still called from `process_payment()` on line 969, though we should not ever fall through to that code. Given that we're deprecating this method, it feels like we should also address our own internal caller.

daledupreez (commented):
I am not sure what we should do about `process_payment_with_saved_payment_method()`, as we still have a caller of that method.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: This is part of the flow we are removing in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4929. I tested purchasing with saved payment methods in both the block and shortcode checkout, and this code wasn't reached.

daledupreez (approved):
This is looking good to me! My only concern is around currently unreachable code, which is being cleaned up/removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4929.

PR #4260, opened 2025-04-24
Explore: implement a basic autoloader

daledupreez:
Fixes STRIPE-505

## Changes proposed in this Pull Request:

This PR explores adding a basic autoloader for the plugin largely as a way to avoid unnecessary file loads across the plugin. I considered using some prefix matching before building the classmaps, but I decided against it as the overhead for building the classmaps is one-time, whereas performing multiple string prefix checks for every class we hit (including other classes) feels like it would quickly add up.

Another small change I included in this PR is to remove class instantiation side-effects from the class files so the effect is clearer from outside the class. I made changes so we're now auto-loading all classes, and I've moved the code that instantiated the classes from the respective class files to the main plugin file.

## Testing instructions

There should be no change in behaviour as a result of this PR, as we're simply changing how we load the files containing the classes. Our unit tests and e2e tests should give us a basic safety net for the core autoloader functions, but it's possible that some edge cases exist.

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [N/A] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5012, opened 2026-02-12
Moving class instantiations outside their definition files

wjrosa:
Towards STRIPE-505

As a small improvement towards the autoloader implementation work ([main PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4260)), I am moving class instantiations outside the class definition files.

## Testing instructions

Code review. Perform some basic smoke testing as both a shopper and a merchant to confirm no regression was introduced.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 10.5.0 - xxxx-xx-xx =
-* Update - Moving class instantiations outside their definition files
+* Update - Move class instantiations from their definition files to the plugin initialization code
 * Update - Deprecate unused non-deferred intent methods in WC_Stripe_UPE_Payment_Gateway
 * Dev - Remove unused frontend code related to UPE
 * Update - Remove all deprecated code up to version 10.1.0
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.5.0 - xxxx-xx-xx =
+* Update - Moving class instantiations outside their definition files
 * Dev - Remove unused frontend code related to UPE
 * Update - Remove all deprecated code up to version 10.1.0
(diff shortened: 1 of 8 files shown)


changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.5.0 - xxxx-xx-xx =
+* Update - Moving class instantiations outside their definition files
```
daledupreez: Nit: I think it would be worth clarifying where we are moving them to.
```suggestion
* Update - Move class instantiations from their definition files to the plugin initialization code
```

includes/class-wc-stripe.php
```diff
@@ -137,6 +138,8 @@ public function init() {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-action-scheduler-service.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-webhook-state.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-webhook-handler.php';
+		new WC_Stripe_Webhook_Handler();
```
daledupreez, line 141: It feels like it would _look_ cleaner to have all these instances created in one code block, but there might be some subtle timing issues and/or dependencies, so probably best to keep them inline in this iteration.

daledupreez (approved):
The changes look OK to me, and both e2e tests and unit tests are passing, so I think we should be good.

includes/class-wc-stripe.php
wjrosa: Yeah, I had the same concern. The first version grouped all instantiations below, after the require calls.

malithsen (approved):
This looks good to me.

PR #5017, opened 2026-02-13
Adaptive pricing support check for products in the cart

Mayisha:
Added 'is_adaptive_pricing_supported' method to check if adaptive pricing is supported for the store setting and the products in the cart.

Note:
Updated some ECE test files irrelevant to the changes of this PR to ensure test cleanups, so that test setups do not conflict with the tests added for this PR.

## Testing instructions
- Add the following filter to allow adaptive pricing and checkout sessions in your store-
` add_filter('wc_stripe_is_checkout_sessions_available', '__return_true');`
- As an admin, enable optimized checkout and adaptive pricing in the settings.
- As a shopper, add a simple product to your cart and go to the checkout page.
- In the developer console check the value of `wc_stripe_upe_params.isAdaptivePricingSupported`. The value should be `1`.
- Add a variable product to your cart. The value of `wc_stripe_upe_params.isAdaptivePricingSupported`. The value should be `1`.
- Add a subscription product to your cart. The value of `wc_stripe_upe_params.isAdaptivePricingSupported`. The value should be empty.
- Add a pre-order charged upon release product to your cart. The value of `wc_stripe_upe_params.isAdaptivePricingSupported`. The value should be empty.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-helper.php
```diff
@@ -1105,6 +1105,57 @@ public static function has_cart_or_checkout_on_current_page() {
...
+		// False if not on the checkout page.
+		if ( ! is_checkout() && ! has_block( 'woocommerce/checkout' ) ) {
+			return false;
+		}
+
+		if ( ! WC()->cart || WC()->cart->is_empty() ) {
+			return true;
+		}
+
+		foreach ( WC()->cart->get_cart() as $cart_item_key => $cart_item ) {
```
wjrosa, line 1147: You could also use `WC_Pre_Orders_Cart::cart_contains_pre_order()` and `WC_Subscriptions_Cart::cart_contains_subscription()`.

wjrosa (approved):
Looks good to me 👍 . Highlighted an alternative way to check for the cart contents (but the current version works fine).

Also, tests needs to be fixed before merging

includes/class-wc-stripe-helper.php
Mayisha: Good suggestion.

Mayisha: I ended up not using these methods as there is no existing method to check if the cart contains pre-order paid upon release. As I need to loop through the items anyway, I am keeping the current version.

tests/phpunit/WC_Stripe_Helper_Test.php
```diff
@@ -1095,22 +1095,141 @@ public function test_build_line_items( bool $itemized = false, array $expected_i
...
+			],
+			'pre-order in cart'                    => [
+				'feature_flag'      => true,
+				'is_checkout'       => true,
+				'adaptive_pricing'  => 'yes',
+				'cart_product_type' => 'pre-order',
+				'expected'          => false,
+			],
+		];
+	}
```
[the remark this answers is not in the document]
Mayisha: Added tests for block checkout.

includes/class-wc-stripe-helper.php
```diff
@@ -1105,6 +1105,57 @@ public static function has_cart_or_checkout_on_current_page() {
 		return is_cart() || is_checkout() || has_block( 'woocommerce/cart' ) || has_block( 'woocommerce/checkout' );
 	}
 
+	/**
+	 * Returns whether adaptive pricing is supported for the current checkout.
+	 *
+	 * When on the checkout page, adaptive pricing is not supported if the cart contains
+	 * any of: subscription products or pre-order (charge upon release) products.
```
daledupreez: Nit: I think it would help if the specific items were listed on their own lines.
```suggestion
	 * When on the checkout page, adaptive pricing is not supported if the cart contains
	 * any of the following:
	 *   - A subscription product.
	 *   - A pre-order product that will be charged upon release.
```

Also, is it OK to handle products using a deposit? I don't think we save the payment method in that case, but I worry that Deposits sets the expectation that it is $X today and $Y later, when the actual amounts in local currency may not be $Y on any given future date.

includes/class-wc-stripe-helper.php
daledupreez: I think it would be a good idea to add a comment to mention why we're using the current approach instead of the helper methods.

tests/phpunit/Helpers/WC_Pre_Orders_Product.php
```diff
@@ -0,0 +1,39 @@
...
+
+	/**
+	 * Set the value to be returned from product_is_charged_upon_release().
+	 *
+	 * @param bool $result The value to return from product_is_charged_upon_release().
+	 * @return void
+	 */
+	public static function set_is_pre_order( bool $result ): void {
+		self::$product_is_charged_upon_release_result = $result;
+	}
```
daledupreez, line 28: The naming here feels a bit weird, as we're setting whether the product is charged upon release, not whether the product is a pre-order.
```suggestion
	/**
	 * Set the value to be returned from product_is_charged_upon_release().
	 *
	 * @param bool $result The value to return from product_is_charged_upon_release().
	 * @return void
	 */
	public static function set_is_pre_order_charged_upon_release( bool $result ): void {
		self::$product_is_charged_upon_release_result = $result;
	}
```

tests/phpunit/WC_Stripe_Helper_Test.php
```diff
@@ -784,15 +784,15 @@ public function test_is_webhook_url( $url, $webhook_url, $expected_result ) {
...
 			'webhook URL with empty comparison query should not match' => [ 'https://example.com/test/?foo=bar', 'https://example.com/test/', false ],
-			'webhook URL with missing parameter should not match'      => [ 'https://example.com/test/?wc-api=wc_stripe', 'https://example.com/test/?wc-api=wc_stripe&foo=bar', false ],
-			'webhook URL with wrong parameter should not match'        => [ 'https://example.com/test/?wc-api=wc_stripe_BAD', 'https://example.com/test/?wc-api=wc_stripe', false ],
-			'webhook URL with extra parameters should match'           => [ 'https://example.com/test/?wc-api=wc_stripe&foo=bar', 'https://example.com/test/?wc-api=wc_stripe', true ],
+			'webhook URL with missing parameter should not match' => [ 'https://example.com/test/?wc-api=wc_stripe', 'https://example.com/test/?wc-api=wc_stripe&foo=bar', false ],
+			'webhook URL with wrong parameter should not match' => [ 'https://example.com/test/?wc-api=wc_stripe_BAD', 'https://example.com/test/?wc-api=wc_stripe', false ],
+			'webhook URL with extra parameters should match' => [ 'https://example.com/test/?wc-api=wc_stripe&foo=bar', 'https://example.com/test/?wc-api=wc_stripe', true ],
```
daledupreez: I think the original spacing should be reinstated. Not sure why it got modified.

tests/phpunit/WC_Stripe_Helper_Test.php
```diff
@@ -1095,22 +1095,174 @@ public function test_build_line_items( bool $itemized = false, array $expected_i
...
+		}
+
+		WC()->cart->empty_cart();
+		$product = null;
+		if ( ! empty( $cart_product_type ) ) {
+			$product = WC_Helper_Product::create_simple_product();
+			WC()->cart->add_to_cart( $product->get_id(), 1 );
+		}
+
+		$actual = WC_Stripe_Helper::is_adaptive_pricing_supported();
```
daledupreez, line 1131: As a thought, it may be worth adding tests that check for additional cases that would exercise more logic, including the following:
 - multiple standard products in the cart
 - multiple standard products plus one subscription in the cart

TIL that pre-orders can't be in a cart with other products.

includes/class-wc-stripe-helper.php
```diff
@@ -1105,6 +1105,57 @@ public static function has_cart_or_checkout_on_current_page() {
...
+				continue;
+			}
+
+			// Subscriptions are not supported with adaptive pricing.
+			if ( class_exists( 'WC_Subscriptions_Product' ) && WC_Subscriptions_Product::is_subscription( $product ) ) {
+				return false;
+			}
+
+			// Pre-order (charge upon release) is not supported with adaptive pricing.
+			if ( class_exists( 'WC_Pre_Orders_Product' ) && WC_Pre_Orders_Product::product_is_charged_upon_release( $product ) ) {
```
daledupreez: Two non-blocking thoughts on this front:
 * Might it be worth adding some defensive checks to ensure the methods exist?
 * Might it be worth running the existence checks outside the loop so they only get run once? This is something of a micro-optimization, as class lookups should be quick, but it was something that struck me as unnecessary work.

daledupreez (commented):
This tests as described, but I do have some feedback and suggestions that may be worth incorporating. My biggest concern is around support for Deposits -- I think we should not support deposits, TBH, even if that's mostly a requirements concern.

includes/class-wc-stripe-helper.php
Mayisha: Good call. I have added deposits in the unsupported list in my latest commits.

tests/phpunit/WC_Stripe_Helper_Test.php
Mayisha: Added.

includes/class-wc-stripe-helper.php
Mayisha: > Might it be worth adding some defensive checks to ensure the methods exist?

Skipping this as we are checking for class exist which should be sufficient here.

> Might it be worth running the existence checks outside the loop so they only get run once?

Sounds good to me. Addressed this feedback.

includes/class-wc-stripe-helper.php
daledupreez: It would also be good to make sure that decision/exclusion is pushed up to the requirements discussion/doc as well.

includes/class-wc-stripe-helper.php
```diff
@@ -1135,6 +1138,15 @@ public static function is_adaptive_pricing_supported(): bool {
...
 
+		$subscriptions_available = class_exists( 'WC_Subscriptions_Product' );
+		$pre_orders_available    = class_exists( 'WC_Pre_Orders_Product' );
+		$deposits_available      = class_exists( 'WC_Deposits_Product_Manager' );
+
+		// We loop through cart items instead of using WC_Pre_Orders_Cart::cart_contains_pre_order()
+		// and WC_Subscriptions_Cart::cart_contains_subscription() because we need to detect pre-orders
+		// that are charged upon release specifically (WC_Pre_Orders_Product::product_is_charged_upon_release),
+		// and there is no cart-level helper for that. Looping also lets us check both subscription and
+		// pre-order in a single pass.
```
daledupreez: Not blocking, but it could be worth simplifying this to say we only want one loop through the cart that checks all possible conditions. The comment is also a bit stale, as it doesn't reference deposits.

includes/class-wc-stripe-helper.php
```diff
@@ -1143,12 +1155,17 @@ public static function is_adaptive_pricing_supported(): bool {
 			}
 
 			// Subscriptions are not supported with adaptive pricing.
-			if ( class_exists( 'WC_Subscriptions_Product' ) && WC_Subscriptions_Product::is_subscription( $product ) ) {
+			if ( $subscriptions_available && WC_Subscriptions_Product::is_subscription( $product ) ) { // @phpstan-ignore-line (guarded by class_exists above)
```
daledupreez: Nit: rather than ignoring the line entirely, I think it would be better to ignore only the specific error types that we have checked for. i.e. use `@phpstan-ignore <error.type>`

tests/phpunit/Helpers/WC_Deposits_Product_Manager.php
```diff
@@ -0,0 +1,35 @@
...
+ * This helper class should ONLY be used for unit tests.
+ */
+class WC_Deposits_Product_Manager {
+
+	/**
+	 * The mocked value for deposits_enabled().
+	 *
+	 * @var bool
+	 */
+	public static bool $deposits_enabled_result = false;
```
daledupreez: Nit: should this be public? It feels like access should be via the helper methods.

tests/phpunit/WC_Stripe_Helper_Test.php
```diff
@@ -936,31 +936,31 @@ public function test_get_localized_error_message_from_response_with_unexpected_d
...
 			],
 			'Object response with string error' => [
 				'response'         => (object) [ 'error' => 'Unexpected data' ],
 				'expected_message' => '',
 			],
-			'Object response with array error'  => [
+			'Object response with array error' => [
 				'response'         => (object) [ 'error' => [ 'message' => 'Unexpected data' ] ],
 				'expected_message' => '',
 			],
```
daledupreez, line 966: Nit: it looks like the whitespace for the array keys was unintentionally removed.

tests/phpunit/WC_Stripe_Helper_Test.php
```diff
@@ -1181,103 +1158,168 @@ public function test_is_adaptive_pricing_supported( bool $feature_flag, bool $is
...
-						'label'  => 'Tax',
-						'amount' => 0,
-					],
-					[
-						'key'    => 'total_shipping',
-						'label'  => 'Shipping',
+						'label' => 'Tax',
 						'amount' => 0,
 					],
+					// Shipping line only added when WC()->cart->needs_shipping(); test cart has no address/methods.
```
daledupreez: Are these fixes meant to be in this PR? They feel unrelated to the main changes in the PR.

includes/class-wc-stripe-helper.php
daledupreez: I think it would be more future-proof if we defensively check that the methods exist -- there is no guarantee that sites will have current versions of the relevant plugins, and/or those helpers may be deprecated or removed in future. I would much rather have defensive code in place to start with so the failure mode doesn't involve fatal errors on merchant sites.

daledupreez (changes requested):
The unit test failures seem specific to this branch, so I think we need to address them before we can be confident in these changes.

tests/phpunit/WC_Stripe_Helper_Test.php
Mayisha: The new test I added was actually causing conflict with this test. I have fixed the conflict in the new test and reverted this change.

tests/phpunit/WC_Stripe_Helper_Test.php
Mayisha: These changes are not present in the final diff; probably it is present in one of the commits.

includes/class-wc-stripe-helper.php
Mayisha: Simplified in [7bbdabc](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5017/commits/7bbdabc6eb17759dbfe3d6f8b964bd8830cb8766)

Mayisha:
> The unit test failures seem specific to this branch, so I think we need to address them before we can be confident in these changes.

It was due to some test data conflicts. Added cleanups in the tests to fix the test.

includes/class-wc-stripe-helper.php
```diff
@@ -1105,6 +1105,71 @@ public static function has_cart_or_checkout_on_current_page() {
...
+
+			// Pre-order (charge upon release) is not supported with adaptive pricing.
+			if ( $pre_orders_available && WC_Pre_Orders_Product::product_is_charged_upon_release( $product ) ) { // @phpstan-ignore class.notFound (guarded by class_exists above)
+				return false;
+			}
+
+			// Deposits are not supported with adaptive pricing.
+			if ( $deposits_available && WC_Deposits_Product_Manager::deposits_enabled( $product->get_id() ) && ! empty( $cart_item['is_deposit'] ) ) { // @phpstan-ignore class.notFound (guarded by class_exists above)
+				return false;
+			}
```
daledupreez, line 1167: Not at all blocking, but we can also mention the method_exists() guard.
```suggestion
			// Subscriptions are not supported with adaptive pricing.
			if ( $subscriptions_available && WC_Subscriptions_Product::is_subscription( $product ) ) { // @phpstan-ignore class.notFound (guarded by class_exists() and method_exists() checks above)
				return false;
			}

			// Pre-order (charge upon release) is not supported with adaptive pricing.
			if ( $pre_orders_available && WC_Pre_Orders_Product::product_is_charged_upon_release( $product ) ) { // @phpstan-ignore class.notFound (guarded by class_exists() and method_exists() checks above)
				return false;
			}

			// Deposits are not supported with adaptive pricing.
			if ( $deposits_available && WC_Deposits_Product_Manager::deposits_enabled( $product->get_id() ) && ! empty( $cart_item['is_deposit'] ) ) { // @phpstan-ignore class.notFound (guarded by class_exists() and method_exists() checks above)
				return false;
			}
```

daledupreez (approved):
I haven't retested, but the changes are fairly minimal since my last round of review, so I think we should get this shipped.

PR #5006, opened 2026-02-11
Add adaptive pricing in classic checkout

Mayisha:
Towards [STRIPE-919](https://linear.app/a8c/issue/STRIPE-919)

## Changes proposed in this Pull Request:
- Creates a checkout session on classic checkout.
- Creates the checkout session UI element with `initCheckout`.
- Renders the currency selector on top of the payment methods list (temporary placement until the design is finalized)
- No checkout functionality implemented in this PR.

## Testing instructions
- Checkout to this branch (`add/adaptive-pricing-in-classic-checkout-2`)
- Run the following commands
```
npm install
npm run up
npm start
```
- Add the following filter to allow adaptive pricing and checkout session in your store-
```
add_filter('wc_stripe_is_checkout_sessions_available', '__return_true');
```
- Connect to a US Stripe account.
- Enable OCS in the Stripe settings page.
- Enable adaptive pricing under the OCS settings.
- Enable card and a couple more payment methods.
- As a logged in shopper, add a product to your cart and go to the checkout page. (shopper must be logged in)
- Verify that, the currency selector is displayed and the available currencies are USD and your local currency.
- Verify that changing the currency hides or displays relevant payment methods.
    - For example, my store currency and Stripe account currency is USD, my local currency is BDT and I have card and WeChat enabled in settings. If I select BDT from the currency selector, only card method is displayed. If I select USD, both card and WeChat is displayed.

| BDT | USD |
|-----|------|
| Screenshot 2026-02-11 at 2 09 53 PM | Screenshot 2026-02-11 at 2 10 08 PM |

## Note
1. If the checkout session ajax fails for you, comment out the [nonce check](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-checkout-sessions-controller.php#L28-L30). (nonce issue to be fixed separately) 
2. Needs to be merged after

- https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977
- https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4993

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
[...]

PR #5019, opened 2026-02-13
Payment processing with adaptive pricing in classic checkout

Mayisha:

- In https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5006 I have rendered the price selector for adaptive pricing on the classic checkout page. In this PR, I have implemented payment processing with the selected currency on checkout.
- Used `action.confirm()` from checkout session element to complete payment on Stripe side.
- Introduced `wc-stripe-checkout-session-id` flag to include in the `/checkout` request and use it in backend to go through the checkout session path to process payment.
- Introduced checkout session-related new order metas in `WC_Stripe_Order_Helper` class.
- Added `process_payment_with_checkout_session` method for payment processing.
- Added a helper method `get_woocommerce_amount_from_stripe_amount` to convert the Stripe amount to WooCommerce amount (example: for 2 decimal currencies 7800 is converted to 78.00)
- Added `get_order_by_checkout_session_id` helper method.
- Added webhook handler for `checkout.session.completed` event (other events will be added in a separate PR)

## Testing instructions
- Checkout to this branch
- Run the following commands
```
npm install
npm run up
npm start
```

- Add the following filter to allow adaptive pricing and checkout session in your store-
` add_filter('wc_stripe_is_checkout_sessions_available', '__return_true');`
- Enable adaptive pricing under the OCS settings.
- Enable card and a couple more payment methods.
- Reconfigure the webhooks.
- As a logged in shopper, add a product to your cart and go to the checkout page. (shopper must be logged in)
- Verify that, the currency selector is displayed and the available currencies are USD and your local currency.
- Select your local currency and pay with card.
- As an admin, verify that the order is successful and the transaction details are present.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1097,6 +1097,12 @@ public function process_payment_intent( $notification ) {
 			return;
 		}
 
+		// Update the order with the payment method ID if it's not already set.
```
wjrosa, line 1100: Not sure if it should be part of this PR, but I think we should skip payment intent event webhook handling when the order contains a checkout session ID, to avoid conflicts

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1397,86 @@ public function process_account_updated( $notification ) {
...
+
+		$order->save();
+
+		$charge = $this->get_latest_charge_from_intent( $intent );
+
+		$charge->is_webhook_response = true;
+		$this->process_response( $charge, $order );
+
+		// Unlock the order
+		$order_helper->unlock_order_payment( $order );
```
wjrosa, line 1477: Maybe we should add a call to `run_webhook_received_action()` (like we do in the other events).

includes/class-wc-stripe-order-helper.php
```diff
@@ -484,6 +505,93 @@ public function update_stripe_setup_intent_id( ?WC_Order $order = null, string $
...
 
+	/**
+	 * Gets the Stripe checkout session ID for order.
+	 *
+	 * @since 10.5.0
+	 *
+	 * @param WC_Order|null $order
+	 * @return false|string|null
+	 */
+	public function get_stripe_checkout_session_id( ?WC_Order $order = null ) {
```
wjrosa, line 516: We can use `get_order_meta` here and in the other getter methods

includes/class-wc-stripe-helper.php
```diff
@@ -952,6 +977,49 @@ public static function get_order_by_setup_intent_id( $intent_id ) {
...
 	}
 
+	/**
+	 * Gets the order by Stripe checkout session ID.
+	 *
+	 * @since 10.5.0
+	 * @param string $checkout_session_id The ID of the checkout session.
+	 * @return WC_Order|bool Either an order or false when not found.
+	 */
+	public static function get_order_by_checkout_session_id( $checkout_session_id ) {
```
wjrosa, line 987: I would consider adding this method to `WC_Stripe_Order_Helper` instead, since the original idea was to place these methods there. Also, `$checkout_session_id` lacks the type

includes/class-wc-stripe-helper.php
```diff
@@ -277,6 +277,31 @@ public static function get_stripe_amount( $total, $currency = '' ) {
...
 	}
 
+	/**
+	 * Converts a Stripe amount (smallest currency unit) to WooCommerce amount.
+	 *
+	 * @param int    $stripe_amount Amount in Stripe's smallest unit (e.g. cents).
+	 * @param string $currency      Currency code (e.g. 'eur', 'usd').
+	 * @return string Formatted amount for display.
+	 */
+	public static function get_woocommerce_amount_from_stripe_amount( $stripe_amount, $currency = '' ) {
```
wjrosa, line 287: It feels like we could have some logic shared with `get_stripe_amount`

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1284,6 +1289,36 @@ private function process_payment_with_deferred_intent( int $order_id ) {
...
+	 * @param int $order_id ID of order to be processed.
+	 * @param string $checkout_session_id ID of checkout session to be processed.
+	 *
+	 * @return array An array with the result of the payment processing, and a redirect URL on success.
+	 */
+	private function process_payment_with_checkout_session( int $order_id, string $checkout_session_id ) {
+		$order = wc_get_order( $order_id );
+
+		if ( ! $order instanceof \WC_Order ) {
+			return [];
```
wjrosa, line 1304: Shouldn't we return a failure here instead?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1284,6 +1289,36 @@ private function process_payment_with_deferred_intent( int $order_id ) {
...
+		}
+
+		$order_helper = WC_Stripe_Order_Helper::get_instance();
+		$order_helper->update_stripe_checkout_session_id( $order, $checkout_session_id );
+		$order->save();
```
wjrosa, line 1309:
```suggestion
		$order->save_meta_data();
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1284,6 +1289,36 @@ private function process_payment_with_deferred_intent( int $order_id ) {
...
 
+	/**
+	 * Process the payment for an order that has a checkout session attached.
+	 *
+	 * @param int $order_id ID of order to be processed.
+	 * @param string $checkout_session_id ID of checkout session to be processed.
+	 *
+	 * @return array An array with the result of the payment processing, and a redirect URL on success.
+	 */
+	private function process_payment_with_checkout_session( int $order_id, string $checkout_session_id ) {
```
wjrosa, line 1300:
```suggestion
	private function process_payment_with_checkout_session( int $order_id, string $checkout_session_id ): array {
```

client/stripe-utils/utils.js
```diff
@@ -326,6 +326,12 @@ export const appendSetupIntentToForm = ( form, setupIntent ) => {
 	);
 };
 
+export const appendCheckoutSessionIdToForm = ( form, checkoutSessionId ) => {
+	form.append(
+		`<input type="hidden" id="wc-stripe-checkout-session-id" name="wc-stripe-checkout-session-id" value="${ checkoutSessionId }" />`
```
wjrosa, line 331: I know we don't have standards in this sense, but personally, I think it is better to have snake_case for name attributes

client/classic/upe/payment-processing.js
```diff
@@ -190,12 +191,12 @@ async function createStripePaymentElement( api, paymentMethodType ) {
 	}
 
 	let elements;
-	let shouldLoadStripeElements = ! isAdaptivePricingSupported;
+	let shouldLoadStripeElements = true;
```
wjrosa, line 194: Is this intentional?

client/classic/upe/payment-processing.js
```diff
@@ -592,45 +607,65 @@ export const processPayment = (
...
-			appendPaymentMethodIdToForm(
-				jQueryForm,
-				paymentMethodObject.paymentMethod.id
-			);
+				const { actions } = loadActionsResult;
 
-			// Append the intent ID to the form if it was previously created through a non-deferred intent.
-			if ( gatewayUPEComponents[ paymentMethodType ].intentId ) {
-				appendPaymentIntentIdToForm(
+				const confirmResult = await actions.confirm( {
```
wjrosa, line 619: Something strange that happens, at least with the block checkout, is that `confirm` kills the code somehow if it does not return an error, so `confirmResult.session.id` wouldn't be available below (not sure if the same happens here). That's why I had to use states to set the session ID

wjrosa (approved):
Nice work! Left some optional suggestions and comments

includes/class-wc-stripe-helper.php
Mayisha: @wjrosa this method is actually the opposite of `get_stripe_amount` method. 
`get_stripe_amount` converts the WooCommece amount to Stripe supported format.
This method converts the Stripe amount to WooCommerce amount.

includes/class-wc-stripe-webhook-handler.php
Mayisha: `run_webhook_received_action` is called at the end of `process_webhook` method which calls this `process_checkout_session` first. So it is already present in the same flow.

includes/class-wc-stripe-webhook-handler.php
Mayisha: Good call but let's explore it in a follow up. So far, what I have seen is that the payment intent event arrives first and does not have any order information. The checkout session event arrives after that which contains the order data.

wjrosa: Yeah, I saw the same behavior. It throws an error and pollutes the logs, basically.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Oh right. We are covered then

includes/class-wc-stripe-helper.php
wjrosa: Gotcha! It makes sense to leave it this way then

Mayisha:
Closing this in favor of
- https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5102

I have split the implementation into frontend and backend in the two PRs above.

PR #5025, opened 2026-02-15
Fix fatal error caused by webhook ID collisions

### LINKED ISSUE #4951: Webhook processing triggers fatal errors when order ID references a refund
### Description

In situations where multiple sites are all using a single Stripe account, all the sites will receive webhooks for the account. This means that many received webhooks will not apply to the current site. For the most part, we already protect against this situation, as we verify that we have a record of the intent we are processing and that it is linked to the order that is referenced in the signature metadata that include in the webhook. However, we can trigger fatal errors when the order ID references a refund.

### Steps to reproduce

This requires a pretty complex setup to trigger, so I don't know that we will be able to reproduce it directly.

### Expected behavior

We should have better guards against this kind of order ID collision, as this should be an ignored webhook, not a fatal error.

### Additional context

**Internal discussion/investigation:** p1769154590337939-slack-C055WHLA98D

#### Code

The reported fatal was in `WC_Stripe_Webhook_Handler->get_order_from_intent()` on line 1521 in the source below. However there are additional calls to `wc_get_order()` on lines 1540 and 1548 that may also require additional guards for refunds.

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/7dadc8fa45b38979bab18e94685ecf3461c9abcd/includes/class-wc-stripe-webhook-handler.php#L1509-L1553

HomerSim:
Fixes #4951

### Description:
This PR addresses a fatal `TypeError` in `get_order_from_intent()`. In multi-site environments sharing a single Stripe account, webhooks intended for Site A are also delivered to Site B. If an ID collision occurs where the ID on Site B refers to a `WC_Order_Refund` object, it triggers a fatal error because subsequent methods strictly expect a `WC_Order` instance.

I added a strict type check `if ( ! $order instanceof WC_Order )` to ensure only valid order objects are processed, allowing the system to gracefully skip refunds or other object types. Additionally, I updated `phpstan-baseline.neon` to reflect the improved type safety.

### Questions for the PR author:
- **How can this code break?:** If `wc_get_order()` returns a non-order object, it could proceed to logic that expects an order.

- **What are we doing to make sure this code doesn't break?:** We added an explicit `instanceof` guard clause, which is the most robust way in PHP to ensure type safety before calling methods dependent on that type.

### Reproduction Confirmed
I successfully reproduced this issue using a local multi-site setup to simulate ID collisions.

### Setup
Two WordPress instances sharing a single Stripe test account to simulate ID collisions:

#### 1. Webhook Configuration
Both sites registered as webhook endpoints in the Stripe Dashboard:
Webhook Setup

#### 2. Database ID Collision Evidence
- **Site A (ngrok):** Contains **Refund #1000040**
image

- **Site B (Cloudflare Tunnel):** Contains **Order #1000040**
image

> *Note: These screenshots confirm that the same ID (#1000040) points to different object types across the two sites.*

### Reproduction Steps
1. Processed a payment for Order #1000040 on Site B.
2. Stripe delivered webhooks to both sites.

3. **Result (Before Fix):** Site A encountered a Fatal Error while processing an irrelevant webhook due to an ID collision.
image

```php
[15-Feb-2026 13:40:47 UTC] PHP Fatal error:  Uncaught TypeError: WC_Stripe_Order_Helper::get_intent_id_from_order(): Argument #1 ($order) must be of type WC_Order, Automattic\WooCommerce\Admin\Overrides\OrderRefund given, called in /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/includes/class-wc-stripe-webhook-handler.php on line 1536 and defined in /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/includes/class-wc-stripe-order-helper.php:918
Stack trace:
#0 /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/includes/class-wc-stripe-webhook-handler.php(1536): WC_Stripe_Order_Helper->get_intent_id_from_order(Object(Automattic\WooCommerce\Admin\Overrides\OrderRefund))
#1 /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/includes/class-wc-stripe-webhook-handler.php(1089): WC_Stripe_Webhook_Handler->get_order_from_intent(Object(stdClass))
#2 /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/includes/class-wc-stripe-webhook-handler.php(1476): WC_Stripe_Webhook_Handler->process_payment_intent(Object(stdClass))
#3 /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/includes/class-wc-stripe-webhook-handler.php(132): WC_Stripe_Webhook_Handler->process_webhook('{\n  "id": "evt_...')
#4 /var/www/html/wp-includes/class-wp-hook.php(341): WC_Stripe_Webhook_Handler->check_for_webhook('')
#5 /var/www/html/wp-includes/class-wp-hook.php(365): WP_Hook->apply_filters('', Array)
#6 /var/www/html/wp-includes/plugin.php(522): WP_Hook->do_action(Array)
#7 /var/www/html/wp-content/plugins/woocommerce/src/Internal/Utilities/LegacyRestApiStub.php(166): do_action('woocommerce_api...')
#8 /var/www/html/wp-content/plugins/woocommerce/src/Internal/Utilities/LegacyRestApiStub.php(95): Automattic\WooCommerce\Internal\Utilities\LegacyRestApiStub::maybe_process_wc_api_query_var()
#9 /var/www/html/wp-includes/class-wp-hook.php(341): Automattic\WooCommerce\Internal\Utilities\LegacyRestApiStub::parse_legacy_rest_api_request(Object(WP))
#10 /var/www/html/wp-includes/class-wp-hook.php(365): WP_Hook->apply_filters(NULL, Array)
#11 /var/www/html/wp-includes/plugin.php(570): WP_Hook->do_action(Array)
#12 /var/www/html/wp-includes/class-wp.php(418): do_action_ref_array('parse_request', Array)
#13 /var/www/html/wp-includes/class-wp.php(821): WP->parse_request('')
#14 /var/www/html/wp-includes/functions.php(1343): WP->main('')
#15 /var/www/html/wp-blog-header.php(16): wp()
#16 /var/www/html/index.php(17): require('/var/www/html/w...')
#17 {main}
  thrown in /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/includes/class-wc-stripe-order-helper.php on line 918
```

4. **Result (After Fix):** Site A ignores the webhook gracefully as it identifies the object as a Refund.

### Evidence
The red box highlights the entire set of webhooks triggered by the same transaction. You can see the transition from **500 (Fatal Error)** to **200 (Success)** after applying the type check guard.

image

image

image

---

### Changelog entry

<details>

Fix - Prevented fatal errors in webhook processing when an ID collision occurs with a refund object.
</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (Detailed instructions provided in the PR description above)


daledupreez (approved):
Thanks for the fix, @HomerSim! Much appreciated!

This is looking good to me. (The E2e test failures are due to permissions on some of the data, so nothing to be concerned about.)

Mayisha (approved):
Thanks for the fix. This looks good to me 👍

HomerSim:
Thanks for the review and the warm welcome, @daledupreez  @Mayisha I'm glad to contribute to the project. 😊

PR #5027, opened 2026-02-16
Agentic Commerce: Checkout session webhook handler

RadoslavGeorgiev:
Fixes STRIPE-900

Adds the webhook handler plumbing for agentic commerce order processing. When an AI agent completes a purchase through Stripe's hosted checkout, Stripe sends a `checkout.session.completed` webhook. Unlike the normal flow where the WooCommerce order already exists, agentic orders must be **created** from the webhook data.

This PR adds the entry point only — it delegates order creation to an order mapper class that will be implemented in STRIPE-902 and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032. The mapper stub currently throws, which the handler catches gracefully.

**What's added:**

- The `process_checkout_session_completed` for handling `checkout.session.completed` in the webhook handler.
- Webhook API version updates in the `WC_Stripe_Account` class.

Note that https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032 was already built on top of this PR. In the latest commits, I brought over some changes from that PR to this one to keep things consistent.

## Testing instructions

This PR is not testable end-to-end on its own because the order mapper is a stub. It becomes fully testable once STRIPE-902 is merged. For now, you can still get to the point where we are about to map the incoming checkout session to an order.

### Account setup

We cannot test this through a connected account, as Stripe has not built that functionality yet. In the interim:

1. Log into the Stripe dashboard.
2. Create a new sandbox.
3. Switch to the sandbox.
4. Navigate to Developers (bottom right) -> API Keys. You should see the publishable and secret keys for the account.
5. In WP Admin, navigate to _WC Stripe Dev_ -> Gateway Settings.
6. Edit the array, replacing `test_publishable_key` and `test_secret_key` with the values from the Stripe dashboard.
7. Go to the _Stripe Account_ section and click the _clear_ link.

At this point, you are connected to the sandbox platform account.

### Testing this PR

1. You need a tunnel. `npm run jt:setup`, then `npm run jt:start`.
2. Set up your webhooks. Regardless of whether you have some already, you need to do this, as there is a new event that we're accepting. Navigate to:
    1. wp-admin
    2. Payments
    3. Click `Manage` on the Stripe row.
    4. Go to the Settings tab.
    5. Click `Configure Connection` in the `Account Details` section.
    6. Click `Reconfigure Webhooks`
    7. Make sure that `Log debug messages` near the bottom of the page is enabled.
3. Navigate to WooCommerce -> Status -> Scheduled Actions -> Pending and click the "Run" link next to the `wc_stripe_agentic_commerce_sync_feed` action. This should upload your local catalog to Stripe.
3. In the Stripe dashboard, go to _Payments -> Agentic Commerce -> External Catalogue_.
4. Click _View Catalogue_
5. Click on any product, and then _Test checkout in Workrench_
6. Add an xDebug breakpoint [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/b8180124f113763b957474c8d946f75a51ef36ef/includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php#L32-L39).
7. Run the flow in workbench.

---

- [x] Covered with tests
- [ ] Tested on mobile (does not apply — server-side webhook handling)

### Changelog entry

- [x] This Pull Request does not require a changelog entry.

<details>

#### Comment

Webhook handler plumbing only — no user-facing behavior until STRIPE-902 (order mapper) is implemented. Will be included in the agentic commerce changelog entry when the full feature ships.

</details>

**Post merge**


RadoslavGeorgiev:
@coderabbitai review

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1572,4 +1577,141 @@ private function get_payment_method_type_from_charge( $charge ) {
...
+					[
+						'session_id'        => $checkout_session->id,
+						'payment_intent_id' => $payment_intent_id,
+						'order_id'          => $existing_order->get_id(),
+					]
+				);
+				$this->resolved_order = $existing_order;
+				return;
+			}
+		}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: One of the next issues requires the order and currently processing webhooks to be put in a session. That will make this check redundant. For now, all we need is to be able to start mapping checkout sessions to orders.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -0,0 +1,41 @@
...
+	 * @throws Exception When the order cannot be created.
+	 */
+	public function create_order_from_checkout_session( object $checkout_session ): WC_Order {
+		throw new Exception(
+			esc_html(
+				sprintf(
+					'Agentic commerce order mapper not yet implemented. Session: %s',
+					$checkout_session->id ?? 'unknown'
+				)
+			)
```
RadoslavGeorgiev: Note: The final implementation (with multiple approvals already) is in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032. Those two PRs will be merged almost simultaneously.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1578,4 +1583,174 @@ private function get_payment_method_type_from_charge( $charge ) {
...
+		// Temporarily override the API version to get the right fields.
+		$override_version = function ( $headers ) {
+			$headers['Stripe-Version'] = WC_Stripe_API::AGENTIC_COMMERCE_API_VERSION;
+			return $headers;
+		};
+		add_filter( 'wc_stripe_request_headers', $override_version );
+
+		try {
+			$url         = $this->build_checkout_session_retrieve_url( $notification->data->object->id );
+			$raw_session = WC_Stripe_API::retrieve( $url );
```
RadoslavGeorgiev, line 1819: This is `$raw_session` because https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032 introduces a DTO that will be `$session`.

RadoslavGeorgiev:
@daledupreez @wjrosa this PR is ready for review. I was waiting for a way to get the `agent_details` field of the payment intent, and now we have it. I've updated the PR with it.

Note that you reviewed a good portion of this work in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032 already.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
daledupreez: That PR is based off this branch, so we should merge it first and then we can merge this branch into `develop`.

includes/class-wc-stripe-api.php
```diff
@@ -13,8 +13,9 @@ class WC_Stripe_API {
 	/**
 	 * Stripe API Endpoint
 	 */
-	const ENDPOINT           = 'https://api.stripe.com/v1/';
-	const STRIPE_API_VERSION = '2025-09-30.clover';
+	const ENDPOINT                     = 'https://api.stripe.com/v1/';
+	const STRIPE_API_VERSION           = '2025-09-30.clover';
+	const AGENTIC_COMMERCE_API_VERSION = '2025-10-29.clover';
```
daledupreez: @wjrosa, @Mayisha, and @RadoslavGeorgiev , do we have a tracking issue somewhere to ensure that we normalize/collapse these values? It feels like that would be a really good thing to track and not forget, even if we can't get to it for a while.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1578,4 +1583,174 @@ private function get_payment_method_type_from_charge( $charge ) {
...
+					'Failed to retrieve checkout session with expand params.',
+					[
+						'url'   => $url,
+						'error' => is_wp_error( $raw_session ) ? $raw_session->get_error_message() : 'Unexpected response from Stripe API.',
+					]
+				);
+				return;
+			}
+
+			if ( empty( $raw_session->payment_intent->agent_details->network_business_profile ?? '' ) ) {
```
daledupreez: It feels like it would be a good idea to add a debug logging entry for this case, as things could mysteriously fail without the logging details.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1578,4 +1583,174 @@ private function get_payment_method_type_from_charge( $charge ) {
...
+				);
+
+				/**
+				 * Fires after an agentic commerce order is created from a checkout session.
+				 *
+				 * @since 10.6.0
+				 * @param WC_Order $order   The created order.
+				 * @param object   $session The checkout session wrapper.
+				 */
+				do_action( 'wc_stripe_agentic_order_created', $order, $raw_session );
```
daledupreez: Do we want to pass in the raw session data? Or should we pass in the session DTO you're working on? (Maybe that's already in the follow-up PR, but want to ask!)

And is it worth passing in either the webhook payload and/or the checkout session ID?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1578,4 +1583,174 @@ private function get_payment_method_type_from_charge( $charge ) {
...
+		);
+
+		// Temporarily override the API version to get the right fields.
+		$override_version = function ( $headers ) {
+			$headers['Stripe-Version'] = WC_Stripe_API::AGENTIC_COMMERCE_API_VERSION;
+			return $headers;
+		};
+		add_filter( 'wc_stripe_request_headers', $override_version );
+
+		try {
```
daledupreez, line 1814: We should initialise `$raw_session` outside the `try` block, as we reference the value in the `catch`, and it is not created immediately in the `try` block
```suggestion
		$raw_session = null;
		try {
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1578,4 +1583,174 @@ private function get_payment_method_type_from_charge( $charge ) {
...
+				);
+
+				/**
+				 * Fires when agentic commerce order creation fails.
+				 *
+				 * @since 10.6.0
+				 * @param Exception $e       The exception that was thrown.
+				 * @param object    $session The checkout session wrapper.
+				 */
+				do_action( 'wc_stripe_agentic_order_creation_failed', $e, $raw_session );
```
daledupreez: As above RE whether we're going to pass in `$session` or `$raw_session`, as well as whether it makes sense to pass in the initial webhook payload and/or the checkout session ID.

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
```diff
@@ -0,0 +1,407 @@
...
+		WC_Stripe_Database_Cache::set( $lock_key, time(), 5 * MINUTE_IN_SECONDS );
+
+		$notification = $this->build_notification( $session_id );
+		$this->handler->process_checkout_session_completed( $notification );
+
+		$resolved = $this->get_resolved_order();
+		$this->assertNull( $resolved );
+
+		// Clean up.
+		WC_Stripe_Database_Cache::delete( $lock_key );
```
daledupreez: Nit: Please clean up before asserting, or shift to a `try/finally`.
```suggestion
		$resolved = $this->get_resolved_order();

		// Clean up.
		WC_Stripe_Database_Cache::delete( $lock_key );

		$this->assertNull( $resolved );
```

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
```diff
@@ -0,0 +1,407 @@
...
+		// Verify no new orders were created.
+		$orders = wc_get_orders(
+			[
+				'meta_key'   => '_stripe_intent_id', // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_key
+				'meta_value' => 'pi_test_duplicate', // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_value
+			]
+		);
+		$this->assertCount( 1, $orders );
+
+		$existing_order->delete( true );
```
daledupreez: Nit: We should clean up before we assert, or use a try/finally.

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
```diff
@@ -0,0 +1,407 @@
...
+		$action_fired = false;
+		add_action(
+			'wc_stripe_agentic_order_creation_failed',
+			function () use ( &$action_fired ) {
+				$action_fired = true;
+			}
+		);
+
+		$this->handler->process_webhook( $body );
+		$this->assertTrue( $action_fired );
```
daledupreez: Nit: I think we should clean up the action before we assert.

We could also verify that the expected arguments are passed to the action.

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
```diff
@@ -0,0 +1,407 @@
...
+	/**
+	 * Gets the resolved_order from the handler via reflection.
+	 *
+	 * @return WC_Order|null
+	 */
+	private function get_resolved_order() {
+		$prop = new \ReflectionProperty( WC_Stripe_Webhook_Handler::class, 'resolved_order' );
+		$prop->setAccessible( true );
+		return $prop->getValue( $this->handler );
+	}
```
daledupreez, line 493: Not blocking, but it may be clearer and more explicit if we pass in the handler instance rather than having an implicit dependency on `$this->handler` -- this is doubly implicit because the calling code needs to ensure that it is using and modifying `$this->handler`, and then calling this function to extract data from it.
```suggestion
	/**
	 * Gets the resolved_order from the handler via reflection.
	 *
	 * @param WC_Stripe_Webhook_Handler $webhook_handler Webhook handler instance.
	 * @return WC_Order|null
	 */
	private function get_resolved_order( WC_Stripe_Webhook_Handler $webhook_handler ) {
		$prop = new \ReflectionProperty( WC_Stripe_Webhook_Handler::class, 'resolved_order' );
		$prop->setAccessible( true );
		return $prop->getValue( $webhook_handler );
	}
```

daledupreez (approved):
This is looking good overall, but I do have some non-blocking feedback and suggestions.

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
```diff
@@ -0,0 +1,407 @@
...
+		// resolved_order should remain null — nothing processed.
+		$resolved = $this->get_resolved_order();
+		$this->assertNull( $resolved );
+	}
+
+	/**
+	 * Tests that concurrent duplicate webhooks are blocked by the lock.
+	 */
+	public function test_concurrent_duplicate_blocked_by_lock() {
+		$this->enable_feature_flag();
```
daledupreez: One more thought: can we move this filter setup to `set_up()`? It feels like it belongs there.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: Hm, I haven't tried to merge a nested branch into `develop` first, and then the main one. Trying it out...

RadoslavGeorgiev:
Note: I merged https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032 into `develop`, meaning that the changes from this PR have been merged. 

This PR will be used to address the non-blocking comments from [this review](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5027#pullrequestreview-3903809745).

includes/class-wc-stripe-api.php
RadoslavGeorgiev: Created STRIPE-981 to track this, once we are close to going live with agentic commerce.

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: The DTO is now being passed (since the follow-up is merged into develop).

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: I'm not sure how it was in this branch when you left the comment, but with the update, the assignment is within the outer `try` block, which only has a `finally`. If something fails with its creation, we have a log & return sequence.

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: It is the full DTO that is being passed, given that the other PR was merged.

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5027/commits/7b677f2a85f3ec11daf26168f928cc20d6fa4802

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
RadoslavGeorgiev: This test is gone now, as Mayisha's changes already include the check for the order.

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5027/commits/52a6a54a4edb3790f4e7f6ede044e3d2cac739a7

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
RadoslavGeorgiev: This test has also been removed, as there is a whole separate handler for the standard checkout session completed webhook.

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
RadoslavGeorgiev: Done [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/52a6a54a4edb3790f4e7f6ede044e3d2cac739a7/tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php#L42)

RadoslavGeorgiev:
@wjrosa I re-requested your review, as Eric is not available at the moment, but merge still requires two approvals. It's pretty quick to review now :)

Screenshot 2026-03-09 at 14 39 19

wjrosa (approved):
Changes look good to me! 👍

PR #5029, opened 2026-02-16
Dev: Require changelog on push to develop

ebinnion:

This PR adds a guard to the pre-push hook in `.husky/pre-push` so pushes are blocked when changelog files are missing.

- Checks the files in the commit(s) being pushed (both existing branch updates and first push of a new branch).
- Fails the push unless both `readme.txt` and `changelog.txt` are included.
- Shows a clear remediation message: `Run: npm run changelog`.
- Skips this check for branch deletion pushes.
- Keeps existing pre-push PHPStan behavior intact.

How can this code break?
- It may block pushes that intentionally do not need changelog updates.
- Edge cases in commit-range detection could cause false positives.

What are we doing to make sure this code doesn’t break?
- Validated shell syntax (`sh -n`).
- Simulated pre-push input and confirmed:
  - fails when `readme.txt`/`changelog.txt` are missing,
  - passes when both are included,
  - skips check for branch deletion pushes.

## Testing instructions

1. Create a commit that does not touch `readme.txt` or `changelog.txt`.
2. Run `git push origin <branch>`.
3. Confirm the push fails and shows:
   - `Error: Missing changelog files in this push.`
   - `Run: npm run changelog`
4. Run `npm run changelog` and commit the updates to `readme.txt` and `changelog.txt`.
5. Push again and confirm it succeeds.
6. Optional: run `git push origin --delete <branch>` and confirm deletion is not blocked by this check.

---

### Changelog entry

<details>

#### Comment

This is an internal developer workflow change (git hook only) and does not affect plugin runtime behavior.

</details>

**Post merge**

## Summary by CodeRabbit

* **Chores**
  * Added pre-push validation to enforce changelog and documentation file updates before code contributions

* **Documentation**
  * Updated release notes and changelog entries reflecting latest development workflow requirements

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -5,13 +5,25 @@ zero=$(git hash-object --stdin </dev/null | tr '[0-9a-f]' '0')
 
 while read local_ref local_oid remote_ref remote_oid
 do
-	range=""
-	if [ "$local_oid" != "$zero" ]
+	if [ "$local_oid" = "$zero" ]
 	then
-		if test "$remote_oid" != "$zero"
+		# Deleting a remote ref; skip changelog and PHPStan checks.
+		continue
+	fi
+
+	range=""
+	changed_files=""
+	if [ "$remote_oid" != "$zero" ]
+	then
+		# Update to existing branch - check all new commits.
+		range="$remote_oid..$local_oid"
+	else
+		# New remote ref - inspect commits not yet on any remote.
+		commits_to_push=$(git rev-list "$local_oid" --not --remotes)
+		if [ -n "$commits_to_push" ]
 		then
-				# Update to existing branch - check all new commits.
-				range="$remote_oid..$local_oid"
+			# Use awk as grep has a non-zero exit code when no matches exist, which causes the pre-push hook to fail.
+			changed_files=$(echo "$commits_to_push" | xargs git diff-tree --no-commit-id --name-only -r | awk '!seen[$0]++')
 		fi
 	fi
 
@@ -46,8 +58,9 @@ do
 			exit 1
 		fi
 
-		# Get the changed files.
+		# Get the changed files in the push.
 		changed_files=$(git diff --name-only "$range")
+
 		# Get the changed PHP files or directories in the push.
 		# Use awk as grep has a non-zero exit code when no matches exist, which causes the pre-push hook to fail.
 		changed_php_files=$(echo "$changed_files" | awk '/.*(\.php|\/)$/ {print $0;}')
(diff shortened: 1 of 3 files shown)


malithsen (commented):
I like the idea of pushing changelog checks closer to the developer to catch omissions early. Only concern I have, which you've already noted in the PR description, is around legitimate scenarios where a changelog update isn't needed (eg: fixing a bug in an unreleased feature). Right now, that's as easy as ticking
a checkbox on the PR template, and CI changelog validation will be omitted. With this pre-push hook, the only bypass that I am aware of is via `--no-verify`, which will also skip PHPStan.

Would it be worth making this a warning rather than a hard block?

.husky/pre-push
```diff
@@ -65,6 +78,23 @@ do
 		fi
 	fi
 
+	if [ "$remote_ref" = "refs/heads/develop" ] && [ -n "$range" ]
```
daledupreez, line 81: I think using `develop` as the target might miss an edge case where we are pushing to a branch that targets another PR branch. That said, in those situations this would likely be a follow-up PR that builds on the underlying PR, so a changelog may or may not make sense.

Another minor edge case would be PRs that are aimed specifically at a release branch first and will then get ported to `develop`. Mostly that would be release note PRs, with one or two bug-fixes, however, so also not a major concern.

So this probably works for most situations.

.husky/pre-push
```diff
@@ -65,6 +78,23 @@ do
...
+	then
+		if [ -n "$changed_files" ]
+		then
+			has_readme_update=$(echo "$changed_files" | awk '/^readme\.txt$/ { found=1 } END { print found+0 }')
+			has_changelog_update=$(echo "$changed_files" | awk '/^changelog\.txt$/ { found=1 } END { print found+0 }')
+
+			if [ "$has_readme_update" -ne 1 ] || [ "$has_changelog_update" -ne 1 ]
+			then
+				echo "Error: Missing changelog files in this push to develop."
```
daledupreez, line 90: To @malithsen's point [about this being a hard stop](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5029#pullrequestreview-3810901999), might it make more sense to run this check after the PHPStan checks? Static analysis issues should be a hard(er) stop, but running these checks first may mean that we unnecessarily prevent the analysis from running.

That said, PHPStan is slow, which is why we have it running on pre-push and not pre-commit. So I can see a quick changelog check being worth running first, as the feedback cycle is quicker.

.husky/pre-push
daledupreez: Checking for `[ "$remote_ref" = "refs/heads/develop" ]` means we bypass this for most workflows, as the remote ref will be a different branch name. In my workflow, I push up to a branch name that mirrors my local branch name, so `remote_ref=refs/heads/<my_branch_name>`.

Maybe we need to change this to check if `[ "$remote_oid" = "$zero" ]` to work out whether we are pushing to a new branch that isn't present on the remote?

daledupreez (changes requested):

ebinnion:
I converted back to draft to consider.

PR #5032, opened 2026-02-18
Agentic Commerce: Initial order processing

RadoslavGeorgiev:
Fixes STRIPE-902

Maps a completed Stripe checkout session into a WooCommerce order. This is the core order-creation logic for agentic commerce — when an AI agent completes a purchase through Stripe, the webhook handler (from STRIPE-900) calls the mapper to build the corresponding WC order.

**New classes:**
- `WC_Stripe_Agentic_Checkout_Session` — typed wrapper around the raw Stripe session object. Provides getter methods with proper return types and fallback logic (e.g. `shipping_details` → `collected_information.shipping_details`), keeping all Stripe data-shape knowledge in one place.
- `WC_Stripe_Agentic_Line_Item` — typed wrapper around a single Stripe line item. Encapsulates null guards on the `price` sub-object and provides getters for `id`, `description`, `quantity`, `amount_total`, `amount_tax`, and `product_id`.
- `WC_Stripe_Agentic_Commerce_Order_Mapper` — takes the session wrapper and creates a WC order: validates required fields, maps line items (using `external_reference` → product ID), maps billing/shipping addresses, stores Stripe metadata, and verifies the order total matches Stripe's amount. Cleans up the order on failure (deletes partially-created orders before re-throwing).

**Other changes:**
- `WC_Stripe_Helper::convert_from_stripe_amount()` — new helper for Stripe → WC amount conversion (handles zero-decimal, two-decimal, and three-decimal currencies).
- Webhook handler updates — creates the session wrapper, uses `$session->is_agentic()` to detect agentic sessions, passes the wrapper to the mapper. API version override wrapped in `try/finally` to prevent filter leak.
- Order mapper uses `payment_complete()` instead of `set_status('processing')` — properly sets `date_paid`, reduces stock, and fires `woocommerce_payment_complete`.

**Not Included:**

Anything related to tax or shipping. Those will be implemented properly in the mapper and checkout when we implement the related hooks.

## Testing instructions

Don't forget to use the right Stripe account. You can find it here: pc4etw-2ax-p2

Automated mostly — the mapper and session wrapper are fully covered by unit tests (220 tests across four test files). No UI or checkout flow changes.

1. `wp option update _wcstripe_feature_agentic_commerce yes`.
2. Navigate to WooCommerce -> Settings -> Payments -> Stripe -> Settings -> Configure Connection -> Reconfigure Webhooks.
3. Make sure that you are tunneled for ☝️ , and that your actual catalog is uploaded to the test site.
4. Make a purchase on the site.
5. Check your local order. The shipping details, main line item, and totals should match.

---

### Changelog entry

<details>

#### Comment

Internal agentic commerce feature behind a feature flag — no user-facing changes.

</details>

**Post merge**


RadoslavGeorgiev:
Note to reviewers: This PR is on top of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5027 which is still in an unclear state. I might need to later tweak it a bit to allow for detection but don't mind that yet.

ebinnion:
@coderabbit review

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,441 @@
...
+		}
+
+		$this->map_address(
+			$order,
+			$shipping_address,
+			$session->get_shipping_name(),
+			$session->get_shipping_phone(),
+			'shipping'
+		);
+	}
```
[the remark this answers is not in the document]
ebinnion: This also came up in my codex review as a high level blocker for blocking digital goods. Of note, ACP doesn't suggest that fulfillment information is required:

https://www.agenticcommerce.dev/docs/reference/checkout#fulfillmentdetails

includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
```diff
@@ -0,0 +1,253 @@
...
+/**
+ * Provides typed access to Stripe checkout session data.
+ *
+ * The Stripe API returns untyped objects from json_decode(). This class
+ * wraps that raw object and provides small, testable getter methods with
+ * proper return types and fallback logic.
+ *
+ * @since 10.5.0
+ */
+class WC_Stripe_Agentic_Checkout_Session {
```
wjrosa, line 24: Nice! I wish we had more DTOs like this one in the extension

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,441 @@
...
+	 * @param string   $name    The name of the address to map.
+	 * @param string   $phone   The phone number of the address to map.
+	 * @param string   $type    The type of address to map ('billing' or 'shipping').
+	 */
+	private function map_address(
+		WC_Order $order,
+		object $address,
+		string $name,
+		string $phone,
+		string $type = 'billing'
```
wjrosa: Maybe we should add a constant in this class for this

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,441 @@
...
+		$customer_id = $session->get_customer_id();
+		if ( '' !== $customer_id ) {
+			$order_helper->update_stripe_customer_id( $order, $customer_id );
+		}
+
+		// Store Stripe currency.
+		$order_helper->update_stripe_currency( $order, $session->get_currency_lowercase() );
+
+		// Store checkout session ID for traceability.
+		$order->update_meta_data( '_stripe_checkout_session_id', $session->get_id() );
```
wjrosa: You still need to call `save_meta_data()`.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1599,48 +1599,86 @@ public function process_checkout_session_completed( $notification ): void {
...
+					'url'      => $url,
+					'expanded' => $expand,
+					'error'    => is_wp_error( $raw_session ) ? $raw_session->get_error_message() : 'Unexpected response from Stripe API.',
+				]
+			);
 			return;
 		}
 
-		$payment_intent_id = $this->get_payment_intent_id_from_checkout_session( $checkout_session );
+		$session = new WC_Stripe_Agentic_Checkout_Session( $raw_session );
```
wjrosa: Thinking further, maybe it is worth calling it just `WC_Stripe_Checkout_Session`, and move to a new `dto` folder, so we can reuse this for the main checkout sessions integration (and later expand with other classes for charges and payment intents).

wjrosa (approved):
Looks good! Left some comments

includes/class-wc-stripe-webhook-handler.php
wjrosa: I made a POC to illustrate an abstract class https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5045

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/f80a6601d2c85c3d578ae4cd4ef7bf08c4911d00

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: I was actually trying to avoid too many `->save()` calls, even if just for meta data. In https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/893cbb1f13139913aef78adac85815f454db08d0, I updated the logic a bit so it looks like this:

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/b2d63d5fe671ded2d309002a89cc7996e0312117/includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php#L47-L65

WDYT?

RadoslavGeorgiev:
@CodeRabbit review

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1599,48 +1599,86 @@ public function process_checkout_session_completed( $notification ): void {
...
+			);
+			$this->resolved_order = $existing_order;
+			return;
 		}
 
 		try {
 			$order_mapper         = new WC_Stripe_Agentic_Commerce_Order_Mapper();
-			$order                = $order_mapper->create_order_from_checkout_session( $checkout_session );
+			$order                = $order_mapper->create_order_from_checkout_session( $session );
 			$this->resolved_order = $order;
```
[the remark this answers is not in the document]
RadoslavGeorgiev: The intent ID is being validated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/4fafd53279b6ae87d63ea5ed7a25ccb26aad857a now.

includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
```diff
@@ -0,0 +1,253 @@
...
+	public function get_line_items(): array {
+		$raw_items = $this->session->line_items->data ?? [];
+
+		return array_map(
+			function ( $item ) {
+				return new WC_Stripe_Agentic_Line_Item( $item );
+			},
+			$raw_items
+		);
+	}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: There will rarely be more than 1 line item for most checkouts. This feels redundant, and I'd rather skip it.

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: While working on Transact Platform, I realized that DTOs are a very useful instrument, but should also be treated with care. I don't want to sneak them in with this PR. 😅 

I created STRIPE-957 to track this. If we're going to implement DTOs, I'd rather also have the shipping and tax calculations in place, to see how the whole thing would shape up first. That's because the current session is very tied to the specific objects that are expanded within it.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,457 @@
...
+			'line2'       => 'address_2',
+			'postal_code' => 'postcode',
+			'state'       => 'state',
+		];
+
+		foreach ( $map as $received_key => $order_key ) {
+			$method_name = sprintf( 'set_%s_%s', $type, $order_key );
+			$order->$method_name( $address->$received_key ?? '' );
+		}
+	}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: This is a set of methods that are foundational to WooCommerce, and would neither change there or here. With the current set of properties, I'd rather let them be.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,457 @@
...
+		if ( '' !== $customer_id ) {
+			$order_helper->update_stripe_customer_id( $order, $customer_id );
+		}
+
+		// Store Stripe currency.
+		$order_helper->update_stripe_currency( $order, $session->get_currency_lowercase() );
+
+		// Store checkout session ID for traceability.
+		$order->update_meta_data( '_stripe_checkout_session_id', $session->get_id() );
+	}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: Yeah, this flow is saving metadata much less frequently then others. I'll keep it as is.

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Commerce_Order_Mapper_Test.php
```diff
@@ -0,0 +1,1039 @@
...
+	 * @return void
+	 */
+	public function test_exception_thrown_when_currency_missing() {
+		$session = $this->build_checkout_session( [ 'currency' => null ] );
+
+		$this->expectException( Exception::class );
+		$this->expectExceptionMessage( 'missing the currency field' );
+
+		$this->mapper->create_order_from_checkout_session( $session );
+	}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/4fff05706491206cc08b62015b72f1dae13c7daa

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,457 @@
...
+			'Agentic order mapper: order created successfully.',
+			[
+				'session_id' => $session->get_id(),
+				'order_id'   => $order->get_id(),
+				'total'      => $order->get_total(),
+			]
+		);
+
+		return $order;
+	}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/4fff05706491206cc08b62015b72f1dae13c7daa

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
wjrosa: It looks nice! But I mean: after calling `update_meta_data`, AFAIK, you need to call `save_meta_data` to persist the information in the order.

includes/class-wc-stripe-webhook-handler.php
wjrosa: I agree. It makes sense to tackle this separately to avoid diverging from this PR's focus 👍.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: My understanding is that once you save the order, all related metadata is saved as well. That's why I'm calling `$order->save()` first, then relying on `complete_payment()`.

wjrosa: If that's true, it is great! I always had to call both so far

includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
```diff
@@ -0,0 +1,253 @@
...
+
+	/**
+	 * Returns the total amount in the smallest currency unit.
+	 *
+	 * @since 10.5.0
+	 * @return int
+	 */
+	public function get_amount_total(): int {
+		return (int) ( $this->session->amount_total ?? 0 );
```
daledupreez: As a pattern, I am wondering whether we should allow for `int|null` and `string|null` returns rather than coercing values to `''` or `0` -- particularly in this case, we probably want to handle a `null` amount quite differently from a real/legitimate `0` amount.

includes/class-wc-stripe-api.php
```diff
@@ -15,7 +15,7 @@ class WC_Stripe_API {
 	 */
 	const ENDPOINT                     = 'https://api.stripe.com/v1/';
 	const STRIPE_API_VERSION           = '2024-06-20';
-	const AGENTIC_COMMERCE_API_VERSION = '2025-10-29.clover';
+	const AGENTIC_COMMERCE_API_VERSION = '2025-12-15.preview';
```
daledupreez, line 18: @RadoslavGeorgiev, do we want to aim for a more recent preview version? It looks like there were version updates (both mainstream and preview) in January and February.

includes/class-wc-stripe-helper.php
```diff
@@ -277,6 +277,30 @@ public static function get_stripe_amount( $total, $currency = '' ) {
...
+	 * Converts a Stripe amount (in smallest currency unit) to a WooCommerce decimal amount.
+	 *
+	 * This is the inverse of `get_stripe_amount()`.
+	 *
+	 * @since 10.5.0
+	 * @param int    $amount   The amount in Stripe's smallest currency unit.
+	 * @param string $currency The three-letter currency code.
+	 * @return float The decimal amount for WooCommerce.
+	 */
+	public static function convert_from_stripe_amount( int $amount, string $currency ): float {
```
daledupreez, line 290: FYI, @Mayisha, as you're implementing very similar code in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5058.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1599,48 +1599,100 @@ public function process_checkout_session_completed( $notification ): void {
...
+			);
 			return;
 		}
+		set_transient( $lock_key, time(), 5 * MINUTE_IN_SECONDS );
 
-		$payment_intent_id = $this->get_payment_intent_id_from_checkout_session( $checkout_session );
+		try {
+			$this->handle_agentic_checkout_session( $notification );
+		} finally {
+			delete_transient( $lock_key );
```
daledupreez: We may want to use `WC_Stripe_Database_Cache::*()` operations instead of transients, as we have seen that transients don't behave that well in the wild. (That can be in a follow-up, though we just need to track it.)

includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
```diff
@@ -0,0 +1,253 @@
...
+	 * Checks whether this checkout session originates from agentic commerce.
+	 *
+	 * A session is agentic when at least one line item has an
+	 * external_reference that resolves to a nonzero integer (product ID).
+	 *
+	 * @since 10.5.0
+	 * @return bool
+	 */
+	public function is_agentic(): bool {
+		foreach ( $this->get_line_items() as $line_item ) {
```
daledupreez, line 256: I think we discussed this on the call yesterday, but there may be a more explicit field to identify this. Maybe not for this PR, but we should track it if we don't include in this PR.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1599,48 +1599,100 @@ public function process_checkout_session_completed( $notification ): void {
...
+		$override_version = function ( $headers ) {
+			$headers['Stripe-Version'] = WC_Stripe_API::AGENTIC_COMMERCE_API_VERSION;
+			return $headers;
+		};
+		add_filter( 'wc_stripe_request_headers', $override_version );
+
+		$url = $this->build_checkout_session_retrieve_url(
+			$notification->data->object->id,
+			WC_Stripe_Agentic_Checkout_Session::get_fields_to_expand()
+		);
```
daledupreez: I think it would be worth moving more of this code into the `try` block to ensure we handle errors and clean up.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1599,48 +1599,100 @@ public function process_checkout_session_completed( $notification ): void {
...
+			return;
+		}
+
+		$session = new WC_Stripe_Agentic_Checkout_Session( $raw_session );
+
+		if ( ! $session->is_agentic() ) {
+			return;
+		}
+
+		$payment_intent_id = $session->get_payment_intent_id();
```
daledupreez: I don't see where we are validating that the intent ID is not empty.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1666,58 +1718,35 @@ public function process_checkout_session_completed( $notification ): void {
...
+	private function build_checkout_session_retrieve_url( string $session_id, array $additional_expand = [] ): string {
+		$url    = 'checkout/sessions/' . $session_id;
+		$expand = array_merge( [ 'payment_intent.agent_details' ], $additional_expand );
 
-		if ( is_object( $payment_intent ) && isset( $payment_intent->id ) ) {
-			return $payment_intent->id;
+		$params = [];
+		foreach ( $expand as $field ) {
+			$params[] = 'expand[]=' . $field;
 		}
```
daledupreez: Should we not be URL-encoding `$session_id` and the `$field` values?
```suggestion
		$url    = 'checkout/sessions/' . raw_url_encode( $session_id );
		$expand = array_merge( [ 'payment_intent.agent_details' ], $additional_expand );

		$params = [];
		foreach ( $expand as $field ) {
			$params[] = 'expand[]=' . raw_url_encode( $field );
}
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,458 @@
  * Creates WooCommerce orders from Stripe agentic checkout session data.
  *
  * @since 10.5.0
- * @see STRIPE-902 for full implementation.
  */
 class WC_Stripe_Agentic_Commerce_Order_Mapper {
 
+	const ADDRESS_TYPE_BILLING  = 'billing';
+	const ADDRESS_TYPE_SHIPPING = 'shipping';
```
daledupreez: We should specify an explicit scope for these constants. I think we can stick with `protected` or `private` unless we foresee other uses.
```suggestion
	private const ADDRESS_TYPE_BILLING  = 'billing';
	private const ADDRESS_TYPE_SHIPPING = 'shipping';
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,458 @@
...
+				sprintf(
+					'wc_create_order() returned an unexpected type for session %s.',
+					$session->get_id()
+				)
+			);
+		}
+
+		$order->set_currency( $session->get_currency() );
+		$order->set_payment_method( 'stripe' );
+		$order->set_payment_method_title( __( 'Stripe', 'woocommerce-gateway-stripe' ) );
```
daledupreez: Not blocking: should we specify something like `Stripe Agentic Checkout` so this is clearly spelled out for merchants?
```suggestion
		$order->set_payment_method_title( __( 'Stripe Agentic Checkout', 'woocommerce-gateway-stripe' ) );
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,458 @@
...
+		$order->add_order_note(
+			__( 'Order created from Stripe agentic commerce checkout session.', 'woocommerce-gateway-stripe' )
+		);
```
daledupreez, line 156: Should we also add the payment intent to the order meta at this early stage?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,458 @@
...
+				);
+			}
+
+			$quantity   = $line_item->get_quantity();
+			$line_total = WC_Stripe_Helper::convert_from_stripe_amount(
+				$line_item->get_amount_total() - $line_item->get_amount_tax(),
+				$currency
+			);
+
+			$product = $this->resolve_product( $product_id, $line_item );
```
daledupreez: Non-blocking nit: it may be worth resolving the product immediately after the `0 === $product_id` check so the product validation stays together, especially because the other code will be unnecessary if we fail to look up the product.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,458 @@
...
+
+			// Let WooCommerce calculate totals from product price × quantity.
+			$item = $this->add_product_to_order( $order, $product, $quantity, $session->get_id() );
+
+			// Verify WC-calculated total matches Stripe's pre-tax line total.
+			$wc_line_total = (float) $item->get_total();
+			if ( abs( $wc_line_total - $line_total ) > 0.01 ) {
+				throw new Exception(
+					sprintf(
+						'Line item price mismatch for product %d: WC calculated %.2f, Stripe expected %.2f.',
```
daledupreez: Two concerns here:
* Is `0.01` small enough for three-decimal currencies?
* For three-decimal currencies, I think the `%.2f` won't accurately reflect the amount.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,458 @@
...
+
+		$this->map_address(
+			$order,
+			$billing_address,
+			$session->get_customer_name(),
+			$session->get_billing_phone(),
+			self::ADDRESS_TYPE_BILLING
+		);
+
+		// Shipping name, phone, and address (optional — not collected for digital goods).
```
daledupreez, line 385: Might it be worth checking whether the goods in the order need shipping before we ignore accept the case where there is no shipping address? Or at least add a note to the order when that occurs?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -16,26 +16,458 @@
...
+		$expected_total = WC_Stripe_Helper::convert_from_stripe_amount(
+			$session->get_amount_total(),
+			$session->get_currency()
 		);
+		$order_total    = (float) $order->get_total();
+
+		if ( abs( $order_total - $expected_total ) > 0.01 ) {
+			throw new Exception(
+				sprintf(
+					'Order total mismatch for session %s: WC total %.2f, Stripe total %.2f.',
```
daledupreez: As above for the three-decimal cases - is `0.01` the right value, and will we lose formatting for three-decimal currencies.

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Checkout_Session_Test.php
```diff
@@ -0,0 +1,498 @@
...
+		$this->assertSame( 'eur', $session->get_currency_lowercase() );
+	}
+
+	/**
+	 * Test get_amount_total.
+	 */
+	public function test_get_amount_total() {
+		$session = new WC_Stripe_Agentic_Checkout_Session( (object) [ 'amount_total' => 2500 ] );
+		$this->assertSame( 2500, $session->get_amount_total() );
+	}
```
daledupreez: Not blocking: should we add a test for the cast to `int`? e.g. `[ 'amount_total' => '2500.3' ]`.

Also not blocking, but these types of tests may be worth pushing into data providers so it's easier to skim the test cases and check that we have good coverage.

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Checkout_Session_Test.php
```diff
@@ -0,0 +1,498 @@
...
+					'shipping_details'      => null,
+					'collected_information' => (object) [
+						'shipping_details' => (object) [ 'name' => 'Collected Name' ],
+					],
+				],
+				'Collected Name',
+			],
+			'empty when all missing'            => [
+				(object) [
+					'customer_details' => (object) [ 'name' => null ],
```
daledupreez, line 167: Is it worth adding a test case for the data object being `(object) []` with no properties?

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Checkout_Session_Test.php
```diff
@@ -0,0 +1,498 @@
...
+	}
+
+	/**
+	 * Test get_customer_id empty when object (expanded customer).
+	 */
+	public function test_get_customer_id_empty_when_object() {
+		$session = new WC_Stripe_Agentic_Checkout_Session(
+			(object) [ 'customer' => (object) [ 'id' => 'cus_expanded' ] ]
+		);
+		$this->assertSame( '', $session->get_customer_id() );
```
daledupreez: At a glance, this looks like it should return `cus_expanded`. Why are we falling back on `''` for this case?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
daledupreez: Also, is there any way for us to verify that the external ID belongs to us and not some other system? Might it make sense to allow each site to specify a prefix?

I ask because we definitely have situations where multiple sites are connected to the same Stripe account. Collisions on product IDs are guaranteed -- we fixed a bug with webhooks that stems from order IDs from another site colliding with refunds on the current site.

tests/phpunit/WC_Stripe_Account_Test.php
```diff
@@ -533,6 +533,56 @@ public function test_reconfigure_webhooks_on_update_with_outdated_api_version()
...
+			'id'             => 'we_123',
+			'url'            => WC_Stripe_Helper::get_webhook_url(),
+			'enabled_events' => WC_Stripe_Account::WEBHOOK_EVENTS,
+			'api_version'    => \WC_Stripe_API::AGENTIC_COMMERCE_API_VERSION,
+			'status'         => 'enabled',
+		];
+
+		$this->account = $this->getMockBuilder( WC_Stripe_Account::class )
+			->setConstructorArgs( [ $this->mock_connect, WC_Helper_Stripe_Api::class ] )
+			->setMethods( [ 'get_existing_webhook', 'configure_webhooks' ] )
```
daledupreez: Nit: `setMethods()` is deprecated in favour of `onlyMethods()`.
```suggestion
			->onlyMethods( [ 'get_existing_webhook', 'configure_webhooks' ] )
```

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
```diff
@@ -213,32 +144,109 @@ function () use ( &$failure_action_fired ) {
...
+
+		$this->assertTrue( $success_action_fired );
+		$this->assertInstanceOf( WC_Order::class, $created_order );
+		$this->assertEquals( 'processing', $created_order->get_status() );
+		$this->assertEquals( '20.00', $created_order->get_total() );
+		$this->assertEquals( 'stripe', $created_order->get_payment_method() );
+		$this->assertEquals( 'pi_test_cs_test_happy', $created_order->get_meta( '_stripe_intent_id', true ) );
+
+		$created_order->delete( true );
+		$product->delete( true );
```
daledupreez: Do we want this cleanup to happen in a finally or similar? At present, this code won't run if an assertion fails.

tests/phpunit/WC_Stripe_Webhook_Handler_Agentic_Test.php
```diff
@@ -213,32 +144,109 @@ function () use ( &$failure_action_fired ) {
...
 
 	/**
-	 * Builds a checkout.session.completed notification object.
+	 * Intercepts HTTP requests to the Stripe checkout sessions API and returns a mock response.
+	 *
+	 * @param object $response_body The mock response body object.
+	 */
+	private function mock_stripe_api_response( $response_body ) {
+		$this->http_filter = function ( $preempt, $args, $url ) use ( $response_body ) {
+			if ( false !== strpos( $url, 'api.stripe.com' ) && false !== strpos( $url, 'checkout/sessions/' ) ) {
```
daledupreez: Two non-blocking nits here:
1. The method name suggests the code will mock a generic response, but it will only mock the checkout sessions API response. It may be worth updating the name so that is clearer. e.g. `mock_stripe_checkout_sessions_response( $response_body )`.
2. It feels like it would be simpler to do a simpler comparison on the URL -- I am not sure why we're using two separate `strpos()` conditions.
```suggestion
	private function mock_stripe_checkout_sessions_response( $response_body ) {
		$this->http_filter = function ( $preempt, $args, $url ) use ( $response_body ) {
			if ( str_starts_with( $url, 'https://api.stripe.com/v1/checkout/sessions/' ) ) {
```

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Commerce_Order_Mapper_Test.php
```diff
@@ -0,0 +1,1075 @@
...
+			],
+			'total_details'    => (object) [
+				'amount_shipping' => 0,
+				'amount_tax'      => 0,
+				'amount_discount' => 0,
+			],
+			'line_items'       => $this->build_line_items(
+				[
+					[
+						'lookup_key'      => (string) $this->default_product->get_id(),
```
daledupreez: Non-blocking nit: I think it's better to have the desired product passed in as an argument to this function so it is explicit in every caller which product is being passed in. As things are written, the reference to the product is 100% implicit and makes it trickier to understand how we "know" the order should have specific properties.

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Commerce_Order_Mapper_Test.php
```diff
@@ -0,0 +1,1075 @@
...
+
+		$order->delete( true );
+	}
+
+	/**
+	 * Test that the order status is set to processing.
+	 *
+	 * @return void
+	 */
+	public function test_order_status_is_processing() {
```
daledupreez, line 236: Should we also add a test case for a virtual product?

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Commerce_Order_Mapper_Test.php
```diff
@@ -0,0 +1,1075 @@
...
+				],
+			]
+		);
+
+		$order = $this->mapper->create_order_from_checkout_session( $session );
+
+		$this->assertEquals( $user_id, $order->get_customer_id() );
+
+		$order->delete( true );
+		wp_delete_user( $user_id );
```
daledupreez, line 274: Nit: this cleanup won't be run if the assertion fails. We could get the value, clean up, and then assert.

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Commerce_Order_Mapper_Test.php
```diff
@@ -0,0 +1,1075 @@
...
+		$this->assertCount( 1, $items );
+
+		$item = reset( $items );
+		$this->assertInstanceOf( WC_Order_Item_Product::class, $item );
+		$this->assertEquals( $product->get_id(), $item->get_product_id() );
+		$this->assertEquals( 1, $item->get_quantity() );
+		$this->assertEquals( '15.00', wc_format_decimal( $item->get_total(), 2 ) );
+
+		$order->delete( true );
+		$product->delete( true );
```
daledupreez, line 358: As above, this cleanup won't run if an assertion fails.

daledupreez (approved):
Changes look good to me overall, though I have plenty of small nits and feedback.

includes/class-wc-stripe-helper.php
Mayisha: Right. I have added a similar method `get_woocommerce_amount_from_stripe_amount`. Only the return statement is different `return wc_format_decimal( $amount, $decimals )`.

includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
RadoslavGeorgiev: I guess this was more of a temporary shortcut for the implementation of this PR. Once we get our hands on the `agentic_details` field of the intent, and can test it, I'll update https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5027 (the base of this PR).

includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
RadoslavGeorgiev: I made the returns nullable in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/7a8999179ae90dd827f7e687c876e4630c12629b and moved the conversion back to a string (where applicable) to the mapper.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: Any reason to go earlier than the current call in `store_stripe_metadata()`? [ref](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/83b26d2855b0be7bec7624dd0f376311543ca7fe/includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php#L411)

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/5cb4d45d4e28ce1109f15c096ad2c4aff5e9635b

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: Changed to `Stripe (Agentic Checkout)` in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/83b26d2855b0be7bec7624dd0f376311543ca7fe.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: > Non-blocking nit: it may be worth resolving the product immediately after the `0 === $product_id` check so the product validation stays together, especially because the other code will be unnecessary if we fail to look up the product.

Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/588b490ce87db9305ec6bed1a0fd5c20b36b7e1f

> we fixed a bug with webhooks that stems from order IDs from another site colliding with refunds on the current site

Can you link to the PR where you fixed this? I'm not sure what structure would make the most sense, and I'm curious what's already working.

daledupreez: The fix was in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5025, but the approach was defensive to prevent fatals rather than clearly identifying whether the order is for the current site.

includes/class-wc-stripe-api.php
RadoslavGeorgiev: I wouldn't mind going to a newer version. The point is that the same version will be used throughout the whole plugin when agentic commerce is enabled, and I do not know whether anything might break because of it. That's why I was trying to be as cautious as possible, and update to the first version where agentic stuff actually works.

What is the API version update process typically like?

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: Alright, done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/f5bf7de78175148ac16d58b118d7f00487252db8

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/de0ab0d02937092147ea6022a6ca16c6eb9c53fe. Though I'm not sure how helpful that is, given that none of those methods are supposed to throw any errors.

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: Good point, it's only validated much later, with the newly loaded checkout session. Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/9593d5be892619a47aa57f067b9cb4f97c74b301

includes/class-wc-stripe-webhook-handler.php
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/9519b0ff334ed528b00b3c2438bd3c88bce27eed

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: For the comparison, I'm using `0.01` to account for 3-decimal currencies.

As for the log, it's `wc_format_decimal()` now.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/53c5775ab7f49afa871f06133d46ead59480a9dd. I'm now checking whether any of the products require shipping, and if so, adding a note to the order.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: Also addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/782a804eccf63a8fe6e75b49bba1b20d88e7705d

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Checkout_Session_Test.php
RadoslavGeorgiev: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5032/commits/6b68415d928a6ea75cf3f88caa94fe073711a242

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: I created STRIPE-968 to track this, though I don't have any specific ideas.

Maybe we can have something like `<id>-<hash>` where `id` is the product ID and `hash` is a random value we generate once per product? 🤔

includes/class-wc-stripe-api.php
daledupreez: We just went through that in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4977, so `STRIPE_API_VERSION` is now at `2025-09-30.clover`, so behind your preview version, but a lot closer than before, as we skipped `acacia` and `basil` to get to `clover`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
daledupreez: I think it depends on how we handle errors or failures. If we're deleting the order, then not a problem. But if we're keeping the order, then knowing what the payment intent was could be helpful. But definitely a question, not a suggestion!

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Commerce_Order_Mapper_Test.php
```diff
@@ -832,6 +836,60 @@ public function test_order_created_without_shipping_details() {
...
+
+		$order = $this->mapper->create_order_from_checkout_session( $session );
+
+		$notes    = wc_get_order_notes( [ 'order_id' => $order->get_id() ] );
+		$contents = array_map( fn( $note ) => $note->content, $notes );
+		$this->assertNotEmpty(
+			array_filter( $contents, fn( $c ) => str_contains( $c, 'no shipping address was provided' ) )
+		);
+
+		$order->delete( true );
```
daledupreez: Nit: We won't delete the order if the assertion fails.

tests/phpunit/AgenticCommerce/WC_Stripe_Agentic_Commerce_Order_Mapper_Test.php
```diff
@@ -832,6 +836,60 @@ public function test_order_created_without_shipping_details() {
...
+		$order = $this->mapper->create_order_from_checkout_session( $session );
+
+		$notes    = wc_get_order_notes( [ 'order_id' => $order->get_id() ] );
+		$contents = array_map( fn( $note ) => $note->content, $notes );
+		$this->assertEmpty(
+			array_filter( $contents, fn( $c ) => str_contains( $c, 'no shipping address was provided' ) )
+		);
+
+		$order->delete( true );
+		$virtual_product->delete( true );
```
daledupreez, line 888: Nit: as above, we won't delete the order or virtual product if an assertion fails.

daledupreez (approved):
Still LGTM! Thanks for all the detailed fixes and improvements, @RadoslavGeorgiev!

includes/class-wc-stripe-api.php
RadoslavGeorgiev: Once the test tool comes back to life, I'll test to see if `2025-09-30.clover` does the job, but I think it won't. We might want to test something newer (non-preview).

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
RadoslavGeorgiev: Yeah, we are deleting the order on failure Execution will either reach the point to store the intent ID, or delete the order. FWIW, it should be a rare case, given that by the time we receive the webhook, an order already exists, just not on the current site.

includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
```diff
@@ -0,0 +1,264 @@
...
+
+	/**
+	 * Returns the expanded payment intent ID, or null when absent.
+	 *
+	 * @since 10.6.0
+	 * @return string|null
+	 */
+	public function get_payment_intent_id(): ?string {
+		return isset( $this->session->payment_intent->id ) ? (string) $this->session->payment_intent->id : null;
+	}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: This session is only intended for use where and how it is used. `build_checkout_session_retrieve_url` in the webhook handler already adds `payment_intent.agent_details`, forcing the intent to be expanded.

includes/agentic-commerce/class-wc-stripe-agentic-line-item.php
```diff
@@ -0,0 +1,119 @@
...
+	 * @since 10.6.0
+	 * @return int
+	 */
+	public function get_product_id(): int {
+		if ( ! isset( $this->item->price ) || ! is_object( $this->item->price ) ) {
+			return 0;
+		}
+
+		return intval( $this->item->price->external_reference ?? '' );
+	}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: We control where those IDs come from in the product feed, and they will never have non-digits in there.

includes/class-wc-stripe-helper.php
```diff
@@ -277,6 +277,30 @@ public static function get_stripe_amount( $total, $currency = '' ) {
...
+		if ( in_array( $currency, self::no_decimal_currencies(), true ) ) {
+			return (float) absint( $amount );
+		}
+
+		if ( in_array( $currency, self::three_decimal_currencies(), true ) ) {
+			return round( $amount / 1000, 3 );
+		}
+
+		return round( $amount / 100, 2 );
+	}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: This method is meant to handle only positive amounts. If the amount were to be negative, Stripe would return a different object type.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1759,4 +1781,137 @@ private function get_payment_method_type_from_charge( $charge ) {
...
+	 * @since 10.6.0
+	 * @param object $notification The webhook notification from Stripe.
+	 */
+	private function handle_agentic_checkout_session( $notification ): void {
+		$checkout_session = $notification->data->object;
+
+		if ( ! WC_Stripe_Feature_Flags::is_agentic_commerce_enabled() ) {
+			WC_Stripe_Logger::error( 'Agentic commerce is disabled, skipping agentic checkout session: ' . $checkout_session->id );
+			return;
+		}
```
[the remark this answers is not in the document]
RadoslavGeorgiev: It is not expected to receive an agentic hook with agentic commerce disabled at all. That's a very serious case, not just a by the way situation.

PR #5044, opened 2026-02-23
Minor order helper class updates

wjrosa:

Refactors `WC_Stripe_Order_Helper` to reduce duplication and replace inline meta key strings with named constants.

- **New constants** — `META_STRIPE_LOCK_PAYMENT` and `META_STRIPE_LOCK_REFUND` replace hardcoded `'_stripe_lock_payment'` / `'_stripe_lock_refund'` strings in the lock/unlock/get methods.
- **Consolidated null-guard** — Multiple getter methods that each duplicated an `if ( null === $order ) { return false; }` guard before calling `get_meta()` now delegate to the existing `get_order_meta()` helper instead.

**Before:**
```php
public function get_stripe_intent_id( ?WC_Order $order = null ) {
    if ( is_null( $order ) ) {
        return false;
    }
    return $order->get_meta( self::META_STRIPE_INTENT_ID, true );
}
```

**After:**
```php
public function get_stripe_intent_id( ?WC_Order $order = null ) {
    return $this->get_order_meta( $order, self::META_STRIPE_INTENT_ID );
}
```

## Testing instructions

Code review should be enough.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
@copilot, update this PR description, update the branch with the base branch, and add entries to the changelog.txt and readme.txt and make this ready for review

Mayisha (approved):
Looks good to me 👍

includes/class-wc-stripe-order-helper.php
malithsen: Shouldn't we also replace this instance?

Otherwise, removing the null guard could lead to issues here.

malithsen (commented):
Found another call to `order->get_meta` that needs adjustment. Otherwise looks good!

includes/class-wc-stripe-order-helper.php
malithsen: Ah I see coderabbit already noted this [above](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5044#pullrequestreview-3853929182), along with another instance of a similiar issue

wjrosa: You're right. Replaced those in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5044/commits/e059f5639ca84ccc9b0b155ee8d745a75e4d6a31

malithsen (approved):
Looks good. Thanks for the refactor!

PR #5047, opened 2026-02-23
Process deferred webhook for on hold and cancelled orders

Mayisha:
Fixes STRIPE-948

The deferred webhook processing only runs `handle_deferred_payment_intent_succeeded()` (which would call `payment_complete()` and store transaction data in order meta) when the order has status pending or failed. If the order has any other status, it skips and never marks the order as paid in this path.

There could be edge cases where this check end up ignoring the event and loses transaction data. For example, when a user takes some time instead of immediately approving the 3DS authentication, it may cause a nonce error on returning to the site. If the user then cancels the order the payment intent succeeded event is never processed for this order due to the above status check, while in the merchant's Stripe account, the payment is already received.

This PR handles this edge case.
- Now we process deferred webhook for on-hold and cancelled orders.
- After processing response, we move the order status to on-hold if it was cancelled before and add a order note for the merchant.

## Testing instructions
- To reproduce the edge case, throw hard coded nonce error from [update_order_status_ajax](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-intent-controller.php#L688) methods (`throw new WC_Stripe_Exception( 'missing-nonce', __( 'Verification failed.', 'woocommerce-gateway-stripe' ) );`)
- As a logged in shopper, add a product to your cart and go to the checkout page.
- Select card payment and use a [3DS card](https://docs.stripe.com/testing#regulatory-cards) to pay.
- Click the place order button.
- Approve the payment in the 3DS modal.
- The checkout should fail.
- Go to **My Account > Orders** and cancel the latest order (this should be in pending payment state at this point).
- Go to the order edit page as an admin. The order should be in the cancelled state.
- Wait for a couple of minutes and refresh the order edit page.
- Go to **WooCommerce > Status > Scheduled Actions > Pending** and confirm that there is no deferred intent webhook action in the pending state for this order. If there is one, then run the action.
- Check the status in the order edit page.
- In `develop` branch, the order should still be in the cancelled state and there should be no transaction ID or related order notes. Check the Stripe dashboard of the connected account and notice that the transaction was successful for this order.
- In this branch, the order should be in 'on-hold' state and there should be relevant order notes.

Screenshot 2026-02-23 at 11 00 38 PM

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,6 +22,7 @@
 * Dev - Add CodeRabbit configuration with Stripe-focused review guidance
 * Dev - Expand AI agent guidance with directory-level AGENTS and CLAUDE context files
 * Fix - Prevent fatal error when order ID in webhook references a refund
+* Fix - Deferred payment_intent.succeeded webhook now processes cancelled or on-hold orders so they are updated with Stripe transaction data when payment succeeded in Stripe
 
 = 10.4.0 - 2026-02-09 =
 * Add - Introduce an endpoint to create Checkout Sessions tokens
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1316,7 +1316,14 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 					}
 
 					// Check if the order is still in a valid state to process the webhook.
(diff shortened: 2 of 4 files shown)


Mayisha:
closing in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5703

Issue #4995, opened 2026-02-05
Checkout Sessions: Payment flow (block checkout)


wjrosa:
A Shopper must be able to complete a purchase successfully using the new integration in the modern Block Checkout (Requires an AJAX endpoint for session creation and frontend work to mount the currency selector and payment element).

PR #5049, opened 2026-02-23
Processing checkout session payments in the block checkout

wjrosa:
Towards STRIPE-945
Towards #4995

Needs to be merged after the following branch:

This PR handles the main payment processing logic for checkout sessions in the block checkout. This shares some backend logic from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5019. I have created new hooks to handle WC events, as besides sharing some logic with the payment intent processing, the code here is much leaner.

## Testing instructions

- Checkout and build this branch on your test environment (`add/processing-checkout-session-payments-in-block-checkout`)
- Merge https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5058
```php
```
- Connect a US Stripe account
- Reconfigure the webhooks. Your store must be publicly accessible
- As a logged-in shopper, add a product to your cart and go to the block checkout page (*shopper must be logged in for now*)
- Select your local currency and pay with card
- As a merchant, verify that the order is successful and the transaction details are present

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
@copilot fix existing frontend unit tests, and add new ones to the files included in this PR

client/blocks/checkout-sessions/hooks.js
```diff
@@ -1,17 +1,115 @@
...
+						return {
+							type: 'error',
+							message: __(
+								'Your payment information is incomplete.',
+								'woocommerce-gateway-stripe'
+							),
+						};
+					}
+
+					if ( errorMessage ) {
```
malithsen, line 60: Would it be worth moving this `errorMessage` block above the `isPaymentElementComplete` block as it seems like `errorMessage` would be more specific than the generic error returned by  `isPaymentElementComplete`

wjrosa: Alright. Moved it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/e61d69e624dc140a96460f55b1f54e1cb5ddc811

malithsen (approved):
This looks good me now. Thanks for the change

client/blocks/checkout-sessions/__tests__/hooks.test.js
```diff
@@ -1,19 +1,153 @@
...
+				hasLoadErrorRef,
+				true
+			);
+			const result = await onPaymentSetupResultPromise;
+			expect( result ).toEqual( {
+				type: 'success',
+				meta: {
+					paymentMethodData: {
+						payment_method: 'stripe',
+						'wc-stripe-is-deferred-intent': true,
```
daledupreez: Do we need this argument? I thought we merged the changes that remove it.

client/blocks/checkout-sessions/hooks.js
```diff
@@ -1,17 +1,115 @@
...
+) => {
+	useEffect(
+		() =>
+			onPaymentSetup( () => {
+				async function handlePaymentProcessing() {
+					if ( hasLoadErrorRef.current ) {
+						return {
+							type: 'error',
+							message: __(
+								'Invalid or missing payment details. Please ensure the provided payment method is correctly entered.',
```
daledupreez: I am not sure what we are asking the shopper to do in this case. If this was a problem loading the checkout session, I can't see what they should or could do regarding the payment method.

client/blocks/checkout-sessions/hooks.js
```diff
@@ -1,17 +1,115 @@
...
+					const { validationStore } = window.wc?.wcBlocksData ?? {};
+					if ( validationStore ) {
+						const store = select( validationStore );
+						const hasValidationErrors = store.hasValidationErrors();
+
+						// Return if there is a validation error on the checkout fields.
+						if ( hasValidationErrors ) {
+							return;
+						}
+					}
```
daledupreez, line 38: Should we check the validation status first so we don't overwrite those errors with something from our code?

client/blocks/checkout-sessions/hooks.js
```diff
@@ -1,17 +1,115 @@
...
+							),
+						};
+					}
+
+					return {
+						type: 'success',
+						meta: {
+							paymentMethodData: {
+								payment_method: 'stripe',
+								'wc-stripe-is-deferred-intent': true,
```
daledupreez: Nit: I thought we removed the related functionality for this flag.
```suggestion
```

client/blocks/checkout-sessions/hooks.js
```diff
@@ -1,17 +1,115 @@
...
+								// The billing information here is relevant to properly create the Stripe Customer object.
+								billing_email: billingAddress.email,
+								billing_first_name: billingAddress.first_name,
+								billing_last_name: billingAddress.last_name,
+								billing_address_1: billingAddress.address_1,
+								billing_address_2: billingAddress.address_2,
+								billing_city: billingAddress.city,
+								billing_state: billingAddress.state,
+								billing_postcode: billingAddress.postcode,
+								billing_country: billingAddress.country,
```
daledupreez: Is there a reason we can't get this information from the standard Woo payload or metadata? It feels like this is somewhat redundant.

daledupreez (commented):
I think this is looking good overall and testing well, but I have some questions I would like to address before I approve.

I also tripped over what I think is a timing issue with the order received page, which happens when the page is rendered before the webhook is received -- note the "Actions" section:

Image

I think that's something we will be addressing separately, but it would be good to confirm.

client/blocks/checkout-sessions/__tests__/hooks.test.js
wjrosa: Not really. Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/ce1e75dddc6593c3f4e2f6b98618aa6fd22b6d65

client/blocks/checkout-sessions/hooks.js
wjrosa: You're right. I replaced the error message in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/621afd8b352f7b8b192068b1f3c331190676f818

client/blocks/checkout-sessions/hooks.js
wjrosa: Can you give some examples of what you mean by "standard Woo payload or metadata"? This code is inherited from our existing flow for payment intents. So we may have an opportunity to improve it there as well

client/blocks/checkout-sessions/hooks.js
wjrosa: You mean moving this block above `if ( hasLoadErrorRef.current ) {`?

client/blocks/checkout-sessions/hooks.js
wjrosa: I did some manual testing here without this information, and everything works just fine! Thanks for flagging it. These fields are necessary for the payment intent flow because we try to create the customer only when processing the payment, while with checkout sessions, we do it when creating the session.

Removed the fields in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/7703afdf2a0247cd67cc40564c5756972f17d81a

wjrosa: To keep the consistency, I added the billing information to the checkout session creation flow in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/7de61aa57e1b4e7584ed008a7665ae9c3796e6a9

client/blocks/checkout-sessions/hooks.js
wjrosa: I went ahead and moved it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/ca894d7182fdb638b42780e0bae3e25e7fbe27ab

client/blocks/checkout-sessions/hooks.js
```diff
@@ -1,17 +1,115 @@
...
+							},
```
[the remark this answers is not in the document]
wjrosa: This is outdated

client/blocks/checkout-sessions/hooks.js
```diff
@@ -1,17 +1,114 @@
...
+
+					return {
+						type: 'success',
+						meta: {
+							paymentMethodData: {
+								payment_method: 'stripe',
+								save_payment_method: 'no', // TODO: Correctly handle this when supporting saved payment methods in the future.
+								wc_stripe_checkout_session_id:
+									checkoutSessionId,
+
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/fcced0b893df7d697f82370e0ef6a8d4cf38c05d

client/blocks/checkout-sessions/__tests__/hooks.test.js
```diff
@@ -1,19 +1,152 @@
...
+						billing_address_1: billingAddress.address_1,
+						billing_address_2: billingAddress.address_2,
+						billing_city: billingAddress.city,
+						billing_state: billingAddress.state,
+						billing_postcode: billingAddress.postcode,
+						billing_country: billingAddress.country,
+					},
+				},
+			} );
+		} );
```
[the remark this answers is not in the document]
wjrosa: This is outdated

client/blocks/checkout-sessions/__tests__/hooks.test.js
```diff
@@ -1,19 +1,153 @@
 import {
-	usePaymentCompleteHandler,
+	usePaymentSetupHandler,
+	useCheckoutSuccessHandler,
 	usePaymentFailHandler,
 } from 'wcstripe/blocks/checkout-sessions/hooks';
 import { useEffect } from '@wordpress/element';
+import { select } from '@wordpress/data';
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/c70b18d6a48f2cd7250c35ccc73be6fa5ec55226

wjrosa:
@daledupreez thanks for the review! As for the action buttons issue, @Mayisha is addressing it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5101

client/api/index.js
```diff
@@ -678,11 +678,27 @@ export default class WCStripeAPI {
 	/**
 	 * Creates a new checkout session.
 	 *
+	 * @param {Object} billingAddress Billing address.
```
daledupreez: Nit: Based on the code, `billingAddress` could be null or at least falsey. I think we should reflect that in the JSDoc.
```suggestion
	 * @param {Object|null} billingAddress Billing address.
```

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -30,7 +32,7 @@ export const CheckoutContainer = ( props ) => {
 			);
 		}
 		return clientSecret;
-	}, [ api, setShouldLoadStripeElements ] );
+	}, [ api, billingAddress, setShouldLoadStripeElements ] );
```
daledupreez: I have two concerns here:
 - When we first load the page, the billing address is likely to be either empty or it may be stale if billing details are for an existing customer. So using the billing address at that point in time feels like it will be wrong at least some of the time.
 - The `useMemo()` now has `billingAddress` as a dependency, which means that address updates will change the return value, and that may cascade into additional updates in the subsequent `useMemo()` and the `CheckoutProvider` render.

Both of those issues feel problematic, especially because I think we should only be sending the billing details across when we attempt to make a payment, and not before.

wjrosa: Alright. Removed the billing information in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5049/commits/a5564147d99109aef1e6e640b0b461171d3ab2ff

daledupreez (approved):
I've retested with the removal of the billing address and all is looking good!

PR #5050, opened 2026-02-24
Add order lock in update order status AJAX

### LINKED ISSUE #5042: `update_order_status_ajax()` missing `lock_order_payment()` causes double `payment_complete()` on credit card 3DS payments
### Prerequisites

- [X] I have carried out troubleshooting steps and I believe I have found a bug.
- [X] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

When a customer pays with a credit card that requires 3D Secure (SCA), which is the majority of European cards, two concurrent processes can both call `$order->payment_complete()` within 1 second of each other, causing:

* **Stock reduced twice** (the `_stock_reduced` flag is read by both processes before either writes it)
* **Duplicate order notes** (two "Stock levels reduced" notes, often in different locales)
* **Duplicate customer emails** (two "Order confirmed" emails)

The root cause is that `update_order_status_ajax()` in `class-wc-stripe-intent-controller.php` (line 672) does **not** call `lock_order_payment()` before invoking `process_order_for_confirmed_intent()`, making it race against `verify_intent_after_checkout()` which **does** acquire a lock.

This is the same class of bug that was fixed for redirect-based payment methods in **PR** woocommerce/woocommerce-gateway-stripe#4437 (v9.6.0, July 2025), but the AJAX path for inline credit card payments was not covered by that fix.

### Environment

| Software | Version |
| -- | -- |
| Stripe Gateway | 10.4.0 |
| WooCommerce | 10.5.1 |
| WordPress | 6.9.1 |
| PHP | 8.2 |
| Server | SiteGround (managed WordPress hosting) |
| Object Cache | **Memcached** (SiteGround SuperCacher) |

### Steps to reproduce

1. Enable an **external persistent object cache** (Memcached, Redis, LiteSpeed Cache, etc.)
2. Configure a WooCommerce store in a European locale (SCA/3DS is enforced)
3. Place an order with a credit card that triggers 3D Secure authentication
4. Complete the 3DS challenge in the modal
5. Observe the order in wp-admin, check order notes and stock levels

The issue is intermittent and timing-dependent (\~1 in 5 orders in our case). It does **not** happen on every order because it requires the AJAX response and the thank-you page redirect to overlap within a tight window.

### Expected behavior

* `payment_complete()` is called once
* Stock is reduced once
* One "processing" order note
* One customer email

### Actual behavior

* `payment_complete()` is called **twice** within \~1 second
* Stock is reduced **twice** (e.g., a product with stock 10 goes to 8 instead of 9)
* **Two** "Order status changed from Pending payment to Processing" notes
* **Two** "Stock levels reduced" notes, the first in the customer's locale (e.g., English/Italian), the second in the site's default locale (Portuguese in our case), proving they come from two separate PHP processes
* **Two** customer confirmation emails

#### The two racing processes

For credit cards with 3DS, the payment flow after the customer completes the 3DS challenge is:

**Path A - AJAX (**`update_order_status_ajax`):

```
JS confirms 3DS → fires wc_ajax: wc_stripe_update_order_status
  → update_order_status_ajax()                          [NO lock_order_payment()]
    → process_order_for_confirmed_intent()
      → process_response()
        → $order->payment_complete()                    ← FIRST payment_complete
```

**Path B - Thank-you page redirect (**`verify_intent_after_checkout`):

```
JS window.location = return_url (fires concurrently with AJAX)
  → verify_intent_after_checkout()
    → lock_order_payment()                              [LOCK acquired — but no competing lock exists]
    → handle_intent_verification_success()
      → process_response()
        → $order->payment_complete()                    ← SECOND payment_complete
```

#### Why the lock doesn't help

`verify_intent_after_checkout()` (line 3943) correctly calls `lock_order_payment()`, but this is useless because `update_order_status_ajax()` **never sets a lock**. Path B's `lock_order_payment()` always sees "no existing lock" and proceeds.

Even the order status guard in `verify_intent_after_checkout()` (line 3937: `$order->has_status(['pending', 'failed'])`) fails under Memcached/Redis because it reads a **stale cached order object** that still shows "pending" even though Path A already changed it to "processing."

#### Reference: the unprotected code path

`class-wc-stripe-intent-controller.php`, line 703:

```php
$gateway->process_order_for_confirmed_intent( $order, $intent_id_received, $save_payment_method );
```

No `lock_order_payment()` call anywhere in `update_order_status_ajax()`. Compare this to every other code path:

| Code Path | File | Lock? |
| -- | -- | -- |
| `process_payment_with_payment_method()` | `class-wc-stripe-upe-payment-gateway.php:1174` | ✅ |
| `process_upe_redirect_payment()` | `class-wc-stripe-upe-payment-gateway.php:1928` | ✅ (added in PR woocommerce/woocommerce-gateway-stripe#4437) |
| `verify_intent_after_checkout()` | `class-wc-stripe-upe-payment-gateway.php:3943` | ✅ |
| `process_payment_intent()` (webhook) | `class-wc-stripe-webhook-handler.php:1096` | ✅ |
| `update_order_status_ajax()` | `class-wc-stripe-intent-controller.php:672` | **❌ MISSING** |

### Proposed fix

Add `lock_order_payment()` to `update_order_status_ajax()`, mirroring the pattern used in PR woocommerce/woocommerce-gateway-stripe#4437 for `process_upe_redirect_payment()`:

```php
public function update_order_status_ajax() {
    $order        = false;
    $order_id     = isset( $_POST['order_id'] ) ? absint( $_POST['order_id'] ) : false;

    try {
        // ... existing nonce and order validation ...

        $save_payment_method = isset( $_POST['payment_method_id'] ) && ! empty( wc_clean( wp_unslash( $_POST['payment_method_id'] ) ) );

+       // Prevent concurrent payment processing (e.g., webhook or redirect handler).
+       if ( $order_helper->lock_order_payment( $order ) ) {
+           WC_Stripe_Logger::info( "Skipping update_order_status_ajax for order $order_id, order payment is already locked." );
+           wp_send_json_success(
+               [ 'return_url' => $this->get_upe_gateway()->get_return_url( $order ) ],
+               200
+           );
+           return;
+       }

        $gateway = $this->get_upe_gateway();
        $gateway->process_order_for_confirmed_intent( $order, $intent_id_received, $save_payment_method );
+
+       $order_helper->unlock_order_payment( $order );

        wp_send_json_success( ... );
+       if ( $order ) {
+           $order_helper->unlock_order_payment( $order );
+       }
        // ... existing error handling ...
    }
}
```

### Deeper issue: `lock_order_payment()` is not atomic under object caches

Even with the above fix, the current locking mechanism (`update_meta_data` + `save_meta_data`) is a classic TOCTOU (time-of-check-time-of-use) race under persistent object caches:

```php
// lock_order_payment() in class-wc-stripe-order-helper.php:1091
public function lock_order_payment( WC_Order $order ): bool {
    if ( $this->is_order_payment_locked( $order ) ) {  // READ from cache
        return true;
    }
    $order->update_meta_data( '_stripe_lock_payment', $new_lock );  // WRITE
    $order->save_meta_data();
    return false;
}
```

Two concurrent PHP processes can both execute `is_order_payment_locked()` before either writes the lock, both see "not locked" and both proceed. This is the exact same pattern that was identified in v8.8.2 when transients were replaced with order meta, but order meta has the same TOCTOU vulnerability when reads are served from Memcached/Redis.

A truly atomic lock would use `wp_cache_add()` (which maps to Memcached's `ADD` command, a single atomic operation that fails if the key already exists) or a database-level `INSERT ... ON DUPLICATE KEY` with a unique constraint.

### Related issues

This is the same family of bugs reported in:

* [#981](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/981>) - Original 2019 report with stack traces showing the race condition
* [#2331](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2331>) - "Potential duplicate webhook actions" - @mattallan confirmed object caching as root cause (Nov 2024)
* [#2463](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2463>) - "Double Notes & Double Emails" - Led to v8.8.2 fixes (transient → meta refactor + async webhooks)
* [#3501](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/3501>) - "Duplicate Vendor/Customer Order Emails" - Confirmed fixed by v8.8.2 for most users
* woocommerce/woocommerce-gateway-stripe#4271 - "Duplicate status changes causes double emails" (Apr 2025) - Fixed by PR woocommerce/woocommerce-gateway-stripe#4437 for redirect methods, but **not for the AJAX credit card path**
* **PR** woocommerce/woocommerce-gateway-stripe#4437 - "Add order locking to payment redirect processing" - Merged into v9.6.0 (Jul 2025)

### Workaround

We are currently using a `woocommerce_payment_complete_reduce_order_stock` filter with `wp_cache_add()` as an atomic mutex to block the second stock reduction. This prevents the double stock reduction symptom but not the duplicate emails/order notes.

Disabling the external object cache also eliminates the issue (as confirmed by @clickstatement in [#2331](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2331>)) but is not a viable production solution.

Mayisha:
Fixes STRIPE-955
Fixes woocommerce/woocommerce-gateway-stripe#5042 

As reported in woocommerce/woocommerce-gateway-stripe#5042 'update_order_status_ajax' was not checking for or acquiring any order lock. So it leaves a possibility for race condition and duplicate order processing.

Added order lock in 'update_order_status_ajax' method as a safety guard to prevent this.

## Testing instructions

* Code review (the real race condition is hard to reproduce).
* Regression test
  * Add a product to your cart and go to the checkout page as a shopper.
  * Add a [3DS card](<https://docs.stripe.com/testing#authentication-and-setup>) for payment and complete the checkout.
  * Verify that the payment is successful.
  * As an admin, check the order and verify that everything looks good.

### Changelog entry

#### Comment

</details>

**Post merge**


wjrosa (approved):
Improvement looks good 👍

Christine-Pinto:
Hey team! I'm experimenting with an AI-powered QA agent that reviews PRs for edge cases. I ran it against this PR to see if it could find anything beyond what CodeRabbit already identified.       
                                                                                                    
  ## Existing Finding (Acknowledged)                                                                
                                                                                                    
  CodeRabbit correctly identified that `wp_send_json_*` calls `exit()`, preventing the `finally` block from running. This is a critical issue that needs addressing.                               
                                                                                                    
  ## Additional Issues Found                                                                        
                                                                                                    
  The agent found two additional concerns that weren't covered in the existing review:              
                                                                                                    
  ### 1. Check-Then-Act Race Condition in Lock Acquisition                                          
                                                                                                    
  The lock implementation itself may have a race condition:                                         
                                                                                                    
  ```php                                                                                            
  public function lock_order_payment( WC_Order $order ): bool {                                     
      if ( $this->is_order_payment_locked( $order ) ) {  // (1) READ                                
          return true;                                                                              
      }                                                                                             
      // ...                                                                                        
      $order->update_meta_data( '_stripe_lock_payment', $new_lock );  // (2) WRITE                  
      $order->save_meta_data();                                                                     
      return false;                                                                                 
  }                                                                                                 
```                                                                                                  
  In high-concurrency scenarios, two requests can both read false at step (1) before either writes  
  at step (2). Both proceed, both overwrite the lock, and both execute                              
  process_order_for_confirmed_intent.                                                               
                                                                                                    
Question: Is this lock acquisition atomic? If not, the race condition this PR aims to fix may still occur.                                                                                      
                                                                                                    
Possible fix: Use add_post_meta with unique=true for atomic lock acquisition, or implement a compare-and-swap mechanism.                                                                                                                                

### 2. False Success When Locked Order Fails          
                                                
```php                                                                                                 
  if ( $is_already_locked ) {                                                                       
      wp_send_json_success(                                                                         
          [ 'return_url' => $gateway->get_return_url( $order ) ],                                   
          200                                                                                       
      );                                                                                            
  }                                                                                                 
```                                                                                               
  If Request A holds the lock and fails (e.g., card declined), Request B has already returned success with get_return_url. The user gets redirected to "Order Received" but the order is actually FAILED.                                                                                  
                                                                                                    
  Question: Should a locked order return success, or should it return an error/retry state?         
                                                                                                    
  ---                                                                                               
  TL;DR: Beyond the finally block issue, the lock acquisition itself may not be atomic, and returning success for locked orders could send users to a success page for failed payments.       
                                                                                                    
  🤖 Analysis by an experimental QA agent that focuses on finding issues other reviewers missed. Feedback welcome!

daledupreez:
We definitely need to fix the AJAX return vs the `finally` block. That was noted by CodeRabbit in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5050#pullrequestreview-3846530850), but it was contracted by default.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -715,7 +716,27 @@ public function update_order_status_ajax() {
...
+					[ 'order_id' => $order_id ]
+				);
+
+				wp_send_json_success(
+					[
+						'return_url' => $gateway->get_return_url( $order ),
+					],
+					200
+				);
+			}
```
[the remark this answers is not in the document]
Mayisha: Addressed this in the latest commits.

client/api/index.js
```diff
@@ -373,19 +373,44 @@ export default class WCStripeAPI {
...
+									_ajax_nonce: nonce,
+								} )
+							);
+							return retryRequest( retryCall, retryCount + 1 );
+						}
+						return response.data.return_url;
+					} );
+
+				return retryRequest( verificationCall );
 			} );
```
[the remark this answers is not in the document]
Mayisha: Fixed.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -733,10 +767,15 @@ public function update_order_status_ajax() {
...
-			if ( $order ) {
+			if ( $order instanceof WC_Order ) {
 				// Remove the awaiting confirmation order meta, don't save the order since it'll be saved in the next `update_status()` call.
 				$order_helper->remove_payment_awaiting_action( $order, false );
 				$order->update_status( OrderStatus::FAILED );
+
+				// Unlock if the lock was not acquired by another code path.
+				if ( ! $is_already_locked ) {
+					$order_helper->unlock_order_payment( $order );
+				}
```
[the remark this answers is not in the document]
Mayisha: Fixed.

tests/phpunit/WC_Stripe_Intent_Controller_Test.php
```diff
@@ -548,6 +550,156 @@ public function test_mandate_options_for_card_setup_intent_for_subscription() {
...
+
+		// Clean up
+		$order->delete();
+		WC_Stripe_Order_Helper::set_instance( $original_helper );
+		Ajax_Test_Helper::remove_hooks();
+
+		$this->assertNotNull( $response, 'Controller output should be valid JSON. Output: ' . $output );
+		$this->assertTrue( $response['success'] );
+		$this->assertArrayHasKey( 'return_url', $response['data'] );
+	}
```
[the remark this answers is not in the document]
Mayisha: Added.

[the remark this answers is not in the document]
Mayisha: Already removed.

client/api/index.js
```diff
@@ -373,19 +373,49 @@ export default class WCStripeAPI {
...
+										paymentMethodToSave || null,
+									_ajax_nonce: nonce,
+								} )
+							);
+							return retryRequest( retryCall, retryCount + 1 );
+						}
+						return response.data.return_url;
+					} );
+
+				return retryRequest( verificationCall );
```
daledupreez, line 418: Not blocking, but there is a _lot_ of nesting here along with heavy use of variables accessible within the scope (e.g. `this`, `ajaxAction`, `orderId`, `paymentMethodToSave`, and `nonce`). Maybe something to tackle in a follow-up, as it feels like it would be easier to follow the logic if we pushed more logic into named `const`s or `function`s.

client/api/index.js
```diff
@@ -373,19 +373,49 @@ export default class WCStripeAPI {
...
+						if ( response.data.status === 'processing' ) {
+							if ( retryCount >= maxRetries ) {
+								throw new Error(
+									__(
+										'Your order is still being processed. Please check your email for confirmation or view your order status in My account → Orders.',
+										'woocommerce-gateway-stripe'
+									)
+								);
+							}
+							const retryAfter = 5000; // 5 seconds
```
daledupreez, line 401: Why 5 seconds? That feels like a pretty long time to wait, especially if we're allowing 5 retries.

includes/class-wc-stripe-intent-controller.php
daledupreez, line 778: I think we need an additional flag to capture whether the code already unlocked the order -- in the code above, we call `$order_helper->unlock_order_payment()` just before calling `wp_send_json_success()`, but we still have this catch block that may end up removing a lock added by another process. It's extremely unlikely especially given how much caching is in play, but adding a local variable to guard against this edge case feels like it would be a good idea, even if it is super-defensive.

tests/phpunit/WC_Stripe_Intent_Controller_Test.php
```diff
@@ -548,6 +550,210 @@ public function test_mandate_options_for_card_setup_intent_for_subscription() {
...
+
+		$response = json_decode( $output, true );
+
+		// Clean up
+		$order->delete();
+		WC_Stripe_Order_Helper::set_instance( $original_helper );
+		Ajax_Test_Helper::remove_hooks();
+
+		$this->assertNotNull( $response, 'Controller output should be valid JSON. Output: ' . $output );
+		$this->assertTrue( $response['success'] );
```
daledupreez, line 599: Not blocking: it may be worth asserting that `$response` is an array, rather than getting indirect warnings or errors because the value is not an array. (This applies to a few of the tests.)
```suggestion
		$this->assertIsArray( $response );
		$this->assertTrue( $response['success'] );
```

phpstan-baseline.neon
```diff
@@ -2730,6 +2700,12 @@ parameters:
 			count: 1
 			path: includes/class-wc-stripe-intent-controller.php
 
+		-
+			message: '#^Unreachable statement \- code above always terminates\.$#'
```
daledupreez, line 2704: This error is extremely suspicious: it looks like 2 code blocks are now unreachable. What are they? Can we remove them?

daledupreez (commented):
I have some feedback from inspection. I plan to test later today.

Mayisha:
### Update
I've reworked this on top of the latest `develop` (the branch had drifted quite a bit since it was opened — the test file had been renamed and the handler rewritten), so this is a force-push over the previous commits. Summary of what changed from the earlier version:

**Dropped**
- The client-side retry/polling loop in `client/api/index.js`. It was added to cover the case where the request holding the lock fails and leaves the order in a bad state — but in this flow that lock is held by the deferred `payment_intent.succeeded` webhook, which is finalizing a payment Stripe has already confirmed succeeded. There's no failure there for a retry to recover from, so the request that finds the order locked can safely return the thank-you URL. The rest of the plugin (`process_upe_redirect_payment()`, `verify_intent_after_checkout()`) already bails this way when it finds the order locked — there's no polling anywhere else — so the AJAX path now follows that same convention. This also drops the 5s × 5 wait and the nesting that came up in review.

**Added / changed**
- `update_order_status_ajax()` now guards before completing the order: if a concurrent request already settled it (status `processing`/`completed`/`on-hold`, or the `_stripe_upe_redirect_processed` flag is set) it returns the thank-you URL without re-processing; otherwise it acquires the payment lock and only processes if it got the lock, releasing it in a `finally`. This mirrors the guard sequence in `process_upe_redirect_payment()`.
- Structured as `if/elseif/else` so the unlock lives only in the branch that acquired the lock. The `catch` blocks never unlock, so no path can remove a lock held by another process — no separate "already unlocked" flag needed.
- Tests cover each skip path: already-settled statuses (data provider), the redirect-processed flag, an already-locked order leaving the existing lock intact, and the lock being released after both success and failure.

**Re: earlier comments**
- Winner's-lock concern — handled by the structure above (unlock scoped to the acquiring branch).
- The "suspicious unreachable code" PHPStan errors are gone; there's no dead `return` after `wp_send_json_success()` now, and I removed the obsolete baseline entries instead of adding an unreachable-code one.
- `assertIsArray()` added before the response assertions.

phpstan-baseline.neon
Mayisha: Not applicable in the latest changes.

tests/phpunit/WC_Stripe_Intent_Controller_Test.php
Mayisha: Added.

client/api/index.js
Mayisha: Not applicable anymore.

includes/class-wc-stripe-intent-controller.php
Mayisha: Gracefully handled in the latest version of the code. See https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5050#issuecomment-5006491907

client/api/index.js
Mayisha: Note applicable after the latest changes. See https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5050#issuecomment-5006491907

Mayisha:
@coderabbitai review

includes/class-wc-stripe-intent-controller.php
```diff
@@ -721,20 +721,53 @@ public function update_order_status_ajax() {
...
+				// winner's lock would defeat the mutual exclusion.
+				WC_Stripe_Logger::info( "Skipping update_order_status_ajax for order $order_id; order payment is already locked." );
+				wp_send_json_success(
+					[
+						'return_url' => $gateway->get_return_url( $order ),
+					],
+					200
+				);
+			} else {
+				$processing_started = true; // Past validation; a failure from here on is a genuine payment failure.
```
wjrosa: Not blocking (AI-assisted):

A concurrent `payment_intent.succeeded` webhook is silently dropped ([class-wc-stripe-webhook-handler.php#L1334](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-webhook-handler.php#L1334) returns 200, so Stripe won't redeliver). 

If this request then dies before unlocking (hard timeout — `finally` won't run), the order stays pending on a captured payment. The checkout-session handler solves this by re-queueing via `defer_webhook_processing()` when locked ([#L1985](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-webhook-handler.php#L1985)); the `payment_intent` branch could do the same.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -721,20 +721,53 @@ public function update_order_status_ajax() {
...
+					$order_helper->unlock_order_payment( $order );
+				}
+
+				wp_send_json_success(
+					[
+						'return_url' => $gateway->get_return_url( $order ),
+					],
+					200
+				);
 			}
```
diegocurbelo: [nit] The same wp_send_json_success() payload is emitted three times across this block, maybe we can re-write it to something more concise:
```suggestion
			$already_settled = $order->has_status( [ OrderStatus::PROCESSING, OrderStatus::COMPLETED, OrderStatus::ON_HOLD ] )
				|| $order_helper->get_stripe_upe_redirect_processed( $order );

			if ( ! $already_settled ) {
				// lock_order_payment() returns true when another request already holds the lock.
				if ( $order_helper->lock_order_payment( $order ) ) {
					// Bail without unlocking — releasing the winner's lock would defeat the mutual exclusion.
					WC_Stripe_Logger::info( "Skipping update_order_status_ajax for order $order_id; order payment is already locked." );
				} else {
					$processing_started = true; // Past validation; a failure from here on is a genuine payment failure.
					try {
						$gateway->process_order_for_confirmed_intent( $order, $intent_id_received, $save_payment_method );
					} finally {
						$order_helper->unlock_order_payment( $order );
					}
				}
			}

			wp_send_json_success( [ 'return_url' => $gateway->get_return_url( $order ) ], 200 );
```

diegocurbelo (approved):
Looks good to me, and it tests well 👍🏼 

Left a non-blocker suggestion.

includes/class-wc-stripe-intent-controller.php
Mayisha: Good call 👍

includes/class-wc-stripe-intent-controller.php
Mayisha: Good idea. I will explore it in a follow-up.

PR #5058, opened 2026-02-24
Adaptive pricing payment processing support in backend

Mayisha:

- Introduced `wc_stripe_checkout_session_id` flag to include in the /checkout request and use it in backend to go through the checkout session path to process payment.
- Introduced checkout session-related new order metas in `WC_Stripe_Order_Helper` class.
- Added `process_payment_with_checkout_session` method for payment processing.
- Added a helper method `get_woocommerce_amount_from_stripe_amount` to convert the Stripe amount to WooCommerce amount (example: for 2 decimal currencies 7800 is converted to 78.00)
- Added `get_order_by_checkout_session_id` helper method.
- Added webhook handler for `checkout.session.completed` event (other events will be added in a separate PR)

## Testing instructions
- Code review.
- Functional testing to be done later with frontend support.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1391,86 @@ public function process_account_updated( $notification ) {
...
+			return;
+		}
+
+		$intent_id = $checkout_session->payment_intent;
+
+		// Store the payment intent ID on the order.
+		$order_helper->add_payment_intent_to_order( $intent_id, $order );
+
+		// Add presentment details if available.
+		if ( ! empty( $checkout_session->presentment_details ) ) {
```
wjrosa: Maybe we can replace this check with an `isset` and pass the required properties

wjrosa (approved):
Change looks good to me! But I agree with coderabbit's suggestions

includes/class-wc-stripe-order-helper.php
```diff
@@ -484,6 +505,81 @@ public function update_stripe_setup_intent_id( ?WC_Order $order = null, string $
...
+	 *
+	 * @since 10.5.0
+	 *
+	 * @param WC_Order|null $order
+	 * @param string $presentment_amount
+	 * @return false|void
+	 */
+	public function update_stripe_presentment_amount( ?WC_Order $order = null, string $presentment_amount = '' ) {
+		return $this->update_order_meta( $order, self::META_STRIPE_PRESENTMENT_AMOUNT, $presentment_amount );
+	}
```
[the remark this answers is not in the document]
Mayisha: Fixed.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -971,6 +976,39 @@ private function process_payment_with_deferred_intent( int $order_id ) {
...
+
+		// Remove cart.
+		if ( WC()->cart ) {
+			WC()->cart->empty_cart();
+		}
+
+		return [
+			'result'   => 'success',
+			'redirect' => $this->get_return_url( $order ),
+		];
```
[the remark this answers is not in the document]
Mayisha: We only update the checkout session order meta here. So it seems unnecessary to lock the order.

includes/class-wc-stripe-helper.php
```diff
@@ -277,6 +277,31 @@ public static function get_stripe_amount( $total, $currency = '' ) {
...
+			$decimals = 0;
+		} elseif ( in_array( $currency, self::three_decimal_currencies(), true ) ) {
+			$amount   = (float) $stripe_amount / 1000;
+			$decimals = 3;
+		} else {
+			$amount   = (float) $stripe_amount / 100;
+			$decimals = 2;
+		}
+		return wc_format_decimal( $amount, $decimals );
+	}
```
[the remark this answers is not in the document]
Mayisha: Added.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1464,6 +1544,9 @@ public function process_webhook( $request_body ) {
 			case 'setup_intent.succeeded':
 			case 'setup_intent.setup_failed':
 				$this->process_setup_intent( $notification );
+				break;
+			case 'checkout.session.completed':
+				$this->process_checkout_session( $notification );
```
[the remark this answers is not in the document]
Mayisha: This is the last case. So break isn't necessary here.

includes/class-wc-stripe-helper.php
```diff
@@ -952,6 +977,49 @@ public static function get_order_by_setup_intent_id( $intent_id ) {
...
+							'key'   => '_stripe_checkout_session_id',
+							'value' => $checkout_session_id,
+						],
+					],
+				]
+			);
+			$order_id = current( $orders ) ? current( $orders )->get_id() : false;
+		} else {
+			$order_id = $wpdb->get_var( $wpdb->prepare( "SELECT DISTINCT ID FROM $wpdb->posts as posts LEFT JOIN $wpdb->postmeta as meta ON posts.ID = meta.post_id WHERE meta.meta_value = %s AND meta.meta_key = %s", $checkout_session_id, '_stripe_checkout_session_id' ) );
+		}
```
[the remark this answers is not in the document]
Mayisha: Added a check.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
daledupreez: I think that the follow-up comment here is relevant, and deserves a clear response as to why the current logic is (not) OK.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -971,6 +976,39 @@ private function process_payment_with_deferred_intent( int $order_id ) {
...
+
+		if ( ! $order instanceof \WC_Order ) {
+			return [
+				'result'   => 'failure',
+				'redirect' => '',
+			];
+		}
+
+		$order_helper = WC_Stripe_Order_Helper::get_instance();
+		$order_helper->update_stripe_checkout_session_id( $order, $checkout_session_id );
```
daledupreez, line 1036: Should we add a check to make sure the checkout session ID is not already set for the order? Or do we have guards higher up the stack to prevent overwriting order details/meta/state?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -971,6 +976,39 @@ private function process_payment_with_deferred_intent( int $order_id ) {
...
+		$order->save_meta_data();
+
+		// Remove cart.
+		if ( WC()->cart ) {
+			WC()->cart->empty_cart();
+		}
+
+		return [
```
daledupreez, line 1047: It may be worth clarifying how things work with the checkout session in place. Simply returning `success` here feels a bit unexpected to me!

includes/class-wc-stripe-helper.php
```diff
@@ -277,6 +277,31 @@ public static function get_stripe_amount( $total, $currency = '' ) {
...
 	}
 
+	/**
+	 * Converts a Stripe amount (smallest currency unit) to WooCommerce amount.
+	 *
+	 * @param int    $stripe_amount Amount in Stripe's smallest unit (e.g. cents).
+	 * @param string $currency      Currency code (e.g. 'eur', 'usd').
+	 * @return string Formatted amount for display.
+	 */
+	public static function get_woocommerce_amount_from_stripe_amount( $stripe_amount, $currency = '' ) {
```
daledupreez: Is it worth adding explicit types to these parameters?
```suggestion
	public static function get_woocommerce_amount_from_stripe_amount( int $stripe_amount, string $currency = '' ) {
```

includes/class-wc-stripe-helper.php
```diff
@@ -952,6 +977,53 @@ public static function get_order_by_setup_intent_id( $intent_id ) {
...
+						[
+							'key'   => '_stripe_checkout_session_id',
+							'value' => $checkout_session_id,
+						],
+					],
+				]
+			);
+			$order_id = current( $orders ) ? current( $orders )->get_id() : false;
+		} else {
+			$order_id = $wpdb->get_var( $wpdb->prepare( "SELECT DISTINCT ID FROM $wpdb->posts as posts LEFT JOIN $wpdb->postmeta as meta ON posts.ID = meta.post_id WHERE meta.meta_value = %s AND meta.meta_key = %s", $checkout_session_id, '_stripe_checkout_session_id' ) );
```
daledupreez, line 1012: Out of interest, why are we using a direct query when HPOS is disabled? Would it not be simpler to use `wc_get_orders()` for both code paths? It would be good to document the why in a comment here, as it's not an obvious choice to me.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1391,120 @@ public function process_account_updated( $notification ) {
...
+	 * @param object $notification The notification from Stripe
+	 * @return void
+	 */
+	public function process_checkout_session( $notification ) {
+		$checkout_session = $notification->data->object;
+
+		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $checkout_session->id );
+
+		if ( ! $order instanceof \WC_Order ) {
+			WC_Stripe_Logger::error( 'Could not find order via checkout session ID: ' . $checkout_session->id );
```
daledupreez, line 1406: Is it worth noting the type of `$order` for debugging purposes?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1391,120 @@ public function process_account_updated( $notification ) {
...
+			return;
+		}
+
+		if ( ! $order->has_status(
+			apply_filters(
+				'wc_stripe_allowed_payment_processing_statuses',
+				[ OrderStatus::PENDING, OrderStatus::FAILED ],
+				$order
+			)
+		) ) {
```
daledupreez: For clarity and to make it easier to document the filter, I think we should expand condition into multiple distinct steps.
```suggestion
		/** This filter is documented in includes/class-wc-stripe-webhook-handler.php */
		$allowed_payment_processing_statuses = apply_filters(
			'wc_stripe_allowed_payment_processing_statuses',
			[ OrderStatus::PENDING, OrderStatus::FAILED ],
			$order
		);
		if ( ! $order->has_status( $allowed_payment_processing_statuses ) ) {
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1391,120 @@ public function process_account_updated( $notification ) {
...
+						$e->getMessage()
+					)
+				);
+			}
+
+			$this->send_failed_order_email( $order->get_id(), $status_update );
+		}
+
+		// Unlock the order
+		$order_helper->unlock_order_payment( $order );
```
daledupreez: Just to be extra safe, does it make sense to have this in a `finally` block? That way any exceptions or failures thrown from the `catch` block (including the error action and sending the email) won't prevent the lock from being released.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1391,120 @@ public function process_account_updated( $notification ) {
...
+
+				$amount = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount(
+					$presentment_details->presentment_amount,
+					$presentment_details->presentment_currency
+				);
+
+				$order->add_order_note(
+					sprintf(
+						/* translators: 1) presentment currency 2) presentment amount */
+						__( 'Presentment amount: %1$s %2$s', 'woocommerce-gateway-stripe' ),
```
daledupreez: Nit: This is not super-helpful wording for merchants. I think we should aim for more user-friendly wording.
```suggestion
						__( 'Local currency purchase via Adaptive Pricing. Amount paid was: %1$s %2$s', 'woocommerce-gateway-stripe' ),
```

includes/class-wc-stripe-webhook-handler.php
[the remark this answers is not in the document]
daledupreez: Personal nit: excluding the `break` sets us up for future bugs if someone misses the need for adding it. I think we should help our future selves and add a `break`. It's the kind of thing that's easy to miss.

tests/phpunit/WC_Stripe_Helper_Test.php
```diff
@@ -279,6 +279,103 @@ public function provide_test_get_stripe_amount(): array {
...
+				'stripe_amount' => 100000,
+				'currency'      => 'bhd',
+				'expected'      => '100.000',
+			],
+			'BHD three-decimal: 100500 fils' => [
+				'stripe_amount' => 100500,
+				'currency'      => 'bhd',
+				'expected'      => '100.500',
+			],
+			'BHD three-decimal: 0' => [
```
daledupreez, line 360: Why don't we have a single-unit test for BHD? Or even some smaller units given that 1, 10, and 100 should all result in non-integer amounts.

phpstan-baseline.neon
```diff
@@ -3546,6 +3552,12 @@ parameters:
 			count: 1
 			path: includes/class-wc-stripe-webhook-handler.php
 
+		-
+			message: '#^Parameter \#1 \$intent of method WC_Stripe_Payment_Gateway\:\:get_latest_charge_from_intent\(\) expects object, object\|false given\.$#'
```
daledupreez: This seems like a legitimate issue we should guard against.

daledupreez (commented):

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1391,120 @@ public function process_account_updated( $notification ) {
...
+						strtoupper( $presentment_details->presentment_currency ),
+						$amount
+					)
+				);
+			}
+
+			$intent = $this->get_intent_from_order( $order );
```
malithsen, line 1468: Maybe worth adding a null guard here
```suggestion
			$intent = $this->get_intent_from_order( $order );
			if ( ! $intent ) {
                              throw new Exception( 'Payment intent not found for checkout session.' );
                        }
```

malithsen (commented):
Looks good overall. Agreeing with existing feedback. I had one additional suggestion.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
Mayisha: @coderabbitai I am still struggling to make sense of why we need the order lock here. 
In all other places where we have the locks, those methods involves multi layer changes; like changing order status, sending requests to Stripe for intent create/update, creating payment tokens etc. This method does not have complex logic and only stores the checkout session ID in the order meta. 🤔

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: I am not sure about this. 
I was hypothetically thinking of a scenario where for some reason the checkout fails and the order is in failed status but we had stored this meta in the order already. On the frontend, if it is part of the same session and user tries to checkout again after getting an error by refreshing the page (and creates new checkout session), shouldn't we allow to save the latest checkout session ID in the order meta? 🤔

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1391,120 @@ public function process_account_updated( $notification ) {
 		WC_Stripe_Logger::debug( 'Cleared account cache after receiving account.updated webhook.' );
 	}
 
+	/**
+	 * Processes the checkout session completed event.
+	 *
+	 * @param object $notification The notification from Stripe
+	 * @return void
+	 */
+	public function process_checkout_session( $notification ) {
```
```suggestion
	public function process_checkout_session( object $notification ): void {
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Done.

includes/class-wc-stripe-webhook-handler.php
Mayisha: Added a check for empty intent.

phpstan-baseline.neon
Mayisha: Fixed and removed this from baseline.

includes/class-wc-stripe-helper.php
Mayisha: Tbh I actually followed the `get_order_by_setup_intent_id` method here.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1391,6 +1391,133 @@ public function process_account_updated( $notification ) {
...
+	 */
+	public function process_checkout_session( object $notification ): void {
+		$checkout_session = $notification->data->object;
+
+		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $checkout_session->id );
+
+		if ( ! $order instanceof \WC_Order ) {
+			WC_Stripe_Logger::error( 'Could not find order via checkout session ID: ' . $checkout_session->id );
+			return;
+		}
```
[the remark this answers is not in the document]
Mayisha: That's a good call. I think we should defer the webhook here in case the webhook arrives earlier than the `process_payment` method call which stores this meta in the order.

I will make the improvement in a follow up PR under STRIPE-921.

includes/class-wc-stripe-helper.php
Mayisha: As per https://developer.woocommerce.com/docs/extensions/core-concepts/wc-get-orders/ `meta_query` is available for HPOS. So when in legacy mode, we need the direct query. I have added this in the comment in docblock.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: That's fair. Just wanted to make sure we thought this through before allowing arbitrary updates to the order.

includes/class-wc-stripe-webhook-handler.php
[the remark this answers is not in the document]
daledupreez: Not sure if you saw my last comment, @Mayisha, so unresolving to check.

daledupreez (approved):
Changes are looking good to me, though we may want to tackle some of my non-blocking comments before merging.

includes/class-wc-stripe-webhook-handler.php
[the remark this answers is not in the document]
Mayisha: I forgot to address this in this PR @daledupreez. But it has been taken care of in the webhook hanlding PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5084/changes#diff-84de206a98f656cf114e4d67162aa7b82d887901669879283fa8e845186374dbR1749-R1758

Mayisha:
> though we may want to tackle some of my non-blocking comments before merging.

@daledupreez just to confirm, did you mean the latest 2 comments or any of the previous comments? 👀

daledupreez:
@Mayisha, I think it was just the one with the `break` that was particularly relevant. There was only one comment, as it turns out. :flushed:

PR #5062, opened 2026-02-25
Rename and autoload checkout sessions ajax handler class

wjrosa:
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4260

As a way to introduce a class autoloader to our extension, we are doing it for new classes. In this PR, I am moving and renaming the first class, responsible for creating checkout sessions, since this feature is behind a feature flag and still under development.

The new class name `WC_Stripe_Checkout_Sessions_Ajax_Handler` better fits the class's purpose, as it introduces a new Ajax handler (not an API endpoint).

I am using the default composer autoloader here, since we don't need to support autoloading for the other existing files.

I am also adding the `vendor` directory to the release package, updating the existing build scripts using the original [POC PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4260) as a reference.

### Regression

- Checkout and build this branch on your test environment (`dev/rename-and-move-checkout-sessions-ajax-handler`)
- Add the following filter to allow adaptive pricing and checkout session in your store:
```
```

- Connect a US Stripe account
- Enable some payment methods
- As a shopper, add a product to your cart and go to any checkout page
- Verify that the currency selector is displayed and the available currencies are USD and your local currency

### Package build

- Run `npm run build`
- Confirm the release package is generated inside the extension directory
- Extract the package
- Confirm the `vendor` directory is there

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
@copilot fix PHPUnit tests for this PR

wjrosa:
@daledupreez @malithsen to make the number of changes here smaller, I decided to remove the namespace inclusion. So, now we are just literally renaming and autoloading the class.

package.json
```diff
@@ -124,7 +124,8 @@
     "postinstall": "composer install && npx playwright install",
     "prebuild": "npm install && composer install",
     "prepare": "is-ci || husky install",
-    "build": "npm run uglify && npm run sass && rimraf build/* && npm run build:webpack && npm run build:i18n && npm run build:release",
+    "build": "npm run uglify && npm run sass && rimraf build/* && npm run build:webpack && npm run build:i18n && npm run build:deps && npm run build:release",
+    "build:deps": "rm -rf vendor && composer install --no-dev --classmap-authoritative",
```
daledupreez: As written, this is going to be problematic for the dev workflow, because all of our dev dependencies are going to be missing after running `build`. Even if we re-add them via `postbuild` or similar, that will really slow down our dev workflow. (Or maybe composer is just slow for me.)

Can we investigate whether we can generate the composer dependencies within the `release` folder instead? That will require making changes in a more intrusive way, but I think it will be worth it.

At an implementation level, that could be as simple as:
 * Copy `composer.json` and `composer.lock` to the `release/woocommerce-gateway-stripe/` directory
 * Change into the `release/woocommerce-gateway-stripe/` directory
 * Run `composer install --no-dev --classmap-authoritative`
 * Remove the copied `composer.json` and `composer.lock` files
 * Profit

The other benefit of an approach like the above is that we can be more confident that the release process is _creating_ the autoload files, and is not copying across any unexpected files from the `vendor` directory.

woocommerce-gateway-stripe.php
```diff
@@ -81,6 +81,12 @@ function woocommerce_gateway_stripe() {
 	static $plugin;
 
 	if ( ! isset( $plugin ) ) {
+		// Attempts to include the default composer autoloader.
+		$autoload_filepath = WC_STRIPE_PLUGIN_PATH . '/vendor/autoload.php';
+		if ( file_exists( $autoload_filepath ) ) {
```
daledupreez, line 86: This approach feels risky for developers as well: we are either going to have no autoload file, in which case some things will mysteriously fail, or our autoloader could be the classmap for `includes/` or it could be the autoloader for any dev dependencies.

I think we need a cleaner process that just works. Can we add `--classmap-authoritative` to more of our scripts to make that happen by default?

daledupreez (changes requested):

package.json
wjrosa: Instead of your suggestion, what do you think about adding a new `build-dev` command instead?

woocommerce-gateway-stripe.php
wjrosa: > in which case some things will mysteriously fail

We would see a fatal error for a file not found. I think it is a clear error. But happy to improve the implementation

> I think we need a cleaner process that just works. Can we add --classmap-authoritative to more of our scripts to make that happen by default?

You mean add the flag to both `postinstall` and `prebuild`?

wjrosa: Added to both in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5062/commits/0f1838a77ff4b32376685f0d9fa8e03ed082bceb

package.json
wjrosa: I applied my suggestion in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5062/commits/2d9d0068c999a441efdc2aaf2d40260ad475961b

package.json
```diff
@@ -121,10 +121,13 @@
...
   },
   "scripts": {
-    "postinstall": "composer install && npx playwright install",
-    "prebuild": "npm install && composer install",
+    "postinstall": "composer install --classmap-authoritative && npx playwright install",
+    "prebuild": "npm install && composer install --classmap-authoritative",
     "prepare": "is-ci || husky install",
-    "build": "npm run uglify && npm run sass && rimraf build/* && npm run build:webpack && npm run build:i18n && npm run build:release",
+    "build": "npm run uglify && npm run sass && rimraf build/* && npm run build:webpack && npm run build:i18n && npm run build:deps && npm run build:release",
+		"build-dev": "npm run uglify && npm run sass && rimraf build/* && npm run build:webpack && npm run build:i18n && npm run build:deps-dev && npm run build:release",
```
daledupreez: I am not a fan of adding a `build-dev` command, because very few developers will think to check whether a special build command is needed. I think it would be better to have one `build` command that keeps the developer experience as expected, and runs `composer install --no-dev --classmap-authoritative` in the `release` directory with the copied `composer.json` and `composer.lock` files.

While that is more expensive now, it's going to be much cheaper in terms of longer-term developer confusion.

daledupreez: Oh, and the unit test failures seem to be stemming from older WooCommerce versions which don't include a class we are failing to autoload.

wjrosa: Alright. Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5062/commits/5d88c472d74dd254b7296075004f7edd6a513a40

malithsen (approved):
This looks good to me

tasks/release.js
```diff
@@ -38,6 +38,13 @@ rm( 'build/*.map' );
 // copy the directories to the release folder
 cp( '-Rf', filesToCopy, targetFolder );
 
+// Install production-only Composer dependencies in the release folder
+cp( 'composer.json', targetFolder );
+cp( 'composer.lock', targetFolder );
+exec( `composer install --no-dev --classmap-authoritative --working-dir=${ targetFolder }` );
+rm( targetFolder + '/composer.json' );
+rm( targetFolder + '/composer.lock' );
+
```
daledupreez, line 47: 😍

daledupreez (changes requested):
The single build routine that loads the dependencies in the release folder is working well.

However, there are still two remaining issues to address for CI:
 * For PHP unit tests, the previous `class_exists()` guards for the unit tests no longer work because the classes are now autoloadable, so we bypass the logic in the setup that skips loading various classes when the interface(s) they implement are not available in the WooCommerce setup.
 * For the E2E tests, we're mapping the development code into the e2e Docker environment, which means we're including the development autoloader, and our autoloaded stubs are causing an exception when trying to activate WooCommerce Subscriptions. We need to make sure we're using the built code for the e2e tests.
   - This also has implications for local e2e tests by developers -- at least some things may break if developers don't (re)build the code in their local environment.

wjrosa:
@daledupreez Thanks for the review and for the tips on the root cause of the issue. I fixed them with the help of Claude in my latest commits

composer.json
```diff
@@ -23,6 +23,15 @@
...
+        "includes/ajax-handlers/"
+      ],
+      "exclude-from-classmap": [
+        "/tests/phpunit/helpers/",
+        "/tests/phpunit/stubs/"
+      ]
+    },
     "autoload-dev": {
       "classmap": ["tests/phpunit/"]
     },
```
daledupreez: Nit: Should the `exclude-from-classmap` entries not go under `autoload-dev` as we only specify `tests/phpunit/` in `autoload-dev`?
```suggestion
    },
    "autoload-dev": {
      "classmap": ["tests/phpunit/"],
      "exclude-from-classmap": [
        "/tests/phpunit/helpers/",
        "/tests/phpunit/stubs/"
      ]
},
```

daledupreez (commented):
Will try to test more extensively tomorrow.

daledupreez (approved):
Thanks for making those changes! I am not going to get to additional manual testing this week, but I think this is looking good.

PR #5067, opened 2026-02-26
Add missing metadata to checkout session

### LINKED ISSUE #5069: Checkout Sessions: Add missing metadata to checkout sessions
Since payment processing for checkout sessions differs from the regular payment intent flow, some of the metadata we set for payment intents is not present on the Stripe side. We need to manually set them when processing the checkout session complete webhook event.

wjrosa:
Fixes STRIPE-962
Fixes #5069

When processing payment intents in the regular flow, we set some metadata on the Stripe side for debugging and support purposes. The checkout session flow is a bit different, and we cannot set these keys when creating the session, as we don't have an order yet. In this PR, I am setting the missing data asynchronously using Action Scheduler, 2 minutes after the `checkout.session.completed` event arrives.

## Testing instructions

- Code review.
- Functional testing to be done later with frontend support.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
@copilot add unit tests for this PR

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1473,6 +1502,21 @@ public function process_checkout_session( $notification ) {
 
 			$charge->is_webhook_response = true;
 			$this->process_response( $charge, $order );
+
+			// Schedule a job to store the remaining metadata to the checkout session.
+			$this->action_scheduler_service->schedule_job(
+				time() + $this->deferred_webhook_delay,
```
malithsen: Should we use a separate delay for this job instead of reusing `deferred_webhook_delay`? If we later update `deferred_webhook_delay` for some reason, it would inadvertently change the metadata delay too.

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
```diff
@@ -884,4 +889,231 @@ public function provide_test_process_webhook_refund_updated() {
...
+
+		// Inject the mock action scheduler service.
+		$prop = new ReflectionProperty( WC_Stripe_Webhook_Handler::class, 'action_scheduler_service' );
+		$prop->setAccessible( true );
+		$prop->setValue( $this->mock_webhook_handler, $mock_scheduler );
+
+		$this->mock_webhook_handler->process_checkout_session( $notification );
+
+		// Assertions are handled by the mock expectation (never called).
+		$this->assertTrue( true );
```
malithsen, line 1125: I guess this empty assertion is needed to prevent PHPUnit from marking this test 'risky'. Actual assertions happen in the `mock_scheduler->expects( $this->never() )` line. Should we add a comment on why this is needed?

malithsen (approved):
Looks good to me on inspection. Left some non-blocking suggestions.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Makes sense! Added a specific value in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5067/commits/0acad30815f8f75f32b218d6f4686dcfdeb83f21

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
wjrosa: I have updated the test comment in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5067/commits/e84084acc8b44caf39f41c261ae6734f675737a1. I used your comment as a reference 👍

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1344,6 +1359,27 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+	/**
+	 * Processes the checkout session metadata update event to store additional metadata on the checkout session object.
+	 *
+	 * @param string $checkout_session_id The checkout session ID.
+	 * @param array $metadata The metadata from the checkout session.
+	 * @return void
+	 */
+	public function process_checkout_session_metadata( string $checkout_session_id, array $metadata ): void {
+		try {
+			$response = WC_Stripe_API::request( [ 'metadata' => $metadata ], 'checkout/sessions/' . $checkout_session_id, 'POST' );
```
daledupreez, line 1371: Is it worth running a `GET` first to confirm what values are already present?

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
```diff
@@ -884,4 +889,231 @@ public function provide_test_process_webhook_refund_updated() {
...
+		try {
+			$handler->process_checkout_session_metadata( $checkout_session_id, $metadata );
+		} catch ( Exception $e ) {
+			$caught = $e;
+		}
+
+		remove_filter( 'pre_http_request', $pre_http_filter );
+
+		$this->assertNotNull( $caught, 'Expected an exception to be thrown.' );
+		$this->assertSame( $error_message, $caught->getMessage() );
```
daledupreez, line 986: Nit: Is it worth confirming that the type of exception was the one we expect?

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
```diff
@@ -884,4 +889,231 @@ public function provide_test_process_webhook_refund_updated() {
...
+			],
+		];
+
+		// Mock the action scheduler service.
+		$mock_scheduler = $this->createMock( WC_Stripe_Action_Scheduler_Service::class );
+		$scheduled_args = null;
+		$mock_scheduler->expects( $this->once() )
+			->method( 'schedule_job' )
+			->with(
+				$this->isType( 'int' ),
```
daledupreez: Nit: I think we should verify that it is an integer slightly larger than the initial timestamp plus the expected delay.

daledupreez (approved):
The changes look good to me, though I do have some non-blocking comments on the tests.

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
wjrosa: Currently, it is always `Exception`. But if you think it is better, I can throw `Stripe_Exception` instead for the API call (like we do in some places).

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
wjrosa: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5067/commits/dc3d107cea91a599f741acbc6568d6e43b354b41 and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5067/commits/b4e56edf1716b523104e8497bddcbb1bbddc10bb

tests/phpunit/WC_Stripe_Webhook_Handler_Test.php
wjrosa: I went ahead and performed the change in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5067/commits/d87a361c3bbf03cf522d5d75bab1d5f313058813

includes/class-wc-stripe-webhook-handler.php
wjrosa: I am not sure. We would just be making an additional API request that is irrelevant most of the time (or always). Since we don't have any other way to update the metadata. I will leave it as-is for now, and we can revisit it later if it makes sense.

PR #5070, opened 2026-02-26
Fix checkout session creation for guest customers

### LINKED ISSUE #5039: Checkout Sessions: Support guest checkout
Currently, an error happens when trying to create a checkout session without a logged-in user, due to missing billing address information. We need to address that and make sure the integration works correctly for guest users.

wjrosa:
Fixes STRIPE-953
Fixes #5039

Currently, an error occurs when trying to create a checkout session without a logged-in user because the billing address is missing. In this PR, I am adding a check for logged-in status and adding the related information accordingly. When a user is a guest, we now skip including the customer key when requesting Stripe's checkout session creation endpoint.

I am taking the opportunity to improve code readability by adding new private methods to build both line items and payment-intent data.

And I am fixing an old issue with the `maybe_create_customer` method, where we were returning `$this->set_id`. This method returns nothing (`void`).

## Testing instructions

- Checkout and build this branch on your test environment (`fix/checkout-sessions-with-guest-customer`)
- Add the following filter to allow adaptive pricing and checkout session in your store:
```
```

- Connect a US Stripe account
- Enable some payment methods
- As a guest shopper, add a product to your cart and go to the classic checkout page
- Verify that the currency selector is displayed and the available currencies are USD and your local currency

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -35,88 +35,33 @@ public function create_checkout_session(): void {
...
+			if ( is_user_logged_in() ) {
+				try {
+					$stripe_customer = new WC_Stripe_Customer( WC()->customer->get_id() );
+					$stripe_customer->maybe_create_customer();
+				} catch ( Exception $e ) {
+					throw new Exception( __( 'Unable to create or retrieve Stripe customer.', 'woocommerce-gateway-stripe' ) );
+				}
+
+				$request['customer'] = WC()->customer->get_id();
+			}
```
wjrosa: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/3655405e61e538c181940c10c3a7b0ad3f2d3a85

wjrosa:
@copilot fix unit tests for this PR

includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
```diff
@@ -134,4 +79,82 @@ public function create_checkout_session(): void {
...
+	 * @return array
+	 */
+	private function build_line_items(): array {
+		$currency   = get_woocommerce_currency();
+		$line_items = [];
+		foreach ( WC_Stripe_Helper::build_line_items() as $raw_line_item ) {
+			if ( 'total_discount' === ( $raw_line_item['key'] ?? '' ) ) {
+				// TODO: Stripe Checkout handles discounts/coupons differently. Skip for now.
+				continue;
+			}
```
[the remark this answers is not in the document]
wjrosa: This will be handled in another PR

includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
```diff
@@ -33,88 +33,32 @@ public function create_checkout_session(): void {
...
+				'payment_intent_data'           => $this->build_payment_intent_data(),
 				'mode'                          => 'payment',
 				'adaptive_pricing'              => [
 					'enabled' => 'true',
 				],
 			];
 
+			if ( is_user_logged_in() ) {
+				try {
+					$stripe_customer = new WC_Stripe_Customer( WC()->customer->get_id() );
```
Mayisha, line 52: Should we check if `WC()->customer` exists as safeguard?

tests/phpunit/WC_Stripe_Checkout_Sessions_Controller_Test.php
Mayisha:
```suggestion
	 * @param array       $customer_data              The customer billing data to set.
	 * @param bool        $is_cart_empty              Whether the cart is empty.
```

The order of the params is wrong in the docblock.

tests/phpunit/WC_Stripe_Checkout_Sessions_Controller_Test.php
```diff
@@ -48,6 +48,7 @@ public function test_create_checkout_session(
 		Ajax_Test_Helper::init_hooks();
 
 		// Set up a logged-in user with billing details.
+		wp_set_current_user( 1 );
```
Mayisha: Shall we add tests for the guest user?

includes/class-wc-stripe-customer.php
```diff
@@ -416,7 +416,9 @@ private function get_billing_data_field( $field, $order = null ) {
 	 */
 	public function maybe_create_customer() {
 		if ( ! $this->get_id() ) {
-			return $this->set_id( $this->create_customer() );
+			$customer_id = $this->create_customer();
+			$this->set_id( $customer_id );
+			return $customer_id;
```
Mayisha, line 434: These changes have added a new PHPStan error. Instead of ignoring the error, shall we cast it to the right type or update the return type of this method accordingly?

Mayisha (commented):
This tests as described. I can see the currency selector on both classic and block checkout pages as a guest user 👍 

Image

tests/phpunit/WC_Stripe_Checkout_Sessions_Controller_Test.php
wjrosa: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/39f31a8149cf4abe6b632ca8012b32f6d04aef4e

includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
wjrosa: Added the extra check in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/39f31a8149cf4abe6b632ca8012b32f6d04aef4e

includes/class-wc-stripe-customer.php
wjrosa: Good call. Fixed this and other related issues in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/1ade98df7825239cae83a50a339af46ca645af53

Mayisha (approved):
Looks good to me after the latest changes 👍

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -132,4 +76,82 @@ public function create_checkout_session(): void {
...
+		$line_items = [];
+		foreach ( WC_Stripe_Helper::build_line_items() as $raw_line_item ) {
+			if ( 'total_discount' === ( $raw_line_item['key'] ?? '' ) ) {
+				// TODO: Stripe Checkout handles discounts/coupons differently. Skip for now.
+				continue;
+			}
+
+			$line_items[] = [
+				'price_data' => [
+					'currency' => strtolower( $currency ),
```
daledupreez: Non-blocking nit: It may be worth running the conversion once outside the loop rather than re-converting for every line item.

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -132,4 +76,82 @@ public function create_checkout_session(): void {
...
+					'product_data' => [
+						'name' => $raw_line_item['label'],
+					],
+					'unit_amount' => $raw_line_item['amount'],
+				],
+				'quantity' => 1, // @TODO: Handle quantity properly if needed.
```
daledupreez: Do we have an issue to track this TODO? It feels like we could easily lose track of the work if we don't.

includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
```diff
@@ -132,4 +76,82 @@ public function create_checkout_session(): void {
...
+			$metadata = array_merge(
+				$metadata,
+				[
+					'customer_name'  => $full_name,
+					'customer_email' => $email,
+				]
+			);
+		}
+
+		$data['metadata'] = apply_filters( 'wc_stripe_payment_metadata', $metadata, null, null );
```
daledupreez, line 122: Nit: We should add a PHPDoc reference to indicate where the filter is documented. And does the existing code for the filter pass in `null` for either of the final two arguments in the current code? I am concerned that passing null may break existing integrations.
```suggestion
		/** Documented in some/file.php */
		$data['metadata'] = apply_filters( 'wc_stripe_payment_metadata', $metadata, null, null );
```

daledupreez: At a more fundamental level, should we be sending this data to Stripe before the shopper decides to make a purchase? Do we send this data as aggressively/early in the current code without Adaptive Pricing enabled? And by that I mean before the shopper has clearly indicated they want to make a purchase using a payment method from the Stripe plugin.

includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
```diff
@@ -33,88 +33,32 @@ public function create_checkout_session(): void {
...
 				'mode'                          => 'payment',
 				'adaptive_pricing'              => [
 					'enabled' => 'true',
 				],
 			];
 
+			if ( is_user_logged_in() && WC()->customer instanceof WC_Customer ) {
+				try {
+					$stripe_customer = new WC_Stripe_Customer( WC()->customer->get_id() );
+					$stripe_customer->maybe_create_customer();
```
daledupreez, line 53: As per my comment RE sending customer metadata, I don't think we should create a customer during checkout session creation, as we don't know whether the shopper will pick a Stripe payment method, so we don't have a good reason to share their personal data with Stripe. I think it _might_ be OK to re-use an existing Stripe customer ID if the shopper already has one, but I think we should consider even this case.

includes/class-wc-stripe-checkout-sessions-controller.php
```diff
@@ -33,88 +33,32 @@ public function create_checkout_session(): void {
...
-					],
-				],
+				'payment_intent_data'           => $this->build_payment_intent_data(),
 				'mode'                          => 'payment',
 				'adaptive_pricing'              => [
 					'enabled' => 'true',
 				],
 			];
 
+			if ( is_user_logged_in() && WC()->customer instanceof WC_Customer ) {
```
daledupreez, line 50: Not blocking, but it feels a bit redundant that we are doing the same customer checks in this code and in `build_payment_intent_data()`. It might be cleaner to do the detection once and then use the result in both blocks of code.

daledupreez (commented):
The code looks OK from inspection, but I have some concerns about what data we send over the wire during checkout session creation.

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Moved outside the loop in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/121b0980d8c52672814f73c73483a9e700bedb8f

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Added the issue ID in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/f1315771ce802bbb19300a5bd5b4627a0b3c4279

includes/class-wc-stripe-checkout-sessions-controller.php
wjrosa: Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/f7395118fa02ce2b8fb5ae1ae1e555d7a2922cc8

includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
wjrosa: Doc added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/3eeba225d9f995076c13e7a140dda6b2dbcf4354 to both filter calls. As for the `null` value, we don't have any code hooking into it in the extension code currently. And there's no way to know if 3rd parties are doing it.

wjrosa: I moved the inclusion of most of the customer information to the checkout sessions during the payment processing in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/d978054709d8e84ec56eb64aeb2e4bd688167e2b, while keeping the flow here the same we have for guests, as discussed today. It required some major changes:
- The `customer` value cannot be updated later. So, I kept this information unchaged
- The child payment intent object cannot be updated. So, I moved the customer-related metadata, and the shipping information to the main checkout session object. I think it is OK since we use this just for debugging

includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
wjrosa: As previously mentioned, we cannot update the customer associated with the checkout session later. So, the only thing I could do here is to remove the creation part. But that would cause a bug in which a new customer would never have an account, always purchasing as a guest

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1041,6 +1095,50 @@ private function process_payment_with_checkout_session( int $order_id, string $c
 			WC()->cart->empty_cart();
 		}
 
+		// Add extra customer information to the checkout session on the Stripe side if the customer is logged in.
+		if ( is_user_logged_in() && WC()->customer instanceof WC_Customer ) {
```
daledupreez: Now that I am seeing that this only occurs for logged-in shoppers, it still feels like we need to think about how this ought to work. I am still hesitant to send customer data to Stripe before we actually have confirmed the customer will be using Stripe for the payment. Can we defer sending customer details for _all_ customers?

wjrosa: We cannot update the customer associated with a checkout session later. So, at least the ID is required here for this to work properly for logged-in users.

daledupreez: Are there any functional downsides to us only sending the customer ID? And do we send the billing, shipping, and name details as part of the payment process? I am just trying to understand when and how the data is transferred, especially later on.

wjrosa: Assuming the merchant keep adaptive pricing enabled, we won't update customer details over the Stripe API anymore (as we did for payment intents). We should add a customer update call somewhere in this feature to keep it consistent. Probably when processing the payment.

Edit: actually, this was passed to the payment intent in the original flow, not to the customer object. I don't see any downsides in this case.

> And do we send the billing, shipping, and name details as part of the payment process?

Shipping details and the name we do. Not billing, as we don't have a way to provide it to the checkout session object (it is not a valid parameter). I don't see issues with the latest changes, since AFAIK, we just use these details for debugging

daledupreez: I noticed some behaviour that may be stemming from the data that is (not) shared for various payment methods [while testing the classic checkout PR earlier today](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5102#pullrequestreview-3979905194). I think we need to make sure we are replicating billing fields into the payment data so shoppers don't need to input data twice, most likely via the `confirm()` API [docs](https://docs.stripe.com/js/react_stripe_js/checkout/confirm) - I noticed that call currently fails for guest checkout in this PR.

But it feels like we should only send shopper data to Stripe at the time that we are trying to check out, and that should be happening whether the shopper is logged in or not.

Maybe we can sync on this topic today?

wjrosa: That's a good idea. Following our chat today, I updated the implementation to collect the data when calling `confirm` in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/0367e3ac74b2629f79c44698f3cdb4cc5fece6ac

client/blocks/checkout-sessions/__tests__/hooks.test.js
```diff
@@ -140,44 +165,174 @@ describe( 'CheckoutSessions hook tests', () => {
...
+				shippingData
+			);
+			await onCheckoutSuccessResultPromise;
+
+			expect( mockConfirm ).toHaveBeenCalledWith(
+				expect.not.objectContaining( { email: expect.anything() } )
+			);
+		} );
+
+		it( 'includes phone from billing-phone DOM element', async () => {
```
daledupreez, line 334: We should also add a fallback for the case where we fall back to `shipping-phone`.

client/blocks/checkout-sessions/hooks.js
```diff
@@ -122,11 +126,54 @@ export const useCheckoutSuccessHandler = (
...
 
+					const billingAddress = billing?.billingAddress;
+					const shippingAddress = shippingData?.shippingAddress;
+
 					const { redirect } = paymentDetails;
 					const { checkout } = checkoutState;
-					const confirmResult = await checkout.confirm( {
+					const confirmArgs = {
+						billingAddress: {
+							name: `${ billingAddress.first_name } ${ billingAddress.last_name }`.trim(),
```
daledupreez: `billingAddress` may be undefined, so I think we should add `?.` accessors and fall back on `''` values.
```suggestion
							name: `${ billingAddress?.first_name ?? '' } ${ billingAddress?.last_name ?? '' }`.trim(),
```

client/blocks/checkout-sessions/hooks.js
```diff
@@ -122,11 +126,54 @@ export const useCheckoutSuccessHandler = (
...
+								country: billingAddress?.country,
+								line1: billingAddress?.address_1,
+								line2: billingAddress?.address_2,
+								state: billingAddress?.state,
+								city: billingAddress?.city,
+								postal_code: billingAddress?.postcode,
+							},
+						},
+						shippingAddress: {
+							name: `${ shippingAddress.first_name } ${ shippingAddress.last_name }`.trim(),
```
daledupreez: As above, `shippingAddress` may be undefined, so we should use safe accessors and fall back on `''` values.
```suggestion
							name: `${ shippingAddress?.first_name ?? '' } ${ shippingAddress?.last_name ?? '' }`.trim(),
```

daledupreez (changes requested):
I just tested this, and I get the following console error when trying to check out:
> IntegrationError: You cannot set the `phoneNumber` in confirm() unless phone_number_collection.enabled is set to `true`.

This seems to be an option that should be set during checkout session creation ([docs](https://docs.stripe.com/api/checkout/sessions/create#create_checkout_session-phone_number_collection)), and we should work out which cases we want to include the phone number for.

client/blocks/checkout-sessions/__tests__/hooks.test.js
wjrosa: Alright. Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/changes/72bd23fbe33e09491a0ae8e4eede000205b6fd48

wjrosa:
Thanks for the review, @daledupreez!

> This seems to be an option that should be set during checkout session creation ([docs](https://docs.stripe.com/api/checkout/sessions/create#create_checkout_session-phone_number_collection)), and we should work out which cases we want to include the phone number for.

I decided to enable the phone collection when the user is logged in (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/7509ff470cc08a44e79d7a124107148abf8753c2). WDYT?

daledupreez (changes requested):
@wjrosa, I am still getting the same IntegrationError for a logged-out user.

I think we need to have the setup logic match the submission logic.

wjrosa:
@daledupreez thanks for another round of review. I was able to reproduce the issue you mentioned and fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/f3332dd741703c75797dee08031d313dfcfdd3be

client/blocks/checkout-sessions/hooks.js
```diff
@@ -145,24 +146,67 @@ export const useCheckoutSuccessHandler = (
...
-							name: `${ shippingAddress?.first_name ?? '' } ${
-								shippingAddress?.last_name ?? ''
-							}`.trim(),
+						returnUrl: redirect,
+					};
+
+					// Only include shipping information if the min. requirement is met.
+					if (
+						shippingAddress?.address_1 &&
+						shippingAddress?.country
```
daledupreez, line 163: Nit: The code below assumes `shippingAddress?.postcode` won't be empty, or that it's OK to send an empty postal code. Is that a valid assumption? The inline comments seem to suggest we should not send any empty values.

client/blocks/checkout-sessions/hooks.js
```diff
@@ -145,24 +146,67 @@ export const useCheckoutSuccessHandler = (
...
+								shippingAddress?.recipient
+									?.split( ' ' )
+									?.slice( 0, 1 )
+									?.join( ' ' ) ?? ''
+							} ${
+								shippingAddress?.recipient
+									?.split( ' ' )
+									?.slice( 1 )
+									?.join( ' ' ) ?? ''
+							}`.trim();
```
daledupreez: Why do we need to manipulate `shippingAddress.recipient` so much? Why can't we just use `shippingAddress.recipient` directly and/or wrapped in `.trim()`? It feels like a code comment would be really helpful.

daledupreez (changes requested):
@wjrosa, I am still getting the following error:

> IntegrationError: You cannot set the `phoneNumber` in confirm() unless phone_number_collection.enabled is set to `true`.

I am testing as follows:
* Configure block checkout to require the shipping and billing phone number fields
* Enable Optimized Checkout and Adaptive Pricing 
* Incognito window with logged-out user
* Add product to checkout
* Navigate to block checkout
* Try to check out using Stripe card with Adaptive Pricing
* See error when trying to check out

Also, it looks like there are now PHP lint issues as well as JS unit test errors.

client/blocks/checkout-sessions/hooks.js
wjrosa: I actually forgot to remove this key here. I am already adding it only when it is set below. I have updated the implementation in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/0534653891a50bfa21dc40a66b2ea9bd9670d22b. It is optional, but cannot be empty

client/blocks/checkout-sessions/hooks.js
wjrosa: I copied some logic from ECE, but it was redundant here. So I simplified it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/02c7ac305bd2105ba9d06adf92b01b9d6751a373

wjrosa:
> @wjrosa, I am still getting the following error: (...)

Thanks for the review, @daledupreez ! I was able to reproduce the bug and fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5070/commits/d595b93f6ad78d176f360b53d440227998a7acb1

malithsen (approved):
Looks good to me

PR #5072, opened 2026-02-27
Clear cache after adding a new payment method

### LINKED ISSUE #3928: Newly added payment methods do not appear and are not set as default on My Account > Payment Methods page
**Describe the bug**
When adding a new payment method via My Account > Payment Methods using the WooCommerce Stripe Gateway (v9.2.0), the newly added method does not appear in the list. Only the first saved payment method is displayed, even though the new method is successfully saved in Stripe.

Additionally, the newly added payment method is not set as the default, even though it is the most recent method saved. This leads to confusion for customers, as they expect the latest payment method to be set as the default.

**To Reproduce**
1. Navigate to My Account > Payment Methods in WooCommerce.
2. Click on Add Payment Method and add a new credit card.
3. Complete the process and return to My Account > Payment Methods.
4. Notice that the newly added payment method does not appear in the list.
5. The previously saved method remains as the only visible payment method.
6. The new method is not set as the default, and the old method is still marked as default.
7. Check the Stripe Dashboard and see that the new method is saved but not shown or set as default in WooCommerce.

**Expected behavior**
When a new payment method is added via My Account > Payment Methods, it should immediately appear in the list.
The newly added method should be automatically set as the default payment method if it's checked as default.
All saved payment methods should be listed, allowing users to manage, select, or delete any of them.

 - WordPress Version : 6.7.2
 - WooCommerce Version : 9.6.2
 - Stripe Plugin Version : 9.2.0
 - Browser [e.g. chrome, safari] and Version : All

Mayisha:
Fixes STRIPE-285
Fixes #3928

Added `cache_clear` in two places which clears the list of payment methods cache retrieved from Stripe.
- After creating a new payment method: this will prevent a possible case when we have the new token in WC token but we are using a stale cache of Stripe payment method list and the token gets deleted [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/payment-tokens/class-wc-stripe-payment-tokens.php#L369-L372).

## Testing instructions
- The issue is sporadic and not easily reproducible. Verify if the code changes and the reason behind it make sense.\
- Regression test.
    - As a shopper, login to the store.
    - Go to My Account > Payment Methods and add a new card.
    - Verify that the first card is set default upon addition. 
    - Check your logs and there should a GET payment_methods request logged.
    - Add a new card. Verify that the card get added.
    - Check your logs and there should a GET payment_methods request logged.
    - Change the default payment method.
    - Check your logs and there should a GET payment_methods request logged.
    - Refresh the page a few times. Go to the checkout page and place orders.
    - Check your logs and verify that there are no unnecessary GET payment_methods requests logged meaning we are using data from the cache when applicable.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


malithsen (approved):
I wasn't able to replicate the issue, but the change seems okay to me.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3103,6 +3103,12 @@ public function add_payment_method() {
 				throw new WC_Stripe_Exception( sprintf( 'New payment token is not an instance of WC_Payment_Token. Token: %s.', print_r( $token, true ) ) );
 			}
 
+			// Clear the cache after saving the token so the payment-methods listing page always
+			// fetches a fresh list from Stripe. Without this, a stale cache populated after
+			// WC_Payment_Token::save() may cause the stale-token cleanup in woocommerce_get_customer_upe_payment_tokens()
+			// to delete the newly created token from the payment methods list before the user sees it.
+			$customer->clear_cache();
```
daledupreez, line 3175: Should this run before or after the action below? I am concerned that developers may have expectations about the state of the data, and our cache invalidation may change that state in ways that developers won't expect.

includes/class-wc-stripe-customer.php
```diff
@@ -960,8 +962,10 @@ public function set_default_source( $source_id ) {
 			'POST'
 		);
 
+		// Clear cache so that the payment methods list from Stripe is refreshed to have the correct default payment method.
+		$this->clear_cache();
```
daledupreez: Nit: why do we want to run this in the case where setting the default returns an error? It feels like the comments should mention why we are clearing the cache regardless of the result.

includes/class-wc-stripe-customer.php
```diff
@@ -986,8 +992,10 @@ public function set_default_payment_method( $payment_method_id ) {
 			'POST'
 		);
 
+		// Clear cache so that the payment methods list from Stripe is refreshed to have the correct default payment method.
```
daledupreez: As above, I am not sure why this should run even when the API call returns an error.

includes/class-wc-stripe-customer.php
Mayisha: You are correct, actually. It is unnecessary to clear cache if the request fails and the default method in Stripe is still same as the one in the cache.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: For the original issue that we are possibly fixing, I think it does not matter whether we clear the cache before or after this action.

About the expected data in the action callback, I believe ideally the callback is supposed to be implemented in a way that handles empty cache and transients of the relevant data 🤔

Mayisha: Let me know if you think moving it after the action sounds safer to you.

daledupreez: I took a closer look at the surrounding code and access patterns, and I think you're right that the order doesn't matter much.

daledupreez (approved):
The changes here LGTM. I do think it would be worth updating the PR description and title to match the changes, as I think they have drifted slightly.

Mayisha:
Merging as BECS test failures are unrelated.

PR #5074, opened 2026-02-27
[FridAI] Add WordPress Playground preview for PRs

wjrosa:
Fixes _N/A — FridAI exploratory initiative; no upstream Linear/GitHub issue._

Adds a one-click WordPress Playground preview for every non-draft PR so reviewers can try the plugin in the browser without spinning up a local environment.

- New workflow `.github/workflows/pr-build-live-branch.yml`: builds the plugin on every non-draft, non-docs-only PR, uploads the zip as an artifact, and posts (or updates) a sticky PR comment with a Playground link.
- New blueprint generator `.github/workflows/scripts/generate-playground-blueprint.js`: emits the Playground steps that install WordPress + WooCommerce + this plugin (from the CI artifact) + the [Stripe Dev Tools extension](https://github.com/woocommerce/woocommerce-gateway-stripe-dev-tools). The generated blueprint proxies the plugin zip through the Playground plugin-proxy service to dodge CORS and lands the reviewer on the Stripe Dev Tools settings screen (`/wp-admin/admin.php?page=wc-stripe-dev`) so test API keys can be entered immediately.
- Modeled after [`woocommerce/woocommerce`'s `pr-build-live-branch.yml`](https://github.com/woocommerce/woocommerce/blob/trunk/.github/workflows/pr-build-live-branch.yml), with the Stripe-specific adjustments above.

**Known limitation (addressed by follow-up #5368):** Playground's CORS proxy strips `Authorization` headers by default, so live calls to `api.stripe.com` from inside the sandbox return 401 and the plugin surfaces "test API keys are no longer valid". The sticky-comment scope copy reflects this: this PR establishes the preview workflow and is useful for admin UI / settings shape / install-flow review even with that caveat. The Stripe-API pass-through fix lands separately in #5368 to keep the review surfaces small.

**How can this code break?**
- The build job uses a dedicated `STRIPE_DEV_TOOLS_TOKEN` to check out the private `woocommerce-gateway-stripe-dev-tools` repo. If that secret is missing or expired, the dev tools step is designed to degrade gracefully — the plugin still ships, just without the dev tools plugin pre-installed.
- Playground itself is a browser sandbox; nothing in the workflow mutates real infrastructure. Failures are scoped to the PR comment and the CI artifact.
- The comment is posted under a sticky marker so repeated pushes update the same comment rather than spamming the thread.

- The workflow is skipped for draft PRs and docs-only changes, so routine PRs don't pay the build cost.
- Changes to the blueprint (dev tools token strategy, dev tools download path, blueprint format, plugin-proxy URL, PHP version) were iterated directly on this branch — the commit history shows each adjustment in isolation so reviewers can bisect if anything regresses.

## Testing instructions

1. Open a non-draft test PR with any code change (e.g., a comment-only edit). Confirm the `Build Live Branch` workflow runs and passes.
2. Check the PR comments: the Playground sticky comment should appear with a `Test this PR in WordPress Playground` link. Pushing a new commit should **update** that comment, not add a second one.
3. Click the Playground link and verify the sandbox boots with:
   - WordPress + WooCommerce installed and activated,
   - this plugin (from the CI artifact) installed and activated,
   - the Stripe Dev Tools extension installed and activated,
   - the admin lands on the Stripe Dev Tools settings page (`/wp-admin/admin.php?page=wc-stripe-dev`) where you can paste test/live API keys and exercise the admin UI.
4. Open a **draft** PR or a docs-only change — confirm the workflow is skipped (so reviewers aren't waiting on a live-branch build that won't help them).
5. (Optional, maintainer-only) Temporarily invalidate `STRIPE_DEV_TOOLS_TOKEN` on a fork and rerun — confirm the dev tools step is treated as optional and the build still produces a usable Playground (just without the dev tools plugin).

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️) — CI-only workflow; validated by triggering on a real PR as described in the testing steps.
-   [x] Tested on mobile (or does not apply) — CI/tooling change; no user-facing UI.

### Changelog entry

Added in `changelog.txt` and `readme.txt` under `10.7.0`:

> Dev - Add WordPress Playground preview workflow for PRs to enable one-click browser testing

<details>

#### Comment

N/A — changelog entry included above.

</details>

**Post merge**


.github/workflows/scripts/generate-playground-blueprint.js
```diff
@@ -0,0 +1,97 @@
+const generatePlaygroundBlueprint = ( runId, prNumber ) => {
+	return {
+		landingPage:
+			'/wp-admin/admin.php?page=wc-settings&tab=checkout&section=stripe',
+
+		preferredVersions: {
+			php: '8.0',
```
daledupreez: Why PHP 8.0? 8.1 reached end-of-life at the end of 2025 and 8.2 is only receiving security fixes until the end of 2026. Given that, I think it would be better to aim for 8.3 or even 8.4. ([PHP Supported Versions](https://www.php.net/supported-versions.php))

.github/workflows/scripts/generate-playground-blueprint.js
```diff
@@ -0,0 +1,97 @@
...
+				},
+				options: {
+					activate: true,
+				},
+			},
+			{
+				step: 'installPlugin',
+				pluginData: {
+					resource: 'url',
+					url: `https://playground.wordpress.net/plugin-proxy.php?org=woocommerce&repo=woocommerce-gateway-stripe&workflow=Build%20Live%20Branch&artifact=plugins-${ runId }&pr=${ prNumber }`,
```
daledupreez: Can we add an inline comment for this step? It looks like there is a _lot_ going on in this URL, and I am not sure how it all hangs together.

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,87 @@
...
+            - name: Setup PHP and Node
+              uses: ./.github/actions/setup-env
+
+            - name: Build plugin zip
+              id: build
+              env:
+                  CURRENT_VERSION: ${{ steps.version.outputs.version }}
+              run: |
+                  # Append run ID and short SHA to make each build version unique and sortable.
+                  VERSUFFIX="${GITHUB_RUN_ID}-g$(git rev-parse --short HEAD)"
```
daledupreez, line 58: Why the `-g<hash>` format? I am not sure what the `g` is for.

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,87 @@
+name: Build Live Branch
```
daledupreez: Might it be more accurate to call this something like `Build plugin and Playground artifacts`? We're doing more than "just" building a specific branch, so it feels like the naming should reflect that.
```suggestion
name: Build plugin and Playground artifacts
```

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,87 @@
...
+        paths-ignore:
+            - 'docs/**'
+            - '**/changelog/**'
+            - '**/changelog.txt'
+            - '**/readme.txt'
+            - '.gitignore'
+            - '**/*.md'
+            - '.github/**'
+            - '!.github/workflows/pr-build-live-branch.yml'
+            - '**/tests/**'
```
daledupreez, line 18: I think we can also add the following paths to the ignore list:
```suggestion
            - '**/tests/**'
            - 'docker/**'
            - '*.md'
            - 'phpstan-stubs/**'
            - 'phpstan*.*'
            - 'phpunit.xml.dist'
```

daledupreez (commented):
I have some comments from inspection, but I hit a nasty blocker while testing the e2e flow: when I try to connect a test Stripe account from within the playground UI, I get multiple CORS errors in the console and the Stripe Connect UI won't load due to a (very legitimate!) Content Security Policy that only allows framing from Stripe-hosted domains.

The console errors are as follows:
```
Access to fetch at 'https://api.woocommerce.com/stripe-sandbox/oauth-init' from origin 'https://playground.wordpress.net' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
Access to fetch at 'https://api.woocommerce.com/stripe-sandbox/oauth-init' from origin 'https://playground.wordpress.net' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
Framing 'https://connect.stripe.com/' violates the following Content Security Policy directive: "frame-ancestors 'self' https://edge-billing.stripe.com https://edge-connect.stripe.com https://edge-dashboard-admin.stripe.com https://edge-dashboard.stripe.com https://edge-docs.stripe.com https://edge-marketplace.stripe.com https://edge-support.stripe.com https://billing.stripe.com https://connect.stripe.com https://dashboard-admin.stripe.com https://dashboard.stripe.com https://docs.stripe.com https://edge-support-conversations.stripe.com https://edge.stripe.com https://marketplace.stripe.com https://stripe.com https://support-admin.corp.stripe.com https://support-conversations.stripe.com https://support.stripe.com". The request has been blocked.
```

.github/workflows/scripts/generate-playground-blueprint.js
wjrosa: Makes sense. Updated it to 8.4. in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5074/commits/63be1ce19615517ffbc8cb22beab85153ceb273f

.github/workflows/pr-build-live-branch.yml
wjrosa: That's just hardcoded as part of the suffix, generated by Claude Code. It is probably named after "Git". I can remove it if you think it is better

.github/workflows/scripts/generate-playground-blueprint.js
wjrosa: AI suggested the following comment:
```
The plugin-proxy.php script serves the latest artifact from the "Build Live Branch" workflow for the PR. The artifact is expected to be named `plugins-{runId}.zip`.
```
WDYT?

daledupreez: How about the following?
```suggestion
					/* The plugin proxy helper fetches the artifact produced by the specified workflow in the target GitHub repo. 
					In this case, it fetches the `plugins-<runId>-<pr>` artifact from the Build Live Branch workflow in the woocommerce/woocommerce-gateway-stripe repo. */
					url: `https://playground.wordpress.net/plugin-proxy.php?org=woocommerce&repo=woocommerce-gateway-stripe&workflow=Build%20Live%20Branch&artifact=plugins-${ runId }&pr=${ prNumber }`,
```

wjrosa: Looks better. Thanks!

wjrosa:
Not sure how we can fix the CORS issue. We can update the Connect server repository to include playground URLs in the whitelist, but that may be a change too specific for this need. Another idea would be to hardcode the same Stripe test account, to keep it always connected by default. But I am not sure how to make it safe enough. Do you have any other ideas, Dale?

daledupreez:
@wjrosa, might we be able to work around some of the restrictions by launching the connect flow in a new tab/window?

wjrosa:
Good idea, Dale. I think it might work. Claude is working on this in the background

wjrosa:
Kinda works, PR is here https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5099 . I am checking the post connection flow, which is showing invalid keys error

wjrosa:
The first request still fails (request to `oauth-init`):
```
Access to fetch at 'https://api.woocommerce.com/stripe-sandbox/oauth-init' from origin 'https://playground.wordpress.net' has been blocked by CORS policy: Response to
   preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

So, I think the only way to move forward with this approach would be to add the playground domain to the WC API.

malithsen (commented):
Mostly leaving a comment to avoid daily automated review pings in Slack 😅 
I will revisit this PR once the feedback on the upstream dependency PR for the Connect server is addressed.

wjrosa:
As a way to unblock this PR, I decided to pursue another approach for now: install the dev tools and open the settings page so we can input the keys manually. The token to download and install the dev tools plugin is still waiting for approval, though.

wjrosa:
Token was approved. The dev tools extension is now being installed and we can set the keys manually 👍

wjrosa:
@copilot resolve the merge conflicts in this pull request

daledupreez:
@wjrosa, could you document how you're copying in the new keys? I have tried multiple ways of updating the settings, but my new keys are not getting picked up or recognised, as I keep getting 401 errors when calling Stripe APIs despite my pasted keys being specified in the debug logs.

As a side note, I _am_ able to trigger the OAuth flow correctly if I open up the playground URL directly in a new tab (i.e. instead of the default iframe that is shown in the initiating window), but the generated keys don't get saved for some reason. I can get to this state by opening wp-admin links in a new tab/window.

There are also some secondary issues that I tripped over:
 * The WooCommerce setup wizard was shown instead of the Stripe dev tools when first loading the site -- I skipped the wizard, and was then taken to the WooCommerce task list
   - We may be able to skip this by using the BluePrints `setSiteOptions` step ([docs](https://wordpress.github.io/wordpress-playground/blueprints/steps#setSiteOptions)) to configure some basic store options and try to bypass the WooCommerce wizard
 * The Stripe dev tools shortcuts are not enabled for the `playground.wordpress.net` domain
 * Clicking on the Edit/Save/Cancel links in the Gateway Settings tab of the dev tools opens a blank new window (at least when accessing the URL directly and not via an iframe)

I have created https://github.com/woocommerce/woocommerce-gateway-stripe-dev-tools/pull/5 to address the two issues with the dev tools.

wjrosa:
@daledupreez 

> @wjrosa, could you document how you're copying in the new keys? I have tried multiple ways of updating the settings, but my new keys are not getting picked up or recognised, as I keep getting 401 errors when calling Stripe APIs despite my pasted keys being specified in the debug logs.

I am opening the WP Playground URL, then in wp-admin, I am going to Stripe _> Dev Tools_> Gateway Settings _> edit and either adding just the keys manually or replacing the whole configuration with my local test environment settings.

Currently, payment methods aren't loading for me on the Stripe settings page (not sure why yet). But I can at least view the other settings and render the Stripe container in the checkout page.

malithsen:
When I copy local keys to the playground instance, I can visit the settings page but the account retrieval doesn't seem to work. I see the following error

```
We couldn't connect to your account, it seems like the test API keys we've saved for you are no longer valid. Please use the Configure connection button below to reconnect.
```

It doesn't seem like requests are reaching Stripe either, I don't see them on Stripe dashboard logs 🤔

.github/workflows/pr-build-live-branch.yml
```diff
@@ -0,0 +1,111 @@
...
+            - ready_for_review
+        paths-ignore:
+            - 'docs/**'
+            - '**/changelog/**'
+            - '**/changelog.txt'
+            - '**/readme.txt'
+            - '.gitignore'
+            - '**/*.md'
+            - '.github/**'
+            - '!.github/workflows/pr-build-live-branch.yml'
```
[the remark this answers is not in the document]
wjrosa: Good catch — fixed in 37e27eb. The `.github/**` ignore was indeed swallowing the script directory; only the workflow file itself was being un-ignored. Added `!.github/workflows/scripts/generate-playground-blueprint.js` so blueprint-only edits rebuild the Playground artifact and don't require a no-op change elsewhere to iterate.

wjrosa:
Dug into @malithsen's report above ([#5074 (comment)](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5074#issuecomment-4309094827)) — same symptom on my side (account retrieval fails, payment methods list empty, no requests in Stripe dashboard logs). Posting findings here so the context is in one place.

### Root cause

Playground runs PHP in a browser WASM sandbox. PHP outbound HTTP (`wp_safe_remote_post`, cURL) is tunneled to the browser's `fetch()` API, which means it's gated by browser CORS. `api.stripe.com` does not return CORS headers permitting `playground.wordpress.net`, so the direct fetch fails.

Since [WP/wp-playground#2076](https://github.com/WordPress/wordpress-playground/pull/2076) (merged 2024-12-13), Playground retries failed fetches through its CORS proxy at `cors.wordpress.net/proxy.php`. **But that proxy strips `Authorization` headers in both directions by default** — see [`packages/playground/php-cors-proxy`](https://github.com/WordPress/wordpress-playground/tree/trunk/packages/playground/php-cors-proxy) and its README:

> authentication headers are blocked in both directions … refuses to pass authentication credentials.

So when the plugin sends `Authorization: Bearer sk_test_…`, the proxy drops the header, Stripe receives an unauthenticated request, returns 401 (or rejects pre-auth without logging it on the dashboard), and the plugin surfaces this as *"test API keys we've saved for you are no longer valid"*. Payment-method listing then comes back empty because `WC_Stripe_UPE_Payment_Gateway` gates method availability on `accounts/retrieve` capabilities, and that call's just been 401'd.

### Fix

The proxy supports a per-request opt-in: setting `X-Cors-Proxy-Allowed-Request-Headers: Authorization, …` on the outbound request lets the listed headers through. I've put up a stacked draft PR — #5368 — that adds a Playground-only mu-plugin (written into the sandbox via a blueprint `writeFile` step, **not** shipped with the released plugin) hooking `http_request_args` for `api.stripe.com` URLs and setting that header. Sticky-comment scope copy is also updated: account connect, method listing, and checkout-against-Stripe should work after the fix; webhooks still won't (no inbound HTTP into the sandbox).

### Ask

For reviewers: I'd appreciate landing this PR (#5074) first as-is, then iterating on the CORS pass-through in #5368. The two changes solve different problems — this PR establishes the preview workflow at all (which still has value for UI/install-flow review even with the Stripe API caveat documented in the sticky comment), and #5368 makes the preview useful for account-bound flows on top of that. Splitting them keeps the review surface small and lets the proxy fix iterate without blocking the workflow rollout.

@malithsen — once CI on #5368 posts a fresh sticky Playground link, would you mind retesting with your test keys? Specifically:

1. Account connection should succeed instead of returning the "keys no longer valid" message.
2. Payment methods should populate on the Stripe settings screen.
3. In browser devtools Network tab, confirm `api.stripe.com` calls go through `cors.wordpress.net/proxy.php` with the `X-Cors-Proxy-Allowed-Request-Headers` request header set, and return 200.

If something still doesn't work end-to-end (especially around POST body shape — the proxy has a ~100KB POST limit, which is plenty for Stripe but worth eyeballing), I'd rather find that out on the stacked PR than after merge.

wjrosa:
Closing in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5376

PR #5078, opened 2026-03-02
Only load Stripe JS on product and cart pages when express checkout is disabled

### LINKED ISSUE #4939: Many scripts are enqueued when we only want Stripe JS enqueued
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4855

### Description

I am splitting this issue out from https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4855 so we can track it more effectively. That original issue reported two closely related problems:
1. We were incorrectly enqueueing express checkout scripts on the product page
2. We were incorrectly enqueueing a number of additional scripts on various other pages

We addressed the bug for (1) in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4893.

However, we are still enqueueing the Stripe JS _and_ various other scripts across the site. [This comment](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4855#issuecomment-3749766142) from #4855 helps to outline the expectations:

> From the original implementation and the related report, [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2065#issuecomment-960527549) has some more context. This implmentation is intentional and this is the expected bahevior to load the Stripe js files on all pages as "Stripe recommends we load the Stripe JS on as many pages as possible to help combat fraud". Sites need to intentionally use the filter and return false if you do not want the scripts to be loaded on product page.

Our core issue at the moment is that we are loading _all_ the scripts, even in the case where we only want the Stripe JS scripts loaded due to express checkout not being enabled. This issue stems from the suggestion in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4855#issuecomment-3749942138) - let's make it possible to enqueue _only_ the Stripe JS files when that's all we need. At present, all of the following unnecessary files are being requested:
* `woocommerce-gateway-stripe/build/upe-blocks.css` - 1.9 kB (1,602 bytes gzipped)
* `woocommerce-gateway-stripe/build/upe-classic.css` - 0.6 kB (227 bytes gzipped)
* `woocommerce-gateway-stripe/assets/css/stripe-link.css` - 0.5 kB (206 bytes gzipped)
* `woocommerce/assets/js/frontend/country-select.min.js` - 1.5 kB (1,130 bytes gzipped)
* `woocommerce/assets/js/frontend/address-i18n.min.js` - 1.3 kB (983 bytes gzipped)
* `woocommerce/assets/js/frontend/checkout.min.js` - 5.4 kB (5,019 bytes gzipped)
* `woocommerce-gateway-stripe/build/upe-classic.js` - 44.2  kB(43,820 bytes gzipped)

That's up to 7 unnecessary HTTP requests and over 50 kB of unnecessary Javascript parsing.

### Steps to reproduce

1. Set up a store with the Stripe plugin active and connected
2. Disable the Stripe plugin
3. Open a browser to your site
4. Open the Network tab in your developer tools
5. Clear the network request history
6. Navigate to a product page and wait for the page to be fully loaded
7. Use the network tab to download a HAR export of all the files requested during the page load
8. Re-enable Stripe, and ensure that all express checkout payment methods are disabled
9. Clear your network request history
10. Reload the product page and wait for the page to be fully loaded
11. Use the network tab to download a HAR export of all the files requested during the page load
12. For each file, use `jq` to extract only the requested URLs: `cat product_requests_with_stripe.har | jq '.log.entries[].request |  .url' > product_requests_with_stripe_urls.txt`
13. Diff the two files, and confirm the additional files are being requested when Stripe is enabled

### Expected behavior

When express checkout payment methods are disabled, we should only enqueue external Stripe resources, and not any resources from the plugin.

zhongruige:
Fixes #4939

From the original issue: When express checkout (ECE) is disabled, product and cart pages load the full Stripe plugin bundle — up to 7 unnecessary HTTP requests and ~50 kB of extra JS/CSS (see #4939 for the full list). The only thing needed on these pages is `stripe.js` itself, which Stripe recommends loading broadly for fraud detection.

This PR adds an early return in `payment_scripts()` for both the UPE gateway and the abstract gateway. When visiting a product or cart page where ECE is not enabled for that location, only `stripe.js` is enqueued and all plugin-specific assets (CSS, JS bundles, localized script data) are skipped.

One callout: to support this, `WC_Stripe_Helper::should_load_scripts_for_ece_location()` is changed from private to public. No logic changes were made to that method.

## Testing instructions

1. Set up a store with the Stripe plugin active and connected.
2. Go to `WooCommerce > Settings > Payments > Stripe > Settings` and disable Express Checkout.
3. Open a product page with your browser's Network tab open.
4. Confirm that `stripe.js` is loaded but the following plugin assets are not present:
  - `upe-blocks.css`
  - `upe-classic.css`
  - `stripe-link.css`
  - `upe-classic.js`
  - `checkout.min.js`
  - `country-select.min.js`
  - `address-i18n.min.js`
5. Navigate to the cart page and confirm the same behavior (only stripe.js, no plugin assets).
6. Go back to Stripe settings and re-enable Express Checkout with Product and Cart locations checked.
7. Revisit the product page and cart page — confirm the full bundle is now loaded (the assets from step 4 should appear).
8. Navigate to the checkout page and confirm it loads the full bundle regardless of ECE settings.

---

-   [x] Covered with tests

### Changelog entry                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
<details>                                                                                                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                                                                                                    
#### Comment

Fix - Only load Stripe JS (not the full UPE bundle) on product and cart pages when express checkout is disabled                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                                                                                                                    
</details>  

**Post merge**


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2002,6 +2002,17 @@ public function payment_scripts() {
 			return;
 		}
 
+		// On product/cart pages where express checkout isn't enabled,
+		// only enqueue Stripe JS — skip the full bundle.
+		if (
+			( is_product() && ! WC_Stripe_Helper::should_load_scripts_for_ece_location( 'product' ) )
+			|| ( is_cart() && ! WC_Stripe_Helper::should_load_scripts_for_ece_location( 'cart' ) )
```
daledupreez: There are some edge cases that we handle via `WC_Stripe_Express_Checkout_Helper->is_product()`and `WC_Stripe_Express_Checkout_Helper->is_cart()`, so it may be worth relying on those helper methods.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
daledupreez: We are currently registering the `stripe` script in both this `if` block and the code below. Rather than having duplication, can we move the `wp_register_script()` call above this `if` block, and then remove the duplicate below?

Or, maybe we can simply register and enqueue the script, and then use the same `if` block and conditions to trigger an early return? It feels like that matches the intent: always enqueue the Stripe JS, but then make a decision about whether to enqueue the other resources.
```suggestion
		wp_register_script( 'stripe', 'https://js.stripe.com/clover/stripe.js', [], null, true );

		wp_enqueue_script( 'stripe' );
		
		$express_checkout_helper = new WC_Stripe_Express_Checkout_Helper();

		// On product/cart pages where express checkout isn't enabled,
		// only enqueue Stripe JS — skip all other resources.
		if (
			( $express_checkout_helper->is_product() && ! WC_Stripe_Helper::should_load_scripts_for_ece_location( 'product' ) )
			|| ( $express_checkout_helper->is_cart() && ! WC_Stripe_Helper::should_load_scripts_for_ece_location( 'cart' ) )
		) {
			return;
		}

	$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';

	wp_register_style( 'stripe_styles', plugins_url( 'assets/css/stripe-styles.css', WC_STRIPE_MAIN_FILE ), [], WC_STRIPE_VERSION );
	wp_enqueue_style( 'stripe_styles' );

```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -403,6 +403,17 @@ public function payment_scripts() {
...
+		// On product/cart pages where express checkout isn't enabled,
+		// only enqueue Stripe JS — skip the full UPE bundle.
+		if (
+			( is_product() && ! WC_Stripe_Helper::should_load_scripts_for_ece_location( 'product' ) )
+			|| ( is_cart() && ! WC_Stripe_Helper::should_load_scripts_for_ece_location( 'cart' ) )
+		) {
+			wp_register_script( 'stripe', 'https://js.stripe.com/clover/stripe.js', [], null, true );
+			wp_enqueue_script( 'stripe' );
+			return;
+		}
```
daledupreez: We should check whether this logic is redundant, as we're calling different helpers further up to return early for cart and product contexts.

includes/class-wc-stripe-helper.php
```diff
@@ -1232,7 +1232,7 @@ public static function should_load_scripts_on_cart_page() {
 	 * @param string $location  Either 'product' or 'cart'. Used to specify which location to check.
 	 * @return boolean True if Stripe's JS should be loaded for the provided location, false otherwise.
 	 */
-	private static function should_load_scripts_for_ece_location( $location ) {
+	public static function should_load_scripts_for_ece_location( $location ) {
```
daledupreez: I think we may need to switch the core of this logic over to the `should_show_ece_on_product_pages()` and `should_show_ece_on_cart_page()` helpers in `WC_Stripe_Express_Checkout_Helper`. One major driver is that this code is not up to date with the addition of Amazon Pay as an express payment method, with the configuration stored in a separate option.

wjrosa:
Broken unit tests are legit, so we need to address those before merging the PR

wjrosa (changes requested):
I agree with Dale's comment. Also, since the new blocks of code are very similar, I would consider adding a helper method that does that and call it where it is needed

includes/class-wc-stripe-helper.php
zhongruige: Thanks for the context there! Will look into it!

zhongruige: Added in these updates here: [6adc75e](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5078/commits/6adc75ed8a9ba831736ffbd78d43cddb8fc1b882)

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
zhongruige: Thanks for flagging, checked this and from my read they're not redundant. The `should_load_scripts_on_product_page()` and `should_load_scripts_on_cart_page()` are merchant-facing kill switches (via filter, default true) that control whether any Stripe scripts load. The ECE check decides what to load: full bundle vs. stripe.js only. So, they're complementary to each other. Happy to make any adjustments, though!

includes/abstracts/abstract-wc-stripe-payment-gateway.php
zhongruige: Good call — updated `should_skip_full_payment_scripts()` to use the helper's methods, which also cover [product_page] shortcode embeds and woocommerce/cart block pages.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
zhongruige: Great suggestion — moved stripe.js registration and enqueue above the check in both methods.

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
```diff
@@ -3552,4 +3553,132 @@ public function provide_test_get_excluded_payment_method_types(): array {
...
+			],
+			'Cart page with ECE on at cart'            => [
+				'page_type'                                => 'cart',
+				'express_checkout'                         => 'yes',
+				'express_checkout_button_locations'         => [ 'cart' ],
+				'upe_checkout_experience_accepted_payments' => [ WC_Stripe_Payment_Methods::CARD ],
+				'amazon_pay_button_locations'               => [],
+				'expected_stripe'                          => true,
+				'expected_upe_classic'                     => true,
+			],
```
daledupreez, line 3590: Nit: We are missing a test case for the cart page with ECE off and Amazon Pay enabled on the cart page.
```suggestion
			'Cart page with ECE off, no Amazon Pay'    => [
				'page_type'                                => 'cart',
				'express_checkout'                         => 'no',
				'express_checkout_button_locations'         => [],
				'upe_checkout_experience_accepted_payments' => [ WC_Stripe_Payment_Methods::CARD ],
				'amazon_pay_button_locations'               => [],
				'expected_stripe'                          => true,
				'expected_upe_classic'                     => false,
			],
			'Cart page with ECE on at cart'            => [
				'page_type'                                => 'cart',
				'express_checkout'                         => 'yes',
				'express_checkout_button_locations'         => [ 'cart' ],
				'upe_checkout_experience_accepted_payments' => [ WC_Stripe_Payment_Methods::CARD ],
				'amazon_pay_button_locations'               => [],
				'expected_stripe'                          => true,
				'expected_upe_classic'                     => true,
			],
			'Cart page with ECE off, Amazon Pay on at cart'    => [
				'page_type'                                => 'cart',
				'express_checkout'                         => 'no',
				'express_checkout_button_locations'         => [],
				'upe_checkout_experience_accepted_payments' => [ WC_Stripe_Payment_Methods::CARD ],
				'amazon_pay_button_locations'               => [ 'cart' ],
				'expected_stripe'                          => true,
				'expected_upe_classic'                     => true,
			],
```

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
```diff
@@ -3552,4 +3553,132 @@ public function provide_test_get_excluded_payment_method_types(): array {
...
+			'Product page with ECE off, Amazon Pay on at product' => [
+				'page_type'                                => 'product',
+				'express_checkout'                         => 'no',
+				'express_checkout_button_locations'         => [],
+				'upe_checkout_experience_accepted_payments' => [ WC_Stripe_Payment_Methods::CARD, WC_Stripe_Payment_Methods::AMAZON_PAY ],
+				'amazon_pay_button_locations'               => [ 'product' ],
+				'expected_stripe'                          => true,
+				'expected_upe_classic'                     => true,
+			],
+			'Checkout page with ECE off'               => [
```
daledupreez: Nit: We should specify that Amazon Pay is also off.
```suggestion
			'Checkout page with ECE off and Amazon Pay off'               => [
```

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Gateway_Test.php
```diff
@@ -3552,4 +3553,132 @@ public function provide_test_get_excluded_payment_method_types(): array {
...
+		}
+
+		$original_settings = WC_Stripe_Helper::get_stripe_settings();
+
+		$stripe_settings                                                  = $original_settings;
+		$stripe_settings['enabled']                                       = 'yes';
+		$stripe_settings['express_checkout']                              = $express_checkout;
+		$stripe_settings['express_checkout_button_locations']             = $express_checkout_button_locations;
+		$stripe_settings['upe_checkout_experience_accepted_payments']     = $upe_checkout_experience_accepted_payments;
+		$stripe_settings['amazon_pay_button_locations']                   = $amazon_pay_button_locations;
```
daledupreez: Non-blocking nit: it looks like there are additional spaces that we can remove.
```suggestion
		$stripe_settings                                              = $original_settings;
		$stripe_settings['enabled']                                   = 'yes';
		$stripe_settings['express_checkout']                          = $express_checkout;
		$stripe_settings['express_checkout_button_locations']         = $express_checkout_button_locations;
		$stripe_settings['upe_checkout_experience_accepted_payments'] = $upe_checkout_experience_accepted_payments;
		$stripe_settings['amazon_pay_button_locations']               = $amazon_pay_button_locations;
```

daledupreez (approved):
The changes are looking good to me, and test well. I also did manual testing with the following combinations:
| Page | Express Checkout | Amazon Pay |
| --- | --- | --- |
| Product | Not active | Not active |
| Product | Active | Not active |
| Product | Not active | Active |
| Classic Cart | Not active | Not active |
| Classic Cart | Active | Not active |
| Classic Cart | Not active | Active |
| Block Cart | Not active | Not active |
| Block Cart | Active | Not active |
| Block Cart | Not active | Active |

I also tested with Amazon Pay, Apple Pay/Google Pay, and Link all disabled.

I have some minor comments around the tests, but things are looking really good.

wjrosa (approved):

zhongruige:
Thank you both for your patience and helping me to review this! 🙏

PR #5080, opened 2026-03-02
Add extension compatibility activation tests workflow

zhongruige:
   
Third-party plugins that integrate with Stripe can fatal when we remove or rename classes. QIT activation tests can catch these, but no workflow runs them today.                                                                              
                                                            
This adds a post-merge workflow (`ci-extension-compat.yml`) that runs QIT activation tests for Stripe paired with integrating extensions, with non-blocking Slack alerts on failure. It uses two complementary QIT mechanisms for testing plugin combinations:

- **`--extension_set=compatibility`** — QIT's built-in set of 16 major extensions (WC Subscriptions, WC Payments, Bookings, etc.). More details in our docs around [Extension Sets](https://qit.woo.com/docs/managed-tests/extension-sets).
- **`-p <slug>`** — Provide individual plugin slugs (WCCOM, WPORG) for specific plugins not in the extension set

Note: Changes to `run-qit.yml` are backwards-compatible — new inputs default to empty strings, and existing callers are unaffected.

## Testing instructions

This is a CI-only change with no runtime impact. More important is ensuring the GitHub action syntax looks correct.

Once this is merged, we can do the following steps to verify:

1. Trigger the workflow manually via Actions > Extension Compatibility > Run workflow
2. Verify full default matrix runs (compatibility set + individual plugins)
3. Confirm existing `ci-merge` and `ci-cron-qit` workflows are unaffected

  ---

  ### Changelog entry

  - [x] This Pull Request does not require a changelog entry. (Comment required below)

  <details>

  #### Comment

  CI workflow change only — no user-facing changes.

  </details>

### DIFF
--- a/.github/workflows/ci-extension-compat.yml
+++ b/.github/workflows/ci-extension-compat.yml
@@ -1,0 +1,112 @@
+name: Extension Compatibility
+
+on:
+  pull_request:
+    branches:
+      - develop
+    types:
+      - closed
+  workflow_dispatch:
+    inputs:
+      plugin-slug:
+        description: 'Test a specific plugin slug (leave empty to run the default matrix)'
+        type: string
+        default: ''
+      extension-set:
+        description: 'QIT extension set to test (e.g. "compatibility"). Mutually exclusive with plugin-slug.'
+        type: string
+        default: ''
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+
+  setup-matrix:
+    if: github.event_name == 'workflow_dispatch' || github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - id: set-matrix
+        env:
+          PLUGIN_SLUG: ${{ inputs['plugin-slug'] }}
+          EXTENSION_SET: ${{ inputs['extension-set'] }}
+        run: |
+          if [[ -n "$PLUGIN_SLUG" && -n "$EXTENSION_SET" ]]; then
+            echo "::error::plugin-slug and extension-set are mutually exclusive"
+            exit 1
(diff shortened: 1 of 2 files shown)


zhongruige:
Tagging you on this one @daledupreez, but let me know if there's anyone else you'd recommend (similar question also for my adjacent PR at https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5078 as well).

.github/workflows/ci-extension-compat.yml
```diff
@@ -0,0 +1,112 @@
...
+        default: ''
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+
+  setup-matrix:
+    if: github.event_name == 'workflow_dispatch' || github.event.pull_request.merged == true
```
daledupreez, line 27: Question: is there a way for us to exit the workflow early if this condition is false? I am asking because every step has this guard condition, and it feels redundant to keep checking that our expected starting condition/state is still true.

daledupreez (approved):
This looks OK to me based on inspection -- it will be much easier to see how it works after we've merged, and we can iterate from there.

daledupreez:
Also, for future workflow development, you can consider using https://github.com/nektos/act - I've had some success testing workflows using the tool.

malithsen (approved):
Thanks for adding this. This looks good to me on code inspection.

zhongruige:
Thanks for sharing! I had heart act has gotten better since I last tried it, I need to check it out to again to see how it works nowadays.

.github/workflows/ci-extension-compat.yml
zhongruige: Good shout! We can definitely simplify this down. Added in on [9918acb](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5080/commits/9918acbac411d386ba974ba372dc2ce9b44889dc)

zhongruige:
Merging--I'll be keeping an eye on this for the first couple days to check for any false-positive or noisy failures and will adjust accordingly. Otherwise, will let this run regularly!

PR #5082, opened 2026-03-02
Disable OCS for the Add payment method page

### LINKED ISSUE #5041: Optimized Checkout shown when adding a payment method
### Description

When Optimized Checkout Suite is enabled for a site, we are also using the same UX and configuration when logged-in users are trying to add a payment method to their account. Given that this use case should give existing users the greatest amount of choice based on the payment methods the store supports, using OCS in this context may be counterproductive, as some payment methods may be hidden from the user via Stripe's algorithm.

The usage also complicates control over which payment methods should be shown, because in this context we only want to show payment methods that can be use for off-session payments, and we have local controls for some payment methods (most notably Bancomat and iDEAL) that may result in our local configuration not actually allowing them to be used as off-session payments.

Also, I am not sure why, but we are showing the "Update the payment method for all of my current subscriptions" toggle twice when OCS is enabled.

### Steps to reproduce

1. Set up a store and connect the store to Stripe -- to reproduce the Bancontact and iDEAL issue, ensure your store is set up to use EUR.
2. Ensure that as many payment methods are enabled as possible. If your store allows Bancontact and/or iDEAL to be enabled, do so.
3. If you have Banconcat and/or iDEAL enabled, navigate to the Stripe settings, and ensure that at least one of the "Enable saved (iDEAL|Bancontact) payments for repeat payments" checkboxes is _unchecked_.
4. As a logged-in user navigate to _My account_ > _Payment methods_ and then click on "Add payment method"
5. See the OCS display for the Stripe payment methods, which may or may not include all the payment methods that are enabled.
6. Note that you may see Bancontact and/or iDEAL in the list despite the local setting flagging them as not re-usable.

### Expected behavior

I think we should show the older split PE UX on this screen, as we should be optimising for choice rather than  conversion, especially given that OCS may or may not show all available options.

### Screenshots

Image

### Additional context

This came up as part of [this code review](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5028#pullrequestreview-3810021728) for some closely related fixes in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5028.

wjrosa:
Fixes STRIPE-954
Fixes #5041

To avoid hiding the Stripe payment method option when adding a payment method on the "My Account" screen, we are disabling the Optimized Checkout Suite on this page.

To achieve that, I introduced a new JS param key: `shouldShowOptimizedCheckout`. This is inspired by the ECE implementation.

Both the block and the shortcode/classic checkout pages remain unchanged, still displaying the Optimized Checkout container.

## Testing instructions

- Checkout and build this branch on your test environment (`update/disable-oc-for-add-payment-method-page`)
- Connect your Stripe account
- Enable some payment methods
- Enable the Optimized Checkout Suite in settings
- Enable saving payment methods

#### Add payment method page

- As a shopper, go to "My Account" > "Add payment method"
- Confirm payment methods are rendered separately, not inside the OC container
- Confirm you can add a new payment method

#### Regression test

- Confirm the Optimized Checkout container is rendered correctly
- Go to the shortcode/classic checkout page
- Confirm the same

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.5.0 - xxxx-xx-xx =
+* Update - Disabling the Optimized Checkout Suite in "Add Payment Method" page
```
```suggestion
* Update - Disable the Optimized Checkout Suite in "Add Payment Method" page
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -523,8 +523,12 @@ public function javascript_params() {
 		// Amazon Pay feature flag.
 		$stripe_params['isAmazonPayAvailable'] = WC_Stripe_Feature_Flags::is_amazon_pay_available();
 
+		$is_add_payment_method               = is_wc_endpoint_url( 'add-payment-method' );
```
Mayisha: Optional nit: We have `is_wc_endpoint_url( 'add-payment-method' )` in some places and `is_add_payment_method_page` in some places. We can make it consistent and use only one of them.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -685,7 +688,7 @@ private function get_enabled_payment_method_config() {
 
 		// If the Optimized Checkout is enabled, we need to return just the card payment method + express methods.
 		// All payment methods are rendered inside the card container.
```
Mayisha, line 738: Nit: Missing mention about the second condition (add payment method page)

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -523,8 +523,12 @@ public function javascript_params() {
 		// Amazon Pay feature flag.
 		$stripe_params['isAmazonPayAvailable'] = WC_Stripe_Feature_Flags::is_amazon_pay_available();
 
+		$is_add_payment_method               = is_wc_endpoint_url( 'add-payment-method' );
+		$stripe_params['isAddPaymentMethod'] = $is_add_payment_method;
+
 		// Optimized Checkout feature flag + setting.
-		$stripe_params['isOCEnabled'] = $this->oc_enabled;
+		$stripe_params['isOCEnabled']                 = $this->oc_enabled;
```
Mayisha: Looks like this param is unused now.

Mayisha (approved):
Thanks for fixing the UX for the shoppers. I have a couple of nits or minor suggestions. Otherwise, the PR looks good to me and tests well. 👍 

When OCS is enabled
✅ OCS on checkout page (block & classic)
✅ Non dynamic payment elements on 'Add payment method' page.
✅ Saving a card method works.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: The URL slug is customisable, so it may be better to use `is_add_payment_method_page()`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: I still see two usages in tests in this branch:

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/dc7b094ed418563fe4994ca8501b5eb21b1f71af/client/styles/upe/__tests__/index.js#L30-L32

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/dc7b094ed418563fe4994ca8501b5eb21b1f71af/client/styles/upe/__tests__/index.js#L116

daledupreez: And if we do keep this client field for backwards compatibility, I would set the value to match the value of the `shouldShowOptimizedCheckout` field below.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -197,20 +197,21 @@ public function is_enabled() {
...
 			if ( $main_stripe_gateway instanceof WC_Stripe_UPE_Payment_Gateway && ! $main_stripe_gateway->is_saved_cards_enabled() ) {
 				return false;
 			}
 		}
 
-		// When OC is enabled, we use the OC payment container to render all the methods.
-		if ( $this->oc_enabled ) {
-			$enabled_methods     = WC_Stripe::get_instance()->get_main_stripe_gateway()->get_upe_enabled_at_checkout_payment_method_ids();
+		// When OC is enabled (and not in "Add payment method" page), we use the OC payment container to render all the methods.
+		if ( $this->oc_enabled && ! is_add_payment_method_page() ) {
```
daledupreez: What about the other two conditions? i.e. the pay for order or change payment method pages?

The change payment method page feels like it would be another case to disable OCS in particular.

daledupreez (commented):
Looking good from inspection; will test later today.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Replaced the function in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5082/commits/ec59b825e3d2f1678ed419cde4567a2f624a1b44

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5082/commits/04685f1e49baa7e77e5cf36e8786bfeac2deca3b

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Removed the setting usage from the remaining unit tests and updated the logic as suggested by Dale in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5082/commits/0bce9af72177a9ed02b098de7aaeb5b73ced35b5

includes/payment-methods/class-wc-stripe-upe-payment-method.php
wjrosa: Added a check for the other pages as well in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5082/commits/d756fffb4642768060b08aea1e27caad7157e5a3

wjrosa:
@copilot add unit tests for the new method `is_invalid_optimized_checkout_page`

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -210,7 +210,7 @@ public function is_available() {
 		}
 
 		// When OC is enabled (and not in "Add payment method" page), we use the OC payment container to render all the methods.
```
daledupreez: Nit: the comment is out of date.
```suggestion
		// When OC is enabled _and_ we are on a page where OC is permitted, we use the OC payment container to render all the methods.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -522,8 +522,10 @@ public function javascript_params() {
...
 		$stripe_params['isAmazonPayAvailable'] = WC_Stripe_Feature_Flags::is_amazon_pay_available();
 
-		// Optimized Checkout feature flag + setting.
-		$stripe_params['isOCEnabled'] = $this->oc_enabled;
+		// Optimized Checkout feature flag + setting + whether we are on any of the pages that should not show OC.
+		$should_show_optimized_checkout               = $this->oc_enabled && ! $this->is_invalid_optimized_checkout_page();
+		$stripe_params['isOCEnabled']                 = $should_show_optimized_checkout;
+		$stripe_params['shouldShowOptimizedCheckout'] = $should_show_optimized_checkout;
 
 		if ( $this->oc_enabled ) {
```
daledupreez: There is a subtle bug here. This should be checking `$should_show_optimized_checkout`, otherwise we could enable adaptive pricing and checkout sessions without enabling optimized checkout. It doesn't work very well!
```suggestion
		if ( $should_show_optimized_checkout ) {
```

daledupreez: As a minor related note, there are two snippets of Javascript that use the value of `getStripeServerData()?.isAdaptivePricingEnabled`, which will be `undefined` when we don't set the `isAdaptivePricingEnabled` key. We may want to ensure those are cast to booleans, but that can be a separate fix.

daledupreez: And I picked all of this up by going to the "special" pages where Optimized Checkout should be disabled, but with OCS and Adaptive Pricing enabled.

wjrosa: Good catch! I went ahead and updated the `isAdaptivePricingEnabled` JS param in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5082/commits/a4c9f163c10e4fe0d7a10ff333c59873edaae590

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.5.0 - xxxx-xx-xx =
+* Update - Disable the Optimized Checkout Suite in "Add Payment Method" page
```
daledupreez: Nit: The scope is slightly larger than the original changelog. We are now disabling Optimized Checkout Suite on the Pay for Order, Add Payment Method, and Change Subscription Payment Method screens.
```suggestion
* Update - Disable the Optimized Checkout Suite in the "Add Payment Method", "Pay for Order", and "Change Subscription Payment Method" screens
```

daledupreez (commented):
Overall, I think we can ship this, though I have one concern: For merchants who rely on the Pay for Order page, we're changing what is effectively a checkout screen. Given that, I think we should keep OCS enabled for that specific screen, at least for now.

daledupreez (changes requested):
Making it clearer that I think we need changes to keep OCS on the Pay for Order screen (at least for now).

changelog.txt
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5082/commits/80372cd772cbdfc3496e8082ac7a1353d329543c

wjrosa:
@daledupreez alright. Updated the main logic in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5082/commits/bcd1799b28a135f995f98c2faa8d7c7adc311e4f

daledupreez (commented):
I tripped over a subtle bug stemming from the fact that `WC_Stripe_UPE_Payment_Method_OC` uses/extends the `WC_Stripe_Subscriptions_Trait` trait and calls `$this->maybe_init_subscriptions();` from its constructor. As a result, we end up with the card element on the add payment page getting two "Update all subscription" checkboxes, one from the main gateway and one from the OC gateway.

Image

That suggests we're instantiating the `WC_Stripe_UPE_Payment_Method_OC` class somewhere when we shouldn't. Based on my testing, I think that is in this code block:

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/5117f758b40a7db66ab9a576633e136d17b61f6d/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L918-L934

When I updated the condition to `if ( $this->oc_enabled && $this->is_valid_optimized_checkout_page() ) :`, that removed the duplicate instance.

We may want stronger protections against those hooks being added in other situations, but that feels like something we should tackle in a separate PR.

wjrosa:
Good catch, Dale. Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5082/commits/0c17f08f8f5124859d4c1a5f0ca1a4a18f11d317

daledupreez (approved):
LGTM! :shipit:

PR #5086, opened 2026-03-03
Adding specific adaptive pricing / currency conversion info to the order received page

### LINKED ISSUE #5088: Checkout Sessions: Frontend work - Order review page
Ensuring the correct display of the presentment currency on the order review page and later order pages.

Figma 0fiyiUe18lGU6kGyQNmXxE-fi

wjrosa:
Fixes STRIPE-971
Fixes #5088
Figma 0fiyiUe18lGU6kGyQNmXxE-fi/Stripe-Payments?node-id=2294-16785&t=j7jlHiYsEgmVSBXw-0

Needs to be merged after the following branch:

(and contains changes from it, please ignore those)

Following the design pieces for the order received page, this PR adds three new pieces of information to it:
- The converted total along the default total, in the order summary table
- A new "Converted total" row in the order details table
- A notice informing the shopper about the conversion

#### Preview

Converted total in order summary table:
Screenshot 2026-03-03 at 18 29 26

Converted total row in order details table:
Screenshot 2026-03-03 at 18 30 02

Conversion notice:
Screenshot 2026-03-03 at 18 31 56

## Testing instructions

- Checkout and build this branch on your test environment (`add/adaptive-pricing-info-in-order-received`)
- Merge https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5058
```php
```
- Connect a US Stripe account
- As a logged-in shopper, add a product to your cart and go to the block checkout page (*shopper must be logged in for now*)
- Select your local currency and pay
- As a shopper, confirm the new information is shown in the order received page

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
@copilot add unit tests for the following new methods:
- `add_converted_currency_information`
- `add_converted_total_table_row`
- `add_currency_conversion_notice`

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -953,6 +961,83 @@ public function attach_currency_selector_element() {
...
+		if ( empty( $checkout_session->presentment_details ) ) {
+			return $formatted_total;
+		}
+
+		$amount = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount(
+			$checkout_session->presentment_details->presentment_amount,
+			$checkout_session->presentment_details->presentment_currency
+		);
+
+		return $formatted_total . ' ($' . $amount . ' ' . strtoupper( $checkout_session->presentment_details->presentment_currency ) . ')';
```
Mayisha: We are using the `$` here before the presentment amount. But the presentment currency may have different currency symbols.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -953,6 +961,83 @@ public function attach_currency_selector_element() {
...
+
+		$rate_amount = wc_format_decimal(
+			$checkout_session->presentment_details->presentment_amount / $checkout_session->amount_total,
+			wc_get_price_decimals()
+		);
+
+		echo '<p class="woocommerce-info">';
+			printf(
+				/* translators: %1$s Converted amount and currency. %2$s Exchange rate and currency. */
+				esc_html__( 'Currency Conversion: You chose to pay %1$s for this order at an exchange rate of 1 USD = %2$s.', 'woocommerce-gateway-stripe' ),
```
Mayisha: Concern about the currency. We are assuming that the store currency or account currency is USD which might not be the case for all stores.

I wanted to make sure if this is intentional in this PR. Are you planning to handle the actual store currency in a follow-up?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -953,6 +961,83 @@ public function attach_currency_selector_element() {
...
+			$checkout_session->presentment_details->presentment_amount,
+			$checkout_session->presentment_details->presentment_currency
+		);
+
+		$rate_amount = wc_format_decimal(
+			$checkout_session->presentment_details->presentment_amount / $checkout_session->amount_total,
+			wc_get_price_decimals()
+		);
+
+		echo '<p class="woocommerce-info">';
```
Mayisha: One observation here. The style of the notice looks different from the design.

Image

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3895,4 +4034,38 @@ private function validate_field( string $field_key, $field_value ) {
...
+						'order_id'            => $order->get_id(),
+						'checkout_session_id' => $checkout_session_id,
+						'error_message'       => $checkout_session->error->message,
+					]
+				);
+
+				return null;
+			}
+
+			WC_Stripe_Database_Cache::set( $cache_key, $checkout_session, HOUR_IN_SECONDS );
```
Mayisha, line 4192: In general, the implementation of `get_checkout_session_from_order` looks good to me. I think it is the right call to cache the retrieved checkout session data.

On another note, so far, we are using this method for the purpose of getting the presentment amount and currency which is fixed for a specific order. What do you think about storing the presentment data in the order meta?

Mayisha (commented):
Working well functionally. 👍  
I have a couple of comments about the hardcoded currency and about notice styles. 

Note: the `Total` row of the order details section also shows the conversion in the bracket which is not in the design and also is redundant here. However, I do not have any suggestion atm to fix this as `woocommerce_get_formatted_order_total` is used for the total and we can not detect if this total is on the order summary section or in the details table.

Image

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -285,6 +285,14 @@ public function __construct() {
 		// Attach the currency selector div to the classic checkout page.
 		add_action( 'woocommerce_review_order_before_payment', [ $this, 'attach_currency_selector_element' ] );
 
+		// Include the converted currency information in the order total on the order received page and in the My Account orders list.
+		add_filter( 'woocommerce_get_formatted_order_total', [ $this, 'add_converted_currency_information' ], 10, 2 );
+
+		// Add a table row showing the converted total in the order review on the checkout page, if the currency was converted.
```
Mayisha: Is this comment correct? I do not see the row on the checkout page. I think it should be the order received page and order details page.

wjrosa:
Per our final design for the order received page (0fiyiUe18lGU6kGyQNmXxE-fi/Stripe-Payments?node-id=2702-208227&p=f&t=X8Hxfo8hRa4PN3nu-0), I removed the "Converted Total" row from the details table in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/commits/8bc641cdb884fb6ddf27d426a7919412fb05c6f3

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Yeah, I am using the default theme style for notices instead of the proposed one

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: You're right. But this row is now removed

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Good catch! Made it dynamic in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/commits/f7389a16cb10a9e0d3923929fd522575526293c1

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: That was actually a mistake on my part. Fixed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/commits/2428c4e4d6dec138c6ecfa8c64d6ac8219d2763a

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: That's a good idea. Added the logic in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/commits/1710a9442ece179d4f5be04589c971fd3ed35c4f

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1006,6 +1011,68 @@ public function attach_currency_selector_element() {
...
+	}
+
+	/**
+	 * Shows a notice to the order received page to inform the customer about the currency conversion
+	 * when the order is paid with a different currency than the store currency.
+	 *
+	 * @return void
+	 */
+	public function add_currency_conversion_notice( WC_Order $order ): void {
+		$checkout_session = $this->get_checkout_session_from_order( $order );
```
Mayisha: @wjrosa as we are storing the data in order meta, I think we should try to get the data from the meta first instead of triggering a cache or API call to fetch the checkout session.

wjrosa: That's a very good idea. Refactored the code in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/commits/5c79f0e9d8bc4516315bc87d9793d65d84d93e6c

wjrosa:
@copilot Update PHPUnit tests to comply with my last commit

Mayisha (approved):
Changes look good to me 👍 

✅ No changes when I pay with the account currency.
✅ When paid with local currency, a notice and presentment amount+currency is displayed on the order received page and **My Account > Orders** page.

**Presentment details**
Image


**Notice**

Image

client/api/index.js
```diff
@@ -678,11 +678,27 @@ export default class WCStripeAPI {
 	/**
 	 * Creates a new checkout session.
 	 *
+	 * @param {Object} billingAddress Billing address.
```
daledupreez: Nit: Based on the code, it looks like `billingAddress` may be falsey and/or nullable. Can we reflect that in the JSDoc?
```suggestion
	 * @param {Object|null} billingAddress Billing address.
```

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -30,7 +32,7 @@ export const CheckoutContainer = ( props ) => {
 			);
 		}
 		return clientSecret;
-	}, [ api, setShouldLoadStripeElements ] );
+	}, [ api, billingAddress, setShouldLoadStripeElements ] );
```
daledupreez: How does the API work if we make repeated calls to this API? With `billingAddress` as a dependency, we may trigger multiple API calls in response to the `billingAddress` changing.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1048,6 +1054,82 @@ public function attach_currency_selector_element() {
 		echo '<div id="wc-stripe-currency-selector" class="wc-stripe-currency-selector" style="margin: 12px 0;"></div>';
 	}
 
+	/**
+	 * Adds the converted currency information to the order total on the order received page when the order is paid with a different currency than the store currency.
```
daledupreez: Nit: The PHPDoc only mentions the order received page, but I believe this also applies to the My Orders page.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1048,6 +1054,82 @@ public function attach_currency_selector_element() {
...
+			return $formatted_total;
+		}
+
+		$currency_symbol = get_woocommerce_currency_symbol( $presentment_currency );
+		$amount          = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount(
+			$presentment_amount,
+			$presentment_currency
+		);
+
+		return $formatted_total . ' (' . $currency_symbol . $amount . ' ' . strtoupper( $presentment_currency ) . ')';
```
daledupreez: Not blocking, but it feels like we may need some way to generate localised currency formatting, and it may be worth pushing that into a helper early. (That probably makes sense for a follow-up rather than this PR.)

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1048,6 +1054,82 @@ public function attach_currency_selector_element() {
...
+		}
+
+		$stripe_amount      = WC_Stripe_Helper::get_stripe_amount( $order->get_total(), $order->get_currency() );
+		$woocommerce_amount = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount(
+			$presentment_amount,
+			$presentment_currency
+		);
+		$rate_amount        = wc_format_decimal(
+			$presentment_amount / $stripe_amount,
+			wc_get_price_decimals()
```
daledupreez: I think this should always use the default number of decimals for the source currency. I tripped over a situation where no decimals were shown for this value, and that can drive incorrect rounding in cases where prices in the store currency use `0` decimals.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4030,4 +4112,69 @@ private function validate_field( string $field_key, $field_value ) {
...
+	private function get_checkout_session_from_order( WC_Order $order ): ?object {
+		$checkout_session_id = WC_Stripe_Order_Helper::get_instance()->get_stripe_checkout_session_id( $order );
+		if ( ! $checkout_session_id ) {
+			return null;
+		}
+
+		$cache_key        = 'checkout_session_' . $checkout_session_id;
+		$checkout_session = WC_Stripe_Database_Cache::get( $cache_key );
+		if ( ! $checkout_session ) {
+			$checkout_session = $this->stripe_request( 'checkout/sessions/' . $checkout_session_id, [], null, 'GET' );
```
daledupreez: Should we invalidate or overwrite this cache when we get webhooks for the checkout session? It feels dangerous to me to keep the checkout session data from immediately after the purchase cached, when we may get updates via webhooks.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -3943,4 +3943,227 @@ public function provide_test_is_valid_optimized_checkout_page() {
...
+
+		ob_start();
+		$this->mock_gateway->add_currency_conversion_notice( $order );
+		$output = ob_get_clean();
+
+		WC_Stripe_Database_Cache::delete( 'checkout_session_' . $checkout_session_id );
+
+		$expected_amount     = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount( 1500, 'eur' );
+		$stripe_order_amount = WC_Stripe_Helper::get_stripe_amount( $order->get_total(), $order->get_currency() );
+		$expected_rate       = wc_format_decimal( 1500 / $stripe_order_amount, wc_get_price_decimals() );
```
daledupreez: Is it worth hard-coding some of these values to ensure we are getting the right data and won't get tripped up by any Woo filters etc?
```suggestion
		$expected_amount = '15.00';
		$expected_rate   = '0.75';
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -308,6 +308,12 @@ public function __construct() {
 		// Attach the currency selector div to the classic checkout page.
 		add_action( 'woocommerce_review_order_before_payment', [ $this, 'attach_currency_selector_element' ] );
 
+		// Include the converted currency information in the order total on the order received page and in the My Account orders list.
+		add_filter( 'woocommerce_get_formatted_order_total', [ $this, 'add_converted_currency_information' ], 10, 2 );
+
+		// Add a notice about currency conversion when the order currency is different from the store currency on the order details page.
+		add_filter( 'woocommerce_order_details_after_order_table', [ $this, 'add_currency_conversion_notice' ], 10 );
```
daledupreez: `add_currency_conversion_notice()` returns `void`, so it should not be added via `add_filter()`. Is `'woocommerce_order_details_after_order_table'` an action? (PHPStan is correctly reporting that as an error.)

daledupreez (changes requested):
The base functionality is working as described, but I think we still need some cleanup.

client/blocks/checkout-sessions/checkout-container.js
wjrosa: I asked Claude to address all your comments/suggestions in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/dd6729dc74879eaa949d8e3c7be01055d7a90384. We are now using `useRef`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/dd6729dc74879eaa949d8e3c7be01055d7a90384.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Replaced `add_filter` with `add_action` in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/dd6729dc74879eaa949d8e3c7be01055d7a90384.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/dd6729dc74879eaa949d8e3c7be01055d7a90384#diff-9356f87eb3d1b3fbc63a5201e8ccc38668d392af16d66e94656a3ee8e101820e

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: On our current webhook handler, we are not retrieving the checkout session object directly (requesting). We just process the incoming event data. The agentic checkout flow does retrieve the checkout session, but it does not implement a cache logic (so it won't share the same data).

client/blocks/checkout-sessions/__tests__/checkout-container.test.js
```diff
@@ -40,6 +40,18 @@ describe( 'CheckoutSessionsContainer', () => {
...
+			billing_first_name: 'John',
+			billing_last_name: 'Doe',
+			billing_address_1: '123 Main St',
+			billing_address_2: '',
+			billing_city: 'Anytown',
+			billing_state: 'CA',
+			billing_postcode: '12345',
+			billing_country: 'US',
+		},
+	};
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/8e676f4ab3afd1910ba6a0fca278e08146446960

client/blocks/checkout-sessions/checkout-container.js
```diff
@@ -16,11 +16,13 @@ const stripePromise = loadStripe();
 export const CheckoutContainer = ( props ) => {
 	const {
 		api,
+		billing: { billingAddress },
 		setPaymentProcessorLoadErrorMessage,
 		setShouldLoadStripeElements,
 	} = props;
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/8e676f4ab3afd1910ba6a0fca278e08146446960

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -308,6 +308,12 @@ public function __construct() {
 		// Attach the currency selector div to the classic checkout page.
 		add_action( 'woocommerce_review_order_before_payment', [ $this, 'attach_currency_selector_element' ] );
 
+		// Include the converted currency information in the order total on the order received page and in the My Account orders list.
+		add_filter( 'woocommerce_get_formatted_order_total', [ $this, 'add_converted_currency_information' ], 10, 2 );
+
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/8e676f4ab3afd1910ba6a0fca278e08146446960

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1006,6 +1012,82 @@ public function attach_currency_selector_element() {
...
+		$stripe_amount      = WC_Stripe_Helper::get_stripe_amount( $order->get_total(), $order->get_currency() );
+		$woocommerce_amount = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount(
+			$presentment_amount,
+			$presentment_currency
+		);
+		$rate_amount        = wc_format_decimal(
+			$presentment_amount / $stripe_amount,
+			wc_get_price_decimals()
+		);
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/8e676f4ab3afd1910ba6a0fca278e08146446960

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1006,6 +1012,82 @@ public function attach_currency_selector_element() {
...
+
+		echo '<p class="woocommerce-info" style="margin-top: 1em;">';
+			printf(
+				/* translators: %1$s Converted amount and currency. %2$s Store currency. %3$s Exchange rate and currency. */
+				esc_html__( 'Currency Conversion: You chose to pay %1$s for this order at an exchange rate of 1 %2$s = %3$s.', 'woocommerce-gateway-stripe' ),
+				esc_html( $woocommerce_amount . ' ' . strtoupper( $presentment_currency ) ),
+				esc_html( get_woocommerce_currency() ),
+				esc_html( $rate_amount . ' ' . strtoupper( $presentment_currency ) )
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/changes/8e676f4ab3afd1910ba6a0fca278e08146446960

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1109,22 +1117,33 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
 			return;
 		}
 
-		$stripe_amount      = WC_Stripe_Helper::get_stripe_amount( $order->get_total(), $order->get_currency() );
+		$stripe_amount = WC_Stripe_Helper::get_stripe_amount( $order->get_total(), $order->get_currency() );
+		if ( $stripe_amount <= 0 ) {
```
daledupreez, line 1121: What would this value be for orders where a coupon brings the payment amount down to 0?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1109,22 +1117,33 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
...
 			$presentment_currency
 		);
-		$rate_amount        = wc_format_decimal(
-			$presentment_amount / $stripe_amount,
-			wc_get_price_decimals()
-		);
+
+		// Use the decimal count for the presentment currency, not the store's price decimal
+		// setting, to avoid incorrect rounding (e.g. JPY stores with 0 decimal places).
+		$presentment_currency_lower = strtolower( $presentment_currency );
```
daledupreez, line 1132: Non-blocking nit: it may also be worth using `$presentment_currency_upper` here so we can re-use the value below, as we have two calls to `strtoupper( $presentment_currency )` below.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -3995,6 +4013,8 @@ public function test_add_converted_currency_information_returns_unchanged_total_
 	 * @return void
 	 */
 	public function test_add_converted_currency_information_appends_converted_currency_info(): void {
+		add_filter( 'woocommerce_is_order_received_page', '__return_true' );
```
daledupreez, line 4016: Is it worth adding a case for My Orders as well, just to provide more basic coverage?

daledupreez (commented):
I have some quick feedback from inspection -- I will aim to tackle manual tests later today, but it may be a few hours.

client/blocks/checkout-sessions/__tests__/checkout-container.test.js
```diff
@@ -52,6 +65,7 @@ describe( 'CheckoutSessionsContainer', () => {
 		render(
 			<CheckoutContainer
 				api={ api }
+				billing={ billing }
```
daledupreez: The `billing` prop is no longer accepted by the `CheckoutContainer` component.
```suggestion
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1052,6 +1058,97 @@ public function attach_currency_selector_element() {
...
+	/**
+	 * Adds the converted currency information to the order total on the order received page and My Account orders when the order is paid with a different currency than the store currency.
+	 *
+	 * @param string   $formatted_total  Total to display.
+	 * @param WC_Order $order            Order data.
+	 */
+	public function add_converted_currency_information( string $formatted_total, WC_Order $order ): string {
+		if ( ! is_order_received_page() && ! is_account_page() ) {
+			return $formatted_total;
+		}
```
daledupreez, line 1070: I just checked and the `woocommerce_get_formatted_order_total` filter can also be triggered from `WC_Abstract_Order->get_formatted_order_total()`, in which case the `$order` argument may not be an instance of `WC_Order`. We should handle that case gracefully without a fatal due to a type error.
```suggestion
	 * @param string               $formatted_total  Total to display.
	 * @param WC_Abstract_Order $order            Order data.
	 */
	public function add_converted_currency_information( string $formatted_total, $order ): string {
		if ( ! $order instanceof WC_Order ) {
			return $formatted_total;
		}
		if ( ! is_order_received_page() && ! is_account_page() ) {
			return $formatted_total;
		}
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1052,6 +1058,97 @@ public function attach_currency_selector_element() {
...
+		$this->maybe_add_presentment_metadata_to_order( $order );
+
+		$presentment_amount   = (int) $order_helper->get_stripe_presentment_amount( $order );
+		$presentment_currency = $order_helper->get_stripe_presentment_currency( $order );
+
+		if ( ! $presentment_amount || ! $presentment_currency ) {
+			return $formatted_total;
+		}
+
+		$currency_symbol = get_woocommerce_currency_symbol( $presentment_currency );
```
daledupreez: Small bug: `$presentment_currency` is in lower-case, so `$currency_symbol` will be empty as `get_woocommerce_currency_symbol()` assumes the value is in upper case.
```suggestion
		$currency_symbol = get_woocommerce_currency_symbol( strtoupper( $presentment_currency ) );
```

daledupreez: Related to the general formatting helper, getting this value right may get us partway to where we would like to be: `$` (or `&#36;`) is the value used for multiple currencies. And the current approach doesn't handle different currency positions (e.g. some countries in Europe use `123.45 €` and not `€ 123.45`).

daledupreez: And we should include the output of the currency lookups in the tests. The bug here was missed because the tests are not looking for the currency symbols.

daledupreez (approved):
Changes are looking good to me in testing, though I noticed one small bug that should be easy to fix.

Image

client/blocks/checkout-sessions/__tests__/checkout-container.test.js
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/commits/3dc5d24b995288acedf7f5fdb497488ac894cf5b

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: I am now checking for the currency symbol in tests (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086/commits/eb1ae6fa6d264f8b1fa56155c5bc841c990590df)

wjrosa: As for the symbol position, I don't see examples in the WC Core code checking/handling that. So, I will leave it like this for now, and we may fix that later (may require a fix or new functions in Core, too).

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Sorry Dale. I missed this comment and some others. Addressig all of them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5151

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5151

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5151

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: It would be `0`, and we would not fall here. The notice is not shown (tested it now). But, in the block checkout, I don't see the new order in the WP-Admin in this case. It is something we need to double-check (but unrelated to the changes here).

PR #5090, opened 2026-03-04
Fix syntax on the Extension Compatibility workflow

zhongruige:

The `setup-matrix` step in the Extension Compatibility workflow fails when neither `plugin-slug` nor `extension-set` is provided (the default path for merged PRs). The `else` branch used a `cat` heredoc to build the matrix JSON, which produced multi-line output with leading whitespace. Since `$GITHUB_OUTPUT` expects `key=value` on a single line, GitHub Actions rejected the subsequent lines with `Invalid format`.

This replaces the `cat` heredoc with `jq -nc`, matching the approach already used by the other two branches. `jq -nc` outputs compact, single-line JSON that is compatible with `$GITHUB_OUTPUT`.

## Testing instructions

No user-facing changes. This is a CI workflow fix.

Verified locally that:
- The old heredoc approach produces 7 lines (broken).
- The new `jq -nc` approach produces 1 line of valid JSON.
- All three branches (`plugin-slug`, `extension-set`, default) produce valid single-line JSON with the correct `include` array structure.

---

### Changelog entry

<details>

#### Comment

CI-only change — fixes the Extension Compatibility workflow. No runtime code modified.

</details>


daledupreez (approved):
Thanks for the fix! I'll get this merged ASAP to get the automations running.

zhongruige:
Awesome thanks confirming that's working now! https://github.com/woocommerce/woocommerce-gateway-stripe/actions/runs/22707676374

PR #5091, opened 2026-03-04
[Proof of concept] Defer express checkout nonce requirements

daledupreez:

This PR explores removing our current logic that forces the generation of a WooCommerce customer session when express checkout is enabled on the product page, which was originally instituted to allow for the generation of nonces in the page payload. Instead, we now have more nuanced logic:
 - If we already have a WooCommerce customer session, we keep the current logic where we send the nonce data in `wc_stripe_express_checkout_params`.
 - If we don't have a WooCommerce customer session yet, we no longer send the nonce details in the initial request.

In the client code, we can then request the nonces (and the creation of a customer session) later on. While this seems to be working for the most part, my current implementation is triggering the AJAX fetch of the nonces during page load, which means we'd only serve one cached product page, as we'd create a session when running the AJAX fetch. It would be better to trigger that logic when the user first interacts with the Express Checkout button. We will just need to ensure that we don't inject our logic synchronously, as most browsers prevent page events from being delayed too long from the initial interaction (e.g. opening a modal). I tripped over this earlier today when debugging some of the related code!

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/express-checkout/utils/index.js
+++ b/client/express-checkout/utils/index.js
@@ -38,7 +38,7 @@ export const getErrorMessageFromNotice = ( notice ) => {
  */
 export const getExpressCheckoutData = ( key ) =>
 	// eslint-disable-next-line camelcase
-	wc_stripe_express_checkout_params[ key ] ?? null;
+	wc_stripe_express_checkout_params?.[ key ] ?? null;
 
 /**
  * Construct Express Checkout AJAX endpoint URL.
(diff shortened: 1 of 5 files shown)


client/express-checkout/utils/index.js
```diff
@@ -56,6 +56,53 @@ export const getExpressCheckoutAjaxURL = (
...
+
+	if ( ! expressCheckoutNoncePromise ) {
+		expressCheckoutNoncePromise = fetch(
+			getExpressCheckoutAjaxURL( 'express_checkout_get_nonces' ),
+			{
+				method: 'POST',
+			}
+		)
+			.then( ( res ) => res.json() )
+			.then( ( data ) => data?.data ?? {} );
```
malithsen, line 87: If the request fails, we'd want to invalidate `expressCheckoutNoncePromise` so that we dont unconditionally _cache_ it

```suggestion
			.then( ( data ) => data?.data ?? {} );
    		.catch( ( err ) => {
                  expressCheckoutNoncePromise = null;
                  throw err;
             });
```

malithsen (commented):
Editing my initial comment after diving into this a bit more. My understanding of the underlying issue is that `set-cookie` header is the main culprit that prevents caching of these pages for guest users. On develop I see the `set-cookie` header in the page load response, which will prevent most servers from caching the product page. 

Image

Absence of that header in this branch itself is a big improvement.

That said, I am not actually sure if the new ajax endpoint needs to be nonce free. AFAICS [wp_create_nonce](https://developer.wordpress.org/reference/functions/wp_create_nonce/) doesn't rely on the session cookie. As a part of the `wp_get_session_token`, it looks at the `wordpress_logged_in_*` which won't exist for guest users. So it seems like all guests produce the same nonce for the same action within the same tick window (12 hours).

If the above is correct, then the nonces could safely remain in the page payload, and the new endpoint could be protected by one of those nonces. Its only job would then be to create the WC session when needed.

client/express-checkout/utils/index.js
```diff
@@ -56,6 +56,53 @@ export const getExpressCheckoutAjaxURL = (
...
+	const nonceFromData = getExpressCheckoutData( 'nonce' );
+	if ( nonceFromData ) {
+		return Promise.resolve( nonceFromData );
+	}
+
+	if ( ! expressCheckoutNoncePromise ) {
+		expressCheckoutNoncePromise = fetch(
+			getExpressCheckoutAjaxURL( 'express_checkout_get_nonces' ),
+			{
+				method: 'POST',
```
wjrosa, line 83: It feels a bit strange to me request in the "get" context using `POST`

daledupreez:
Closing this PR in favour of the _much_ simpler changes in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5100.

PR #5097, opened 2026-03-05
Refactor order-received page check to use is_order_received_page

JeroenSormani:

Does the same check, but through the existing appropriate conditional-function, which also has a filter hook included in it. 

Reason for this change came from having a custom thank-you page configured, but when canceling a payment it would continue to send the customer to the TY page instead of back to the checkout as it should without the custom TY page.
This issue only exists with this payment gateway, not the original ones. 

### Changelog entry

<details>

#### Comment 
Don't think this warrants a changelog comment as its still doing the same thing.

</details>

I think the change is pretty straight forward, but please let me know if you need any further information.


wjrosa:
Thanks for your contribution, @JeroenSormani ! Looks good to me 👍

wjrosa (approved):

Mayisha (approved):
Thanks for the refactor. Looks good to me 👍

PR #5102, opened 2026-03-06
Process order with checkout sessions on classic checkout

Mayisha:

- Used `action.confirm()` from checkout session element to complete payment on Stripe side.
- Appending `wc_stripe_checkout_session_id` flag to include in the `/checkout` request which is expected in the backend to go through the checkout session path to process payment.

## Testing instructions

- Add the following filter to allow adaptive pricing and checkout session in your store-
```
```
- Enable adaptive pricing under the OCS settings.
- Enable card and a couple more payment methods.
- As a shopper, add a product to your cart and go to the classic/shortcode checkout page.
- Verify that, the currency selector is displayed and the available currencies are USD and your local currency.
- Select your local currency and pay with card.
- As an admin, verify that the order is successful and the transaction details are present.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/stripe-utils/utils.js
```diff
@@ -505,6 +505,12 @@ export const getUpeSettings = () => {
 	return upeSettings;
 };
 
+export const appendCheckoutSessionIdToForm = ( form, checkoutSessionId ) => {
+	form.append(
+		`<input type="hidden" id="wc_stripe_checkout_session_id" name="wc_stripe_checkout_session_id" value="${ checkoutSessionId }" />`
+	);
+};
```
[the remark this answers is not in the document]
Mayisha: Addressed.

client/stripe-utils/utils.js
```diff
@@ -505,6 +505,12 @@ export const getUpeSettings = () => {
 	return upeSettings;
 };
 
+export const appendCheckoutSessionIdToForm = ( form, checkoutSessionId ) => {
+	form.append(
+		`<input type="hidden" id="wc_stripe_checkout_session_id" name="wc_stripe_checkout_session_id" value="${ checkoutSessionId }" />`
+	);
```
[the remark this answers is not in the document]
Mayisha: Addressed.

client/classic/upe/payment-processing.js
```diff
@@ -602,45 +613,65 @@ export const processPayment = (
...
 
-			// Append the intent ID to the form if it was previously created through a non-deferred intent.
-			if ( gatewayUPEComponents[ paymentMethodType ].intentId ) {
-				appendPaymentIntentIdToForm(
+				appendCheckoutSessionIdToForm(
 					jQueryForm,
-					gatewayUPEComponents[ paymentMethodType ].intentId
+					confirmResult.session.id
+				);
+			} else {
```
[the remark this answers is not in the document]
Mayisha: Added check for `loadAction` before calling.

client/classic/upe/payment-processing.js
```diff
@@ -602,45 +613,65 @@ export const processPayment = (
...
+				} );
 
-			// Append the intent ID to the form if it was previously created through a non-deferred intent.
-			if ( gatewayUPEComponents[ paymentMethodType ].intentId ) {
-				appendPaymentIntentIdToForm(
+				appendCheckoutSessionIdToForm(
 					jQueryForm,
-					gatewayUPEComponents[ paymentMethodType ].intentId
+					confirmResult.session.id
+				);
```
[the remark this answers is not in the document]
Mayisha: Good catch 👍 
Added a check for error in confirm request.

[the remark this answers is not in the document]
Mayisha: Handled this case already.

client/classic/upe/payment-processing.js
```diff
@@ -602,45 +613,65 @@ export const processPayment = (
...
+					() => {
+						// Provide a callback to flag that a redirect has occurred.
+						stopFormSubmission = true;
+					}
+				);
+
+				if ( stopFormSubmission ) {
+					return;
+				}
 			}
```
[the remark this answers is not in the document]
Mayisha: Block checkout is not part of this PR.

client/stripe-utils/utils.js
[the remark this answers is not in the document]
Mayisha: Addressed.

client/classic/upe/payment-processing.js
```diff
@@ -485,7 +479,24 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+		isAdaptivePricingEnabled &&
+		elements &&
+		typeof elements.loadActions === 'function'
+	) {
+		const actions = await elements.loadActions();
+
+		if ( actions.type === 'error' ) {
+			showErrorPaymentMethod( actions?.error?.message, domElement );
+			// Setting the flag to true to prevent the form from being submitted.
+			component.hasLoadError = true;
```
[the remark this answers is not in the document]
Mayisha: Addressed.

client/classic/upe/__tests__/payment-processing.test.js
```diff
@@ -0,0 +1,490 @@
...
+
+				expect( mockActions.confirm ).toHaveBeenCalledWith( {
+					returnUrl: window.location.href,
+					redirect: 'if_required',
+				} );
+				expect(
+					stripeUtils.appendCheckoutSessionIdToForm
+				).toHaveBeenCalledWith( form, 'cs_session_xyz' );
+				expect( form.trigger ).toHaveBeenCalledWith( 'submit' );
+			} );
```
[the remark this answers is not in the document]
Mayisha: Added.

wjrosa (approved):
Looks good to me and works as expected! 🎉 

Image

client/classic/upe/payment-processing.js
```diff
@@ -48,6 +48,8 @@ export function initializeUPEComponents() {
 			hasLoadError: false,
 		};
 	}
+	// Reset so processPayment runs fully when called again (e.g. after re-init or in tests).
+	hasCheckoutCompleted = false;
```
daledupreez, line 67: Nit: Can we shift this variable initialisation to the top of this file? I almost missed it in it's current location, which is at the start of a function halfway down the file.

client/classic/upe/payment-processing.js
```diff
@@ -487,7 +485,38 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+	// Call loadActions() after mounting the elements with the Checkout Session API to check if there are any errors.
+	const component = gatewayUPEComponents[ paymentMethodType ];
+	const elements = component.elements;
+	const isAdaptivePricingEnabled =
+		getStripeServerData()?.isAdaptivePricingEnabled;
+	if (
+		isAdaptivePricingEnabled &&
+		elements &&
+		typeof elements.loadActions === 'function'
+	) {
```
daledupreez, line 563: Not at all blocking, but it feels like it could be worth using an early return here.
```suggestion
	const isAdaptivePricingEnabled =
		getStripeServerData()?.isAdaptivePricingEnabled;
	if (
		! isAdaptivePricingEnabled ||
		! elements ||
		typeof elements.loadActions !== 'function'
	) {
		return component;
	}
```

client/classic/upe/payment-processing.js
```diff
@@ -487,7 +485,38 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+		try {
+			const actions = await elements.loadActions();
+
+			if ( actions.type === 'error' ) {
+				showErrorPaymentMethod( actions?.error?.message, domElement );
```
daledupreez: Can we add a fallback message if we don't think `actions.error.message` will always be defined?
```suggestion
				showErrorPaymentMethod(
					actions?.error?.message ?? __(
						'Failed to load payment method. Please refresh the page and try again.',
						'woocommerce-gateway-stripe'
					),
					domElement
				);
```

client/classic/upe/payment-processing.js
```diff
@@ -487,7 +485,38 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+				// Setting the flag to true to prevent the form from being submitted.
+				component.hasLoadError = true;
+			}
+		} catch ( error ) {
+			showErrorPaymentMethod(
```
daledupreez: This mimics the `actions.type === 'error'` logic above. Might it make sense to consider using a `let errorMessage;` and maybe a finally block that actions on the variable? That way the main error handling is consolidated.

client/stripe-utils/utils.js
```diff
@@ -505,6 +505,21 @@ export const getUpeSettings = () => {
 	return upeSettings;
 };
 
+export const appendCheckoutSessionIdToForm = ( form, checkoutSessionId ) => {
+	// If the element already exists, remove it first, to avoid duplicates.
+	const existingElement = form.find( 'input#wc_stripe_checkout_session_id' );
+	if ( existingElement.length ) {
+		existingElement.remove();
```
daledupreez: Not blocking: might it be easier to re-use the element if it already exists rather than removing and re-creating it?

client/classic/upe/__tests__/payment-processing.test.js
```diff
@@ -0,0 +1,518 @@
...
+};
+
+const BASE_SERVER_DATA = {
+	paymentMethodsConfig: {
+		card: { supportsDeferredIntent: true, title: 'Card' },
+	},
+	currency: 'usd',
+	cartTotal: 1000,
+	isCheckout: true,
+	isOCEnabled: false,
```
daledupreez: Nit: This should be updated to `shouldShowOptimizedCheckout`.
```suggestion
	shouldShowOptimizedCheckout: false,
```

daledupreez (commented):
The changes are working as expected, and test well. I do have some feedback that may or may not be worth addressing before we merge, but none of it was blocking.

Mayisha:
@daledupreez this is ready for another round of review.

client/classic/upe/payment-processing.js
```diff
@@ -45,14 +46,14 @@ import { handleDisplayOfSavingCheckbox } from 'wcstripe/optimized-checkout/handl
  * @type {Object<string, UPEComponent>}
  */
 const gatewayUPEComponents = {};
-const paymentMethodsConfig = getStripeServerData()?.paymentMethodsConfig;
-const isAdaptivePricingEnabled =
-	getStripeServerData()?.isAdaptivePricingEnabled;
+let hasCheckoutCompleted = false;
```
daledupreez, line 49: 😍

client/classic/upe/payment-processing.js
```diff
@@ -552,7 +551,46 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+			component.hasLoadError = true;
+		}
+	} catch ( error ) {
+		loadActionsError = error?.message;
+		component.hasLoadError = true;
+	}
+
+	if ( loadActionsError ) {
+		showErrorPaymentMethod(
+			loadActionsError ??
```
daledupreez: Question: do we want to show `loadActionsError` to shoppers? It feels like we should log the value, so we have the details, but I don't know that we should show it to shoppers.

client/classic/upe/payment-processing.js
```diff
@@ -685,45 +723,92 @@ export const processPayment = (
...
-				paymentMethodObject.paymentMethod.id
-			);
+				const sessionId = confirmResult?.session?.id;
+				if ( ! sessionId ) {
+					throw new Error(
+						__(
+							'Payment could not be completed. Please try again.',
+							'woocommerce-gateway-stripe'
+						)
+					);
```
daledupreez, line 765: Given that our fallback messages are translated, I am not sure if it makes sense to throw the error returned from Stripe -- it seems like we may end up showing the error to shoppers.

daledupreez (approved):
The happy path for this PR is working as expected, but there are some issues we need to address in follow-ups:
 *  Redirect-based payments don't seem to be handling the redirect back to the site -- I keep getting returned to checkout with no clear indication as to what is going on
 * It doesn't look like we are pulling in data from the billing page, as payment methods that require additional fields are exposing embedded fields that may replicate billing fields, e.g. address fields for SEPA, name fields for SEPA and Bancontact
 * When we're using checkout sessions, I am seeing a console error that `<var>.elements.update()` is not a function, which looks like it may come from the following two code locations:
     - Classic checkout:
https://github.com/woocommerce/woocommerce-gateway-stripe/blob/507d995f0713974ae4aeca4d1ebc064549b78a5b/client/classic/upe/deferred-intent.js#L180-L202
     - Blocks checkout:
https://github.com/woocommerce/woocommerce-gateway-stripe/blob/507d995f0713974ae4aeca4d1ebc064549b78a5b/client/blocks/upe/upe-deferred-intent-creation/payment-processor.js#L282-L301

Note that I was logged in, but I assume the same issues with details will apply to that case too.

@wjrosa, could I ask you to create one or more issues to track the things I found? I think we have at least three issues to address, but I am not sure how best to split up the work, or if there are things I may have missed:
 - Getting the right user data across to the Stripe element/checkout sessions from the billing details
 - Getting redirect payment methods to work
 - Adding guards to prevent calls to `elements.update()` when the function is not available

client/classic/upe/payment-processing.js
Mayisha: I followed the existing pattern from here https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/client/classic/upe/payment-processing.js#L523-L527. We show the load error for upe elements.

client/classic/upe/payment-processing.js
Mayisha: You are right about the translation. On the other hand, I think showing the specific message from Stripe would be helpful. For example, I simulated an error scenario and got this notice which explains the exact error.

Screenshot 2026-03-24 at 11 16 15 AM

Mayisha:
Thanks for the review @daledupreez. I will create the follow-up issues. cc: @wjrosa

PR #5109, opened 2026-03-10
Making ECE debug logs optional

wjrosa:
See p1768294774721149-slack-C055WHLA98D / p1773142676536059-slack-C055WHLA98D

In this PR, I am making the existing ECE debug logs optional, being added only when verbose debug mode is enabled (using the `wc_stripe_is_verbose_debug_mode_enabled` filter). The goal is to reduce noise on merchants' log files.

## Testing instructions

- Checkout and build this branch on your test environment (`update/ece-debug-logs-optional`)
- Enable verbose debug logs with the filter:
```php
add_filter( 'wc_stripe_is_verbose_debug_mode_enabled', '__return_true' );
```
- Connect your Stripe account
- Enable ECE methods (Google Pay, Apple Pay, Amazon Pay)
- Disable ECE display in the cart page
- Go to your cart. ECE buttons should not be shown
- As a merchant, go to your Stripe logs page
- Confirm the following entry is logged:
```
Stripe Express Checkout buttons display on cart is disabled.
```
- Remove the filter above
- Repeat the test steps
- Confirm the entry is no longer added to your Stripe logs

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.6.0 - xxxx-xx-xx =
+* Update - Logging reasons for ECE buttons not showing only when verbose debug mode is enabled
 * Dev - Remove unused frontend code: legacy blocks payment request API helpers, related normalize utilities, and unused Stripe icon component
 
 = 10.5.0 - 2026-03-09 =
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 = 10.6.0 - xxxx-xx-xx =
-* Update - Logging reasons for ECE buttons not showing only when verbose debug mode is enabled
+* Update - Express Checkout button logging will only occur when verbose debug mode is enabled
 * Update - Disable the Optimized Checkout Suite in the "Add Payment Method" and "Change Subscription Payment Method" screens
 * Dev - Remove unused frontend code: legacy blocks payment request API helpers, related normalize utilities, and unused Stripe icon component
 
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -792,7 +792,7 @@ class WC_Stripe_Express_Checkout_Helper {
 				if ( $hide_based_on_tax !== $hide_based_on_tax_filtered ) {
 					WC_Stripe_Logger::debug( 'Stripe Express Checkout is hidden due to the tax setup being overridden by the filter.' );
 				} else {
-					WC_Stripe_Logger::debug( 'Stripe Express Checkout is hidden due to product/cart not requiring shipping and tax being based on customer\'s billing or shipping address.' );
+					WC_Stripe_Logger::debug( "Stripe Express Checkout is hidden due to product/cart not requiring shipping and tax being based on customer's billing or shipping address." );
 				}
 			}
 			return false;
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -693,7 +693,7 @@ class WC_Stripe_Express_Checkout_Helper {
 			&& ! $this->allowed_items_in_cart()
 		) {
 			if ( WC_Stripe_Helper::is_verbose_debug_mode_enabled() ) {
-				WC_Stripe_Logger::debug( 'Some items in cart are not compatible with Stripe Express Checkout. ' );
+				WC_Stripe_Logger::debug( 'Some items in cart are not compatible with Stripe Express Checkout.' );
 			}
 			return false;
 		}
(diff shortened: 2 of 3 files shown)


Mayisha (approved):
Looks good to me. 👍 

Non-blocker idea for later. We are checking `is_verbose_debug_mode_enabled()` before all the conditional logging (here and in other places). It might be a good idea to introduce a `WC_Stripe_Logger::verbose_debug` method and have the `is_verbose_debug_mode_enabled` check there only. These conditional logs can call the `verbose_debug` logger method.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -651,31 +651,39 @@ public function is_page_supported() {
...
-			];
-			// phpcs:ignore WordPress.PHP.DevelopmentFunctions
-			WC_Stripe_Logger::debug( 'Stripe Express Checkout live mode requires SSL', [ 'server_details' => $server_details ] );
+			if ( WC_Stripe_Helper::is_verbose_debug_mode_enabled() ) {
+				$server_details = [
+					'url'   => get_permalink(),
+					'https' => isset( $_SERVER['HTTPS'] ) ? wp_unslash( $_SERVER['HTTPS'] ) : '', // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+					'port'  => isset( $_SERVER['SERVER_PORT'] ) ? wp_unslash( $_SERVER['SERVER_PORT'] ) : '', // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+				];
+				// phpcs:ignore WordPress.PHP.DevelopmentFunctions
```
daledupreez: Nit: Do we still need this sniff? I don't see any functions that might trigger the sniff on the following line.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -685,21 +693,27 @@ public function should_show_express_checkout_button() {
 			WC_Stripe_Helper::has_cart_or_checkout_on_current_page()
 			&& ! $this->allowed_items_in_cart()
 		) {
-			WC_Stripe_Logger::debug( 'Some items in cart are not compatible with Stripe Express Checkout. ' );
+			if ( WC_Stripe_Helper::is_verbose_debug_mode_enabled() ) {
+				WC_Stripe_Logger::debug( 'Some items in cart are not compatible with Stripe Express Checkout. ' );
```
daledupreez: Non-blocking nit: I can't unsee that a number of these messages end with a space after the `.`. 🙈 It may be worth cleaning these up, but it's _definitely_ not a blocker.
```suggestion
				WC_Stripe_Logger::debug( 'Some items in cart are not compatible with Stripe Express Checkout.' );
```

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -753,18 +779,22 @@ public function should_show_express_checkout_button() {
...
 		if ( $hide_based_on_tax_filtered ) {
-			if ( $hide_based_on_tax !== $hide_based_on_tax_filtered ) {
-				WC_Stripe_Logger::debug( 'Stripe Express Checkout is hidden due to the tax setup being overridden by the filter.' );
-			} else {
-				WC_Stripe_Logger::debug( 'Stripe Express Checkout is hidden due to product/cart not requiring shipping and tax being based on customer\'s billing or shipping address.' );
+			if ( WC_Stripe_Helper::is_verbose_debug_mode_enabled() ) {
+				if ( $hide_based_on_tax !== $hide_based_on_tax_filtered ) {
+					WC_Stripe_Logger::debug( 'Stripe Express Checkout is hidden due to the tax setup being overridden by the filter.' );
+				} else {
+					WC_Stripe_Logger::debug( 'Stripe Express Checkout is hidden due to product/cart not requiring shipping and tax being based on customer\'s billing or shipping address.' );
```
daledupreez: Nit: To make this message easier to find when searching, it may be worth switching to `"` for the string so we can remove the `\'` value.
```suggestion
					WC_Stripe_Logger::debug( "Stripe Express Checkout is hidden due to product/cart not requiring shipping and tax being based on customer's billing or shipping address." );
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.6.0 - xxxx-xx-xx =
+* Update - Logging reasons for ECE buttons not showing only when verbose debug mode is enabled
```
daledupreez: This entry is a bit confusing to me due to the "not showing only ... enabled" phrasing. I think we should reword this, and we should also use "Express Checkout" instead of "ECE".
```suggestion
* Update - Express Checkout button logging will only occur when verbose debug mode is enabled
```

daledupreez (approved):
Overall, this is working as I would expect it to. The one issue I tripped over is that the log message I saw was `No Stripe Express Checkout options are enabled in the current context.` after I disabled showing Amazon Pay and Apple Pay/Google Pay in the cart vai the two setting UIs. It may be worth updating the testing instructions to be more explicit as to how to get the expected message, or updating the testing instructions.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: We don't really need it. Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5109/commits/e7dc7141944b22a6751a5c49323263e7dd1da3e1

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: I noticed the same. I wanted to avoid moving away from the main purpose of this PR, but took the freedom to remove all of them already in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5109/commits/221a717fb13a05a61641d48b81440889d0546591

PR #5113, opened 2026-03-11
Simplify CSS detection logic

daledupreez:
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5034

This PR is a small refactor stemming from my investigations in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5034. This particular change should simplify our logic for detecting CSS values for the Stripe embedded elements and drastically improve the performance, via two changes:
 1. I've changed our loop logic to iterate over the fields we actually care about. This should be much faster than the previous logic, because we have around 50 possible properties we want to check, while Chrome returns ~400 basic CSS properties and another ~150 properties for CSS variables on my test site. So this should deliver a 90% reduction in the number of iterations that are needed.
 2. I have also changed the logic to use direct access via camel-cased property names, which also means we can avoid calling `dashedToCamelCase()` for every iteration. The direct lookup should be a _lot_ less work than the function call which itself performs a regex replace across the string value.

## Testing instructions

* Set up a test store using the code from `develop`
* Connect to Stripe and enable multiple payment methods, especially cards
* Add a product to your cart, and navigate to classic checkout in one tab and block checkout in a second tab
* Do the same using a mobile browser
* Expand one or more payment methods, and inspect the styles for the elements (especially the input fields for cards)
* Now switch to this branch and rebuild the UI: `git checkout refactor/stripe-style-calculations-expand-fonts && npm run build`
* Reload both the classic and block checkout tabs, and verify that no visuals have changed.
* Reload the mobile browser windows as well.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


malithsen (approved):
Thanks for improving this. Looks good to me!

Mayisha (approved):
Thanks for the refactor. Looks good to me 👍 

Following is an AI feedback I got, which I think we should consider.
**Note**: In upe-styles.js, textFontTransitionProps still lists '-webkit-font-smoothing' and '-moz-osx-font-smoothing'. On many browsers getComputedStyle() exposes these as webkitFontSmoothing and mozOsxFontSmoothing, so styles['-webkit-font-smoothing'] can be undefined and those two will never be applied. If you care about them, the allowed list should use the camelCase names.

daledupreez:
Good catch, @Mayisha (or your bot!). Fixed in 961d18aaa32cc931868d823f9ad6c6c3d6cec03b.

PR #5119, opened 2026-03-11
Remove strict param from existing public method

Mayisha:
Removed the strict param from the existing public methods.

## Testing instructions
- Code review

### Changelog entry

<details>

#### Comment 

Removed a strict param check which was added in another PR going in the same release. So reverting this part does not need a new changelog entry.

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -981,7 +981,7 @@ class WC_Stripe_Customer {
 	 * @return bool True if the default payment method was set successfully, false otherwise.
 	 * @throws WC_Stripe_Exception
 	 */
-	public function set_default_payment_method( $payment_method_id ) {
+	public function set_default_payment_method( $payment_method_id ): bool {
 		$response = WC_Stripe_API::request(
 			[
 				'invoice_settings' => [
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -953,7 +953,7 @@ class WC_Stripe_Customer {
 	 * @return bool True if the default source was set successfully, false otherwise.
 	 * @throws WC_Stripe_Exception
 	 */
-	public function set_default_source( $source_id ) {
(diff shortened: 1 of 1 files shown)


Mayisha:
@daledupreez should I ship this change in 10.6.0? Or do you think it's fine to add the strict parameter in the major release?

includes/class-wc-stripe-customer.php
```diff
@@ -953,7 +953,7 @@ public function detach_payment_method( $payment_method_id ) {
 	 * @return bool True if the default source was set successfully, false otherwise.
 	 * @throws WC_Stripe_Exception
 	 */
-	public function set_default_source( string $source_id ): bool {
+	public function set_default_source( $source_id ) {
```
daledupreez: I think we can keep the `bool` return type, as it's clear that we only return `true|false`.
```suggestion
	public function set_default_source( $source_id ): bool {
```

includes/class-wc-stripe-customer.php
```diff
@@ -981,7 +981,7 @@ public function set_default_source( string $source_id ): bool {
 	 * @return bool True if the default payment method was set successfully, false otherwise.
 	 * @throws WC_Stripe_Exception
 	 */
-	public function set_default_payment_method( string $payment_method_id ): bool {
+	public function set_default_payment_method( $payment_method_id ) {
```
daledupreez: As above, I think we can keep the explicit return type.
```suggestion
	public function set_default_payment_method( $payment_method_id ): bool {
```

daledupreez (commented):
@Mayisha, I think we should ship the updates from this PR in 10.6.0, mostly because adding the strict types to the method parameters can cause runtime exceptions for code that was previously running fine. If the supplied argument was not a valid source ID or payment method ID, the related API call would fail, and we would get a return value. And adding the strict types would break that in a really bad way.

daledupreez (approved):
Thanks for taking this on! :shipit:

PR #5122, opened 2026-03-12
Adding adaptive pricing specific information to emails

### LINKED ISSUE #5120: Checkout Sessions: Add specific adaptive pricing information to the order received email
Figma: 0fiyiUe18lGU6kGyQNmXxE-fi/Stripe-Payments?node-id=2702-208227&p=f&t=X8Hxfo8hRa4PN3nu-0

We need to add specific adaptive pricing information to the order-received email as part of the Adaptive Pricing project. This means adding the converted amount next to the total and a notice showing the conversion rate.

wjrosa:
Fixes STRIPE-983
Fixes #5120
Figma 0fiyiUe18lGU6kGyQNmXxE-fi/Stripe-Payments?node-id=2702-208227&p=f&t=X8Hxfo8hRa4PN3nu-0

Following the design pieces for the order received email template, this PR adds three new pieces of information to it:
- The converted total along the default total, in the order summary table
- A notice informing the shopper about the conversion after the summary table

Preview:
Screenshot 2026-03-12 at 18 26 11

## Testing instructions

- Checkout and build this branch on your test environment (`add/adaptive-pricing-info-to-emails`)
```php
```
- Connect a US Stripe account
- As a shopper, add a product to your cart and go to the block checkout page
- Select your local currency and pay
- As a shopper, confirm the new information is shown in the order received email*

\* You may need to test this using a JN website or something similar.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
@copilot Add changelog entries to both changelog.txt and readme.txt files, and unit tests for the new method introduced in this PR.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1088,6 +1091,43 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
...
 	}
 
+	/**
+	 * Shows a notice to the order received page to inform the customer about the currency conversion
+	 * when the order is paid with a different currency than the store currency.
+	 *
+	 * @return void
+	 */
+	public function add_email_currency_conversion_notice( WC_Order $order ): void {
+		$checkout_session = $this->get_checkout_session_from_order( $order );
```
Mayisha: We can check the order meta in this method first before triggering the API call, similar to how it is done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1088,6 +1091,43 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
 		echo '</p>';
 	}
 
+	/**
+	 * Shows a notice to the order received page to inform the customer about the currency conversion
+	 * when the order is paid with a different currency than the store currency.
+	 *
+	 * @return void
+	 */
+	public function add_email_currency_conversion_notice( WC_Order $order ): void {
```
Mayisha: I see a lot of similarities here with the `add_currency_conversion_notice` method? Is consolidating these two a possible option?

Mayisha (changes requested):
The notice in the email looks good for the shopper. But I think we should state it differently in the email received by the admin/merchant. The statement `you chose to pay...` is not correct here.

Image

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: You're right. Done in my latest commits

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: That's a good idea. I reduced the duplicate code in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/commits/d5ad7111a4af5da0822657c7f4d542b6ac634c14

wjrosa:
@copilot Update and fix PHPUnit tests based on the latest changes on this PR

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1093,41 +1085,70 @@ public function add_converted_currency_information( string $formatted_total, WC_
...
+				$order->get_total(),
+				[
+					'currency' => $order_currency,
+					'in_span'  => false,
+				]
+			) . ' ' . strtoupper( $order_currency );
 			printf(
-				/* translators: %1$s Converted amount and currency. %2$s Store currency. %3$s Exchange rate and currency. */
+			/* translators: %1$s Converted amount and currency. %2$s Store currency. %3$s Exchange rate and currency. %4$s Original store amount. */
+				esc_html__( 'Adaptive Pricing Applied: The customer opted to pay %1$s (%2$s = %3$s). Your settlement remains unchanged at the original store price of %4$s.', 'woocommerce-gateway-stripe' ),
```
Mayisha:
```suggestion
				esc_html__( 'Adaptive Pricing Applied: The customer opted to pay %1$s (1 %2$s = %3$s). Your settlement remains unchanged at the original store price of %4$s.', 'woocommerce-gateway-stripe' ),
```

There should be a `1` otherwise the notice the notice says-

Image

Mayisha (approved):
Thanks for addressing the feedback. The admin notice needs one minor fix. Otherwise, looks good to me 👍

phpstan-baseline.neon
```diff
@@ -4800,6 +4800,12 @@ parameters:
 			count: 1
 			path: includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
 
+		-
+			message: '#^Filter callback return statement is missing\.$#'
+			identifier: return.missing
+			count: 1
+			path: includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```
wjrosa: @daledupreez do you know if it is possible to add the `line` information here?

daledupreez: I know that PHPStan's JSON format includes more specific information, but that's much harder to diff and inspect.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1072,29 +1075,19 @@ public function add_converted_currency_information( string $formatted_total, WC_
 			return $formatted_total;
 		}
 
-		$order_helper = WC_Stripe_Order_Helper::get_instance();
-
-		$checkout_session_id = $order_helper->get_stripe_checkout_session_id( $order );
-		if ( ! $checkout_session_id ) {
+		$presentment_data = $this->get_presentment_data_from_order( $order );
+		if ( empty( $presentment_data ) || ! ( $presentment_data['amount'] ?? null ) || ! ( $presentment_data['currency'] ?? null ) ) {
```
daledupreez: Might it be cleaner to have `$this->get_presentment_data_from_order()` return `null` when we don't have the amount and currency? Given that this is private logic, it feels like it would be easier to check for null or assume we have the data we need.
```suggestion
		$presentment_data = $this->get_presentment_data_from_order( $order );
		if ( null === $presentment_data ) {
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1104,52 +1097,72 @@ public function add_converted_currency_information( string $formatted_total, WC_
...
+		$notice_data = $this->get_currency_conversion_notice_data( $order );
+		if ( empty( $notice_data ) ) {
 			return;
 		}
 
-		$woocommerce_amount = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount(
-			$presentment_amount,
-			$presentment_currency
-		);
+		echo '<div style="margin-top: 1em; border: solid 1px #007CBA; border-radius: 4px; background-color: #F6F5F8; padding: 1em 2em;">';
```
daledupreez: Do we use these colours in our other emails? I worry that our colour scheme could easily clash with merchant's email colour schemes.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1104,52 +1097,72 @@ public function add_converted_currency_information( string $formatted_total, WC_
...
+					'currency' => $order_currency,
+					'in_span'  => false,
+				]
+			) . ' ' . strtoupper( $order_currency );
 			printf(
-				/* translators: %1$s Converted amount and currency. %2$s Store currency. %3$s Exchange rate and currency. */
+			/* translators: %1$s Converted amount and currency. %2$s Store currency. %3$s Exchange rate and currency. %4$s Original store amount. */
+				esc_html__( 'Adaptive Pricing Applied: The customer opted to pay %1$s (1 %2$s = %3$s). Your settlement remains unchanged at the original store price of %4$s.', 'woocommerce-gateway-stripe' ),
+				esc_html( $converted_amount ),
+				esc_html( $store_currency ),
```
daledupreez: I think this should be the order currency, just to be consistent.
```suggestion
				esc_html( $order_currency ),
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1104,52 +1097,72 @@ public function add_converted_currency_information( string $formatted_total, WC_
...
+			);
+		} else {
+			printf(
+			/* translators: %1$s Converted amount and currency. %2$s Store currency. %3$s Exchange rate and currency. */
 				esc_html__( 'Currency Conversion: You chose to pay %1$s for this order at an exchange rate of 1 %2$s = %3$s.', 'woocommerce-gateway-stripe' ),
-				esc_html( $woocommerce_amount . ' ' . strtoupper( $presentment_currency ) ),
-				esc_html( strtoupper( $order->get_currency() ) ),
-				esc_html( $rate_amount . ' ' . strtoupper( $presentment_currency ) )
+				esc_html( $converted_amount ),
+				esc_html( $store_currency ),
```
daledupreez: This should also be the order currency.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4236,4 +4249,67 @@ private function maybe_add_presentment_metadata_to_order( WC_Order $order ): voi
...
+		// Use the decimal count for the presentment currency, not the store's price decimal
+		// setting, to avoid incorrect rounding (e.g. JPY stores with 0 decimal places).
+		$presentment_currency_lower = strtolower( $presentment_data['currency'] );
+		$rate_decimals              = 2;
+		if ( in_array( $presentment_currency_lower, WC_Stripe_Helper::no_decimal_currencies(), true ) ) {
+			$rate_decimals = 0;
+		} elseif ( in_array( $presentment_currency_lower, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
+			$rate_decimals = 3;
+		}
+		$rate_amount = wc_format_decimal( $presentment_data['amount'] / $stripe_amount, $rate_decimals );
```
daledupreez, line 4339: This was an issue called out in the main order received PR: we need to make sure we're keeping the same minimum units, as the ratio will be wrong if the two currencies use different numbers of decimals.
As an explicit example, if we convert 5.99 USD to 159 JPY, where USD has two decimals and JPY has none, the rate would be computed as 159/599, when it should be 15900/599.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -314,6 +314,9 @@ public function __construct() {
 		// Add a notice about currency conversion when the order currency is different from the store currency on the order details page.
 		add_action( 'woocommerce_order_details_after_order_table', [ $this, 'add_currency_conversion_notice' ], 10 );
 
+		// Add a notice about currency conversion in the order confirmation emails when the order currency is different from the store currency.
+		add_filter( 'woocommerce_email_after_order_table', [ $this, 'add_email_currency_conversion_notice' ], 10, 4 );
```
daledupreez: I think this should be an action, not a filter.
```suggestion
		add_action( 'woocommerce_email_after_order_table', [ $this, 'add_email_currency_conversion_notice' ], 10, 4 );
```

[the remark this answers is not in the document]
daledupreez: The first issue above is accurate -- the converted total is not being added in the email due to the early returns in `add_converted_currency_information()`.

daledupreez (changes requested):
The admin email looked OK to me, but the customer one did not, as the converted amount was missing:

Image

Also, I am assuming that this PR is intentionally limiting the scope to non-EU use cases, but that is not explicitly mentioned anywhere.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/commits/20a63afaf20ff0b470a06fca9ab21a408e9dc659

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: We don't have notices in any other existing email. And I am not sure how we could adapt this notice to use the store's email style 🤔 . We would have to either keep it this way or reduce the number of styles applied so it can fit better any existing style.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
wjrosa: Fix applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/commits/cbf4335c0267bdf463ca680673dd2eb38d30da8c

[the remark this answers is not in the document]
wjrosa: The fix did not work, so I decided to just remove that early return instead in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/commits/ec283c8ded7826dd2b7dafd89c5f61ec876783a5

We already check if the checkout session ID is not present there.

wjrosa:
@copilot fix PHPUnit tests on this PR

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -314,6 +314,9 @@ public function __construct() {
 		// Add a notice about currency conversion when the order currency is different from the store currency on the order details page.
 		add_action( 'woocommerce_order_details_after_order_table', [ $this, 'add_currency_conversion_notice' ], 10 );
 
+		// Add a notice about currency conversion in the order confirmation emails when the order currency is different from the store currency.
+		add_action( 'woocommerce_email_after_order_table', [ $this, 'add_email_currency_conversion_notice' ], 10, 4 );
```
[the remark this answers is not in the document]
daledupreez: This is also flagged in the PHPStan report, but we need to handle plain text emails.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: Should we add a filter so these colours can be customised?

phpstan-baseline.neon
```diff
@@ -2167,15 +2167,9 @@ parameters:
...
-			message: '#^Expected 3 @param tags, found 2\.$#'
-			identifier: paramTag.count
-			count: 1
-			path: includes/class-wc-stripe-feature-flags.php
-
-		-
-			message: '#^Method WC_Stripe_Feature_Flags\:\:is_oc_available\(\) should return bool but returns string\.$#'
-			identifier: return.type
-			count: 1
+			message: '#^Unreachable statement \- code above always terminates\.$#'
```
daledupreez: I think we should work out what code path is causing these errors.

phpstan-baseline.neon
```diff
@@ -4704,6 +4698,12 @@ parameters:
 			count: 1
 			path: includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
 
+		-
+			message: '#^Callback expects 1\-2 parameters, \$accepted_args is set to 4\.$#'
```
daledupreez: This should be addressed as part of the filter setup.

daledupreez (changes requested):

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Good idea! Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/commits/b878efaf347cf8bd1440ebd91b8342b71abd3d60

phpstan-baseline.neon
wjrosa: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/commits/95b7a66a16ed766078f7a2715ff57a8b9d3c9504

phpstan-baseline.neon
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/changes/0df1c77e52e0f753442551d1896e1a1da539cb78#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efa

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/changes/0df1c77e52e0f753442551d1896e1a1da539cb78#diff-995edee38ad4f8387e58ebd52c31bcc04c56cc2448d331b1cf5e0b35c57b9efa

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1112,25 +1112,88 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
...
-	 * @param WC_Order $order Order data.
-	 * @param bool     $sent_to_admin Whether the email is being sent to admin or customer.
+	 * @param WC_Order        $order         Order data.
+	 * @param bool            $sent_to_admin Whether the email is being sent to admin or customer.
+	 * @param bool            $plain_text    Whether the email is plain text (no HTML).
+	 * @param WC_Email|null   $email         The email object (unused, accepted for hook compatibility).
 	 * @return void
 	 */
-	public function add_email_currency_conversion_notice( WC_Order $order, bool $sent_to_admin = false ): void {
+	public function add_email_currency_conversion_notice( WC_Order $order, bool $sent_to_admin = false, bool $plain_text = false, $email = null ): void {
```
daledupreez: Not blocking, but we don't need to accept the email arg if we don't use it -- we can simply pass 3 instead of 4 for the `add_filter()` command for the email handling.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1112,25 +1112,88 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
...
+		);
+
+		$inline_style = sprintf(
+			'margin-top: 1em; border: solid 1px %s; border-radius: %s; background-color: %s; padding: 1em 2em;',
+			esc_attr( $styles['border-color'] ?? '#007CBA' ),
+			esc_attr( $styles['border-radius'] ?? '4px' ),
+			esc_attr( $styles['background-color'] ?? '#F6F5F8' )
+		);
+
+		echo '<div style="' . esc_attr( $inline_style ) . '">';
```
daledupreez, line 1196: We're double-escaping the values here because we are calling `esc_attr()` for each value and then calling `esc_attr()` for the `style` attribute's value. We should remove the `esc_attr()` calls for each value from the filter.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -4324,4 +4324,127 @@ public function test_add_email_currency_conversion_notice_outputs_notice_with_co
 		$this->assertStringContainsString( $expected_rate . ' EUR', $output );
 		$this->assertStringContainsString( '</div>', $output );
 	}
+
+	/**
+	 * Test that the wc_stripe_adaptive_pricing_email_notice_styles filter allows customising the notice colours.
```
daledupreez, line 4324: Not blocking, but there is a lot of repetition across these tests.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1104,52 +1093,134 @@ public function add_converted_currency_information( string $formatted_total, WC_
...
 			return;
 		}
 
-		$woocommerce_amount = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount(
-			$presentment_amount,
-			$presentment_currency
+		/**
+		 * Filters the inline styles applied to the adaptive pricing currency conversion notice in order emails.
+		 *
+		 * @since 9.5.0
```
daledupreez: Nit: This version should be bumped to 10.6.0.
```suggestion
		 * @since 10.6.0
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1104,52 +1093,134 @@ public function add_converted_currency_information( string $formatted_total, WC_
...
-		$presentment_currency_lower = strtolower( $presentment_currency );
-		$rate_decimals              = 2;
-		if ( in_array( $presentment_currency_lower, WC_Stripe_Helper::no_decimal_currencies(), true ) ) {
-			$rate_decimals = 0;
-		} elseif ( in_array( $presentment_currency_lower, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
-			$rate_decimals = 3;
-		}
-		$rate_amount = wc_format_decimal( $presentment_amount / $stripe_amount, $rate_decimals );
+		$inline_style = sprintf(
+			'margin-top: 1em; border: solid 1px %s; border-radius: %s; background-color: %s; padding: 1em 2em;',
```
daledupreez: Nit: I think we should add a `margin-bottom: 1em;` rule to ensure we have space after the box as well. This was a bit cramped in my email.

Image

daledupreez (approved):
I have some minor comments from code review, including some items we should address before merging, but I think this is looking good overall.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/changes/cd0212ec360cf6384293a649dea27d5a1d6b0092

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/changes/cd0212ec360cf6384293a649dea27d5a1d6b0092

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/changes/cd0212ec360cf6384293a649dea27d5a1d6b0092

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5122/changes/dc77bad1971012346ddc89dab0cf56dc3243ed72

PR #5125, opened 2026-03-13
Declutter and refine test card messaging

pierorocca:
## Description

Shortens the verbose test mode messaging on checkout, adds a "Test Mode" badge to Blocks checkout radio labels, and adds copy-to-clipboard for test card/account numbers — matching the polish of WooPayments' implementation.

Fixes STRIPE-985

- **Shorter copy:** Each payment method's `get_testing_instructions()` now returns a single-line instruction instead of a full paragraph
  - CC: "**Test mode:** use card `4242 4242 4242 4242` with any expiry and CVC. More test cards."
  - SEPA: "**Test mode:** use account `AT611904300234573201`. More test methods."
  - BLIK: "**Test mode:** use any 6-digit number."
- **Copy-to-clipboard:** `<number>` tags in instructions are expanded into copy buttons via `expand_copy_button_markup()`. Clicking copies the number (without spaces) and shows a snackbar notification. Uses CSS `mask-image` SVG icons for copy/check states.
- **Test Mode badge:** Yellow badge appears on the Blocks checkout radio label when the payment method is selected (hidden by default, shown via CSS `input:checked` sibling selector).
- **All three checkout surfaces covered:**
  - Blocks (split PE): badge on each method radio + copy buttons
  - Blocks (OC): badge on "Stripe" radio + copy buttons in per-method instruction divs
  - Shortcode: bold "Test mode:" prefix + copy buttons (no badge — no radio labels)

## Testing instructions

1. Enable test mode in Stripe settings
2. Go to **Blocks checkout**:
   - Verify "Test Mode" badge appears next to the selected payment method label
   - Verify shorter test instructions appear below the payment element
   - Click the card number — verify it copies `4242424242424242` (no spaces) and shows a snackbar
3. Enable **Optimized Checkout** and verify:
   - Badge appears on the "Stripe" radio when selected
   - Switching sub-methods in the PE toggles the correct instruction div
   - Copy button works for each method's test number
4. Switch to **shortcode checkout**:
   - Verify "**Test mode:** use card..." with copy button (no badge)
5. Disable test mode — verify no badge, no instructions on any surface

### Stripe test cards

| Scenario | Card |
|----------|------|
| Success | 4242 4242 4242 4242 |
| SEPA | AT61 1904 3002 3457 3201 |

## Pre-review checklist

- [x] Added changelog entry
- [x] Added/updated tests (PHPUnit: instruction strings + `expand_copy_button_markup` data provider)
- [x] All PHP tests pass (1479 tests, 3488 assertions)
- [x] All JS tests pass (553 tests)
- [x] PHP lint, JS lint, CSS lint all pass

---
Pre-existing bug -> Test copy on OCS is covered when "Save my info" is checked.

### Before
image

### After

https://github.com/user-attachments/assets/93cbf6a6-f5b5-4ce3-baec-6b2254d750e0

https://github.com/user-attachments/assets/82521f52-8395-4ee2-8277-b39b53eefdee

https://github.com/user-attachments/assets/b9f1a6d1-dcfb-4e2b-be4a-df7c1ab6a374

https://github.com/user-attachments/assets/84c29943-c6e7-4cec-9f86-c97a854ab2c0

https://github.com/user-attachments/assets/46dc598e-9a55-4451-a392-fdafdc26b373

image

## Updated
Removed "Test mode:" from Blocks checkout as it's redundant and missed during testing.
image


client/blocks/upe/copy-test-number.js
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	event.preventDefault();
+
+	const number = copyNumberButton.querySelector( 'span' )?.innerText;
+	if ( ! number ) {
+		return;
+	}
+
+	navigator.clipboard.writeText( number.replace( /\s/g, '' ) );
```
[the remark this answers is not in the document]
pierorocca: Fixed. Wrapped `clipboard.writeText()` in `.then()/.catch()` so the success state and snackbar only appear after the promise resolves.

client/stripe-utils/copy-test-number.scss
```diff
@@ -0,0 +1,48 @@
...
+	i {
+		order: 1;
+		display: block;
+		width: 1.2em;
+		height: 1.2em;
+		mask-image: url('./icons/copy.svg?asset');
+		mask-size: contain;
+		mask-repeat: no-repeat;
+		mask-position: center;
+		background-color: currentColor;
```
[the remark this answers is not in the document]
pierorocca: No change. The project's Stylelint config does not enforce `value-keyword-case` — CI linting passes with `currentColor`. Both casings are valid CSS; `currentColor` is the original spec spelling.

client/blocks/upe/copy-test-number.scss
```diff
@@ -0,0 +1,48 @@
...
+		opacity: 0.7;
+	}
+
+	&:active i {
+		transform: scale(0.9);
+	}
+
+	&:focus {
+		outline: none;
+	}
```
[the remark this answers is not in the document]
pierorocca: Fixed. Replaced `outline: none` on `&:focus` with a visible `outline: 2px solid currentColor` on `&:focus-visible`, so keyboard users see a focus ring while mouse clicks remain clean.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -350,6 +350,25 @@ public static function get_testing_instructions_for_optimized_checkout() {
...
+	 */
+	public static function expand_copy_button_markup( $instructions ) {
+		$copy_label = esc_attr__( 'Copy to clipboard', 'woocommerce-gateway-stripe' );
+
+		return preg_replace(
+			'/<number>(.*?)<\/number>/',
+			'<button type="button" class="js-wc-stripe-copy-test-number" aria-label="' . $copy_label . '" title="' . $copy_label . '"><i></i><span>$1</span></button>',
+			$instructions
+		);
+	}
```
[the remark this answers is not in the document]
pierorocca: Fixed. Added `?? $instructions` fallback after `preg_replace()` — resolves the PHPStan failure.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -3847,4 +3847,42 @@ public function provide_test_is_valid_optimized_checkout_page() {
...
+			'string without number tag' => [
+				'input'    => '<strong>Test mode:</strong> use any 6-digit number.',
+				'expected' => '<strong>Test mode:</strong> use any 6-digit number.',
+			],
+			'empty string' => [
+				'input'    => '',
+				'expected' => '',
+			],
+		];
+	}
```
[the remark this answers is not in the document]
pierorocca: Fixed. Added a `string with multiple number tags` case to the data provider.

client/stripe-utils/__tests__/copy-test-number.test.js
```diff
@@ -0,0 +1,134 @@
...
+	it( 'works when clicking a child element inside the button', () => {
+		createButton( '4000056655665556' );
+		const icon = document.querySelector(
+			'.js-wc-stripe-copy-test-number i'
+		);
+		click( icon );
+
+		expect( writeText ).toHaveBeenCalledWith( '4000056655665556' );
+	} );
+} );
```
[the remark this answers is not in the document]
pierorocca: Fixed. Added a test case that sets `navigator.clipboard` to `undefined` and asserts no throw and no success state. Also added a null-safe guard (`navigator.clipboard?.writeText`) in the source to handle non-HTTPS contexts.

client/stripe-utils/copy-test-number.js
```diff
@@ -0,0 +1,47 @@
...
+
+	event.preventDefault();
+
+	const number = copyNumberButton.querySelector( 'span' )?.innerText;
+	if ( ! number ) {
+		return;
+	}
+
+	navigator.clipboard
+		.writeText( number.replace( /\s/g, '' ) )
```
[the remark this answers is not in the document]
pierorocca: Fixed. Added `if ( ! navigator.clipboard?.writeText ) { return; }` guard before the `writeText` call. This prevents synchronous throws in non-HTTPS or unsupported contexts, before the promise `.catch()` handler.

client/blocks/upe/styles.scss
```diff
@@ -35,3 +35,26 @@ button.stripe-gateway-stripelink-modal-trigger:hover {
...
+}
+
+.wc-block-checkout__payment-method
+	.wc-block-components-radio-control-accordion-option
+	input:checked
+	~ div
+	.wc-block-components-radio-control__label
+	.wc-stripe-test-mode-badge {
+	display: inline-block;
+}
```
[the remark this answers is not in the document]
pierorocca: Fixed. Added a matching selector for `.wc-block-components-radio-control__option` alongside the existing accordion variant. The badge now shows when checked in both radio option structures.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -917,31 +936,50 @@ public function payment_fields() {
...
 							],
+							'button' => [
+								'type'       => [],
+								'class'      => [],
+								'aria-label' => [],
+								'title'      => [],
+							],
+							'i'      => [],
+							'span'   => [],
 						]
```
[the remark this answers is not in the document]
pierorocca: Fixed. Extracted shared allowed tags into `$allowed_tags` variable, reused across both OC and non-OC branches. OC branch merges in the additional `div` tag via `array_merge()`.

client/blocks/upe/copy-test-number.js
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	navigator.clipboard
+		.writeText( number.replace( /\s/g, '' ) )
+		.then( () => {
+			window.wp?.data
+				?.dispatch( 'core/notices' )
+				?.createInfoNotice(
+					__(
+						'Test number copied to clipboard!',
```
daledupreez: Not blocking: Maybe we can simplify this to "Copied to clipboard!", as what we are copying should be pretty clear, and it will be more general for any value we want to make copyable. But that may be unnecessary.

client/blocks/upe/copy-test-number.js
```diff
@@ -0,0 +1,51 @@
+import { __ } from '@wordpress/i18n';
+import './copy-test-number.scss';
+
+/**
+ * Event-delegation handler for copy-to-clipboard buttons in test mode instructions.
+ * Copies the test card/account number and shows a snackbar notification.
+ */
+document.addEventListener( 'click', function ( event ) {
+	const copyNumberButton = event.target?.closest(
+		'.js-wc-stripe-copy-test-number'
```
daledupreez: Nit: We generally use `wc-stripe-` as our prefix for everything. It would be better to keep to that, even for this fairly specific use case.
```suggestion
		'.wc-stripe-copy-test-number'
```

client/blocks/upe/copy-test-number.scss
```diff
@@ -0,0 +1,54 @@
...
+	border: none !important;
+	background-color: transparent !important;
+	font-weight: normal;
+	display: inline-flex;
+	cursor: pointer;
+	color: inherit !important;
+	font-size: inherit;
+	padding: 2px 1px !important;
+	align-items: center;
+	box-shadow: none !important;
```
daledupreez: The `!important` rules here are something of a red flag for me. We should add an explicit comment as to why these values need the flag, and maybe group them so it's easier to see which items they actually apply to. I know from context that this single level class selector will generally get lose out in terms of selectivity by other layered classes, and we don't know where in our DOM the elements will end up. But that needs to be explicit for future readers.
```suggestion
	font-weight: normal;
	display: inline-flex;
	cursor: pointer;
	font-size: inherit;
	align-items: center;
	/* Important are needed because this selector has low selectivity and we don't know where in the page/DOM we will be. */
	border: none !important;
	background-color: transparent !important;
	color: inherit !important;
	padding: 2px 1px !important;
	box-shadow: none !important;
```

includes/payment-methods/class-wc-stripe-upe-payment-method-sepa.php
```diff
@@ -62,10 +62,12 @@ public function get_testing_instructions( $show_optimized_checkout_instruction =
...
-			esc_html__( '%1$sTest mode:%2$s use the test account number AT611904300234573201. Other payment methods may redirect to a Stripe test page to authorize payment. More test card numbers are listed %3$shere%4$s.', 'woocommerce-gateway-stripe' ),
+			/* translators: 1) HTML strong open tag 2) HTML strong closing tag 3) number open tag 4) number closing tag 5) HTML anchor open tag 6) HTML anchor closing tag */
+			esc_html__( '%1$sTest mode:%2$s use account %3$sAT611904300234573201%4$s. %5$sMore test methods%6$s.', 'woocommerce-gateway-stripe' ),
 			'<strong>',
 			'</strong>',
+			'<number>',
+			'</number>',
 			'<a href="https://docs.stripe.com/testing?payment-method=sepa-direct-debit#non-card-payments" target="_blank">',
 			'</a>'
 		);
```
daledupreez, line 73: It feels like it may be easier to have the replacement logic live closer to the instructions implementation so it's easier to see and replicate the right way of doing things.
```suggestion
		return sprintf(
			/* translators: 1) HTML strong open tag 2) HTML strong closing tag 3) number open tag 4) number closing tag 5) HTML anchor open tag 6) HTML anchor closing tag */
			esc_html__( '%1$sTest mode:%2$s use account %3$sAT611904300234573201%4$s. %5$sMore test methods%6$s.', 'woocommerce-gateway-stripe' ),
			'<strong>',
			'</strong>',
			$this->get_copy_button_tag(),
			'</span></button>',
			'<a href="https://docs.stripe.com/testing?payment-method=sepa-direct-debit#non-card-payments" target="_blank">',
			'</a>'
		);
	}

protected function get_copy_button_tag(): string {
	$copy_label = esc_attr__( 'Copy to clipboard', 'woocommerce-gateway-stripe' );
	
	return '<button type="button" class="js-wc-stripe-copy-test-number" aria-label="' . $copy_label . '" title="' . $copy_label . '"><i></i><span>';
}
```

daledupreez (commented):
I have some initial feedback based on inspection -- it's looking good, but it feels like we may want to move the logic closer to the code where we are generating the instructions.

includes/payment-methods/class-wc-stripe-upe-payment-method-sepa.php
daledupreez: Not blocking, but I am open to discussion on this one.

client/blocks/upe/copy-test-number.js
pierorocca: Fixed. Simplified to "Copied to clipboard\!".

client/blocks/upe/copy-test-number.js
pierorocca: Fixed. Renamed to `wc-stripe-copy-test-number` across JS, SCSS, PHP, and tests.

client/blocks/upe/copy-test-number.scss
pierorocca: Fixed. Grouped the `\!important` overrides together with an explanatory comment about the low-specificity selector needing to win against deeply nested checkout/theme styles.

includes/payment-methods/class-wc-stripe-upe-payment-method-sepa.php
pierorocca: I'd lean toward keeping the current approach. The `<number>` placeholder tag separates content from presentation — each payment method only declares *what* the test number is, while `expand_copy_button_markup()` in the gateway owns *how* it renders. If the button HTML changes (attributes, structure, accessibility), only one method needs updating. Moving the markup into each subclass via `get_copy_button_tag()` would scatter the closing tags (`</span></button>`) across 20+ sprintf calls, which is fragile. Happy to discuss further if you see downsides to the current split.

pierorocca: @daledupreez Claude here is disagreeing. I'll defer to you to resolve this one as I'm only able to get 90% of the way there before I'm out of my element.

elizaan36:
Really nice enhancement @pierorocca !
I don't think the icon should change to a checkmark when copied. Could we keep the icon consistent and present a tooltip or snackbar to indicate "Copied to clipboard"?

Also maybe for a separate issue. Is there anything we can do about the spacing here? The checkbox needs more top padding, this is potentially an accessibility issue.
Screenshot 2026-03-17 at 6 34 57 PM

elizaan36:
Looks like the spacing was fixed in [this issue](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5124
) actually. I added another question there about the double checkboxes.

pierorocca:
> I don't think the icon should change to a checkmark when copied. Could we keep the icon consistent and present a tooltip or snackbar to indicate "Copied to clipboard"?

This is copying the WooPayments behavior so would like to keep things consistent if possible. The snackbar is present on Blocks. I forgot it was there because it renders way out of my central vision on my screen. A tool-tip as an in-place affordance would be a good upgrade over the snack bar.

pierorocca:
If you don't mind @daledupreez adding @diegocurbelo as a 2nd reviewer since you'll be traveling the next few days.

diegocurbelo (approved):
The changes here look good to me overall.

Not a blocker (supporting all themes it's really hard, and it's only for test mode), but when using Storefront the number is not aligned:

Image

pierorocca:
> The changes here look good to me overall.
> 
> Not a blocker (supporting all themes it's really hard, and it's only for test mode), but when using Storefront the number is not aligned:
> 
> Image

Good catch! I tested classic checkout on Storefront but not block checkout. I'll compare against WooPayments as I hope this was solved there. Thanks @diegocurbelo .

pierorocca (commented):
Thanks for catching this @diegocurbelo! Fixed in c01fc822.

**Root cause:** Themes like Storefront set an explicit `font-family` on all `button` elements (e.g. `"Source Sans Pro"`), which overrides the parent paragraph's font. The copy button already used `!important` overrides for `border`, `background`, `color`, `padding`, and `box-shadow` — but was missing `font-family`, so the card number rendered in the theme's button font instead of the surrounding text font.

**Fix:** Added `font-family: inherit !important` to `.wc-stripe-copy-test-number`, consistent with the existing override pattern. Verified on Storefront (classic + blocks checkout) and Twenty Twenty-Four (blocks checkout).

### Storefront + classic checkout

image

### Storefront + block checkout

image

### Block theme and checkout
image

pierorocca:
image

Better but still not there. Working on one additional fix.

pierorocca:
**Update on the Storefront font mismatch fix (23099731):**

Three CSS issues in `.wc-stripe-copy-test-number` caused the test card number to look different from surrounding text on themes like Storefront:

1. **Missing `font-family: inherit`** — Storefront sets `font-family: "Source Sans Pro"` on all `button` elements, overriding the parent paragraph's font. Added `font-family: inherit !important` (consistent with existing `!important` overrides on the same selector).

2. **`vertical-align: middle` → `baseline`** — For an `inline-flex` element, `middle` aligns the element's midpoint to the parent's midpoint (not the text baseline), causing a subtle vertical shift. Changed to `baseline` so the button text sits on the same line as surrounding text.

3. **`line-height: 1.2em` → `inherit`** — The hardcoded `1.2em` differed from the paragraph's line-height (Storefront uses `1.618`, blocks themes vary), making the number appear slightly taller. Changed to `inherit` so it matches.

Net diff:
```diff
 .wc-stripe-copy-test-number {
-    line-height: 1.2em;
-    vertical-align: middle;
+    line-height: inherit;
+    vertical-align: baseline;
     ...
+    font-family: inherit !important;
 }
```

image

includes/payment-methods/class-wc-stripe-upe-payment-method-sepa.php
pierorocca: Going to mark this as resolved given this was 3 weeks ago and the implemented solution is less fragile.

client/classic/upe/index.js
```diff
@@ -9,6 +9,7 @@ import { initializeAlwaysExpandedOptimizedCheckout } from './always-expanded-opt
 import { legacyHashchangeHandler } from './legacy-support';
 import './style.scss';
 import './deferred-intent.js';
+import 'wcstripe/blocks/upe/copy-test-number';
```
malithsen: Having this defined in the blocks directory and then cross-importing to classic seems awkward. I think `wcstripe/stripe-utils/` will be a better home for this.

client/stripe-utils/copy-test-number.js
```diff
@@ -0,0 +1,48 @@
...
+			window.wp?.data
+				?.dispatch( 'core/notices' )
+				?.createInfoNotice(
+					__( 'Copied to clipboard!', 'woocommerce-gateway-stripe' ),
+					{
+						id: 'wc-stripe/test-number-copied',
+						type: 'snackbar',
+						context: 'wc/checkout/payments',
+					}
+				);
```
malithsen, line 44: Nit: As written, this is run on both blocks and classic, but the toast will only show up on the blocks checkout. We should consider extracting this out to a separate function and document that notice only appears on blocks. (or work on a separate mechanism for classic checkout)

malithsen (approved):
Looks good and works as described. Left two non-blocking comments.

client/classic/upe/index.js
pierorocca: Moved to `client/stripe-utils/copy-test-number.js` along with the SCSS, icons, and test file. Updated imports in both blocks and classic entry points.

client/stripe-utils/copy-test-number.js
pierorocca: Added a JSDoc note documenting that the snackbar only renders on Blocks checkout (where `core/notices` is connected to the DOM). On classic checkout the icon toggle (copy → check) still provides visual feedback. The optional chaining on `window.wp?.data?.dispatch` already handles the classic case gracefully.

PR #5135, opened 2026-03-16
Fix default UX for single payment elements

daledupreez:
Fixes STRIPE-989

In 10.5.x, we updated to the Clover version of the Stripe.js library. This update ended up having two side-effects in situations where Optimized Checkout Suite is not enabled:
 - Even when rendering only one payment method, we are getting an accordion layout that adds a payment icon and title wrapper
Screenshot 2026-03-16 at 12 49 43

 - For US merchants who also have Link enabled, the Card payment element includes a secondary Bank accordion item that results in the card element staying contracted by default after select the "Credit / Debit Card" option in checkout.
Screenshot 2026-03-16 at 12 46 42

_The UX is similar across classic and blocks checkout._

This PR addresses the above by ensuring that we specify `layout: 'tabs'` when creating a payment element with OCS disabled (as per [the Stripe docs](https://docs.stripe.com/js/elements_object/create_payment_element#payment_element_create-options-layout).

## Testing instructions

* Check out this branch
* Set up a connection to Stripe, ideally for a US address and for a store using USD
* Disable Optimized Checkout
* Ensure that multiple payment methods are enabled, including Link and Card
* Navigate to classic checkout
* Verify that the Card element (and all other payment methods) show up without an accordion wrapper
  - If you have a US store, verify that the Bank option for the Card element is available, but we expand the Card option by default
* Navigate to blocks checkout
* Verify that the Card element (and all other payment methods) show up without an accordion wrapper
* Enable Optimized Checkout Suite
* Verify that nothing has changed in both block checkout and classic checkout

Optionally, you can also verify the current UX by checking things against `develop`.

#### Card with no Bank option - block checkout

Screenshot 2026-03-16 at 15 01 49

#### Card with Bank option - blocks checkout

Screenshot 2026-03-16 at 14 59 53

#### Card with no Bank option - classic checkout

Screenshot 2026-03-16 at 15 01 23

#### Card with Bank option - classic checkout

Screenshot 2026-03-16 at 15 00 19

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa (approved):
Looks good to me! Thanks for adding specific unit tests!

malithsen (approved):
Thanks for the fix. Both classic and block checkout pages look good.

PR #5136, opened 2026-03-16
Implement manual approval hook for agentic checkout finalize_checkout

RadoslavGeorgiev:
Fixes STRIPE-908

Implements the `v1.delegated_checkout.finalize_checkout` manual approval hook. When Stripe sends this event before charging the customer, we validate each line item for product existence, purchasability, and stock availability, then respond with `approved` or `declined`.

- New `WC_Stripe_Agentic_Commerce_Manual_Approval` class handles validation logic
- Adds `get_quantity()` and `get_name()` to `WC_Stripe_Agentic_Customize_Checkout_Line_Item` (reused for both hooks)
- Reuses `WC_Stripe_Agentic_Customize_Checkout_Event` for event parsing
- Provides `wc_stripe_agentic_approve_order` filter for custom merchant approval/decline logic (return `null` to approve, a string to decline with that reason)
- Wires up the handler in `process_agentic_hook()`, replacing the placeholder

## Testing instructions

This hook is triggered by Stripe during agentic checkout. Trigger one through your dashboard for each scenario below, and make sure it passes.

| Result | Test |
| - | - |
| ✅ Pass | Stock not managed |
| ✅ Pass | Stock managed and sufficient |
| 🟥  Fail | Not in stock |
| 🟥  Fail | Switch the product to a draft, making it unpurchaseable. |

---

- [x] Covered with tests

### Changelog entry

- [x] This Pull Request does not require a changelog entry. (Comment required below)

<details>

#### Comment

Internal agentic commerce feature, not yet user-facing.

</details>


wjrosa (approved):
Changes look good to me 👍

includes/agentic-commerce/class-wc-stripe-agentic-commerce-manual-approval.php
```diff
@@ -0,0 +1,118 @@
...
+		/**
+		 * Filters the manual approval decision for an agentic checkout order.
+		 *
+		 * Return null to approve, or a non-empty string to decline with that reason.
+		 *
+		 * @since 10.6.0
+		 * @param string|null                                $decline_reason Null to approve, or a decline reason string.
+		 * @param WC_Stripe_Agentic_Customize_Checkout_Event $event          The finalize checkout event.
+		 */
+		$decline_reason = apply_filters( 'wc_stripe_agentic_approve_order', $decline_reason, $event );
```
daledupreez: Not blocking: might it be worth passing in the invalid line item into the filter so the implementer doesn't need to loop over the events to find the incorrect one?
```suggestion
		$decline_reason = apply_filters( 'wc_stripe_agentic_approve_order', $decline_reason, $event, $invalid_line_item );
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-manual-approval.php
```diff
@@ -0,0 +1,118 @@
...
+	 * @throws Exception When product resolution fails.
+	 */
+	private function validate_line_item( WC_Stripe_Agentic_Customize_Checkout_Line_Item $line_item ): ?string {
+		$product_id = (int) $line_item->get_sku_id();
+		$product    = WC_Stripe_Agentic_Commerce_Product_Resolver::resolve_product( $product_id );
+
+		if ( ! $product->is_purchasable() ) {
+			return sprintf(
+				/* translators: %s: product name */
+				__( '%s is not available for purchase.', 'woocommerce-gateway-stripe' ),
```
daledupreez: Not blocking: For internal use and for the filter, might it be worth using a code + description object rather than just a translated string? It feels like filtering or tweaking the wording would be much simpler if we returned a `not_purchasable` or `not_in_stock` code in addition to the reason.

daledupreez (approved):
This is looking good to me based on inspection. I did have two non-blocking comments, but no need to address them in this PR.

RadoslavGeorgiev:
Resolved the coderabbit feedback above as:
1. The type mismatch could be for older versions, but this has not been released.
2. It commented on tests that I had failed to `git add`
It's all good :)

PR #5138, opened 2026-03-16
Fix PHPStan issues after #5128

RadoslavGeorgiev:
#5128 changed how classes are autoloaded and this led to inconsistencies in the PHPStan baseline. This PR addresses them.

Here is Claude's explanation:

> **Root cause:** The rename commit (`c4e1f5671`) changed `autoload-dev` from PSR-4 to `classmap`, which made PHPStan discover the test stub classes (`WC_Subscriptions_Product`, `WC_Pre_Orders_Product`, etc.). This caused:
> 
> 1. **91 stale baseline entries** — old "class not found" / "function not found" suppressions that no longer match because the classes are now visible
> 2. **~119 new errors** — the stubs have incomplete method signatures, causing type mismatch errors (these are now baselined)
> 3. **6 inline `@phpstan-ignore` comments** in `class-wc-stripe-helper.php` that suppressed errors that no longer occur — removed
> 
> **Changes on `develop`:**
> - `phpstan-baseline.neon` — regenerated (old: stale suppressions, new: stub type mismatches)
> - `includes/class-wc-stripe-helper.php` — removed 6 stale `@phpstan-ignore` inline comments

Note that the stubs having incomplete signatures, causing type mismatches, is not ideal, so I pushed https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5138/commits/ac14449f103983e2d9780044ec22c08470b79fe3, fixing stubs and making the PR net negative in terms of changed lines.

### Changelog entry

<details>

#### Comment 

Internal changes, nothing to be released.

</details>

### DIFF
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -1219,9 +1219,9 @@ class WC_Stripe_Helper {
 			return true;
 		}
 
-		$subscriptions_available = class_exists( 'WC_Subscriptions_Product' ) && method_exists( 'WC_Subscriptions_Product', 'is_subscription' ); // @phpstan-ignore function.impossibleType
-		$pre_orders_available    = class_exists( 'WC_Pre_Orders_Product' ) && method_exists( 'WC_Pre_Orders_Product', 'product_is_charged_upon_release' ); // @phpstan-ignore function.impossibleType
-		$deposits_available      = class_exists( 'WC_Deposits_Product_Manager' ) && method_exists( 'WC_Deposits_Product_Manager', 'deposits_enabled' ); // @phpstan-ignore function.impossibleType
+		$subscriptions_available = class_exists( 'WC_Subscriptions_Product' ) && method_exists( 'WC_Subscriptions_Product', 'is_subscription' );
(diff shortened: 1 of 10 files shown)


daledupreez:
@wjrosa, I think we need to work out how to handle the changed and/or incorrect analyses that are being performed as a result of the autoloader changes. Maybe we should [exclude the test helper directory](https://getcomposer.org/doc/04-schema.md#exclude-files-from-classmaps) from the classmap?

wjrosa:
@daledupreez I think that should do it 👍

RadoslavGeorgiev:
@daledupreez @wjrosa I'm not sure what to do with the PR in this case 🤣 

Given that this PR unblocks checks (and local commits), with a net decrease in PHPStan entries, are we good?

> incorrect analyses that are being performed as a result of the autoloader changes

Is this about "~119 new errors — the stubs have incomplete method signatures, causing type mismatch errors (these are now baselined)"? The later commit fixed the method signatures, and I think we are in a good place?

wjrosa (approved):
IMHO we are good to go here :). Thanks for fixing the issues!

malithsen (approved):

PR #5146, opened 2026-03-17
Fix missing payment tokens when OCS is enabled

### LINKED ISSUE #5081: Non-card saved payment methods not shown when Optimized Checkout is enabled
When the Optimized Checkout Suite is enabled, and the customer has saved a non-card payment method (e.g., SEPA), that method is not available for selection in the checkout page (both block and shortcode/classic checkout).

A SEPA token is saved, but it is not shown below:

Image

wjrosa:
Fixes STRIPE-969
Fixes #5081

When OCS is enabled, all payment methods are rendered under the consolidated Stripe gateway element rather than their individual sub-gateway IDs. This caused saved tokens for methods like SEPA, Bancontact, ACH, etc., to be invisible at checkout
   — They were stored under sub-gateway IDs (e.g., stripe_sepa_debit), but only the main stripe gateway was queried.

Changes:
  - When OCS is enabled, and the stripe gateway is queried, sub-gateway tokens are merged into the result, so all saved methods appear at checkout.
  - When OCS is enabled, it skips the individual `is_enabled()` check per payment method, since all reusable methods are accessible through the OCS element regardless of their individual toggle state.
  - Refactors `woocommerce_get_customer_payment_tokens` into smaller, focused private methods (`sync_and_retrieve_customer_payment_tokens`, `categorize_existing_tokens`, `get_sub_gateway_ids`) for clarity.
  - Deprecates `customer_has_saved_methods()` with `wc_deprecated_function()`.
  - Adds PHPUnit tests covering OCS token merging, the enabled-state bypass, and the refactored token categorization logic.

## Testing instructions

- Checkout and build this branch on your test environment (`fix/missing-saved-tokens-when-oc-is-enabled`)
- Connect your Stripe account
- Enable the Optimized Checkout Suite feature
- Enable some non-card payment methods
- As a merchant, go to the "My account" page > "Add payment method"
- Confirm you are able to add a non-card payment method (i.e., SEPA)
- Now, add a product to your cart
- Go to the blocks checkout page
- Confirm the saved method is there
- Repeat the test for the classic checkout

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4008,6 +4009,39 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
...
+	 * @since 10.6.0
+	 * @return array
+	 */
+	public function get_tokens() {
+		$tokens = parent::get_tokens();
+		if ( ! is_user_logged_in() ) {
+			return $tokens;
+		}
+
+		if ( $this->oc_enabled ) {
```
wjrosa: This is the main fix for the classic checkout.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -126,8 +128,12 @@ public static function get_token_from_request( array $request ) {
 	 * @since 4.1.0
 	 * @param int $customer_id
 	 * @return bool
+	 *
+	 * @deprecated 10.6.0 - This method is deprecated and will be removed in a future version.
 	 */
 	public static function customer_has_saved_methods( $customer_id ) {
```
wjrosa, line 148: The methods deprecated here have not been used for a while. I decided to use the opportunity to flag them.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
 	 *
-	 * @return array
+	 * @return WC_Payment_Token[]
+	 */
+	public function woocommerce_get_customer_payment_tokens( $tokens, $customer_id, $gateway_id ) {
+		if ( ! is_user_logged_in() ) {
+			return $tokens;
+		}
+
+		return $this->sync_and_retrieve_customer_payment_tokens( $tokens, $customer_id, $gateway_id );
```
wjrosa, line 185: I created this intermediary method just to make it clearer what we are actually doing. We are not just retrieving the tokens here

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+	/**
+	 * Syncs the payment tokens between Stripe and WooCommerce and returns the updated tokens list.
+	 *
+	 * @param array  $tokens      Existing tokens.
+	 * @param int    $customer_id WC customer ID.
+	 * @param string $gateway_id  WC Gateway ID.
+	 *
+	 * @return WC_Payment_Token[] Array of payment tokens after syncing with Stripe.
+	 */
+	private function sync_and_retrieve_customer_payment_tokens( array $tokens, int $customer_id, string $gateway_id ): array {
```
wjrosa, line 197: I tried to make this method a bit easier to read, but I am not sure I was able to. It is still very confusing.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -848,6 +1008,85 @@ public function get_account_saved_payment_methods_list_item_sepa( $item, $paymen
...
 	}
-}
 
-new WC_Stripe_Payment_Tokens();
+	/**
+	 * Returns the list of active reusable payment method types based on the enabled gateways and their settings.
+	 *
+	 * @return array List of active reusable payment method types.
+	 */
+	private function get_active_reusable_payment_method_types(): array {
```
wjrosa, line 1068: Created this helper method to isolate the logic

includes/class-wc-stripe.php
```diff
@@ -203,6 +203,8 @@ public function init() {
 		new WC_Stripe_Order_Handler();
 
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/payment-tokens/class-wc-stripe-payment-tokens.php';
+		new WC_Stripe_Payment_Tokens();
```
wjrosa, line 208: We forgot to externalize the instantiation for this one.

wjrosa:
@copilot Add unit tests for the changes on this PR

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4125,6 +4126,39 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
 		return $this->validate_field( $field_key, $value );
 	}
 
+	/**
+	 * Returns a users saved tokens for this gateway.
```
daledupreez: Nit: This should be `user's`. I can't unsee it! 😉 
```suggestion
	 * Returns a user's saved tokens for this gateway.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: I have two comments here:
1. We need to check whether we're on a valid OCS page before we do anything here.
2. This is less important, but this feels like a good candidate for an early return.
```suggestion
		if ( ! $this->oc_enabled || ! $this->is_valid_optimized_checkout_page ) {
			return $tokens
		}
```

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+		return $tokens;
+	}
+
+	/**
+	 * Splits the existing token list into stored (valid) and deprecated buckets,
+	 * ignoring tokens that don't belong to a reusable UPE gateway.
+	 *
+	 * @param WC_Payment_Token[] $tokens All existing tokens for the customer.
+	 * @return array{stored: array<string, WC_Payment_Token>, deprecated: array<string, WC_Payment_Token>}
+	 */
```
daledupreez, line 319: The PHPDoc doesn't specify that the function only returns Stripe tokens -- other tokens are ignored and are not returned.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -848,6 +1008,85 @@ public function get_account_saved_payment_methods_list_item_sepa( $item, $paymen
...
+		}
+
+		// Remove the APM tokens from before Split PE was in place.
+		foreach ( $deprecated_tokens as $token ) {
+			unset( $tokens[ $token->get_id() ] );
+			$token->delete();
+		}
+
+		// Re-add the action after cleanup is done.
+		add_action( 'woocommerce_payment_token_deleted', [ $this, 'woocommerce_payment_token_deleted' ], 10, 2 );
```
daledupreez: It feels like it would be a good idea to use `try/finally` for these operations, just to make sure we're cleaning up correctly.
```suggestion
		try {
		// Remove the payment methods that no longer exist in Stripe's side.
			foreach ( $stored_tokens as $token ) {
				unset( $tokens[ $token->get_id() ] );
				$token->delete();
			}

			// Remove the APM tokens from before Split PE was in place.
			foreach ( $deprecated_tokens as $token ) {
				unset( $tokens[ $token->get_id() ] );
				$token->delete();
			}
		} finally {
			// Re-add the action after cleanup is done.
			add_action( 'woocommerce_payment_token_deleted', [ $this, 'woocommerce_payment_token_deleted' ], 10, 2 );
		}
```

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -848,6 +1008,85 @@ public function get_account_saved_payment_methods_list_item_sepa( $item, $paymen
...
+	 */
+	private function get_active_reusable_payment_method_types(): array {
+		$gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+
+		// Retrieve the payment methods for the enabled reusable gateways.
+		$reusable_payment_method_types = array_keys( self::UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD );
+
+		// When OCS is enabled, all reusable payment methods can be used via the consolidated OCS element.
+		// Return all reusable types so existing tokens are verified against Stripe and not incorrectly orphaned.
+		if ( $gateway->is_oc_enabled() ) {
```
daledupreez: We also need to check if we're in a context where Optimized Checkout can be enabled.
```suggestion
		if ( $gateway->is_oc_enabled() && $gateway->is_valid_optimized_checkout_page() ) {
```

daledupreez: Maybe we need a new `is_optimized_checkout_active()` helper? It feels like we are going to get this wrong if we don't have something like that to get the right information.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -420,6 +572,14 @@ public function get_account_saved_payment_methods_list_item( $item, $payment_tok
 			return $item;
 		}
 
+		// When OCS is enabled, all saved payment methods are processed through the single consolidated OCS element
+		// (gateway ID 'stripe'). Remap sub-gateway tokens (e.g. stripe_sepa_debit) to the main stripe gateway ID
+		// so that PaymentUtils includes them in the blocks checkout saved methods list.
+		$main_gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+		if ( $main_gateway->is_oc_enabled() && WC_Stripe_UPE_Payment_Gateway::ID !== $payment_token->get_gateway_id() ) {
```
daledupreez: Another location where we need to check whether OCS is enabled _and_ active.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+		}
+
+		$tokens = $this->cleanup_invalid_tokens( $tokens, $stored_tokens, $deprecated_tokens );
+
+		// When OCS is enabled and the main stripe gateway is queried (e.g. blocks checkout), also include
+		// tokens for sub-gateways (stripe_sepa_debit, stripe_bancontact, etc.) since all payment methods
+		// are rendered under the single 'stripe' gateway in that context.
+		// The filter is kept active so each sub-gateway call goes through the normal sync path.
+		// There is no recursion risk because the OCS block only runs when gateway_id === 'stripe'.
+		if ( $gateway->is_oc_enabled() && WC_Stripe_UPE_Payment_Gateway::ID === $gateway_id ) {
```
daledupreez: Another location for enabled + active.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+			foreach ( $this->get_sub_gateway_ids() as $sub_gateway_id ) {
+				$tokens = array_merge( $tokens, WC_Payment_Tokens::get_customer_tokens( $customer_id, $sub_gateway_id ) );
+			}
```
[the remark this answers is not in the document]
daledupreez: I may need to pick up my PR to switch the payment method data to the database cache -- it feels like some sites may end up with lots of API calls if transients are not retained. 😬

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+					continue;
+				}
+
+				// Retrieve the real APM behind SEPA PaymentMethods.
+				$payment_method_type = $this->get_original_payment_method_type( $payment_method );
+
+				// The corresponding method for the payment method type is not enabled, skipping.
+				// When OCS is enabled, allow all reusable methods regardless of their individual enabled state,
+				// since they are all accessible through the consolidated OCS element.
+				if ( ! $gateway->is_oc_enabled() && ! $gateway->payment_methods[ $payment_method_type ]->is_enabled() ) {
```
daledupreez: Another enabled + active check.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+					$this->is_valid_payment_method_id( $payment_method->id, $payment_method_type ) &&
+					( empty( $gateway_id ) || $this->is_valid_payment_method_type_for_gateway( $payment_method_type, $gateway_id ) )
+				) {
+					$token                      = $this->add_token_to_user( $payment_method, $customer, $payment_method_ids );
+					$tokens[ $token->get_id() ] = $token;
+				}
+			}
+
+			// Re-add the filter after we're done adding missing tokens to prevent unnecessary recursion.
+			add_filter( 'woocommerce_get_customer_payment_tokens', [ $this, 'woocommerce_get_customer_payment_tokens' ], 10, 3 );
```
daledupreez: It feels like the filter removal should happen first in the `try` block, and we should move the re-add to a `finally` block. We should be very confident that we restore things after the try block, especially considering that non-Stripe exceptions could be thrown.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+
+			if ( $this->is_token_deprecated( $token ) ) {
+				$deprecated[ $token->get_token() ] = $token;
+				continue;
+			}
+
+			$stored[ $token->get_token() ] = $token;
+		}
+
+		return compact( 'stored', 'deprecated' );
```
daledupreez: Personal preference: it feels like we could just as easily return an array manually. I don't like `compact`, and I prefer being explicit about what is going on. Definitely not blocking.
```suggestion
		return [
			'stored'     => $stored,
			'deprecated' => $deprecated,
		);
```

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+	}
+
+	/**
+	 * Returns the gateway IDs for all sub-gateways (every reusable gateway except the main 'stripe' gateway).
+	 *
+	 * @return string[]
 	 */
-	public function woocommerce_get_customer_payment_tokens( $tokens, $user_id, $gateway_id ) {
-		return $this->woocommerce_get_customer_upe_payment_tokens( $tokens, $user_id, $gateway_id );
+	private function get_sub_gateway_ids(): array {
```
daledupreez: Nit: The method name should explicitly mention "reusable" in some way, and not just have that be part of the PHPDoc. It would be easy to get this smaller set of IDs when just reading the method name.
```suggestion
	private function get_reusable_sub_gateway_ids(): array {
```

daledupreez (changes requested):
I have a good amount of small feedback, but I have a bigger question: should checkout work with the saved payment method? Because it doesn't at the moment, at least when using a non-card saved token. 😬 (I was testing with a Klarna token.)

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: Added the new helper with unit tests and implemented it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/39c4980ebfc854b41b7c508675a8630bacecc24d

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/8ab15453057d1f214868c4655408b77a39d17232

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: Moved both in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/65ca5b88c510ada8ae7cd4edcbe891f613139c41

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/e445838331553bb6d89429597f6f14158f4bcd1b

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/0cdfd7d3ea54c2807a82c51210a0fce44f0c7436

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/8579f34ce57f1e55c681ca1516a292abbb9f39a5

wjrosa:
Thanks for the review, @daledupreez ! I have addressed all your suggestions. And I just tested some transactions with a saved Klarna token in both the blocks and the classic checkout. It is working correctly for me. Which issue are you experiencing?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4023,6 +4024,39 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
...
+
+				// Already handled by the parent method.
+				if ( WC_Stripe_Payment_Methods::CARD === $stripe_id ) {
+					continue;
+				}
+
+				$gateway_id    = WC_Stripe_Payment_Tokens::UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD[ $stripe_id ];
+				$method_tokens = WC_Payment_Tokens::get_customer_tokens( get_current_user_id(), $gateway_id );
+				$tokens        = array_merge( $tokens, $method_tokens );
+			}
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/908d47707a3ea005d5b9dcbd73b5cada8496d3fd

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+	 * @param int    $customer_id WC customer ID.
+	 * @param string $gateway_id  WC Gateway ID.
+	 *
+	 * @return WC_Payment_Token[] Array of payment tokens after syncing with Stripe.
+	 */
+	private function sync_and_retrieve_customer_payment_tokens( array $tokens, int $customer_id, string $gateway_id ): array {
+		// Not a reusable payment gateway.
+		if ( empty( $gateway_id ) || ! in_array( $gateway_id, self::UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD, true ) ) {
+			return $tokens;
+		}
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/786462a6e2b7e79e878db2397e0a1b2de2eeeb9d

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +157,152 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+			}
+
+			// Re-add the filter after we're done adding missing tokens to prevent unnecessary recursion.
+			add_filter( 'woocommerce_get_customer_payment_tokens', [ $this, 'woocommerce_get_customer_payment_tokens' ], 10, 3 );
+		} catch ( WC_Stripe_Exception $e ) {
+			wc_add_notice( $e->getLocalizedMessage(), 'error' );
+			WC_Stripe_Logger::error( 'Error getting customer payment tokens (upe) for customer: ' . $customer_id, [ 'error_message' => $e->getMessage() ] );
+		}
+
+		$tokens = $this->cleanup_invalid_tokens( $tokens, $stored_tokens, $deprecated_tokens );
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/6555a5513939422c18a88a92d9b5629ed79e41d2

Mayisha (commented):
I got two issues during the functional testing.

1. Methods not enabled at checkout are displayed in the saved list.

Steps to reproduce:
- Set store currency to USD.
- Disable OCS.
- Enable card, ACH and ACH.
- As a shopper, add a card and an ACH payment method in **My Account**.
- Change store currency to Euro.
- Enable SEPA.
- As a shopper, add a SEPA payment method in **My Account**.
- Enable OCS.
- Notice that the saved payment method list has the card, ACH and SEPA methods though ACH is not supported for the currency Euro.

Image

2. Payment not working for non-card methods.

Steps to reproduce:
- Set store currency to USD.
- Disable OCS.
- Enable card, ACH and ACH.
- As a shopper, add a card and an ACH payment method in **My Account**.
- Select the saved ACH method for payment and click the checkout page.
- Checkout fails with the following error notice.

Image

wjrosa:
Good catch, @Mayisha ! I fixed the issues in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/changes/15bfceb8aca0d92c41b669fb515e257243a98774

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -4391,4 +4441,31 @@ public function test_add_email_currency_conversion_notice_outputs_plain_text_for
...
+		// guard were absent.
+		$this->mock_gateway->oc_enabled = true;
+
+		$tokens = $this->mock_gateway->get_tokens();
+
+		// For a guest session, parent::get_tokens() returns an empty array because
+		// WooCommerce's get_customer_tokens() only returns tokens for logged-in users.
+		$this->assertIsArray( $tokens );
+		$this->assertEmpty( $tokens, 'get_tokens() should return no tokens for a guest user even when OCS is enabled.' );
+	}
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/65425168d83831066817a21db7b8df877e48d109

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4197,6 +4206,43 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
...
+			if ( $this->id === $gateway_id ) {
+				continue;
+			}
+
+			$method_tokens = WC_Payment_Tokens::get_customer_tokens( get_current_user_id(), $gateway_id );
+			$tokens        = array_merge( $tokens, $method_tokens );
+		}
+
+		return array_unique( $tokens );
+	}
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/65425168d83831066817a21db7b8df877e48d109

phpstan-baseline.neon
```diff
@@ -10060,7 +10060,13 @@ parameters:
...
 			message: '#^Function remove_action invoked with 4 parameters, 2\-3 required\.$#'
 			identifier: arguments.count
-			count: 2
+			count: 3
+			path: includes/payment-tokens/class-wc-stripe-payment-tokens.php
+
+		-
+			message: '#^Function remove_filter invoked with 4 parameters, 2\-3 required\.$#'
+			identifier: arguments.count
+			count: 1
```
Mayisha: We should remove the 4th parameter from `remove_filter` instead of ignoring this.

Mayisha (commented):
Thanks for the fix @wjrosa. I did another round of functional testing and confirmed that the [second issue](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146#pullrequestreview-3993658381) is fixed and the first issue is fixed for block checkout. But in classic checkout, I still see the first issue. Could you please check if you can reproduce this?

> Methods not enabled at checkout are displayed in the saved list.
 

Image

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -780,6 +780,15 @@ public function is_valid_optimized_checkout_page(): bool {
 		return ! $this->is_on_add_payment_method_page() && ! $this->is_changing_payment_method_for_subscription();
 	}
 
+	/**
+	 * Checks if the optimized checkout is active based on the setting and whether we are on a valid page to show it.
+	 *
+	 * @return bool True if the optimized checkout is active, false otherwise.
+	 */
+	public function is_optimized_checkout_active(): bool {
```
daledupreez: 😍

daledupreez (commented):
I am getting a payment method that is inactive shown in both block and classic checkouts, _and_ I can use that payment method to make payments. In my case, I set up my store using USD, and used the add payment methods page to add Klarna and ACH as saved payment methods. I then made a purchase with each from block and classic checkout. I then disabled Klarna in the Stripe settings, but Klarna was still shown in both block and classic checkouts, and I was still able to make a payment with the saved token. (I did not change the store currency - it stayed as USD.)

wjrosa:
Thanks for the review, @daledupreez and @Mayisha ! I fixed the main issue with Claude in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/0566ea397fab0daf548341c00d7be9d9c22118eb

phpstan-baseline.neon
wjrosa: Removed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146/commits/82bd62aa67740aa7cd3bce827bcee35af02cb6c0

Mayisha (commented):
@wjrosa unfortunately I still see [this issue in classic checkout](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146#pullrequestreview-4030771383)

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -151,14 +171,189 @@ public static function customer_has_saved_methods( $customer_id ) {
...
+
+			$tokens = $this->cleanup_invalid_tokens( $tokens, $stored_tokens, $deprecated_tokens );
+
+			// When OCS is enabled and the main stripe gateway is queried (e.g. blocks checkout), also include
+			// tokens for sub-gateways (stripe_sepa_debit, stripe_bancontact, etc.) since all payment methods
+			// are rendered under the single 'stripe' gateway in that context.
+			// The filter has already been removed (above) so these calls read raw WooCommerce DB
+			// tokens without triggering the sync path. There is no recursion risk because the filter
+			// is absent for the duration of this try block.
+			// Only merge tokens for sub-gateways whose payment method is currently enabled.
```
Mayisha, line 281: Cursor review here:

Medium – Tokens for methods unavailable at checkout can still be surfaced under OCS

In the OCS merge path, sub-gateway tokens are included when is_enabled() is true, but not gated by is_enabled_at_checkout(). That means methods enabled in settings but unavailable for current checkout context (currency/capability constraints) can still appear as saved tokens.

This could be the reason behind unsupported methods appearing in classic checkout that I see [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5146#pullrequestreview-4067279894)

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4197,6 +4206,60 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
...
+			return $tokens;
+		}
+
+		// Track gateway IDs already represented in $tokens to avoid fetching the same sub-gateway twice.
+		$fetched_gateway_ids = [];
+		foreach ( $tokens as $token ) {
+			$fetched_gateway_ids[ $token->get_gateway_id() ] = true;
+		}
+
+		foreach ( $this->get_upe_enabled_payment_method_ids() as $stripe_id ) {
```
Mayisha, line 4432: Cursor review:

Medium – Tokens for methods unavailable at checkout can still be surfaced under OCS

In the OCS merge path, sub-gateway tokens are included when is_enabled() is true, but not gated by is_enabled_at_checkout(). That means methods enabled in settings but unavailable for current checkout context (currency/capability constraints) can still appear as saved tokens.

The same comment for this part also.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
Mayisha: Some more AI explanation with example

Trace for ACH in a EUR store (OCS active):                                                                                                                                                                   
                  
  // Main sync loop — line 241:                                                                                                                                                                                
  ! $method_obj->is_enabled_at_checkout()  →  true (ACH unsupported in EUR)                                                                                                                                    
  → unset( $stored_tokens['pm_ach_xxx'] )   // token won't be deleted                                                                                                                                          
  → unset( $tokens[ $excluded_token->get_id() ] )  // removed from result                                                                                                                                      
                                                                                                                                                                                                               
  // Sub-gateway merge — line 296:                                                                                                                                                                             
  ! $method_obj->is_enabled()  →  false (admin toggle is on)                                                                                                                                                   
  → $tokens = array_merge( $tokens, WC_Payment_Tokens::get_customer_tokens( $customer_id, 'stripe_us_bank_account' ) )                                                                                         
                                      // ← reads DB, finds the same ACH token, adds it back                                                                                                                    
                                                                                                                                                                                                               
  The ACH token that was correctly excluded by the currency check at line 241 is re-inserted at line 296. The final result shows a token that should be hidden.

wjrosa (commented):
Thanks for the follow-up, @Mayisha! I've pushed a fix in 55473a19.

**Root cause**: The sub-gateway token merge (which runs when OCS is enabled and the main `stripe` gateway is queried) was checking `is_enabled()` (the toggle) but not `is_enabled_at_checkout()` (currency/capability). So a method like ACH that is toggled on but unsupported for the current currency (e.g. EUR) had its stored DB tokens merged back in *after* the sync loop had already correctly excluded them.

**Fix**:
- Added `is_enabled_at_checkout()` to the sub-gateway merge guard in `sync_and_retrieve_customer_payment_tokens`
- Added the same check to the `get_tokens()` sub-gateway loop in `WC_Stripe_UPE_Payment_Gateway` (defense-in-depth for classic checkout)
- Added a PHPUnit test covering the enabled-but-unavailable-at-checkout path

Could you please test the classic checkout again with the latest commit?

wjrosa (commented):
(wrong comment by Claude deleted)

diegocurbelo (approved):
The changes look good to me, and it tests as described!

I also verified that inactive payment methods are no longer shown in the checkout (both block and classic).

Mayisha (approved):
Looks good to me. I only see the saved token for available payment methods on both classic and block checkout pages. Checkout with saved method is also working 👍

PR #5148, opened 2026-03-18
Prevent modification of express checkout payment methods on upgrades

daledupreez:

This PR addresses two possible causes of unexpected re-enablement of express checkout methods on upgrades.

The underlying issues are:
* When we are upgrading and we detect that setting synchronization is disabled (via the `pmc_enabled` sub-setting being set to `'no'`, we remove the `pmc_enabled` sub-setting to see if the new version can successfully enable settings synchronisation.
* The logic that attempts to enable settings synchronization mostly lives in `WC_Stripe_Payment_Method_Configurations::maybe_migrate_payment_methods_from_db_to_pmc()`, which includes the following block of code that auto-enables express checkout methods:

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/e2087af6ace561dd0dcda02b23ec5b76cb2f6827/includes/class-wc-stripe-payment-method-configurations.php#L594-L605

* The logic that replicates the PMC configuration into the local data was skipping express checkout payment methods, so previously enabled express checkout methods could end up disabled.

This PR sets a new `skip_pmc_express_checkout_defaults` sub-setting during the upgrade process when `pmc_enabled` is reset so that we continue to perform the existing settings synchronization checks for new installs or sites where we haven't attempted to (re)enable settings synchronization, but we don't default any express checkout methods on.

The PR also ensures that when we pick a PMC, any express checkout payment methods that were enabled in the PMC are kept enabled.

## Testing instructions

* Install version 10.5.1 or older on a test site - note that there were infinite loop issues in 10.4.0 and 10.5.0 to be wary of
* Make some changes to the express checkout payment methods, e.g. to disable either Apple Pay/Google Pay or Amazon Pay (note that Amazon Pay is only available in 10.4.0 and newer)
* Ensure that the `pmc_enabled` sub-setting is set to `no`, e.g. `wp option patch update woocommerce_stripe_settings pmc_enabled no`
* Upload and activate a plugin zip built using the code from this branch
* Confirm that all disabled express checkout payment methods remain disabled, and all enabled express checkout payment methods remain enabled
* Confirm that the `skip_pmc_express_checkout_defaults` sub-setting is `'yes'`: `wp option pluck woocommerce_stripe_settings skip_pmc_express_checkout_defaults`
* Check the Stripe log under _WooCommerce_ > _Status_ > _Logs_ and verify that you see an error recorded with the following text: `Settings synchronization eligibility will be re-checked after upgrade`

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -592,7 +592,7 @@ class WC_Stripe_Payment_Method_Configurations {
 		}
 
 		// Add default express checkout methods to the list if express checkout is enabled
-		if ( ! empty( $stripe_settings['express_checkout'] ) && 'yes' === $stripe_settings['express_checkout'] ) {
+		if ( ! empty( $stripe_settings['express_checkout'] ) && 'yes' === $stripe_settings['express_checkout'] && 'yes' !== ( $stripe_settings['skip_pmc_express_checkout_migration'] ?? 'no' ) ) {
 			$enabled_payment_methods = array_merge(
 				$enabled_payment_methods,
 				[ WC_Stripe_Payment_Methods::GOOGLE_PAY, WC_Stripe_Payment_Methods::APPLE_PAY ]
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -618,9 +618,8 @@ class WC_Stripe_Payment_Method_Configurations {
 					$available_payment_method_ids[] = $payment_method_id;
 				}
 
-				// We want to also include payment methods enabled in the PMC, except for express payment methods.
+				// Add all payment methods enabled in the PMC that are not enabled locally.
 				if (
-					! in_array( $payment_method_id, WC_Stripe_Payment_Methods::EXPRESS_PAYMENT_METHODS, true ) &&
 					! in_array( $payment_method_id, $enabled_payment_methods, true ) &&
 					isset( $payment_method->display_preference->value ) && 'on' === $payment_method->display_preference->value
 				) {
--- a/includes/class-wc-stripe-payment-method-configurations.php
+++ b/includes/class-wc-stripe-payment-method-configurations.php
@@ -592,7 +592,11 @@ class WC_Stripe_Payment_Method_Configurations {
 		}
 
 		// Add default express checkout methods to the list if express checkout is enabled
-		if ( ! empty( $stripe_settings['express_checkout'] ) && 'yes' === $stripe_settings['express_checkout'] && 'yes' !== ( $stripe_settings['skip_pmc_express_checkout_migration'] ?? 'no' ) ) {
+		if (
+			! empty( $stripe_settings['express_checkout'] ) &&
+			'yes' === $stripe_settings['express_checkout'] &&
+			'yes' !== ( $stripe_settings['skip_pmc_express_checkout_migration'] ?? 'no' )
+		) {
 			$enabled_payment_methods = array_merge(
 				$enabled_payment_methods,
 				[ WC_Stripe_Payment_Methods::GOOGLE_PAY, WC_Stripe_Payment_Methods::APPLE_PAY ]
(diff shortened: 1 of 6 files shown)


includes/class-wc-stripe-payment-method-configurations.php
```diff
@@ -592,7 +592,7 @@ public static function maybe_migrate_payment_methods_from_db_to_pmc( $force_migr
 		}
 
 		// Add default express checkout methods to the list if express checkout is enabled
-		if ( ! empty( $stripe_settings['express_checkout'] ) && 'yes' === $stripe_settings['express_checkout'] ) {
+		if ( ! empty( $stripe_settings['express_checkout'] ) && 'yes' === $stripe_settings['express_checkout'] && 'yes' !== ( $stripe_settings['skip_pmc_express_checkout_migration'] ?? 'no' ) ) {
```
malithsen: nit: Given the length, I'd consider extracting this condition to a variable.

malithsen (approved):
Thanks for the fix. Looks good to me

Mayisha (commented):
@daledupreez could you check if you can reproduce the following scenario?

- Install an older version.
- Enable Google Pay, Link and keep Amazon Pay disabled.
- Disable sync `wp option patch update woocommerce_stripe_settings pmc_enabled no`
- Upload the build of this branch.
- Check that `sync` is now enabled
- ⚠️ Google Pay got disabled while Link is still enabled.

daledupreez:
@Mayisha, I'll test that now. Did you test with 10.4.0? Or some other version? (Just to check.)

includes/class-wc-stripe-payment-method-configurations.php
daledupreez: I pushed this into a multi-line condition earlier but forgot to push that up. 127364d2c85bc524d436ece7786b5f632114d483

Let me know if that looks more readable or if you think we should switch to a variable.

malithsen: Looks much better, thanks!

Mayisha:
> Did you test with 10.4.0? Or some other version?

In my whole testing flow, I first upgraded from 10.4.0 to 10.5.1 to reproduce the bug. Then I upgraded from 10.5.1 to this build to test the fix.

daledupreez:
@Mayisha, excellent catch! The problem stems from two facts:
 - Our PMC updates don't update the `upe_checkout_experience_accepted_payments`, so that remains as whatever stale value it was (in my case `card` and `link` were enabled)
 - Our PMC => local replication was explicitly skipping express checkout methods

I've addressed the second issue in 47dc4cccde63e6253dd010670aa1fb91cf5435c9.

tests/phpunit/admin/migrations/class-migrate-payment-methods-from-db-to-pmc-test.php
```diff
@@ -216,4 +216,99 @@ public function test_migration_executed_and_amazon_pay_enabled_correctly( ?bool
...
+		$this->mock_payment_method_configurations(
+			[ 'google_pay', 'link', 'apple_pay' ], // enabled in PMC
+			[ 'card', 'amazon_pay' ]               // disabled in PMC
+		);
+
+		// Apple Pay, Google Pay, and Link should be enabled; Amazon Pay should be disabled.
+		$this->expect_payment_method_configurations_update(
+			[ 'card', 'google_pay', 'link', 'apple_pay' ],
+			[ 'amazon_pay' ]
+		);
```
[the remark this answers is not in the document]
daledupreez: I renamed the flag so it refers to skipping defaults rather than the migration, as it only skipped the defaults. So I think this should be considered resolved, as the bypass should be limited to defaulting on of payment methods, not replication from the PMC.

wjrosa (approved):
Change looks good and works as expected:

> Confirm that all disabled express checkout payment methods remain disabled, and all enabled express checkout payment methods remain enabled

Image
(I disabled Amazon Pay)

> Confirm that the skip_pmc_express_checkout_defaults sub-setting is 'yes': wp option pluck woocommerce_stripe_settings skip_pmc_express_checkout_defaults

Image

> Check the Stripe log under WooCommerce > Status > Logs and verify that you see an error recorded with the following text: Settings synchronization eligibility will be re-checked after upgrade

Image

malithsen (approved):
Looks good, works as described.

PR #5151, opened 2026-03-19
Minor improvements to order received code

wjrosa:
See https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086#discussion_r2960566246

In this simple PR, I am addressing some suggestions I missed from Dale [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5086#discussion_r2960566246). Basically:
- Adding a specific and new unit test to check for the "My Account" page (`test_add_converted_currency_information_appends_converted_currency_info_on_my_orders_page`)
- Added a check for the `WC_Order` instance in `add_converted_currency_information`
- Added some lowercased currency variables to reduce duplication

I also added a new helper method (`get_currency_decimals`) to reduce duplication when checking for a currency's decimal places.

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -323,14 +323,7 @@ public function get_allowed_shipping_countries() {
...
 	public static function get_stripe_currency_decimals() {
-		$currency = strtolower( get_woocommerce_currency() );
-		if ( in_array( $currency, WC_Stripe_Helper::no_decimal_currencies(), true ) ) {
-			return 0;
-		} elseif ( in_array( $currency, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
-			return 3;
-		}
-
-		return 2;
+		return WC_Stripe_Helper::get_currency_decimals( get_woocommerce_currency() );
```
daledupreez, line 326: Not blocking, but it would be good to migrate callers over to the new function. But I think that should happen in follow-up PRs, _not this one_.

daledupreez (changes requested):
The AI issue looks like a legitimate bug we should fix. Things are looking OK otherwise.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1129,22 +1134,18 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
...
-		if ( in_array( $presentment_currency_lower, WC_Stripe_Helper::no_decimal_currencies(), true ) ) {
-			$rate_decimals = 0;
-		} elseif ( in_array( $presentment_currency_lower, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
-			$rate_decimals = 3;
-		}
-		$rate_amount = wc_format_decimal( $presentment_amount / $stripe_amount, $rate_decimals );
+		$presentment_currency       = strtolower( $presentment_currency ); // Make sure the original currency code is in lowercase.
+		$presentment_currency_upper = strtoupper( $presentment_currency );
+		$rate_decimals              = WC_Stripe_Helper::get_currency_decimals( $presentment_currency );
+		$rate_amount                = wc_format_decimal( $presentment_amount / $stripe_amount, $rate_decimals );
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5151/changes/cb3d733b894ccde213d37e16d2a7d7334b50b1b1

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1081,20 +1085,21 @@ public function add_converted_currency_information( string $formatted_total, WC_
...
 
-		$currency_symbol = get_woocommerce_currency_symbol( strtoupper( $presentment_currency ) );
+		$currency_symbol = get_woocommerce_currency_symbol( $presentment_currency_upper );
 		$amount          = WC_Stripe_Helper::get_woocommerce_amount_from_stripe_amount(
 			$presentment_amount,
 			$presentment_currency
 		);
 
-		return $formatted_total . ' (' . $currency_symbol . ' ' . $amount . ' ' . strtoupper( $presentment_currency ) . ')';
+		return $formatted_total . ' (' . $currency_symbol . ' ' . $amount . ' ' . $presentment_currency_upper . ')';
```
daledupreez, line 1086: Idea: it feels like it may be worth adding a filter for this in a follow-up PR. I am thinking something like the following.


```suggestion
		$converted_currency_information = $formatted_total . ' (' . $currency_symbol . ' ' . $amount . ' ' . $presentment_currency_upper . ')';
		/*
		 * TODO: PHPDoc
		 */
		return apply_filters( 'wc_stripe_adaptive_pricing_converted_amount( $converted_currency_information, $formatted_total, $currency_symbol, $amount, $presentment_currency_upper );
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1132,22 +1137,19 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
...
-			$rate_decimals = 0;
-		} elseif ( in_array( $presentment_currency_lower, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
-			$rate_decimals = 3;
-		}
-		$rate_amount = wc_format_decimal( $presentment_amount / $stripe_amount, $rate_decimals );
+		$presentment_currency       = strtolower( $presentment_currency ); // Make sure the original currency code is in lowercase.
+		$presentment_currency_upper = strtoupper( $presentment_currency );
+		$rate_decimals              = WC_Stripe_Helper::get_currency_decimals( $presentment_currency );
+		// Divide major-unit amounts so the rate is correct for currencies with different decimal exponents (e.g. JPY↔USD).
+		$rate_amount = wc_format_decimal( (float) $woocommerce_amount / $order_total, $rate_decimals );
```
daledupreez: While this prevents the mismatch issue, I am still not sure if this is the right way to display conversion rates. As an example, if the presentment currency is JPY (or any other 0-decimal currency), we would always round the conversion to the nearest whole digit, while the conversion rate will almost always be some non-round number. I think we should use at least two decimal digits, but three may be useful as a baseline, simply because it means we have a better representation of the conversion rate.

wjrosa: That's a good point, Dale. I addressed your comment with Claude in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5151/commits/3716a6a8c38d76094cd05b46172bd143785b3ad4

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1130,24 +1135,22 @@ public function add_currency_conversion_notice( WC_Order $order ): void {
...
-		} elseif ( in_array( $presentment_currency_lower, WC_Stripe_Helper::three_decimal_currencies(), true ) ) {
-			$rate_decimals = 3;
-		}
-		$rate_amount = wc_format_decimal( $presentment_amount / $stripe_amount, $rate_decimals );
+		$presentment_currency       = strtolower( $presentment_currency ); // Make sure the original currency code is in lowercase.
+		$presentment_currency_upper = strtoupper( $presentment_currency );
+		// Use at least 3 decimal places for the rate regardless of the presentment currency's own decimal
+		// exponent — a conversion rate is not a currency amount, and rounding e.g. 149.567 JPY/USD to 150
+		// gives a misleading representation.
+		$rate_decimals = max( WC_Stripe_Helper::get_currency_decimals( $presentment_currency ), 3 );
```
daledupreez: This feels like a no-op, because the result will always be 3, as the presentment currency will always have 0, 2, or 3 decimals.
```suggestion
		$rate_decimals = 3;
```

daledupreez (approved):
Changes are LGTM, but I think we can simply use 3 for the number of decimals in the exchange rate.

PR #5153, opened 2026-03-20
Fix payment intent using global currency instead of order currency

### LINKED ISSUE #4925: Incorrect currency in payment intent via AJAX in multicurrency stores
### Description

The [WC_Stripe_Intent_Controller::create_payment_intent](<https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/includes/class-wc-stripe-intent-controller.php#L382>) function could create a payment intent with an incorrect currency if the order's currency does not match the currency returned by the `get_woocommerce_currency` function.<br>This issue could happen on sites that use a multicurrency plugin.

### Steps to reproduce

1. Create an order with a specific currency (for example, USD)
2. Add the filter

```
add_filter(
    'woocommerce_currency',
    function( $value ) {
        return 'CAD';
     }
);
```

3. Call `WC_Stripe_Intent_Controller::create_payment_intent` to create a payment intent for the order of step 1. The payment intent is created with an incorrect currency.

### Suggested Fix

Use the` order->get_currency()` function to get the currency instead of `get_woocommerce_currency` function in the WC_Stripe_Intent_Controller::create_payment_intent:

```
public function create_payment_intent( $order_id = null, $payment_method_type = null ) {
    $amount   = WC()->cart->get_total( false );
    $currency = get_woocommerce_currency();
    $order    = wc_get_order( $order_id );
    if ( is_a( $order, 'WC_Order' ) ) {
    		$amount   = $order->get_total();
    		$currency = $order->get_currency();
    }
...
    
```

### Additional notes

The issue relates to support tickets **10617845** and **10692120** on [WooCommerce.com](<http://WooCommerce.com>).

zhongruige:
Fixes STRIPE-883
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4925

When `create_payment_intent` is called with an order ID, the payment intent was always created using the global WooCommerce currency (`get_woocommerce_currency()`) rather than the order's currency (`$order->get_currency()`). This causes incorrect charges when a multicurrency plugin changes the global currency but the order was placed in a different currency.

This PR moves the currency resolution so that:
* When an order is provided, both the amount and the currency are taken from the order.
* When no order is provided (cart-only flow), the global WooCommerce currency is used as before.

## Testing instructions

1. Install and activate a multicurrency plugin (e.g., WooCommerce Multicurrency or WPML).
2. Configure the store's base currency to USD and add a secondary currency (e.g., CAD or EUR).
3. Switch the storefront currency to the secondary currency and add a product to the cart.
4. Proceed to checkout using a Stripe payment method (card, Klarna, etc.).
5. Verify that the payment intent is created with the correct (order) currency — not the store's base currency.
6. Repeat on both classic checkout and Blocks checkout.
7. Confirm that a normal single-currency checkout (no multicurrency plugin) still works as expected.

---

### Changelog entry

<details>

#### Comment

Fix - Use the order currency instead of the global store currency when creating a payment intent, resolving incorrect charges in multicurrency setups.

</details>

**Post merge**


zhongruige:
Hi @daledupreez! I took a look at the no-order-ID path (mentioned in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4925#issuecomment-3738941278)) and I believe it's already safe. When no order exists, both `WC()->cart->get_total()` and `get_woocommerce_currency()` are driven by the same multicurrency filter during the customer's session, so the amount and currency stay in sync. Though definitely let me know if I may have completely overlooked something!

daledupreez (approved):
The changes look good to me based on inspection, but I've requested another Quark reviewer as I won't get to manual testing this week.

wjrosa (approved):
Thanks for fixing this issue! Code is good and works as expected:

Image

I have used the [CURCY](https://br.wordpress.org/plugins/woo-multi-currency/) plugin.

PR #5170, opened 2026-03-25
Fix: put subscription on hold when Stripe Radar blocks a renewal payment

### LINKED ISSUE #5169: plugin is creating multiple new payment intents in rapid succession even after Radar blocks the first attempt, leading to an elevated block rate.
### Description

The WooCommerce Stripe Gateway plugin (by Stripe) is creating multiple new payment intents in rapid succession even after Radar blocks the first attempt, leading to an elevated block rate.

The plugin is:

Creating a new payment intent for each retry instead of reusing the existing one
Not checking the outcome of the previous attempt before retrying
Not stopping retries when Radar blocks a payment with outcome.type: blocked
Retrying the same payment method that was blocked, rather than prompting for an alternative
### Steps to reproduce

It happens on subset of renewal payments from WooCommerce subscription, using SEPA/direct debit. 
cannot share customer details here... 

### Expected behavior

The plugin should:

Check the outcome object after each payment attempt
When outcome.type is blocked and outcome.reason is high_risk_score, stop automatic retries
Display a clear message to the customer explaining the payment was blocked
Prompt the customer to use a different payment method
Use proper idempotency keys to prevent duplicate payment intent creation during genuine retries

### System report

<details>

`
### WordPress Environment ###

WC Version: 10.6.1
Legacy REST API Package Version: De Legacy REST API plugin is niet geïnstalleerd op deze site.
Log Directory Writable: ✔
WP Version: 6.9.4
WP Multisite: –
WP Memory Limit: 512 MB
WP Debug Mode: –
WP Cron: ✔
WP Environment Type: production
Language: nl_NL
External object cache: ✔

Version: 10.5.2
Account ID: acct_1S6VXv1MGOho8ede
Test Mode Enabled: ✔
					Nee
OAuth Connected: ✔
					Ja
Sync Enabled: ✔
					Ja
Optimized Checkout Enabled: Nee
Enabled Payment Methods: apple_pay,card,google_pay,ideal
Express Checkout: ✔
					Ingeschakeld (checkout)
Auth and Capture: ✔
					Ja
Logging: ✔
					Ja

### Abonnementen ###

WCS_DEBUG: ✔ Nee
Subscriptions Mode: ✔ Live
Subscriptions Live URL: https://medyne.nl
Subscriptions-core Library Version: 8.3.0
Subscription Statuses: trash: 1
wc-active: 9
wc-cancelled: 12
wc-on-hold: 8
wc-pending: 5
wc-pending-cancel: 1

WooCommerce Account Connected: ❌ No
Custom Retry Rules: ✔ No
Custom Retry Rule Class: ✔ No
Custom Raw Retry Rule: ✔ No
Custom Retry Rule: ✔ No
Retries Migration Status: ✔ Completed
Report Cache Enabled: ✔ Ja
Cache Update Failures: ✔ 0 fout

### Ondersteuning betaalmethode ###

:: products
refunds
tokenization
subscriptions
subscription_cancellation
subscription_suspension
subscription_reactivation
subscription_amount_changes
subscription_date_changes
subscription_payment_method_change
subscription_payment_method_change_customer
subscription_payment_method_change_admin
multiple_subscriptions

Stripe: products
refunds
tokenization
add_payment_method
subscriptions
subscription_cancellation
subscription_suspension
subscription_reactivation
subscription_amount_changes
subscription_date_changes
subscription_payment_method_change
subscription_payment_method_change_customer
subscription_payment_method_change_admin
multiple_subscriptions

</details>

wjrosa:
## Summary

Fixes #5169.

When Stripe Radar blocks a subscription renewal charge the plugin correctly skips the **in-request** retry (`card_declined` is non-retryable), but it only marks the renewal order as `failed`. WC Subscriptions then schedules its own retry attempts (e.g. after 1 day, 3 days, …). Each retry creates a brand-new Stripe charge that Radar blocks again, inflating the block rate and generating misleading fraud signals.

- Adds `is_charge_blocked_by_radar( $response )` to `WC_Stripe_Subscriptions_Trait`. After a `card_declined` error it fetches the underlying Stripe charge (`WC_Stripe_API::retrieve`) and checks `outcome.type === 'blocked'`.
- When a Radar block is detected, the parent subscription(s) are put **on hold** (instead of staying active) so WC Subscriptions does not schedule further retries against a payment method that Radar has already flagged as high risk.
- The renewal order itself is still correctly marked `failed` — only the _subscription_ status changes to prevent the retry loop.
- `is_charge_blocked_by_radar()` lives in the trait (not the abstract gateway) because the trait is used by both `WC_Stripe_UPE_Payment_Gateway` and individual payment method classes (e.g. SEPA) that do not inherit from the abstract gateway.

## Test plan

- [ ] Automated: `npm run test:php -- --filter=WC_Stripe_Subscription_Renewal_Test` — all 5 tests pass, including the new `test_renewal_radar_blocked_puts_subscription_on_hold`
- [ ] Manual: trigger a SEPA or card subscription renewal against a Stripe test-mode Radar block rule; confirm the renewal order is `failed` and the parent subscription moves to `on-hold`
- [ ] Confirm a normal card-declined renewal (issuer decline, not Radar) still leaves the subscription active so WC Subscriptions can retry normally


includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -373,6 +373,41 @@ public function scheduled_subscription_payment( $amount_to_charge, $renewal_orde
...
+		}
+
+		$charge = WC_Stripe_API::retrieve( "charges/{$charge_id}" );
+
+		if ( empty( $charge ) || ! empty( $charge->error ) ) {
+			return false;
+		}
+
+		return isset( $charge->outcome->type ) && 'blocked' === $charge->outcome->type;
+	}
```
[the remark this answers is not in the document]
Mayisha: Good idea. We should either return the error message from this method or update the order note to be generic instead of mentioning specifically `high risk`

changelog.txt
```diff
@@ -21,6 +21,7 @@
 * Fix - Improve performance of CSS style lookups
 * Add - Initial implementation of always-expanded Optimized Checkout Suite in shortcode checkout
 * Dev - Collapse PHPUnit tests using data providers to reduce duplication and improve test isolation
+* Fix - Put subscription on hold when Stripe Radar blocks a renewal payment to prevent WC Subscriptions from scheduling further retries that would also be blocked
```
Mayisha: We can drop the last part to make it concise.
```suggestion
* Fix - Put subscription on hold when Stripe Radar blocks a renewal payment to prevent WC Subscriptions from scheduling further retry attempts
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -373,6 +373,41 @@ public function scheduled_subscription_payment( $amount_to_charge, $renewal_orde
...
+		} elseif ( ! empty( $response->error->payment_intent->latest_charge ) && is_string( $response->error->payment_intent->latest_charge ) ) {
+			// Fallback: Payment Intents API may surface the charge ID via latest_charge.
+			$charge_id = $response->error->payment_intent->latest_charge;
+		}
+
+		if ( empty( $charge_id ) ) {
+			return false;
+		}
+
+		$charge = WC_Stripe_API::retrieve( "charges/{$charge_id}" );
```
Mayisha, line 403: Should we wrap this in try-catch?

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -373,6 +373,41 @@ public function scheduled_subscription_payment( $amount_to_charge, $renewal_orde
...
+
+		if ( empty( $charge ) || ! empty( $charge->error ) ) {
+			return false;
+		}
```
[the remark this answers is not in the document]
Mayisha: 👍

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -451,8 +486,9 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
 			// It's only a failed payment if it's an error and it's not of the type 'authentication_required'.
 			// If it's 'authentication_required', then we should email the user and ask them to authenticate.
 			if ( ! empty( $response->error ) && ! $is_authentication_required ) {
-				// We want to retry.
-				if ( $this->is_retryable_error( $response->error ) ) {
+				// We want to retry — unless Stripe Radar blocked the charge, in which case retrying
+				// would just create another blocked charge and inflate the block rate.
+				if ( $this->is_retryable_error( $response->error ) && ! $this->is_charge_blocked_by_radar( $response ) ) {
```
Mayisha: `is_charge_blocked_by_radar` calls the Stripe API to retrieve the charge object. We call this method here and also in the catch block. How about we check it here, keep a flag/variable and throw an early error. In catch block, we check the flag/variable instead of calling the method again?

Mayisha (commented):
Left a couple of suggestions from code review and agreed with a couple of AI feedback.

diegocurbelo (approved):
Looks good to me.

I agree with @Mayisha's comment about the duplicated `is_charge_blocked_by_radar()` call, we should use a variable, but it's non blocking and we can iterate on that and supporting different block reasons (as suggested by rabbit) on future PRs.

```php
// Before the retry/non-retry branching:
$is_radar_blocked = $this->is_charge_blocked_by_radar( $response );
```

includes/compat/trait-wc-stripe-subscriptions.php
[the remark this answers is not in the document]
wjrosa: Makes sense. Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5170/changes/30cb7662d130ba1fb38f1da9ffa94b7e54133430

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -510,6 +545,21 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+			// would create a new charge that Radar would block again, inflating the block rate.
+			if ( isset( $response ) && ! empty( $response->error ) && $this->is_charge_blocked_by_radar( $response ) ) {
+				$subscriptions = function_exists( 'wcs_get_subscriptions_for_renewal_order' )
+					? wcs_get_subscriptions_for_renewal_order( $renewal_order )
+					: [];
+				foreach ( $subscriptions as $subscription ) {
+					$subscription->update_status(
+						OrderStatus::ON_HOLD,
+						__( 'Stripe Radar blocked this payment as high risk. The subscription has been put on hold to prevent further blocked payment attempts.', 'woocommerce-gateway-stripe' )
+					);
```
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5170/commits/7b4e7842df7172e30d39b4b58825f5e7b3fd8106

includes/compat/trait-wc-stripe-subscriptions.php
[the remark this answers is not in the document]
wjrosa: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5170/changes/093bf18b6ffc076908915c52a0b7feb9d18f76f2

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: No try-catch needed here — `WC_Stripe_API::retrieve()` never throws. On failure it returns `null` (when the invalid-API-key rate limit is exceeded) or a `WP_Error` (on network/HTTP error). The `is_wp_error( $charge ) || empty( $charge )` guard added in 093bf18 already covers both cases before any property access.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Good idea! Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5170/changes/70d797d18dc65081ea5cd137d9675cc344139a55

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -510,6 +555,43 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
 
 			$renewal_order->update_status( OrderStatus::FAILED );
 
+			// If the payment was blocked by Stripe Radar, suspend the parent subscription(s)
+			// so that WC Subscriptions does not schedule further retry attempts. Each retry
+			// would create a new charge that Radar would block again, inflating the block rate.
+			if ( false !== $radar_reason ) {
+				switch ( $radar_reason ) {
+					case 'rule':
+						$radar_note = __( 'Stripe Radar blocked this payment due to a custom Radar rule. The subscription has been put on hold to prevent further blocked payment attempts.', 'woocommerce-gateway-stripe' );
```
Mayisha, line 564: Nit: We are adding an order note that says the subcription has been put on hold. But the subscriptions update happens later in the code which may fail (considering the catch block below). 
We can either change the wording here and say 'subscription will be put on-hold' or add the order status later after subscription update.

Mayisha (approved):
Thanks for fixing the issue @wjrosa. Looks good to me 💯

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: That's fair. Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5170/commits/90deeb0d43ee284a2caa3e7c74017ff493c3ec55

PR #5175, opened 2026-03-26
Fix UPE style transition keys for font smoothing properties to prevent warning

### LINKED ISSUE #5155: CSS logic for font smoothing needs to be updated
### Description

In testing various things today, I noticed console warnings when rendering Stripe in development:
> [Stripe.js] stripe.elements(): "webkitFontSmoothing" is not a supported property for ".Text"

The warnings applied to `.Text`, `.Input`, `.Text--redirect`, and `.Input--invalid`. Per the [Stripe appearance API documentation](https://docs.stripe.com/elements/appearance-api?api-integration=paymentintents&q=accordion#supported-css-properties), only `-webkit-font-smoothing` and `-moz-osx-font-smoothing` are supported.

This is a bug introduced by #5113 earlier this week, as that code changed to use the camel case property names.

### Steps to reproduce

1. Enable Stripe on a site and create a test Stripe connection
2. Enable the Stripe payment method
3. Add a product to your cart
4. Open your browser console
5. Navigate to checkout
6. You should see the warnings appear if your theme specifies either of those CSS rules

### Expected behavior

There should be no warnings related to CSS properties.

Mayisha:
Fixes STRIPE-STRIPE-996
Fixes #5155 

This PR reverts the commit https://github.com/woocommerce/woocommerce-gateway-stripe/commit/961d18aaa32cc931868d823f9ad6c6c3d6cec03b added in PR https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5155 where we changed the keys `MozOsxFontSmoothing` and `webkitFontSmoothing` in camelCase which is not supported by the Stripe element. 

## Testing instructions
- Enable Stripe on a site and connect to a Stripe account.
- Enable some Stripe payment methods
- Open your browser console
- Navigate to the checkout page
- In `develop`, you should see the following warnings appear if your theme specifies either of those CSS rules.

- In this branch, these warnings should not be there.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (approved):
LFTM 👍🏼 

**BEFORE**
Image

**AFTER**
No warnings in the console

daledupreez (commented):
@Mayisha, I think we need to map the properties into the appearance fields rather than use the old property names. The issue is that direct access via `computedStyles[ dashedPropertyName ]` doesn't get the related styles from the original object.

daledupreez:
Or, expressed differently, we need to make sure that all font smoothing properties from the computed styles are correctly being set into the appearance fields.

wjrosa (approved):
Looks good to me 👍 . Feel free to request my review again if Dale's comment requires a rework

Mayisha:
> I think we need to map the properties into the appearance fields rather than use the old property names. The issue is that direct access via computedStyles[ dashedPropertyName ] doesn't get the related styles from the original object.

@daledupreez I have updated this PR. Could you please check and confirm if this is the solution you proposed?

client/styles/upe/index.js
```diff
@@ -315,6 +319,14 @@ export const getFieldStyles = ( selector, upeElement ) => {
...
+		if ( fontSmoothingProps.includes( property ) ) {
+			const mappedProperty = computedStylePropertyMaps[ property ];
+			if ( typeof styles[ mappedProperty ] !== 'undefined' ) {
+				filteredStyles[ property ] = styles[ mappedProperty ];
+			}
+			continue;
+		}
+
 		if ( typeof styles[ property ] !== 'undefined' ) {
 			filteredStyles[ property ] = styles[ property ];
```
daledupreez: While this gets the job done, I am not a fan of the implementation as it splits the mapping constants and the processing logic and it has two code paths that do the same thing. I think this could be cleaner if we implemented a function like `getSourcePropertyName( propertyName )` in `upe-styles.js` and then use a single code path in this logic? (We could then export only the function, drop `fontSmoothingProps`, and keep computedStylePropertyMaps local.)

```js
export const getSourcePropertyName = ( propertyName )  => {
	return computedStylePropertyMaps[ propertyName ] || propertyName;
};
```


```suggestion
		const sourceProperty = getSourcePropertyName( property );

		if ( typeof styles[ sourceProperty ] !== 'undefined' ) {
			filteredStyles[ property ] = styles[ sourceProperty ];
```

client/styles/upe/upe-styles.js
```diff
@@ -8,6 +8,16 @@ const paddingColorProps = [
...
 ];
 
+export const fontSmoothingProps = [
+	'-webkit-font-smoothing',
+	'-moz-osx-font-smoothing',
+];
+
+export const computedStylePropertyMaps = {
+	'-webkit-font-smoothing': 'WebkitFontSmoothing',
+	'-moz-osx-font-smoothing': 'MozOsxFontSmoothing',
```
daledupreez, line 13: Good catch on `WebkitFontSmoothing` and `webkitFontSmoothing` both being valid.

client/styles/upe/index.js
Mayisha: Thanks for explaining in detail. I have updated the code as per this suggestion.

Mayisha:
> Looks good to me 👍 . Feel free to request my review again if Dale's comment requires a rework

@wjrosa requested you review again.

wjrosa (approved):
Latest changes still looks good to me 👍

PR #5177, opened 2026-03-26
Stripe 139 mandate updated webhook

anant1811:
Fixes  https://linear.app/a8c/issue/STRIPE-139/add-order-notes-when-mandate-for-recurring-payments-is-cancelled-or

 ## Summary

  Handle `mandate.updated` Stripe webhooks for Indian recurring payments and other mandate-based payment methods. When a customer cancels
  or pauses a mandate (e.g., via SMS notification), the order and subscription statuses are updated accordingly, and order notes are added
   for merchant visibility.

  ### Changes proposed in this Pull Request:

  * Add `get_order_by_mandate_id()` to `WC_Stripe_Order_Helper` with HPOS-compatible dual-path query (custom orders table + legacy
  postmeta)
  * Add `mandate.updated` case to webhook dispatcher in `process_webhook()`
  * Implement `process_webhook_mandate_updated()` handler with:
    - Mandate state mapping: `inactive` + revocation → cancelled, `inactive` → on-hold, `active` → reactivate subscription (only if
  on-hold)
    - Order notes for all mandate state changes
    - Idempotency guards against webhook redelivery
    - Subscription status updates when WooCommerce Subscriptions is active
    - Input sanitization for revocation reason
  * Fix pre-existing missing `break;` after `checkout.session.completed` case
  * Add 10 PHPUnit test cases covering all mandate states, edge cases, and idempotency

  Fixes https://linear.app/a8c/issue/STRIPE-139/add-order-notes-when-mandate-for-recurring-payments-is-cancelled-or

  ## How can this code break?

  - If the Stripe `mandate.updated` webhook payload structure differs from documented (mitigated by null checks and early returns on
  malformed payloads)
  - On legacy (non-HPOS) stores if the meta query path has issues (mitigated by following the established dual-path pattern from existing
  `get_order_by_*` methods)

  ## What are we doing to make sure this code doesn't break?

  - 10 PHPUnit test cases: `npm run test:php -- --filter=test_process_webhook_mandate_updated`
  - Follows established webhook handler patterns (dispute handler, charge handler)
  - HPOS compatibility via the same dual-path approach used by all existing order lookups


anant1811:
Just adding that all Code Rabbit suggestions have been implemented now.

zhongruige (changes requested):
Hi @anant1811!

There's a few things that need to be adjusted here:
- The pipeline has a merge conflict, can you look into resolving it?
- There's a few linter checks as well that need some updating
- There's two markdown files included in `docs/superpowers` which I don't think we want included in here either, could you look into removing them?

Thanks! Let me know if you have any questions, especially around the pipeline failures.

anant1811:
Green on the previously-failing checks after the two follow-up commits:

- **Check for changelog updates** ✅ — added entries in `changelog.txt` and `readme.txt` under the 10.6.0 section (using the correct per-file format: `YYYY-MM-DD - version X.Y.Z` in changelog, `= X.Y.Z - YYYY-MM-DD =` in readme).
- **Ensure milestone is assigned** ✅ — set to 10.6.0.
- **PHP linting** ✅ — PHPCBF auto-fixed the 6 array-alignment errors in the new test provider entry.
- **PHPStan Analysis** ✅ — added `@return void` to `process_webhook_mandate_updated()`, tightened `get_order_by_mandate_id()` with `is_array()` + `instanceof WC_Order` guards (the method now honors its declared `?WC_Order` return type), and removed two stale baseline patterns for `object::$payment_method_details` / `$payment_method_type` on the webhook handler since the new mandate code guards both with `isset()`.

Also merged in `develop` (f20bb484) to resolve the DIRTY state that had been preventing GitHub Actions from creating check suites at all on the previous two pushes. E2E and compat matrix jobs still running — no failures so far.

anant1811:
Thanks @zhongruige — addressed all three in commits b7e93ae8 (superpowers cleanup), f20bb484 (develop merge to clear the conflict), and 582a8010 (lint + phpstan fixes):

- Merge conflict with `develop` resolved.
- Lint (PHPCS alignment, PHPStan types + return annotations, stale baseline entries) now green.
- Removed `docs/superpowers/plans/` and `docs/superpowers/specs/` — those were brainstorming/plan artefacts that shouldn't have landed in the plugin tree.

daledupreez:
@coderabbitai resume

includes/class-wc-stripe-order-helper.php
```diff
@@ -697,6 +699,62 @@ public function update_stripe_mandate_id( ?WC_Order $order = null, string $manda
 		return $this->update_order_meta( $order, self::META_STRIPE_MANDATE_ID, $mandate_id );
 	}
 
+	/**
+	 * Gets the most recent non-terminal order by Stripe mandate ID.
+	 *
+	 * @since 10.2.0
```
Mayisha:
```suggestion
	 * @since 10.9.0
```

includes/class-wc-stripe-order-helper.php
```diff
@@ -697,6 +699,62 @@ public function update_stripe_mandate_id( ?WC_Order $order = null, string $manda
...
+			return null;
+		}
+
+		global $wpdb;
+
+		if ( WC_Stripe_Woo_Compat_Utils::is_custom_orders_table_enabled() ) {
+			$orders = wc_get_orders(
+				[
+					'limit'      => 1,
+					'status'     => [ OrderStatus::PENDING, OrderStatus::PROCESSING, OrderStatus::ON_HOLD ],
```
Mayisha: If we are considering `Processing` status here, we should also consider the `Completed` status. For digital products, the order transitions to the completed status instead of processing.

includes/class-wc-stripe-order-helper.php
```diff
@@ -697,6 +699,62 @@ public function update_stripe_mandate_id( ?WC_Order $order = null, string $manda
...
+			if ( is_array( $orders ) && ! empty( $orders ) && $orders[0] instanceof WC_Order ) {
+				return $orders[0];
+			}
+
+			return null;
+		}
+
+		$order_id = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT DISTINCT ID FROM $wpdb->posts as posts LEFT JOIN $wpdb->postmeta as meta ON posts.ID = meta.post_id WHERE meta.meta_value = %s AND meta.meta_key = %s AND posts.post_status IN ('wc-pending', 'wc-processing', 'wc-on-hold') ORDER BY posts.post_date DESC LIMIT 1",
```
Mayisha:
```suggestion
				"SELECT DISTINCT ID FROM $wpdb->posts as posts LEFT JOIN $wpdb->postmeta as meta ON posts.ID = meta.post_id WHERE meta.meta_value = %s AND meta.meta_key = %s AND posts.post_status IN ('wc-pending', 'wc-processing', 'wc-completed', 'wc-on-hold') ORDER BY posts.post_date DESC LIMIT 1",
```

Similarly, if we are considering the processing status we should consider completed as well.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
 	}
 
+	/**
+	 * Process webhook for mandate status updates.
+	 *
+	 * Handles mandate cancellation (revocation), pause (inactive), and
+	 * reactivation (active) for Indian recurring payments and other
+	 * mandate-based payment methods.
+	 *
+	 * @since 10.2.0
```
Mayisha:
```suggestion
	 * @since 10.9.0
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
+
+		if ( ! isset( $mandate->id, $mandate->status ) ) {
+			WC_Stripe_Logger::warning( 'mandate.updated webhook received with missing mandate ID or status.' );
+			return;
+		}
+
+		$mandate_id     = $mandate->id;
+		$mandate_status = $mandate->status;
+
+		WC_Stripe_Logger::info(
```
Mayisha: Nit. Can we move this log after the order check ( `if ( ! $order )` ) so that we log this only when there is no chance of bailing after the log.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
+		);
+	}
+
+	/**
+	 * Extracts the revocation reason from a mandate's payment_method_details.
+	 *
+	 * The revocation reason is nested under payment_method_details.{type}
+	 * where {type} varies by payment method (e.g., card, au_becs_debit, acss_debit).
+	 *
+	 * @since 10.2.0
```
Mayisha:
```suggestion
	 * @since 10.9.0
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1913,6 +2112,10 @@ public function process_webhook( $request_body ) {
 			case 'checkout.session.async_payment_failed':
 				$checkout_session_deferred = $this->process_checkout_session_failure( $notification );
 				break;
+
+			case 'mandate.updated':
```
Mayisha, line 2939: `mandate.updated` is never registered, so Stripe won't deliver it

We need to add this event to `WC_Stripe_Account::WEBHOOK_EVENTS` (class-wc-stripe-account.php)

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
+			return sanitize_text_field( $mandate->payment_method_details->$type->revocation_reason );
+		}
+
+		return null;
+	}
+
+	/**
+	 * Generates the order note message for a mandate status change.
+	 *
+	 * @since 10.2.0
```
Mayisha:
```suggestion
	 * @since 10.9.0
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
+	 * @param object $mandate The Stripe mandate object.
+	 * @return string|null The revocation reason, or null if not revoked.
+	 */
+	private function get_mandate_revocation_reason( $mandate ): ?string {
+		if ( ! isset( $mandate->payment_method_details, $mandate->payment_method_type ) ) {
+			return null;
+		}
+
+		$type = $mandate->payment_method_type;
```
Mayisha: As per the [documentation](https://docs.stripe.com/api/mandates/object) there is no `payment_method_type` field in the mandate object. 
This line and the check above will always return null.

Mayisha: The `type` can be found in `$mandate->payment_method_details->type`

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
+			$order->update_status( $target_order_status, $note );
+		} else {
+			$order->add_order_note( $note );
+			$order->save();
+		}
+
+		// Update subscription status if applicable.
+		$this->update_subscription_for_mandate( $order, $mandate_status, $revocation_reason );
+
+		WC_Stripe_Logger::info(
```
Mayisha, line 751: Nit: We can add the additional data in the context array in most of the logging where applicable.

```
WC_Stripe_Logger::info(
      'Mandate update processed.',
      [
          'mandate_id'          => $mandate_id,
          'order_id'            => $order->get_id(),
          'mandate_status'      => $mandate_status,
          'payment_method_type' => $mandate->payment_method_details->type ?? 'unknown',
      ]
  );
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
+		);
+	}
+
+	/**
+	 * Updates the subscription status based on a mandate status change.
+	 *
+	 * Only updates if WooCommerce Subscriptions is active and the order
+	 * has associated subscriptions.
+	 *
+	 * @since 10.2.0
```
Mayisha:
```suggestion
	 * @since 10.9.0
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
+
+		$subscriptions = wcs_get_subscriptions_for_order( $order );
+
+		if ( empty( $subscriptions ) ) {
+			return;
+		}
+
+		foreach ( $subscriptions as $subscription ) {
+			if ( 'inactive' === $mandate_status && $revocation_reason ) {
+				$subscription->update_status( 'cancelled', __( 'Subscription cancelled due to mandate revocation.', 'woocommerce-gateway-stripe' ) );
```
Mayisha: I think we should not cancel the whole subscription even if there is a revocation reason for this mandate. The presence of a revocation reason does not mean that the customer wants to cancel this subscription.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,205 @@ public function process_webhook_dispute_closed( $notification ) {
...
+		if ( $target_order_status && $order->has_status( $target_order_status ) ) {
+			WC_Stripe_Logger::info(
+				sprintf( 'Order %d already has status %s, skipping mandate update.', $order->get_id(), $target_order_status )
+			);
+			return;
+		}
+
+		// For non-actionable states (active, pending), check if the last order note already
+		// matches this mandate status to avoid duplicate notes on webhook redelivery.
+		if ( ! $target_order_status ) {
```
Mayisha: Is this note check necessary? If a similar webhook was actually received before the order status is supposed to be in the `target_order_status` and this note check seems redundant.
I would recommend we drop this check.

Mayisha (changes requested):
Thanks @anant1811 for adding the new webhook event for mandate updates. I have a few suggestions from code review.

includes/class-wc-stripe-webhook-handler.php
anant1811: Good catch — fixed in 5b1b2c46. Added `mandate.updated` to `WC_Stripe_Account::WEBHOOK_EVENTS` so Stripe actually delivers the event.

includes/class-wc-stripe-webhook-handler.php
anant1811: You're right — fixed in 5b1b2c46.\n\n`get_mandate_revocation_reason()` now reads the payment method type from `payment_method_details->type` and the reason from `payment_method_details->{type}->revocation_reason`. Confirmed against the [Stripe mandate object docs](https://docs.stripe.com/api/mandates/object) (there is no top-level `payment_method_type`, so the previous check always returned null).

includes/class-wc-stripe-webhook-handler.php
anant1811: Agreed — changed in 5b1b2c46. An inactive mandate (whether paused or revoked) now puts the subscription **on hold** instead of cancelling it; the revocation only affects the note wording so the merchant can follow up. Also added a terminal-state guard so already-cancelled/expired subscriptions are left untouched, plus a try/catch around `update_status()`.

includes/class-wc-stripe-webhook-handler.php
anant1811: Dropped in 5b1b2c46 — idempotency now relies solely on the `has_status()` guard.

One tradeoff worth flagging: a revocation webhook arriving *after* the order is already on-hold (e.g. a pause followed by a revoke) is now skipped, so the specific "revoked by the customer" note won't be added in that sequence. If you'd rather not lose that, I can track the last-processed mandate status in order meta instead — let me know your preference.

includes/class-wc-stripe-order-helper.php
anant1811: Added `OrderStatus::COMPLETED` to the HPOS query (and `wc-completed` to the legacy query) in 5b1b2c46.

includes/class-wc-stripe-order-helper.php
anant1811: Done in 5b1b2c46 — added `wc-completed` to the legacy query as suggested.

includes/class-wc-stripe-webhook-handler.php
anant1811: Moved the processing log to after the `if ( ! $order )` bail in 5b1b2c46.

includes/class-wc-stripe-webhook-handler.php
anant1811: Done in 5b1b2c46 — the processing and completion logs now use structured context arrays.

includes/class-wc-stripe-order-helper.php
anant1811: Bumped to `@since 10.9.0` in 5b1b2c46.

includes/class-wc-stripe-webhook-handler.php
anant1811: Bumped to `@since 10.9.0` in 5b1b2c46.

includes/class-wc-stripe-webhook-handler.php
anant1811: Bumped to `@since 10.9.0` in 5b1b2c46.

includes/class-wc-stripe-webhook-handler.php
anant1811: Bumped to `@since 10.9.0` in 5b1b2c46.

includes/class-wc-stripe-webhook-handler.php
anant1811: Bumped to `@since 10.9.0` in 5b1b2c46.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,204 @@ public function process_webhook_dispute_closed( $notification ) {
...
+					'limit'    => 1,
+				]
+			);
+			$note_preview   = $this->get_mandate_order_note( $mandate_id, $mandate_status, $revocation_reason );
+			if ( ! empty( $existing_notes ) && $existing_notes[0]->content === $note_preview ) {
+				WC_Stripe_Logger::info(
+					sprintf( 'Duplicate mandate.updated webhook for order %d with status %s, skipping.', $order->get_id(), $mandate_status )
+				);
+				return;
+			}
```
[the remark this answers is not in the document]
anant1811: Added the `: void` return type in 5b1b2c46.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,204 @@ public function process_webhook_dispute_closed( $notification ) {
...
+		}
+
+		foreach ( $subscriptions as $subscription ) {
+			if ( 'inactive' === $mandate_status && $revocation_reason ) {
+				$subscription->update_status( 'cancelled', __( 'Subscription cancelled due to mandate revocation.', 'woocommerce-gateway-stripe' ) );
+			} elseif ( 'inactive' === $mandate_status ) {
+				$subscription->update_status( 'on-hold', __( 'Subscription paused due to mandate becoming inactive.', 'woocommerce-gateway-stripe' ) );
+			} elseif ( 'active' === $mandate_status && $subscription->has_status( 'on-hold' ) ) {
+				$subscription->update_status( 'active', __( 'Subscription reactivated due to mandate becoming active.', 'woocommerce-gateway-stripe' ) );
+			}
```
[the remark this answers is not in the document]
anant1811: Done in 5b1b2c46 — added a terminal-state guard (skips `cancelled`/`expired`/`pending-cancel` subscriptions) and wrapped each `update_status()` call in a try/catch that logs failures, so one failing update can't leave the webhook half-applied.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -1378,4 +1378,226 @@ public function test_process_checkout_session_does_not_schedule_metadata_job_on_
...
+		$this->mock_webhook_handler->process_webhook_mandate_updated( $notification );
+
+		// The newer order should be updated to on-hold.
+		$final_new_order = wc_get_order( $new_order->get_id() );
+		$this->assertSame( OrderStatus::ON_HOLD, $final_new_order->get_status() );
+
+		// The older completed order should remain unchanged.
+		$final_old_order = wc_get_order( $old_order->get_id() );
+		$this->assertSame( OrderStatus::COMPLETED, $final_old_order->get_status() );
+	}
```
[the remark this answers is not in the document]
anant1811: Added subscription-sync coverage in 5b1b2c46: `test_process_webhook_mandate_updated_pauses_subscription` (data-driven for both paused and revoked mandates), `test_process_webhook_mandate_updated_reactivates_subscription`, and `test_process_webhook_mandate_updated_skips_terminal_subscription`.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,204 @@ public function process_webhook_dispute_closed( $notification ) {
...
+	 * Process webhook for mandate status updates.
+	 *
+	 * Handles mandate cancellation (revocation), pause (inactive), and
+	 * reactivation (active) for Indian recurring payments and other
+	 * mandate-based payment methods.
+	 *
+	 * @since 10.2.0
+	 * @param object $notification The webhook notification from Stripe.
+	 */
+	public function process_webhook_mandate_updated( $notification ) {
```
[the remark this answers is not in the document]
anant1811: Reworked the idempotency in 5b1b2c46. The fragile last-note comparison is gone; the handler now relies on the `has_status()` guard for the actionable (inactive → on-hold) path. Note that revocation no longer cancels — both paused and revoked mandates resolve to on-hold — which removes the status/note ambiguity this comment flagged. Happy to add per-mandate meta tracking if stronger idempotency is wanted.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,212 @@ public function process_webhook_dispute_closed( $notification ) {
...
+			// Never resurrect or mutate subscriptions that have already reached a terminal state.
+			if ( $subscription->has_status( [ 'cancelled', 'expired', 'pending-cancel' ] ) ) {
+				continue;
+			}
+
+			try {
+				if ( 'inactive' === $mandate_status && ! $subscription->has_status( 'on-hold' ) ) {
+					// An inactive mandate can no longer be charged, so the subscription is put on
+					// hold rather than cancelled: a revocation does not necessarily mean the customer
+					// wants to end the subscription, and the merchant can ask them to re-authorize.
```
Mayisha, line 864:
```suggestion
					// An inactive mandate can no longer be charged, so the subscription is put on hold
```
Nit: This is sufficient for code comment.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,212 @@ public function process_webhook_dispute_closed( $notification ) {
...
+					$note = $revocation_reason
+						? __( 'Subscription put on hold because the customer revoked the Stripe mandate.', 'woocommerce-gateway-stripe' )
```
Mayisha, line 866: Can we surely tell that the _customer_ has revoked the mandate? 🤔 

May be adding the generic note `Subscription put on hold because the Stripe mandate became inactive` for both cases would be better. And in case of the presence of a revocation reason, We can append the reason in the note.

includes/class-wc-stripe-order-helper.php
```diff
@@ -697,6 +699,62 @@ public function update_stripe_mandate_id( ?WC_Order $order = null, string $manda
...
+			$wpdb->prepare(
+				"SELECT DISTINCT ID FROM $wpdb->posts as posts LEFT JOIN $wpdb->postmeta as meta ON posts.ID = meta.post_id WHERE meta.meta_value = %s AND meta.meta_key = %s AND posts.post_status IN ('wc-pending', 'wc-processing', 'wc-completed', 'wc-on-hold') ORDER BY posts.post_date DESC LIMIT 1",
+				$mandate_id,
+				self::META_STRIPE_MANDATE_ID
+			)
+		);
+
+		if ( ! empty( $order_id ) ) {
+			$order = wc_get_order( $order_id );
+			if ( $order instanceof WC_Order && $order->get_status() !== OrderStatus::TRASH ) {
```
Mayisha, line 842: The check for Trash status is redundant here as we are retrieving the order from DB based on specific order statuses ('wc-pending', 'wc-processing', 'wc-completed', 'wc-on-hold')

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,212 @@ public function process_webhook_dispute_closed( $notification ) {
...
+		);
+
+		$this->resolved_order = $order;
+
+		// A revocation reason is only present when the customer has revoked the mandate.
+		$revocation_reason = $this->get_mandate_revocation_reason( $mandate );
+
+		// An inactive mandate (whether paused or revoked) can no longer be charged, so the
+		// order is put on hold. A revocation does not cancel the order outright — it only
+		// changes the wording of the note so the merchant can follow up with the customer.
```
Mayisha, line 725:
```suggestion
		// An inactive mandate (whether paused or revoked) can no longer be charged, so the
		// order is put on hold.
```
Nit: This is sufficient as code comment.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,212 @@ public function process_webhook_dispute_closed( $notification ) {
...
+	 * @param WC_Order    $order             The WooCommerce order.
+	 * @param string      $mandate_status    The new mandate status.
+	 * @param string|null $revocation_reason The revocation reason, if applicable.
+	 */
+	private function update_subscription_for_mandate( WC_Order $order, string $mandate_status, ?string $revocation_reason ): void {
+		if ( ! WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() || ! function_exists( 'wcs_get_subscriptions_for_order' ) ) {
+			return;
+		}
+
+		$subscriptions = wcs_get_subscriptions_for_order( $order );
```
Mayisha, line 848: AI feedback

🔴 High — Subscription sync silently breaks after the first renewal
  
  The core feature (pause/resume a subscription when its mandate changes) won't fire for any subscription that has renewed at least once.

  The chain:

  1. get_order_by_mandate_id() (class-wc-stripe-order-helper.php:707) returns the most recent non-terminal order carrying that mandate ID, ORDER BY ... date DESC LIMIT 1.
  2. The mandate ID is stored on every order that processes a charge — parent and each renewal order (abstract-wc-stripe-payment-gateway.php:1805-1808, reused via get_mandate_for_subscription). So after the
  first renewal, the most recent order with that mandate is a renewal order, not the parent.
  3. update_subscription_for_mandate() then calls wcs_get_subscriptions_for_order( $order ) (class-wc-stripe-webhook-handler.php:687) with default args. Default order_type is ['parent', 'switch'] (confirmed
  in wcs-order-functions.php:48) — renewal orders are excluded. So $subscriptions comes back empty and the subscription is never put on hold or reactivated.

  Net effect: the feature works for a brand-new subscription (parent is the only/most-recent order), then silently stops working from the first renewal onward — exactly the long-lived recurring-payment
  scenario this targets. 

  Fix: pass [ 'order_type' => 'any' ], matching the existing pattern already used elsewhere in this codebase for the same reason (trait-wc-stripe-subscriptions.php:1433):
  $subscriptions = wcs_get_subscriptions_for_order( $order, [ 'order_type' => 'any' ] );

changelog.txt
```diff
@@ -47,6 +47,7 @@ xxxx-xx-xx - version 10.8.0
 * Dev - Move additional classes to autoloader, including email and migration classes
 * Add - Register Stripe gateway capabilities with the WordPress Abilities API for agent access (default off; opt in via the `wc_stripe_abilities_enabled` filter)
 * Dev - Add e2e tests for subscriptions and Optimized Checkout Suite and support stricter payment confirmation checks
+* Add - Handle Stripe `mandate.updated` webhooks to update order/subscription status and add order notes when a mandate is revoked, paused, or reactivated for Indian recurring payments and other mandate-based payment methods
```
```suggestion
* Add - Handle Stripe `mandate.updated` to update order and subscription status for mandate-based payments
```

readme.txt
```diff
@@ -198,5 +198,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Dev - Move additional classes to autoloader, including email and migration classes
 * Add - Register Stripe gateway capabilities with the WordPress Abilities API for agent access (default off; opt in via the `wc_stripe_abilities_enabled` filter)
 * Dev - Add e2e tests for subscriptions and Optimized Checkout Suite and support stricter payment confirmation checks
+* Add - Handle Stripe `mandate.updated` webhooks to update order/subscription status and add order notes when a mandate is revoked, paused, or reactivated for Indian recurring payments and other mandate-based payment methods
```
Mayisha, line 203:
```suggestion
* Add - Handle Stripe `mandate.updated` to update order and subscription status for mandate-based payments
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -515,6 +515,212 @@ public function process_webhook_dispute_closed( $notification ) {
...
+			);
+			return;
+		}
+
+		// Generate order note for this mandate state change.
+		$note = $this->get_mandate_order_note( $mandate_id, $mandate_status, $revocation_reason );
+
+		// Update order status for actionable states (note is passed to update_status which adds it).
+		// For non-actionable states, add the note manually and save.
+		if ( $target_order_status ) {
```
Mayisha, line 741:
```suggestion
		if ( $target_order_status && ! $order->get_transaction_id() ) {
```

If the order is already charged successfully and has a transaction ID, we should not move the order to on-hold. Moving the subscription to on-hold is fine in this case to prevent next renewal attempt.

Mayisha (changes requested):
Thanks @anant1811 for addressing the feedback. I have a few more suggestions from code inspection.

diegocurbelo (commented):
Hey @anant1811!. Do you want me to take over this pull request and make the changes suggested in the review?

diegocurbelo (changes requested):
Hey @anant1811!. Do you want me to take over this pull request and make the changes suggested in the review?

anant1811:
My apologies, I missed the comments on this one. Yes, okay to take over @diegocurbelo !

wjrosa:
Diego is handling other things this week, so I will take it over instead now 👍

includes/class-wc-stripe-webhook-handler.php
wjrosa: Great catch, this was the most serious one. Fixed in 6ab51134e: the lookup now passes `order_type => any`, with a test asserting the args, and the date ordering got an ID tie-break so the newest-order contract is deterministic.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Agreed, done in 6ab51134e: one generic note for both cases, with the revocation reason appended when present.

includes/class-wc-stripe-order-helper.php
wjrosa: Removed in 6ab51134e.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -676,6 +676,212 @@ public function process_webhook_dispute_closed( $notification ) {
...
+	 * @since 10.9.0
+	 * @param object $notification The webhook notification from Stripe.
+	 * @return void
+	 */
+	public function process_webhook_mandate_updated( $notification ): void {
+		$mandate = $notification->data->object ?? null;
+
+		if ( ! isset( $mandate->id, $mandate->status ) ) {
+			WC_Stripe_Logger::warning( 'mandate.updated webhook received with missing mandate ID or status.' );
+			return;
```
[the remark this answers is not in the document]
wjrosa: Done in 6ab51134e: mandate, id, status, and payment_method_details are type-checked before any typed call, with PHPUnit cases for array and object values.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -676,6 +676,212 @@ public function process_webhook_dispute_closed( $notification ) {
...
+				if ( 'inactive' === $mandate_status && ! $subscription->has_status( 'on-hold' ) ) {
+					// An inactive mandate can no longer be charged, so the subscription is put on
+					// hold rather than cancelled: a revocation does not necessarily mean the customer
+					// wants to end the subscription, and the merchant can ask them to re-authorize.
+					$note = $revocation_reason
+						? __( 'Subscription put on hold because the customer revoked the Stripe mandate.', 'woocommerce-gateway-stripe' )
+						: __( 'Subscription put on hold because the Stripe mandate became inactive.', 'woocommerce-gateway-stripe' );
+					$subscription->update_status( 'on-hold', $note );
+				} elseif ( 'active' === $mandate_status && $subscription->has_status( 'on-hold' ) ) {
+					$subscription->update_status( 'active', __( 'Subscription reactivated because the Stripe mandate became active.', 'woocommerce-gateway-stripe' ) );
```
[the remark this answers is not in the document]
wjrosa: Done in 6ab51134e: pausing records a per-mandate hold marker on the subscription, and an active event only lifts holds carrying that marker for that mandate, with data-provider coverage for the no-marker and other-mandate cases.

includes/class-wc-stripe-order-helper.php
```diff
@@ -803,6 +803,64 @@ public function update_stripe_mandate_id( ?WC_Order $order = null, string $manda
 		return $this->update_order_meta( $order, self::META_STRIPE_MANDATE_ID, $mandate_id );
 	}
 
+	/**
+	 * Gets the most recent non-terminal order by Stripe mandate ID.
+	 *
+	 * @since 10.9.0
```
daledupreez, line 809: Nit: Please update the version.
```suggestion
	 * @since 11.1.0
```

includes/class-wc-stripe-order-helper.php
```diff
@@ -803,6 +803,64 @@ public function update_stripe_mandate_id( ?WC_Order $order = null, string $manda
...
+			if ( is_array( $orders ) && ! empty( $orders ) && $orders[0] instanceof WC_Order ) {
+				return $orders[0];
+			}
+
+			return null;
+		}
+
+		$order_id = $wpdb->get_var(
+			$wpdb->prepare(
+				"SELECT DISTINCT ID FROM $wpdb->posts as posts LEFT JOIN $wpdb->postmeta as meta ON posts.ID = meta.post_id WHERE meta.meta_value = %s AND meta.meta_key = %s AND posts.post_status IN ('wc-pending', 'wc-processing', 'wc-completed', 'wc-on-hold') ORDER BY posts.post_date DESC, posts.ID DESC LIMIT 1",
```
daledupreez, line 846: In both cases here, I think we should pass the permitted order status values through a filter so we can pick up any custom order statuses from stores.

daledupreez, line 846: I have some concerns about the way the database is going to optimise this query as a result of using a LEFT (OUTER) JOIN. We are requiring non-null `meta.post_id` and `meta.meta_value` values, so this may be better written as a strict inner `JOIN` -- that way MySQL/MariaDB can make a better decision about whether to start the query from the posts table or the postmeta table. We _probably_ want it to start from the meta_key index given that the number of matching mandates should be really small. In testing locally, the queries with `LEFT JOIN` and `JOIN` seem to have the same plan, but I think we should aim for the inner join.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -682,6 +688,240 @@ public function process_webhook_dispute_closed( $notification ) {
...
 
+	/**
+	 * Process webhook for mandate status updates.
+	 *
+	 * An inactive mandate (whether paused or revoked by the customer) puts the order
+	 * and any linked subscription on hold, and an active mandate reactivates an on-hold
+	 * subscription. Order notes are added for every state change so merchants have
+	 * visibility. Used for Indian recurring payments and other mandate-based methods.
+	 *
+	 * @since 10.9.0
```
daledupreez, line 699: Nit: Please bump the version.
```suggestion
	 * @since 11.1.0
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -682,6 +688,240 @@ public function process_webhook_dispute_closed( $notification ) {
...
+				'mandate_id'     => $mandate_id,
+				'mandate_status' => $mandate_status,
+				'order_id'       => $order->get_id(),
+			]
+		);
+
+		$this->resolved_order = $order;
+
+		// A revocation reason is only present when the customer has revoked the mandate.
+		$revocation_reason = $this->get_mandate_revocation_reason( $mandate );
```
daledupreez, line 738: Nit: We don't use this value if the `$target_order_status` check triggers an early return, so it would be cleaner to compute the value after that check.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -682,6 +688,240 @@ public function process_webhook_dispute_closed( $notification ) {
...
+
+		// An inactive mandate (whether paused or revoked) can no longer be charged, so the
+		// order is put on hold.
+		$target_order_status = ( 'inactive' === $mandate_status ) ? OrderStatus::ON_HOLD : null;
```
daledupreez, line 742: Not blocking: is it worth offering a filter for the target status?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -682,6 +688,240 @@ public function process_webhook_dispute_closed( $notification ) {
...
+
+		// Idempotency guard: skip if the order is already in the target status (e.g. on webhook redelivery).
+		if ( $target_order_status && $order->has_status( $target_order_status ) ) {
+			WC_Stripe_Logger::info(
+				sprintf( 'Order %d already has status %s, skipping mandate update.', $order->get_id(), $target_order_status )
+			);
```
daledupreez, line 748: Not blocking, but it would be good to include the mandate ID and status in the logging context.
```suggestion
			WC_Stripe_Logger::info(
				sprintf( 'Order %d already has status %s, skipping mandate update.', $order->get_id(), $target_order_status ),
				[
					'mandate_id'     => $mandate_id,
					'mandate_status' => $mandate_status,
				]
			);

```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -682,6 +688,240 @@ public function process_webhook_dispute_closed( $notification ) {
...
+		);
+	}
+
+	/**
+	 * Extracts the revocation reason from a mandate's payment_method_details.
+	 *
+	 * The revocation reason is nested under payment_method_details.{type}
+	 * where {type} varies by payment method (e.g., card, au_becs_debit, acss_debit).
+	 *
+	 * @since 10.9.0
```
daledupreez, line 784: Nit: version.
```suggestion
	 * @since 11.1.0
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -682,6 +688,240 @@ public function process_webhook_dispute_closed( $notification ) {
...
+			return sanitize_text_field( $mandate->payment_method_details->$type->revocation_reason );
+		}
+
+		return null;
+	}
+
+	/**
+	 * Generates the order note message for a mandate status change.
+	 *
+	 * @since 10.9.0
```
daledupreez, line 814: Nit: version
```suggestion
	 * @since 11.1.0
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -682,6 +688,240 @@ public function process_webhook_dispute_closed( $notification ) {
...
+		);
+	}
+
+	/**
+	 * Updates the subscription status based on a mandate status change.
+	 *
+	 * Only updates if WooCommerce Subscriptions is active and the order
+	 * has associated subscriptions.
+	 *
+	 * @since 10.9.0
```
daledupreez, line 860: Nit: version
```suggestion
	 * @since 11.1.0
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -682,6 +688,240 @@ public function process_webhook_dispute_closed( $notification ) {
...
+				} elseif ( 'active' === $mandate_status && $subscription->has_status( 'on-hold' ) ) {
+					// Only lift holds this handler created for this mandate.
+					if ( $mandate_id !== $subscription->get_meta( self::META_MANDATE_HOLD ) ) {
+						continue;
+					}
+
+					$subscription->delete_meta_data( self::META_MANDATE_HOLD );
+					$subscription->save();
+					$subscription->update_status( 'active', __( 'Subscription reactivated because the Stripe mandate became active.', 'woocommerce-gateway-stripe' ) );
+				}
```
daledupreez, line 911: Not blocking, but it may make sense to push this subscriptions-specific logic into code that is specific to subscriptions. It doesn't feel right to have this code in the generic webhook handler.

daledupreez, line 911: Can we add logging for the `else` case here so we know when and why we fall through?
```suggestion
					$subscription->update_status( 'active', __( 'Subscription reactivated because the Stripe mandate became active.', 'woocommerce-gateway-stripe' ) );
				} else {
					WC_Stripe_Logger::info(
						'No subscription modifications performed due to mandate update',
						[
							'subscription_id'     => $subscription->get_id(),
							'subscription_status' => $subscription->get_status(),
							'mandate_id'          => $mandate_id,
							'mandate_status'      => $mandate_status,
						]
					);
				}
```

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+			$this->assertMatchesRegularExpression( $expected_note_pattern, $notes[0]->content );
+		}
+	}
+
+	/**
+	 * Provider for `test_process_webhook_mandate_updated`.
+	 *
+	 * @return array
+	 */
+	public function provide_test_process_webhook_mandate_updated() {
```
daledupreez, line 5081: Nit: All of these test cases rely on `card` payments. I think it would be good to switch some of them to other payment methods, like SEPA.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+					],
+				],
+			],
+		];
+
+		// Should not throw an exception.
+		$this->mock_webhook_handler->process_webhook_mandate_updated( $notification );
+
+		// Verify no errors — test passes if no exception thrown.
+		$this->assertTrue( true );
```
daledupreez, line 5164: Is it worth mocking the logger so we can confirm the expected logging occurs?

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+	}
+
+	/**
+	 * Test that process_webhook_mandate_updated handles malformed payload gracefully.
+	 */
+	public function test_process_webhook_mandate_updated_malformed_payload() {
```
daledupreez, line 5170: Not blocking: this might be easier to maintain as a basic test with a data provider.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+			'data' => (object) [ 'object' => [ 'id' => 'mandate_mock_789' ] ],
+		];
+
+		$this->mock_webhook_handler->process_webhook_mandate_updated( $notification5 );
+		$this->assertTrue( true );
+	}
+
+	/**
+	 * A malformed payment_method_details (array type feeding the variable-property
+	 * access) must not fatal; the status update still happens with the generic note.
```
daledupreez, line 5236: Nit: we mention the generic note, but we don't assert it is added to the order.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+
+		$this->mock_webhook_handler->process_webhook_mandate_updated( $notification );
+
+		// The newer order should be updated to on-hold.
+		$final_new_order = wc_get_order( $new_order->get_id() );
+		$this->assertSame( OrderStatus::ON_HOLD, $final_new_order->get_status() );
+
+		// The older order should remain unchanged.
+		$final_old_order = wc_get_order( $old_order->get_id() );
+		$this->assertSame( OrderStatus::COMPLETED, $final_old_order->get_status() );
```
daledupreez, line 5308: I think we should also verify that a note is added to the new order, and not to the older order.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+	 * @dataProvider provide_test_mandate_subscription_sync
+	 *
+	 * @param string|null $revocation_reason The revocation reason, or null.
+	 */
+	public function test_process_webhook_mandate_updated_pauses_subscription( $revocation_reason ) {
+		$mandate_id = 'mandate_sub_pause';
+
+		$order = WC_Helper_Order::create_order();
+		$order->set_status( OrderStatus::PROCESSING );
+		$order->update_meta_data( '_stripe_mandate_id', $mandate_id );
```
daledupreez, line 5354: Nit: This raw update should probably be handled by the order helper to make sure we're setting the right flags etc.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+		$this->assertSame( 'on-hold', $subscription->get_status() );
+
+		// Renewal orders carry the mandate too; the default lookup excludes them.
+		$this->assertSame( [ 'order_type' => 'any' ], WC_Subscriptions_Helpers::$wcs_get_subscriptions_for_order_args );
+
+		// The hold is recorded against the mandate so only this handler's holds get lifted.
+		$this->assertSame( $mandate_id, $subscription->get_meta( '_stripe_mandate_hold' ) );
+
+		WC_Subscriptions_Helpers::$wcs_get_subscriptions_for_order      = null;
+		WC_Subscriptions_Helpers::$wcs_get_subscriptions_for_order_args = null;
```
daledupreez, line 5374: Nit: This should be handled via a `try/finally` block, as the reset won't happen if an assertion fails.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+	 *
+	 * @param string|null $hold_marker     The `_stripe_mandate_hold` meta on the subscription, or null for none.
+	 * @param string      $expected_status The expected subscription status after the webhook.
+	 */
+	public function test_process_webhook_mandate_updated_reactivates_subscription( $hold_marker, $expected_status ) {
+		$mandate_id = 'mandate_sub_reactivate';
+
+		$order = WC_Helper_Order::create_order();
+		$order->set_status( OrderStatus::PROCESSING );
+		$order->update_meta_data( '_stripe_mandate_id', $mandate_id );
```
daledupreez, line 5403: As above RE using an official helper.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+		}
+		WC_Subscriptions_Helpers::$wcs_get_subscriptions_for_order = [ $subscription ];
+
+		$notification = $this->build_mandate_notification( $mandate_id, 'active' );
+		$this->mock_webhook_handler->process_webhook_mandate_updated( $notification );
+
+		$this->assertSame( $expected_status, $subscription->get_status() );
+		if ( 'active' === $expected_status ) {
+			$this->assertSame( '', $subscription->get_meta( '_stripe_mandate_hold' ), 'The lifted hold marker must be cleared.' );
+		}
```
daledupreez, line 5419: We should also verify that the value is unchanged in the other code branch.
```suggestion
		if ( 'active' === $expected_status ) {
			$this->assertSame( '', $subscription->get_meta( '_stripe_mandate_hold' ), 'The lifted hold marker must be cleared.' );
		} elseif ( null !== $hold_marker ) {
			$this->assertSame( $hold_marker, $subscription->get_meta( '_stripe_mandate_hold' ), 'The hold marker must be left as-is.' );
		} // Not sure if there should be an else - maybe it should verify the value is still ''.
```

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+	public function test_process_webhook_mandate_updated_skips_terminal_subscription() {
+		$mandate_id = 'mandate_sub_terminal';
+
+		$order = WC_Helper_Order::create_order();
+		$order->set_status( OrderStatus::PROCESSING );
+		$order->update_meta_data( '_stripe_mandate_id', $mandate_id );
+		$order->save();
+
+		$subscription = new WC_Subscription();
+		$subscription->set_status( 'cancelled' );
```
daledupreez, line 5449: Nit: This test should use a data provider to make sure all terminal subscription statuses are explicitly tested.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -5014,4 +5014,447 @@ private function interpolate_live_and_test_modes( array $test_cases, string $val
...
+		WC_Subscriptions_Helpers::$wcs_get_subscriptions_for_order = [ $subscription ];
+
+		$notification = $this->build_mandate_notification( $mandate_id, 'inactive', 'customer_request' );
+		$this->mock_webhook_handler->process_webhook_mandate_updated( $notification );
+
+		// The cancelled subscription must not be revived or modified.
+		$this->assertSame( 'cancelled', $subscription->get_status() );
+
+		WC_Subscriptions_Helpers::$wcs_get_subscriptions_for_order = null;
```
daledupreez, line 5458: Nit: We should use `try/finally` to ensure the static variable is cleaned up.
```suggestion
		try {
			$this->mock_webhook_handler->process_webhook_mandate_updated( $notification );

			// The cancelled subscription must not be revived or modified.
			$this->assertSame( 'cancelled', $subscription->get_status() );
		} finally {
			WC_Subscriptions_Helpers::$wcs_get_subscriptions_for_order = null;
		}
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 11.1.0
+* Add - Handle Stripe `mandate.updated` to update order and subscription status for mandate-based payments
```
daledupreez, line 4: Nit: we should specify that we're handling webhooks.
```suggestion
* Add - Handle Stripe `mandate.updated` webhooks to update order and subscription status for mandate-based payments
```

daledupreez (commented):
I have lots of feedback based on inspection.

Issue #189, opened 2017-03-17
Trying to get property of non-object in /woocommerce-gateway-stripe/includes/class-wc-stripe-apple-pay.php on line 131


cloudsurferuk:
This has only started with today's update when Apple Pay was introduced.

Apple Pay is disabled the control panel however I receive the error "Trying to get property of non-object in /home/aviationwales/public_html/wp-content/plugins/woocommerce-gateway-stripe/includes/class-wc-stripe-apple-pay.php on line 131" on pages.

roykho:
@cloudsurferuk is your WP debug on?

cloudsurferuk:
It was yes, presumably I wouldn't have seen the error otherwise.

roykho:
So I just tried replicating your issue with WP debug on and I don't see this notice. Does it show on any particular page?

cloudsurferuk:
It appears on the dashboard and preview pages. Not on live pages as far as I can see.

roykho:
Would you mind please take a screenshot of it and link it here?

cloudsurferuk:
![2017-03-17](https://cloud.githubusercontent.com/assets/11637840/24057043/e0c879ba-0b3d-11e7-80f6-988583936dfb.png)

roykho:
Ok thanks for that and if I were to replicate this preview how do I do that? Is this specific to a plugin you're using or a theme? Because I have already tried preview via products page, blog page and regular page and none of the preview had errors for me.

roykho:
Ok I put in a check for object type before using which should fix your issue. Please download from master branch and let me know.

cloudsurferuk:
thanks, will do that tomorrow and let you know.

cloudsurferuk:
Hi, I can confirm the error is gone. Thanks for sorting it.

PR #5179, opened 2026-03-26
Fix Dependabot #189: remove locutus (CVE-2026-32304) via @woocommerce/currency

wjrosa:
## Summary

- Removes `@woocommerce/currency` from `devDependencies`, eliminating the `locutus` transitive dependency that carries CVE-2026-32304 (CVSS 9.8, RCE via `create_function()`).
- Adds a Jest mock for `@woocommerce/currency` so tests that transitively import `statement-preview/index.js` continue to pass.

## Context

**Vulnerability:** [GHSA-vh9h-29pq-r5m8](https://github.com/advisories/GHSA-vh9h-29pq-r5m8) / CVE-2026-32304 — `locutus <= 3.0.13` allows RCE via unsanitized input in `create_function()`. Fixed in `locutus@3.0.14`.

**Why not just bump locutus?** The patched version (`locutus@3.0.14`) requires Node >= 22, but this project is pinned to Node 20 with `engine-strict=true` in `.npmrc`. A version override is therefore not installable.

**Why is removing `@woocommerce/currency` safe?**
- It is [externalized by `@woocommerce/dependency-extraction-webpack-plugin`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/develop/webpack.config.js) — the package is never bundled into the plugin; at runtime it loads from WooCommerce's global `wc.currency`.
- No existing Jest tests directly import or exercise `@woocommerce/currency` or `@woocommerce/number`.
- All 98 Jest test suites pass after the change.

## Test plan

- [x] `npm run test:js` — 98 suites, 612 tests, all passing
- [x] `npm run phpstan` — no errors (run by the push hook)
- [ ] Manual: verify statement preview still renders correctly in the admin UI (it loads `@woocommerce/currency` from the WooCommerce global as before)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -81,6 +81,7 @@ module.exports = {
 		'import/core-modules': [
 			'@woocommerce/blocks-checkout',
 			'@woocommerce/blocks-registry',
+			'@woocommerce/currency',
 			'@woocommerce/settings',
 			'@wordpress/i18n',
 			'@wordpress/is-shallow-equal',
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,6 +22,7 @@
 * Fix - Improve performance of CSS style lookups
 * Add - Initial implementation of always-expanded Optimized Checkout Suite in shortcode checkout
 * Dev - Collapse PHPUnit tests using data providers to reduce duplication and improve test isolation
+* Dev - Remove @woocommerce/currency dev dependency to resolve locutus CVE-2026-32304 (GHSA-vh9h-29pq-r5m8)
 
 = 10.5.3 - 2026-03-19 =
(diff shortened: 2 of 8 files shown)


diegocurbelo (approved):
The change look good, and the statement preview renders correctly.

Mayisha (approved):
Looks good to me. 👍 

Nit: Additionally, it would be good to add `'@woocommerce/currency'` in `.eslintrc.js` to prevent the ESLint warning.

Image

wjrosa:
Makes sense. Added it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5179/commits/0ce84dec4d9cd2ee5a74d20a73cbd7ea242c33f1

PR #5180, opened 2026-03-26
fix: BLIK and ACSS visibility in Optimized Checkout mode

kingsleyinfo:
## Summary

Non-deferred-intent payment methods (BLIK, ACSS) were hidden at checkout when Optimized Checkout (OC) is enabled. These methods cannot render inside the OC Payment Element's deferred mode, so they now get their own separate payment method tabs alongside the OC container.

Fixes STRIPE-946.

### Changes
- **`is_available()`**: Non-deferred-intent methods fall through to standard availability checks instead of being suppressed by the OC container gate
- **`get_enabled_payment_method_config()`**: Non-deferred methods are included in the JS payment method config alongside the OC container
- **`get_excluded_payment_method_types()`**: Non-deferred methods are excluded from the Payment Element to avoid duplication with their separate tabs

## Test Coverage

All new code paths have test coverage.

```
CODE PATH COVERAGE: 4/4 paths tested (100%)
  ★★★ Non-deferred method is_available() in OC mode (BLIK + ACSS via @dataProvider)
  ★★★ Non-deferred methods excluded from PE
  ★★  Non-deferred methods in enabled config
  ★★  OC container still registered
```

- PHPUnit: 1638 tests, 3718 assertions — all pass
- Jest: 612 tests — all pass
- PHPStan: no errors

## Pre-Landing Review

No issues found.

## Adversarial Review

Claude adversarial subagent: no actionable findings. 9 items triaged — all pre-existing, theoretical, or false positives.

## Plan Completion

6/7 DONE, 1 PARTIAL (manual browser testing limited by inactive Stripe test account — server-side data verified via console inspection), 1 CHANGED (ACSS implemented alongside BLIK — same pattern).

## Test plan
- [x] All PHP tests pass (1638 tests, 3718 assertions)
- [x] All JS tests pass (612 tests)
- [x] PHPStan clean
- [x] Server-side data verified: `paymentMethodsConfig` includes BLIK alongside OC container
- [x] `excludedPaymentMethodTypes` correctly includes `blik` and `acss_debit`
- [ ] Manual browser verification requires activated Stripe account with BLIK capability


Luc45 (approved):
## Review: fix: BLIK and ACSS visibility in Optimized Checkout mode

Clean, well-scoped fix. The root cause is clear: when OC is enabled, `is_available()` only returns true for `stripe_id === 'oc'`, so BLIK and ACSS (which can't render inside the OC Payment Element because `supports_deferred_intent = false`) silently disappear.

I set up a WC 10.6.2 environment with the Stripe gateway on this branch and stepped through `is_available()` with xdebug. Confirmed that:
- BLIK hits the OC gate at line 215 (`oc_enabled=true`)
- With the fix, it enters the new branch at line 218 (`supports_deferred_intent()` returns `false`)
- Falls through to `is_enabled_at_checkout() && parent::is_available()` — the standard availability path
- Without the fix (trunk), BLIK would hit line 223 (`return OC === $this->stripe_id`) which is always `false` for BLIK — confirming the bug

Couldn't do full browser verification (no Stripe test account with BLIK capability), but the server-side logic is correct.

The three changes work together well:
1. `is_available()` — non-deferred methods bypass OC gate to standard checks ✓
2. `get_enabled_payment_method_config()` — non-deferred methods get separate entries in the JS config ✓
3. `get_excluded_payment_method_types()` — non-deferred methods excluded from Payment Element to avoid duplication ✓

Test coverage is thorough (data provider for both BLIK and ACSS, existing 1638 tests pass).

### Re: CodeRabbit comments

I reviewed the four CodeRabbit comments. The one flagged as "Major" — claiming the test doesn't exercise the OC path because `OC_FEATURE_FLAG_NAME` isn't set — contradicts what I observed at runtime. In my xdebug session with the same settings the test uses (`pmc_enabled=yes`, `optimized_checkout_element=yes`), `oc_enabled` was `true`. I also read `is_oc_available()` (`class-wc-stripe-feature-flags.php:235-257`) and it only checks `pmc_enabled`, not the `OC_FEATURE_FLAG_NAME` option. So this appears to be a false positive.

That said — could you double-check that the test is actually hitting the OC branch and not the legacy path? My xdebug session was on a live env, not inside PHPUnit, so there could be differences in how the feature flag resolves during tests.

The centralization suggestion (extracting non-deferred filtering into a helper) is valid style feedback for a follow-up but not a blocker for this fix.

Looks good to me.

kingsleyinfo:
## Re: Double-checking the OC test path (per @Luc45's request)

Traced the full call chain to confirm the test exercises the OC branch in PHPUnit, not the legacy path.

### Call chain

```
Constructor (line 137):
  $this->oc_enabled = WC_Stripe_Feature_Flags::is_oc_available()
                      && 'yes' === $this->get_option('optimized_checkout_element')

is_available() (line 215):
  if ($this->oc_enabled && $main_stripe_gateway->is_valid_optimized_checkout_page())
    → line 218: if (!$this->supports_deferred_intent())
      → return $this->is_enabled_at_checkout() && parent::is_available()   ← the fix
```

### Every condition satisfied in the test

| Condition | Value | Why |
|---|---|---|
| `pmc_enabled = 'yes'` | ✅ | Set at line 810 via `update_main_stripe_settings()` before mock construction |
| `optimized_checkout_element = 'yes'` | ✅ | Set at line 811, read by constructor |
| `$this->oc_enabled` | ✅ `true` | Computed at construction time from the two above |
| `is_valid_optimized_checkout_page()` | ✅ `true` | Returns `!is_add_payment_method_page() && !is_changing_payment_method_for_subscription()` — both false in PHPUnit (no HTTP context) |
| `supports_deferred_intent()` | ✅ `false` | BLIK constructor hardcodes this |

### OC_FEATURE_FLAG_NAME is irrelevant

`is_oc_available()` (`class-wc-stripe-feature-flags.php:235-257`) only checks `pmc_enabled`. It does **not** check `OC_FEATURE_FLAG_NAME` — that constant is deprecated (docblock: "will be removed in version 9.9.0"). Behavior is identical in PHPUnit and live environments, consistent with what @Luc45 observed in xdebug.

### No caching risk

Settings are written via `update_option()` **before** mock construction. The singleton gateway is explicitly reset (lines 817-824) so the fresh instance reads updated settings. `$this->oc_enabled` is computed at construction time from the options DB — no stale cache possible.

### Conclusion

The test correctly enters the OC branch at line 215, then the non-deferred sub-branch at line 218. **CodeRabbit's "Major" flag is a false positive.** No risk of the test silently passing through the legacy path.

daledupreez:
@coderabbitai resume

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-method-test.php
```diff
@@ -807,6 +807,72 @@ public function test_non_card_methods_are_not_available_when_optimized_checkout_
...
+		$mocked_payment_method = $this->getMockBuilder( $payment_method_class )
+			->onlyMethods( $mocked_methods )
+			->getMock();
+
+		$mocked_payment_method->method( 'get_capabilities_response' )->willReturn( self::MOCK_ACTIVE_CAPABILITIES_RESPONSE );
+		$mocked_payment_method->method( 'get_woocommerce_currency' )->willReturn( $currency );
+		$mocked_payment_method->method( 'is_subscription_item_in_cart' )->willReturn( false );
+
+		$this->assertFalse( $mocked_payment_method->supports_deferred_intent() );
+		$this->assertTrue( $mocked_payment_method->is_available() );
```
daledupreez, line 855: This test should also use the `woocommerce_is_checkout` filter to ensure that the code recognises we are in a checkout context. As things stand, the code we want to test will fail the `$main_stripe_gateway->is_valid_optimized_checkout_page()` check because `is_checkout()` will return `false` by default.

daledupreez (commented):
This is working as described, but the unit tests are not covering the case they are supposed to. We also don't have any tests that confirm that BLIK and/or ACSS are correctly reflected in the `paymentMethodConfig` Javascript params.

@kingsleyinfo, let me know if you'd like to handle the test updates, or whether you'd like me to take those changes on.

One minor thing I did trip over was that BLIK was shown _before_ the OC element initially in the block checkout, but I can't reproduce that locally or see how that was happening in production. That said, the current logic for this is super-implementation dependent, as the way we serialise the array data to JSON for the `paymentMethodConfig` effectively drives the payment method order for block checkout, which doesn't feel very robust to me. I'll take on an improvement to that logic once this PR gets merged.

frankremmy:
Adding another confirmed reproduction from this forum report: https://wordpress.org/support/topic/blik-not-showing-at-checkout-despite-being-enabled-in-stripe-and-woocommerce/

The merchant has BLIK enabled in both Stripe and WooCommerce, with PLN currency and Poland store location, but BLIK does not appear at checkout when Optimized Checkout Suite is enabled. 

I was also able to reproduce the issue in testing with the same general setup. A workaround is to disable Optimized Checkout Suite.

diegocurbelo (commented):
When both OC and Adaptive Pricing are enabled, the new BLIK/ACSS tabs on Blocks checkout renders the wrong component:

`client/blocks/upe/upe-deferred-intent-creation/payment-elements.js` picks the container based only on `isAdaptivePricingEnabled`:
- line ~221: `useState( ! stripeServerData?.isAdaptivePricingEnabled )`
- line ~236: `if ( isAdaptivePricingSupported && ! shouldLoadStripeElements )` → `CheckoutContainer`

Neither checks `supportsDeferredIntent`, so on a store with OCS + Adaptive Pricing (default-on for eligible stores since 10.8.0), the BLIK/ACSS tab this PR introduces renders a second full Checkout-Session OC container. And submitting from that tab takes the checkout-session path using the `stripe_blik` gateway, so the order gets attributed to BLIK while paying with whatever was selected inside the embedded container.

The classic implementation already gates this correctly (`client/classic/upe/payment-processing.js:304`: `isAdaptivePricingEnabled && supportsDeferredIntent`); 
_Blocks needs this same gate._ 

wjrosa:
Closing this in favor of #5966.

This branch was opened before the OC logic was extracted into the dedicated `WC_Stripe_OCS_Payment_Gateway` subclass. After that refactor, the diff here no longer applies cleanly. Two of the conflicts are semantic, not mechanical: the OC config change now belongs in `WC_Stripe_OCS_Payment_Gateway::get_enabled_payment_method_config()`, and the exclusion loop moved into a repurposed method, so taking either side would either break the fix or leave dead code.

Same fix, re-ported against the current architecture, is in #5966.

PR #5182, opened 2026-03-27
Disable stripe developer widget

### LINKED ISSUE #5167: Make it possible to disable the Stripe developer widget
With our recent updates to the Stripe JS library, Stripe now renders an assistant/tool when test mode is enabled. That's OK if the store is a test site, but if a site admin is trying to explore the Stripe configuration or integration on a real site, the debug tool will show up for _all_ site visitors.

Image

We should work out how we can make this configurable in a way that works for merchants who have live sites. The assistant can be disabled per the Stripe docs, https://docs.stripe.com/sdks/stripejs-testing-assistant#hide-the-testing-assistant, but I think we need to work out what the right approach is for managing that visibility within the Stripe plugin. Given that this is a recent addition, it may make sense to have it disabled by default.

Given that this is a change in behaviour, we may want to classify this as a bug rather than an enhancement request. We may want to disable the tool by default in an initial PR, and then explore better ways to manage the visibility in follow-up work.

Mayisha:
Towards STRIPE-1000
Fixes #5167

In this PR, we are disabling the new Stripe developer widget or testing assistant from the checkout pages. This feature was enabled unintentionally as part of Stripe JS update.
In a follow-up PR, we will make it configurable for test sites only as suggested in the issue.

## Testing instructions
- As an admin, enable OCS and Adaptive Pricing on the Stripe settings page.
- Select Stripe as the payment method.
- In `develop` notice that a developer widget appears on the page as soon as adaptive pricing is available.

Screenshot 2026-03-27 at 11 50 43 AM

- In this branch, confirm that the developer widget is absent.
- Check on both classic and block checkout pages.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@
 * Fix - Improve performance of CSS style lookups
 * Add - Initial implementation of always-expanded Optimized Checkout Suite in shortcode checkout
 * Dev - Collapse PHPUnit tests using data providers to reduce duplication and improve test isolation
+* Dev - Hide Stripe's testing assistant on checkout page
 
 = 10.5.3 - 2026-03-19 =
 * Fix - Restore default layout when Optimized Checkout is disabled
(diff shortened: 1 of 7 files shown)


wjrosa (approved):
Nice implementation 👍 . Looks good

diegocurbelo (approved):
The changes look good 👍🏼

PR #5189, opened 2026-03-28
Fix UPE font loading race: re-compute appearance after web fonts load

### LINKED ISSUE #5186: UPE appearance: getComputedStyle reads fonts before web fonts finish loading
## Description

`getAppearance()` in `client/styles/upe/index.js` calls `getComputedStyle` to sample the page's fonts for Stripe Payment Element theming, but doesn't wait for `document.fonts.ready`. When web fonts haven't loaded yet, it returns fallback generic families (e.g. `sans-serif` instead of `"Inter"`), causing the PE to render with the wrong font.

## Root Cause

`initializeUPEAppearance` in `client/stripe-utils/utils.js` is called synchronously when constructing Stripe Elements options. It calls `getAppearance()` → `getFieldStyles()` → `getComputedStyle()` immediately, before `document.fonts.ready` resolves.

## Impact

On pages where web fonts load slowly (large font files, slow connections), Stripe PE will render with fallback system fonts instead of the theme's web fonts.

## Possible Approaches

1. **Await fonts before creating Elements** — Make the PE initialization async, wait for `document.fonts.ready` before calling `initializeUPEAppearance`. Requires changes in `payment-elements.js`.
2. **Deferred re-computation** — Compute appearance eagerly (may get fallback fonts), then after `document.fonts.ready` resolves, re-compute and call `elements.update({ appearance })`. Requires storing a reference to the Stripe Elements instance.
3. **Pre-load appearance on the server** — Compute and cache the appearance config server-side during page render. The appearance is already cached client-side via `appearanceCache` — extending to server-side would avoid the race entirely.

## Notes

* WooPayments has the same issue ([WOOPMNT-5907](https://linear.app/a8c/issue/WOOPMNT-5907/upe-appearance-wrong-background-color-in-input-fields-and-incorrect)) and did not fix it in their PR [#11427](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/11427>) either.
* Related to [STRIPE-1008](https://linear.app/a8c/issue/STRIPE-1008/upe-appearance-wrong-background-color-transparent-inputs-font-loading) (other UPE appearance issues, now fixed).

pierorocca:
## Description

Fixes a race condition where `getComputedStyle()` reads fallback font families (e.g. `sans-serif` instead of `"Inter"`) because web fonts haven't loaded yet when the Stripe Payment Element appearance is computed. After fonts load, the PE now re-computes its appearance with the correct fonts.

Fixes #5186

- **`invalidateAppearanceCache()`** — New export in `client/stripe-utils/utils.js` that clears the in-memory appearance cache, forcing `initializeUPEAppearance()` to re-compute from current page styles
- **Blocks checkout** — `useEffect` in `PaymentProcessor` that waits for `document.fonts.ready`, busts the stale cache, re-computes appearance, and calls `elements.update({ appearance })`
- **Classic checkout** — Same deferred re-computation pattern after `stripe.elements()` creation, using the same `shouldExpandOptimizedCheckout` parameter for OC support
- **jsdom guard** — Optional chaining on `document.fonts` since jsdom doesn't implement the FontFaceSet API

## How it works

1. PE initializes immediately with whatever fonts are available (may be fallback generics)
2. `document.fonts.ready` resolves when all requested font faces finish loading
3. The stale appearance cache is invalidated and re-computed with correct `fontFamily` values
4. `elements.update({ appearance })` applies the updated appearance to the live PE
5. If fonts are already loaded at mount time (`document.fonts.status === 'loaded'`), no re-computation occurs

The `elements.update()` pattern is already used in this codebase for `setupFutureUsage` updates.

## Testing instructions

1. Open DevTools → Network → throttle to **Slow 3G**
2. Go to **Blocks checkout** with a product in cart
3. Observe: PE initially renders with system fonts, then updates to the theme's web fonts after they load
4. Repeat on **Classic checkout**
5. Repeat with **Optimized Checkout** enabled
6. On **Fast 3G / no throttle**: verify no unnecessary re-computation (fonts load before PE)

## Pre-review checklist

- [x] Added changelog entry (both changelog.txt and readme.txt)
- [x] All JS tests pass (612 tests, 98 suites)
- [x] JS lint passes
- [x] Covers blocks checkout, classic checkout, and OC

---


pierorocca:
## Review Report

**Branch:** `fix/upe-font-loading-race`
**Commits:** 5

### Overall Verdict: **APPROVE**

Clean deferred re-computation pattern using `document.fonts.ready` + `elements.update()` — both already proven in the codebase. Covers blocks, classic, and OC checkout. No payment flow impact.

### Architecture Compliance — PASS
- `invalidateAppearanceCache()` colocated with `initializeUPEAppearance()` in `client/stripe-utils/utils.js`
- Blocks: `useEffect` with `[elements]` dependency — matches existing effects in the same component
- Classic: `document.fonts.ready.then()` — appropriate for vanilla JS context
- Both paths pass correct parameters (`'true'` for blocks, `'false' + shouldExpandOptimizedCheckout` for classic/OC)

### Security — PASS
- No user input, no server calls, no data flow changes. `document.fonts.ready` is a read-only browser API.

### Performance — PASS
- Early exit when fonts already loaded (`document.fonts?.status === 'loaded'`)
- Non-blocking async — doesn't delay rendering or checkout flow
- One-time cost — cache repopulated on re-compute, no repeated work

### Test Coverage — PASS
- 612/612 JS tests pass, no regressions
- `document.fonts` not available in jsdom — optional chaining guards prevent test failures
- Visual testing (throttled network) is the appropriate verification for this fix

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -301,6 +305,20 @@ const PaymentProcessor = ( {
...
+		if ( ! elements || document.fonts?.status === 'loaded' ) {
+			return;
+		}
+
+		document.fonts?.ready?.then( () => {
+			invalidateAppearanceCache();
+			const appearance = initializeUPEAppearance( 'true' );
+			elements.update( { appearance } );
+		} );
+	}, [ elements ] );
```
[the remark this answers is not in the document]
pierorocca: Fixed in c9f84514. Added `useEffect` cleanup with cancellation guard.

client/classic/upe/payment-processing.js
```diff
@@ -270,6 +271,19 @@ async function createStripePaymentElement( api, paymentMethodType ) {
...
+	if ( document.fonts?.status !== 'loaded' ) {
+		document.fonts?.ready?.then( () => {
+			invalidateAppearanceCache();
+			const appearance = initializeUPEAppearance(
+				'false',
+				shouldExpandOptimizedCheckout
+			);
+			elements.update( { appearance } );
+		} );
+	}
```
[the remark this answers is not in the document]
pierorocca: Fixed in c9f84514. Added `typeof elements?.update !== 'function'` guard for the `stripe.initCheckout()` (Adaptive Pricing) path.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -301,6 +305,28 @@ const PaymentProcessor = ( {
...
+
+		let cancelled = false;
+		document.fonts?.ready?.then( () => {
+			if ( cancelled ) {
+				return;
+			}
+			invalidateAppearanceCache();
+			const appearance = initializeUPEAppearance( 'true' );
+			elements.update( { appearance } );
+		} );
```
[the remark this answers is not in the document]
pierorocca: Fixed in 5f806187. Removed the `document.fonts?.status === 'loaded'` early exit. `document.fonts.ready` retains its resolved state, so `.then()` runs immediately (as a microtask) when fonts are already loaded — no race window.

client/classic/upe/payment-processing.js
```diff
@@ -270,6 +271,22 @@ async function createStripePaymentElement( api, paymentMethodType ) {
...
+			if ( typeof elements?.update !== 'function' ) {
+				return;
+			}
+			invalidateAppearanceCache();
+			const appearance = initializeUPEAppearance(
+				'false',
+				shouldExpandOptimizedCheckout
+			);
+			elements.update( { appearance } );
+		} );
```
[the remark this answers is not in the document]
pierorocca: Fixed in 5f806187. Removed the status gate and moved `invalidateAppearanceCache()` before the `elements?.update` check so the stale cache is cleared even when the Checkout object doesn't support `.update()`.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -308,6 +312,28 @@ const PaymentProcessor = ( {
...
```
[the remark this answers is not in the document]
pierorocca: Fixed. Added `typeof elements?.update !== 'function'` guard before the `elements.update()` call, matching the existing pattern in the classic path.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -308,6 +312,31 @@ const PaymentProcessor = ( {
...
+				return;
+			}
+			const appearance = initializeUPEAppearance( 'true' );
+			elements.update( { appearance } );
+		} );
+
+		return () => {
+			cancelled = true;
+		};
+	}, [ elements ] );
```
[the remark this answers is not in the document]
pierorocca: This is a React-specific lifecycle concern that doesn't apply to the classic checkout path. `createStripePaymentElement()` is an imperative function that runs once per page load — there's no unmount, no cleanup return, and `elements` is a local variable scoped to that single invocation. Adding a cancellation guard here would either be dead code or, as the suggested snippet shows (`cancelled = true` at function end), would cancel the callback before fonts load. The Blocks path needs the guard because React effects can re-run and unmount; this path cannot.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -308,6 +312,31 @@ const PaymentProcessor = ( {
...
+		if ( ! elements ) {
+			return;
+		}
+
+		let cancelled = false;
+		document.fonts?.ready?.then( () => {
+			if ( cancelled ) {
+				return;
+			}
+			invalidateAppearanceCache();
```
malithsen: I'm having trouble replicating the original issue, but looking at the code, I think this will always invalidate the appearance cache for all requests.

The only scenario where the re-computation is actually needed is when fonts load after the initial appearance was computed. But this line unconditionally busts the cache every time. So even on fast connections, cache never survives.

Maybe we should compare the font families, and only invalidate if they actually changed.

malithsen (changes requested):
Requesting changes as I think we should add a guard to prevent unnecessary cache invalidations.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
pierorocca: Fixed. Both blocks and classic now compare the cached `fontFamily` with a freshly computed value after `document.fonts.ready` resolves. If they match (fonts were already loaded at PE init time), the cache isn't invalidated and `elements.update()` is skipped entirely.

pierorocca: Great catch @malithsen. Thanks for the reviews.

diegocurbelo: The changes only skips `elements.update()`... the cache invalidation + getComputedStyle() recomputation happens on every page load, even when the fonts never change.

The fix should compare fonts before invalidating.
Something like: read the fontFamily from a fresh getComputedStyle() on the sampling element, compare it to what's in the cached appearance, and only invalidate + recompute if they actually differ.

diegocurbelo (commented):
The code changes look good in general, I left [a comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5189#discussion_r3042060289) on an alternative to avoid invalidating the cache on each page load

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
pierorocca: Fixed. Added `sampleFontFamily()` that reads fontFamily from a single `getComputedStyle()` call on the text selector. Both blocks and classic now compare the live font against the cached appearance *before* calling `invalidateAppearanceCache()` — no cache bust or full recomputation occurs when fonts haven't changed.

pierorocca: Updated. `sampleFontFamily()` now does a single `getComputedStyle().fontFamily` read using the same selector-array-to-querySelector coercion as `getFieldStyles`, so it resolves the exact same DOM element. Both blocks and classic compare this live value against the cached appearance before deciding whether to invalidate and recompute.

Could you take another look @diegocurbelo and scrutinize the approach? Want to make sure there aren't other edge cases I'm missing.

diegocurbelo (approved):
Thanks for the changes @pierorocca!

LGTM 👍🏼

client/AGENTS.md
```diff
@@ -49,6 +50,8 @@ For repository-wide rules, always read the root `AGENTS.md` first.
 - Mixing display amounts and API minor-unit amounts in per-feature logic.
 - Safari inline element baseline gap: Stripe Express Checkout Element iframes are inline-level replaced elements. Safari adds extra whitespace below the text baseline (the "image gap"), inflating container height beyond the visible button. Fix with `font-size: 0; line-height: 0` on the container, not by targeting margins.
 - Using block layout (`padding-bottom` on children) for vertical button spacing instead of flexbox/grid with explicit `gap`. Block layout margin/padding behavior varies across browsers; flex/grid `gap` is deterministic.
+- Using `document.fonts.status` as a gate before subscribing to `document.fonts.ready`. The `ready` promise retains its resolved state — `.then()` runs as a microtask even when fonts are already loaded. A `status` check creates a race window. Always subscribe unconditionally.
+- Calling `elements.update()` without checking the object type. The Adaptive Pricing path uses `stripe.initCheckout()` which returns a Checkout object that may lack `.update()`. Always guard with `typeof elements?.update === 'function'`.
```
malithsen: nit: I think we should reserve these files for generic guidelines and avoid things that are too specific. 

That said, including these might not really have a practical impact given the increasing context lengths of newer models.

malithsen (approved):
Looks good, thanks for the updates. I left a non-blocking comment on the update to the AGENTS.md file; feel free to ignore.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
pierorocca: Addressed in 35098673. Both the blocks and classic paths now compare `sampleFontFamily()` against the cached `appearance.variables.fontFamily` before invalidating — the cache only busts when fonts actually differ:

```js
const cachedFont = initializeUPEAppearance(...)?.variables?.fontFamily;
const liveFont = sampleFontFamily( true );
if ( ! liveFont || liveFont === cachedFont ) {
    return;
}
invalidateAppearanceCache();
```

On fast connections where the font is already loaded, `liveFont === cachedFont` and no invalidation occurs.

client/AGENTS.md
pierorocca: Agreed — I'll remove the overly specific entries. The Safari baseline gap and `document.fonts.status` race pitfalls are broadly applicable enough to keep, but the `elements.update()` guard is too implementation-specific. I'll trim it down.

PR #5191, opened 2026-03-30
Fix STRIPE-643: Return error response when PMC update fails in settings

thuautp:
## Summary

- When the Stripe API rejects a Payment Method Configuration (PMC) update, the settings page was silently swallowing the error and returning HTTP 200
- This caused the frontend to display "Settings saved." even though the payment method changes were not actually persisted
- Fix threads a `WP_Error` return value through the three-layer call chain so the REST endpoint returns 500 on failure, allowing the frontend's existing error-notice handler to fire correctly

## Bug

[STRIPE-643: Display error in Payment Methods page when PMC update fails](https://linear.app/a8c/issue/STRIPE-643/display-error-in-payment-methods-page-when-pmc-update-fails)

## Changes

- `includes/class-wc-stripe-payment-method-configurations.php` — `update_payment_method_configuration()` now returns `WP_Error` on failure (no config found, or Stripe API error) and `true` on success, instead of `void`
- `includes/payment-methods/class-wc-stripe-upe-payment-gateway.php` — `update_enabled_payment_methods()` propagates the return value from the PMC layer
- `includes/admin/class-wc-rest-stripe-settings-controller.php` — `update_settings()` checks the result and returns HTTP 500 if the PMC update failed; `update_enabled_payment_methods()` returns the result instead of `void`
- `tests/phpunit/admin/class-wc-rest-stripe-settings-controller-test.php` — regression test asserting a 500 response is returned when the Stripe API rejects the PMC update

## Testing

- Regression test added: `test_update_settings_returns_500_when_pmc_update_fails` — mocks the Stripe API to return an error on `update_payment_method_configurations` and asserts the settings endpoint returns 500
- QIT automated tests will cover the happy path (no regression to normal save flow)
- To verify manually: on a PMC-enabled site, trigger a Stripe API error on save (e.g. temporarily use invalid credentials) and confirm the admin now shows "Error saving settings." instead of "Settings saved."

## Notes


zhongruige (changes requested):
Hi @thuautp! Thanks for getting this in!

I think there might be quite a few unintended changes on this PR--I see 97 files changed, for example, including changes to GitHub actions files. It looks like we should only have these changes here, based on the PR description:

- `includes/class-wc-stripe-payment-method-configurations.php` — `update_payment_method_configuration()` now returns `WP_Error` on failure (no config found, or Stripe API error) and `true` on success, instead of `void`
- `includes/payment-methods/class-wc-stripe-upe-payment-gateway.php` — `update_enabled_payment_methods()` propagates the return value from the PMC layer
- `includes/admin/class-wc-rest-stripe-settings-controller.php` — `update_settings()` checks the result and returns HTTP 500 if the PMC update failed; `update_enabled_payment_methods()` returns the result instead of `void`
- `tests/phpunit/admin/class-wc-rest-stripe-settings-controller-test.php` — regression test asserting a 500 response is returned when the Stripe API rejects the PMC update

I would recommend closing this out and starting fresh, as there's also a lot of merge conflicts that would need to be resolved as well. 

Let me know if you have any questions or if there's anything I can help with!

thuautp:
closing in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5388

PR #5194, opened 2026-03-30
Fix console error when toggling save payment method with Adaptive Pricing

Mayisha:
Fixes STRIPE-998

On classic checkout with Optimized Checkout, changing the “save payment information” checkbox runs code that calls `elements.update`. When Adaptive Pricing is active, the Payment Element is created via `stripe.initCheckout()`, which returns a Checkout instance that does not expose `elements.update`, causing an uncaught TypeError in the console.

Call `elements.update` only when the function is defined.

Added this function check instead of `isAdaptivePricingEnabled` flag check (which is sent from the server side) because the integration can still use the standard `stripe.elements()` path as a fallback when AP session creation fails, and that path does support update.

## Testing instructions
- Enable adaptive pricing flag by adding this code snippet `add_filter('wc_stripe_is_checkout_sessions_available', '__return_true');`
- Enable optimized checkout and adaptive pricing from the settings page.
- As a shopper, add a product to your cart and go to the classic checkout page.
- Select Stripe for payment.
- Open the developer console of your browser.
- When the adaptive pricing element is loaded, click the save payment method checkbox.
- In `develop` branch, you'll see the following console error

```
Uncaught TypeError: component.elements.update is not a function
```
- In this branch, there should be no console errors.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -31,6 +31,7 @@
 * Dev - Hide Stripe's testing assistant on checkout page
 * Dev - Treat misaligned statements as errors in PHPCS ruleset
 * Fix - Put subscription on hold when Stripe Radar blocks a renewal payment to prevent WC Subscriptions from scheduling further retry attempts
+* Fix - Prevent JavaScript error in `elements.update` when using checkout sessions with adaptive pricing
 
 = 10.5.3 - 2026-03-19 =
 * Fix - Restore default layout when Optimized Checkout is disabled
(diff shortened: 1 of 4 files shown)


diegocurbelo (approved):
LGTM 👍🏼

wjrosa (approved):
Fix makes sense and works as intended 👍

PR #5198, opened 2026-03-31
Handle adaptive pricing async success event

Mayisha:
Towards STRIPE-921

Handled `checkout.session.async_payment_succeeded` event. For async payments this event handler will move the order to it's final status.

## Testing instructions

- Code review. Check if the logic looks good to you.

I haven't found a payment method that works with adaptive pricing but needs this async webhook only. I tested with ACH with USD and confirmed that this event is received in the logs.
**Note that** there is an existing bug that sets the payment method as card, though it was paid with ACH. This will be handled separately.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1487,7 +1496,7 @@ public function process_checkout_session( object $notification ): void {
 		 */
 		$allowed_payment_processing_statuses = apply_filters(
 			'wc_stripe_allowed_payment_processing_statuses',
-			[ OrderStatus::PENDING, OrderStatus::FAILED ],
+			[ OrderStatus::PENDING, OrderStatus::FAILED, OrderStatus::ON_HOLD ],
```
Mayisha, line 1553: Included on-hold status here to support the async_payment_succeeded event

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1437,17 +1437,26 @@ public function process_account_updated( $notification ) {
...
 	 * @param object $notification The notification from Stripe
 	 * @return void
 	 */
-	public function process_checkout_session( object $notification ): void {
+	public function process_checkout_session_success( object $notification ): void {
 		$checkout_session = $notification->data->object;
-		$session_id       = $checkout_session->id;
+
+		if ( ! isset( $checkout_session->id ) ) {
+			WC_Stripe_Logger::debug( 'Checkout session ID is missing from the event data.' );
```
daledupreez: Should this not be an error or warning? And maybe we should log the data that we received as well.
```suggestion
			WC_Stripe_Logger::error( 'Checkout session ID is missing from the event data.', [ 'notification' => $notification ] );
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1437,17 +1437,26 @@ public function process_account_updated( $notification ) {
...
+	 * - checkout.session.completed event; Fires when a Stripe Checkout session is completed.
+	 * - checkout.session.async_payment_succeeded event; Fires when an asynchronous payment method on a Stripe Checkout session succeeds.
 	 *
 	 * This webhook exists for both standard post-payment checkout
 	 * sessions, as well as agentic checkout sessions.
 	 *
 	 * @param object $notification The notification from Stripe
 	 * @return void
 	 */
-	public function process_checkout_session( object $notification ): void {
```
daledupreez, line 1486: This method was included in the 10.5.3 release. I think we need to keep the old method with a deprecation notice. We should not remove the method entirely, as it is technically a public API.

daledupreez (changes requested):
This looks OK to me based on inspection, but I am not sure whether testing with ACH in USD does or does not exercise this code path based on the wording in the main PR.

That said, I think we need to keep the old method around as it was available in the 10.5 release.

diegocurbelo (approved):
The changes look good to me 👍🏼 

Approving this, but as Dale mentioned, we should keep `process_checkout_session()` with a deprecation notice before merging.

includes/class-wc-stripe-webhook-handler.php
Mayisha: Done.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1437,17 +1437,26 @@ public function process_account_updated( $notification ) {
...
+	public function process_checkout_session_success( object $notification ): void {
 		$checkout_session = $notification->data->object;
-		$session_id       = $checkout_session->id;
+
+		if ( ! isset( $checkout_session->id ) ) {
+			WC_Stripe_Logger::error( 'Checkout session ID is missing from the event data.', [ 'notification' => $notification ] );
```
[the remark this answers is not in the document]
Mayisha: Removed. The data is already present in the `Webhook received` log before this one.

malithsen (approved):

PR #5203, opened 2026-03-31
Add Agentic Commerce intro, onboarding guide, and settings UI

wjrosa:
Fixes STRIPE-1034
Depends on https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5214

## Summary

- Adds an **About Agentic Commerce** intro card with a description, a link to the Stripe docs, and a numbered onboarding guide pointing merchants to Stripe Dashboard > Payments > Agentic Commerce
- Adds an **Agentic Commerce Settings** card with an enable/disable toggle and a webhook secret input field (revealed when the toggle is on)
- Extends the REST controller with `GET /settings` and `POST /settings` endpoints that persist `is_enabled` (stored as `wc_stripe_agentic_commerce_enabled` — a merchant-controlled setting, distinct from the developer feature flag `_wcstripe_feature_agentic_commerce`) and `webhook_secret` (`wc_stripe_agentic_commerce_webhook_secret` option)
- The inventory tracker now gates both hook registration and scheduled sync on the merchant `is_enabled` setting; when it is off, no stock-change hooks are registered and no incremental syncs run
- Updates the webhook handler to fall back to the DB-stored secret when the `AGENTIC_COMMERCE_WEBHOOK_SECRET` constant is not defined or empty in `wp-config.php`
- Adds two PHPStan baseline entries (unavoidable: `WP_REST_Request` generic type and `booleanAnd.rightAlwaysFalse` on the constant truthiness check)
- Loads the REST controller in the PHPUnit bootstrap so the controller tests actually run
- Adds 8 new PHP tests and 11 new JS tests

Onboarding guide + setting preview:
Screenshot 2026-03-31 at 17 28 07

## Test plan

- [ ] JS unit tests pass: `npm run test:js -- --testPathPattern=agentic-commerce`
- [ ] PHP unit tests pass: `npm run test:php`
- [ ] PHP lint passes: `npm run lint:php`
- [ ] PHPStan passes: `npm run phpstan`
- [ ] Merge https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5214
- [ ] Navigate to WooCommerce > Settings > Payments > Agentic Commerce tab — confirm the intro card and onboarding steps are visible
- [ ] Toggle **Enable Agentic Commerce** on — confirm the webhook secret field appears
- [ ] Enter a webhook secret and click **Save Settings** — confirm success notice and value is persisted
- [ ] Toggle off and save — confirm webhook secret field disappears and feature is disabled
- [ ] With **Enable Agentic Commerce** toggled off, trigger a stock change on any product — confirm no entry is added to the `wc_stripe_agentic_pending_inventory` option and no inventory sync is scheduled
- [ ] With **Enable Agentic Commerce** toggled on, trigger a stock change — confirm the pending inventory option is populated and a sync is scheduled
- [ ] With the feature on and a webhook secret set, confirm `v1.delegated_checkout.*` webhooks are validated against the stored secret (not the constant)


client/settings/agentic-commerce/index.js
```diff
@@ -20,492 +30,261 @@ const CardTitle = styled.h2`
...
+							value={ decodeURIComponent( webhookURLForDisplay ) }
+							autoComplete="off"
+							disabled={ true }
+						/>
+						<Button
+							variant="primary"
+							isBusy={ isSavingSettings }
+							onClick={ handleCopy }
+						>
+							{ webhookURLCopied ? 'Copied!' : 'Copy' }
```
diegocurbelo: These two strings should be wrapped in __() for translation:
```suggestion
							{ webhookURLCopied
								? __( 'Copied!', 'woocommerce-gateway-stripe' )
								: __( 'Copy', 'woocommerce-gateway-stripe' ) }
```

client/settings/agentic-commerce/sync-status.js
```diff
@@ -0,0 +1,513 @@
...
+	margin-bottom: 20px;
+`;
+
+const CardTitle = styled.h2`
+	font-size: 14px;
+	font-weight: 600;
+	margin: 0 0 16px;
+	padding-bottom: 8px;
+	border-bottom: 1px solid #eee;
+`;
```
diegocurbelo: `Card` and `CardTitle` are defined identically here and in `client/settings/agentic-commerce/index.js`, we should extract them into a shared file (or importing them from index.js) to avoid duplication.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -0,0 +1,513 @@
...
+		padding: 4px 0;
+	}
+`;
+
+const Actions = styled.div`
+	display: flex;
+	gap: 8px;
+	margin-top: 16px;
+	align-items: center;
+`;
```
diegocurbelo: Same as `Card` and `CardTitle`, consider moving this to a shared file.

client/settings/agentic-commerce/index.js
```diff
@@ -20,492 +30,261 @@ const CardTitle = styled.h2`
...
-		return sprintf(
-			/* translators: %d: number of minutes until next sync */
-			__(
-				'Next automatic sync: in %d minutes.',
-				'woocommerce-gateway-stripe'
-			),
-			minutes
-		);
+	const handleCopy = () => {
+		navigator.clipboard.writeText( webhookURLForDisplay );
```
diegocurbelo: We should await/catch a failed promise from `writeText()`, and we can try a fallback (like `document.execCommand('copy')`) or just show a notice.

diegocurbelo (commented):
The code changes look good to me, I left some suggestions.

I also, had a problem during the test plan (step 5)
Affter merging `Add merchant-controlled enabled toggle for Agentic Commerce` https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5214

I started getting these errors:
Image

Image

elizaan36:
hi @wjrosa, Thanks for putting this together. I'll share my design recommendations.

- Placement in settings: Agentic commerce should appear just above the advanced settings section and below Payments & transactions. I'll share a mockup of what this would look like within the current 2col layout. It's hard to tell from your screenshot where the section appears, maybe this is already the placement.
- When the feature is not enabled, only the checkbox and help text will show. This prevents overwhelm and allows easier access to the rest of settings.
- "Agentic commerce" is not capitalized unless is appears at the beginning of a sentence, and we use sentence case throughout settings.
- The "Learn more" link appears in the left column to reduce the size of the settings section.
- The save action shouldn't appear within the agentic commerce settings section, it remains a global element on the page.

Stripe - General

- Upon enabling the feature, setup instructions and webhook form field field will appear. I'll share a mockup of settings when the feature is enabled.

Enabled

- The step by step instructions are simplified and endpoint URL is consolidated onto one line to easily copy. Let me know if that's possible. Also removed additional helper text that felt repetitive.
- According to the [demo Rado shared](https://wcstripe.wordpress.com/2026/03/19/stripe-agentic-commerce-handover/#agentic-onboarding), the webhook needs to be copied from Developers > Webhooks so I updated the copy to reflect that.

Let me know if you have any questions about these notes. Thanks! cc @pierorocca

client/settings/agentic-commerce/index.js
wjrosa: Fixed in the previous commit (4761f638). Thanks for catching!

client/settings/agentic-commerce/sync-status.js
wjrosa: Extracted `Card`, `CardTitle`, and `Actions` into a shared `styled.js` file. Both `index.js` and `sync-status.js` now import from there.

client/settings/agentic-commerce/sync-status.js
wjrosa: Removed the duplicate `Actions` component here — now imported from the shared `styled.js` file.

client/settings/agentic-commerce/index.js
wjrosa: Updated `handleCopy` to properly await the promise, with a `document.execCommand('copy')` fallback for browsers without the clipboard API, and an error notice shown on failure.

wjrosa:
@elizaan36 thank you for reviewing the designs here already. I did not reach your team as I intend to tackle design changes in another iteration/PR. I will keep this one as is, so we can keep the main functionality. Once I have the other PR, I can tag you there

wjrosa:
@diegocurbelo, this error is due to the lack of the composer autoloader classmap file. Try to run `composer dump-autoload` and see if it fixes the issue for you

diegocurbelo (approved):
The code changes look good, and it works as described.

I have a couple comments, though:

- The `GET /settings` returns the full webhook secret (class-wc-rest-stripe-agentic-commerce-controller.php:156), the frontend uses type="password" on the input, but the secret is still sent on every page load, we should consider masking it in the response (or not sending it and should ****) and only accepting a new value on POST (don't round-trip the real secret). Similar to how the old manual/shared keys functionally worked.

- The error notices could use a bottom separation:
Image

Image

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker-test.php
```diff
@@ -87,7 +88,22 @@ public function test_constants() {
 	 *
 	 * @return void
 	 */
+	/**
+	 * Test register_hooks does not attach hooks when the merchant setting is disabled.
+	 *
+	 * @return void
+	 */
```
diegocurbelo: There are two consecutive doc blocks, the first one (test_constants doc) should be removed since it's orphaned.

wjrosa: Fixed — removed the orphaned doc block. Thanks!

wjrosa:
Thanks for the review, @diegocurbelo!

Both points addressed in the latest commit:

1. **Webhook secret masking** — `GET /settings` now returns `****` when a secret is stored, so the real value is never sent to the client. `POST /settings` skips the update when it receives `****`, preserving the stored secret when the user saves without changing the field — same pattern as the manual/shared keys.

2. **Notice spacing** — Added `margin-bottom: 16px` to the settings Notice so it has proper separation from the content below.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -0,0 +1,504 @@
...
+										<td>
+											<code>{ lastSync.file_id }</code>
+										</td>
+									</tr>
+								) }
+							</tbody>
+						</DetailsTable>
+
+						{ nextSyncLabel() && (
+							<p className="description">{ nextSyncLabel() }</p>
```
Mayisha, line 377: Nit: nextSyncLabel is called twice.

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
```diff
@@ -178,6 +214,50 @@ public function trigger_sync() {
...
+	 * @return WP_REST_Response
+	 */
+	public function get_agentic_settings(): WP_REST_Response {
+		$secret = (string) get_option( self::WEBHOOK_SECRET_OPTION, '' );
+		return rest_ensure_response(
+			[
+				'is_enabled'     => 'yes' === get_option( WC_Stripe_Agentic_Commerce_Integration::ENABLED_OPTION, 'no' ),
+				// Never expose the real secret. Return '****' when one is stored so
+				// the frontend knows a secret exists without round-tripping the value.
+				'webhook_secret' => '' !== $secret ? '****' : '',
```
Mayisha, line 230: As mentioned in another comment, I think we should show the secret similar to how Stripe shows it. My first impression from functional test was that it is a bug that I am always seeing 4 dots/stars.

Image

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -125,7 +125,9 @@ public function check_for_webhook() {
 		$is_agentic_hook = 0 === strpos( $event_type, 'v1.delegated_checkout.' );
 
 		$secret = $is_agentic_hook
-			? ( defined( 'AGENTIC_COMMERCE_WEBHOOK_SECRET' ) ? AGENTIC_COMMERCE_WEBHOOK_SECRET : '' )
+			? ( defined( 'AGENTIC_COMMERCE_WEBHOOK_SECRET' ) && AGENTIC_COMMERCE_WEBHOOK_SECRET
```
Mayisha, line 128: Just a thought, should we give the value from the settings preference here?

Mayisha (approved):
I have tested the feature together with https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5260 and listed my observations in this [comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5260#pullrequestreview-4142029228).

client/settings/agentic-commerce/sync-status.js
wjrosa: Good catch! Fixed in 1a0bec31 — `nextSyncLabel()` is now called once and the memoised value is used in the JSX.

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
wjrosa: Agreed — it did read like a bug at a glance. In 1a0bec31 the masked placeholder returned by `GET /settings` is now `whsec_********************************`, matching Stripe's `whsec_` prefix style rather than four dots. `POST /settings` still skips the update when the client echoes the mask back verbatim, so saving without editing the field preserves the stored secret.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Good thought — done in 1a0bec31. The webhook handler now reads the secret directly from `WEBHOOK_SECRET_OPTION`, and I've removed the `AGENTIC_COMMERCE_WEBHOOK_SECRET` constant override (and its PHPStan stub) since the settings UI is now the single source of truth.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -0,0 +1,509 @@
...
+	return new Date( timestamp * 1000 ).toLocaleString();
+};
+
+const humanTimeDiff = ( timestamp ) => {
+	if ( ! timestamp ) return '';
+	const diffSec = Math.floor( Date.now() / 1000 ) - timestamp;
+	if ( diffSec < 60 ) return __( 'just now', 'woocommerce-gateway-stripe' );
+	if ( diffSec < 3600 ) {
+		const m = Math.floor( diffSec / 60 );
+		if ( m === 1 )
```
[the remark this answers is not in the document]
wjrosa: Good call — fixed in c975f743. `sync-status.js` now imports `_n` and `sprintf` from `@wordpress/i18n` and uses `_n()` for the `%d minute(s) ago`, `%d hour(s) ago`, `in %d minute(s)`, and overdue warning strings. The custom `%d`-only `sprintf` helper has been removed.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -0,0 +1,509 @@
...
+	return (
+		<>
+			<p className="description">
+				{ __(
+					'Monitors the product feed sync status for the Agentic Commerce integration.',
+					'woocommerce-gateway-stripe'
+				) }{ ' ' }
+				<a
+					href="https://dashboard.stripe.com/data-management/import-sets"
+					target="_blank"
```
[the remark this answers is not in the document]
wjrosa: Good catch. Restored in c975f743 — `importSetsUrl` and `logsUrl` are now read from `wc_stripe_settings_params.agentic_commerce_import_sets_url` / `agentic_commerce_logs_url`, with the hardcoded live URLs kept only as fallbacks. Test-mode merchants now land on `dashboard.stripe.com/test/...` as before.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-files-api-delivery.php
```diff
@@ -299,6 +299,20 @@ private function upload_to_files_api( string $file_path ): array {
 	 * @throws Exception If the request fails.
 	 */
 	private function create_import_set( string $file_id, string $standard_data_format ): array {
+		/**
+		 * Allows the ImportSet API request to be short-circuited (e.g. for testing).
+		 *
+		 * @since 10.6.0
```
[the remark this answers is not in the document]
wjrosa: Fixed in c975f743 — bumped to `@since 10.7.0`.

client/settings/agentic-commerce/index.js
```diff
@@ -1,579 +1,300 @@
...
+							<OnboardingSteps>
+								<li>
+									{ interpolateComponents( {
+										mixedString: __(
+											'Go to {{agenticLink}}Payments > Agentic Commerce{{/agenticLink}} in your Stripe Dashboard.',
+											'woocommerce-gateway-stripe'
+										),
+										components: {
+											agenticLink: (
+												<ExternalLink href="https://dashboard.stripe.com/agentic-commerce" />
```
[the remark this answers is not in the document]
wjrosa: Good catch — same class of issue as c975f743. Fixed in 6e2d29a0: the onboarding step's `ExternalLink` now uses `https://dashboard.stripe.com/test/agentic-commerce` when `isTestMode` is true, mirroring how `webhookURLForDisplay` already picks the test webhook URL.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker-test.php
```diff
@@ -463,7 +476,7 @@ public function test_generate_inventory_feed_includes_all_pending_products() {
 	// -------------------------------------------------------------------------
 
 	/**
-	 * Test sync_inventory skips when feature flag is disabled.
+	 * Test sync_inventory skips when the merchant setting is disabled.
```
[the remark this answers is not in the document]
wjrosa: Good catch — the filter was dead code since `sync_inventory()` moved to `ENABLED_OPTION`. Fixed in 8518c3db: dropped the `add_filter` line and clarified the comment that the test relies on `ENABLED_OPTION` defaulting to `'no'` when unset, so there's nothing left leaking to subsequent tests.

client/settings/agentic-commerce/index.js
```diff
@@ -1,626 +1,305 @@
...
+			setIsFeatureEnabled( result.is_enabled );
+			setWebhookSecret( result.webhook_secret ?? '' );
+		} catch {
+			// Settings fetch failure is non-fatal; defaults remain.
 		} finally {
-			if ( ! background ) {
-				setIsLoading( false );
-			}
+			setIsLoadingSettings( false );
 		}
```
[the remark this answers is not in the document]
wjrosa: Fixed in bff8971c. Added an `areSettingsLoaded` boolean that flips to `true` only on a successful GET response. The form (toggle + secret input + Save button) is now only rendered when both `! isLoadingSettings` and `areSettingsLoaded` are true; on failure we surface a `Notice` with the GET error and the form stays hidden, so there's no way to POST stale defaults back. `handleSaveSettings()` also has a defensive early `return` for the same reason. New Jest test `does not render the settings form when the GET /settings request fails` asserts the form is locked, the error notice renders, and no destructive POST is issued. All 30 panel Jest tests pass.

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
```diff
@@ -93,6 +113,35 @@ public function register_routes(): void {
...
+						],
+						'webhook_secret' => [
+							'description'       => __( 'Webhook signing secret for Agentic Commerce delegated checkout events.', 'woocommerce-gateway-stripe' ),
+							'type'              => 'string',
+							'validate_callback' => 'rest_validate_request_arg',
+						],
+					],
+				],
+			]
+		);
```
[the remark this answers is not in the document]
wjrosa: Fixed in bff8971c. Added a private `is_merchant_enabled()` check that reads `WC_Stripe_Agentic_Commerce_Integration::ENABLED_OPTION` and gates `trigger_sync()` with a 409 response and a new `stripe_agentic_commerce_disabled` error code when the toggle is off. Scope: the gate is on the **write** path (manual sync) — `get_status()` is left as a read-only endpoint so admins can still inspect history after disabling. New regression test `test_trigger_sync_returns_409_when_merchant_toggle_off` seeds `ENABLED_OPTION = 'no'`, fires the route, and asserts the 409 + the error code AND that the Stripe HTTP layer was never called. The set-up fixture for existing tests now sets `ENABLED_OPTION = 'yes'` so they continue to exercise the success path; `test_get_settings_returns_defaults` deletes it explicitly to assert the off-by-default response.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -125,7 +125,7 @@ public function check_for_webhook() {
 		$is_agentic_hook = 0 === strpos( $event_type, 'v1.delegated_checkout.' );
 
 		$secret = $is_agentic_hook
-			? ( defined( 'AGENTIC_COMMERCE_WEBHOOK_SECRET' ) ? AGENTIC_COMMERCE_WEBHOOK_SECRET : '' )
+			? (string) get_option( WC_REST_Stripe_Agentic_Commerce_Controller::WEBHOOK_SECRET_OPTION, '' )
 			: $this->secret;
```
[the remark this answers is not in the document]
wjrosa: Fixed in bff8971c. Took option (a) from the suggestion: moved `WEBHOOK_SECRET_OPTION` to `WC_Stripe_Agentic_Commerce_Integration`. That class is in the Composer autoload classmap (`includes/agentic-commerce/`), so the const is reachable from any request — including webhook deliveries that come in via `woocommerce_api_wc_stripe` and never trigger `rest_api_init`. The webhook handler now reads `WC_Stripe_Agentic_Commerce_Integration::WEBHOOK_SECRET_OPTION` instead. The controller still references it (in `get_agentic_settings` / `update_agentic_settings`) and `MASKED_WEBHOOK_SECRET` stays on the controller since it is UI-presentation only. All test references and the JS test fixture comment were updated.

client/settings/agentic-commerce/index.js
```diff
@@ -1,626 +1,305 @@
...
-					__(
-						'Failed to load sync status.',
-						'woocommerce-gateway-stripe'
-					),
+				path: '/wc/v3/wc_stripe/agentic-commerce/settings',
 			} );
+			setIsFeatureEnabled( result.is_enabled );
+			setWebhookSecret( result.webhook_secret ?? '' );
+		} catch {
+			// Settings fetch failure is non-fatal; defaults remain.
```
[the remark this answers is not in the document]
wjrosa: Already addressed in bff8971c — same root cause as the CodeRabbit Major thread on this hunk, fixed alongside it. Implementation is structurally identical to your suggestion, just inverted polarity: an `areSettingsLoaded` boolean flips to `true` only on a successful GET response. The form (toggle, secret input, Save button) is conditionally rendered on `! isLoadingSettings && areSettingsLoaded`, the Save button's `disabled` includes `! areSettingsLoaded`, and `handleSaveSettings()` has a defensive early `return` for the same reason. The error notice now surfaces the GET failure instead of being silently swallowed.

New Jest test `does not render the settings form when the GET /settings request fails` asserts the form is locked, the error notice renders, and **no destructive POST is issued** — which is exactly the regression you flagged.

PR #5204, opened 2026-03-31
Fix: Ensure cart is cleared after successful payment in process_payment_with_payment_method

kaushikasomaiya:
Closes STRIPE-880

## Summary
- Adds \`WC()->cart->empty_cart()\` in \`process_payment_with_payment_method\` immediately after a successful payment, before returning the success response
- This prevents duplicate orders when a hook throws an exception or corrupts the JSON response after payment is complete but before the client-side redirect lands on the order-received page (where WooCommerce core normally clears the cart)
- Adds a PHPUnit test asserting \`empty_cart()\` is called exactly once on a successful payment

### Scenario 1 — Classic checkout: hook throws an exception after charge

Deploy this snippet, then checkout with a test card on classic checkout:

```
add_action( 'wc_gateway_stripe_process_response', function() {
    throw new \Exception( 'no idea what went wrong' );
} );
```

**Before fix:** An error is shown. Retrying places a second order — both are successfully charged (duplicate order).

**After fix:** An error is still shown, but the cart is cleared. Retrying checkout starts with an empty cart, preventing a duplicate charge.

---

### Scenario 2 — Block checkout: hook outputs content that breaks the JSON response

Deploy this snippet, then checkout with a test card on block checkout:

```
add_action( 'woocommerce_payment_complete', function() {
    echo 'unwanted output  - breaks the JSON';
} );
```

**Before fix:** Block checkout fails to parse the malformed JSON and shows _"Something went wrong when placing the order."_ Retrying places a second successful order.

**After fix:** Same error is shown, but the cart is already cleared server-side. Retrying starts with an empty cart, preventing a duplicate charge.

---

- [ ] Confirm Scenario 1 no longer results in duplicate orders
- [ ] Confirm Scenario 2 no longer results in duplicate orders
- [ ] PHPUnit: \`test_process_payment_with_payment_method_clears_cart_on_success\` passes in CI


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1490,6 +1490,10 @@ private function process_payment_with_payment_method( int $order_id ) {
 
 			$order_helper->unlock_order_payment( $order );
 
+			if ( WC()->cart ) {
```
kaushikasomaiya, line 1550: I think this is still not the ideal place to clear cart.. eg exceptions happening before it will not allow code to reach here. Though what this can help with is the scenarios where eg on Block Checkout corrupted JSON causes redirect fail

kaushikasomaiya: Perhaps it can be before this; however I am afraid it needs to be checked by Quark team if this can cause any other unwanted issues. https://github.com/woocommerce/woocommerce-gateway-stripe/blob/ae7241bdcba0eebe6ff12e3dca08537ba0dfe60d/includes/abstracts/abstract-wc-stripe-payment-gateway.php#L652

zhongruige: Agree, I plan to tag in Quark for a final review before we merge as well, especially for payment gateway plugins so we're extra careful.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 = 10.6.0 - xxxx-xx-xx =
+* Fix - Ensure cart is cleared after successful payment in process_payment_with_payment_method
```
wjrosa: The entry in readme.txt is missing (use this same entry)

zhongruige: Thanks, happened to be doing some other adjustments here so made this change, too.

zhongruige (approved):
Changes are looking good on my end, but let's make sure someone from Quard reviews and approves this before we merge anything in @kaushikasomaiya. Nice work!

daledupreez:
@coderabbitai resume

malithsen (commented):
> Scenario 1 — Classic checkout: hook throws an exception after charge

This scenario still fails for me. The cart doesn't clear for me after the error.

I also noticed an unexpected behavior in the 3DS flow. If 3DS verification fails, this branch clears the cart.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
malithsen: Yeah, I agree this likely isn't the right place for it. I left a couple of observations in a top-level [comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5204#pullrequestreview-4438890108), which are both results of this placement. 

I haven't tested this, but what if we use the `woocommerce_payment_complete` hook to clear the cart? It runs before `wc_gateway_stripe_process_response` and should prevent the issues I noted.

kaushikasomaiya: > I also noticed an unexpected behavior in the 3DS flow. If 3DS verification fails, this branch clears the cart.

@malithsen great catch! I totally missed this flow. 

>  but what if we use the woocommerce_payment_complete hook to clear the cart? 

Yes I think if we can limit this to just when Stripe payment is involved and Stripe calls payment complete, that would be great. 

There's also another place where cart clearing can be done  https://github.com/woocommerce/woocommerce-gateway-stripe/blob/f4dbee44e097f6aaff986fa6146947a57e37f3ea/includes/abstracts/abstract-wc-stripe-payment-gateway.php#L667 although it needs proper assessment of what can be the side effects if any because it is probably a shared method.

Or do we just discard this PR and issue and leave things as is?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1717,6 +1725,10 @@ private function process_payment_with_confirmation_token( int $order_id ) {
 
 			$return_url = $this->get_return_url( $order );
 
+			if ( WC()->cart ) {
+				WC()->cart->empty_cart();
```
wjrosa, line 1729: Extending @malithsen's 3DS catch to this path: `process_payment_with_confirmation_token()` returns `success` (and now clears the cart) even when the intent has no charge yet — the `if ( $charge )` guard above implies a 3DS-pending intent reaches here with no charge. If so, a failed authentication leaves the cart already empty here too, same as the payment-method branch.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -5888,6 +5888,122 @@ public function test_add_email_currency_conversion_notice_outputs_plain_text_for
...
+			->willReturn( $mock_payment_method );
+
+		$mock_cart = $this->getMockBuilder( 'WC_Cart' )
+			->disableOriginalConstructor()
+			->onlyMethods( [ 'empty_cart' ] )
+			->getMock();
+
+		$mock_cart->expects( $this->once() )->method( 'empty_cart' );
+
+		WC()->cart = $mock_cart;
```
wjrosa, line 5946: These tests replace the global `WC()->cart` with a mock but never restore it, so the mock leaks into later tests in the run (the `empty_cart()` `once()` expectation can also trip a base tearDown that empties the cart). Worth stashing the original and restoring it in a `finally`/tearDown. Same for the other new test at L6000.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -5888,6 +5888,122 @@ public function test_add_email_currency_conversion_notice_outputs_plain_text_for
...
+
+		$response = $this->mock_gateway->process_payment( $order_id );
+
+		$this->assertEquals( 'success', $response['result'] );
+  }
```
wjrosa, line 6005: Nit: this brace is indented with spaces; the rest of the file uses a tab (cf. L5951). PHPCS will flag it.

wjrosa (commented):
A few notes that I didn't see covered in the existing threads (steering clear of the placement / `woocommerce_payment_complete` discussion and the 3DS-on-the-payment-method-branch catch already raised).

daledupreez:
@kaushikasomaiya, I just want to follow up to see if you plan to continue work on this PR in the near future. If you do, please let us know, otherwise we'd like to close the PR so we can make sure can focus on active PRs.

kaushikasomaiya:
Hey @daledupreez I think we can close this one because there are some blockers described above. That said, I'd like to highlight that we regularly receive duplicate order issues described in STRIPE-858 - it would be great if either cart can be cleared properly on successful payment or we deploy some mechanism to catch duplicate orders/charge on retries. I might come back with additional analysis later but this one is Ok to be closed.

daledupreez:
I fully agree that we need to prevent duplicate orders! But I'll go ahead and close this PR for now.

PR #5190, opened 2026-03-30
Fix TypeError when deferred Stripe webhooks run via Action Scheduler

Mayisha:
Fixes STRIPE-1004
Fixes #5184 

Action Scheduler serializes scheduled job arguments. For `wc_stripe_deferred_webhook`, the Stripe webhook stdClass becomes nested arrays on load. `process_deferred_webhook()` then passed that value to `run_webhook_received_action( string $webhook_type, object $notification )`, which triggered a TypeError and failed every deferred job (notably redirect flows such as iDEAL / Klarna).

## Changes proposed in this Pull Request:
- Add `normalize_deferred_webhook_notification_to_object()` method to turn deserialized array payloads back into an object
- Call it at the start of `process_deferred_webhook()` to ensure the data passed to `run_webhook_received_action` is of correct type.

## Testing instructions
- Enable iDEAL or Klarna as a payment method
- Place a test order using iDEAL or Klarna
- Check out to `develop` branch.
- Complete the payment on the bank redirect page
- Go to **WooCommerce > Status > Action Scheduler**
- Search for the `wc_stripe_deferred_webhook` action
- The action should be failed with the TypeError message above
- Check out to this branch and follow the same steps.
-  `wc_stripe_deferred_webhook` actions should be successful now.

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5205, opened 2026-03-31
Fix: normalize array notification in deferred webhook handler (STRIPE-821)

stuartduff:
## Summary

Fixes a `TypeError` that caused deferred webhook jobs to fail and be retried indefinitely by Action Scheduler.

**Root cause:** Action Scheduler serializes job args to JSON and deserializes them with `json_decode( $args, true )` — the `true` flag converts all JSON objects to PHP associative arrays. This means the `stdClass` webhook notification stored by `defer_webhook_processing()` always arrives at `process_deferred_webhook()` as a PHP array. That array was passed unchanged to `run_webhook_received_action()`, which declares a strict `object` type hint, throwing a `TypeError`. Because `TypeError` extends `Error` (not `Exception`), it also bypassed the `catch ( Exception $e )` block, so Stripe's error logger never fired and Action Scheduler received a bare fatal — retrying the job indefinitely.

**Changes:**
- Normalize array `$notification` to `stdClass` via `json_encode()` + `json_decode()`, with a defensive `is_string()` guard on the encode result and `is_object()` guard on the decode result
- Widen `catch ( Exception $e )` → `catch ( \Throwable $e )` so `TypeError` and other `Error` subclasses are logged by `WC_Stripe_Logger` before being re-thrown to Action Scheduler
- Add `is_object( $notification )` guard before calling `run_webhook_received_action()` (also resolves a PHPStan type error and removes a stale baseline entry)
- Update `@param` docblock to reflect that arrays are accepted and normalized

**Test coverage:**
- `test_process_deferred_webhook_normalizes_array_notification` — simulates exact Action Scheduler deserialization via `json_decode( json_encode( $data ), true )`, verifies no `TypeError` is thrown and `wc_stripe_webhook_received` action receives an `object`
- `test_process_deferred_webhook_null_notification_throws_expected_exception` — covers the `null` notification path through the unsupported-type exception route

Fixes [STRIPE-821](https://linear.app/a8c/issue/STRIPE-821/investigate-array-notification-passed-to-stripe-webhook-handler)

Related Zendesk tickets: #10691273, #10472513, #10460364

## Test plan

- [ ] Run `npm run test:php -- --filter WC_Stripe_Webhook_Handler_Test` — 36/36 pass
- [ ] Run `npm run phpstan` — no errors
- [ ] Verify in a test store that deferred payment intent webhooks process correctly (trigger a `payment_intent.succeeded` and confirm the Action Scheduler job completes without error)

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1307,11 +1307,23 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	 *
 	 * Deferred webhooks are scheduled by @see defer_webhook_processing().
 	 *
-	 * @param string $webhook_type    The webhook event name/type.
-	 * @param array  $additional_data Additional data passed to the scheduled job.
-	 * @param stdClass $notification  The webhook notification payload.
+	 * @param string         $webhook_type    The webhook event name/type.
+	 * @param array          $additional_data Additional data passed to the scheduled job.
+	 * @param stdClass|array $notification    The webhook notification payload. Action Scheduler
+	 *                                        deserializes stored args via json_decode( $args, true ),
+	 *                                        so this may arrive as an array and is normalized to stdClass.
 	 */
 	public function process_deferred_webhook( $webhook_type, $additional_data, $notification = null ) {
+		// Action Scheduler deserializes stored args with json_decode( $args, true ), which converts
+		// stdClass objects to associative arrays. Re-hydrate to stdClass recursively so downstream
+		// code (and the strict object type hint on run_webhook_received_action) receives an object.
+		if ( is_array( $notification ) ) {
+			$normalized = json_decode( json_encode( $notification ) );
+			if ( is_object( $normalized ) ) {
+				$notification = $normalized;
+			}
+		}
+
 		try {
 			switch ( $webhook_type ) {
 				case 'payment_intent.succeeded':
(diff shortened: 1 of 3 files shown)


wjrosa:
Is this one also ready for review?

kaushikasomaiya:
@stuartduff @wjrosa This might be duplicate of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5190/changes

stuartduff:
Hey @kaushikasomaiya, the linear issue I picked up was still marked as todo. However, that [STRIPE-1004](https://linear.app/a8c/issue/STRIPE-1004) does look the same as [STRIPE-821](https://linear.app/a8c/issue/STRIPE-821/investigate-array-notification-passed-to-stripe-webhook-handler) :/

wjrosa:
Do you see something this one covers and the other does not? If so, you can target the other branch and keep the relevant changes

daledupreez:
@Mayisha, pinging you to work out whether this duplicates or overlaps with your fix in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5190.

Mayisha:
> @Mayisha, pinging you to work out whether this duplicates or overlaps with your fix in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5190.

@daledupreez it is a duplicate and the fix is logically the same (syntactically a bit different). We can close this PR as the other one https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5190 is already shipped.

PR #5209, opened 2026-04-01
Fallback to split PE from OCS when Stripe is not the first gateway

Mayisha:
Fixes STRIPE-1028

- Added helper methods
    - `record_first_gateway_id_from_available_list`
    - `clear_first_available_payment_gateway_record`
    - `is_stripe_gateway_first_in_available_list`
- Used `wc_payment_gateways_initialized` action to call `record_first_gateway_id_from_available_list`. We are setting the static variable only if it is empty. (Note: used cache instead of static variable at first, but coderabbit flagged some legit issues, so moved to static variable)
- Added the check in the relevant backend payment flow also where only `is_oc_enabled` was checked previously. (to prevent showing split in frontend and using oc flow in backend)

## Testing instructions
- Enable a few Stripe payment methods.
- Go to **WooCommerce > Settings > Payment methods**.
- Make Stripe the first method on the list.
- Verify that Stripe methods are loaded using the dynamic elements (**OCS**)
- As the admin go to **WooCommerce > Settings > Payment methods** and move Stripe to the non-first position (2nd, 3rd or last).
- As a shopper, verify that Stripe methods are loaded using split payment elements (**not OCS**)
- Verify that the available list of payment methods is correct.
- Now disable OCS from the Stripe settings page.
- Verify as a shopper that the list of payment methods rendered on the checkout page is correct and in the correct order.
- Test for both Stripe in the first position and other positions in the settings.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha:
@coderabbitai review

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -121,20 +121,28 @@ abstract class WC_Stripe_UPE_Payment_Method extends WC_Payment_Gateway {
...
-		$this->supports_deferred_intent = true;
-		$this->oc_enabled               = WC_Stripe_Feature_Flags::is_oc_available() && 'yes' === $this->get_option( 'optimized_checkout_element' );
+		$this->enabled                           = $is_stripe_enabled && in_array( static::STRIPE_ID, $this->get_upe_enabled_payment_method_ids(), true ) ? 'yes' : 'no'; // @phpstan-ignore-line (STRIPE_ID is defined in classes using this class)
+		$this->id                               = WC_Stripe_UPE_Payment_Gateway::ID . '_' . static::STRIPE_ID; // @phpstan-ignore-line (STRIPE_ID is defined in classes using this class)
+		$this->has_fields                        = true;
+		$this->testmode                          = WC_Stripe_Mode::is_test();
+		$this->supports                          = [ PaymentGatewayFeature::PRODUCTS, PaymentGatewayFeature::REFUNDS ];
+		$this->supports_deferred_intent          = true;
+		$this->oc_enabled                        = WC_Stripe_Feature_Flags::is_oc_available() && 'yes' === $this->get_option( 'optimized_checkout_element' );
+		$this->is_first_available_payment_method = WC_Stripe_Helper::is_stripe_first_available_gateway( WC_Stripe_UPE_Payment_Gateway::ID );
```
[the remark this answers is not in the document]
Mayisha: Reverted the use of local variable and using helper method directly.

includes/class-wc-stripe.php
```diff
@@ -328,6 +328,9 @@ public function init() {
 		// cards when the Optimized Checkout is enabled, etc.
 		add_action( 'wc_payment_gateways_initialized', [ $this, 'maybe_toggle_payment_methods' ] );
 
+		// Cache the first available payment gateway.
+		add_filter( 'woocommerce_available_payment_gateways', [ 'WC_Stripe_Helper', 'cache_first_available_payment_gateway' ], 99999 );
```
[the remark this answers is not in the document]
Mayisha: Using static variable instead of cache.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -610,7 +612,8 @@ static function ( $domain ) {
 		}
 
 		// Optimized Checkout feature flag + setting + whether we are on any of the pages that should not show OC.
-		$should_show_optimized_checkout                 = $this->oc_enabled && $this->is_valid_optimized_checkout_page();
+		// Keep OCS gated to cases where Stripe is the first available gateway.
+		$should_show_optimized_checkout                 = $this->should_use_optimized_checkout_payment_method_layout();
 		$stripe_params['isOCEnabled']                   = $should_show_optimized_checkout;
 		$stripe_params['shouldShowOptimizedCheckout']   = $should_show_optimized_checkout;
 		$stripe_params['shouldExpandOptimizedCheckout'] = $should_show_optimized_checkout && WC_Stripe_Feature_Flags::should_expand_ocs_in_legacy_checkout();
```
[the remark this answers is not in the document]
Mayisha: Fixed

wjrosa:
Hey @Mayisha ! Is this ready for review? Asking because it is in draft

Mayisha:
> Hey @Mayisha ! Is this ready for review? Asking because it is in draft

@wjrosa this is ready for review.

includes/class-wc-stripe-helper.php
```diff
@@ -740,6 +747,51 @@ public static function get_upe_settings_enabled_payment_method_ids( $gateway ) {
...
+		self::$first_gateway_id_from_available_list = null;
+	}
+
+	/**
+	 * Memoizes the first gateway ID from the available gateways list (woocommerce_available_payment_gateways).
+	 *
+	 * @param array<string, WC_Payment_Gateway> $gateways Available payment gateways keyed by gateway ID.
+	 * @return array<string, WC_Payment_Gateway>
+	 */
+	public static function record_first_gateway_id_from_available_list( $gateways ) {
```
wjrosa:
```suggestion
	public static function record_first_gateway_id_from_available_list( array $gateways ): array {
```

includes/class-wc-stripe-helper.php
```diff
@@ -740,6 +747,51 @@ public static function get_upe_settings_enabled_payment_method_ids( $gateway ) {
...
+		return $gateways;
+	}
+
+	/**
+	 * Whether the Stripe UPE gateway is first among WooCommerce's currently available payment gateways.
+	 *
+	 * @param string|null $stripe_gateway_id Main Stripe gateway id. Defaults to `stripe` (see WC_Stripe_UPE_Payment_Gateway::ID).
+	 * @return bool
+	 */
+	public static function is_stripe_first_available_gateway( $stripe_gateway_id = null ) {
```
wjrosa:
```suggestion
	public static function is_stripe_first_available_gateway( ?string $stripe_gateway_id ): bool {
```

includes/class-wc-stripe-helper.php
```diff
@@ -740,6 +747,51 @@ public static function get_upe_settings_enabled_payment_method_ids( $gateway ) {
...
+		if ( null === $stripe_gateway_id ) {
+			$stripe_gateway_id = 'stripe';
```
wjrosa: I would consider using the existing constant `WC_Stripe_UPE_Payment_Gateway::ID`

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1362,7 +1374,7 @@ private function process_payment_with_payment_method( int $order_id ) {
 			$upe_payment_method            = $this->payment_methods[ $selected_payment_type ] ?? null;
 			$response_args                 = [];
 
-			if ( $this->oc_enabled && isset( $payment_method_details->type ) ) {
+			if ( $this->oc_enabled && WC_Stripe_Helper::is_stripe_first_available_gateway( self::ID ) && isset( $payment_method_details->type ) ) {
```
wjrosa: Since this is repeated so many times, I would add a helper method for it

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -780,6 +781,17 @@ public function is_valid_optimized_checkout_page(): bool {
...
 	}
 
+	/**
+	 * Whether Optimized Checkout layout should be used for this request.
+	 *
+	 * When false, split Payment Element path is used.
+	 *
+	 * @return bool
+	 */
+	public function should_use_optimized_checkout_payment_method_layout(): bool {
```
wjrosa, line 809: Maybe this method can be used in the remaining places

includes/class-wc-stripe-helper.php
wjrosa: Claude comment:
> The method takes a `$stripe_gateway_id` parameter (defaulting to 'stripe') and checks whether that ID is first — it's a generic "is this gateway first?" check. The name implies it always checks for Stripe specifically, but it's called with `self::ID` throughout the gateway class. A clearer name: `is_gateway_first_in_available_list( $gateway_id )`. Minor, but worth noting since this becomes a stable API in `WC_Stripe_Helper`.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -212,7 +212,7 @@ public function is_available() {
 		}
 
 		// When OC is enabled _and_ we are on a page where OC is permitted, we use the OC payment container to render all the methods.
-		if ( $this->oc_enabled && $main_stripe_gateway->is_valid_optimized_checkout_page() ) {
+		if ( $this->oc_enabled && $main_stripe_gateway->is_valid_optimized_checkout_page() && WC_Stripe_Helper::is_stripe_first_available_gateway( WC_Stripe_UPE_Payment_Gateway::ID ) ) {
```
wjrosa: Claude comment:
> `is_available()` on sub-payment-methods (`stripe_card`, `stripe_sepa`, etc.) is called during the loop that builds the `woocommerce_available_payment_gateways` list. The `record_first_gateway_id_from_available_list hook` fires at priority 99999, i.e. after that loop. So on the first `get_available_payment_gateways()` call in any request:

>  1. `is_available()` runs → `is_stripe_first_available_gateway()` → static is null → returns false
>  2. The OCS collapsing block is skipped → individual Stripe sub-methods are marked available
>  3. Then the filter fires at 99999 → static is set

>  The result on the first call: split PE methods appear in the WC payment list at the same time the main stripe gateway's `payment_fields()` renders the OCS container (because `should_use_optimized_checkout_payment_method_layout()` is called after the list is built and now gets the correct value). This creates a double-render situation or incorrect method list on a fresh request. Pre-existing behavior (`$this->oc_enabled` && `$main_stripe_gateway->is_valid_optimized_checkout_page()`) had no timing issue because those properties are available immediately.

>  Suggestion: Instead of gating on `is_stripe_first_available_gateway()` inside `is_available()`, consider connecting the collapsing logic to a hook that runs after `woocommerce_available_payment_gateways` has fired (e.g., priority 100000+ on the same filter to prune the list), or populate the static earlier (e.g., on `wc_payment_gateways_initialized`).

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -780,6 +781,17 @@ public function is_valid_optimized_checkout_page(): bool {
...
+		return $this->oc_enabled && $this->is_valid_optimized_checkout_page() && WC_Stripe_Helper::is_stripe_first_available_gateway( self::ID );
+	}
+
 	/**
```
wjrosa, line 840: Claude comment:

> File: `includes/payment-methods/class-wc-stripe-upe-payment-gateway.php`, `set_payment_method_title_for_order` (~line 2394) and `create_token_from_setup_intent` (~line 2621):

> `woocommerce_available_payment_gateways` does not fire during webhook processing (payment_intent.succeeded, etc.) or WooCommerce Subscriptions cron renewals. In those contexts `$first_gateway_id_from_available_list` is always null, so `is_stripe_first_available_gateway()` always returns false.

>  Concrete impact:

>  - `set_payment_method_title_for_order` is called from the webhook handler. If the order was placed via OCS (`$payment_method_type === WC_Stripe_Payment_Methods::OC`), the guard now fails → `$payment_methods[OC]` is never created → payment title is not set correctly on the order.
>  - create_token_from_setup_intent won't resolve the payment method instance via type in OCS flows called outside a frontend request.
>  - Subscription renewal payments through OCS-placed subscriptions will hit the wrong code paths in `prepare_payment_information_from_request`.

>  The original `$this->oc_enabled` check was correct for these flows because it reflects a stored setting, not a per-request computation. You could separate the concern: `$this->oc_enabled` for "was this order placed via OCS?" (safe in all contexts), and `is_stripe_first_available_gateway()` only for "should we render OCS on this checkout?" (frontend only).

includes/class-wc-stripe-helper.php
Mayisha: Renamed with `is_stripe_gateway_first_in_available_list`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: I assume you meant using this method to avoid the repetitions you flagged [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5209#discussion_r3024763580). This method is not suitable for those cases, as it checks the 'is_valid_optimized_checkout_page' which is irrelevant and might give us a wrong value in the backend payment processing code flow.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Good call. I am creating a helper for this.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: > The original $this->oc_enabled check was correct for these flows because it reflects a stored setting, not a per-request computation. You could separate the concern: $this->oc_enabled for "was this order placed via OCS?" (safe in all contexts), and is_stripe_first_available_gateway() only for "should we render OCS on this checkout?" (frontend only).

This is a good suggestion. Let me see if I can apply this.

Mayisha: I reviewed this suggestion, and it would require a bit of refactoring to mark "was this order placed via OCS?" and detect it across all paths.
I have fixed the check in `set_payment_method_title_for_order` and reviewed the usage of the new `is_oc_active` helper. I think this is sufficient.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
Mayisha: Addressed.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -797,6 +798,30 @@ public function is_valid_optimized_checkout_page(): bool {
...
 
+	/**
+	 * Whether Optimized Checkout layout should be used for this request.
+	 *
+	 * When false, split Payment Element path is used.
+	 *
+	 * @return bool
+	 */
+	public function should_use_optimized_checkout_payment_method_layout(): bool {
+		return $this->oc_enabled && $this->is_valid_optimized_checkout_page() && WC_Stripe_Helper::is_stripe_gateway_first_in_available_list( self::ID );
```
wjrosa: This logic is being repeated across the code, I would consider creating a new method just for it

malithsen (commented):
@Mayisha If Stripe is the first "enabled gateway" but not the first one on the list, shouldn't we use OCS? 

Image

Right now, when the gateway list is like the above, OCS is not enabled.

wjrosa (changes requested):
Code is good and almost works as intended. It works the first time, but after putting Stripe as the first gateway again, I don't see OCS anymore.

includes/class-wc-stripe-helper.php
```diff
@@ -740,6 +747,49 @@ public static function get_upe_settings_enabled_payment_method_ids( $gateway ) {
...
+		if ( null !== self::$first_gateway_id_from_available_list ) {
+			return;
+		}
+
+		$gateways = $gateways->payment_gateways();
+
+		if ( is_array( $gateways ) && [] !== $gateways ) {
+			self::$first_gateway_id_from_available_list = array_key_first( $gateways );
+		}
+	}
```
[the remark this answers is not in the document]
Mayisha: I will loop through the list to find the first available one.

Mayisha:
> @Mayisha If Stripe is the first "enabled gateway" but not the first one on the list, shouldn't we use OCS?

@malithsen I have updated the `record_first_gateway_id_from_available_list` method to consider `is_available()` as well.

includes/class-wc-stripe-helper.php
```diff
@@ -740,6 +747,58 @@ public static function get_upe_settings_enabled_payment_method_ids( $gateway ) {
...
+
+	/**
+	 * Whether the Stripe UPE gateway is first among WooCommerce's currently available payment gateways.
+	 *
+	 * @param string|null $stripe_gateway_id Main Stripe gateway id. Defaults to `stripe` (see WC_Stripe_UPE_Payment_Gateway::ID).
+	 * @return bool
+	 */
+	public static function is_stripe_gateway_first_in_available_list( ?string $stripe_gateway_id = null ): bool {
```
malithsen: Does this function need a parameter or can we always rely on `WC_Stripe_UPE_Payment_Gateway::ID` within the function?

malithsen (approved):
Thanks for the fixes. This works as described. Left a non-blocking comment.

includes/class-wc-stripe-helper.php
Mayisha: Right. This is actually redundant. 👀

wjrosa (approved):
Code is good and works as expected now 👍 

PS: We need to fix the unit tests before merging

PR #5215, opened 2026-04-01
Fix: update CC expiry metadata when re-adding a saved card

### LINKED ISSUE #4991: New Expiry not shown for saved cards
### Description

We are having issues where a user adds a new payment method and its not fully reflected in my account. We have an issue where the user has a card that expires 02/26, they have been issued a new card that expires in 02/31. 

We can see in the logs and stripe that the user has added a new card (5 times), but do to the check if the payment method is new, only checking the fingerprint. Its not updating the meta data, only the pm_ token

This seems to relate to - d https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4077#issuecomment-2790183666

### Steps to reproduce

I can reproduce this on a staging environment, by

Log into my account and add a card that expires on 01/27

Image

Add another card but with a later expiry

Image

When i look back at my saved payments i see only the initial date

Image

Even though in the logs, i can see a new pm for all of them (there is 3, i only mention 2 here)

Image
Its correct in db

Image

### Expected behavior

I would expect the expiry date to be updated, if its the same "card", as this does change. This has resulted in lots of confusion by both customer and site admins. 

### System report

I can replicate this on any site running woocommerce 9.9.6 (even 9.9.3) and stripe gateway 10.3.1

nicdwilson:
## Summary

Fixes #4991.

When a customer re-adds a card that was already saved (same card number → same Stripe fingerprint), only the Stripe payment method ID was being updated. The expiry month, expiry year, and card type were never refreshed, so the **My Account → Payment Methods** page continued to show the old expiry date even after the card was reissued with a new one.

- When a duplicate CC token is found by fingerprint, also check whether `exp_month`, `exp_year`, or `card_type` have changed and update them if so
- Consolidate `clear_cache()` + `save()` behind a single `$token_updated` flag to avoid unnecessary writes
- Zero-pad `exp_month` before comparison to match WooCommerce's stored format (e.g. `2` → `'02'`), preventing spurious updates
- Fix orphan-cleanup bug: when `add_token_to_user` updates a token's Stripe payment method ID (re-issued card with new `pm_*` ID), the old key remained in `$stored_tokens`, causing the orphan-cleanup loop to immediately delete the just-updated token. After `add_token_to_user` returns, we now scan `$stored_tokens` by WC token ID and remove any stale old-key entry before the cleanup runs.

## Test plan

- [ ] `npm run test:php -- --filter WC_Stripe_Payment_Tokens` passes (36 tests, 68 assertions)
- [ ] Two new PHPUnit tests added:
  - `test_add_token_to_user_updates_cc_expiry_on_duplicate` — verifies expiry is updated when the same card is re-added with a new expiry date
  - `test_add_token_to_user_does_not_update_when_cc_metadata_unchanged` — verifies no unnecessary save/cache-clear when metadata is already current
- [ ] Manual: save a card, re-add the same card number with a different expiry date, confirm My Account shows the updated expiry
- [ ] Manual: re-add a card that returns a new `pm_*` ID (re-issued card), confirm the saved token persists and is not deleted


zhongruige (approved):
Looks good to me! Let's make sure Quark reviews this first before we merge anything in.

gin0115:
@nicdwilson thanks for running with this :)

diegocurbelo (commented):
I think this use case was covered/fixed by https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5308, right?

nicdwilson:
Looks like it.

PR #5216, opened 2026-04-01
Fix STRIPE-843: Remove GBP from Alipay supported currencies

Snaacks:
## Summary

- GBP was listed as a supported currency for Alipay on GB Stripe accounts, but Stripe's API rejects it with: *"The currency provided (gbp) is invalid. Payments with alipay support the following currencies: cny, eur."*
- This removes GBP from Alipay's `$supported_currencies` list and drops the `GB` case from `get_supported_currencies()`, so UK stores with GBP currency no longer see Alipay at checkout — preventing the error entirely.
- UK stores where the customer pays in CNY continue to work as before.

## Bug

[STRIPE-843: Fix Alipay currency support](https://linear.app/a8c/issue/STRIPE-843/fix-alipay-currency-support)

## Changes

- `includes/payment-methods/class-wc-stripe-upe-payment-method-alipay.php`: removed `POUND_STERLING` from `$supported_currencies` and removed `case 'GB'` from `get_supported_currencies()` (GB now falls to `default` → `[CNY]`)
- `tests/phpunit/PaymentMethods/WC_Stripe_UPE_Payment_Method_Alipay_Test.php`: new test file covering `get_supported_currencies()` for GB, US, AU, FR, and unknown countries, plus a direct assertion that GBP is not in the supported list

## Testing

- New PHPUnit tests added and passing locally (run `npm run test:php` with Docker)
- No manual test environment — relying on QIT automated regression tests + code review

## Notes


wjrosa:
Hi @Snaacks ! Is this PR ready for review?

Snaacks:
@wjrosa Hey! It is ready for review. Apologies, I got sidetracked and forgot to comment after finishing with this.

wjrosa (dismissed):
Looks good to me. But you are missing changelog entries in changelog.txt and readme.txt (add them before merging the PR)

includes/payment-methods/class-wc-stripe-upe-payment-method-alipay.php
```diff
@@ -57,9 +56,6 @@ public function get_supported_currencies() {
 			case 'CA':
 				$currency = [ WC_Stripe_Currency_Code::CANADIAN_DOLLAR, WC_Stripe_Currency_Code::CHINESE_YUAN ];
 				break;
-			case 'GB':
-				$currency = [ WC_Stripe_Currency_Code::POUND_STERLING, WC_Stripe_Currency_Code::CHINESE_YUAN ];
-				break;
```
Mayisha, line 62: The error says `The currency provided (gbp) is invalid. Payments with alipay support the following currencies: cny, eur.`

So we should still keep this block but change the POUND_STERLING with EURO.

I have tested and verified that the payment element loads fine if the store currency is CNY and Euro.

Mayisha (changes requested):
I think we should update the supported currency list for `GB` instead of removing it completely.

diegocurbelo (changes requested):
The error quoted in the PR description is: “Payments with alipay support the following currencies: cny, eur”. That `CNY`, `EUR` set is what Stripe returns for a EUR-settlement account. A GB-settlement account returns `CNY`, `GBP`. 

So the account that produced the error wasn’t settling in GBP (it was using EUR), removing the plugin’s GB mapping is not the correct fix, that would drop Alipay for every UK store on GBP; including the ones where it works today.


The base `get_supported_currencies()` runs its result through `wc_stripe_{$id}_upe_supported_currencies`, but 

A solution could be to modify `get_supported_currencies()` to call `parent::...` so a store with an unusual account setup can use the filter `wc_stripe_{$id}_upe_supported_currencies` to override the supported currencies.

daledupreez:
@Snaacks, given the contents of [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5216#pullrequestreview-4987798302) regarding the merchant's settlement currency, I don't think we will be in a position to merge this PR. Please feel free to re-open the PR if you think we should tackle this differently, and we will definitely take another look.

PR #5220, opened 2026-04-02
Add product deletion tracking to Agentic Commerce inventory tracker

wjrosa:
Fixes STRIPE-1036

## Summary

- Extends `WC_Stripe_Agentic_Commerce_Inventory_Tracker` to track product deletions and trash events in addition to stock changes
- Hooks into `woocommerce_before_delete_product` and `woocommerce_trash_product` — both trigger `track_product_archive()`
- Batches pending archives in `wc_stripe_agentic_pending_archives` and syncs them via a new Action Scheduler action (`wc_stripe_agentic_commerce_sync_archives`) using the same 60-second delay pattern as inventory updates
- Sends a `product_catalog_feed` ImportSet with `id` and `availability: out_of_stock`, marking products as unavailable on Stripe without permanently removing them from the catalog
- Removes a product from pending inventory updates when it is deleted or trashed, since the stock quantity is no longer relevant

### Automated

- [ ] Run `npm run up && npm run test:php -- --filter=WC_Stripe_Agentic_Commerce_Inventory_Tracker_Test` to confirm all tests pass
- [ ] `test_register_hooks_attaches_archive_hooks` — delete and trash hooks are wired
- [ ] `test_track_product_archive_stores_pending_archive` — archive stored with id + timestamp
- [ ] `test_track_product_archive_removes_from_pending_inventory` — stock update removed on archive
- [ ] `test_track_product_archive_batches_multiple_products` — multiple archives batched
- [ ] `test_track_product_archive_stops_accumulating_at_threshold` — MAX_PENDING_UPDATES respected
- [ ] `test_generate_archive_feed_*` — feed has correct `id`, `availability` columns, and `out_of_stock` value
- [ ] `test_sync_archives_*` — skips when disabled/empty, clears on success, retains on failure, defers to full sync at threshold

### Manual

- [ ] Enable Agentic Commerce in plugin settings and confirm a valid Stripe secret key is configured
- [ ] Open Action Scheduler (**WooCommerce → Status → Scheduled Actions**) — note there is no pending `wc_stripe_agentic_commerce_sync_archives` action at this point
- [ ] **Trash a product:** go to **Products**, move any product to trash
  - Within ~5 seconds an `wc_stripe_agentic_commerce_sync_archives` action should appear in Action Scheduler scheduled ~60 seconds out
  - After it runs, verify in **WooCommerce → Status → Logs** that an "Archive feed uploaded" entry appears with a valid `import_set_id`
  - Verify in your Stripe dashboard (Data Management → ImportSets) that the ImportSet was created with `product_catalog_feed` format and the product row has `availability: out_of_stock`
- [ ] **Permanently delete a product:** empty the trash for a product
  - Confirm the same Action Scheduler + log flow as above
  - Verify the product is marked `out_of_stock` (not deleted) in the Stripe catalog
- [ ] **Delete a product that has a pending stock update:** update a product's stock quantity (triggers the inventory sync), then immediately delete it
  - Confirm the product is removed from `wc_stripe_agentic_pending_inventory` and added to `wc_stripe_agentic_pending_archives` (check via **WP CLI**: `wp option get wc_stripe_agentic_pending_inventory` and `wp option get wc_stripe_agentic_pending_archives`)
  - Only one archive feed (not an inventory feed) should be uploaded for that product
- [ ] **Feature disabled:** disable Agentic Commerce, trash a product, and confirm no archive feed is uploaded and the pending option is retained for retry once re-enabled


includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
```diff
@@ -218,6 +273,128 @@ public function sync_inventory(): void {
...
+				'Agentic Commerce: Archive feed uploaded',
+				[
+					'archives'      => count( $pending ),
+					'import_set_id' => $result['import_set_id'] ?? null,
+					'status'        => $result['status'] ?? 'unknown',
+				]
+			);
+
+			// Clear pending archives on success.
+			delete_option( self::PENDING_ARCHIVES_OPTION );
```
Mayisha: We should delete the diff here also, similar to how you did it in the inventory sync PR.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
```diff
@@ -218,6 +273,128 @@ public function sync_inventory(): void {
...
+
+			// Clean up the temporary file.
```
Mayisha, line 491: We should clean up on catch also. Or moving it to the `finally` block should work.

Mayisha (commented):
Code looks good to me, but I do not see any scheduled action after deleting a product.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
wjrosa: Done — `sync_archives()` now uses `array_diff_key` to diff against the option value re-read after upload, preserving any archive events that arrived while the upload was in progress. Same pattern as `sync_inventory()`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
wjrosa: Done — the catch block now cleans up the temp file and reschedules the action for retry, matching the pattern in `sync_inventory()`.

wjrosa:
@Mayisha Good catch — you were right that there was no scheduled action, and the root cause was the choice of hooks.

`woocommerce_before_delete_product` and `woocommerce_trash_product` are only fired inside WooCommerce's data store `delete()` method, which is only called when deletion goes through the WooCommerce REST API or programmatic `$product->delete()`. The standard WordPress admin UI 'Move to Trash' / 'Delete Permanently' flow calls `wp_trash_post()` / `wp_delete_post()` directly, bypassing the data store entirely — so `track_product_archive()` was never being called and no action was ever scheduled for admin UI deletions.

Fixed in ad949583 by switching to the WordPress-level `before_delete_post` and `wp_trash_post` hooks, which fire for all deletion paths (admin UI, REST API, WP-CLI, programmatic). A new `maybe_track_product_archive()` dispatcher filters to `post_type === 'product'` only so orders, pages, and other post types are ignored.

(PS: debugging, fixing and reply created with the help of Claude)

Mayisha (approved):
Thanks for the fix @wjrosa. I can see the scheduled action now after a product deletion and also the relevant logs when the action runs 👍

diegocurbelo (commented):
The log entries and the scheduled action work as described, but the file is not processed correctly by Stripe, and the product is still listed with stock in the product feed:

Image
Image

Image

**File:** [stripe-archive-feed-2026-04-08-dca4658e8ab5cd0073fdb4a3669e1174.csv](https://github.com/user-attachments/files/26578815/stripe-archive-feed-2026-04-08-dca4658e8ab5cd0073fdb4a3669e1174.csv)
**Error log:** [import_set_errors_stripe-archive-feed-2026-04-08-dca4658e8ab5cd0073fdb4a3669e1174.csv](https://github.com/user-attachments/files/26578827/import_set_errors_stripe-archive-feed-2026-04-08-dca4658e8ab5cd0073fdb4a3669e1174.csv)

wjrosa:
Thanks for catching this, Diego! The issue was that the archive feed was only sending `id` and `availability` columns, but since it's delivered as a `product_catalog_feed` ImportSet, Stripe requires all the mandatory fields.

I pushed a fix in b5402fdb that captures the full product data (via the product mapper) at deletion/trash time — before the product is removed — and includes all schema columns in the archive CSV. The `availability` field is forced to `out_of_stock` regardless of the product's current stock status.

Could you give it another try?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
```diff
@@ -63,6 +77,13 @@ public function register_hooks(): void {
 		add_action( 'woocommerce_product_set_stock', [ $this, 'track_stock_change' ] );
 		add_action( 'woocommerce_variation_set_stock', [ $this, 'track_stock_change' ] );
 		add_action( self::SCHEDULED_ACTION, [ $this, 'sync_inventory' ] );
+		// Use WordPress-level hooks rather than WooCommerce data-store hooks
+		// (woocommerce_before_delete_product / woocommerce_trash_product) because
+		// the data-store hooks only fire through the REST API or programmatic
+		// $product->delete() calls, not through the WordPress admin UI.
+		add_action( 'before_delete_post', [ $this, 'maybe_track_product_archive' ] );
+		add_action( 'wp_trash_post', [ $this, 'maybe_track_product_archive' ] );
+		add_action( self::ARCHIVE_SCHEDULED_ACTION, [ $this, 'sync_archives' ] );
```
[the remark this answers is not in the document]
wjrosa: Good catch — added an `untrash_post` hook (`maybe_cancel_pending_archive`) that removes the product from `PENDING_ARCHIVES_OPTION` when restored. Also added test coverage for this flow. Fixed in c1d4a073.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
```diff
@@ -101,6 +122,90 @@ public function track_stock_change( \WC_Product $product ): void {
...
+			return;
+		}
+
+		$product = wc_get_product( $post_id );
+		if ( ! $product ) {
+			return;
+		}
+
+		$this->track_product_archive( $product );
+	}
```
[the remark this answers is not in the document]
wjrosa: Good point — updated `maybe_track_product_archive` (and `maybe_cancel_pending_archive`) to accept both `product` and `product_variation` post types. Added a test for variation archive tracking. Fixed in c1d4a073.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
```diff
@@ -238,4 +343,149 @@ public function sync_inventory(): void {
...
+				]
+			);
+
+			// Clean up the temporary file even on failure.
+			// $feed is non-null here: if generate_archive_feed() returned null we
+			// would have returned early before any exception could be thrown.
+			$file_path = $feed->get_file_path();
+			if ( ! empty( $file_path ) && file_exists( $file_path ) ) {
+				wp_delete_file( $file_path );
+			}
```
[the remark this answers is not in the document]
wjrosa: Investigated this — PHPStan confirms `$feed` is always non-null in the catch block because `generate_archive_feed()` only returns `null` (triggering an early return) or a feed instance. Any exception from the delivery layer happens after `$feed` is assigned. Adding a null/instanceof guard causes a PHPStan `alwaysTrue` error. Keeping the existing pattern with the explanatory comment.

diegocurbelo (approved):
Thanks for the fix @wjrosa.

Now the file is correctly processed by Stripe:
Image
[stripe-archive-feed-2026-04-13-318dbf0ff04b764c151d8a7d03215a22.csv](https://github.com/user-attachments/files/26688709/stripe-archive-feed-2026-04-13-318dbf0ff04b764c151d8a7d03215a22.csv)

And is correctly shown as `out-of-stock`:

Image

PR #4556, opened 2025-08-04
Remove validation error check from classic checkout before pm creation

Mayisha:
Fixes STRIPE-650
Fixes #4549 

## Changes proposed in this Pull Request:
In https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4502 we added a check to see whether there are any validation error on the checkout page before creating a payment method in Stripe. 
For block checkout, we used `validationStore` and the check is straightforward as the validation check is done first before the `/checkout` request.
But for shortcode checkout, we had to do [this workaround](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4502/files#r2212851652) and explicitly trigger the validation check. However, this might not work well when there are custom checkout fields (like the reported issue) where the check prevents a payment method creation but tries to create a payment intent as the validation passed in the core.

In this PR, I have removed the validation check from the shortcode checkout to unblock the users. Creating a couple of extra payment methods is harmless compared to blocking users from paying.

## Testing instructions
- Enable card in Stripe settings page.
- Add a product to your cart and go to the shortcode checkout page.
- Fill in all the required fields on the checkout page, fill in the card details and click the place order button.
- The payment should be successful without any errors.
- Add another product to your cart and go to the shortcode checkout page.
- Leave some required fields empty, fill in the card details and click the place order button.
- There should be a validation error notice on the page.
- Now fill in all the required fields and click the place order button again.
- Payment should be successful without any errors.

PR #5222, opened 2026-04-03
Fix STRIPE-652: Prevent unnecessary payment method creation on shortcode checkout validation errors

kamlesh08:
Fixes STRIPE-652

When a customer clicks "Place Order" on the shortcode (classic) checkout without filling in required fields (email, billing details, etc.), the plugin calls Stripe's `createPaymentMethod` API before WooCommerce's server-side validation rejects the form. This creates unnecessary payment methods on Stripe's side.

### Background

This was previously fixed in PR #4502 (STRIPE-589) for both block and shortcode checkout. However, the shortcode fix was reverted in PR #4556 (STRIPE-650) because it triggered jQuery `validate` and `blur` events on all form fields, which caused side effects on sites with custom themes — specifically, the payment method was not attached to the PaymentIntent, breaking checkout entirely.

### This fix

Instead of triggering validation events (which mutate the DOM and can interfere with theme/plugin handlers), this PR adds a **read-only DOM check** that inspects visible required field values directly:

- Checks `.validate-required:visible` wrappers for empty `input.input-text`, `select`, or `input[type="checkbox"]` fields
- Uses the same field selectors as WC core's `checkout.js` validation
- **No event triggers, no DOM mutations, no CSS class changes** — purely reads field values
- Visibility filtering uses jQuery `:visible` in the caller, while the utility function uses plain DOM (no jQuery dependency, fully testable in jsdom)

When empty required fields are found, the handler returns without calling `processPayment()`, letting WC's standard AJAX submission proceed to server-side validation which displays the appropriate error messages.

### How can this code break?

- If a theme doesn't use `.validate-required` on required field wrappers (non-standard, but possible). In that case, the check passes through and the existing behavior is preserved.
- If a custom required field uses a non-standard input type not matched by the selector. Same fallback — existing behavior preserved.
- The check only catches **empty** fields, not invalid values (e.g. malformed email). Invalid-but-present values still trigger a Stripe PM creation, then fail server-side. This matches the block checkout behavior.

- The utility function (`hasEmptyRequiredFields`) is extracted, exported, and covered by 9 unit tests
- All 623 existing JS tests continue to pass
- The fix only applies to `form.checkout` (shortcode checkout) — Pay for Order and Add Payment Method flows are unaffected

### Test 1: Empty fields should NOT create a payment method

1. Set up a Stripe test site with the shortcode checkout (`[woocommerce_checkout]`)
2. Add a product to cart and go to checkout
3. Open browser DevTools → Network tab, filter by `stripe.com`
4. Without filling in any billing details, enter card details in the Stripe card element and click "Place Order"
5. **Expected:** No request to `/v1/payment_methods` on Stripe. WooCommerce shows validation errors for the empty required fields.

### Test 2: Filled fields should process payment normally

1. Same setup as Test 1
2. Fill in all required billing fields (name, email, address, etc.)
3. Enter test card `4242424242424242` with any future expiry and any CVC
4. Click "Place Order"
5. **Expected:** Payment processes normally. A `/v1/payment_methods` request is made to Stripe, followed by successful order completion.

### Test 3: Partial fields

1. Fill in some required fields but leave email empty
2. Enter card details and click "Place Order"
3. **Expected:** No Stripe PM creation. WooCommerce shows "Billing Email address is a required field" error.

### Test 4: Hidden conditional fields

1. Uncheck "Ship to a different address" (if visible)
2. Leave shipping fields empty (they should be hidden)
3. Fill in all visible billing fields and card details
4. Click "Place Order"
5. **Expected:** Payment processes normally. Hidden shipping fields are not checked.

---

- [x] Covered with tests (9 new unit tests for `hasEmptyRequiredFields`)
- [ ] Tested on mobile (applies — should be verified on mobile checkout)

### Changelog entry

> Fix - Prevent unnecessary Stripe payment method creation when shortcode checkout has empty required fields.

**Post merge**

- [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions)

---

This PR was created during the [Woo Bug Blitz](https://woohappyp2.wordpress.com/2026/03/18/extensions-bug-blitz/) via team Ceres' [`/bug-blitz` Claude skill](https://github.com/Automattic/public-resources-squad/blob/trunk/.claude/skills/bug-blitz/skill.md). The fix was authored with Claude Code assistance by a Happiness Engineer who encounters this issue in support tickets.


Luc45 (commented):
Tested on QIT (WC 10.6.2, PHP 8.2) with xdebug enabled but no Stripe test keys configured. I could verify the client-side validation logic and server-side paths, but not the full payment flow end-to-end.

The `hasEmptyRequiredFields` early-return logic is sound. With empty required billing fields, the WC AJAX checkout now proceeds to server-side validation instead of firing `processPayment()` prematurely. I confirmed via xdebug breakpoints that `send_ajax_failure_response` fires with the expected validation errors, and `process_order_payment` is never hit. On the second attempt with all fields filled, `processPayment()` fires correctly with clean state.

Edge cases I could verify without Stripe keys: guest checkout (same DOM structure), Select2 enhanced selects (underlying value preserved), country change toggling state field requirement, T&C checkbox in a `.validate-required` wrapper, custom required fields via `woocommerce_checkout_fields` filter, double-submit retry after validation, and the Pay for Order page (completely unaffected, no `.validate-required` wrappers). All 624 existing JS tests pass.

What I could not verify: whether the happy path still completes a real Stripe payment after the check passes, and whether the interaction with Stripe.js initialization has any side effects when `processPayment()` is deferred.

**This needs end-to-end testing with Stripe test keys. A maintainer with proper Stripe test keys should review this to verify the full payment flow end-to-end.**

Two informational notes: (1) Required `<textarea>` fields aren't matched by the selector, but this matches WC core's own `checkout.js` validation and is the same behavior as `develop`. (2) The T&C checkbox behavior change is a user-visible improvement worth mentioning in the PR description.

Luc45:
The ACSS E2E test (`acss.spec.js:49` - `customer can save and reuse ACSS payment method`) is failing on this PR while passing on other branches. The `hasEmptyRequiredFields` check might be interfering with the ACSS payment flow. This reinforces my earlier note about needing end-to-end verification from a maintainer with Stripe test keys.

diegocurbelo (approved):
The code change look good to me, and it tests well! 👍🏼

PR #5224, opened 2026-04-06
Allow customers to save payment methods during checkout with adaptive pricing

Mayisha:
Fixes STRIPE-927

- Pass `savePaymentMethod` into Checkout `confirm()` based on the save payment method checkbox in classic checkout.
- Pass `savePaymentMethod` in `save_payment_method` metadata and `confirm() based on the save payment method checkbox in block checkout.
- Enable saving a payment method in the checkout session during create request via `payment_method_save: 'enable'` in `create_checkout_session` method. (without this the `savePaymentMethod` param in JS won't work)
- Keep redisplay of saved methods and the Stripe rendered save payment method checkbox off via Elements (`savedPaymentMethod.enableRedisplay: 'never'` and `savedPaymentMethod.enableSave: 'never'`) on classic and Blocks.
- Keep a flag in order meta in `process_payment_with_checkout_session` if save payment method checkbox was selected by the shopper.. At this point we do not have the payment method ID as during checkout session flow Stripe implicitly creates a payment method during `confirm()`.
- In webhook handler, we check for the flag and create a payment token in WooCommerce for this payment method (the payment method is already saved and attached to the customer in Stripe as we use `savePaymentMethod` in JS)

## Testing instructions
- Enable Optimized checkout and Adaptive Pricing from the Stripe settings page.
- As a logged in shopper, add a product to your cart and go to the classic checkout page.
- Use card method with your local currency (adaptive pricing) and check the save payment method checkbox.
- Click the checkout button and wait for the order to complete.
- Go to **My Account > Payment methods** page and verify that the card is saved successfully.
- Test and confirm that the payment method is not saved when the save payment method checkbox is unchecked during checkout.
- Test the above in block checkout and verify that it works on block checkout too.

**Note:** Test with card as the non-card payment methods are not fully functional now with adaptive pricing.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1266,7 +1266,7 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
 		}
 
 		if ( is_string( $checkout_session_id ) && ! empty( $checkout_session_id ) ) {
-			return $this->process_payment_with_checkout_session( $order_id, $checkout_session_id );
+			return $this->process_payment_with_checkout_session( $order_id, $checkout_session_id, $save_payment_method, $selected_payment_type );
```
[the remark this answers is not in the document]
Mayisha: Tested and confirmed that saving payment method is working fine in block checkout.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3187,7 +3196,7 @@ private function get_express_payment_type_from_request() {
 	 * @param stdClass $payment_method_object  The payment method object retrieved from Stripe.
 	 * @param string   $payment_method_type    The payment method type, like `card`, `sepa_debit`, etc.
 	 */
-	protected function handle_saving_payment_method( WC_Order $order, $payment_method_object, string $payment_method_type ) {
+	public function handle_saving_payment_method( WC_Order $order, $payment_method_object, string $payment_method_type ) {
```
wjrosa: I would use the opportunity to add missing types here since this is becoming public:

```suggestion
	public function handle_saving_payment_method( WC_Order $order, stdClass $payment_method_object, string $payment_method_type ): void {
```

includes/class-wc-stripe-order-helper.php
```diff
@@ -516,6 +523,34 @@ public function update_stripe_checkout_session_id( ?WC_Order $order = null, stri
 		return $this->update_order_meta( $order, self::META_STRIPE_CHECKOUT_SESSION_ID, $checkout_session_id );
 	}
 
+	/**
+	 * Gets whether the payment method should be saved to the store after a checkout session payment.
+	 *
+	 * @param WC_Order|null $order
+	 * @return bool
+	 */
+	public function get_stripe_save_payment_method( ?WC_Order $order = null ): bool {
```
wjrosa: I would consider using a more Boolean-esque naming, like `should_save_stripe_payment_method`. Below it would be `update_should_save_stripe_payment_method` and `delete_should_save_payment_method`. Or  something similar

wjrosa (approved):
Looks good to me! Left some suggestions

malithsen (approved):
Looks good to me

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Good idea.

PR #5234, opened 2026-04-07
Add exit survey to capture merchant feedback on deactivation and gateway disable

pierorocca:
Fixes STRIPE-1011

Implements an in-plugin exit survey using Crowdsignal to capture structured data on why merchants churn from the Stripe extension or fail to activate after install. The survey appears as a modal with an embedded iframe when a merchant takes any of these actions:

### Trigger surfaces

| Trigger | Page | Value | Mechanism |
|---------|------|-------|-----------|
| Plugin deactivation | Plugins page | `plugins_page_deactivate` | Intercepts deactivate link click |
| Gateway disable | WC Payments page | `payments_page_disable` | `apiFetch` middleware detects `URLSearchParams` toggle request |
| Gateway toggle off | Legacy Payments list (WC L-2) | `gateway_toggle_disable` | Intercepts `woocommerce_toggle_gateway_enabled` AJAX |
| "Enable Stripe" unchecked + Save | Stripe settings | `settings_disable` | Detects enabled→disabled transition after save |

### Key implementation details

- **Crowdsignal iframe modal** — shared `ExitSurveyModal` component embeds the survey at `woocommerce.survey.fm/stripe-exit-survey` with iframe `sandbox` for defense-in-depth
- **Hidden fields** — Stripe account ID, store ID, plugin/WC/WP versions, and trigger type passed as Crowdsignal `q_N_text` URL parameters
- **7-day cooldown** — persisted via REST endpoint (`POST /wc/v3/wc_stripe/exit-survey/dismiss`) to `wp_options` (non-autoloaded), with in-memory update to prevent re-show on retry
- **Graceful degradation** — if React app fails to mount or params are missing, deactivation/disable proceeds normally
- **Cache-only account lookup** — survey params use `WC_Stripe_Database_Cache::get()` directly to avoid triggering live Stripe API calls
- **Shared helper** — `WC_Stripe_Helper::get_exit_survey_params()` deduplicates the localized params array across controllers

### New files
- `includes/admin/class-wc-rest-stripe-exit-survey-controller.php` — REST endpoint for cooldown
- `includes/admin/class-wc-stripe-plugins-page-controller.php` — enqueues JS on Plugins screen
- `client/components/exit-survey-modal/` — shared modal component + styles
- `client/entrypoints/plugins-page/` — deactivate link interception
- `client/entrypoints/payment-gateways/payments-page-survey.js` — WC Payments page `apiFetch` middleware

### Modified files
- `client/entrypoints/payment-gateways/payment-gateways-confirmation.js` — shows survey after disable confirm (legacy)
- `client/entrypoints/payment-gateways/index.js` — mounts PaymentsPageSurvey component
- `client/settings/settings-manager/index.js` — detects Enable Stripe unchecked after save
- `includes/admin/class-wc-stripe-payment-gateways-controller.php` — localizes survey params, `$_GET` fallback for new Payments page
- `includes/admin/class-wc-stripe-settings-controller.php` — merges survey params into settings localization
- `includes/class-wc-stripe-helper.php` — shared `get_exit_survey_params()` helper
- `includes/class-wc-stripe.php` — registers new controllers and REST routes
- `webpack.config.js` — adds `plugins-page` entry point
- `uninstall.php` — cleans up cooldown option

### Plugin deactivation survey (Plugins page)
1. Go to **Plugins** in the WordPress admin
2. Click **Deactivate** next to "Stripe for WooCommerce"
3. A modal should appear with the Crowdsignal exit survey
4. Complete or dismiss the survey — the plugin should then deactivate normally
5. Reactivate the plugin, then try deactivating again within 7 days — the survey should NOT appear (cooldown)

### Gateway disable survey (WC Payments page)
1. Ensure Stripe is enabled
2. Go to **WooCommerce > Payments** (the core Payments settings page)
3. Click **Disable** on the Stripe gateway
4. After the disable completes, a survey modal should appear
5. Dismiss the survey

### Settings disable survey (Stripe settings)
1. Go to **WooCommerce > Settings > Payments > Stripe** (the Stripe settings section)
2. Uncheck **Enable Stripe**
3. Click **Save changes**
4. A survey modal should appear after save

### Cooldown verification
1. After dismissing the survey, check `wp_options` for `wc_stripe_exit_survey_last_shown` — should contain a recent ISO 8601 timestamp
2. Triggering the survey again within 7 days should skip it automatically on all surfaces

---

- [x] Covered with tests (PHPUnit for REST controller, Jest for modal component, plugins page, and payment gateways confirmation)

### Changelog entry

Included in this PR.

---

### From payments providers page:

https://github.com/user-attachments/assets/c8b147af-ae7d-4f61-bf2b-819449418f68

### Cooldown, doesn't trigger

https://github.com/user-attachments/assets/e242fa51-4ff7-4e2e-9157-b7556bc5e8c9

### Stripe Settings

https://github.com/user-attachments/assets/5a6c3559-643e-48bc-8c17-c3bfafce633d

### Plugins page

https://github.com/user-attachments/assets/2f4743ff-87df-49cf-8f6f-abb29c0f88fd

### Data Collection, note the source trigger is captured correctly.
image

image

image


pierorocca:
## 🤖 Automated Review Report

**Verdict: REQUEST CHANGES** — 1 critical bug, 2 test gaps

### Required Changes

| # | Severity | Issue |
|---|----------|-------|
| 1 | **CRITICAL** | PHP fatal: `WC_REST_Stripe_Exit_Survey_Controller::OPTION_NAME` referenced before class file is loaded |
| 2 | HIGH | Missing Jest tests for `payment-gateways-confirmation.js` exit survey flow |
| 3 | HIGH | Missing Jest tests for `plugins-page/index.js` |

### Architecture: FLAG
- Load-order bug: exit survey controller class referenced in `init()` but file only loaded in `register_routes()` (runs later on `rest_api_init`)
- All layer boundaries, registration patterns, and conventions are correct

### Security: PASS
- Proper `manage_woocommerce` permission, CSRF via wp.apiFetch nonce, iframe origin validation, no sensitive data exposed

### Performance: PASS
- Scripts scoped to specific admin pages, caching used correctly, event listeners cleaned up, zero idle rendering cost

### Test Coverage: FLAG
- REST controller: 4 PHPUnit tests ✅
- ExitSurveyModal: 11 Jest tests ✅
- Plugins page app: no tests ❌
- Payment gateways exit survey flow: no tests ❌

### Non-blocking Suggestions
- Add `sandbox` attribute to iframe
- Set `autoload=false` on cooldown option
- Deduplicate localized params array

---

pierorocca:
## Review changes applied

### Required changes (all resolved)

**1. [CRITICAL] PHP class load-order fix** (`d40facc5`)
- `WC_REST_Stripe_Exit_Survey_Controller::OPTION_NAME` was referenced in `WC_Stripe_Plugins_Page_Controller` and `WC_Stripe_Payment_Gateways_Controller` before the class file was loaded (it's only `require_once`'d in `register_routes()` on `rest_api_init`, which runs after `init()`). This would cause a PHP fatal on any admin page load.
- **Fix:** Replaced the class constant references with inline string `'wc_stripe_exit_survey_last_shown'` in both controllers.

**2. [HIGH] Added Jest tests for payment-gateways exit survey flow** (`d40facc5`)
- `PaymentGatewaysConfirmation` had no test coverage for the new exit survey integration path.
- **Added 4 tests:** cooldown inactive → show survey, cooldown active → skip, undefined params → skip, survey close → proceed with disable.

**3. [HIGH] Added Jest tests for plugins-page** (`d40facc5`)
- `PluginsPageApp` had no test coverage.
- **Added 4 tests:** deactivate link interception, cooldown active → skip, survey dismiss → proceed, missing deactivate link → render nothing.

### Non-blocking suggestions (all applied)

**4. iframe sandbox attribute** (`4d7556ee`)
- Added `sandbox="allow-scripts allow-forms allow-same-origin"` to the Crowdsignal iframe. Limits what embedded content can do (no top-frame navigation, no popups) as defense-in-depth.

**5. autoload=false on cooldown option** (`4d7556ee`)
- Set third arg to `false` on `update_option()` for `wc_stripe_exit_survey_last_shown`. Prevents the option from being autoloaded on every page for sites without a persistent object cache — it's only read on 2 specific admin pages.

**6. Shared survey params helper** (`4d7556ee`)
- Extracted `WC_Stripe_Helper::get_exit_survey_params()` to deduplicate the localized params array that was copy-pasted between `WC_Stripe_Plugins_Page_Controller` and `WC_Stripe_Payment_Gateways_Controller`. Prepares for future surveys.

---

client/entrypoints/plugins-page/index.js
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		deactivateLink.current = link;
+
+		const handleClick = ( event ) => {
+			// If cooldown is active, skip survey and let deactivation proceed normally.
+			if (
+				isCooldownActive(
+					wcStripePluginsPageParams.exitSurveyLastShown
+				)
```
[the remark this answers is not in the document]
pierorocca: Fixed in f1b5e48c. Added `typeof wcStripePluginsPageParams === 'undefined'` guard in the click handler, with a test covering this path.

client/entrypoints/payment-gateways/__tests__/payment-gateways-confirmation.test.js
```diff
@@ -0,0 +1,151 @@
...
+		trigger: () => {
+			// no-op for toggle clicks in tests
+		},
+		length: selectorStr.includes( 'woocommerce-input-toggle--disabled' )
+			? 0
+			: 1,
+		removeClass: jest.fn().mockReturnValue( {
+			length: 0,
+		} ),
+	};
```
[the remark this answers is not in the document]
pierorocca: Acknowledged — the jQuery trigger mock is intentionally a no-op since `proceedWithDisable()` calls jQuery to re-trigger the toggle click, which we can't meaningfully assert without a full jQuery + DOM simulation. The test verifies the survey modal appears and closes correctly; the actual toggle behavior is covered by the existing `DisableConfirmationModal` tests and manual testing.

pierorocca:
Duplicate title in modal has been removed. Only the Crowdsignal title remains.

image

pierorocca:
## Latest changes (63b532b)

- **Removed duplicate modal title** — the WordPress Modal no longer renders its own "Stripe for WooCommerce" header. The Crowdsignal survey has its own branded header inside the iframe.
- **Close button floats over iframe** — modal header is positioned absolute so the X aligns with the Crowdsignal header row.
- **Sandbox attribute reapplied** with `allow-popups allow-popups-to-escape-sandbox` — the privacy policy link now opens in a new tab (required `target="_blank"` on the Crowdsignal survey side).

---

includes/class-wc-stripe.php
```diff
@@ -250,7 +250,10 @@ public function init() {
 			}
 
 			require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-stripe-payment-gateways-controller.php';
-			new WC_Stripe_Payment_Gateways_Controller();
+			new WC_Stripe_Payment_Gateways_Controller( $this->account );
+
+			require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-stripe-plugins-page-controller.php';
```
wjrosa: Since this is a new class, it should be autoloaded (by composer autoload, map generated with composer dump-autoload)

includes/class-wc-stripe.php
```diff
@@ -855,6 +858,10 @@ public function register_routes() {
 
 		$oc_setting_toggle_controller = new WC_Stripe_REST_OC_Setting_Toggle_Controller( $this->get_main_stripe_gateway() );
 		$oc_setting_toggle_controller->register_routes();
+
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-rest-stripe-exit-survey-controller.php';
```
wjrosa: Same as my previous comment.

includes/class-wc-stripe-helper.php
```diff
@@ -2307,4 +2307,25 @@ public static function get_currency_decimals( string $currency_code ): int {
...
+	 * @param WC_Stripe_Account $account Stripe account instance.
+	 * @return array Associative array of survey parameters for wp_localize_script.
+	 */
+	public static function get_exit_survey_params( WC_Stripe_Account $account ) {
+		// Read account data from cache only — avoid triggering a live Stripe API call.
+		$account_cache = WC_Stripe_Database_Cache::get( WC_Stripe_Account::ACCOUNT_CACHE_KEY );
+		$account_data  = is_array( $account_cache ) ? $account_cache : [];
+
+		return [
+			'exitSurveyLastShown' => get_option( 'wc_stripe_exit_survey_last_shown', null ),
```
wjrosa: Usually, we prefer snake_case as array keys (`stripe_account_id`, `wc_store_id`, etc).

includes/class-wc-stripe-helper.php
```diff
@@ -2307,4 +2307,25 @@ public static function get_currency_decimals( string $currency_code ): int {
...
 		return 2;
 	}
+
+	/**
+	 * Build the localized survey params array shared across admin controllers.
+	 *
+	 * @param WC_Stripe_Account $account Stripe account instance.
+	 * @return array Associative array of survey parameters for wp_localize_script.
+	 */
+	public static function get_exit_survey_params( WC_Stripe_Account $account ) {
```
wjrosa:
```suggestion
	public static function get_exit_survey_params( WC_Stripe_Account $account ): array {
```

includes/admin/class-wc-stripe-plugins-page-controller.php
```diff
@@ -0,0 +1,93 @@
...
+		add_action( 'admin_footer', [ $this, 'render_container' ] );
+	}
+
+	/**
+	 * Enqueue the plugins page script and styles.
+	 *
+	 * @param string $hook_suffix The current admin page hook suffix.
+	 * @return void
+	 */
+	public function enqueue_scripts( $hook_suffix ) {
```
wjrosa:
```suggestion
	public function enqueue_scripts( string $hook_suffix ) {
```

includes/admin/class-wc-stripe-payment-gateways-controller.php
```diff
@@ -70,13 +81,22 @@ public function enqueue_payments_scripts() {
...
 		$this->register_payments_scripts();
 
+		// Check via WC globals (legacy) or $_GET (new React Payments page).
 		$is_payment_methods_page = (
 			is_admin() &&
-			$current_tab && ! $current_section
-			&& 'checkout' === $current_tab
+			(
+				( $current_tab && ! $current_section && 'checkout' === $current_tab ) ||
+				( isset( $_GET['page'] ) && 'wc-settings' === $_GET['page'] && isset( $_GET['tab'] ) && 'checkout' === $_GET['tab'] && ! isset( $_GET['section'] ) ) // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
wjrosa: I would consider adding a helper method to check this page since we have the same logic already in the current code

includes/admin/class-wc-stripe-payment-gateways-controller.php
```diff
@@ -35,7 +46,7 @@ public function __construct() {
 	public function register_payments_scripts() {
 		$payment_gateways_script_asset_path = WC_STRIPE_PLUGIN_PATH . '/build/payment-gateways.asset.php';
 		$payment_gateways_script_asset      = file_exists( $payment_gateways_script_asset_path )
-			? require_once $payment_gateways_script_asset_path
+			? require $payment_gateways_script_asset_path
```
wjrosa: Is this change intentional? `require_once` is safer to avoid errors when, for some reason, the script tries to include the same file multiple times.

includes/admin/class-wc-stripe-payment-gateways-controller.php
```diff
@@ -10,12 +10,23 @@
  * @since 5.6.0
  */
 class WC_Stripe_Payment_Gateways_Controller {
+
+	/**
+	 * The Stripe account instance.
+	 *
+	 * @var WC_Stripe_Account
+	 */
+	private $account;
```
wjrosa:
```suggestion
	private WC_Stripe_Account $account;
```

includes/admin/class-wc-rest-stripe-exit-survey-controller.php
```diff
@@ -0,0 +1,57 @@
...
+ * Handles the exit survey dismiss REST endpoint.
+ */
+class WC_REST_Stripe_Exit_Survey_Controller extends WC_Stripe_REST_Base_Controller {
+
+	/**
+	 * Endpoint path.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'wc_stripe/exit-survey';
```
wjrosa, line 20:
```suggestion
	protected string $rest_base = 'wc_stripe/exit-survey';
```

client/components/exit-survey-modal/index.js
```diff
@@ -0,0 +1,176 @@
...
+ *
+ * Displays a Crowdsignal survey in an iframe inside a WordPress Modal.
+ * Persists a 7-day cooldown via REST on close.
+ *
+ * @param {Object}   props                Props.
+ * @param {string}   props.trigger        'deactivate' or 'disable'.
+ * @param {Function} props.onRequestClose Called after the modal is closed and cooldown is persisted.
+ * @param {Object}   props.surveyParams   Hidden field values from wp_localize_script.
+ */
+const ExitSurveyModal = ( { trigger, onRequestClose, surveyParams } ) => {
```
wjrosa, line 94: Personally, I think we are creating code tied to the survey iframe. I recommend opening the form URL in a new tab instead.

wjrosa (approved):
Code is good and works as expected. Left some suggestions

client/components/exit-survey-modal/index.js
pierorocca: Intentional — the iframe modal keeps the merchant in context rather than redirecting to a separate tab. This matches the WooPayments exit survey pattern and is proven to drive higher participation rates. A redirect would be a bigger interruption to the merchant's flow since we're already intercepting their deactivate/disable action.

includes/admin/class-wc-stripe-payment-gateways-controller.php
pierorocca: Yes, intentional. `require_once` returns `true` (not the array) on subsequent includes when the file was already loaded. If the asset file gets included earlier in the request by another controller, the second `require_once` returns `true` instead of the dependency/version array, causing `wp_register_script` to silently fail. The CSS enqueued but the JS didn't — this was the root cause of the Payments page survey not loading. `require` always returns the file's return value.

diegocurbelo (approved):
The changes look good to me, and it tests as descirbed.

One question though, should we also show the survey when the merchant disconnects the account (which effectively disables the gateway)?:

Image

includes/class-wc-stripe.php
pierorocca: Fixed. Both new classes added to Composer classmap in composer.json and require_once calls removed.

includes/class-wc-stripe-helper.php
pierorocca: Fixed. All keys renamed to snake_case: exit_survey_last_shown, stripe_account_id, wc_store_id, plugin_version, wc_version, wp_version. Updated all JS consumers and tests.

includes/class-wc-stripe-helper.php
pierorocca: Fixed. Added `: array` return type.

includes/admin/class-wc-stripe-plugins-page-controller.php
pierorocca: Fixed. Added `string` type hint.

includes/admin/class-wc-stripe-payment-gateways-controller.php
pierorocca: Fixed. Extracted `WC_Stripe_Helper::is_admin_payments_page()` and use it in the controller.

includes/admin/class-wc-rest-stripe-exit-survey-controller.php
pierorocca: Cannot add native type — parent `WP_REST_Controller::$rest_base` doesn't declare a native type, and PHPStan rejects the override. Kept the `@var string` docblock instead.

includes/admin/class-wc-stripe-payment-gateways-controller.php
pierorocca: Fixed. Updated to PHP 7.4+ typed property.

includes/class-wc-stripe.php
pierorocca: Fixed. Same approach — added to Composer classmap, require_once removed.

pierorocca:
> One question though, should we also show the survey when the merchant disconnects the account (which effectively disables the gateway)?:

Good catch @diegocurbelo. Do you have a sense of when a merchant would disconnect before disabling or deactivating? For sure from Test to Live which we don't want that to trigger a survey.

Let's leave this out for now as I don't know how to separate out the Test to Live and Live to Test switching and I'd think they'd eventually actually disable or deactivate the plugin. This should be good enough to get a decent sample.

includes/admin/class-wc-rest-stripe-exit-survey-controller.php
pierorocca: @wjrosa please check the accuracy of this statement above.

includes/admin/class-wc-stripe-payment-gateways-controller.php
pierorocca: Correction — I initially suspected `require_once` returning `true` on a double-include was causing the JS not to load. After further investigation, nothing else includes this asset file, so that wasn't the root cause. The actual issue was the `$is_payment_methods_page` condition failing because the new WC Payments React page doesn't set the WC globals (`$current_tab`, `$current_section`). The fix is the `$_GET` fallback now in `WC_Stripe_Helper::is_admin_payments_page()`. Reverted back to `require_once` to match codebase convention.

wjrosa (approved):
Still working correctly 👍

PR #5237, opened 2026-04-08
Fix: scope set_cookie_on_current_request to checkout context only

### LINKED ISSUE #4875: Cart Emptied After Account Registration When Stripe + WooPayments + Jetpack Are Active
We discovered this issue while debugging unexpected cart resets for **Sensei LMS course on a Team51 partner site**, but we later found it is **not specific to Sensei** and can be reproduced on a **clean WordPress installation**.

When **WooCommerce Stripe**, **WooPayments**, and **Jetpack (connected with Sync enabled)** are all active, a customer who registers an account from the **My Account** page after adding a product to the cart is logged in successfully, but **their cart is emptied**.

---

## Environment

* WordPress: fresh install
* WooCommerce: latest
* WooCommerce Stripe Gateway (official): active, **not need to configure**
* WooPayments: active, **no need to configure**
* Jetpack: active and **connected to a **[**WordPress.com**](<http://WordPress.com>)** account****<br>****- Jetpack Sync has to be enabled (should be the case by default)**
  * `jetpack_sync_settings_disable = 0` in `wp_options`

---

## Steps to Reproduce

1. Install WooCommerce, complete basic setup, and create a product.
2. Go to **WooCommerce → Settings → Accounts & Privacy**:
   * Disable **Guest checkout**
   * Disable  **Allow customers to create an account During checkout**
3. Install **WooPayments** (no configuration needed).
4. Install the **WooCommerce Stripe plugin** (no configuration needed).
5. Install **Jetpack**, connect it to a [WP.com](<http://WP.com>) account.
6. Add a product to the cart.
7. Register a new account from the **My Account** page.

**Result:**  <br>After registration/login, the cart is wiped instead of being transferred to the newly registered customer.

---

## Related Context & Findings

This behavior closely resembles a previously fixed issue involving **nonce mismatches during 3DS flows** when Jetpack Sync was enabled:

* Original issue reference:  <br>[https://github.com/woocommerce/woocommerce/issues/34464#issuecomment-1596243046](<https://github.com/woocommerce/woocommerce/issues/34464#issuecomment-1596243046>)
* WooPayments fix:
  * PR: [https://github.com/Automattic/woocommerce-payments/pull/8084](<https://github.com/Automattic/woocommerce-payments/pull/8084>)
  * Commit: [https://github.com/Automattic/woocommerce-payments/pull/8084/commits/e346ff06b1d5a02b317299a503b59db3a0ff67eb](<https://github.com/Automattic/woocommerce-payments/pull/8084/commits/e346ff06b1d5a02b317299a503b59db3a0ff67eb>)

---

## So why is it happening again when it's fixed already?

The WooCommerce Stripe plugin contains the **same method**, using the **same hooks and priority**, but **without** the conditional logic that was added to WooPayments:

```php
/**
 * Proceed with current request using new login session (to ensure consistent nonce).
 */
public function set_cookie_on_current_request( $cookie ) {
    $_COOKIE[ LOGGED_IN_COOKIE ] = $cookie;
}
```

Locations:

* Legacy checkout:  <br>[https://github.com/woocommerce/woocommerce-gateway-stripe/blob/c5c0fe2bd15b4923f22c0b686f92256a59cb9637/includes/class-wc-gateway-stripe.php#L847](<https://github.com/woocommerce/woocommerce-gateway-stripe/blob/c5c0fe2bd15b4923f22c0b686f92256a59cb9637/includes/class-wc-gateway-stripe.php#L847>)
* UPE checkout:  <br>[https://github.com/woocommerce/woocommerce-gateway-stripe/blob/c5c0fe2bd15b4923f22c0b686f92256a59cb9637/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L348](<https://github.com/woocommerce/woocommerce-gateway-stripe/blob/c5c0fe2bd15b4923f22c0b686f92256a59cb9637/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L348>)

My theory is that this is a **race condition caused by overlapping hooks between WooPayments and the Stripe plugin**.

## Suggested Fixes

1. **From Jetpack's side**
   * I don’t have enough context around how the Jetpack connection manages authentication and session state to confidently suggest a fix there. 😅
2. **Stripe plugin alignment with WooPayments (tested below)**
   * Align the Stripe plugin logic with the fix already implemented in WooPayments.
   * Not sure whether it will cause additional side effects, but I can confirm I'm no longer able to replicate the issue after adding the following:

```php

/* 
* To replace current methods in:
* woocommerce-gateway-stripe/includes/class-wc-gateway-stripe.php
* woocommerce-gateway-stripe/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
*/
public function set_cookie_on_current_request( $cookie ) {
    if (
        defined( 'WOOCOMMERCE_CHECKOUT' ) &&
        WOOCOMMERCE_CHECKOUT &&
        did_action( 'woocommerce_created_customer' ) > 0
    ) {
        $_COOKIE[ LOGGED_IN_COOKIE ] = $cookie;
    }
}
```

---

thuautp:
## Summary

- Fixes a race condition where Stripe's `set_cookie_on_current_request` fired on **all** `set_logged_in_cookie` actions (including My Account registration), not just during checkout
- When Stripe + WooPayments + Jetpack are all active, this caused the session cookie to be overwritten mid-request during registration, causing WooPayments/Jetpack to read a stale nonce and blank the cart
- The fix mirrors the guard already used by WooPayments: only update `$_COOKIE` when `WOOCOMMERCE_CHECKOUT` is defined **and** `woocommerce_created_customer` has fired (i.e. we're in a genuine guest-checkout new-customer flow)

## Root Cause

`set_cookie_on_current_request` was hooked unconditionally onto `set_logged_in_cookie`:

```php
// Fires on every login, including My Account registration
add_action( 'set_logged_in_cookie', [ $this, 'set_cookie_on_current_request' ] );
```

The method then blindly updated `$_COOKIE` for the current PHP request, which interfered with WooPayments + Jetpack's own session handling during My Account registration.

## Fix

```php
public function set_cookie_on_current_request( $cookie ) {
    if (
        defined( 'WOOCOMMERCE_CHECKOUT' ) &&
        WOOCOMMERCE_CHECKOUT &&
        did_action( 'woocommerce_created_customer' ) > 0
    ) {
        $_COOKIE[ LOGGED_IN_COOKIE ] = $cookie;
    }
}
```

## Test plan

- [x] `test_set_cookie_on_current_request_does_not_set_cookie_outside_checkout` — no cookie set outside checkout (My Account registration)
- [x] `test_set_cookie_on_current_request_does_not_set_cookie_during_checkout_before_customer_creation` — no cookie set during checkout before guest customer is created
- [x] `test_set_cookie_on_current_request_sets_cookie_during_checkout_after_customer_creation` — cookie IS set during checkout after new customer created (the 3DS nonce-consistency use case)
- [x] Manually verify: register on My Account page with Stripe + WooPayments + Jetpack active → cart should remain intact
- [x] Manually verify: guest checkout creating a new account → payment still processes correctly

Fixes #4875


daledupreez:
@thuautp, the PR seems to include a number of unrelated changes beyond the cookie fix. Can you clean up the PR to the changes you documented in the PR description? It definitely sounds like this would be a valuable fix.

thuautp:
> @thuautp, the PR seems to include a number of unrelated changes beyond the cookie fix. Can you clean up the PR to the changes you documented in the PR description? It definitely sounds like this would be a valuable fix.

@daledupreez done. Please review and let me know if anything else needs to be fixed. Thanks!

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -395,10 +395,19 @@ public function filter_saved_payment_methods_list( $list, $customer_id ) {
 	/**
 	 * Proceed with current request using new login session (to ensure consistent nonce).
 	 *
+	 * Only applies during checkout after a customer has been created, to avoid a race condition
+	 * with WooPayments + Jetpack that causes the cart to be emptied on My Account registration.
```
daledupreez: Can we also reference the issue for more details?
```suggestion
	 * Only make this change during checkout after a customer has been created, to avoid a race condition
	 * that causes the cart to be emptied on My Account registration.
	 *
	 * @see https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4875
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -395,10 +395,19 @@ public function filter_saved_payment_methods_list( $list, $customer_id ) {
...
+	 *
 	 * @param string $cookie New cookie value.
 	 */
 	public function set_cookie_on_current_request( $cookie ) {
-		$_COOKIE[ LOGGED_IN_COOKIE ] = $cookie;
+		if (
+			defined( 'WOOCOMMERCE_CHECKOUT' ) &&
+			WOOCOMMERCE_CHECKOUT &&
```
daledupreez: Is it worth calling `is_checkout()` rather than relying on the constant?

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -4372,6 +4372,56 @@ public function test_add_email_currency_conversion_notice_outputs_plain_text_for
...
+	public function test_set_cookie_on_current_request_sets_cookie_during_checkout_after_customer_creation() {
+		define( 'WOOCOMMERCE_CHECKOUT', true );
+		do_action( 'woocommerce_created_customer', 1, [], false );
+
+		$gateway = new WC_Stripe_UPE_Payment_Gateway();
+		$gateway->set_cookie_on_current_request( 'test_cookie_value' );
+
+		$this->assertSame( 'test_cookie_value', $_COOKIE[ LOGGED_IN_COOKIE ] ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput
+	}
+
```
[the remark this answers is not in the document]
daledupreez: The `WOOCOMMERCE_CHECKOUT = false` case would be good to add. It would also be cleaner to use a data provider for the test cases, but that's not a blocker. I ended up adding this in f33c91307989f3afe041432c33c24139bed51bdf.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -378,10 +378,19 @@ public function filter_saved_payment_methods_list( $list, $customer_id ) {
...
+			did_action( 'woocommerce_created_customer' ) > 0
+		) {
+			$_COOKIE[ LOGGED_IN_COOKIE ] = $cookie;
+		}
```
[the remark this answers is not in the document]
daledupreez: The protections didn't exist before this PR, so this is not a blocker.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -395,10 +395,19 @@ public function filter_saved_payment_methods_list( $list, $customer_id ) {
...
```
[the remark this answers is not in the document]
daledupreez: I don't believe we need this -- logins don't occur within the Store API context, so we shouldn't trip over this specific edge case.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: I think we can tackle this in a follow-up if we decide it's worth it.

daledupreez (approved):
@thuautp, I just reviewed the changes and I think the core change is ready to merge! Thanks for contributing!

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: I updated the code to use `is_checkout()` in commit 9ab5d217aedcb1a1aa2f8c9d2c183eae2f8bc9ad as the tests were hanging with the constant redefinition.

Issue #4315, opened 2025-05-15
Prevent excessive Stripe API calls when keys are invalid


daledupreez:
### Description

When the current Stripe API key becomes invalid (which we need to investigate separately), we become unable to populate our local cached data, and we end up triggering large numbers of API requests to Stripe. In turn, this can cause notable slow-downs on the site due to us tying up thread waiting for HTTP 401 responses from Stripe.

### Steps to reproduce

We saw some sites trip over this yesterday as a result of their keys somehow being invalidated. The issue can also be manually triggered by doing the following with the Stripe plugin: (h/t @annemirasol)
1. If your site has a connection to Stripe, disconnect it
2. Connect (or reconnect) to Stripe in live mode
3. Switch to test mode
4. See repeated API calls

### Expected behavior

We should ensure that we have better protections in place, which may require a number of changes, including the following possibilities when we detect repeated HTTP 401 responses from Stripe (especially for basic API calls like the `account` API):
 * We should prevent further API calls from being made, as we likely need to reconnect - exact approach here is to be discussed and confirmed
 * We should probably disable functionality within the plugin, e.g. express checkout and payment buttons, most likely by updating our state to reflect that we are no longer connected to Stripe. (This may address the first item above as well.)
 * We should highlight the problem to users in the UI -- if we're in live mode, this probably warrants a really intrusive admin banner.
 * We should send the user an email (especially if we're in live mode), that flags the problem on the site.
   - If we're the only or primary payment method on the site, we should include language that flags the issue even more prominently.

PR #5238, opened 2026-04-08
Fix: cache failed Stripe account lookups

### LINKED ISSUE #5239: Cache failed Stripe account lookups to avoid repeated 401 calls
### Description

Admin pages can become very slow when the Stripe account lookup fails because the plugin repeatedly retries the Stripe `/v1/account` endpoint instead of caching the failed account lookup.

In my testing, a single admin page load made hundreds of sequential uncached calls to `https://api.stripe.com/v1/account`. With invalid or missing API credentials, each request returns HTTP 401 and blocks the request thread, producing 50+ second admin page loads.

### Steps to reproduce

1. Set up a WooCommerce Stripe Gateway site with Stripe enabled and test mode active.
2. Configure invalid test Stripe keys, for example an invalid `sk_test_...` value.
3. Enable Express Checkout / Apple Pay settings so account-country/payment-method availability checks run in admin.
4. Clear the Stripe account data cache.
5. Load a WooCommerce admin/settings page that initializes the Stripe gateway/payment methods.
6. Inspect the WooCommerce Stripe logs or outgoing HTTP traffic.

### Expected behavior

When account data retrieval fails due to invalid credentials or another failed response, the account layer should cache the failure briefly and return the expected empty account response during that TTL instead of retrying Stripe repeatedly within the same admin request/page load.

### Actual behavior

`WC_Stripe_Account::get_cached_account_data()` caches successful account data, but failed responses are returned as an empty array without writing any cache state. Subsequent callers, including Apple Pay registration/payment method availability checks and ACH account-country checks, treat that empty result as a cache miss and call Stripe again.

### Screenshots

N/A. This is visible through server-side HTTP/API logs rather than the rendered UI.

### System report

<details>

Reproduced locally on the repository Docker environment:

- WooCommerce Stripe Gateway: 10.5.3 / develop
- WordPress: 6.9.x local test site
- WooCommerce: 10.6.x local test site
- PHP: 7.4 Docker image
- Stripe mode: test mode with invalid test secret key

</details>

### Additional context

This is related to the broader, closed issue #4315, but the current failure path has a specific account-cache gap that still allows repeated `/v1/account` calls before the generic API invalid-key threshold can protect the request.

Validation from the local fix branch:

- Invalid test keys configured with Stripe logging enabled.
- Stripe settings admin page loaded through Playwright Chromium.
- WooCommerce Stripe log showed exactly one `GET account` request and one 401 for the page load.
- DB cache contained `wcstripe_cache_test_account_data_failure` with a 60s TTL.

dmallory42:
Fixes #5239

`WC_Stripe_Account::get_cached_account_data()` only cached successful, non-empty account responses. When Stripe returned a 401 for an invalid key, callers received `[]`, but no account-cache state was written. Subsequent admin/UI payment-method checks treated that as a cache miss and re-entered the same Stripe `/v1/account` request path repeatedly.

This PR adds a short-lived negative cache for failed account lookups. Failed lookups now return the expected empty account response without repeatedly retrying Stripe during the failure TTL. The failure marker is cleared on successful account fetches and when the account cache is cleared.

This is a focused follow-up to the earlier #4315. It keeps the behavior scoped to the account data cache layer instead of changing each Apple Pay/payment-method caller.

## Testing instructions

1. Set up the local Docker site with `npm run up`.
2. Configure Stripe test mode with an invalid test secret key.
3. Enable Stripe logging and Express Checkout / Apple Pay settings.
4. Clear Stripe cache rows matching `wcstripe_cache_%`.
5. Load the Stripe settings admin page.
6. Confirm the page loads normally and the WooCommerce Stripe log contains only one `GET account` request / 401 for the page load.
7. Confirm the database contains `wcstripe_cache_test_account_data_failure` with a short TTL.

Automated checks run:

- `npm install`
- `npm run test:php -- --filter WC_Stripe_Account_Test` passed: 28 tests, 43 assertions
- `./vendor/bin/phpcs --standard=phpcs.xml.dist -n includes/class-wc-stripe-account.php tests/phpunit/class-wc-stripe-account-test.php tests/phpunit/helpers/class-wc-helper-stripe-api.php` passed
- `npm run phpstan` passed via the pre-push hook
- Browser sanity check with Playwright Chromium against the local Docker admin Stripe settings page with invalid test keys: page loaded with HTTP 200 in about 1.1s; WooCommerce Stripe log showed one `GET account` request and one 401; DB cache contained `wcstripe_cache_test_account_data_failure` with a 60s TTL.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (commented):
I'm a little confused as to why we need to add another layer of 401 detection, we already have a circuit breaker to handle the 401 in place:

In WC_Stripe_API::retrieve() (line 301-393) we:
1. Track consecutive 401 errors via `INVALID_API_KEY_ERROR_COUNT_CACHE_KEY`
2. After 5 consecutive 401s, returns null immediately without any API call
3. Use a 2-hour TTL on the block
4. Reset on successful responses or key updates

I'm not a fan of caching invalid values.... but if we are going to implement this, we should also explicitly invalidate the cache when starting the re-connect flow(s) to avoid stale data (for that minute) that would likely bring back the re-connect loop notice/banner problems we had last year.

dmallory42:
Thanks for looking, @diegocurbelo! You likely have more context than I do on the original 401/circuit-breaker work, so I may be missing some history here.

The reason I approached it this way is that I hit this as a real issue locally: once account lookup failed, multiple admin/payment-method paths kept treating the empty account response as a cache miss and repeatedly retrying /v1/account, which made the page very slow before the existing API-level breaker had a chance to help.

So I was thinking of this less as adding a new circuit breaker, and more as filling an account-cache gap. `WC_Stripe_API::retrieve()` is still the broader invalid-key safety valve, but  `WC_Stripe_Account::get_cached_account_data()` currently doesn’t distinguish between “cache miss” and “recent failed lookup”, which is what seems to trigger the fan-out.

That said, I agree the recovery paths should explicitly clear or bypass this failure state, especially reconnect / OAuth-init flows, so we don’t hold onto stale failure state during a user-initiated reconnect.

Let me know what you think!

malithsen (commented):
I'm a bit hesitant about introducing changes to account caching, given the repeated issues we've had last year 😅 Given that manual key entry is no longer possible, it seems to me that a merchant running into this specific scenario is very rare. It might still be possible to achieve the same outcome for OAuth failures, but I haven't been able to replicate it so far.

wjrosa:
@dmallory42 Took a closer look at the concerns here and have a few suggestions that I think address Diego's and Malith's points.

**On the "why a second 401 layer" question (@diegocurbelo):** the existing `INVALID_API_KEY_ERROR_COUNT` circuit breaker in `WC_Stripe_API::retrieve()` only trips after 5 consecutive 401s and then locks out for 2 hours. The bug in #5239 is the fan-out that happens *before* the 5th 401 — a single admin page initializing UPE methods + ECE/Apple Pay + ACH checks can fire dozens to hundreds of `/v1/account` calls in one request, all before the breaker engages. The 60s failure marker is a same-request short-circuit, not a duplicate of the 2-hour lockout. Worth calling that out explicitly in the PR description so it doesn't read as a duplicate mechanism.

**On the stale-data / re-connect banner risk (@diegocurbelo):** this is a real concern as-is. `WC_Stripe_Connect::save_stripe_keys()` clears `INVALID_API_KEY_ERROR_COUNT_CACHE_KEY` (line 289) but never calls `account->clear_cache()`, so a warm `ACCOUNT_CACHE_FAILURE_KEY` could survive a successful OAuth re-connect for up to 60s and re-trigger the very banner loop you're worried about. Suggest adding `WC_Stripe::get_instance()->account->clear_cache();` to `WC_Stripe_Connect::clear_caches_after_key_update()` so all OAuth/key-update paths wipe both the account data and the failure marker atomically.

**On reproducibility (@malithsen):** agreed the manual-key trigger is now rare, but the code gap is verifiable from `develop` independent of trigger — any path that leaves Stripe returning a non-success (revoked OAuth token, mid-migration legacy keys, transient Stripe-side disruption) hits the same amplification. As an optional tightening, we could fire the failure marker only when `WC_Stripe_API::retrieve()` returns `null` (the explicit auth-failure signal at line 401), so transient network errors don't lock callers out for 60s. Happy to defer that as a follow-up if we want to keep scope tight.

**On the CodeRabbit nitpick:** the force-refresh bypass test is worth adding — it's a one-case data provider and locks down the documented `$force_refresh = true` semantics. The suggested diff looks good.

---
_Drafted with assistance from Claude Code; reviewed and posted by me._

daledupreez:
@dmallory42, I am going to close this PR for the time being, as it hasn't received any updates in a few months. Please feel free to re-open this PR or a new one when you're in a position to work on the underlying issue. Thanks for taking this on to start with!

PR #5241, opened 2026-04-08
Add Link express checkout customize settings page

diegocurbelo:
Fixes STRIPE-1038

Adds a dedicated Customize settings page for Link by Stripe, giving merchants independent control over button placement and size — matching the pattern already used by Amazon Pay.

Previously, Link shared location settings with Apple Pay / Google Pay and had no button size control, so merchants couldn't manage Link placement independently or size it to match other express checkout buttons.

- _Settings page_: Adds a new `?area=link` admin page with enable/disable checkbox, location checkboxes (Product, Cart, Checkout), button size radio (Small/Default/Large), and a live Link button preview.
- _Main settings_: Replaces the "Read more" link on the Link row with a "Customize" link navigating to the new page.
- _Backend_: Adds `link_button_locations` and `link_button_size` settings to the settings definition, REST API, and express checkout helper. Link now resolves its own locations instead of falling through to Apple Pay / Google Pay defaults.
- _Frontend_: It now applies the Link-specific button height when rendering the Link button instance.

Screenshot 2026-04-08 at 11 53 59

Screenshot 2026-04-08 at 11 54 11

## Testing instructions

1. Navigate to WooCommerce > Settings > Payments > Stripe > Express checkouts and verify the Link row shows "Customize" instead of "Read more"
2. Click "Customize" and verify the settings page loads with enable checkbox, location checkboxes, button size radio, and a live Link button preview
3. Toggle locations and size, save, reload — verify settings persist
4. Disable Link on a specific location (e.g. product page), visit that page on the storefront, and verify the Link button does not appear
5. Change Link button size, visit the storefront, and verify the Link button renders at the configured height
Verify Apple Pay / Google Pay and Amazon Pay settings and behavior are unaffected

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/entrypoints/link-settings/__tests__/link-settings-locations.test.js
```diff
@@ -0,0 +1,116 @@
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import LinkSettingsSection from '../link-settings-section';
+import { useEnabledPaymentMethodIds, useLinkLocations } from 'wcstripe/data';
+
+jest.mock( 'wcstripe/data', () => ( {
+	useEnabledPaymentMethodIds: jest.fn(),
+	useLinkLocations: jest.fn(),
+	useLinkButtonSize: jest.fn().mockReturnValue( [ 'default' ] ),
+} ) );
```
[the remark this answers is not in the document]
diegocurbelo: Adding an unused import just for symmetry with the mock adds noise.

client/entrypoints/link-settings/express-checkout-preview-component.js
```diff
@@ -0,0 +1,90 @@
...
+		<InlineNotice icon status="error" isDismissible={ false }>
+			{ __(
+				'Failed to preview the Link by Stripe button. ' +
+					'Ensure your store uses HTTPS on a publicly available domain ' +
+					"and you're viewing this page in a Safari or Chrome browser. " +
+					'Your device must be configured to use Link.',
+				'woocommerce-gateway-stripe'
+			) }
+		</InlineNotice>
+	);
```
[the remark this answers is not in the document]
diegocurbelo: This component is fundamentally a Stripe Elements wrapper.
The cost-to-value ratio of mocking Stripe Elements for this thin wrapper is poor.
The important behavior (size mapping) is already tested indirectly through the PHP get_link_button_height() test and the settings interaction tests.

client/entrypoints/link-settings/link-settings-section.js
```diff
@@ -0,0 +1,178 @@
...
+				'Large {{helpText}}(56 px){{/helpText}}',
+				'woocommerce-gateway-stripe'
+			)
+		),
+		value: 'large',
+	},
+];
+
+const LinkSettingsSection = () => {
+	const [ size, setSize ] = useLinkButtonSize();
```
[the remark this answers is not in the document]
diegocurbelo: Fixed in [3513864](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5241/commits/35138641d050b0d2939d684ca715d87c7e8561c4)

client/entrypoints/link-settings/__tests__/link-settings-locations.test.js
```diff
@@ -0,0 +1,123 @@
...
+		expect( updateLinkLocationsHandler ).toHaveBeenLastCalledWith( [
+			'product',
+		] );
+
+		await userEvent.click( screen.getByText( 'Checkout' ) );
+
+		expect( updateLinkLocationsHandler ).toHaveBeenLastCalledWith( [
+			'checkout',
+		] );
+	} );
```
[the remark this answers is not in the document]
diegocurbelo: The component gets linkLocations from the useLinkLocations hook, which is mocked to always return ['checkout', 'product', 'cart']. The mock never changes between clicks because there's no real store behind it.

The current test correctly validates that the handler is called with the right argument for each individual toggle

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -251,6 +266,8 @@ public function get_settings() {
 				/* Settings > Express checkouts */
 				'amazon_pay_button_size'                => $this->gateway->get_validated_option( 'amazon_pay_button_size' ),
 				'amazon_pay_button_locations'           => $this->gateway->get_validated_option( 'amazon_pay_button_locations' ),
+				'link_button_size'                      => $this->gateway->get_validated_option( 'link_button_size' ),
+				'link_button_locations'                 => $this->gateway->get_validated_option( 'link_button_locations' ),
```
daledupreez, line 270: I am wondering if this is the right approach here: it feels like we should structure button sizes and locations quite differently. In my mind, that is driven by the following thoughts:
 - All our express checkout buttons on a page should use the same size - we should not _allow_ different sizes to be specified for buttons that will be displayed together
    - That said, we might want to allow different button sizes in different contexts, e.g. large buttons on the product page, and smaller buttons in the cart and checkout
  - Controlling which express methods are shown on which pages feels wrong when it is all done from this level -- I've flagged this in some of our design discussions, but it feels like it would be way more powerful to have a single screen that controls which express checkout methods show up in specific contexts.
     - That doesn't dictate how we store this data, but it feels like it would be much more efficient to store `express_checkout_locations` that has a structure like the following (which would also allow for ordering of express checkout buttons):
```php
$express_checkout_locations = [
	'product'  => [ 'amazon_pay', 'link', 'apple_pay', 'google_pay' ],
	'cart'     => [ 'apple_pay', 'google_pay', 'link' ],
	'checkout' => [ 'link' ],
];
```

That may be somewhat ambitious for this PR, but I do worry about the possible issues stemming from another somewhat redundant value that we will need to migrate.

daledupreez (commented):
I am also not seeing any logic that replicates the current effective Link button locations, which should be the same as the Apple Pay/Google Pay locations.

includes/admin/stripe-settings.php
```diff
@@ -278,6 +278,34 @@
...
+			'type'        => 'select',
+			'description' => __( 'Select the size of the button.', 'woocommerce-gateway-stripe' ),
+			'default'     => 'default',
+			'desc_tip'    => true,
+			'options'     => [
+				'small'   => __( 'Small (40px)', 'woocommerce-gateway-stripe' ),
+				'default' => __( 'Default (48px)', 'woocommerce-gateway-stripe' ),
+				'large'   => __( 'Large (56px)', 'woocommerce-gateway-stripe' ),
+			],
+		],
```
Mayisha, line 308: An observation here. Previously Link used to share the settings of Google/Apple Pay. Suppose Google/Apple Pay were disabled on the cart and checkout page, it made Link disabled only product page as well. 

But after switching to this branch Link got enabled on the product page. May be we need a migration for the location.

Another observation is, after switching to this branch Link customization page shows that Link is enabled on product, cart, checkout all three pages. However, I do not see the Link button on checkout page until I click the save button on Link customization page.

To reproduce-
- Install the latest version of Stripe in a JN site.
- Enable ECE for cart and checkout.
- Build this branch and update the plugin of the JN site with this build.
- Go to Link customization page and see all three locations are enabled.
- Go to the checkout page and see that Link is not there.
- Click the save button on Link customization page.
- Check the checkout page again and see that Link is not present.

includes/admin/class-wc-rest-stripe-settings-controller.php
Mayisha: +1 to using the same size for all ECE buttons. Came across this view when the sizes are different. However this only happens in product and classic pages. Block cart, checkout looks fine and takes the same size for all.

Image

However, I think this is a separate UX concern to decide where the button size setting should be.

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -568,6 +586,29 @@ private function update_amazon_pay_settings( WP_REST_Request $request ) {
...
+	 * Updates appearance attributes of the Link by Stripe button.
+	 *
+	 * @param WP_REST_Request $request Request object.
+	 *
+	 * @return void
+	 */
+	private function update_link_settings( WP_REST_Request $request ) {
+		$attributes = [
+			'link_button_size'      => 'link_button_size',
+			'link_button_locations' => 'link_button_locations',
```
Mayisha, line 599: Nit: we can avoid the key-value pair here and only use the values array.

diegocurbelo: Nice catch, I also updated the same occurrence in the amazon_pay_settings in [c35e7ba](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5241/commits/c35e7ba5f06dd5f1edb298ae204e33dbb3f17d53)

includes/admin/stripe-settings.php
diegocurbelo: Added the migration in [b8e6d22](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5241/commits/b8e6d221c3340de7dc18e9ce69a6f276e6e9db32)

includes/admin/class-wc-rest-stripe-settings-controller.php
diegocurbelo: > I am wondering if this is the right approach here: it feels like we should structure button sizes and locations quite differently.

This is a fair point and I think a unified data model is the cleaner direction.
However, restructuring the settings to something like `[ location => [ methods ] ]` would need matching migrations, and settings-UI unification, and as Mayisha mentioned, we have a similar problem with the button sizing; we need a separate UX for the unified button size setting.

I propose we ship this PR as-is so the Link customize page (and the migration that preserves existing merchants' effective locations) lands, and we'll track the unified express-checkout locations/sizing model in STRIPE-1093.

Mayisha: One clarification here. My proposal is to unify the size only so that the UX looks nice. I think the buttons should have their own location settings.

client/entrypoints/express-checkout/index.js
```diff
@@ -356,8 +356,22 @@ jQuery( function ( $ ) {
...
+			if ( expressPaymentType === EXPRESS_PAYMENT_METHOD_SETTING_LINK ) {
+				const linkHeight = parseInt(
+					getExpressCheckoutData( 'link_button_height' ) || '48',
+					10
+				);
+				buttonStyleSettings.buttonHeight = Math.min(
+					Math.max( linkHeight, 40 ),
+					55
+				);
+			}
```
daledupreez: Two comments here:
 - It feels like the Link-specific logic should be in the shared function, not outside of it
 - It also looks like we may not be setting Amazon Pay settings correctly based on this logic being new (at least for classic cart/checkout)

The first issue may be something for this PR, but the second should be tackled in a separate PR.

client/entrypoints/link-settings/express-checkout-preview-component.js
```diff
@@ -0,0 +1,91 @@
...
+import InlineNotice from 'components/inline-notice';
+import { PAYMENT_METHOD_LINK } from 'wcstripe/stripe-utils/constants';
+
+const buttonSizeToPxMap = {
+	small: 40,
+	default: 48,
+	large: 56,
+};
+
+const ExpressCheckoutPreviewComponent = ( { size } ) => {
```
daledupreez, line 17: Not blocking, but I think this may be the third copy of this element. It would be worth considering whether we can refactor the component for re-use.

That can be something we do in a follow-up PR.

client/entrypoints/link-settings/express-checkout-preview-component.js
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	return (
+		<InlineNotice icon status="error" isDismissible={ false }>
+			{ __(
+				'Failed to preview the Link by Stripe button. ' +
+					'Ensure your store uses HTTPS on a publicly available domain ' +
+					"and you're viewing this page in a Safari or Chrome browser. " +
+					'Your device must be configured to use Link.',
+				'woocommerce-gateway-stripe'
```
daledupreez, line 83: Nit: I am not sure the Safari/Chrome restriction applies to Link. It might apply to Apple Pay/Google Pay, but I am doubtful of that. I am also not sure that the device needs any configuration, or if Link will only preview on public URLs.
```suggestion
				'Failed to preview the Link by Stripe button. ' +
					'Ensure your store uses HTTPS on a publicly available domain.',
				'woocommerce-gateway-stripe'
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-files-api-delivery.php
```diff
@@ -276,7 +276,7 @@ private function upload_to_files_api( string $file_path ): array {
 			}
 		} finally {
 			// Note: Intelephense falsely reports curl_close is deprecated.
-			curl_close( $curl_handle );
+			curl_close( $curl_handle ); // phpcs:ignore Generic.PHP.DeprecatedFunctions.Deprecated
```
daledupreez: This feels like it should not be part of this PR.

includes/class-wc-stripe-status.php
```diff
@@ -158,7 +158,15 @@ public function render_status_report_section() {
 					<mark class="yes"><span class="dashicons dashicons-yes"></span>
 					<?php
 					if ( $express_checkout_helper->is_express_checkout_enabled() ) {
-						$express_checkout_enabled_locations = $express_checkout_helper->get_button_locations();
+						$express_checkout_enabled_locations = array_values(
```
daledupreez, line 161: Good catch!

includes/migrations/class-migrate-link-button-locations.php
```diff
@@ -0,0 +1,47 @@
...
+ * Class Migrate_Link_Button_Locations
+ *
+ * Before Link had a dedicated `link_button_locations` setting, it inherited its
+ * locations from the payment request (Apple Pay / Google Pay) button. This
+ * migration copies those locations into the new setting on upgrade so existing
+ * merchants keep their configured Link placement.
+ *
+ * @since 10.7.0
+ */
+class Migrate_Link_Button_Locations {
```
daledupreez: Nit: I think we should include either a namespace or `WC_Stripe` in the class name to prevent unintended collisions.
```suggestion
class Migrate_WC_Stripe_Link_Button_Locations {
```

tests/phpunit/admin/migrations/class-migrate-link-button-locations-test.php
```diff
@@ -0,0 +1,70 @@
...
+	public function set_up() {
+		parent::set_up();
+
+		// Seed the settings option so update_main_stripe_settings() does not hit the
+		// "option does not exist yet" branch of the pre_update_option_woocommerce_stripe_settings
+		// filter, which would otherwise merge in gateway defaults (including link_button_locations)
+		// and break these assertions.
+		add_option( WC_Stripe_Helper::SETTINGS_OPTION, [] );
+
+		$this->migration = new Migrate_Link_Button_Locations();
```
daledupreez: We are not using this instance for any mocking, so it feels somewhat redundant. It may be cleaner if each test method instantiates a new migration instance.

includes/class-wc-stripe.php
```diff
@@ -216,10 +216,12 @@ public function init() {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-allowed-payment-request-button-types-update.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-sepa-tokens-for-other-methods-settings-update.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-migrate-payment-request-data-to-express-checkout-data.php';
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-migrate-link-button-locations.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-account.php';
 
 		new Allowed_Payment_Request_Button_Types_Update();
 		new Migrate_Payment_Request_Data_To_Express_Checkout_Data();
+		new Migrate_Link_Button_Locations();
```
daledupreez: I don't like that we are auto-instantiating all this migration code on all requests, when we generally want them to trigger code on version updates.

Not for this PR, but it may be worth considering this as a cleanup item for a future PR.

daledupreez (approved):
I didn't test the upgrade path extensively, but things are looking good to me.

We may want to consider whether

includes/agentic-commerce/class-wc-stripe-agentic-commerce-files-api-delivery.php
diegocurbelo: Removed in [b07061a](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5241/commits/b07061af60d34e79065354316ddb6d6866093fbd)

includes/migrations/class-migrate-link-button-locations.php
diegocurbelo: Renamed in [d07d898](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5241/commits/d07d89810b6b05f052d977c651c5583e8a5d3e56)

includes/class-wc-stripe.php
diegocurbelo: Refactored to use the migrations merged into develop

tests/phpunit/admin/migrations/class-migrate-link-button-locations-test.php
diegocurbelo: Refactored in [f1b8ca2](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5241/commits/f1b8ca2ed8cb63b14f4d4af0199909b752919a7b)

client/entrypoints/link-settings/express-checkout-preview-component.js
diegocurbelo: Updated in [1577a31](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5241/commits/1577a31a0dfe98b3d713dc31cc0a5c4499dacbc3)

client/entrypoints/link-settings/express-checkout-preview-component.js
diegocurbelo: Created a new issue to track this: STRIPE-1200

client/entrypoints/express-checkout/index.js
diegocurbelo: Fair point, moved the logic to the shared function in [d1df92e](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5241/commits/d1df92e79da82b4797c896f657c6d7a0c80770e2)

And created an issue (STRIPE-1201) to track the Amazon Pay Appearance Size setting not being used

diegocurbelo:
Added 3 follow up issues:
- STRIPE-1093: Unify express checkout button locations & sizing across methods
- STRIPE-1200: Extract a shared Express Checkout preview component for the settings pages (Google Pay/Apple Pay/Amazon Pay/Link)
- STRIPE-1201: Amazon Pay button size setting is not applied (the value from GPay/Apple pay is used)

client/entrypoints/link-settings/link-settings-section.js
```diff
@@ -0,0 +1,188 @@
...
+
+	const checkoutPageId = getSetting( 'storePages' )?.checkout?.id;
+
+	return (
+		<Card className="express-checkout-settings">
+			<CardBody>
+				<Notice status="warning" isDismissible={ false }>
+					{ interpolateComponents( {
+						mixedString: __(
+							'Some appearance settings may be overridden by the express payment section of the ' +
```
Mayisha: We should apply the `isButtonStyleOverridden` [check](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/fix/STRIPE-1038/add-link-express-checkout-cutomize-settings-ui/client/entrypoints/express-checkout-settings/express-checkout-settings-section.js#L162) in the Link settings page.

Mayisha (approved):
Nice work. 💯 

I have one feedback regarding `isButtonStyleOverridden`. Otherwise looks good to me.

PR #5245, opened 2026-04-08
Fix handling of payer phone in checkout sessions

wjrosa:
Fixes STRIPE-1069

This PR fixes two related issues with payer phone handling in Checkout Sessions:

1. **Conditional phone collection on Stripe side**: `phone_number_collection` is now only enabled in the Checkout Session request when WooCommerce's checkout phone field is set to "Required". Previously it was always enabled for logged-in users, which caused Stripe to require a phone number even when the store had it set to optional or hidden.

2. **Conditional phone forwarding on the client side**: The phone number read from the DOM is now only sent to Stripe when both conditions are met — the customer is logged in AND phone collection is required. Previously it was sent whenever the customer was logged in regardless of the phone field setting.

Both changes align Stripe's phone collection behavior with the store's WooCommerce settings (`WooCommerce > Settings > Accounts & Privacy > Phone field`).

## Testing instructions

**Prerequisites:** Enable the Optimized Checkout Suite and have a logged-in test account.

**Scenario 1 — Phone field set to Required (default)**
1. Go to WooCommerce > Settings > Accounts & Privacy. Set "Phone field" to **Required**.
2. Add a product to the cart and proceed to the Blocks checkout.
3. Complete checkout using the Optimized Checkout (Checkout Sessions) flow.
4. Confirm the Stripe Checkout Session is created successfully and the phone number is collected/forwarded without errors.

**Scenario 2 — Phone field set to Optional**
1. Set "Phone field" to **Optional**.
2. Add a product to the cart and proceed to the Blocks checkout.
3. Complete checkout using the Optimized Checkout flow.
4. Confirm the Stripe Checkout Session is created without `phone_number_collection` being requested, and checkout completes without errors (Stripe should not prompt for or require a phone number).

**Scenario 3 — Phone field Hidden**
1. Set "Phone field" to **Hidden**.
2. Same steps as Scenario 2 — confirm phone is neither collected nor sent.

**Scenario 4 — Guest user (any phone setting)**
1. Log out and checkout as a guest.
2. Confirm checkout completes successfully (phone forwarding is skipped for guest users).

---

- [x] Covered with tests
- [ ] Tested on mobile (does not apply)

### Changelog entry

- [x] Fix - Only collect payer phone in Checkout Sessions when WooCommerce's checkout phone field is required, and only send it to Stripe when the customer is logged in and phone collection is required

**Post merge**

### DIFF
--- a/includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
+++ b/includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
@@ -56,8 +56,11 @@ class WC_Stripe_Checkout_Sessions_Ajax_Handler {
 					throw new Exception( __( 'Unable to create or retrieve Stripe customer.', 'woocommerce-gateway-stripe' ) );
 				}
 
-				$request['customer']                = $stripe_customer->get_id();
-				$request['phone_number_collection'] = [ 'enabled' => true ];
+				$request['customer'] = $stripe_customer->get_id();
+
+				if ( 'required' === get_option( 'woocommerce_checkout_phone_field', 'required' ) ) {
+					$request['phone_number_collection'] = [ 'enabled' => 'true' ];
+				}
 			}
 
 			$checkout_session = WC_Stripe_API::request( $request, 'checkout/sessions' );
--- a/changelog.txt
+++ b/changelog.txt
@@ -40,7 +40,7 @@ xxxx-xx-xx - version 10.6.0
 * Dev - Add incremental inventory sync for Agentic Commerce: tracks stock changes via WooCommerce hooks and uploads a minimal inventory_feed CSV to Stripe one minute after the first change
 * Dev - Collapse PHPUnit tests using data providers to reduce duplication and improve test isolation
 * Fix - Fix UPE style transition keys for font smoothing properties
-* Fix - Only collect payer phone in Checkout Sessions when WooCommerce's checkout phone field is required, and only send it to Stripe when the customer is logged in and phone collection is required
+* Fix - Only collect and send payer phone in Checkout Sessions when the WooCommerce phone field is required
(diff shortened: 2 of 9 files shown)


includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
```diff
@@ -56,8 +56,11 @@ public function create_checkout_session(): void {
 					throw new Exception( __( 'Unable to create or retrieve Stripe customer.', 'woocommerce-gateway-stripe' ) );
 				}
 
-				$request['customer']                = $stripe_customer->get_id();
-				$request['phone_number_collection'] = [ 'enabled' => true ];
+				$request['customer'] = $stripe_customer->get_id();
+
+				if ( 'required' === get_option( 'woocommerce_checkout_phone_field', 'required' ) ) {
+					$request['phone_number_collection'] = [ 'enabled' => 'true' ];
```
wjrosa: There was a bug here, it was supposed to be `'true'` instead of `true`.

Mayisha (approved):
Thanks for fixing the issue. Checkout with adaptive pricing is now working without any errors in classic checkout.

I have one concern here @wjrosa. Looks like in classic checkout as the phone field is marked as non-required, we are ignoring the phone number completely even if user inputs a phone number in that non-required field. Shouldn't we use the phone number in the request if it is present?

Approving to unblock the bug but I think we should consider the above case.

wjrosa:
Thanks for the review, @Mayisha !

> Shouldn't we use the phone number in the request if it is present?

That was my idea with the original implementation, but unfortunately, if we flag the checkout session as collecting the phone number, we need to provide it later in the flow (becomes required).

We could check if the shopper has a phone number when creating the checkout session (in the ajax handler) and set the `phone_number_collection` flag. But if the shopper deletes the phone number before submitting the checkout form, the form will fail. I'm not sure if that's a good trade-off.

diegocurbelo (approved):
The changes look good to me, and it works as described 🚢

PR #5240, opened 2026-04-08
Add optimized checkout performance notice with “move Stripe gateways to top” action

Mayisha:
Towards STRIPE-1030

## Changes proposed in this Pull Request:
- Added `OptimizedCheckoutFirstMethodNotice` component to show the notice
- Added a new REST endpoint in `WC_REST_Stripe_Settings_Controller` to move Stripe gateways to the top of `woocommerce_gateway_order`.
- Added helper methods in `WC_Stripe_Helper` to:
    - Detect whether Stripe is first among real gateways (ignoring _wc_pes_* entries).
    - Move Stripe gateways (stripe + stripe_*) to the top while preserving internal relative order.

Screenshot 2026-04-09 at 2 34 50 AM

## Testing instructions
- Go to **WooCommerce > Settings > Payments** page
- Enable COD, Cheque, PayPal etc., a few non-Stripe gateways.
- Move Stripe to 1st position among the gateways
- Go to the Stripe settings page and scroll down to the advanced settings section.
- Enable Optimized checkout
- There should be no notices below the OCS checkbox
- Now move Stripe to the 3rd/4th position in **WooCommerce > Settings > Payments** page
- Refresh the Stripe settings page
- You should see a notice below the OCS checkbox.
- Click the `Move to top` button.
- Refresh the **WooCommerce > Settings > Payments** page and verify that Stripe is moved to the first position
- The relative order of the other payment gateways should be unchanged.

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5247, opened 2026-04-08
Add optimized checkout performance notice on WooCommerce payments settings page

Mayisha:
Towards [STRIPE-1030](https://linear.app/a8c/issue/STRIPE-1030)

- `wp_localize_script` on `woocommerce_stripe_payment_gateways_page` with a minimal object.
- inject `OptimizedCheckoutFirstMethodNotice` into the Stripe list item (under `.woocommerce-list__item-text`), with a short MutationObserver fallback if the row appears after load.

Screenshot 2026-04-09 at 2 54 37 AM

## Testing instructions
- Go to the Stripe settings page and enable OCS
- Go to **WooCommerce > Settings > Payments** page
- Enable COD, Cheque, PayPal etc., a few non-Stripe gateways.
- Move Stripe to 1st position among the gateways.
- Refresh the page and confirm that there is no notice.
- Now move Stripe to the 3rd/4th position.
- Refresh the page and you should see the notice below the Stripe description
- Move Stripe to 1st position.
- Refresh the page and the notice should be gone (we have room for improvement here, the notice should be gone without refresh)
- Now move Stripe to the 3rd/4th position again.
- You should see the notice below the Stripe description
- Click the Move to top button and verify that Stripe is moved to the first position
- The relative order of the other payment gateways should be unchanged.

### Changelog entry

<details>

#### Comment 

The generic changelog added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5240 should cover this one as too.

</details>

**Post merge**

### DIFF
--- a/client/entrypoints/payment-gateways/index.js
+++ b/client/entrypoints/payment-gateways/index.js
@@ -66,7 +66,11 @@ function runWithObserver() {
 			observer.disconnect();
 		}
 	} );
-	observer.observe( document.body, { childList: true, subtree: true } );
+
+	const observerTarget = document.querySelector(
+		'.wc-settings-prevent-change-event'
+	);
+	observer.observe( observerTarget, { childList: true, subtree: true } );
 	window.setTimeout( () => observer.disconnect(), 15000 );
 }
 
--- a/client/entrypoints/payment-gateways/index.js
+++ b/client/entrypoints/payment-gateways/index.js
@@ -73,9 +73,9 @@ function runWithObserver() {
 // If the Optimized Checkout is enabled and the notice should be shown, run the notice injection.
 if (
 	// eslint-disable-next-line camelcase
-	wc_stripe_settings_params?.is_oc_enabled === '1' &&
+	!! wc_stripe_settings_params?.is_oc_enabled &&
 	// eslint-disable-next-line camelcase
-	wc_stripe_settings_params?.show_stripe_first_method_notice === '1'
+	!! wc_stripe_settings_params?.show_stripe_first_method_notice
 ) {
 	if ( document.readyState === 'loading' ) {
 		document.addEventListener( 'DOMContentLoaded', runWithObserver );
--- a/client/entrypoints/payment-gateways/index.js
+++ b/client/entrypoints/payment-gateways/index.js
@@ -24,7 +24,7 @@ function injectNotice() {
 	if (
 		! stripeElement ||
 		stripeElement.querySelector(
-			'.wc-stripe-optimized-checkout-first-method-notice'
+			'.wc-stripe-payment-gateways-oc-notice-wrapper'
 		) !== null
 	) {
 		return false;
--- a/client/entrypoints/payment-gateways/index.js
+++ b/client/entrypoints/payment-gateways/index.js
@@ -1,12 +1,80 @@
+/* global wc_stripe_settings_params */
 import React from 'react';
 import { createRoot } from 'react-dom/client';
+import OptimizedCheckoutFirstMethodNotice from '../../settings/advanced-settings-section/optimized-checkout-first-method-notice';
 import PaymentGatewaysConfirmation from './payment-gateways-confirmation';
 
 const paymentGatewaysContainer = document.getElementById(
 	'wc-stripe-payment-gateways-container'
 );
+
 if ( paymentGatewaysContainer ) {
 	createRoot( paymentGatewaysContainer ).render(
 		<PaymentGatewaysConfirmation />
 	);
+}
+
+/**
+ * Injects the Optimized Checkout first method notice into the Stripe payment methods list.
+ *
+ * @return {boolean} True if the notice was injected, false otherwise.
+ */
+function injectNotice() {
+	const StripeElement = document.querySelector( '#stripe' );
+	if (
+		! StripeElement ||
+		StripeElement.querySelector(
+			'.wc-stripe-optimized-checkout-first-method-notice'
+		) !== null
+	) {
+		return false;
+	}
+
+	const StripeDescription = StripeElement.querySelector(
+		'.woocommerce-list__item-text'
+	);
+
+	if ( ! StripeDescription ) {
+		return false;
+	}
+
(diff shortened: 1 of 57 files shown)


client/settings/advanced-settings-section/optimized-checkout-first-method-notice.js
wjrosa: I think it would be good to add JSDoc to all new code

Mayisha: Added.

includes/admin/class-wc-stripe-payment-gateways-controller.php
```diff
@@ -77,6 +77,15 @@ public function enqueue_payments_scripts() {
...
 		if ( $is_payment_methods_page ) {
+			$gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+			wp_localize_script(
+				'woocommerce_stripe_payment_gateways_page',
+				'wc_stripe_settings_params',
+				[
+					'show_stripe_first_method_notice' => WC_Stripe_Helper::should_show_stripe_first_method_notice(),
+					'is_oc_enabled'                   => $gateway->is_oc_enabled(),
+				]
+			);
```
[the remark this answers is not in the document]
Mayisha: Fixed in JS

wjrosa (approved):
Code is good, but it is not working correctly for me. I have OCS enabled, but I never see the notice below Stripe after putting it in the last position 🤔

wjrosa (changes requested):
Selected approve instead by mistake *

Mayisha:
> Code is good, but it is not working correctly for me. I have OCS enabled, but I never see the notice below Stripe after putting it in the last position 

@wjrosa is the option `wc_stripe_show_stripe_first_method_notice` set as `no` in your test site (may be from testing the other PR)?

includes/admin/class-wc-stripe-payment-gateways-controller.php
```diff
@@ -77,6 +77,15 @@ public function enqueue_payments_scripts() {
 		);
 
 		if ( $is_payment_methods_page ) {
+			$gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+			wp_localize_script(
+				'woocommerce_stripe_payment_gateways_page',
+				'wc_stripe_settings_params',
```
malithsen, line 83: We already have a [script with this exact name](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/06fe488b0246d75b2622cce7f92a3185100cbb29/includes/admin/class-wc-stripe-settings-controller.php#L268) which exposes params in the gateway settings page. This is not currently colliding because they are on different pages, but just to avoid confusion and make searches easy, can we rename this to something else, maybe: `wc_stripe_payment_gateways_page_params`

client/entrypoints/payment-gateways/index.js
```diff
@@ -1,12 +1,85 @@
...
+function runWithObserver() {
+	if ( injectNotice() ) {
+		return;
+	}
+	const observer = new MutationObserver( () => {
+		if ( injectNotice() ) {
+			observer.disconnect();
+		}
+	} );
+	observer.observe( document.body, { childList: true, subtree: true } );
```
malithsen: `document.body` seems excessive. Is there a more targeted container that we can observe?

malithsen (commented):
Thanks for the changes! This worked well the first time, but after moving Stripe to 3rd/4th position again, I no longer see the notice.

includes/admin/class-wc-stripe-payment-gateways-controller.php
Mayisha: It was actually intentional. In `OptimizedCheckoutFirstMethodNotice` component I am reading the flag from `wc_stripe_settings_params` but it was absent on this page. So I registered the same name script on the wc settings page with minimal (only the needed) params.

client/entrypoints/payment-gateways/index.js
Mayisha: Addressed in [d50e6f3](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5247/commits/d50e6f36dad91d8bc138067cdbcc8b316631ac40)

Mayisha:
> Thanks for the changes! This worked well the first time, but after moving Stripe to 3rd/4th position again, I no longer see the notice.

@malithsen Thanks for the feedback. I was also able to reproduce this after several moves and `move to top` action. 
When we move the gateway on settings page, `set_stripe_gateways_in_list` method was expected to be called. However, the hook was not working. I moved the hook in [d0e5af0](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5247/commits/d0e5af00588cbb89ce71f9388db97bf8d07750dc) and it seems to be working now.

cc: @wjrosa you might have faced the same issue.

client/entrypoints/payment-gateways/index.js
```diff
@@ -1,12 +1,89 @@
...
+		if ( injectNotice() ) {
+			observer.disconnect();
+		}
+	} );
+
+	const observerTarget = document.querySelector(
+		'.wc-settings-prevent-change-event'
+	);
+	observer.observe( observerTarget, { childList: true, subtree: true } );
+	window.setTimeout( () => observer.disconnect(), 15000 );
```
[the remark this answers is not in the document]
Mayisha: Fallback added [cc52a0b](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5247/commits/cc52a0badb9403e64e0b23b3ab3add88f08006ed)

wjrosa:
> > Code is good, but it is not working correctly for me. I have OCS enabled, but I never see the notice below Stripe after putting it in the last position
> 
> @wjrosa is the option `wc_stripe_show_stripe_first_method_notice` set as `no` in your test site (may be from testing the other PR)?

My issue was both. Updated the branch and deleted the option, and now it is working as expected 👍

wjrosa (approved):

malithsen (approved):
Thanks for the fixes. Looks good to me now.

PR #5166, opened 2026-03-23
Render Adaptive Pricing currency selector immediately above test instructions

wjrosa:
Fixes STRIPE-987
Fixes #5129

## Changes proposed in this Pull Request

The Adaptive Pricing currency selector was previously rendered via the `woocommerce_review_order_before_payment` hook, which placed it above the entire payment section (above the heading, saved payment methods list, etc.). This moves the rendering into `payment_fields()` directly, so the currency selector appears immediately above the test mode instructions and payment element where it belongs.

The old `attach_currency_selector_element()` method is deprecated in 10.6.0 and now calls `wc_deprecated_function()`.

Preview:
Screenshot 2026-04-02 at 16 43 45

## Testing instructions

**Prerequisites:** Stripe account with Adaptive Pricing enabled. Store running in test mode with Optimized Checkout Suite (OCS) enabled.

1. Go to a product page and add an item to the cart.
2. Proceed to the classic checkout (shortcode-based, not Blocks).
3. In the **Payment** section, confirm the currency selector (the Stripe currency dropdown) appears **immediately above** the test card instructions and the payment element — not above the payment method heading or saved cards list.
4. Select a different currency from the selector and confirm the payment element updates to reflect the new currency.
5. Complete a test purchase and confirm the order is placed successfully in the selected currency.
6. Repeat with a logged-in customer who has saved payment methods, and confirm the currency selector still appears below the saved methods list and above the payment element.

---

- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

### Changelog entry

* Fix - Render the Adaptive Pricing currency selector immediately above the payment element in classic checkout by moving it into payment_fields() instead of the woocommerce_review_order_before_payment hook

**Post merge**

- [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
- [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
- [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5248, opened 2026-04-08
Move test mode instructions above the Adaptive Pricing currency selector

wjrosa:
See: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5166#issuecomment-4202491223

## Summary

Addresses review feedback from [#5166 (comment)](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5166#issuecomment-4202491223).

- Moves the test copy (test mode instructions) to render **before** the Adaptive Pricing currency selector in `payment_fields()`
- New render order: description → test instructions → currency selector → payment element
- Adds a PHPUnit test asserting the test-copy-before-selector ordering when testmode and adaptive pricing are both active

## Test plan

- [ ] Enable Optimized Checkout Suite and Adaptive Pricing; set store to test mode
- [ ] Visit classic checkout — confirm test mode instructions appear **above** the currency selector, which in turn appears above the payment element
- [ ] Verify existing test `test_payment_fields_renders_currency_selector_conditionally` still passes
- [ ] Verify new test `test_payment_fields_renders_test_copy_before_currency_selector` passes

### Changelog entry

* Fix - Move test mode instructions above the Adaptive Pricing currency selector in classic checkout

cc @pierorocca 

### DIFF
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -998,9 +998,6 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 			$show_optimized_checkout = $this->oc_enabled && $this->is_valid_optimized_checkout_page();
 			$show_adaptive_pricing   = $show_optimized_checkout && $this->is_adaptive_pricing_supported();
 
-			if ( $show_adaptive_pricing ) {
-				echo '<div id="wc-stripe-currency-selector" class="wc-stripe-currency-selector" style="margin: 12px 0;"></div>';
-			}
 			// Output the form HTML.
 			?>
 			<?php if ( ! empty( $this->get_description() ) ) : ?>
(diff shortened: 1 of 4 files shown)


pierorocca:
Thanks @wjrosa! We don't want to break up the actual shopper experience in half. Having the instructions take up a single line followed by the experience is the right approach.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1053,6 +1050,10 @@ public function payment_fields() {
 					<?php
 				endif;
 			endif;
+
+			if ( $show_adaptive_pricing ) :
+				echo '<div id="wc-stripe-currency-selector" class="wc-stripe-currency-selector" style="margin: 12px 0;"></div>';
```
Mayisha:
```suggestion
				echo '<div id="wc-stripe-currency-selector" class="wc-stripe-currency-selector" style="margin-top: 12px;"></div>';
```

Nit: I think it would look better if we remove the bottom margin after the re-arrangement of the elements, as the gap below the currency selector looks bigger now.

**With bottom margin**
Image

Mayisha (approved):
Looks good to me. Left a small nit style suggestion. 💯

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
pierorocca: I'll elevate this to a required change as the gap looks broken.

malithsen (approved):
Looks good to me 👍

PR #5251, opened 2026-04-09
Defer checkout sessions webhooks

Mayisha:
Towards STRIPE-921

- If the checkout session events arrive before we set the checkout session ID in order meta in `process_payment_with_checkout_session` the event won't be processed correctly as `get_order_by_checkout_session_id` won't return the order. To avoid this scenario, added a safeguard here by deferring the webhook if order is not found.
- Process immediately if order is found.
- No logic changed in the main handler method.

## Testing instructions
- Enable OCS and adaptive pricing.
- As a shopper, checkout with adaptive pricing.
- Confirm that there is no regression.
- Hardcode `true` in [this if block](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/1a1fb400fa993b1f765257d071431b969abe69fa/includes/class-wc-stripe-webhook-handler.php#L1524) to intentionally send the webhook to deferred path.
- As a shopper, checkout with adaptive pricing.
- Check the order as an admin. It should be in a pending state.
- Wait for a while and check the order again.
- It should be in processing state with proper order notes.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1389,6 +1389,10 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 				case 'checkout.session.async_payment_succeeded':
 					$this->handle_checkout_session_success( $notification );
 					break;
+				case 'checkout.session.expired':
+				case 'checkout.session.async_payment_failed':
+					$this->handle_checkout_session_failure( $notification );
+					break;
 				default:
 					throw new Exception( "Unsupported webhook type: {$webhook_type}" );
 					break;
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1731,6 +1731,38 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 			return;
 		}
 
+		$session_id = $checkout_session->id;
+
+		// Look for an order. If order exists, process the webhook immediately.
+		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $checkout_session->id );
+
+		// If order does not exist, defer the webhook processing.
+		// This might happen if a webhook arrived before the order metadata was stored.
+		if ( ! $order instanceof \WC_Order ) {
+			WC_Stripe_Logger::debug( "Deferring processing of {$notification->type} ($session_id) asynchronously." );
+
+			$this->defer_webhook_processing(
+				$notification,
+				[
+					'session_id' => $session_id,
+				]
+			);
+			return;
+		}
+
+		// If order exists, process the webhook immediately.
+		$this->handle_checkout_session_failure( $notification );
+	}
+
+	/**
+	 * Handles a deferred checkout session failure event.
+	 *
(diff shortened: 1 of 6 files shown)


Mayisha:
@coderabbitai review

Mayisha:
@coderabbitai review this PR

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1500,16 +1508,49 @@ public function process_checkout_session( object $notification ): void {
 	 * sessions, as well as agentic checkout sessions.
 	 *
 	 * @param object $notification The notification from Stripe
-	 * @return void
+	 * @return bool True if the event was deferred for async processing, false if handled inline.
 	 */
-	public function process_checkout_session_success( object $notification ): void {
+	public function process_checkout_session_success( object $notification ): bool {
```
wjrosa, line 1513: It might be dangerous to set the `bool` return type here since this is a public method. I think it might be better to just remove the type instead

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1500,16 +1508,49 @@ public function process_checkout_session( object $notification ): void {
...
 		if ( ! isset( $checkout_session->id ) ) {
 			WC_Stripe_Logger::error( 'Checkout session ID is missing from the event data.' );
-			return;
+			return false;
+		}
+
+		$session_id = $checkout_session->id;
+
+		// Look for an order. If order exists, process the webhook immediately.
+		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $checkout_session->id );
```
```suggestion
		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $session_id );
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1687,14 +1728,47 @@ public function process_checkout_session_success( object $notification ): void {
 	 *
 	 * @param object $notification The Stripe notification containing the checkout session data.
 	 */
-	public function process_checkout_session_failure( object $notification ): void {
+	public function process_checkout_session_failure( object $notification ): bool {
```
wjrosa, line 1731: Same as my previous comment

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1687,14 +1728,47 @@ public function process_checkout_session_success( object $notification ): void {
...
 		if ( ! isset( $checkout_session->id ) ) {
 			WC_Stripe_Logger::debug( 'Checkout session ID is missing from the event data.' );
-			return;
+			return false;
+		}
+
+		$session_id = $checkout_session->id;
+
+		// Look for an order. If order exists, process the webhook immediately.
+		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $checkout_session->id );
```
```suggestion
		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $session_id );
```

wjrosa (approved):
Looks good to me 👍 . Left some suggestions

includes/class-wc-stripe-webhook-handler.php
Mayisha: @wjrosa this method was introduced in the same version. So it is not a breaking change.

diegocurbelo (approved):
LGTM and tests as described!

Issue #4007, opened 2025-03-04
Improve Link by Stripe display in Payment Methods to reduce user confusion


namiokuzono:
**Is your feature request related to a problem? Please describe.**

The current display of "Link by Stripe" payment method on the My Account > Payment method page shows "Expires: N/A". 

This is causing confusion among merchants' customers who may interpret it as an error or incomplete information. The problem is exacerbated because _Link by Stripe_ is a secure payment method storage system, much like other Express Payment method such as Apple Pay, and doesn't have a traditional expiration date like individual credit cards.

**Describe the solution you'd like**

I would like to propose improving the UI/UX by addressing the confusion around the "Expires: N/A" label to merchants' customers. Here are two potential solutions:

- **Retrieve and Display the Actual Expiration Date:** If the Stripe API allows, retrieve the expiration date of the underlying card used with Link by Stripe and display it instead of "N/A." This would align with how traditional credit card payment methods are displayed.
- **Replace "Expires: N/A" with a More User-Friendly Label:** If retrieving the expiration date is not possible, replace "Expires: N/A" with a clearer label such as "No Expiration" or "Stored Securely with Stripe Link." This would make it clear that the payment method does not have an expiration date and is securely stored.

**Other possibilities could also be:**
- Add a "Manage in Link" button that directs users to their Link Dashboard.
- Include a tooltip or information icon that explains the nature of Link by Stripe when users hover over or click it. 

Above improvement would provide clear information to merchants' customers, offer a way to manage their Link account, and educate them about this payment method's unique characteristics.

**Describe alternatives you've considered**

I am not sure if there are any workarounds to this as I am not able to test this myself.

**Additional context**

This issue was reported by a merchant who noticed customer confusion regarding the "Expires: N/A" display for Link by Stripe. 9475145-zd-a8c 

Internal: p1740962269124609-slack-C7U3Y3VMY

Implementing this change would improve the user experience for customers using Link by Stripe and reduce confusion about the validity or status of this payment method.

PR #5257, opened 2026-04-10
Fix STRIPE-329: Clear 'Expires' display for Link by Stripe payment tokens

stuartduff:
## Summary

- Link by Stripe is a wallet, not a card — it has no expiration date
- WooCommerce core sets `expires` to "N/A" for all non-credit-card payment tokens, which confuses customers who see it on My Account > Payment Methods
- This fix sets `$item['expires']` to an empty string in the Stripe plugin's `get_account_saved_payment_methods_list_item()` filter, so the Expires cell is blank instead of showing "N/A"

## Bug

[STRIPE-329: Improve Link by Stripe display in Payment Methods to reduce user confusion](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4007)

## Changes

- `includes/payment-tokens/class-wc-stripe-payment-tokens.php`: Added `$item['expires'] = ''` to the `LINK` case in `get_account_saved_payment_methods_list_item()`
- `tests/phpunit/PaymentTokens/WC_Stripe_Payment_Tokens_Test.php`: Added `test_link_token_clears_expires()` to assert the expires field is cleared for Link tokens

## Testing

- New PHPUnit test added: `test_link_token_clears_expires()` — simulates the WooCommerce-set "N/A" value and asserts it is overridden with an empty string
- No manual test environment set up; relying on automated tests + code review

## Notes


diegocurbelo (commented):
Works as expected.

But, I think that if we remove the "N/A" from Link, we should also remove it for the other payment methods as well, otherwise it might be confusing. cc @pierorocca 

Image

pierorocca:
Yeah keep it clean and remove the other N/As. The omission of info says the same thing as N/A. Good callout @diego.curbelo.

malithsen (commented):
Code looks good to me. Just expanding on Diego's comment. If the merchant has other gateways and the shopper has payment methods with no expiry saved from them, they can still see "N/A"s and empty expiry together, which can still cause confusion. I wonder if a more descriptive label like "Managed by Link" would help here.

daledupreez:
@stuartduff, I just want to follow up on this PR. Are you in a position to tackle the open requests above? Please let us know if you plan to pick this work up in the near future.

stuartduff:
Hey @daledupreez I have some bug blitz time tomorrow and I'll take a look for us then :)

stuartduff:
## Proposal: blank the Expires cell for all non-card Stripe tokens

Thanks @diegocurbelo and @malithsen; both points are fair, and I'd like to lock a decision in here before touching the code again.

### The proposal

Extend this PR from Link only to **every non-card Stripe payment token**, so the Expires cell renders empty instead of "N/A":

SEPA IBAN, Bacs Direct Debit, Cash App Pay, ACH, ACSS Debit, BECS Direct Debit, Link, Amazon Pay, Klarna.

Apple Pay and Google Pay are unaffected. They're stored as `CC` tokens with a genuine card expiry, so they keep showing real dates (the 12/28 in Diego's screenshot).

This is what @pierorocca called for above: *"keep it clean and remove the other N/As. The omission of info says the same thing as N/A."*

### Why this fully resolves the screenshot

Every row in Diego's screenshot is a Stripe token. So the exact inconsistency shown there, Link blank while Cash App Pay and SEPA IBAN still say "N/A", disappears completely with this change.

### On the cross-gateway point, correct, but core's to fix

`'expires' => 'N/A'` is hardcoded as the default for *every* payment token in WooCommerce core (`includes/wc-account-functions.php`), and is only overwritten for `cc` tokens. A gateway plugin can override its own tokens and nothing else. So if a shopper has a non-expiring token saved from a different gateway, that gateway's "N/A" will still sit alongside our blanks, and there's no way to reach it from this repo.

Worth noting that WooPayments has the same behaviour today — its token service filters the Link, SEPA and Amazon Pay list items but never clears `expires`, so those show "N/A" as well. That reinforces that the general fix is changing core's default, which I'd suggest raising as a separate core issue rather than blocking this PR on.

### Why not a descriptive label like "Managed by Link"

I dug into this option, and there are three concerns up for discussion; they lean towards leaving it for now:

1. The column header is **Expires**. "Managed by Link" isn't an expiry value, so it reads as a mismatch under that
heading.
2. It restates what the Method column already says on the same row: "Stripe Link (email@example.com)".
3. It needs bespoke translated copy for nine mely in a narrow column.

That said, the instinct behind it is right: the question a shopper actually has is "is this still valid, and can I do anything about it?" The Expires column just isn't where that gets answered. The better home for it is the **"Manage in Link"** action button suggested in the original issue description, which I'd keep as a follow-up enhancement rather than fold in here.

### What changes in the PR

- Replace the Link-specific `$item['expires'] = ''` with a single type guard covering any non-card Stripe token — including ones added in future, so this bug doesn't get reintroduced each time a new payment method lands.
- Extend the unit test to assert the blanking across a representative set of non-card types, plus a card token that keeps its expiry.

Does this work as the decision? A 👍 from @diegocurbelo, @malithsen and @pierorocca and I'll get the PR updated.

daledupreez:
@pierorocca, could you weigh in on this? I think the scope has become a lot bigger with the request to cover all Stripe payment methods, given that we may differ from the default for other payment gateways and the default in WooCommerce Core.

pierorocca:
For the existing shortcode based My Account > Payment Methods removing the N/A **for now** works for me even if other gateways might still end up with an N/A displayed. Having a mixed approach within a gateway is worse than the potential for gateways to show mixed N/A. Guarding future methods is reasonable. <br><br>At some point when My Account can be blockified, I expect we'd move away from the credit card centric table design to something more modern, brand forward, and payment method agnostic. e.g. a card component.

stuartduff:
Updated as proposed above, now that @pierorocca has confirmed the direction.

**What changed**

The Link-specific `$item['expires'] = ''` is gone. In its place, one type guard covers every non-card Stripe token:

```php
if ( ! $payment_token instanceof WC_Payment_Token_CC ) {
    $item['expires'] = '';
}
```

`WC_Payment_Token_CC` is the core class that defines `get_expiry_month()`/`get_expiry_year()`, and every non-card Stripe token — SEPA, Bacs Direct Debit, Cash App Pay, ACH, ACSS Debit, BECS Direct Debit, Link, Amazon Pay, Klarna — extends plain `WC_Payment_Token` instead. So there's no list of method IDs to maintain and future payment methods are covered automatically.

Apple Pay and Google Pay are stored as `WC_Stripe_Payment_Token_CC`, so they keep showing the underlying card's real expiry — the 12/28 in @diegocurbelo's screenshot is unaffected.

**On placement:** the guard sits after the branding `switch` rather than before it. Putting it earlier narrows `$payment_token` into a union type, which rewrites the message text of twelve already-baselined PHPStan errors in the switch below and forces unrelated baseline churn.

**Tests:** `test_link_token_clears_expires` is replaced with a `@dataProvider` test covering thirteen cases — the nine non-card types blanked, a plain card plus Apple Pay and Google Pay retaining `12/28`, and a non-Stripe token left at "N/A" (@malithsen's cross-gateway case). Reverting the source change fails exactly eight of them, so the coverage isn't vacuous.

Full PHPUnit, PHPCS and PHPStan run clean locally with no new failures. Also merged `develop` in, since the branch was well behind, and added the changelog and readme entries.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -693,6 +693,16 @@ public function get_account_saved_payment_methods_list_item( $item, $payment_tok
 				break;
 		}
 
+		// Only card tokens carry a real expiry date. WooCommerce core seeds every token's
+		// "Expires" cell with "N/A" and overwrites it for cards only, so wallets and bank
+		// debits read as missing data rather than "does not expire". Blank the cell for any
+		// non-card Stripe token instead — including ones added in future, which is why this
+		// is a type guard rather than another case in the switch above. Apple Pay and Google
+		// Pay are stored as card tokens, so they keep showing the underlying card's expiry.
```
daledupreez: I think we should simplify this comment.
```suggestion
		// Only card tokens have a real expiry date. WooCommerce defaults every token's
		// "Expires" value with "N/A", but only overwrites it for cards, so all other payment
		// methods have "N/A". We override the value to '' for all non-card tokens for consistency.
		// Note that Apple Pay and Google Pay are stored as card tokens, so they show an expiry date.
```

daledupreez (approved):
The changes look good to me! Thanks for sticking with this, @stuartduff! I'll get this merged shortly.

Screenshot 2026-08-27 at 10 14 13

Testing included most payment methods that we save, including iDEAL->SEPA and Bancontact->SEPA.

Issue #4184, opened 2025-04-02
WP Admin order Currency Display with countries that use $ sign


namiokuzono:
**Is your feature request related to a problem? Please describe.**

For when the site uses a different currency setting than the one for Stripe account, the WP Admin order area shows the Stripe account's currency in WP Admin order area. 

When the site uses currency such as Australian Dollars while Stripe account uses US dollars, the display can look confusing: 

![Image](https://github.com/user-attachments/assets/86645931-231b-4ad4-80ea-d4afa1332232)

**Describe the solution you'd like**

When the site currency differs from the one with Stripe account, either 1) Automatically displays the payout currency with letters such as "USD" 2) Add a setting to change the display from $ sign to letter based like USD.

**Describe alternatives you've considered**

No alternative. 

**Additional context**

Most of merchants likely will not have a different currency setting for the site from the one in Stripe account, however, this scenario could be possible when the merchant has a multi-site for each country with the same business (same Stripe account).

namiokuzono:
Out of curiosity, I tested also with WooPayments with this scenario, and it seems the possible payout amount is not a feature on WooPayments. For this, I've created another feature request here: https://github.com/Automattic/woocommerce-payments/issues/10667#issue-2965271002

PR #5259, opened 2026-04-10
Fix STRIPE-385: Append currency code when Stripe and order currencies differ

stuartduff:
## Summary

- When a store uses a different currency than the Stripe account (e.g. AUD store with a USD Stripe account), both display `$` in the WP Admin order totals — the Stripe Fee and Stripe Payout amounts are ambiguous
- When the Stripe balance transaction currency differs from the order currency, append the 3-letter code after the amount (e.g. `-$0.59 USD`)
- When currencies match, display is unchanged — no impact on the majority of merchants

## Bug

[STRIPE-385: WP Admin order Currency Display with countries that use $ sign](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4184)

## Changes

- `includes/payment-methods/class-wc-stripe-upe-payment-gateway.php`: In `display_order_fee()` and `display_order_payout()`, compute the formatted amount before the HTML block; when `$order->get_currency() !== $stripe_currency`, append `esc_html( strtoupper( $currency ) )` to the output
- `tests/phpunit/PaymentMethods/WC_Stripe_UPE_Display_Order_Fee_Test.php`: New test class covering all four scenarios (fee/payout × same/different currency)

## Testing

- New PHPUnit tests added covering same-currency (no code appended) and different-currency (code appended) for both `display_order_fee` and `display_order_payout`
- No manual test environment set up; relying on automated tests + code review
- To verify manually: set up a store in AUD with a USD Stripe account, place and complete an order, view in WP Admin — Stripe Fee and Stripe Payout rows should show `$X.XX USD`

## Notes


tests/phpunit/PaymentMethods/WC_Stripe_UPE_Display_Order_Fee_Test.php
```diff
@@ -0,0 +1,120 @@
+<?php
```
Mayisha: The test file name should be `class-wc-stripe-upe-display-order-fee-test.php`

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Display_Order_Fee_Test.php
```diff
@@ -0,0 +1,120 @@
...
+
+namespace WooCommerce\Stripe\Tests\PaymentMethods;
+
+use WC_Stripe_Order_Helper;
+use WC_Stripe_UPE_Payment_Gateway;
+use WooCommerce\Stripe\Tests\WC_Mock_Stripe_API_Unit_Test_Case;
+use WooCommerce\Stripe\Tests\Helpers\UPE_Test_Helper;
+use WooCommerce\Stripe\Tests\Helpers\WC_Helper_Order;
```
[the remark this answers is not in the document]
Mayisha: This is a valid concern, the tests fail with a fatal error.

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Display_Order_Fee_Test.php
Mayisha: Instead of adding a new test file, these tests should go in `tests/phpunit/class-wc-stripe-payment-gateway-test.php`

Mayisha (dismissed):
Functionally looking good to me. But the test file needs fixes as mentioned below and we should also fix the phpstan issues.

**Stripe account: USD, Store: AUD**

Image

**Stripe account: Euro, Store: AUD**
Image

daledupreez (commented):
The PHPStan warnings are also somewhat valid. It may be worth adding a guard/check after we call `$order = wc_get_order();` in these functions to make sure we have an order object:
```php
	if ( ! $order instanceof WC_Order ) {
		return <expected_return>;
	}
```

daledupreez:
@stuartduff, please let us know if you plan to keep working on this, or whether we should provide help with getting the changes wrapped up and merged.

tests/phpunit/PaymentMethods/WC_Stripe_UPE_Display_Order_Fee_Test.php
stuartduff: Thanks for the thorough review and the screenshots confirming the fix! I've addressed all the points in 2c2f8e27:

- PHPStan — added an if ( ! $order instanceof WC_Order ) { return; } guard after wc_get_order() in both display_order_fee() and display_order_payout(). This also let me drop three now-obsolete phpstan-baseline.neon entries, so npm run phpstan is green.
- Test imports / fatal error — you were right, the fabricated namespaces caused a fatal. Resolving this by relocating the tests (below) into the global-namespaced gateway test file, so the plain use imports resolve correctly.
- Test location — moved all four tests into tests/phpunit/class-wc-stripe-payment-gateway-test.php as you suggested, and deleted the standalone PaymentMethods/ file (which also makes the file-naming nit moot).
- Extra coverage — added test_display_order_fee_invalid_order_returns_early and the payout equivalent to lock in the new guard behaviour.

All six tests pass locally (npm run test:php -- --filter display_order → OK (6 tests, 6 assertions)), alongside clean PHPStan and PHPCS runs.

stuartduff:
> The PHPStan warnings are also somewhat valid. It may be worth adding a guard/check after we call `$order = wc_get_order();` in these functions to make sure we have an order object:
> 
> ```
> 	if ( ! $order instanceof WC_Order ) {
> 		return <expected_return>;
> 	}
> ```

@daledupreez Good catch — agreed the PHPStan warnings were valid. Both methods now dereference $order directly ($order->get_currency()), so I've added exactly the guard you suggested after wc_get_order() in display_order_fee() and display_order_payout():

```
if ( ! $order instanceof WC_Order ) {
    return;
}
```

Since the guard narrows $order to WC_Order, the three related phpstan-baseline.neon entries (get_stripe_fee/get_stripe_net/get_stripe_currency parameter-type ignores) were no longer matched, so I removed them too. npm run phpstan is clean, and I added invalid-order early-return tests to cover the new guards. Fixed in 2c2f8e27.

Hopefully this is okay; if not, I'll take you up on your offer of providing some help with getting the changes wrapped up and merged for this one. :)

daledupreez:
@stuartduff, that's looking good -- could you also add a changelog entry to `changelog.txt` and `readme.txt`?

stuartduff:
Hey @daledupreez, Great to hear, and thanks for your patience and help. Much appreciated!

I noticed you updated the milestone to 10.9.0, so I've added the entries referencing that version number.

daledupreez (approved):
Thanks for persisting with this, @stuartduff! This is looking great and I'll get this merged shortly.

I will create a small follow-up PR to use a data provider for the unit tests so they are a bit more compact, but I think we need the main feature merged as a starting point.

PR #5254, opened 2026-04-09
Fix payment method title on order received page

Mayisha:
This fixes a regression caused by https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5210 and https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5211.

## Changes proposed in this Pull Request:
- In https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5211 we removed the JS update of the OCS title on the classic checkout page and updated it in the backend. But that caused regression. In this PR, I am re-adding the JS override to set the label on classic checkout.
- Also fixes the wrong title for split PE

Screenshot 2026-04-10 at 11 52 50 AM

## Testing instructions
- Enable OCS and adaptive pricing.
- As a shopper, go to the classic checkout page.
- Verify the payment method label reads "Payment options" in classic checkout.
- Complete checkout with card.
- In `develop` notice that in the order received page and in the order notes the title is `Payment options`
- In this branch, the title should be `Stripe` on the order received page and in the order notes.

### Changelog entry

-   [x] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

The changelog for https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5210 should be enough as all 3 PRs are going in the same version.

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5261, opened 2026-04-10
Confirm checkout session with user data in classic checkout for guest user

Mayisha:
Related to STRIPE-999

In block checkout, we send the user data in `confirm` request if it was missing during initial checkout session creation. This part was missing for classic checkout. This PR generates the args for `confirm` request from checkout form fields if the corresponding data is missing in the created checkout session.

## Testing instructions
- Enable OCS and adaptive pricing.
- Open the store in a new incognito window (make sure your checkout page is not pre-filled with any data from previous checkout)
- As a shopper, go to the classic checkout page and try to pay with adaptive pricing.
- In `develop` the checkout will fail with an error message related to `email` being empty in `confirm`.
- In this branch, checkout should be successful.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/stripe-utils/utils.js
```diff
@@ -624,6 +624,136 @@ export const getDefaultValues = ( forCheckoutSession = false ) => {
 	};
 };
 
+/**
+ * Reads a value from a field by element id.
+ *
+ * @param {string} id HTML id attribute.
+ * @return {string} Field value, or empty string when missing.
+ */
+export const getFieldValue = ( id ) => {
```
wjrosa: Do we need to export all these functions?

client/classic/upe/__tests__/payment-processing.test.js
daledupreez: I am not sure why we need this `finally`, given that this is replicating the setup immediately before the `try`. And if we don't need the `finally`, we may not need the `try` either.

client/classic/upe/payment-processing.js
```diff
@@ -848,12 +852,14 @@ export const processPayment = (
 				}
 
 				const { actions } = loadActionsResult;
+				const session = await actions.getSession();
 
 				const shouldSavePaymentMethod = jQueryForm
```
daledupreez: Nn-blocking nit: If we're only using `shouldSavePaymentMethod` when the user is logged in, it may make sense to compute the value in the corresponding `if` block.

client/stripe-utils/utils.js
```diff
@@ -624,6 +624,136 @@ export const getDefaultValues = ( forCheckoutSession = false ) => {
...
+			},
+		};
+		result.billingAddress = billingAddress;
+	}
+
+	if ( ! currentSession?.shippingAddress ) {
+		const shipLine1 = getFieldValue( 'shipping_address_1' );
+		const shipCountryRaw = getFieldValue( 'shipping_country' );
+
+		if ( shipLine1 && shipCountryRaw ) {
```
daledupreez: Nit: We should add an inline comment to note why we are specifically checking these two values.

client/stripe-utils/utils.js
```diff
@@ -624,6 +624,136 @@ export const getDefaultValues = ( forCheckoutSession = false ) => {
...
+export const getFieldValue = ( id ) => {
+	const value = document.getElementById( id )?.value;
+	return typeof value === 'string' ? value.trim() : '';
+};
+
+/**
+ * Normalizes a WooCommerce country code for Stripe (uppercase ISO 3166-1 alpha-2 when applicable).
+ *
+ * @param {string} country Raw country from the form.
+ * @return {string|undefined} Normalized country or undefined when empty.
```
daledupreez: The JSDoc here seems wrong. It looks like we won't return `undefined`, but the function also doesn't enforce that a string is returned if `country` is some other type.
```suggestion
 * @return {string} If country is a two-letter string, return the uppercase value. If country is some other string, return that value. For falsy values of country, return ''.
```

client/blocks/checkout-sessions/hooks.js
daledupreez, line 169: The PR description and testing instructions don't mention this change. Should this be in this PR? If so, we should document it.

client/classic/upe/__tests__/payment-processing.test.js
daledupreez: As above, I am not sure why we need this `finally` -- if any other tests are expecting specific values for the return, they should be explicitly setting that up and not depending on upstream cleanup.

client/stripe-utils/utils.js
```diff
@@ -624,6 +624,136 @@ export const getDefaultValues = ( forCheckoutSession = false ) => {
...
+	}
+
+	if ( ! currentSession?.phoneNumber ) {
+		const billingOrShippingPhone =
+			getFieldValue( 'billing_phone' ) ||
+			getFieldValue( 'shipping_phone' );
+
+		if ( isLoggedIn && isPayerPhoneRequired && billingOrShippingPhone ) {
+			result.phoneNumber = billingOrShippingPhone;
+		}
```
daledupreez: Why are we only sending the phone number when the user is logged in? That's not an obvious condition to me. I would expect the phone number to be required regardless of logged-in state.

wjrosa (approved):
Code is good and works as expected 👍 

Probably unrelated issues:

1 - The checkout sessions payment container is not loaded in the classic checkout page if Link is enabled. I get the following error in console:

Image

I had to disable Link to test this PR.

2 - The payment method name here is wrong, but it shows something else:

Image

client/classic/upe/payment-processing.js
```diff
@@ -438,6 +438,10 @@ async function createStripePaymentElement( api, paymentMethodType ) {
 			paymentElementOptions
 		);
 	} else {
+		paymentElementOptions = {
+			...paymentElementOptions,
+			...getUpeSettings(),
+		};
```
daledupreez, line 449: This change is causing the following JS console error for me when loading classic checkout with Link and Adaptive Pricing enabled:
```
IntegrationError: Invalid createPaymentElement() parameter: options.terms.link is not an accepted parameter.
```

daledupreez (changes requested):
The happy path is working, but I am not sure about the changes affecting block checkout, and the error when Link is enabled is a blocker.

Mayisha:
Thank you @wjrosa for the feedback.

> 1 - The checkout sessions payment container is not loaded in the classic checkout page if Link is enabled. I get the following error in console:

Good catch. Fixed in [14e99dc](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5261/commits/14e99dc867adf5c1a78c3f5864a08201366211f5)

> 2 - The payment method name here is wrong, but it shows something else:

It is fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5254

client/classic/upe/payment-processing.js
Mayisha: Fixed in [14e99dc](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5261/commits/14e99dc867adf5c1a78c3f5864a08201366211f5)

client/stripe-utils/utils.js
Mayisha: This is currently aligned with the block checkout implementation. It is to avoid a complicated scenario which we can revisit in a follow-up.
Basically, in the backend, we create the checkout session with a phone number flag only for logged-in users so we can send the phone in frontend for a logged-in user. In case of a logged in user we are sure that there would be a phone number because the mismatch in backend flag vs the frontend key would result in failure. 
I will create a follow-up issue to see if we can effectively handle the scenario.

client/blocks/checkout-sessions/hooks.js
Mayisha: I don't see any changes in this file. This PR only targets the classic checkout 🤔

client/classic/upe/__tests__/payment-processing.test.js
Mayisha: Right. Fixed it.

wjrosa (approved):
Thanks for addressing the issue! Looks good to me 👍

client/stripe-utils/utils.js
Mayisha: Ignore my previous comment. The issue I mentioned was actually related to the required vs non-required status of the field. The payer phone is not blocked by login status. I have removed the check in 84e1da6e72eb64a992c663083ff350aa9e61b042

client/stripe-utils/utils.js
```diff
@@ -624,6 +624,137 @@ export const getDefaultValues = ( forCheckoutSession = false ) => {
...
+			if ( shipCity ) {
+				shippingAddress.address.city = shipCity;
+			}
+
+			const shipPostcode = getFieldValue( 'shipping_postcode' );
+			if ( shipPostcode ) {
+				shippingAddress.address.postal_code = shipPostcode;
+			}
+
+			result.shippingAddress = shippingAddress;
```
diegocurbelo, line 719: [Nit] Why not build the full object with `|| undefined` like billing does?

```suggestion
			const shippingAddress = {
				name: shipName,
				address: {
					country: shipCountry,
					line1: shipLine1,
					line2: getFieldValue( 'shipping_address_2' ) || undefined,
					state: getFieldValue( 'shipping_state' ) || undefined,
					city: getFieldValue( 'shipping_city' ) || undefined,
					postal_code: getFieldValue( 'shipping_postcode' ) || undefined,
				},
			};
			result.shippingAddress = shippingAddress;
```

diegocurbelo (approved):
The code looks good 👍🏼 , I left an optional suggestion for the shippingAddress build function.

It tests as described, and the Payment method is correctly shown in the order received page:

Image

client/stripe-utils/utils.js
Mayisha: Refactored.

PR #5270, opened 2026-04-13
EU compliance for adaptive pricing

Mayisha:
Fixes STRIPE-1018

0fiyiUe18lGU6kGyQNmXxE-fi-2702_208852

- Added ECB interbank fee to the notice on the order received page and email the EU shoppers.
- Added a `AdaptivePricingDisclosure` component to show below the currency selector.

**Note**: I will replace `X%` with actual value before merging.

## Testing instructions
- Enable OCS and adaptive pricing
- As a shopper, go to the checkout page and use a billing address from any EEA country.
- Verify that the ECB rate is displayed below the currency selector.

Screenshot 2026-04-13 at 5 08 26 PM

- Complete checkout with card in local currency. 
- On the order received page, confirm that the adaptive pricing notice contains the ECB rate.
- Check the email sent to the shopper, confirm that the adaptive pricing notice contains the ECB rate.
- Check and confirm that the ECB rate is not displayed below the currency selector and in the adaptive pricing notice on the order received page and email for US/non-EU billing addresses.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


changelog.txt
```diff
@@ -66,6 +66,8 @@ xxxx-xx-xx - version 10.6.0
 * Add - Allow customers to save payment methods during checkout with adaptive pricing
 * Fix - Restrict Checkout Session saved payment method options to logged-in customers so guest checkout session creation succeeds
 * Add - Add an admin notice and one-click action to move Stripe payment methods to the top of WooCommerce payment gateway order for Optimized Checkout
+* Update - Allow Adaptive Pricing for merchant accounts based in EEA countries
```
wjrosa, line 70: Is the inclusion of two changelog entries intentional?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1170,15 +1173,26 @@ public function add_converted_currency_information( string $formatted_total, $or
...
 	 * Shows a notice to the order received page to inform the customer about the currency conversion
 	 * when the order is paid with a different currency than the store currency.
 	 *
-	 * @param WC_Abstract_Order $order The order object.
+	 * @param WC_Order $order The order object.
 	 *
 	 * @return void
 	 */
-	public function add_currency_conversion_notice( WC_Abstract_Order $order ): void {
+	public function add_currency_conversion_notice( WC_Order $order ): void {
```
wjrosa: Is this change intentional? I set this to `WC_Abstract_Order` due to a request from Dale, as he mentioned this can be a refund order as well

client/classic/upe/payment-processing.js
```diff
@@ -482,6 +487,67 @@ function mountCurrencySelectorElement( elements ) {
...
+function mountAdaptivePricingDisclosure() {
+	const container = document.getElementById(
+		'wc-stripe-adaptive-pricing-disclosure'
+	);
+	if ( ! container ) {
+		return;
+	}
+
+	const isAdaptivePricingEnabled =
+		container.getAttribute( 'isAdaptivePricingEnabled' ) === 'true';
```
wjrosa, line 559: Not sure if that's the best approach to persist the information. Maybe using session storage would make more sense?

wjrosa (approved):
Code is good and works as expected. Left some suggestions/questions

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Thanks for calling this out. I thought it was unintentional as I was getting phpstan error in a downstream method. Let me check and handle this.

changelog.txt
Mayisha: Yes, I intentionally added two dedicated entries.

client/classic/upe/payment-processing.js
Mayisha: I am also not sure what would be best here. But the intention is to set the flag only after the currency selector mounts and this should not persist on page refresh or re-mounts as the currency selectors availability might change during those events.

wjrosa: Hum, maybe we could check for the `ready` in the currency selector element?

client/classic/upe/payment-processing.js
```diff
@@ -33,6 +36,7 @@ import {
 } from 'wcstripe/stripe-utils/constants';
 import { handleDisplayOfPaymentInstructions } from 'wcstripe/optimized-checkout/handle-display-of-payment-instructions';
 import { handleDisplayOfSavingCheckbox } from 'wcstripe/optimized-checkout/handle-display-of-saving-checkbox';
+import { AdaptivePricingDisclosure } from 'wcstripe/blocks/checkout-sessions/adaptive-pricing-disclosure';
```
diegocurbelo: We should move `adaptive-pricing-disclosure`  to a shared location and avoid importing it from `wcstripe/blocks*`

client/blocks/checkout-sessions/styles.scss
```diff
@@ -0,0 +1,30 @@
...
+		color: inherit;
+		text-decoration: underline;
+		cursor: pointer;
+		vertical-align: baseline;
+	}
+
+	.wc-stripe-adaptive-pricing-disclosure__toggle:hover,
+	.wc-stripe-adaptive-pricing-disclosure__toggle:focus {
+		color: #1d2327;
+	}
```
diegocurbelo: [Nit] For consistency with `.wc-stripe-adaptive-pricing-disclosure__summary` this can be moved to inseude `.wc-stripe-adaptive-pricing-disclosure__toggle`:
```
&:hover,
&:focus {
    ...
}
```

diegocurbelo (commented):
The code changes look good to me in general, just one small detail:
- I think the `AdaptivePricingDisclosure` component should be in a shared location, not inside the blocks folder.

-- 
> Note: I will replace X% with actual value before merging.

Is this something that can change frequently? If so, we can store it in the plugin settings (DB) so it's easier to update in the future (either via Stripe API or by a field in the plugin settings page)

client/classic/upe/payment-processing.js
Mayisha: > maybe we could check for the ready in the currency selector element?

I’m not sure I fully understand how to keep the flag by checking for `ready` on the currency selector. Could you clarify what you have in mind? 🤔

Mayisha:
Thanks for the review @diegocurbelo

>  think the AdaptivePricingDisclosure component should be in a shared location, not inside the blocks folder.

I totally forgot to move that to a shared location 🤦‍♀️ 

> Is this something that can change frequently? If so, we can store it in the plugin settings (DB) so it's easier to update in the future (either via Stripe API or by a field in the plugin settings page)

My understanding is it will be one fixed value.

diegocurbelo (approved):
Thanks for the clarification @Mayisha 👍🏼

PR #4914, opened 2026-01-05
Fix warning for Amazon Pay in Express Checkout block

daledupreez:
Fixes STRIPE-876
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4913

## Changes proposed in this Pull Request:

This PR ensures that we don't show the Amazon Pay gateway as enabled when in an admin context, as we want to bypass warnings that appear in the Express Checkout block. This re-uses and refactors the existing code that does the same thing for Link.

## Testing instructions

* Check out `develop` and run a build from that branch
* Enable the Amazon Pay feature flag: `wp option update _wcstripe_feature_amazon_pay yes`
* Connect to Stripe
* Ensure that Amazon Pay is enabled in the Stripe settings
* Navigate to the _Pages_ section in the left nav
* Open the Cart or Checkout page from the page
* If the page is still using the classic/shortcode content, switch to blocks
* Click on the Express Checkout block
* Confirm that you see a warning about **Amazon Pay** not supporting the current block in the inspector
* Navigate to the Cart (or Checkout) page, and confirm you also see the warning for the Express Checkout block in the other context
* Now check out this branch and run a build from this branch
* Repeat the tests, and confirm that you no longer see a warning for Amazon Pay

| Before | After |
|--------|-----|
| Screenshot 2026-01-05 at 20 05 45  | Screenshot 2026-01-05 at 20 05 06 |

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [N/A] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5279, opened 2026-04-14
Fix: Restore refund via gateway for Amazon Pay orders

maxlaf:
Fixes STRIPE-1081

`WC_Stripe::add_gateways()` filtered `WC_Stripe_UPE_Payment_Method_Amazon_Pay` out of the WooCommerce gateway registry whenever `is_admin()` was `true`. Because the WooCommerce order edit page is an admin page, the `stripe_amazon_pay` gateway was absent from the registry there, causing two issues:

1. **"Refund via Gateway" button missing** — WooCommerce checks `$payment_gateways[$order->payment_method]->supports('refunds')` to decide whether to show the button. With no gateway registered, the check fails and only the manual refund option is shown.
2. **Stripe dashboard transaction link missing** — `get_transaction_url()` is called on the registered gateway instance; with none present the link cannot be generated.

Stripe Link is unaffected because Link orders are stored with `payment_method = 'stripe'` (the main gateway), not `stripe_link`.

**Root cause introduced in:** PR #4914 / version 10.3.0 (STRIPE-876) — the `is_admin()` guard was added to suppress block-editor warnings, but it was too broad.

**Fix:** Narrow the condition to `is_admin() && ! $this->is_order_management_context()`. A new private helper `is_order_management_context()` returns `true` on the legacy CPT order screen (`shop_order`), the HPOS order screen (`woocommerce_page_wc-orders`), and during the `woocommerce_refund_line_items` AJAX action, so Amazon Pay (and all other gateways) remain registered where refund processing requires them.

## Testing instructions

**Manual:**
1. Enable Amazon Pay in Stripe settings and place a test order using Amazon Pay.
2. Open the order in WooCommerce admin.
3. Confirm the **"Refund via Stripe"** button is visible in the order totals box.
4. Confirm the **transaction ID at the top of the order** is a clickable link to the Stripe dashboard.
5. Process a partial refund and verify it succeeds via the gateway.

**Automated:**
```
cd /path/to/woocommerce-gateway-stripe
composer install
./vendor/bin/phpunit tests/phpunit/class-wc-stripe-test.php --filter test_add_gateways
```
Two new cases cover the regression: `amazon pay included on legacy CPT order edit page` and `amazon pay included on HPOS order edit page`.

- [ ] I have created a changelog entry for this change


maxlaf:
Closing to reopen with a clean single commit.

PR #5284, opened 2026-04-14
Add ECE support for subscription change payment method page

### LINKED ISSUE #4792: ECE support for subscriptions' "Change payment" page
### Description

Originally reported here: p1762877300408759-slack-C055WHLA98D

WC Stripe allows to pay for subscription products with ECE (Amazon Pay/Google Pay).
The payment method generated by the ECE is normally saved in the customer's profile.

However, if the customer wants to change the payment method for one of their subscriptions, they can't choose any ECE method, because no ECE buttons are displayed for them.

### Steps to reproduce

1. As a merchant, set up a simple subscription product
2. Enable one of the ECE methods (Amazon Pay/Google Pay) in the WC Stripe settings
3. As a customer, purchase the subscription product with Amazon Pay or Google Pay
4. As a customer, navigate to My Account > Subscriptions
5. Select one of the subscriptions
6. Click "Change payment"
7. On the "Change payment" page, no ECE methods are displayed

### Expected behavior

If the customer can purchase a subscription product with an ECE method, they should also be able to change the payment method as another ECE method

### Screenshots

ECE Available on subscription product:

Image

But not available on the "Change payment method" page:
Image

wjrosa:
Fixes #4792

- Enables Express Checkout Element buttons (Apple Pay, Google Pay, Link) on the WooCommerce Subscriptions "Change payment method" page. Previously, ECE buttons were explicitly blocked on this page, forcing customers to use the standard payment form even when they originally paid with an ECE method.
- Uses Stripe Elements in `setup` mode (no payment amount) for the change-payment-method context, creating a SetupIntent instead of a PaymentIntent.
- The payment method is created client-side and submitted through the existing checkout form, reusing the server-side `process_change_subscription_payment_with_deferred_intent()` flow.
- Persists the express checkout type to the subscription so the payment method title reflects the actual express method used (e.g. "Apple Pay (Stripe)" / "Google Pay (Stripe)") on both the success path and the 3DS-redirect path.

### How it works

1. **PHP** — `WC_Stripe_Express_Checkout_Element::init()` no longer early-returns for `change_payment_method`. It registers a minimal set of hooks (scripts + checkout button rendering + title persistence).
2. **PHP** — A new `WC_Stripe_Express_Checkout_Helper::is_change_payment_method_page()` detects the context, with a dedicated visibility check that bypasses cart-dependent validations.
3. **JS** — `startExpressCheckout()` creates Stripe Elements with `mode: 'setup'` and `paymentMethodCreation: 'manual'`. (Amazon Pay is auto-hidden by the Stripe SDK in this mode and is therefore not surfaced on the change-payment-method page.)
4. **JS** — On confirm, `handleChangePaymentMethodFlow()` creates a Stripe PaymentMethod, populates the checkout form's hidden fields (including `express_checkout_type`), and submits the form.
5. **Server** — The existing subscription change payment flow processes the SetupIntent and updates the subscription. `update_subscription_payment_method_title` (hooked on `wc_stripe_change_subs_payment_method_success`) sets the express label on the success path; for 3DS, the type is persisted to subscription meta before redirect and re-applied via a new `wc_stripe_after_set_payment_method_title_for_confirmed_intent` action after `process_order_for_confirmed_intent()` runs.

## Testing instructions

1. Set up a subscription product and enable Apple Pay/Google Pay in Stripe settings.
2. Purchase a subscription with Apple Pay or Google Pay.
3. Navigate to **My Account → Subscriptions → select subscription → Change Payment**.
4. Verify ECE buttons (Apple Pay/Google Pay) appear on the change-payment-method page.
5. Complete the payment method change with Apple Pay/Google Pay and verify the subscription's payment method is updated successfully.
6. Verify the subscription's payment method title in the admin reflects the express method used (e.g. "Apple Pay (Stripe)" / "Google Pay (Stripe)"), not just "Stripe".
7. Enable Link in Stripe settings and verify the Link express checkout button also appears on the change-payment-method page; complete a method change with Link and confirm the subscription updates successfully.
8. Verify regular checkout ECE still works (product page, cart, checkout).
9. Verify pay-for-order ECE still works.
10. Verify ECE buttons do **not** appear on the switch-subscription page.

---

- [x] Covered with tests

### Changelog entry

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -629,7 +629,52 @@ public function get_product() {
...
+	private function should_show_ece_on_change_payment_method_page() {
+		if ( ! WC_Stripe::get_instance()->connect->is_connected() ) {
+			return false;
+		}
+
+		if ( ! $this->testmode && ! is_ssl() ) {
+			return false;
+		}
+
+		$available_gateways = WC()->payment_gateways->get_available_payment_gateways();
```
malithsen:
```suggestion
		$available_gateways = WC()->payment_gateways()->get_available_payment_gateways();
```

Calling the method, instead of the propery should avoid the phpstan-baseline bump

client/express-checkout/payment-flow.js
```diff
@@ -251,3 +251,86 @@ export const handleConfirmationTokenFlow = async ( {
...
+			expressTypeInput.val( event.expressPaymentType ?? '' );
+		} else {
+			form.append(
+				jQuery( '<input>' )
+					.attr( 'type', 'hidden' )
+					.attr( 'name', 'wc-stripe-express-checkout-type' )
+					.val( event.expressPaymentType ?? '' )
+			);
+		}
+
```
malithsen, line 323: Do we capture `wc-stripe-express-checkout-type` somewhere in the backend? I can't seem to find it.

client/entrypoints/express-checkout/index.js
```diff
@@ -339,15 +339,30 @@ jQuery( function ( $ ) {
...
-					expressPaymentType,
-					hasFreeTrial
-				) && {
+				...( ( isChangePaymentMethod ||
+					isManualPaymentMethodCreation(
+						expressPaymentType,
+						hasFreeTrial
+					) ) && {
 					paymentMethodCreation: 'manual',
 				} ),
```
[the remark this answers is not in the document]
wjrosa: Good catch. Fixed by passing `isChangePaymentMethod` through `isManualPaymentMethodCreation()` instead of bypassing it with OR logic. Now `isManualPaymentMethodCreation(expressPaymentType, isChangePaymentMethod || hasFreeTrial)` correctly excludes Amazon Pay from manual creation in setup mode, since the utility already handles that exclusion.

client/express-checkout/payment-flow.js
```diff
@@ -251,3 +251,86 @@ export const handleConfirmationTokenFlow = async ( {
...
+		if ( paymentGatewayInput.length ) {
+			paymentGatewayInput.prop( 'checked', true );
+		}
+
+		// Submit the form — the server-side handler processes the SetupIntent.
+		form.trigger( 'submit' );
+	} catch ( e ) {
+		return handlePaymentFlowException( event, e, abortPayment );
+	}
+};
```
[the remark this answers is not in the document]
wjrosa: Agreed — will add test coverage for the change-payment-method flow in a follow-up. The hidden-field/form-submission path and the non-regression check for regular ECE checkout are good test targets.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -629,7 +629,52 @@ public function get_product() {
...
+	/**
+	 * Returns true if the current page is a subscription change payment method page.
+	 *
+	 * @return boolean
+	 */
+	public function is_change_payment_method_page() {
+		return isset( $_GET['change_payment_method'] ) // phpcs:ignore WordPress.Security.NonceVerification
+			&& function_exists( 'wcs_is_subscription' )
+			&& wcs_is_subscription( absint( $_GET['change_payment_method'] ) ); // phpcs:ignore WordPress.Security.NonceVerification
+	}
```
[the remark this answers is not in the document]
wjrosa: The detection already validates the subscription ID via `wcs_is_subscription( absint( $_GET['change_payment_method'] ) )`, which confirms the value is a real subscription. This can't be triggered on unrelated pages by simply carrying the query arg — it would need a valid subscription ID. I think the current check is sufficient.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Good call — fixed. Using `WC()->payment_gateways()` method call now, and updated the phpstan-baseline count accordingly.

client/express-checkout/payment-flow.js
wjrosa: Good question. The existing backend handlers read `$_POST['express_checkout_type']` (with underscores), but this hidden field uses `wc-stripe-express-checkout-type` (with hyphens). For the change-payment-method flow, the form submission goes through WooCommerce Subscriptions' standard handler, which calls the gateway's `process_change_subscription_payment_method()`. The express checkout type isn't strictly needed for that flow since it's only used for order meta (payment method title like "Apple Pay" vs "Google Pay"), but I should either rename the field to match the existing backend convention or add a backend handler to read it. I'll address this.

client/express-checkout/payment-flow.js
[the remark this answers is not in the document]
wjrosa: yes, please

client/express-checkout/payment-flow.js
wjrosa: Fixed — renamed the hidden field from `wc-stripe-express-checkout-type` to `express_checkout_type` so it matches what the backend reads in `$_POST['express_checkout_type']` at `WC_Stripe_Express_Checkout_Element::update_payment_method_title_for_express_checkout()`.

changelog.txt
```diff
@@ -1,8 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.6.0
-* Fix - Accept regional language names for Spanish provinces (e.g., Basque "Gipuzkoa") in Apple Pay and express checkout address validation
-* Fix - Restore missing saved payment tokens when Optimized Checkout Suite is enabled
```
daledupreez, line 7: Were these entries removed intentionally?

readme.txt
```diff
@@ -146,8 +146,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.6.0 - xxxx-xx-xx =
-* Fix - Accept regional language names for Spanish provinces (e.g., Basque "Gipuzkoa") in Apple Pay and express checkout address validation
-* Fix - Restore missing saved payment tokens when Optimized Checkout Suite is enabled
+* Add - Allow shoppers to change a subscription payment method using Express Checkout (Apple Pay, Google Pay, Amazon Pay)
```
daledupreez: Question: is Link supposed to work? It's not clear from this context that it is, nor does the PR specify why it's not expected to work.

daledupreez (commented):
I have a high-level concern: should merchants not be able to choose whether express checkout methods are visible on the change payment method page? In my mind, it is a distinct context from checkout, even if it is using the same underlying primitives.

client/express-checkout/payment-flow.js
```diff
@@ -251,3 +251,86 @@ export const handleConfirmationTokenFlow = async ( {
...
```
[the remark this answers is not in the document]
wjrosa: Fixed in 7297ca38b. Added `import jQuery from 'jquery';` at the top of `client/express-checkout/payment-flow.js` so the three `jQuery(...)` calls in `handleChangePaymentMethodFlow` resolve at runtime. Matched the same import pattern used by the sibling modules (`utils/index.js`, `compatibility/wc-product-page.js`).

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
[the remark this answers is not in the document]
wjrosa: Fixed in 7297ca38b. Both raw `isset( $_GET['change_payment_method'] )` checks now delegate to `$this->express_checkout_helper->is_change_payment_method_page()`:

- **Line 77** (hook registration): the minimal-hook-set branch now only activates when the subscription ID resolves to a real subscription via `wcs_is_subscription()`, matching the helper contract.
- **Line 241** (localized JS flag `is_change_payment_method`): uses the same helper so the JS and PHP gating can't diverge on an invalid subscription ID.

Keeps the helper as the single source of truth for detecting this context.

changelog.txt
wjrosa: Good catch, that was unintentional — the rebase on top of develop accidentally dropped those four entries. Restored in 7297ca38b (paratest, Spanish provinces Apple Pay fix, saved tokens OCS fix, and the Checkout Sessions redirect flow fix).

readme.txt
wjrosa: Confirmed Link is supported — updated in 92d8efaef. No code change was needed:

- `client/entrypoints/express-checkout/index.js` already includes Link in the `expressPaymentTypes` list (gated by `isLinkEnabled`), so the Link button renders on the change-payment page when the merchant has Link enabled.
- `isManualPaymentMethodCreation('link', …)` returns `true` (only Amazon Pay is excluded from `paymentMethodCreation: 'manual'`), so the setup-mode flow uses `stripe.createPaymentMethod({ elements })` for Link the same way it does for card-backed types. Per [Stripe's docs](https://docs.stripe.com/js/elements_object/create.md), only Amazon Pay and Klarna are excluded from that combination.
- `handleChangePaymentMethodFlow()` is payment-method-type-agnostic — it takes the returned `paymentMethod.id` and feeds it through the existing checkout form, and the server-side `process_change_subscription_payment_with_deferred_intent()` path handles Link PaymentMethods the same as any other.

Updated:
- `changelog.txt` and `readme.txt` entries now read "Apple Pay, Google Pay, Amazon Pay, Link".
- PR description updated to include Link in the scope and added a Link-specific test-plan item.

Test plan for Link is covered by the existing manual-verification checklist — enable Link in Stripe settings, navigate to a subscription's Change Payment page, and confirm the Link button renders and the change goes through.

wjrosa:
@daledupreez Good call — wiring this up as a dedicated merchant toggle rather than piggybacking on the checkout-page setting.

Tracked in #5312 and implemented in #5310, which targets this branch so it can land either stacked or rebased onto `develop` after this PR merges. Summary:

- Adds `change_payment_method` as a multiselect option for both `express_checkout_button_locations` (Apple Pay / Google Pay / Link) and `amazon_pay_button_locations`, with a matching "Change payment method (subscriptions)" checkbox in both settings UIs.
- Decouples `should_show_ece_on_change_payment_method_page()` from `should_show_ece_on_checkout_page()` and exposes a `wc_stripe_show_express_checkout_on_change_payment_method` filter for programmatic overrides.
- Includes the new location in the default list so the out-of-the-box behavior matches what this PR ships; merchants can opt out of the change-payment context without touching the checkout toggle.

Happy to collapse the two PRs into one if you'd prefer the toggle ships with the feature itself — let me know.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -73,13 +73,16 @@ public function init() {
...
 
-		// Don't load for change payment method page.
-		if ( isset( $_GET['change_payment_method'] ) ) {
+		// For change payment method page, only load the minimal set of hooks needed.
+		if ( $this->express_checkout_helper->is_change_payment_method_page() ) {
+			add_action( 'wp_enqueue_scripts', [ $this, 'scripts' ] );
+			add_action( 'woocommerce_checkout_before_customer_details', [ $this, 'display_express_checkout_button_html' ], 1 );
+			add_filter( 'woocommerce_gateway_title', [ $this, 'filter_gateway_title' ], 10, 2 );
 			return;
 		}
```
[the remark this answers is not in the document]
wjrosa: Addressed in 7ae613a0.

The hook structure is now: `update_subscription_payment_method_title` runs on `wc_stripe_change_subs_payment_method_success` for the success path (reads `$_POST['express_checkout_type']`), and `maybe_apply_express_title_after_confirmed_intent` runs on a new `wc_stripe_after_set_payment_method_title_for_confirmed_intent` action for the 3DS-redirect path. For the redirect path, the type is persisted to `_wc_stripe_express_checkout_type` subscription meta before redirect (in `process_change_subscription_payment_with_deferred_intent`) and read back during AJAX confirmation; the meta is cleaned up after applying. PHPUnit coverage added for both paths in `class-wc-stripe-express-checkout-element-test.php`.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.7.0
+* Add - Allow shoppers to change a subscription payment method using Express Checkout (Apple Pay, Google Pay, Amazon Pay, Link)
```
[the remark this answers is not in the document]
wjrosa: Addressed in 564a50f6 — dropped Amazon Pay from both `changelog.txt` and `readme.txt` change-payment entries.

client/express-checkout/payment-flow.js
```diff
@@ -251,3 +252,86 @@ export const handleConfirmationTokenFlow = async ( {
...
+		// Ensure the payment method is set to stripe.
+		const paymentGatewayInput = form.find(
+			'input[name="payment_method"][value="stripe"]'
+		);
+		if ( paymentGatewayInput.length ) {
+			paymentGatewayInput.prop( 'checked', true );
+		}
+
+		// Submit the form — the server-side handler processes the SetupIntent.
+		form.trigger( 'submit' );
```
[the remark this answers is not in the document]
wjrosa: Addressed in 7ae613a0. Hooked `update_subscription_payment_method_title` on `wc_stripe_change_subs_payment_method_success` — it reads `$_POST['express_checkout_type']` and applies the express label ("Apple Pay (Stripe)" / "Google Pay (Stripe)") via `set_payment_method_title()`. Also covered the 3DS-redirect path: the type is persisted to `_wc_stripe_express_checkout_type` subscription meta before redirect and re-applied via a new `wc_stripe_after_set_payment_method_title_for_confirmed_intent` action after the AJAX confirmation runs through `process_order_for_confirmed_intent()`. Tests added in `class-wc-stripe-express-checkout-element-test.php`.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -639,6 +684,11 @@ public function is_page_supported() {
 	 * @return  boolean  True if express checkout elements are supported on current page, false otherwise
 	 */
 	public function should_show_express_checkout_button() {
+		// For subscription change payment method, only check basic requirements.
+		if ( $this->is_change_payment_method_page() ) {
+			return $this->should_show_ece_on_change_payment_method_page();
```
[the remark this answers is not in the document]
wjrosa: Addressed in 564a50f6. Added two `@dataProvider`-driven tests in `tests/phpunit/payment-methods/class-wc-stripe-express-checkout-helper-test.php`:

1. `test_is_change_payment_method_page` — covers query arg missing, `wcs_is_subscription()` true, and `wcs_is_subscription()` false.
2. `test_should_show_express_checkout_button_change_payment` — exercises the change-payment dispatch with all checks passing, plus per-condition negatives: not connected, `stripe` gateway missing, ECE disabled, and `should_show_ece_on_checkout_page()` false.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -490,6 +502,86 @@ public function add_order_meta( $order_id, $posted_data ) {
...
+
+		$express_checkout_type = $order->get_meta( '_wc_stripe_express_checkout_type' );
+		if ( empty( $express_checkout_type ) ) {
+			return;
+		}
+
+		if ( $this->apply_express_checkout_title_to_order( $order, $express_checkout_type ) ) {
+			$order->delete_meta_data( '_wc_stripe_express_checkout_type' );
+			$order->save();
+		}
```
daledupreez: If there was junk in the `_wc_stripe_express_checkout_type` meta field, `$this->apply_express_checkout_title_to_order()` will return false, and we won't clean up. Is that what we want? Or should we unconditionally remove the meta field, regardless of whether it was handled correctly?

daledupreez (commented):
@wjrosa, I am not seeing the express checkout options showing up when trying to change the payment method for subscriptions initially purchased via a variety of different payment methods, including Google Pay. Can you confirm this is working as expected and/or share a screenshot of what it is supposed to look like?

client/entrypoints/express-checkout/index.js
```diff
@@ -339,14 +339,28 @@ jQuery( function ( $ ) {
...
+					amount: options.total,
+				} ),
 				currency: options.currency,
 				...( isManualPaymentMethodCreation(
 					expressPaymentType,
-					hasFreeTrial
+					isChangePaymentMethod || hasFreeTrial
 				) && {
 					paymentMethodCreation: 'manual',
 				} ),
```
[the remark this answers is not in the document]
wjrosa: Acknowledged — Jest test for the setup-mode `elements()` config and an E2E for the change-payment flow are valuable, but out of scope for this PR. Tracking as a follow-up; resolving the thread.

client/express-checkout/payment-flow.js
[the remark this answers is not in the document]
wjrosa: Acknowledged — the hidden-field/form-submission contract is real, but adding Jest + E2E coverage is bigger than this PR's scope. Tracking as a follow-up; resolving the thread.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
wjrosa: Yes, that's the right call — handled in 73feccc06. The marker is now cleared unconditionally so junk meta, an unmappable type, or an aborted/cancelled redirect can't leak a stale title into a later confirmed-intent update.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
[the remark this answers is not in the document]
wjrosa: Tried this in 73feccc06 and reverted in 57d5471d8 — adding `is_account_page()` broke ECE rendering on the actual change-payment screen. WC Subscriptions exposes that page as the My Account `view-subscription` endpoint, and `is_page(myaccount)` returns false on endpoints in many theme setups. The "unrelated page activates setup-mode ECE" risk is theoretical: without a change-payment form on the page the rendered button has nothing to submit to and no-ops, so the cure is worse than the disease here.

wjrosa:
Quick post-mortem on why ECE wasn't actually rendering on the change-payment screen until f605cb6ce: `display_express_checkout_button_html` was hooked to `woocommerce_checkout_before_customer_details` (the multi-section checkout template's action). WC Subscriptions renders the change-payment flow with its own `checkout/form-change-payment-method.php` template via `WC_Subscriptions_Change_Payment_Gateway`, which never fires that action — so the JS bundle loaded with no DOM mount point. Switched to `before_woocommerce_pay` (the action the gateway fires immediately before rendering its template) and the buttons paint as expected.

daledupreez (commented):
@wjrosa, manual testing is still not working correctly for me:
* When I immediately interact with the express checkout buttons, the "Use this payment method for all of my current subscriptions" checkbox remains selected, but the shopper may not have seen it -- I think we should not apply the payment change automatically for all subscriptions when express checkout methods are involved.
* The shopper-facing payment method doesn't seem to be updating correctly, regardless of the "change all" toggle from above - when I updated to a new Google Pay card, the shopper-facing value shows some other credit card as the payment method.
  - In my testing, I used two separate Mastercard numbers via Google Pay, but the subscription continued to show the Visa 4242 credit card.
  - The subscription admin shows the Google Pay method, but it also shows some notes saying that the payment method was changed from Google Pay to Credit Card

wjrosa:
Thanks for the careful repro, @daledupreez — both findings are real and addressed in ecd57c40:

1. **Cascade checkbox** — the deferred-intent change-payment flow now suppresses `update_all_subscriptions_payment_method` whenever the submission carries `express_checkout_type`. Covers both the immediate path (WCS reads `$_POST` directly inside its own `update_payment_method`) and the post-3DS path (the meta-stored `_delayed_update_payment_method_all` flag).
2. **Shopper-facing card not updating** — root cause was that the deferred-intent flow updates `_stripe_source_id` and `_payment_method` on the subscription but never calls `$subscription->add_payment_token()`, so My Account keeps rendering the previous token. Added a shared `WC_Stripe_Express_Checkout_Helper::replace_subscription_payment_token()` and call it after `handle_saving_payment_method()` in both the no-3DS branch and the post-3DS hook.

Scoped the token fix to ECE for now (gated on `express_checkout_type`) since the same gap likely affects non-ECE "new card on change-payment" submissions and is worth fixing with its own regression coverage. Tracked in #5382 along with the misleading `"…to Credit Card"` WCS-generated note (which your override already corrects in the visible title — the misleading copy is in the WCS-written note that lands before our hook runs).

daledupreez:
@wjrosa, this is still not working for me. I had the following initial state:
 * Subscription 1:  Purchased using Stripe test card 4242
 * Subscription 2: Purchased using ACH
 * Subscription 3: Purchased using Google Pay via card ending in 4242

I then tried to change the payment method for subscription 3 to Apple Pay. That showed as switching the payment method to Stripe via card ending in 4242.

I then tried to change the payment method for subscription 3 to Apple Pay. That also showed the payment method as the 4242 credit card. I then tried to change the payment method for subscription 3 to Google Pay, but using a different test card. That also shows as being via credit card ending in 4242 via the front end, but _is_ showing as Google Pay via the subscription edit page. I then tried to change the payment method to Link (using a Mastercard ending in 4444), but the display is incorrect for the shopper and merchant. The subscription order notes are also not super-helpful (see below for more).

#### General updated screens for customer

Screenshot 2026-05-13 at 14 52 00

The payment method is always card with 4242, regardless of the new payment method.

Screenshot 2026-05-13 at 14 57 58

#### Subscription details

Screenshot 2026-05-13 at 15 07 44

The sequence of payment method changes was: ACH (original purchase) -> Apple Pay -> Google Pay -> Link.

I was testing via https://previous-tuna.jurassic.ninja/ if you want to play around there.

wjrosa:
Thanks for pinging me on Slack, Dale. I was able to reproduce your issues when testing Link (Google Pay was fine for me). I fixed the issue in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5284/commits/9602c44f90d48875a333fe717b703f942c6ad86d. Can you give it another try please?

daledupreez:
Thanks for addressing those issues, @wjrosa! That's definitely looking better, though there are some features of the UX that aren't great and likely need to be fixed before we can merge.

For testing, I purchased a subscription using a Mastercard (4444) where I signed up to Link so I would be signed in. That showed up as "Stripe Link" with my email address in the subscription page, so all was well to start.

I then changed the payment method to Google Pay, then to Apple Pay, and then back to Link. (Screenshots are below.) I noticed the following issues:

 * For Apple Pay and Google Pay, the shopper-facing payment method shows up as the underlying card, and there is no mention of Apple Pay or Google Pay.
    - I tested purchasing a new subscription with Google Pay, and the saved payment method also stores the card details without the payment type, so I think this is a pre-existing behaviour.
 * The order notes are a bit haywire still, despite the main payment method being correct. In particular:
   - When changing from Link to Google Pay, the note specifies "Credit / Debit Card" to "Credit / Debit Card"
   - When changing from Google Pay to Apple Pay, the note specifies "Google Pay (Stripe)" to "Credit / Debit Card"
   - When changing from Apple Pay to Link, the note specifies "Apple Pay (Stripe)" to "Credit / Debit Card"

#### Payment method changed from Link to Google Pay

Screenshot 2026-05-13 at 21 26 41
Screenshot 2026-05-13 at 21 27 24

#### Payment method changed from Google Pay to Apple Pay

Screenshot 2026-05-13 at 21 29 12

Screenshot 2026-05-13 at 21 28 50

#### Payment method changed from Apple Pay to Link

Screenshot 2026-05-13 at 21 30 16

Screenshot 2026-05-13 at 21 30 33

#### All saved payment methods

Screenshot 2026-05-13 at 21 39 41

The list here is Link, Apple Pay, and Google Pay, but that's not obvious. 😁

wjrosa:
Note copy fixed in 07f260c, @daledupreez.

**1. Saved-method shows only the card** — pre-existing, as you noted. Tracked separately in #5437; fix proposed in #5439.

**2. "...to Credit Card" note** — WCS writes the note from `$gateway->get_title()` *before* `wc_stripe_change_subs_payment_method_success` lets us override the title. Hooked the existing `woocommerce_subscription_note_new_payment_method_title` filter to substitute the express label; covers both the no-3DS and post-3DS paths. Tests added.

"From" side: now correct when the previous change ran through ECE. Link → GP still shows "Credit / Debit Card" on the "from" — same item-#1 root cause (your Mastercard-with-Link-signup stored "Credit Card" as the title). That stored title isn't changed by #5439 (which only affects saved-token display in My Account), so the "from" side of the note will only update on future change-payment events; a separate item is needed if we want it backfilled.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -286,6 +286,22 @@ public function process_change_subscription_payment_with_deferred_intent( $subsc
 			];
 		}
 
+		$express_checkout_type          = isset( $_POST['express_checkout_type'] ) && is_string( $_POST['express_checkout_type'] ) // phpcs:ignore WordPress.Security.NonceVerification.Missing
+			? wc_clean( wp_unslash( $_POST['express_checkout_type'] ) ) // phpcs:ignore WordPress.Security.NonceVerification.Missing
+			: '';
+		$is_express_checkout_submission = '' !== $express_checkout_type;
```
daledupreez, line 294: I think we need to ensure that we have a valid, supported express checkout type before we proceed.
```suggestion
		$express_checkout_type = in_array( $express_checkout_type, WC_Stripe_Payment_Methods::EXPRESS_PAYMENT_METHODS, true ) ? $express_checkout_type : '';

		$is_express_checkout_submission = '' !== $express_checkout_type;
```

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -552,10 +558,19 @@ public function maybe_apply_express_title_after_confirmed_intent( $order ) {
...
 		$this->apply_express_checkout_title_to_order( $order, $express_checkout_type );
+
+		// Mirror the no-3DS path: associate the new WC payment token with the
+		// subscription so the shopper-facing My Account list reflects it.
+		$payment_method_id = $order->get_meta( '_wc_stripe_express_checkout_payment_method_id' );
+		if ( ! empty( $payment_method_id ) ) {
+			WC_Stripe_Express_Checkout_Helper::replace_subscription_payment_token( $order, $payment_method_id );
+		}
+
+		// One-shot: clear the markers even if the override didn't apply, so
```
daledupreez: What does the "one-shot" have to do with anything?

I am slightly more worried about the fact that these order meta fields are supposed to be cleaned up. Might it be better to store the values in some other location so it's less likely that they linger after their short-term purpose?

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -564,26 +579,86 @@ public function maybe_apply_express_title_after_confirmed_intent( $order ) {
...
+		if ( WC_Stripe_Payment_Methods::GOOGLE_PAY === $express_checkout_type ) {
+			return WC_Stripe_Payment_Methods::GOOGLE_PAY_LABEL . $suffix;
+		}
+		if ( WC_Stripe_Payment_Methods::LINK === $express_checkout_type ) {
+			// Match the title produced by the standard Link path (set_payment_method_title_for_order)
+			// and expected by WC_Stripe_UPE_Payment_Method_Link::filter_gateway_title — bare "Link", no suffix.
+			return WC_Stripe_Payment_Methods::LINK_LABEL;
+		}
+
+		return '';
```
daledupreez, line 627: Do we need to support/offer Amazon Pay in this logic as well?

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -67,6 +67,39 @@ public static function get_payment_method_title_suffix() {
...
+	 * Replaces the subscription's saved payment tokens with the WC token that
+	 * matches the given Stripe payment method ID. The change-payment flow
+	 * updates `_stripe_source_id` / `_payment_method` but not `_payment_tokens`,
+	 * so without this My Account keeps rendering the previously saved card.
+	 *
+	 * @param WC_Order $subscription      The subscription being updated.
+	 * @param string   $payment_method_id The Stripe payment method ID just attached to the subscription.
+	 * @return bool                       Whether a matching token was attached.
+	 */
+	public static function replace_subscription_payment_token( $subscription, $payment_method_id ) {
```
daledupreez, line 80: Nit: I don't think there is any part of this logic that is specific to express checkout. It may be worth considering another class to host this API.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -67,6 +67,39 @@ public static function get_payment_method_title_suffix() {
...
+		if ( ! $subscription instanceof WC_Order || empty( $payment_method_id ) ) {
+			return false;
+		}
+
+		$user_id = $subscription->get_user_id();
+		if ( ! $user_id ) {
+			return false;
+		}
+
+		$tokens = WC_Payment_Tokens::get_customer_tokens( $user_id, 'stripe' );
```
daledupreez: Question: do we have a constant for this value? It feels like we should, as it's not clear to me why we're passing in 'stripe' for this specific code.

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-element-test.php
```diff
@@ -179,6 +179,217 @@ public function provide_test_add_order_meta() {
...
+		$subscription->set_payment_method( 'stripe' );
+		$subscription->save();
+
+		$_POST['express_checkout_type'] = $checkout_type;
+
+		$filtered = $this->element->filter_change_payment_method_note_title( 'Credit Card', 'stripe', $subscription );
+
+		$this->assertSame( $expected, $filtered );
+
+		unset( $_POST['express_checkout_type'] );
```
daledupreez: Nit: we should reset `$_POST` before we assert.
```suggestion
		unset( $_POST['express_checkout_type'] );

		$this->assertSame( $expected, $filtered );
```

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-element-test.php
```diff
@@ -179,6 +179,217 @@ public function provide_test_add_order_meta() {
...
+		);
+
+		// Unrecognized express type: pass through.
+		$_POST['express_checkout_type'] = 'paypal';
+		$this->assertSame(
+			'Credit Card',
+			$this->element->filter_change_payment_method_note_title( 'Credit Card', 'stripe', $subscription )
+		);
+
+		unset( $_POST['express_checkout_type'] );
```
daledupreez: Nit: We should unset `$_POST` even if the assertion fails.
```suggestion
		try {
			$this->assertSame(
				'Credit Card',
				$this->element->filter_change_payment_method_note_title( 'Credit Card', 'stripe', $subscription )
			);
		} finally {
			unset( $_POST['express_checkout_type'] );
		}
```

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-element-test.php
```diff
@@ -179,6 +179,217 @@ public function provide_test_add_order_meta() {
...
+		$this->element->update_subscription_payment_method_title();
+		$this->assertSame( 'Stripe', $subscription->get_payment_method_title() );
+
+		// Unrecognized type should also leave the title unchanged.
+		$_POST['express_checkout_type'] = 'paypal';
+		$this->element->update_subscription_payment_method_title();
+		$this->assertSame( 'Stripe', $subscription->get_payment_method_title() );
+
+		unset( $_GET['change_payment_method'], $_POST['express_checkout_type'] );
+		WC_Subscriptions::$wcs_get_subscription = null;
```
daledupreez: We should make sure we clean up regardless of whether the assertions fail or succeed.
```suggestion
		$_GET['change_payment_method'] = $subscription->get_id();
		unset( $_POST['express_checkout_type'] );

		$this->element->update_subscription_payment_method_title();
		try {
			$this->assertSame( 'Stripe', $subscription->get_payment_method_title() );

			// Unrecognized type should also leave the title unchanged.
			$_POST['express_checkout_type'] = 'paypal';
			$this->element->update_subscription_payment_method_title();
			$this->assertSame( 'Stripe', $subscription->get_payment_method_title() );
		} finally {
			unset( $_GET['change_payment_method'], $_POST['express_checkout_type'] );
			WC_Subscriptions::$wcs_get_subscription = null;
		}
```

daledupreez (commented):
This is now testing well for me -- thanks for sticking with the multiple rounds of issues!

I have some feedback based on inspection, and I think the validation checks are a blocker for merging.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -67,6 +67,39 @@ public static function get_payment_method_title_suffix() {
...
+		foreach ( $tokens as $token ) {
+			if ( $token->get_token() === $payment_method_id ) {
+				$subscription->delete_meta_data( '_payment_tokens' );
```
malithsen: Question: Directly manipulating the meta seems dangerous. Is there an API we can use?

malithsen: Looking at `add_payment_token`, it seems like it will update the token, so this line might not be necessary.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Applied — `EXPRESS_PAYMENT_METHODS` allow-list is in place; an out-of-list value collapses to `''` before the submission check.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
wjrosa: Reworked. Dropped the "one-shot" language and explained why: order meta is the store because it survives the 3DS redirect (WC session can expire across the round-trip), and the cleanup now runs unconditionally so junk values can't linger even when `apply_express_checkout_title_to_order()` returned false.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
wjrosa: Added — new `AMAZON_PAY_LABEL` constant, a case in `get_express_checkout_method_title`, and an `amazon_pay` row in the title data provider. Amazon Pay now resolves to `"Amazon Pay (Stripe)"` like the other wallet paths.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Fair point — none of `replace_subscription_payment_token` is ECE-specific anymore (it's called from the trait and the intent controller too). Leaving the move as a follow-up so this PR stays scoped; happy to relocate it if you'd prefer.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Switched to `WC_Stripe_UPE_Payment_Gateway::ID` (the canonical `"stripe"` constant in the codebase). Same value, just stops hard-coding the literal.

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-element-test.php
wjrosa: Already applied in the current diff (`unset` on line 294 before `assertSame`).

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-element-test.php
wjrosa: Already applied — `try { ... } finally { unset(...) }` around the third assertion (lines 352-359).

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-element-test.php
wjrosa: Already applied — `try/finally` wraps the assertions, the `finally` unsets both `$_GET['change_payment_method']` and `$_POST['express_checkout_type']` and clears the `wcs_get_subscription` mock (lines 384-394).

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Switched to `$subscription->get_data_store()->update_payment_token_ids( $subscription, [ $token->get_id() ] )` — that's the public API and works under both CPT and HPOS. Re: your follow-up on `add_payment_token` updating: it actually **appends** (`$token_ids[] = $token->get_id();` in WC core), so without an explicit replace we'd have left the old token alongside the new one. Confirmed with a new test that locks in the replace behaviour.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -67,6 +67,44 @@ public static function get_payment_method_title_suffix() {
...
+		foreach ( $tokens as $token ) {
+			if ( $token->get_token() === $payment_method_id ) {
+				// `_payment_tokens` is in WC's internal-meta-keys list, so
+				// `delete_meta_data()` is a no-op on it; `add_payment_token()`
+				// then appends rather than replaces, leaving the stale token
+				// alongside the new one. The data-store API rewrites the list
+				// in one call (and works under both CPT and HPOS).
+				// PHPStan can't follow WC_Data_Store's generic factory return type to
+				// the order data store that owns update_payment_token_ids().
+				$subscription->get_data_store()->update_payment_token_ids( $subscription, [ $token->get_id() ] ); // @phpstan-ignore-line method.notFound
```
daledupreez: Nit: I think we should move the PHPStan commentary into the ignore comment.
```suggestion
				$subscription->get_data_store()->update_payment_token_ids( $subscription, [ $token->get_id() ] ); // @phpstan-ignore-line method.notFound (PHPStan can't follow the generic get_data_store() return type to the order data store that implements update_payment_token_ids().)
```

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
daledupreez: As long as we take that on as a fast-follow, I am OK merging as-is. I just want to make sure we get the location right before we release -- it will be a lot faster and less complicated to change if we don't have to maintain backwards compatibility for a few releases.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2455,6 +2455,18 @@ public function process_order_for_confirmed_intent( $order, $intent_id, $save_pa
 
 		$this->save_intent_to_order( $order, $intent );
 		$this->set_payment_method_title_for_order( $order, $payment_method_type );
+
+		/**
+		 * Fires after the payment method title is set on a confirmed intent, allowing
+		 * extensions (e.g. express checkout) to override the title for special cases.
+		 *
+		 * @since 10.7.0
```
malithsen:
```suggestion
		 * @since 10.8.0
```

daledupreez (approved):
Thanks for the persistence! I've re-tested with OCS on and off, and all is looking better. We can still improve the way we display the token details, but you have multiple PRs tackling that problem!

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -492,6 +509,164 @@ public function add_order_meta( $order_id, $posted_data ) {
...
+		}
+
+		$express_checkout_type = '';
+
+		// No-3DS path: the form submission is in progress, so $_POST is authoritative.
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing
+		if ( ! empty( $_POST['express_checkout_type'] ) && ! is_array( $_POST['express_checkout_type'] ) ) {
+			// phpcs:ignore WordPress.Security.NonceVerification.Missing
+			$express_checkout_type = sanitize_text_field( wp_unslash( $_POST['express_checkout_type'] ) );
+		} elseif ( $subscription instanceof WC_Order ) {
```
malithsen: Shouldn't this be?

```suggestion
		} elseif ( $subscription instanceof WC_Subscription ) {
```

Docblock type for the parameter is already `WC_Subscription`. Also, since `WC_Subscription` extends `WC_Order`, this check will probably pass, but it's worth tightening this.

malithsen (approved):
The changes look good to me on inspection. Left a couple of comments that are still worth addressing.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
wjrosa: Done in b9bf5bf8e — agreed, the WCS filter only ever hands us a `WC_Subscription` and the docblock already declares it, so tightening matches the contract and locks out a stray non-subscription order ever picking up the express-type meta.

PR #5287, opened 2026-04-15
Use paratest in CI workflow for faster PHP test execution

wjrosa:
## Summary

- Switches the CI PHP test runner (`bin/phpunit.sh`) from sequential `phpunit` to parallel `paratest` execution.
- Uses `phpunit-parallel.xml.dist` (without `<report>` section) and `--no-coverage` for clean parallel runs.
- Sets `PARATEST=true` during test environment setup so per-worker databases are created.
- Exports DB connection details (`WP_TESTS_DB_NAME`, `WORDPRESS_DB_HOST`, `MYSQL_ROOT_PASSWORD`) for the bootstrap safety net.
- Caches `/tmp/wordpress` and `/tmp/wordpress-tests-lib` in `php-tests.yml` and the `ci-tests-with-coverage` composite action, keyed on WP+WC version.
- Stubs `wp-content/uploads/woocommerce-placeholder.webp` in the test bootstrap to work around a WooCommerce 10.7.0-rc.1 regression (see below).

This builds on #5139 which added paratest for local development.

## Why the CI speedup is smaller than local

Running paratest locally (8–16 CPUs, warm caches, Docker MySQL) produces a 3–4× wallclock reduction. On GitHub-hosted runners the gain is closer to 1.5–2×. The main causes:

1. **4 vCPUs on `ubuntu-22.04` runners.** Paratest scales ~linearly with cores, so the ceiling in CI is ~4×, versus 8–16× on most developer machines. `install-wp-tests.sh` confirms this: only 4 worker DBs get provisioned per job.
2. **Per-worker bootstrap dominates a small suite.** With ~88 test files, each worker pays the full WP + WooCommerce + plugin bootstrap + MySQL connect cost. Paid 4×, that fixed cost can be 30–50% of wallclock.
3. **Matrix already parallelizes.** `php-tests.yml` fans out to 15+ jobs (WC × PHP × WP) with `max-parallel: 16`. The runner pool is already saturated, so shaving 40% off one job doesn't help overall pipeline time unless it's the slowest matrix cell.
4. **Shared MySQL service is slower than Docker MySQL on localhost.** All 4 workers hit the same instance; WP_UnitTestCase constantly truncates/recreates tables. DB I/O is the real bottleneck for WP tests.
5. **Cold caches every run.** No warm OS page cache, MySQL buffer pool, or PHP opcache across runs.

The caching commit (#567f421b) removes the 20–40 s per-cell cost of downloading WP core and the SVN test-suite checkout on every run, which is one of the bigger unavoidable costs on cold runners.

Further gains are possible with larger GitHub-hosted runners (8-core or 16-core), but that requires org-level provisioning and is out of scope for this PR.

## Why the woocommerce-placeholder.webp stub is necessary

The first post-merge runs of `compatibility.yml` (the `WC=beta` matrix) surfaced a flaky set of failures like:

```
PHP Warning: exif_imagetype(/tmp/wordpress/wp-content/uploads/woocommerce-placeholder.webp):
  Failed to open stream: No such file or directory
```

Root cause: **WooCommerce 10.7.0-rc.1 fails to generate the WebP placeholder** on plugin activation. Any test that creates an order or renders a product triggers `wp_getimagesize()` on the missing file. Because `phpunit.xml.dist` sets `convertWarningsToExceptions="true"`, those warnings become test errors — affecting 3–5 tests per run, in a varying subset depending on test ordering (hence the flakiness).

The bootstrap stub copies WC's bundled `assets/images/placeholder.webp` (falling back to `placeholder.png` for older WC versions) into the expected uploads path if it's missing. This:

- Preserves test coverage across the entire compat matrix — nothing is skipped.
- Is a no-op on fixed WC releases where the placeholder already exists.
- Can be removed once a fixed WooCommerce release is in the matrix.

Verified the failure is unrelated to the paratest or caching changes: the same errors appeared on the develop-merge commit (`e75bf23a`) before the caching commit, and the same workflow is failing on several other unrelated PRs (`add/payment-method-domain-class`, `add/refund-domain-class`, etc.) opened around the same time.

## Test plan

- [ ] Verify CI PHP tests pass across all matrix combinations (PHP 7.4–8.3, WP L/L-1, WC L/L-1/L-2).
- [ ] Verify `compatibility.yml` (`WC=beta`) passes now that the placeholder stub is in place.
- [ ] Confirm test count matches sequential run.
- [ ] Compare CI run time before and after.


.github/actions/ci-tests-with-coverage/action.yml
```diff
@@ -19,6 +19,14 @@ runs:
         path: ~/.cache/composer/
         key:  ${{ runner.os }}-composer-${{ hashFiles('composer.lock') }}
 
+    - name: Cache WordPress core and test suite
+      uses: actions/cache@v4
+      with:
+        path: |
+          /tmp/wordpress
+          /tmp/wordpress-tests-lib
+        key: wp-testenv-${{ env.WP_VERSION }}-wc-${{ env.WC_VERSION }}
```
Mayisha, line 28: AI suggested to add `restore-keys: wp-testenv-` in both the yml files.

bin/phpunit.sh
```diff
@@ -4,4 +4,7 @@ if [[ $RUN_PHPCS == 1 || $SHOULD_DEPLOY == 1 ]]; then
 	exit
 fi
 
-./vendor/bin/phpunit
+./vendor/bin/paratest \
+	--configuration phpunit-parallel.xml.dist \
+	--no-coverage \
+	"$@"
```
Mayisha, line 10: AI suggestion: We should check the coverage flag and use the `--no-coverage` based on that flag.

```
COVERAGE_FLAGS="--no-coverage"                                                                                                                                                                               
  if [ "${XDEBUG_MODE_PHPUNIT:-}" = "coverage" ]; then                                                                                                                                                         
      COVERAGE_FLAGS="--coverage-cobertura php-coverage.xml"                                                                                                                                                   
  fi                                                                                                                                                                                                           
                                                                                                                                                                                                               
  ./vendor/bin/paratest \
      --configuration phpunit-parallel.xml.dist \                                                                                                                                                              
      $COVERAGE_FLAGS \
      "$@" 
```

Mayisha (approved):
Looks good to me and the tests are clean in this PR. I have added some AI suggestions to consider.

daledupreez (approved):
This looks reasonable to me, and it does look like there is a speedup from 5-6 minutes per run to 3-4 minutes per run, which is noticeable, but not a 2x speedup.

.github/actions/ci-tests-with-coverage/action.yml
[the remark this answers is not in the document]
wjrosa: Done in 9f01c461 — added `hashFiles('bin/install-wp-tests.sh', 'bin/run-ci-tests.bash')` plus `runner.os` to the cache key, with progressively broader `restore-keys:` fallbacks so partial cache hits still speed up cold runs.

.github/workflows/php-tests.yml
```diff
@@ -84,6 +84,14 @@ jobs:
           path: ~/.cache/composer/
           key:  ${{ runner.os }}-composer-${{ hashFiles('composer.lock') }}
 
+      - name: Cache WordPress core and test suite
+        uses: actions/cache@v4
+        with:
+          path: |
+            /tmp/wordpress
+            /tmp/wordpress-tests-lib
+          key: wp-testenv-${{ matrix.wordpress }}-wc-${{ matrix.woocommerce }}
```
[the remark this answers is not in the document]
wjrosa: Done in 9f01c461 — cache key now includes `runner.os` and `hashFiles('bin/install-wp-tests.sh', 'bin/run-ci-tests.bash')`, so changes to the installer (including the `wp-tests-config.php` regeneration logic) will invalidate the cache.

.github/actions/ci-tests-with-coverage/action.yml
wjrosa: Added in 9f01c461 — both workflows now have `restore-keys:` fallbacks progressively widening from `wp-testenv-<os>-<wp>-<wc>-` down to `wp-testenv-<os>-`, so a changed installer hash still reuses most of the cached test environment.

bin/phpunit.sh
wjrosa: Done in 9f01c461 — `bin/phpunit.sh` now gates coverage flags behind `XDEBUG_MODE_PHPUNIT=coverage`, mirroring the pattern already used by `bin/run-tests-parallel.sh`.

PR #5288, opened 2026-04-15
Replace Docker with wp-env for local development

wjrosa:

Replaces the custom Docker setup with `@wordpress/wp-env` for local development, aligning with how WooCommerce core and most other WP plugins manage their dev environments.

### Why wp-env?

- **Standard tool, less code to maintain.** wp-env is the official Gutenberg-team-maintained local-dev tool for WP plugins. Every contributor who has worked on Gutenberg, WooCommerce core, or any modern WP block plugin already knows it. Switching means we delete a custom `docker-compose.yml` + `bin/docker-setup.sh` and ride on a tool that has dedicated maintainers and a much larger user base.
- **Cleaner first-run experience.** `npm run up` boots WP + WC + Storefront with one command. No manual MySQL bootstrap, no separate `bin/docker-setup.sh` step, no hand-coding the WC install hook.
- **Built-in WP / WC / PHP version switching.** `.wp-env.json` pins the matrix declaratively (this PR uses PHP 8.0, WC 9.9, WP latest). Contributors testing against L-1 or L-2 just edit the JSON; no Dockerfile churn.
- **Cheaper context-switching across PRs.** Each branch can carry its own `.wp-env.json` overrides for the WP/WC versions it needs to validate, without contributors having to remember which `docker-compose` flags to pass.
- **Easier integration with WordPress's own test suite.** `wp-env run tests-cli` is a one-liner for PHPUnit; `npm run test:php:wp-env` wraps that. The current Docker setup needs separate `bin/run-tests.sh` plumbing.

The legacy Docker setup is preserved as `docker:up` / `docker:down` while the migration is being evaluated.

**Core changes**
- Added `.wp-env.json` (PHP 8.0, port 8072, WooCommerce + Storefront auto-installed).
- Added `@wordpress/env` as a devDependency.
- Repointed `npm run up` / `npm run down` at wp-env; preserved the previous behaviour as `npm run docker:up` / `npm run docker:down`.
- Added `npm run test:php:wp-env` for running PHPUnit through wp-env.
- Added `bin/run-tests-wp-env.sh` runner script.
- Updated `AGENTS.md` and `docs/DOCKER.md` so the first command a new contributor runs is the wp-env one.

### How can this code break?

- **Two parallel local-dev paths.** The legacy Docker setup stays around as `docker:up`/`docker:down`. Contributors splitting their time between branches that drift in/out of wp-env support will hit confusion. Mitigated by the doc updates and by making wp-env the default `up`/`down` target.
- **wp-env's own image churn.** wp-env pulls fresh WP/WC/PHP images on `up` after a `wp-env destroy`. First-run cost on a clean machine is non-trivial; documented in `docs/DOCKER.md`.
- **Test harness divergence.** `npm run test:php` still hits the legacy Docker stack; `npm run test:php:wp-env` runs against wp-env. Until CI runs both, a passing test on one stack doesn't guarantee the other. The `test:php:parallel` paratest path remains Docker-only.
- **Coexistence with [PR #5443](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443).** That PR refactors the Docker setup to be worktree-friendly. The two PRs aim at overlapping pain points (parallel branch environments, port collisions). If both land, the Docker stack becomes a maintained-but-unused secondary path. We should decide which one becomes the primary local-dev tool before merging either.
- **Tooling that shells out to the legacy `docker compose` plumbing.** Scripts in `bin/` that hard-code `docker compose -p woocommerce_stripe` will keep working only against the legacy `docker:up` path.

- The full CI matrix on this branch (linting, JS tests, PHP tests, E2E) still runs against the legacy Docker stack — CI is not switched to wp-env, so the regression surface there is zero.
- Both local paths are documented side by side in `AGENTS.md` and `docs/DOCKER.md` with explicit commands so contributors don't have to guess.
- The legacy `docker:up`/`docker:down` aliases stay until we see fleet-wide confirmation that wp-env covers every flow today's setup handles (xdebug, JT, webhook listener, etc.).

## Testing instructions

1. Run `npm install` to pull `@wordpress/env`.
2. Run `npm run up`. wp-env should bring up WordPress, WooCommerce, and Storefront at `http://localhost:8072`. Verify the Stripe plugin is active in **wp-admin → Plugins**.
3. Run `npm run test:php:wp-env`. PHPUnit should run through the wp-env `tests-cli` container and pass.
4. Run `npm run down` and confirm the wp-env containers stop.
5. Run `npm run docker:up`. The legacy Docker stack should still come up on `http://localhost:8072`. Stop with `npm run docker:down`.
6. Confirm every CI workflow on this PR goes green — they continue to use the legacy Docker stack, so the wp-env addition should not break them.

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️) — tooling migration; the existing CI suite continues to exercise the legacy Docker stack while the wp-env path is validated by contributors.
-   [ ] Tested on mobile (or does not apply) — does not apply.

### Changelog entry

`Dev - Replace Docker with wp-env for local development` — added to `changelog.txt` and `readme.txt` under the unreleased 10.8.0 section.

**Post merge**

-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply) — contributor docs already updated in this PR


wjrosa:
Given the latest changes to the local environment with worktrees, this is no longer worth it.

PR #5106, opened 2026-03-09
Add deprecation notices to methods deprecated in older versions

wjrosa:
Addresses review feedback from the original code-removal PR.

## Changes proposed in this Pull Request

This PR adds `wc_deprecated_function()` calls to methods and properties that were marked as deprecated in their docblocks but never had a runtime deprecation notice. This makes deprecations "louder" so callers can detect them.

**What this PR does:**
- Adds `wc_deprecated_function()` to deprecated methods in:
  - `WC_Stripe_Payment_Gateway`: `validate_minimum_order_amount()`, `get_owner_details()`, `lock_order_payment()`, `unlock_order_payment()`, `get_order_existing_lock()`, `is_order_payment_locked()`, `lock_order_refund()`, `unlock_order_refund()`
  - `WC_REST_Stripe_Settings_Controller`: `dismiss_customization_notice()`
  - `WC_Stripe_Helper`: `get_stripe_currency()`, `update_stripe_currency()`, `get_stripe_fee()`, `update_stripe_fee()`
  - `WC_Stripe_Logger`: `log()`
  - `WC_Stripe_Subscriptions_Utilities_Trait`: `is_manual_renewal_required()`
  - `WC_Stripe_Express_Checkout_Helper`: `get_booking_id_from_cart()`
  - `WC_Stripe_UPE_Payment_Gateway`: `get_testing_instructions_for_optimized_checkout()`, `is_sepa_tokens_for_other_methods_enabled()`
  - `WC_Stripe_UPE_Payment_Method`: `is_sepa_tokens_for_other_methods_enabled()`
- Fixes consistency: all replacement method names in `wc_deprecated_function()` calls now include fully-qualified class names and trailing `()`
- Updates PHPStan baseline to remove entries fixed by improved type annotations (`@param WC_Order` instead of `@param object`, added `@return void`)

**What this PR does NOT do:**
- Actual removal of deprecated code — that is handled in the follow-up PR: #5133 (branched from this one)

## Testing instructions

Code review. Perform some basic smoke testing to make sure nothing breaks.

---

- [x] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

### Changelog entry

`* Update - Add deprecation notices to methods and properties that were deprecated without them in older versions`

**Post merge**

- [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
[...]

PR #5294, opened 2026-04-16
Remove deprecated code up to version 10.1.0

wjrosa:
2nd attempt after reverting it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5293

Follow-up to #5106. Branches from that PR and adds the actual removals.

Removes deprecated code up to version 10.1.0 after #5106 has already added runtime `wc_deprecated_function()` notices to allow users to discover and update their usage. Some items are kept as deprecation shims (loud notice on each call/access) instead of being fully removed, so external dashboards and extensions that still call them get a clear signal before the next removal pass.

### Removed

- `WC_Stripe::STRIPE_GATEWAY_SETTINGS_OPTION_NAME` constant (deprecated 8.7.0)
- `WC_REST_Stripe_Account_Keys_Controller::STRIPE_GATEWAY_SETTINGS_OPTION_NAME` constant (deprecated 8.7.0)
- `WC_Stripe_Connect::SETTINGS_OPTION` constant (deprecated 8.7.0)
- `WC_Stripe_Webhook_State::get_testmode()` method (deprecated 8.9.0)
- `WC_Stripe_Payment_Tokens::get_account_saved_payment_methods_list_item_sepa()` method (deprecated 8.4.0)
- `WC_Stripe_Intent_Controller::maybe_process_upe_redirect()` method (deprecated 8.3.0)
- `WC_Stripe_Admin_Notices::subscriptions_check_environment()` method (deprecated 9.6.0)
- `WC_Stripe_Subscriptions_Helper::get_some_detached_subscriptions()` method (deprecated 9.6.0)
- `WC_Stripe_Subscriptions_Utilities_Trait::is_subscriptions_enabled()` method (deprecated 9.2.0)
- `WC_Stripe_Express_Checkout_Ajax_Handler::ajax_pay_for_order()` method (deprecated 9.2.0)
- `WC_Stripe_Express_Checkout_Helper::use_blocks_api()` method (deprecated 9.2.0)
- `WC_Stripe_Feature_Flags::ECE_FEATURE_FLAG_NAME` constant (deprecated 10.0.0)
- `WC_Stripe_UPE_Payment_Gateway::SUCCESSFUL_INTENT_STATUS` constant (deprecated 9.1.0)
- `WC_Stripe_UPE_Availability_Note` class (deprecated 10.2.0)

### Kept as deprecation shims

- `WC_Stripe_Feature_Flags::OC_FEATURE_FLAG_NAME` constant — kept (still referenced in feature flag map).
- `WC_Stripe_Feature_Flags::is_stripe_ece_enabled()`, `is_upe_preview_enabled()`, `is_upe_checkout_enabled()` — kept as stubs returning `true` (`@deprecated` doc annotations).
- `WC_Stripe_REST_UPE_Flag_Toggle_Controller` class (deprecated 10.2.0) — kept; `get_flag()` and `set_flag()` now emit `wc_deprecated_function()` notices on each call. The class previously had only an `@deprecated` doc which most tools / developers don't surface.
- `WC_Stripe_UPE_Payment_Gateway::$sepa_tokens_for_other_methods` (deprecated 10.0.0) — kept as `private`. `__set()` and `__get()` magic methods emit `wc_deprecated_function()` on read or write so external code hits a loud notice rather than silently shadowing or receiving `null`.
- `WC_Stripe_UPE_Payment_Gateway::$spe_enabled` (deprecated 9.5.0) — kept as `private`, same `__set()` / `__get()` deprecation pattern.

## Testing instructions

Code review. Perform basic smoke testing to confirm nothing breaks.

---

### Changelog entry

`* Remove - Remove deprecated code up to version 10.1.0`

**Post merge**

cc @daledupreez


includes/admin/class-wc-stripe-express-checkout-controller.php
```diff
@@ -46,7 +46,7 @@ public function admin_scripts(): void {
 
 		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
 		$params          = [
-			'key'                     => WC_Stripe_Mode::is_test() ? $stripe_settings['test_publishable_key'] : $stripe_settings['publishable_key'],
+			'key'                     => WC_Stripe_Mode::is_test() ? ( $stripe_settings['test_publishable_key'] ?? '' ) : ( $stripe_settings['publishable_key'] ?? '' ),
```
daledupreez: Nit: Not that I am opposed to the fix, but it doesn't feel appropriate to include it in this PR.

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
```diff
@@ -1,92 +0,0 @@
...
-	/**
-	 * Update the data.
-	 *
-	 * @param WP_REST_Request $request Full data about the request.
-	 *
-	 * @deprecated 10.2.0 UPE is generally available and this endpoint will be removed in a future release.
-	 *
-	 * @return WP_REST_Response
-	 */
-	public function set_flag( WP_REST_Request $request ) {
```
daledupreez, line 78: Just having `@deprecated` annotations is not going to be caught by most developers or tools. Rather than removing this file entirely, can we keep the code but add explicit deprecated notices when the methods get run?

tests/phpunit/class-wc-stripe-feature-flags-test.php
```diff
@@ -23,8 +23,7 @@ public function set_up() {
 	 * Test for `is_oc_available`.
 	 *
 	 * @param bool $pmc_enabled Whether the Payment Method Configuration API is enabled.
-	 * @param string $filter_function The filter function to apply.
-	 * @param bool   $expected     The expected result.
+	 * @param bool   $expected  The expected result.
```
daledupreez, line 27: The parameter docs should be reinstated, as the `$filter_function` argument is still being supplied.
```suggestion 
	 * @param string $filter_function The filter function to apply.
	 * @param bool   $expected  The expected result.
```

changelog.txt
```diff
@@ -18,6 +18,7 @@ xxxx-xx-xx - version 10.8.0
 * Fix - Show better message when express checkout preview is not available
 * Dev - Add Claude Code skills and review rules under .claude/ to capture repo-specific contributor guidance
 * Dev - Move some independent classes into autoloader
+* Update - Remove deprecated code up to version 10.1.0
```
daledupreez: Should we use `Remove` for these entries?

daledupreez (commented):
Overall, this is looking good, though I have some minor comments, and one suggestion to make sure our code is louder about deprecations before we remove the code.

> * `WC_Stripe_UPE_Payment_Gateway::$sepa_tokens_for_other_methods` — made private, deprecation notice via __set() when WP_DEBUG is on
> * `WC_Stripe_UPE_Payment_Gateway::$spe_enabled` — made private, deprecation notice via __set() when WP_DEBUG is on

There is no `__get()` implementation, so the code or the description is wrong. I think we should implement `__get()` so we can be louder about continued access.

Also, as a more general check: how long have all the deprecation notices been in the code? I know we added a lot of them a few months ago, despite the original light/quiet deprecations happening much earlier. Before we can remove these deprecations, I think we need to triple-check that the deprecation notices have been in place for a few releases.

wjrosa:
@daledupreez addressed your top-level review in 52548cdbd:

- Restored `$sepa_tokens_for_other_methods` and `$spe_enabled` as private properties with both `__set()` and `__get()` deprecation notices (fair call — read access was silently dropping to `null` before).
- Kept `WC_Stripe_REST_UPE_Flag_Toggle_Controller` as a shim with `wc_deprecated_function()` on both endpoints; dropped the now-removed `WC_Stripe_UPE_Availability_Note` side-effects from `set_flag()`.
- Reverted the out-of-scope null-coalescing tweak.
- Changelog entries switched to `Remove -`.
- PR description updated to reflect the actual state.

On the deprecation-age check: I've gone through the `@deprecated` tags on each removed item — the oldest are 8.3.0 / 8.4.0 / 8.7.0 (well over a year in code), the youngest is 10.2.0 (the now-shimmed UPE flag toggle controller). The `wc_deprecated_function` runtime notices from #5106 went in around 9.7.0–10.1.0 for the older items; for items added later we kept them as shims rather than removing.

includes/admin/class-wc-stripe-express-checkout-controller.php
wjrosa: Reverted in 52548cdbd — that snuck in during a merge resolution. Will file separately if it ever matters.

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
wjrosa: Kept as a shim in 52548cdbd — both endpoints now emit `wc_deprecated_function()` on each call. Dropped the dependent `WC_Stripe_UPE_Availability_Note` references from `set_flag()` since that class is removed in this PR.

changelog.txt
wjrosa: Switched both `changelog.txt` and `readme.txt` to `Remove -` in 52548cdbd.

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
```diff
@@ -6,7 +6,16 @@
...
+ * REST controller for the legacy UPE feature flag.
+ *
+ * UPE has been generally available for several releases and the underlying
+ * feature flag was already a no-op; the constants this endpoint used to
+ * toggle have since been removed from the codebase. The class is kept as a
+ * deprecation shim so any external dashboard / extension still calling the
+ * `wc_stripe/upe_flag_toggle` endpoint continues to receive the long-standing
+ * "always on" answer while emitting a loud deprecation notice that tooling
+ * (Query Monitor, `WP_DEBUG_LOG`, error log scrapers) actually surfaces. The
+ * `@deprecated` PHPDoc alone was too quiet to be caught.
```
daledupreez: I think we can simplify this to something much shorter.
```suggestion
 * This class is fully deprecated, but is being kept in the code for backwards compatibility.
 * It will be fully removed in an upcoming release.
```

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
```diff
@@ -67,25 +78,18 @@ public function get_flag() {
...
-	 *
-	 * @param WP_REST_Request $request Full data about the request.
+	 * Update the flag. No-op shim: UPE is already generally available, and the
+	 * note-cleanup side-effects this endpoint used to perform run from elsewhere
+	 * now. Always reports success so legacy callers don't break.
 	 *
 	 * @deprecated 10.2.0 UPE is generally available and this endpoint will be removed in a future release.
 	 *
+	 * @param WP_REST_Request<array<string, mixed>> $request Full data about the request.
+	 *
```
daledupreez: As above, I think we can simplify this comment to be way less verbose. Also as a minor nit, this PR changed the PHPDoc order.
```suggestion
	 * Deprecated no-op. Kept in the code for backwards compatibility.
	 *
	 * @param WP_REST_Request<array<string, mixed>> $request Full data about the request.
	 *
	 * @deprecated 10.2.0 UPE is generally available and this endpoint will be removed in a future release.
	 *
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -110,13 +101,18 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 	public $saved_cards;
 
 	/**
-	 * Should SEPA tokens be used for other payment methods (iDEAL and Bancontact)
+	 * Should SEPA tokens be used for other payment methods (iDEAL and Bancontact).
+	 *
+	 * Kept as a private property so external code that previously set or read
+	 * this hits __set / __get and gets a loud `wc_deprecated_function` notice
+	 * (the doc-only `@deprecated` was too quiet). Replaced by
+	 * `$sepa_tokens_for_ideal` and `$sepa_tokens_for_bancontact`.
```
daledupreez: As above, let's simplify these comments.
```suggestion
	 * Should SEPA tokens be used for other payment methods (iDEAL and Bancontact).
	 *
	 * Previously public property that is now accessed via `__get()` after deprecation.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -135,11 +131,15 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 	/**
 	 * Is Single Payment Element enabled?
 	 *
+	 * Kept as a private property so external code that previously set or read
+	 * this hits __set / __get and gets a loud `wc_deprecated_function` notice
+	 * (the doc-only `@deprecated` was too quiet). Replaced by `$oc_enabled`.
+	 *
```
daledupreez: As with other locations, let's simplify the comment.
```suggestion
	 * Deprecated property that was previously public and is now private for backwards compatibility.
	 * Deprecation notices are triggered via {@see __get()} and {@see __set()}.
	 *
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4939,4 +4939,54 @@ private function get_presentment_data_from_order( $order ): ?array {
 			'currency' => $currency,
 		];
 	}
+
+	/**
+	 * Trap writes to deprecated public properties so external code that hasn't
+	 * migrated emits a `wc_deprecated_function` notice instead of silently
+	 * shadowing a now-private property.
```
daledupreez: As above, let's simplify the wording.
```suggestion
	/**
	 * Magic setter to add warnings for deprecated properties.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4939,4 +4939,54 @@ private function get_presentment_data_from_order( $order ): ?array {
...
+	 *
+	 * @param string $name  Property name.
+	 * @param mixed  $value Value being assigned.
+	 * @return void
+	 */
+	public function __set( $name, $value ) {
+		if ( 'sepa_tokens_for_other_methods' === $name ) {
+			wc_deprecated_function( static::class . '::$' . $name, '10.0.0', '$sepa_tokens_for_ideal and $sepa_tokens_for_bancontact' );
```
daledupreez, line 4949: Should these be more direct `wc_doing_it_wrong()` or `_doing_it_wrong()` calls?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4939,4 +4939,54 @@ private function get_presentment_data_from_order( $order ): ?array {
...
+
+		// Preserve PHP's default behaviour for unknown properties so we don't
+		// silently swallow typos elsewhere.
+		$this->$name = $value;
+	}
+
+	/**
+	 * Trap reads of deprecated public properties so external code that hasn't
+	 * migrated emits a `wc_deprecated_function` notice instead of silently
+	 * receiving null.
```
daledupreez: Another simplification.
```suggestion
	 * Magic getter to add warnings for deprecated properties.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4939,4 +4939,54 @@ private function get_presentment_data_from_order( $order ): ?array {
...
+		}
+		if ( 'spe_enabled' === $name ) {
+			wc_deprecated_function( static::class . '::$' . $name, '9.5.0', '$oc_enabled' );
+			return $this->spe_enabled;
+		}
+
+		// PHP would emit a notice and return null here for an undefined
+		// property; mirror that explicitly to keep behaviour predictable.
+		trigger_error( esc_html( 'Undefined property: ' . static::class . '::$' . $name ), E_USER_NOTICE ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
+		return null;
```
daledupreez, line 4992: We should check whether our inheritance tree has a method we should be dispatching to before we record the error.
```suggestion
		if ( method_exists( parent::class, '__get' ) ) {
			return parent::__get( $name );
		}

		// PHP would emit a notice and return null here for an undefined
		// property; mirror that explicitly to keep behaviour predictable.
		trigger_error( esc_html( 'Undefined property: ' . static::class . '::$' . $name ), E_USER_NOTICE ); // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_trigger_error
		return null;
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4939,4 +4939,54 @@ private function get_presentment_data_from_order( $order ): ?array {
...
+	 * @param string $name Property name.
+	 * @return mixed
+	 */
+	public function __get( $name ) {
+		if ( 'sepa_tokens_for_other_methods' === $name ) {
+			wc_deprecated_function( static::class . '::$' . $name, '10.0.0', '$sepa_tokens_for_ideal and $sepa_tokens_for_bancontact' );
+			return $this->sepa_tokens_for_other_methods;
+		}
+		if ( 'spe_enabled' === $name ) {
+			wc_deprecated_function( static::class . '::$' . $name, '9.5.0', '$oc_enabled' );
```
daledupreez, line 4981: As for the setter, should these be doing_it_wrong calls?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4939,4 +4939,54 @@ private function get_presentment_data_from_order( $order ): ?array {
...
+			$this->spe_enabled = $value;
+			return;
+		}
+
+		// Preserve PHP's default behaviour for unknown properties so we don't
+		// silently swallow typos elsewhere.
+		$this->$name = $value;
```
daledupreez, line 4966: We should check if our inheritance tree has a magic setter before we fall back on this.
```suggestion
		if ( method_exists( parent::class, '__set' ) {
			parent::__set( $name, $value );
			return;
		}

		// Preserve PHP's default behaviour for unknown properties so we don't
		// silently swallow typos elsewhere.
		$this->$name = $value;
```

includes/class-wc-stripe.php
```diff
@@ -870,6 +863,7 @@ public function register_routes() {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-rest-stripe-account-keys-controller.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-stripe-rest-oc-setting-toggle-controller.php';
 
+		// Deprecated 10.2.0; kept as a shim that loudly logs each call. See the class for context.
```
daledupreez: I don't think this comment is needed.

tests/phpunit/class-wc-stripe-feature-flags-test.php
```diff
@@ -45,14 +45,13 @@ public function test_is_oc_available( $pmc_enabled, $filter_function, $expected
 		$actual = WC_Stripe_Feature_Flags::is_oc_available();
 
 		// Clean up
+		if ( ! empty( $filter_function ) ) {
```
daledupreez, line 48: Good catch!

daledupreez (commented):
I have a good amount of feedback, mostly around inline comments, though I think we may need to tweak the magic getter and setter implementations a little bit before we can merge.

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
wjrosa: Applied verbatim in 7964e32f5.

includes/admin/class-wc-stripe-rest-upe-flag-toggle-controller.php
wjrosa: Applied in 28c6073f0 — picked up the simpler wording and the PHPDoc order.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Applied in cab7f45c6.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Applied in df6836f06.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Applied in 9026073ab.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Applied in 7ab069da8.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Kept as `wc_deprecated_function()`: the property itself is deprecated, which matches the semantics of `wc_deprecated_function` (`_doing_it_wrong` is for misuse, not removal). It also already routes through `_deprecated_function` under the hood and respects `WP_DEBUG`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Same reasoning as on the setter — `wc_deprecated_function()` is the right semantic for a deprecated property, `_doing_it_wrong` would imply misuse rather than removal.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Applied in 84fa00126.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Applied in 9b445cfcf — also fixed a missing `)` after `'__set'` in the suggestion so PHP could parse it.

includes/class-wc-stripe.php
wjrosa: Removed in 9b445cfcf — the class docblock already covers it.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -110,13 +101,15 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 	public $saved_cards;
 
 	/**
-	 * Should SEPA tokens be used for other payment methods (iDEAL and Bancontact)
+	 * Should SEPA tokens be used for other payment methods (iDEAL and Bancontact).
+	 *
+	 * Previously public property that is now accessed via `__get()` after deprecation.
```
daledupreez: Not blocking: we should also reference `__set()`.
```suggestion
	 * Previously public property that is now private for backwards compatibility.
	 * Deprecation notices are triggered via {@see __get()} and {@see __set()}.
```

daledupreez (approved):
Thanks for the cleanup! Let's get this shipped!

PR #5299, opened 2026-04-17
Add WC_Stripe_Payment_Method domain wrapper

wjrosa:
## Summary

- Introduces `WC_Stripe_Payment_Method`, a typed domain wrapper around a Stripe PaymentMethod object (accepts `\Stripe\PaymentMethod` / `\Stripe\StripeObject` from the SDK or `stdClass` from legacy `json_decode`).
- Consolidates the heaviest property-scatter in the codebase — PaymentMethod fields are currently accessed at 40+ call sites across 11 files (payment-tokens layer, UPE payment-method subclasses, UPE gateway, subscriptions compat trait, customer, helper). The wrapper folds those into:
  - **Type predicates** (`is_card`, `is_type`, `is_payment_method_object`) so call sites stop repeating `$pm->type === WC_Stripe_Payment_Methods::CARD` and the `$pm->object === 'payment_method'` check used to disambiguate legacy source objects.
  - **Canonical card-brand fallback chain** — `card.display_brand` → `card.networks.preferred` → `card.brand` — centralized in `get_card_brand()`, replacing the verbatim 3-way null-coalesce copy-pasted in `class-wc-stripe-upe-payment-method-cc.php:97` and `class-wc-stripe-payment-tokens.php:573`.
  - **Typed card accessors** (`last4`, `fingerprint`, `exp_month`, `exp_year`, `country`, `funding`) plus the two business predicates the UPE gateway needs against card country/funding: `is_prepaid_card()` and `is_india_card()`.
  - **Sub-object accessors** for `sepa_debit` (last4, fingerprint), `us_bank_account` (last4, fingerprint, bank_name, account_type) with `has_us_bank_details()` for the ACH payment-token early-return guard, and `link` (email).
  - **Billing-detail accessors** (`get_billing_email` with optional fallback, `get_billing_name`, `get_billing_phone`, `get_billing_address`) that fold in the `?? ''` pattern sprinkled through the payment-tokens and subscriptions code.
  - **Expanded-or-string customer ID** resolution via `get_customer_id()`, and `get_metadata_value()` for consistent null-safety on metadata reads.
- Fail-fast constructor guard (throws `InvalidArgumentException` on unrelated object types) plus a `raw()` escape hatch for incremental migration.

## Why this shape

Same rationale as the prior domain-class PRs (#5276, #5277, #5278, #5298):

- **Type safety at the boundary.** PHPStan can verify property access across all 11 call sites that currently touch `$payment_method->...` directly.
[...]

PR #5298, opened 2026-04-17
Add WC_Stripe_Refund domain wrapper

wjrosa:
## Changes proposed in this Pull Request:

Introduces `WC_Stripe_Refund`, a typed domain wrapper around the `stdClass` produced by `json_decode` of a Stripe Refund response. Consolidates refund scatter currently duplicated across the webhook handler paths into a single, tested seam:

- Status predicates: `is_succeeded`, `is_pending`, `is_failed`, `is_canceled`, `is_requires_action`, plus `is_failed_or_canceled()` for the terminal-failure branch the webhook handler inspects.
- Decimal-aware `get_amount_decimal()` that handles zero-decimal currencies (JPY, KRW, etc.) correctly, with an optional currency override for legacy webhook payloads where the charge currency lives on the notification rather than on the refund itself.
- Expanded-or-string ID resolution for `charge`, `payment_intent`, `balance_transaction`, and `failure_balance_transaction`.
- Reason accessors: `get_reason()` (Stripe code), `get_failure_reason()`, and `get_friendly_failure_reason()` which delegates to `WC_Stripe_Helper::get_refund_reason_description()` for end-user display.
- `get_receipt_number()`, `get_metadata_value()`, and a `raw()` escape hatch for incremental migration.

**Why now.** The `$refund_object->amount / 100` vs. no-decimal-currency branch appears three times in `class-wc-stripe-webhook-handler.php` (`process_webhook_refund`, `process_webhook_refund_updated`, `get_refund_amount`), and the `in_array( $status, ['failed', 'canceled'], true )` predicate is inlined alongside it. Centralising the math and predicates removes the next decimal-currency regression risk and gives PHPStan something to verify across the refund path.

**Scope.** This PR only introduces the class (with tests). The wrapper is autoloaded via `WC_Stripe::init()` but no call site uses it yet — that migration lands in a follow-up so the wrapper can be reviewed in isolation. Candidate call sites for the follow-up:

- `WC_Stripe_Webhook_Handler::process_webhook_refund()` — `includes/class-wc-stripe-webhook-handler.php:740`
- `WC_Stripe_Webhook_Handler::process_webhook_refund_updated()` — `includes/class-wc-stripe-webhook-handler.php:833`
- `WC_Stripe_Webhook_Handler::get_refund_amount()` — `includes/class-wc-stripe-webhook-handler.php:1055`
- The `$refund_object->balance_transaction` access at `includes/class-wc-stripe-webhook-handler.php:557` (partial-capture path)

[...]

PR #5301, opened 2026-04-17
Add WC_Stripe_Refund domain wrapper and migrate webhook + gateway refund paths

wjrosa:
## Summary

- Introduces `WC_Stripe_Refund`, a typed domain wrapper around a Stripe Refund object (the raw `stdClass` returned by `WC_Stripe_API::request()`). No Stripe PHP SDK dependency — the wrapper works directly against `stdClass` payloads from `json_decode()`.
- Migrates the webhook handler and abstract gateway refund paths to use the wrapper:
  - `class-wc-stripe-webhook-handler.php::process_webhook_refund`, `process_webhook_refund_updated`, `get_refund_amount`, and the partial-capture branch of `process_webhook_charge_succeeded` wrap the stdClass response in `WC_Stripe_Refund` once and use typed accessors for the rest of the method.
  - `abstracts/abstract-wc-stripe-payment-gateway.php::process_refund` wraps the refund response in the `! empty( $response->id )` branch and uses `get_id()` / `get_amount()` / `get_balance_transaction_id()`.
- The in-place amount normalization (`! in_array( strtoupper( $currency ), NO_DECIMAL_CURRENCY_CODES ) ? $amount / 100 : $amount`) is collapsed into `$stripe_refund->get_amount_decimal( $order_currency )` — the optional currency override supports the webhook path, which sources the currency from the order rather than the refund.
- The `in_array( $status, [ 'failed', 'canceled' ], true )` predicate is replaced by `$stripe_refund->is_failed_or_canceled()`.
- Balance-transaction and failure-balance-transaction expanded/string access is now typed via `get_balance_transaction_id()` / `get_failure_balance_transaction_id()`.
- PHPStan baseline is refreshed: the migration eliminates 3 pre-existing `Access to an undefined property object::...` warnings in the webhook handler and 1 in the abstract gateway. Two small baseline entries are added to cover the new `WC_Stripe_Refund` constructor type narrowing (the `$response` variable in the gateway is loosely typed as `object` from upstream assignments).

## Why this shape

This is a sibling of #5298 / #5299 (the SDK-based domain wrappers), but targeted at `develop` and independent of the Stripe PHP SDK. Same rationale:

- **Type safety at the boundary.** PHPStan can verify refund property access; the baseline-delta proves the reduction in unknown-property warnings.
- **Consolidates scatter.** The decimal-currency amount normalization repeats 3 times in the webhook handler alone — now it's one `get_amount_decimal()` call.
- **Read-only domain object.** The wrapper does not hold state, does not perform I/O, and does not mutate the order — it just exposes typed getters with consistent null-safety.
- **Backwards-compatible.** `get_refund_object()`'s signature is unchanged; the wrapper is constructed at the call site from the same stdClass.

## Test plan

- [x] Run PHPUnit: `npm run test:php -- --filter=WC_Stripe_Refund_Test|WC_Stripe_Webhook_Handler` — 95 tests pass (31 new + existing webhook handler suite).
- [x] Run PHPStan: `npm run phpstan` — clean after baseline refresh.
- [ ] In a local Docker environment, trigger a refund via the Stripe Dashboard (webhook path) on a zero-decimal-currency order (JPY) and verify the amount is formatted correctly without the off-by-100x bug.
- [ ] In a local Docker environment, trigger a failed refund and verify `wc_stripe_refund_status` / `wc_stripe_refund_failure_reason` order meta are populated with the friendly description.

### DIFF
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -1308,8 +1308,9 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 			);
 
 		} elseif ( ! empty( $response->id ) ) {
+			$stripe_refund    = new WC_Stripe_Refund( $response );
 			$formatted_amount = wc_price(
-				WC_Stripe_Helper::convert_from_stripe_amount( (int) $response->amount, $order->get_currency() ),
+				WC_Stripe_Helper::convert_from_stripe_amount( (int) $stripe_refund->get_amount(), $order->get_currency() ),
 				[ 'currency' => $order->get_currency() ]
 			);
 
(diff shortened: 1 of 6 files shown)


wjrosa:
Postponing this work for now.

PR #5302, opened 2026-04-17
Add WC_Stripe_Setup_Intent domain wrapper and migrate intent-controller + gateway

wjrosa:
## Summary

- Introduces `WC_Stripe_Setup_Intent`, a typed domain wrapper around a Stripe SetupIntent `stdClass` payload. No Stripe PHP SDK dependency.
- Migrates the two hottest paths:
  - **`class-wc-stripe-intent-controller.php`**: `create_setup_intent_ajax` (save-a-card flow), `create_setup_intent_for_payment_method_ajax` (PaymentMethod-type setup), and `create_and_confirm_setup_intent_ajax` (add-payment-method via shortcode/account page) wrap the stdClass response in `WC_Stripe_Setup_Intent` and use typed predicates and accessors.
  - **`abstracts/abstract-wc-stripe-payment-gateway.php`**: `save_intent_to_order`'s `setup_intent` branch wraps `$intent` and uses `get_id()` / `get_mandate_id()` (the latter handles both string and expanded mandate forms). `setup_intent()` (the SCA verification path for APMs that don't require 3DS) also migrates to typed accessors.
- The `! empty( $intent->error )` / `$intent->error->message` pattern becomes `has_error()` / `get_error_message()`.
- The `in_array( $status, SUCCESSFUL_SETUP_INTENT_STATUSES, true )` check becomes `is_successful_for_setup()`.

## Why this shape

Same rationale as #5301 (Refund on develop):

- **Type safety at the boundary.** PHPStan can now verify SetupIntent property access across the intent controller and the gateway; the baseline delta reflects the reduction in unknown-property warnings (`$id`, `$status`, `$client_secret`).
- **Consolidates scatter.** The `SUCCESSFUL_SETUP_INTENT_STATUSES` membership check and the error-pair pattern were each open-coded in 2+ places.
- **Read-only domain object.** No I/O, no state, no order mutation — just typed getters with consistent null-safety.
- **Backwards-compatible.** `create_and_confirm_setup_intent()` and `create_setup_intent()` still return stdClass for external callers; the wrapper is constructed at the call site.

## Test plan

- [x] Run PHPUnit: `npm run test:php -- --filter=WC_Stripe_Setup_Intent_Test|WC_Stripe_Intent_Controller|WC_Stripe_Payment_Gateway_Test` — 104 tests pass (21 new + existing controller/gateway suites).
- [x] Run PHPStan: `npm run phpstan` — clean after baseline refresh (net reduction of ignored-pattern counts as property accesses move into the wrapper).
- [ ] In a local Docker environment, add a new payment method via the My Account → Payment Methods flow and verify 3DS authentication completes.
- [ ] In a local Docker environment, place a free-trial subscription order and confirm the mandate ID is stored on the order.

### DIFF
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -1741,11 +1741,13 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 				$order_helper->update_stripe_mandate_id( $order, $charge->payment_method_details->acss_debit->mandate );
 			}
 		} elseif ( 'setup_intent' === $intent->object ) {
-			$order_helper->update_stripe_setup_intent_id( $order, $intent->id );
+			$setup_intent = new WC_Stripe_Setup_Intent( $intent );
+			$order_helper->update_stripe_setup_intent_id( $order, (string) $setup_intent->get_id() );
 
 			// Add mandate for free trial subscriptions.
-			if ( isset( $intent->mandate ) ) {
-				$order_helper->update_stripe_mandate_id( $order, $intent->mandate );
+			$mandate_id = $setup_intent->get_mandate_id();
+			if ( null !== $mandate_id ) {
+				$order_helper->update_stripe_mandate_id( $order, $mandate_id );
 			}
 		}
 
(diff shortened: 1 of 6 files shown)


wjrosa:
Postponing this work for now.

PR #5303, opened 2026-04-17
Add WC_Stripe_Charge domain wrapper and migrate mandate + radar + webhook paths

wjrosa:
## Summary

- Introduces `WC_Stripe_Charge`, a typed domain wrapper around a Stripe Charge `stdClass` payload. No Stripe PHP SDK dependency.
- Migrates three concrete scatter points:
  - **Mandate extraction** in `abstracts/abstract-wc-stripe-payment-gateway.php::save_intent_to_order`. The two-branch check for `payment_method_details.card.mandate` and `payment_method_details.acss_debit.mandate` folds into a single `$charge->get_mandate_id()` call.
  - **Radar block detection** in `compat/trait-wc-stripe-subscriptions.php::is_charge_blocked_by_radar`. The inline `isset( $charge->outcome->type ) && 'blocked' === ...` / `$charge->outcome->reason` pair becomes `$charge->is_blocked_by_radar()` / `get_radar_block_reason()`.
  - **Webhook `charge.succeeded`** in `class-wc-stripe-webhook-handler.php::process_webhook_charge_succeeded`. The stdClass is wrapped once at the top; `captured`, `balance_transaction` (expanded-or-string), and `id` access all go through typed getters.
- Fixes an incidental test-fixture bug along the way: `WC_Stripe_Webhook_Handler_Test::test_process_webhook_charge_succeeded_skipped_for_synchronous_payment_methods` used to pass `balance_transaction` as an expanded object without an `id` — that shape would never arrive via a real webhook and masked a latent bug where the old code passed the whole object to `update_fees()` (which then tried to concatenate it into a URL). The fixture is updated to a proper string ID.

## Why this shape

Same rationale as #5301 (Refund) and #5302 (SetupIntent) — typed boundary, consolidated scatter, read-only domain object:

- **Mandate extraction** was previously 4 lines of nested `isset()` checks at a single call site; it's the kind of business rule that should live on the domain object.
- **Radar blocked state** is a predicate that applies in more than one place (subscriptions trait today, likely the Amazon Pay / non-subscription paths later) — getting it onto the wrapper sets us up for a consistent treatment.
- **Charge ID resolution** (`$charge->id` vs. `$charge->latest_charge`) and **expanded/string balance_transaction** are both patterns that would otherwise keep appearing in each call site that inspects a Charge.

## Test plan

- [x] Run PHPUnit: `npm run test:php -- --filter=WC_Stripe_Charge_Test|WC_Stripe_Webhook_Handler|WC_Stripe_Subscription_Renewal|WC_Stripe_Payment_Gateway_Test` — 156 tests pass (27 new + existing webhook / gateway / renewal suites).
- [x] Run PHPStan: `npm run phpstan` — clean.
- [ ] In a local Docker environment, complete an Indian-card Subscription renewal and confirm the mandate ID is stored on the order.
- [ ] In a local Docker environment, trigger a Radar-blocked subscription renewal and confirm the radar-block reason appears in the renewal-failure order note.

### DIFF
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -1733,12 +1733,12 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 			// TODO: Refactor and add mandate ID support for other payment methods, if necessary.
 			// The mandate ID is not available for the intent object, so we need to fetch the charge.
 			// Mandate ID is necessary for renewal payments for certain payment methods and Indian cards.
-			$charge = $this->get_latest_charge_from_intent( $intent );
-
-			if ( isset( $charge->payment_method_details->card->mandate ) ) {
-				$order_helper->update_stripe_mandate_id( $order, $charge->payment_method_details->card->mandate );
-			} elseif ( isset( $charge->payment_method_details->acss_debit->mandate ) ) {
-				$order_helper->update_stripe_mandate_id( $order, $charge->payment_method_details->acss_debit->mandate );
+			$latest_charge = $this->get_latest_charge_from_intent( $intent );
+			if ( $latest_charge instanceof stdClass ) {
+				$mandate_id = ( new WC_Stripe_Charge( $latest_charge ) )->get_mandate_id();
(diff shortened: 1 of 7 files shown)


wjrosa:
Postponing this work for now.

PR #5310, opened 2026-04-17
Add merchant setting for ECE on the change-payment-method page

### LINKED ISSUE #5312: Merchant toggle for Express Checkout on the subscription Change Payment Method page
## Context

In review of #5284 (ECE support for the subscription "Change payment method" page), [Dale raised](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5284#pullrequestreview-4124045576) that the change-payment-method page is a distinct context from checkout and merchants should be able to choose whether express checkout methods are visible there — independently of the checkout-page location toggle.

This issue tracks the work to add that merchant-facing toggle.

## Proposal

Extend the existing location-toggle pattern (Product / Cart / Checkout) with a fourth "Change payment method (subscriptions)" option for both:

- `express_checkout_button_locations` — Apple Pay, Google Pay, Link.
- `amazon_pay_button_locations` — Amazon Pay.

Include the new location in the default list so the out-of-the-box behavior matches #5284's intent; merchants can uncheck it if they want the change-payment page to differ from the checkout page.

## Scope

- **Backend** (`includes/admin/stripe-settings.php`): add `change_payment_method` as a valid multiselect option for both ECE and Amazon Pay, and include it in `$wc_stripe_default_express_checkout_locations`.
- **Backend** (`includes/payment-methods/class-wc-stripe-express-checkout-helper.php`): `should_show_ece_on_change_payment_method_page()` should consult the new location instead of piggybacking on `should_show_ece_on_checkout_page()`. `is_enabled_for_current_context()` should route through the new location when `is_change_payment_method_page()` is true. Expose a `wc_stripe_show_express_checkout_on_change_payment_method` filter for programmatic overrides.
- **Frontend**: add a "Change payment method (subscriptions)" checkbox to the Apple/Google Pay express-checkout settings section (`client/entrypoints/express-checkout-settings/express-checkout-settings-section.js`) and the Amazon Pay settings section (`client/entrypoints/amazon-pay-settings/amazon-pay-settings-section.js`), wired to the existing `useExpressCheckoutLocations` / `useAmazonPayLocations` hooks.
- **REST**: the settings controller auto-derives the allowed enum from the form-fields options, so extending the multiselect config is sufficient — no validator changes required.

## Implementation

PR: #5310 (branched from `fix/4792-ece-change-payment-method`, will rebase onto `develop` once #5284 merges).

wjrosa:
Fixes STRIPE-1087
Fixes #5312

Follow-up to #5284 addressing [Dale's review feedback](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5284#pullrequestreview-4124045576): the subscription "Change payment method" page is a distinct context from checkout, and merchants should be able to opt in or out independently of the checkout-page toggle.

**What changes**

- Adds a `change_payment_method` location option to `express_checkout_button_locations` (Apple Pay / Google Pay / Link). Only registered/defaulted when WC Subscriptions is active — without Subscriptions the option doesn't appear in the multiselect, isn't part of the default, and the ECE settings UI hides the checkbox via a new `is_subscriptions_active` flag in `wc_stripe_express_checkout_settings_params`.
- Adds a "Change payment method (subscriptions)" checkbox to the Apple/Google Pay express-checkout settings section, wired to the same hooks as the existing checkbox list.
- Decouples the change-payment check from the checkout toggle — `WC_Stripe_Express_Checkout_Helper::should_show_ece_on_change_payment_method_page()` consults a dedicated `should_show_ece_on_change_payment_method_location()` method (which exposes the `wc_stripe_show_express_checkout_on_change_payment_method` filter), rather than piggybacking on `should_show_ece_on_checkout_page()`.
- Routes `is_enabled_for_current_context()` through the new location when on the change-payment page, so per-type rendering and the JS `is_change_payment_method` flag both respect the toggle.
- Adds a one-shot migration `WC_Stripe_Express_Checkout_Add_Change_Payment_Method_Location_Update` that opts in existing merchants who had every previous default location (product/cart/checkout) enabled — preserving the pre-PR "everything on" state without re-enabling the new location for merchants who later disable it. The migration is gated on `WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled()` and deliberately does **not** set its one-shot flag when Subscriptions isn't installed, so the next plugin update will run it if the merchant installs Subscriptions later.

> [!NOTE]
> Amazon Pay does not support the subscription change-payment flow, so the toggle is intentionally not added to the Amazon Pay settings or `amazon_pay_button_locations`.

The REST settings controller auto-derives the allowed enum from the form-fields options array, so extending the multiselect config with the new option extends the schema without additional validator changes.

**Why this shape**

Matching the existing location-toggle pattern (checkout / product / cart) means merchants get a familiar UI and no new concepts. Independently from this change, the change-payment context already had its own helper (`is_change_payment_method_page`) and render path, so the new toggle slots in cleanly without refactoring existing flows.

**How can this code break?**

- *A merchant who customized `express_checkout_button_locations` before this PR landed* won't get `change_payment_method` from the migration (it only acts on the unmodified default set) and will have to toggle it on themselves. The toggle UI and validator changes make that a one-click change.
- *Mocked unit tests for `should_show_express_checkout_button_change_payment` had to be re-wired* to mock `should_show_ece_on_change_payment_method_location` instead of `should_show_ece_on_checkout_page`, since the gate no longer calls the latter on the change-payment branch.

- `WC_Stripe_Express_Checkout_Helper_Test` covers the new gate path with the merchant-toggle mock.
- `WC_Stripe_Express_Checkout_Add_Change_Payment_Method_Location_Update_Test` covers the migration's five branches (all-defaults → append, partial set → skip, already-present → skip, flag-set → skip, non-array → skip) plus the new `subscriptions-not-installed → skip and leave flag unset` case.
- `WC_REST_Stripe_Settings_Controller_Test` exercises the extended enum.
- Jest `express-checkout-settings-locations` suite covers the checkbox wiring.

## Testing instructions

> Assume the tester has admin access to a WooCommerce store with this plugin active and does not have intimate knowledge of Stripe internals.

1. Run `npm run test:php -- --filter='WC_Stripe_Express_Checkout|WC_REST_Stripe_Settings_Controller|WC_Stripe_Express_Checkout_Add_Change_Payment_Method_Location_Update'` — all tests pass.
2. Run `npm run phpstan` — no new errors.
3. Run `npm run test:js -- --testPathPattern='express-checkout-settings-locations'` — all tests pass.
4. With WC Subscriptions **active**, navigate to **WooCommerce → Stripe → Customize Express Checkouts**. Confirm a new "Change payment method (subscriptions)" checkbox is visible alongside Checkout / Product page / Cart.
5. **Express Checkout off-toggle**: uncheck "Change payment method (subscriptions)" and save. On a subscription view page, click "Change payment method" and confirm Apple Pay / Google Pay / Link buttons do NOT render. Re-check it, save, and confirm the buttons reappear.
6. **Programmatic override**: add `add_filter( 'wc_stripe_show_express_checkout_on_change_payment_method', '__return_false' );` in a mu-plugin and confirm the buttons hide regardless of the merchant setting.
7. **Subscriptions not installed**: deactivate WC Subscriptions and reload the Customize Express Checkouts page. Confirm the "Change payment method (subscriptions)" checkbox no longer appears and the `change_payment_method` option is absent from the multiselect dropdown on the main Stripe settings page. Verify the option `wc_stripe_express_checkout_cpm_location_migrated` is **not** in the database (e.g. via `wp option get wc_stripe_express_checkout_cpm_location_migrated`), so the migration will run once Subscriptions is installed.

---

-   [x] Tested on mobile (or does not apply) — admin-only setting; no checkout-flow change.

### Changelog entry

* Add - Add a setting to control whether Express Checkout is shown on the WooCommerce Subscriptions change payment method page

<details>

#### Comment

n/a — entry included above.

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -670,13 +672,36 @@ private function should_show_ece_on_change_payment_method_page() {
...
 		}
 
 		return true;
 	}
 
+	/**
+	 * Returns true if any express checkout buttons are enabled for the subscription
+	 * change payment method location, false otherwise.
+	 *
+	 * @since 10.6.0
```
diegocurbelo: Should be `10.7.0`

```suggestion
	 * @since 10.7.0
```

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -670,13 +672,36 @@ private function should_show_ece_on_change_payment_method_page() {
...
+	 * @return boolean
+	 */
+	public function should_show_ece_on_change_payment_method_location() {
+		$should_show = $this->should_show_ece_on_location( 'change_payment_method' );
+
+		/**
+		 * Filters whether Express Checkout buttons should appear on the
+		 * WooCommerce Subscriptions "Change payment method" page.
+		 *
+		 * @since 10.6.0
```
diegocurbelo:
```suggestion
		 * @since 10.7.0
```

diegocurbelo (commented):
The code changes look good, and it tests well.

Should we include a migration for this?
I had all 3 locations enabled before this PR; now they are still enabled, but the new option it's not 🤷🏼 

Image

wjrosa:
@diegocurbelo Good call — added the migration in a513e404e. `Express_Checkout_Add_Change_Payment_Method_Location_Update` runs once on `woocommerce_stripe_updated` and appends `change_payment_method` only when the stored locations include all three pre-PR defaults (product/cart/checkout) and don't already have the new key. Gated by a one-shot `wc_stripe_express_checkout_cpm_location_migrated` flag so a merchant who later removes the new location doesn't get it re-added on the next version bump. Tests cover the four interesting branches + a defensive non-array case.

client/entrypoints/amazon-pay-settings/amazon-pay-settings-section.js
```diff
@@ -174,6 +174,24 @@ const AmazonPaySettingsSection = () => {
...
+					<li>
+						<CheckboxControl
+							disabled={ ! isAmazonPayEnabled }
+							checked={
+								isAmazonPayEnabled &&
+								amazonPayLocations.includes(
+									'change_payment_method'
+								)
+							}
+							onChange={ makeLocationChangeHandler(
```
daledupreez: Not blocking: is it worth memoizing the result of the function to avoid unnecessary re-renders? `useCallback()` could be relevant. (This applies in both pages.)

includes/admin/stripe-settings.php
```diff
@@ -197,9 +197,10 @@
 			'desc_tip'          => true,
 			'class'             => 'wc-enhanced-select',
 			'options'           => [
-				'product'  => __( 'Product', 'woocommerce-gateway-stripe' ),
-				'cart'     => __( 'Cart', 'woocommerce-gateway-stripe' ),
-				'checkout' => __( 'Checkout', 'woocommerce-gateway-stripe' ),
+				'product'               => __( 'Product', 'woocommerce-gateway-stripe' ),
+				'cart'                  => __( 'Cart', 'woocommerce-gateway-stripe' ),
+				'checkout'              => __( 'Checkout', 'woocommerce-gateway-stripe' ),
+				'change_payment_method' => __( 'Change payment method (subscriptions)', 'woocommerce-gateway-stripe' ),
```
daledupreez: Should we only add this if Subscriptions is available?

includes/admin/stripe-settings.php
```diff
@@ -5,7 +5,7 @@
 
 $is_gte_wc6_6 = defined( WC_VERSION ) && version_compare( WC_VERSION, '6.6', '>=' );
 
-$wc_stripe_default_express_checkout_locations = [ 'product', 'cart', 'checkout' ];
+$wc_stripe_default_express_checkout_locations = [ 'product', 'cart', 'checkout', 'change_payment_method' ];
```
daledupreez: It feels like we should only default this on when Subscriptions is enabled.

includes/migrations/class-express-checkout-add-change-payment-method-location-update.php
```diff
@@ -0,0 +1,89 @@
...
+ * location enabled into the new `change_payment_method` location, so the
+ * upgrade preserves the "everything on" state instead of silently leaving
+ * the new location off.
+ *
+ * Runs once. Merchants who customized their location set (i.e. didn't have
+ * all three of product/cart/checkout enabled) are left alone.
+ *
+ * @since 10.7.0
+ */
+class Express_Checkout_Add_Change_Payment_Method_Location_Update {
```
daledupreez: We should always use some sort of prefix or namespace to ensure we don't collide with other code.
```suggestion
class WC_Stripe_Express_Checkout_Add_Change_Payment_Method_Location_Update {
```

includes/migrations/class-express-checkout-add-change-payment-method-location-update.php
```diff
@@ -0,0 +1,89 @@
...
+	/**
+	 * Option flag used to ensure the migration only runs once. Reading and
+	 * writing a top-level option (rather than a gateway sub-option) keeps the
+	 * gate independent of the settings array we mutate.
+	 */
+	const MIGRATION_FLAG_OPTION = 'wc_stripe_express_checkout_cpm_location_migrated';
```
daledupreez: The visibility for this constant should be explicit. I would prefer private or protected, but we could also make it public if that's truly needed (though I don't think it is).

includes/migrations/class-wc-stripe-express-checkout-add-change-payment-method-location-update.php
```diff
@@ -0,0 +1,89 @@
...
+	 */
+	public function maybe_migrate() {
+		// One-shot guard: never run a second time, so a merchant who later
+		// removes the new location doesn't have it added back on the next
+		// version bump.
+		if ( 'yes' === get_option( self::MIGRATION_FLAG_OPTION ) ) {
+			return;
+		}
+
+		$gateway   = $this->get_gateway();
```
daledupreez: High level question: should we run the migration if Subscriptions is not installed? I feel that would be more confusing to merchants who don't have it installed.

includes/migrations/class-wc-stripe-express-checkout-add-change-payment-method-location-update.php
```diff
@@ -0,0 +1,89 @@
...
+		$locations = $gateway->get_option( 'express_checkout_button_locations' );
+
+		// Treat anything other than an array as "not the all-three default" —
+		// don't touch it.
```
daledupreez, line 83: Why are we only adding the new value if all three are still using their default values?

includes/migrations/class-express-checkout-add-change-payment-method-location-update.php
```diff
@@ -0,0 +1,89 @@
...
+
+		update_option( self::MIGRATION_FLAG_OPTION, 'yes' );
+	}
+
+	/**
+	 * Returns the main Stripe payment gateway.
+	 *
+	 * @return WC_Stripe_Payment_Gateway
+	 */
+	public function get_gateway() {
```
daledupreez: Can we inline this function? We're only calling it from one location, so it doesn't feel necessary to create a new public function for it.

includes/class-wc-stripe.php
```diff
@@ -239,6 +239,7 @@ public function init() {
 			new Allowed_Payment_Request_Button_Types_Update();
 			new Migrate_Payment_Request_Data_To_Express_Checkout_Data();
 			new Sepa_Tokens_For_Other_Methods_Settings_Update();
+			new Express_Checkout_Add_Change_Payment_Method_Location_Update();
```
daledupreez: I am realising after my changes this week to add the `self::$instance === $this` checks that I _really_ dislike this pattern of creating separate object instances for every request that hook themselves into actions via their constructors and linger in memory regardless of context. It also means that we need to look at every migration file to understand what the code is doing and how that code is hooked in.
:endrant:

Not a blocker for this PR, but something I may try to restructure/refactor soon.

tests/phpunit/admin/migrations/class-express-checkout-add-change-payment-method-location-update-test.php
```diff
@@ -0,0 +1,141 @@
...
+	public function set_up() {
+		parent::set_up();
+
+		delete_option( Express_Checkout_Add_Change_Payment_Method_Location_Update::MIGRATION_FLAG_OPTION );
+
+		$this->gateway_mock = $this->getMockBuilder( WC_Stripe_UPE_Payment_Gateway::class )
+									->disableOriginalConstructor()
+									->getMock();
+
+		$this->migration = $this->getMockBuilder( Express_Checkout_Add_Change_Payment_Method_Location_Update::class )
```
daledupreez: Not blocking, but I have a pattern question here: it doesn't feel right to me to have instance properties that could be shared across methods but aren't only because they get created in `set_up()` before each test run. I think it's cleaner to have a helper method for each object type so it is clearer that the mock is fresh in all cases. But that's not a blocker in any way.

tests/phpunit/admin/migrations/class-express-checkout-add-change-payment-method-location-update-test.php
```diff
@@ -0,0 +1,141 @@
...
+								->disableOriginalConstructor()
+								->setMethods( [ 'get_gateway' ] )
```
daledupreez: Nit: `setMethods()` is deprecated in PHPUnit 9 and has been removed in newer versions of PHPUnit, we should use `onlyMethods()`.
```suggestion
								->onlyMethods( [ 'get_gateway' ] )
```

client/entrypoints/amazon-pay-settings/amazon-pay-settings-section.js
```diff
@@ -174,6 +174,24 @@ const AmazonPaySettingsSection = () => {
 							label={ __( 'Cart', 'woocommerce-gateway-stripe' ) }
 						/>
 					</li>
+					<li>
+						<CheckboxControl
```
daledupreez: Is Amazon Pay supported? I thought it was not based on the main PR.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.8.0
+* Add - Add a "Change payment method (subscriptions)" visibility toggle to the Express Checkout and Amazon Pay settings so merchants can opt in or out of showing the buttons on the subscription change-payment screen independently of the checkout-page toggle
```
daledupreez: This description is _really_ long. We should shorten it. Also, I believe Amazon Pay is not available for this flow.
```suggestion
* Add - Add a setting to control whether Express Checkout is shown on the WooCommerce Subscriptions change payment method page
```

wjrosa: Applied your suggestion. Also trimmed the matching line in `readme.txt` and dropped Amazon Pay from this PR entirely (see thread on the Amazon Pay checkbox).

client/entrypoints/amazon-pay-settings/amazon-pay-settings-section.js
wjrosa: Good catch — Amazon Pay isn't supported on the change-payment-method page. Removed the checkbox here and dropped `change_payment_method` from `amazon_pay_button_locations` (options + default) in `stripe-settings.php`.

client/entrypoints/amazon-pay-settings/amazon-pay-settings-section.js
wjrosa: Leaving as-is for now. The closures are only consumed by unmemoized `CheckboxControl` children, so memoizing them doesn't actually save re-renders today. Happy to revisit if/when we wrap these in `React.memo`.

includes/admin/stripe-settings.php
wjrosa: Done. The `change_payment_method` option is only added to `express_checkout_button_locations` when `WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled()` is true, and the ECE settings UI hides the checkbox via a new `is_subscriptions_active` flag in `wc_stripe_express_checkout_settings_params`.

includes/admin/stripe-settings.php
wjrosa: Done — when Subscriptions isn't enabled, the default is now `[ 'product', 'cart', 'checkout' ]` and `change_payment_method` is appended only when it is.

includes/migrations/class-express-checkout-add-change-payment-method-location-update.php
wjrosa: Renamed both the class and the file to `WC_Stripe_Express_Checkout_Add_Change_Payment_Method_Location_Update`. Updated the composer classmap, the `new …` site in `class-wc-stripe.php`, and the corresponding test file.

includes/migrations/class-express-checkout-add-change-payment-method-location-update.php
wjrosa: Made it `private`.

includes/migrations/class-wc-stripe-express-checkout-add-change-payment-method-location-update.php
wjrosa: Agreed. The migration now early-returns when Subscriptions isn't enabled and deliberately does **not** set the one-shot flag, so if the merchant installs Subscriptions later, the next `woocommerce_stripe_updated` firing will run the migration. Wrapped the check in a protected `is_subscriptions_enabled()` seam so the test can stub it without juggling the global `WC_Subscriptions` class. New test case: `test_skips_and_does_not_set_flag_when_subscriptions_not_enabled`.

includes/migrations/class-wc-stripe-express-checkout-add-change-payment-method-location-update.php
wjrosa: Design intent: opt in merchants who had the equivalent of "everything on" so the new location lights up the same way without surprising those who'd deliberately disabled cart/product/checkout. If any of the three is off, the location set is treated as a customization and left alone.

includes/migrations/class-express-checkout-add-change-payment-method-location-update.php
wjrosa: Kept it as a method but switched it to `protected` — it exists as a test seam for the partial mock so we don't have to wire `woocommerce_gateway_stripe()` in the unit test. Now that there's a second seam (`is_subscriptions_enabled()`), inlining one of them and not the other felt inconsistent.

includes/class-wc-stripe.php
wjrosa: Noted — agreed this pattern has aged poorly. Happy to follow your refactor when it lands.

tests/phpunit/admin/migrations/class-express-checkout-add-change-payment-method-location-update-test.php
wjrosa: Refactored: replaced the shared `$migration` instance property with a `build_migration( $subscriptions_enabled )` helper, so every test reads "this is a fresh mock built for these inputs" at the call site. Kept the shared `$gateway_mock` since every test wants the same contract against the gateway.

tests/phpunit/admin/migrations/class-express-checkout-add-change-payment-method-location-update-test.php
wjrosa: Done — switched to `onlyMethods()`.

includes/admin/stripe-settings.php
```diff
@@ -5,8 +5,21 @@
...
+
+$wc_stripe_express_checkout_location_options = [
+	'product'  => __( 'Product', 'woocommerce-gateway-stripe' ),
+	'cart'     => __( 'Cart', 'woocommerce-gateway-stripe' ),
+	'checkout' => __( 'Checkout', 'woocommerce-gateway-stripe' ),
+];
+
 $wc_stripe_default_express_checkout_locations = [ 'product', 'cart', 'checkout' ];
 
+if ( $wc_stripe_subscriptions_enabled ) {
```
daledupreez: Not blocking, but this code could run in a global scope, so it may be worth minimising the number of "local" variables.
```suggestion
if ( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled()) {
```

client/entrypoints/express-checkout-settings/express-checkout-settings-section.js
```diff
@@ -226,6 +227,29 @@ const ExpressCheckoutSettingsSection = () => {
...
+										isExpressCheckoutEnabled &&
+										expressCheckoutLocations.includes(
+											'change_payment_method'
+										)
+									}
+									onChange={ makeLocationChangeHandler(
+										'change_payment_method'
+									) }
+									label={ __(
+										'Change payment method (subscriptions)',
```
daledupreez: I think we should be explicit about this being for WooCommerce Subscriptions, and not other subscription options.
```suggestion
										'Change payment method for WooCommerce Subscriptions',
```

includes/migrations/class-wc-stripe-express-checkout-add-change-payment-method-location-update.php
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * Opt merchants who already had every previous express-checkout button
+ * location enabled into the new `change_payment_method` location, so the
+ * upgrade preserves the "everything on" state instead of silently leaving
+ * the new location off.
+ *
+ * Runs once. Merchants who customized their location set (i.e. didn't have
+ * all three of product/cart/checkout enabled) are left alone.
+ *
+ * @since 10.7.0
```
daledupreez: Nit: Should be 10.8.0
```suggestion
 * @since 10.8.0
```

daledupreez (approved):
I have one comment about the wording in the UI, but I think this is ready to ship otherwise.

PR #5313, opened 2026-04-17
Align agentic commerce sync UI with WP design system components

wjrosa:
Child PR of #5260. Addresses @elizaan36's latest review feedback on the agentic commerce sync status UI.

### UI swapped to WPDS primitives

| Before | After |
|---|---|
| Custom `Card` / `CardTitle` emotion-styled components (`styled.js`) | `Card` + `CardHeader` + `CardBody` from `@wordpress/components` |
| Ad-hoc `StatusBadge` emotion span with hand-rolled tone colors | Shared repo `Pill` primitive (already used by `PaymentMethodUnavailablePill`) with tone modifier classes |
| Unicode emoji icons: ✓ ✗ ⚠ ⏳ ? | Real `@wordpress/icons` — `check`, `close`, `warning`, `pending`, `help` |
| Flex `Actions` emotion div | `Flex` from `@wordpress/components` |
| Raw `<a>` to Stripe Dashboard | `ExternalLink` from `@wordpress/components` |

The WordPress `Badge` component is not available in the version of `@wordpress/components` this repo depends on (`^25.0.0`; Badge landed in a later release). Using the repo's shared `Pill` primitive is the closest equivalent and matches the pattern already used elsewhere in the Stripe extension — which is the alignment @elizaan36 asked for.

`styled.js` (Card / CardTitle / Actions emotion components) is deleted — nothing imports it after this change.

### Content + layout tweaks

- **Sentence case** status labels throughout: "Partial success", "Creating records", "Sync now", "View logs", "Product feed status", "Recent syncs", "Last sync", "Products synced", "Last sync error:".
- **Widen `.col-status` from 160px → 200px** so "Creating records" fits on one line (Diego's request).

### Not in this PR

- No backend / REST API / options changes. The history row cap is owned by the backend (`SYNC_HISTORY_LIMIT`) and is unchanged.

Preview:
Screenshot 2026-05-12 at 14 21 06

## Test plan

- [ ] Open **WooCommerce → Settings → Payments → Stripe → Settings** with the feature flag enabled.
- [ ] Toggle "Enable agentic commerce" on; verify the Product feed status and Recent syncs panels use WP `Card` / `CardHeader` / `CardBody` chrome (matches the rest of the settings page).
- [ ] Hover/click the status pills and verify the icons are proper WP Dashicons (not unicode).
- [ ] Run through each status state ("Success", "Processing", "Creating records", "Queued", "Validating", "Failed", "Partial success", "Unknown") — all should fit one line in the Status column at normal admin widths.
- [ ] Verify "Sync now" / "View logs" buttons still function.
- [ ] Run `npm run lint:js` — clean.
- [ ] Run `npm run test:js -- client/settings/agentic-commerce/` — 28/28 pass.
- [ ] Run `npm run build:webpack` — builds cleanly.

cc @elizaan36


client/settings/agentic-commerce/sync-status.js
```diff
@@ -1,44 +1,27 @@
...
-import { Card, CardTitle, Actions } from './styled';
+import { check, close, help, info, pending, warning } from '@wordpress/icons';
 import apiFetch from '@wordpress/api-fetch';
-import { __, _n, sprintf } from '@wordpress/i18n';
-import { Button, Notice } from '@wordpress/components';
+import { __, sprintf } from '@wordpress/i18n';
+import {
+	Button,
+	Card,
+	CardBody,
```
daledupreez: Note: we have a local override of `CardBody` in `client/settings/card-body.js`. It may be better to use the same component for now so we stay consistent and don't have diverging styles/implementations.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -120,54 +103,138 @@ const HistoryTable = styled.table`
...
+		outline-offset: 1px;
+		border-radius: 2px;
+	}
+
+	svg {
+		fill: currentColor;
+	}
+`;
+
+const getStatusConfig = ( status ) => {
```
daledupreez, line 204: Not blocking, but it may be worth adding JSDoc for the return structure.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -120,54 +103,138 @@ const HistoryTable = styled.table`
...
-		label: __( 'Failed', 'woocommerce-gateway-stripe' ),
-		className: 'error',
-		icon: '✗',
-	},
-	succeeded_with_errors: {
-		label: __( 'Partial Success', 'woocommerce-gateway-stripe' ),
-		className: 'warning',
-		icon: '⚠',
-	},
+const StatusPill = styled( Pill )`
```
daledupreez, line 105: Not blocking, but might it be worth pushing this into `client/components/` or `client/settings/` so it's easier to re-use? We can also do it later, hence this not being blocking.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -182,21 +249,19 @@ const humanTimeDiff = ( timestamp ) => {
...
 	}
 	const h = Math.floor( diffSec / 3600 );
+	if ( h === 1 ) return __( '1 hour ago', 'woocommerce-gateway-stripe' );
 	return sprintf(
 		/* translators: %d: number of hours */
-		_n( '%d hour ago', '%d hours ago', h, 'woocommerce-gateway-stripe' ),
+		__( '%d hours ago', 'woocommerce-gateway-stripe' ),
 		h
 	);
 };
```
daledupreez: These overrides should be reverted to the `_n()` calls from upstream. (I can't suggest that cleanly via the UI.)

client/settings/agentic-commerce/sync-status.js
```diff
@@ -265,7 +325,7 @@ const AgenticCommerceSyncStatus = () => {
 
 	const { last_sync: lastSync, history, next_sync: nextSync } = data ?? {};
 
-	const computeNextSyncLabel = () => {
+	const nextSyncLabel = () => {
```
daledupreez: Nit: it feels like the upstream method name is clearer. Also, it feels like it would be worth wrapping the computation in `useMemo()` or `useCallback()`.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -274,18 +334,21 @@ const AgenticCommerceSyncStatus = () => {
...
 			/* translators: %d: number of minutes until next sync */
-			_n(
-				'Next automatic sync: in %d minute.',
+			__(
 				'Next automatic sync: in %d minutes.',
-				minutes,
 				'woocommerce-gateway-stripe'
 			),
 			minutes
 		);
```
daledupreez: This code block should also be reverted to the `_n()` logic from upstream.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
 					'Monitors the product feed sync status for the agentic commerce integration.',
 					'woocommerce-gateway-stripe'
 				) }{ ' ' }
-				<a
-					href={ importSetsUrl }
-					target="_blank"
-					rel="noopener noreferrer"
-				>
+				<ExternalLink href="https://dashboard.stripe.com/data-management/import-sets">
```
daledupreez: Question: is there a reason we're not using test links as we were before?

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
-						'woocommerce-gateway-stripe'
-					) }
-				</CardTitle>
-
-				{ isLoading && (
-					<p>{ __( 'Loading…', 'woocommerce-gateway-stripe' ) }</p>
-				) }
-				{ ! isLoading && ! hasError && ! lastSync && (
-					<p>
+			<Card style={ { marginBottom: '20px' } }>
```
daledupreez: This custom style feels wrong -- we should be handling that in some way that isn't specific to this component.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
+				<CardHeader>
+					<h2 style={ { margin: 0, fontSize: '14px' } }>
```
daledupreez: As above, inline styles don't feel right here.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
-									{ __(
-										'Import ID',
+								</tbody>
+							</DetailsTable>
+
+							{ nextSyncLabel() && (
+								<p className="description">
+									{ nextSyncLabel() }
+								</p>
+							) }
```
daledupreez: Two nits here:
* I think we changed the upstream code to store the return value from the function in a local variable to avoid calling `nextSyncLabel()` twice
* I think we should be extremely wary of using a very generic class name like `description`. If there is a specific reason for it, we should add a comment as to why.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
+						</Button>
+					</Flex>
+				</CardBody>
+			</Card>
+
+			<Card style={ { marginBottom: '20px' } }>
+				<CardHeader>
+					<h2 style={ { margin: 0, fontSize: '14px' } }>
+						{ __( 'Recent syncs', 'woocommerce-gateway-stripe' ) }
+					</h2>
```
daledupreez: As above, these inline styles should be shifted out of this location so they can be shared.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
+					{ ! isLoading && ! hasError && ! history?.length && (
+						<p>
+							{ __(
+								'No sync history available.',
+								'woocommerce-gateway-stripe'
+							) }
+						</p>
+					) }
+					{ ! isLoading && !! history?.length && (
+						<HistoryTable>
```
daledupreez, line 593: Question: is it worth using a `DataView` for this instead of a custom table?

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
+											'woocommerce-gateway-stripe'
+										) }
+									</th>
+									<th className="col-import-id">
+										{ __(
+											'Import ID',
+											'woocommerce-gateway-stripe'
 										) }
-									</td>
+									</th>
```
daledupreez: I think we should use `wc-stripe-`-prefixed class names to avoid style collisions. (This also applies to the `td` elements in the body below.)
```suggestion
									<th className="wc-stripe-agentic-sync-col-timestamp">
										{ __(
											'Timestamp',
											'woocommerce-gateway-stripe'
) }
									</th>
									<th className="wc-stripe-agentic-sync-col-products">
										{ __(
											'Products',
											'woocommerce-gateway-stripe'
										) }
									</th>
									<th className="wc-stripe-agentic-sync-col-status">
										{ __(
											'Status',
											'woocommerce-gateway-stripe'
										) }
									</th>
									<th className="wc-stripe-agentic-sync-col-import-id">
										{ __(
											'Import ID',
											'woocommerce-gateway-stripe'
) }
									</th>
```

client/settings/agentic-commerce/sync-status.js
```diff
@@ -1,44 +1,27 @@
...
-		color: #0c5460;
-	}
-	&.unknown {
-		background: #e2e3e5;
-		color: #383d41;
-	}
-`;
+import Pill from 'wcstripe/components/pill';
+
+const HISTORY_ROW_LIMIT = 5;
```
daledupreez: Is it worth making this some sort of control or dropdown rather than showing only 5 rows in all cases?

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
+											</td>
+											<td className="col-status">
+												<SyncStatusBadge
+													status={ entry.status }
+												/>
+												{ entry.error && (
+													<Tooltip
+														text={ entry.error }
+													>
+														<ErrorInfoButton
```
daledupreez: This button doesn't have an event handler or a target href. That feels like a bug.

client/settings/agentic-commerce/sync-status.js
wjrosa: Done in d791a8861.

client/settings/agentic-commerce/sync-status.js
wjrosa: Added in d791a8861.

client/settings/agentic-commerce/sync-status.js
wjrosa: Reverted in d791a8861.

client/settings/agentic-commerce/sync-status.js
wjrosa: Renamed + cached in d791a8861. Skipped `useMemo` since the body reads `Date.now()` — memoising on `nextSync` alone would freeze the countdown.

client/settings/agentic-commerce/sync-status.js
wjrosa: Reverted in d791a8861.

client/settings/agentic-commerce/sync-status.js
wjrosa: Restored in d791a8861 — oversight in the WPDS pass.

client/settings/agentic-commerce/sync-status.js
wjrosa: Done in d791a8861 — both Cards now share a `SectionCard = styled( Card )`.

client/settings/agentic-commerce/sync-status.js
wjrosa: Same — moved into a shared `SectionHeading = styled.h2`.

client/settings/agentic-commerce/sync-status.js
wjrosa: Both done in d791a8861: cached in a local `nextSyncLabel` const, and renamed the `.description` p to a co-located `NextSyncNote` styled component.

client/settings/agentic-commerce/sync-status.js
wjrosa: Folded into the shared `SectionCard`/`SectionHeading` styled components.

client/settings/agentic-commerce/sync-status.js
wjrosa: Adopted the suggested prefix in d791a8861.

client/settings/agentic-commerce/sync-status.js
wjrosa: Good catch — replaced with a focusable `<span tabIndex={0} role="img" aria-label="Sync error: …">` in d791a8861. Tooltip is the only action, so a button was misleading.

client/settings/agentic-commerce/sync-status.js
wjrosa: Server caps to 5 (commit 67a26f7). Happy to add a 'show more' if we want deeper history, but I'd lift the server cap first.

client/settings/agentic-commerce/sync-status.js
wjrosa: Agreed — leaving co-located for now since this is the only consumer.

client/settings/agentic-commerce/sync-status.js
wjrosa: Considered it — `DataView` is overkill for a fixed 4-col / 5-row card. Happy to revisit if scope grows.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -182,21 +249,19 @@ const humanTimeDiff = ( timestamp ) => {
...
+			__( '%d minutes ago', 'woocommerce-gateway-stripe' ),
 			m
 		);
 	}
 	const h = Math.floor( diffSec / 3600 );
+	if ( h === 1 ) return __( '1 hour ago', 'woocommerce-gateway-stripe' );
 	return sprintf(
 		/* translators: %d: number of hours */
-		_n( '%d hour ago', '%d hours ago', h, 'woocommerce-gateway-stripe' ),
+		__( '%d hours ago', 'woocommerce-gateway-stripe' ),
```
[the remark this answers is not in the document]
wjrosa: Restored both branches in d791a8861.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
+										<tr>
+											<th>
+												{ __(
+													'Products synced',
+													'woocommerce-gateway-stripe'
+												) }
+											</th>
+											<td>
+												{ lastSync.products.toLocaleString() }
+											</td>
```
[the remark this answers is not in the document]
wjrosa: Tightened to `typeof === 'number'` for both occurrences in d791a8861 (covers null + undefined, plays nice with `eqeqeq`).

client/settings/agentic-commerce/sync-status.js
```diff
@@ -294,249 +357,285 @@ const AgenticCommerceSyncStatus = () => {
...
 								</tr>
-							) ) }
-						</tbody>
-					</HistoryTable>
-				) }
+							</thead>
+							<tbody>
+								{ history
+									.slice( 0, HISTORY_ROW_LIMIT )
+									.map( ( entry, i ) => (
```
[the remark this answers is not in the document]
wjrosa: Removed in d791a8861 — server already caps at 5 (commit 67a26f7).

client/settings/agentic-commerce/sync-status.js
```diff
@@ -228,6 +256,23 @@ const getStatusConfig = ( status ) => {
...
 
+/**
+ * Statuses that indicate Stripe is still processing the ImportSet. While the
+ * latest entry is in any of these states, the dashboard polls so the badge
+ * flips to a terminal status without requiring a manual reload. Mirrors the
+ * server-side REFRESHABLE_STATUSES list in the REST controller.
+ */
+const NON_TERMINAL_STATUSES = [
+	'queued',
+	'validating',
```
daledupreez, line 271: Question: is this a status we want to keep?

client/settings/agentic-commerce/sync-status.js
```diff
@@ -80,82 +62,198 @@ const HistoryTable = styled.table`
...
+		fill: currentColor;
+	}
+`;
+
+const IntroDescription = styled.p`
+	margin-top: 16px;
+`;
+
+const SectionCard = styled( Card )`
+	margin-bottom: 20px;
```
daledupreez, line 176: Nit: The margin below the final card ends up being 44px, as we get this 20px bottom margin for the "Recent syncs" card (or whichever card is shown in last position) as well as the 24px bottom margin from the outer `SettingsSection` component (which technically comes from the `SettingsSectionWrapper` inline item. It would be better if the bottom margin did not apply to the last child.

But maybe we need a follow-up PR to tidy that aspect up.

daledupreez (approved):
Changes are looking good to me overall, though I do have some smaller comments that may drive some tweaks.

client/settings/agentic-commerce/sync-status.js
wjrosa: I think it is safer to keep those for now. Stripe has been actively changing their status strings (that's why we have multiple versions). We can revisit this in the future

client/settings/agentic-commerce/sync-status.js
wjrosa: Applied here rather than in a follow-up — added `&:last-child { margin-bottom: 0; }` to `SectionCard` so the gap collapses to the wrapper's 24px (or 0 when the section is itself last). Mirrors the same rule already on `SettingsSectionWrapper` and `DescriptionWrapper`.

PR #42, opened 2016-08-03
Remove renewal payment retry

thenbrent:
Fixes #37

#### Changes proposed in this Pull Request:

This patch removes the renewal payment retry system introduced with SHA: fb1cdf410 which would delete the card and potentially the customer tokens from a renewal order and use the tokens on the customer's account instead. This lead to issues outlined in #37.

It also sneaks in a typo fix with SHA: 90ff54830

-------------------
- [x] Make sure your changes respect [WordPress' coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/).
- [x] Did you make changes, or create a **new .js file**? If **Gruntfile.js** exists in the repo, make sure to run `grunt`.

PR #5317, opened 2026-04-20
Fix Agentic Commerce product lookup by SKU

wjrosa:
### Root cause

The Agentic Commerce manual approval validator and tax calculator were treating Stripe's `sku_id` line-item field as if it were a WooCommerce product ID. Two call sites had an explicit `(int)` cast on the SKU:

- `WC_Stripe_Agentic_Commerce_Manual_Approval::validate_line_item()`: `$product_id = (int) $line_item->get_sku_id();`
- `WC_Stripe_Agentic_Commerce_Tax_Calculator::calculate_line_item_taxes()`: `resolve_product( (int) $product_id )` (variable name was misleading — the value was actually the SKU string).

PHP string-to-int coercion truncates at the first non-numeric character, so realistic SKUs produced broken lookups:
- `"ABC-123"` → `0`
- `"PROD-001"` → `0`
- `"42-blue"` → `42` (could silently resolve to an **unrelated** product whose ID happens to be 42)

`WC_Stripe_Agentic_Commerce_Product_Resolver::resolve_product( 0 )` then failed resolution, so every checkout with a non-numeric SKU errored out — and a checkout with a numeric-prefixed SKU could point at the wrong product.

The bug was hidden by tests because the shared fixture used `(string) $product->get_id()` as `sku_id`. Every test SKU was a numeric string equal to a real product ID, so the cast happened to round-trip.

### Fix

- Route both flows through `wc_get_product_id_by_sku()` — the correct WooCommerce API for resolving a SKU to a product.
- Guard against empty `sku_id` with a distinct `'Line item ... has no sku_id'` error. Without this, `wc_get_product_id_by_sku('')` matches any product with empty `_sku` meta, which could silently resolve a malformed request to an unintended product.
- Unknown SKUs now throw `'Product not found for line item ... with SKU "..."'` at extraction time; the now-unreachable zero guard in `calculate_line_item_taxes` is removed.
- Existing tax-calculator, manual-approval, and customization-hook tests (plus the shared trait helper) were relying on the buggy `(string) $product->get_id()` as `sku_id`. They now use `$product->get_sku()` with unique SKUs per product and updated assertions for the new contract.
- Adds matching 10.7.0 entries to `changelog.txt` and `readme.txt`.

## Test plan
- [x] `npm run test:php` — full suite green (1832 tests, 4178 assertions)
- [x] `npm run phpstan` — passed locally via pre-push hook
- [ ] Manual: simulate a Stripe `customize_checkout` and `finalize_checkout` webhook against a product with a non-numeric SKU and confirm the lookup resolves correctly

### DIFF
--- a/includes/agentic-commerce/class-wc-stripe-agentic-commerce-tax-calculator.php
+++ b/includes/agentic-commerce/class-wc-stripe-agentic-commerce-tax-calculator.php
@@ -36,13 +36,23 @@ class WC_Stripe_Agentic_Commerce_Tax_Calculator {
 		$line_items = [];
 
 		foreach ( $event->get_line_items() as $line_item ) {
-			$product_id = wc_get_product_id_by_sku( $line_item->get_sku_id() );
+			$sku = $line_item->get_sku_id();
+			if ( '' === $sku ) {
+				throw new Exception(
+					sprintf(
+						'Line item %s has no sku_id.',
+						$line_item->get_id()
+					)
+				);
+			}
+
+			$product_id = wc_get_product_id_by_sku( $sku );
 			if ( ! $product_id ) {
 				throw new Exception(
 					sprintf(
 						'Product not found for line item %s with SKU "%s".',
 						$line_item->get_id(),
-						$line_item->get_sku_id()
+						$sku
 					)
 				);
 			}
--- a/includes/agentic-commerce/class-wc-stripe-agentic-commerce-tax-calculator.php
+++ b/includes/agentic-commerce/class-wc-stripe-agentic-commerce-tax-calculator.php
@@ -27,7 +27,8 @@ class WC_Stripe_Agentic_Commerce_Tax_Calculator {
 	 *
 	 * @since 10.6.0
 	 * @param WC_Stripe_Agentic_Customize_Checkout_Event $event The customization hook event.
-	 * @return array<string,string> The line items hash. Line item ID => SKU ID.
+	 * @return array<string,int> The line items hash. Line item ID => Product ID.
+	 * @throws Exception When a SKU cannot be resolved to a product.
 	 */
 	public function extract_line_items_from_customization_hook(
 		WC_Stripe_Agentic_Customize_Checkout_Event $event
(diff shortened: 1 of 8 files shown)


includes/agentic-commerce/class-wc-stripe-agentic-commerce-tax-calculator.php
```diff
@@ -96,25 +118,16 @@ public function calculate(
...
 	 * @param WC_Stripe_API_Address  $address      The tax address.
 	 * @return array The line item tax response.
 	 */
 	private function calculate_line_item_taxes(
 		string $line_item_id,
-		string $product_id,
+		int $product_id,
 		WC_Stripe_API_Address $address
 	): array {
-		if ( '' === $product_id ) {
```
Mayisha, line 108: QQ: Don't we need the empty check for product ID anymore?

Mayisha (approved):
Looks good to me from code inspection 👍

malithsen (approved):
Code changes make sense on the surface, but it's not super clear to me what this PR is fixing/addressing.
Specifically, what causes the following behavior?
`Non-numeric SKUs were silently casting to 0 and failing resolution.`

Maybe we can improve the PR description to include more details on the underlying issue.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-tax-calculator.php
wjrosa: Not anymore — the check moved into `extract_line_items_from_customization_hook`, which now throws both when the SKU is empty and when `wc_get_product_id_by_sku()` can't resolve it. By the time `calculate_line_item_taxes` runs, `$product_id` is guaranteed to be a valid non-zero int (and the param is now typed `int`, so an empty string isn't representable here anyway).

wjrosa:
@malithsen thanks for the review! Expanded the PR description with the concrete root cause.

The tl;dr: two call sites had an explicit `(int)` cast on `$line_item->get_sku_id()` — PHP string→int coercion truncates at the first non-numeric character, so `"ABC-123"` → `0`, `"PROD-001"` → `0`, and (worst case) `"42-blue"` → `42` which could silently resolve to an unrelated product #42. `resolve_product(0)` then failed, so any non-numeric SKU broke the webhook. Tests missed it because the shared fixture used `(string) $product->get_id()` as the SKU, so every test SKU was coincidentally a valid numeric product ID.

PR #5176, opened 2026-03-26
Add Agentic Commerce admin dashboard for product feed monitoring

wjrosa:
## Summary

Adds a **React tab** inside the Stripe settings page (`WooCommerce → Settings → Payments → Stripe`) for monitoring the Agentic Commerce product feed sync. The tab is only visible when the agentic commerce feature flag is enabled.

- **Dashboard UI** — React tab showing last sync status badge, timestamp, product count, ImportSet/File IDs, next scheduled sync countdown, manual Sync Now button with loading state, inline success/error notices, and a sync history table (last 20 entries, newest first)
- **REST API** — `WC_REST_Stripe_Agentic_Commerce_Controller` with two endpoints:
  - `GET /wc/v3/wc_stripe/agentic-commerce/status` — returns `{ last_sync, history, next_sync }`
  - `POST /wc/v3/wc_stripe/agentic-commerce/sync` — triggers a manual sync, resets the automatic sync window
- **Sync persistence** — `store_sync_result()` persists each sync result to `wc_stripe_agentic_last_sync` and `wc_stripe_agentic_sync_history` options (non-autoloaded), with a filterable history limit (10–50, default 50)
- **Settings integration** — exposes `is_agentic_commerce_enabled`, test-mode-aware Stripe dashboard URL, and `admin_url()`-based logs URL via `wc_stripe_settings_params`; hides Save Changes button on the Agentic Commerce tab

### ImportSet lifecycle and UI states

Each sync creates a Stripe `ImportSet` whose status advances through the following lifecycle. On every dashboard load the controller re-polls any entry in a **non-terminal** state and persists the latest status.

| Stripe `ImportSet.status` | UI badge | Terminal? |
| --- | --- | --- |
| `pending` | ⏳ Processing | no (re-polled on dashboard load) |
| `creating_records` | ⏳ Creating records | no (re-polled on dashboard load) |
| `succeeded` | ✓ Success | yes |
| `succeeded_with_errors` | ⚠ Partial Success | yes |
| `failed` | ✗ Failed | yes |

Expected end state for a successful sync is **Success**, **Partial Success**, or **Failed**. Non-terminal states are typically visible for seconds to a minute, depending on catalog size.

### Design decisions

- **Local styled components** — `Card`, `StatusBadge`, `HistoryTable`, etc. are scoped to this panel. They are dashboard-specific presentational wrappers that don't map to reusable primitives elsewhere in the plugin.
[...]

PR #5319, opened 2026-04-20
Fix Agentic Commerce synced product count and Unknown sync status

### LINKED ISSUE #5320: Agentic Commerce: synced product count is inflated and sync status gets stuck on "Unknown"
### Description

Two related bugs in the Agentic Commerce admin dashboard (introduced in #5176):

**1. Synced product count does not match Stripe's ImportSet count**

The dashboard reports the number of products *iterated* from the database rather than the number of rows that actually ended up in the CSV delivered to Stripe. Any product skipped by `FeedValidatorInterface::validate_entry()` (missing brand, missing image, GTIN/MPN both empty, missing category, malformed price, etc.) is silently dropped from the CSV but still counted. Observed in review: dashboard showed 45, Stripe's ImportSet showed 31.

**2. Manual sync badge stays on "Unknown"**

After clicking **Sync Now**, the status badge renders as **Unknown** even when internal state shows `pending` and Stripe eventually processes the upload. Two causes:

- `store_sync_result()` falls back to the literal string `'unknown'` when `$result['status']` is missing, and the frontend's `STATUS_CONFIG` has no entry for `unknown` (so it renders the fallback badge).
- `refresh_pending_sync_statuses()` only matches `pending`, so entries stored as `unknown` are never re-polled and the UI stays stuck.

On top of both issues, there is no per-product diagnostic to explain *why* a product was excluded from the feed, which makes the count discrepancy hard to investigate from a merchant's perspective.

### Steps to reproduce

1. Install a store with a mix of products where some fail feed validation (e.g. products missing a brand, or with neither GTIN nor MPN).
2. Go to **WooCommerce → Settings → Payments → Stripe → Agentic Commerce dashboard**.
3. Click **Sync Now**.
4. Compare the \"Synced products\" number on the dashboard to the row count on the corresponding ImportSet in the Stripe Dashboard.
5. Observe the sync status badge immediately after clicking **Sync Now**.

### Expected behavior

- The synced product count on the dashboard matches the number of rows Stripe reports for the ImportSet.
- Products that are excluded from the feed are logged with their ID, SKU, and the specific validation reason so merchants can fix the underlying data.
- After clicking **Sync Now**, the badge reads **Processing** and transitions to **Success** / **Partial Success** once Stripe finishes — it never gets stuck on **Unknown**.

### Additional context

Reported by @mayisha in her review of #5176: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5176#pullrequestreview-4141655204

Fix in progress: #5319

wjrosa:
Fixes STRIPE-1092
Fixes #5320

Addresses [Mayisha's review](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5176#pullrequestreview-4141655204) on #5176, which reported:

1. The dashboard's **synced product count** did not match Stripe's ImportSet count (Mayisha saw 45 vs 31).
2. The manual sync badge stayed on **Unknown** even though the internal logs showed `pending` and Stripe eventually processed the upload.

### Root cause — product count

`ProductWalker::walk()` returns `$progress->processed_items`, which is the number of products iterated from the database, not the number of rows that ended up in the CSV. Inside `ProductWalker::iterate()` any row that fails `FeedValidatorInterface::validate_entry()` is silently `continue`d. Because we stored the iterated count, the dashboard always over-reported whenever any product failed validation (e.g. missing brand, missing image, GTIN/MPN both empty, category missing, malformed price format, etc.).

### Root cause — Unknown status

`store_sync_result()` fell back to the literal string `'unknown'` whenever `$result['status']` was missing. The frontend's `STATUS_CONFIG` has no entry for `unknown`, so it rendered the "Unknown" fallback badge. It also meant `refresh_pending_sync_statuses()` would skip the entry (since it only matched `pending`), so the UI got stuck.

### Changes

- `WC_Stripe_Agentic_Commerce_Csv_Feed`: expose `get_entry_count()` — the CSV already tracked this internally.
- `WC_Stripe_Agentic_Commerce_Integration::sync_feed()`:
  - Use the CSV entry count as the authoritative synced-product number.
  - Log a rolled-up `skipped_products` warning when the delta is non-zero.
  - When Stripe's ImportSet creation response succeeds (we got an `import_set_id`) but omits a `status`, default to `pending` rather than `unknown` so the dashboard renders Processing and the refresh picks it up.
  - When the local validator dropped products (`$skipped_count > 0`) and Stripe returned `succeeded`, upgrade to `succeeded_with_errors` so the dashboard's "Partial Success" badge matches the warning logged for the skips.
- `client/settings/agentic-commerce/sync-status.js`: poll the status endpoint every 5s while the latest sync is in any non-terminal status, so the badge flips to a terminal state without a manual reload. Mirrors the server-side `REFRESHABLE_STATUSES` list.
- `WC_Stripe_Agentic_Commerce_Feed_Validator::validate_entry()`: emit a per-product warning log that includes the product ID and exact failure reasons so merchants can diagnose gaps.
- `WC_REST_Stripe_Agentic_Commerce_Controller::refresh_pending_sync_statuses()`: also refresh entries stored as `unknown`, so any rows persisted by earlier syncs can still resolve.

## Testing instructions

Automated:

- [x] `npm run test:php -- --filter='WC_Stripe_Agentic_Commerce_Csv_Feed_Test|WC_Stripe_Agentic_Commerce_Integration_Test|WC_Stripe_Agentic_Commerce_Feed_Validator_Test|WC_REST_Stripe_Agentic_Commerce_Controller_Test'` passes (new tests for `get_entry_count` and `refresh_unknown`; pre-existing suite still green).
- [x] `npm run phpstan` clean.
- [x] `vendor/bin/phpcs` clean on the changed files.

Manual:

1. On a store with at least one product that will fail feed validation (e.g. missing brand, missing image, GTIN/MPN both empty, category missing, or a malformed price), open **WooCommerce → Stripe → Agentic Commerce** and click **Sync Now**.
2. Confirm the dashboard's synced product count matches Stripe's "rows processed" in the ImportSet (not the total product count).
3. Open the WooCommerce logs and confirm each skipped product is listed with its product ID and the exact validation failure reason(s), plus a rolled-up `skipped_products` warning.
4. Click **Sync Now** again and confirm the status badge reads **Processing** or **Creating records** (whichever Stripe returns) immediately — not **Unknown**. The dashboard polls every 5s while the status is non-terminal, so the badge should flip on its own without a manual reload.
5. Once Stripe finishes, the badge transitions to **Success** if Stripe accepted everything *and* the local validator dropped nothing, or **Partial Success** if either side reported failures.

---

### Changelog entry

<details>

#### Comment

</details>

**Post merge**


includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator.php
```diff
@@ -82,7 +82,23 @@ public function validate_entry( array $row, \WC_Product $product ): array {
...
 		 * @param \WC_Product $product Product object.
 		 */
-		return apply_filters( 'wc_stripe_agentic_commerce_validation_errors', $errors, $row, $product );
+		$errors = apply_filters( 'wc_stripe_agentic_commerce_validation_errors', $errors, $row, $product );
+
+		// Log rows that will be dropped from the feed so merchants can explain
+		// the gap between the number of products iterated and the number
+		// delivered to Stripe. The walker itself silently skips failing rows.
+		if ( ! empty( $errors ) ) {
+			WC_Stripe_Logger::warning(
```
daledupreez: This is going to log an entry _every_ time an invalid product appears _and_ include a backtrace. On large stores, that's going to generate a _lot_ of disk writes and the log file is going to be huge. Can we make this an `info()` instead?
```suggestion
			WC_Stripe_Logger::info(
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -297,6 +297,14 @@ function ( WalkerProgress $progress ) {
 				}
 			);
 
+			// Use the CSV entry count as the authoritative "synced" number — the
+			// walker returns the count of products *iterated*, which includes rows
+			// the validator silently dropped before they made it into the feed.
+			$total_products = $feed instanceof WC_Stripe_Agentic_Commerce_Csv_Feed
+				? $feed->get_entry_count()
```
daledupreez, line 360: Is it worth pushing this helper method into the WooCommerce interfaces?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -378,6 +378,20 @@ public function end(): void {
...
+	 * Get the number of entries written to the feed.
+	 *
+	 * This reflects the rows actually persisted to the CSV, so it matches
+	 * what is delivered to Stripe — unlike the ProductWalker's iterated count,
+	 * which includes rows dropped by the validator.
+	 *
+	 * @since 10.6.0
+	 * @return int Number of CSV rows written (excluding the header row).
+	 */
+	public function get_entry_count(): int {
```
daledupreez, line 391: Is it worth proposing this or a similar change as something that ought to belong in the WooCommerce `FeedInterface` interface?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
```diff
@@ -378,6 +378,20 @@ public function end(): void {
...
 	}
 
+	/**
+	 * Get the number of entries written to the feed.
+	 *
+	 * This reflects the rows actually persisted to the CSV, so it matches
+	 * what is delivered to Stripe — unlike the ProductWalker's iterated count,
+	 * which includes rows dropped by the validator.
+	 *
+	 * @since 10.6.0
```
daledupreez: Nit: Let's bump the version.
```suggestion
	 * @since 10.7.0
```

daledupreez (commented):
@wjrosa, could you share some sample invalid products, possibly as CSV or XML export to help with testing? I am not close enough to the code to know what constitutes a valid or invalid product for agentic commerce purposes.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator.php
```diff
@@ -82,7 +82,23 @@ public function validate_entry( array $row, \WC_Product $product ): array {
...
+				'Agentic Commerce: Product skipped by feed validator',
+				[
+					'product_id' => $product->get_id(),
+					'sku'        => $row['id'] ?? '',
```
diegocurbelo: The mapper uses the `product_id` (technically `postID`) as `sku` when interacting with Stripe; and I think it's confusing to log both here; they are always going to be the same `product_id` == `sku`, maybe we should remove `sku` or rename it `stripe_sku` 🤷🏼

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -319,22 +327,48 @@ function ( WalkerProgress $progress ) {
...
+			// ImportSet is always being processed by Stripe — reporting "unknown"
+			// hides that from the dashboard and blocks the pending-status refresh.
+			$import_set_id = $result['import_set_id'] ?? '';
+			$status        = $result['status'] ?? '';
+			if ( '' === $status && '' !== $import_set_id ) {
+				$status = 'pending';
+			}
+			if ( '' === $status ) {
+				$status = 'unknown';
+			}
```
diegocurbelo: Should we add a test to ensure we always store status = 'pending' (not 'unknown') when the response has an `import_set_id` but no `status` to prevent regression here?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator.php
wjrosa: Good catch — the mapper's `get_id()` returns `(string) $product->get_id()`, so `$row['id']` is always the same post ID we're already logging as `product_id`. Dropped the `sku` key from the skip log in 91c73714 rather than renaming, since the log already carries the product ID merchants need to locate the row.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Added in 91c73714. Pulled the fallback out of `sync_feed()` into `normalize_delivery_status()` and covered it with a data-provider test that asserts the key invariants:

- explicit status from Stripe wins
- `import_set_id` present + no status → `pending` (the regression guard you asked for)
- `import_set_id` missing or result empty → `unknown` (true failure path)

Keeps the sync_feed call site to one line and means a future refactor can't silently flip the fallback back to `unknown`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Good call — opened https://github.com/woocommerce/woocommerce/pull/64394 to add `get_entry_count()` to `FeedInterface` (and implement it on `JsonFileFeed`). Once that lands, we can drop the `instanceof` check on line 304 and just call `$feed->get_entry_count()` directly.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-csv-feed.php
wjrosa: Agreed — opened https://github.com/woocommerce/woocommerce/pull/64394 to add `get_entry_count(): int` to `FeedInterface` with a reference implementation on `JsonFileFeed`. When that ships we can keep `WC_Stripe_Agentic_Commerce_Csv_Feed::get_entry_count()` as the interface implementation and simplify the call site in the integration.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -480,7 +532,7 @@ public static function update_pending_statuses( array $status_updates ): void {
 			return;
 		}
 
-		$non_terminal_statuses = [ 'pending', 'creating_records' ];
+		$non_terminal_statuses = [ 'pending', 'creating_records', 'unknown' ];
```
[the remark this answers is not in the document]
wjrosa: Fair point — docblock updated in b66d9c2b to list all three non-terminal statuses.

daledupreez:
@wjrosa, it looks like there are a _lot_ of conflicts for this PR. LMK if you can get to resolving them.

wjrosa:
Unit tests issues unrelated. Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5366

wjrosa:
Thanks @daledupreez . I just fixed the conflicts with the help of AI.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator.php
```diff
@@ -82,7 +82,22 @@ public function validate_entry( array $row, \WC_Product $product ): array {
...
+		// delivered to Stripe. The walker itself silently skips failing rows.
+		if ( ! empty( $errors ) ) {
+			WC_Stripe_Logger::info(
+				'Agentic Commerce: Product skipped by feed validator',
+				[
+					'product_id' => $product->get_id(),
+					'errors'     => $errors,
+				]
+			);
+		}
```
diegocurbelo: I'm a bit worried about the impact on the log size.

If validation errors are somewhat rare then it's fine, but if they might be frequent, then we should consider moving the log to the caller, and using a sampling (first N per sync) + the total error count

wjrosa: Makes sense. Thanks for flagging it! Updated it in 3320c2ef7

Validator no longer logs per row. It accumulates failing rows into a `$collected_errors` array (capped at 50, overflow tracked separately) and exposes `get_collected_errors()`. The integration caches the validator on a property, so the walker and the post-walk caller share the same instance, with the instance reset at the start of each `sync_feed()`, so successive syncs don't pool errors. The existing rolled-up warning now includes per-product details as context.

Net per sync: one warning entry (was: one warning + N info lines), up to 50 examples in the context, bounded payload regardless of catalog size.

diegocurbelo (commented):
The count and error log works as described, and there is no `Unknown` badge.

However, the testing instructions do not match the UI... Instead of `Processing` the badge says `⏳ Creating records`, and it does not refresh/update, and after manually reloading the page, the status is `Success` not `Partial Success` (but the logs show that 1 product failed)

Image

---

Image

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator-test.php
```diff
@@ -611,4 +611,103 @@ function ( $errors ) {
...
+
+		$collected = $validator->get_collected_errors();
+
+		$this->assertSame( 0, $collected['truncated'] );
+		$this->assertArrayHasKey( $failing->get_id(), $collected['products'] );
+		$this->assertArrayNotHasKey( $passing->get_id(), $collected['products'] );
+		$this->assertNotEmpty( $collected['products'][ $failing->get_id() ] );
+
+		$failing->delete( true );
+		$passing->delete( true );
```
daledupreez: Nit: these should be handled via a `finally` or cleaned up before we run any assertions.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator-test.php
```diff
@@ -611,4 +611,103 @@ function ( $errors ) {
...
+		}
+
+		$collected = $validator->get_collected_errors();
+
+		$this->assertCount( $max, $collected['products'] );
+		$this->assertSame( 5, $collected['truncated'] );
+
+		foreach ( $products as $product ) {
+			$product->delete( true );
+		}
```
daledupreez: Nit: we should clean up before we run any assertions, or we should use a `finally` block.

wjrosa:
@diegocurbelo Thanks — both real bugs. Fixed in 76768b1cc:

- **Partial Success badge:** `normalize_delivery_status()` now folds in `skipped_count` and upgrades `succeeded` → `succeeded_with_errors` when the validator dropped products. Test cases added.
- **Auto-refresh:** dashboard polls `/status` every 5s while the latest sync is non-terminal and clears the interval once it flips to a terminal state.
- **Label drift:** testing instructions now say "Processing or Creating records" — Stripe chooses which one to return; both render the in-flight badge.

Re-test scenario should now show the badge flipping on its own and "Partial Success" when 1 product fails validation.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator-test.php
wjrosa: Same fix in 0bca4afaf — `try/finally` now guards the loop's cleanup so the bulk-created products are reaped regardless of assertion outcome.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator-test.php
wjrosa: Fixed in 0bca4afaf — wrapped the body in `try { … } finally { $failing->delete(); $passing->delete(); }` so the products are reaped even when an assertion throws.

diegocurbelo (commented):
The status refresh is fixed... but half the time it ends with an `? Unknown` status.

Also, with 8 total products (1 invalid) it correctly show 7 products synced but the status is still `Success` instead of `Partial Success`

Image

---

There is no additional error or warning in the logs other than the expected:

Image

wjrosa:
Thanks for catching both, fixed in 28f9809f9.

The `? Unknown` was Stripe surfacing `validating_records`, which `STATUS_CONFIG` didn't map — added that entry so in-flight syncs render Validating instead of the fallback chip. Stripe keeps changing and adding new statuses all the time.

The Partial Success miss was that the `succeeded → succeeded_with_errors` upgrade only ran inside `normalize_delivery_status()` at initial sync, where Stripe's status is still `pending`, so it never fired. I now persist `skipped_products` per history entry and re-apply the rule in `update_pending_statuses()`, so the refresh poll upgrades the terminal status once Stripe finishes.

diegocurbelo (approved):
LGTM 👍🏼

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -541,28 +568,54 @@ public static function get_sync_history(): array {
...
+	 * badge ("Partial Success") matches the warning logged for the skips.
+	 *
 	 * @since 10.7.0
-	 * @param array $result Delivery result from the Files API.
+	 * @param array $result        Delivery result from the Files API.
+	 * @param int   $skipped_count Count of products dropped by the local validator.
 	 * @return string Normalized status string.
 	 */
-	private static function normalize_delivery_status( array $result ): string {
+	public static function normalize_delivery_status( array $result, int $skipped_count = 0 ): string {
```
daledupreez: Does this method really need to be public? I am not sure where else we would need this logic.

client/settings/agentic-commerce/sync-status.js
```diff
@@ -158,6 +163,22 @@ const STATUS_CONFIG = {
...
 
+/**
+ * Statuses that indicate Stripe is still processing the ImportSet. While the
+ * latest entry is in any of these states, the dashboard polls so the badge
+ * flips to a terminal status without requiring a manual reload. Mirrors the
+ * server-side REFRESHABLE_STATUSES list in the REST controller.
+ */
+const NON_TERMINAL_STATUSES = [
+	'queued',
+	'validating',
```
daledupreez: Should this not be `validating_records` to match the status from Stripe?

daledupreez (commented):
Based on inspection, the changes generally look good to me, but I have two questions that I'd like to get answered before approving.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Good call — dropped to `private` and updated the test to reach it through Reflection.

client/settings/agentic-commerce/sync-status.js
wjrosa: Yep — Stripe sends `validating_records` (matches `creating_records`). Renamed it everywhere: `STATUS_CONFIG`, `NON_TERMINAL_STATUSES`, `REFRESHABLE_STATUSES`, the integration helper, and the tests.

changelog.txt
```diff
@@ -12,6 +12,7 @@ xxxx-xx-xx - version 10.7.0
 * Fix - Detect Agentic Commerce sessions via payment_intent.agent_details so their checkout.session.completed webhooks aren't skipped
 * Add - Add Agentic Commerce settings UI with feature introduction, onboarding guide, enable/disable toggle, and webhook secret management
 * Fix - Surface PHP Throwables from the Agentic Commerce checkout.session.completed flow so fatals are logged, the order rollback runs, and Action Scheduler marks the job failed
+* Fix - Report Agentic Commerce dashboard product count from delivered CSV rows so it matches Stripe's ImportSet, surface "Partial Success" when the local validator drops products, and auto-refresh the badge while a sync is in flight
```
daledupreez: This description is really long. I think we can shorten it.
```suggestion
* Fix - Improve handling of skipped products and in-flight syncs in Agentic Commerce sync dashboard
```

daledupreez (approved):
Based on inspection, the updates LGTM. My only comment is that we can shorten the changelog entry.

PR #5324, opened 2026-04-21
Fix failing E2E tests

diegocurbelo:

Fixes the `customer can use Link express checkout` E2E suite in `tests/e2e/tests/express-checkout/express-checkout.spec.js`, which was failing on three separate flows. All fixes are scoped to `tests/e2e/tests/express-checkout/utils.js`.

- **Pay button selector (popup)** — the Link auth flow opens a separate window, but the assertion was searching for `getByRole('button', { name: 'Continue payment' })` on the parent `page`. Switched to `popup.getByTestId('pay-button')`, which both targets the correct browser context and uses a stable test id instead of localizable button copy.
- **Link button render timeout (product page)** — after #5100 deferred WC session creation on product pages, the ECE takes longer to mount on first load in a clean test context (no prior session from `addToCart`). Bumped `expect(linkButton).toBeVisible()` to 60s.
- **Popup open timeout (block cart)** — `context.waitForEvent('page')` was inheriting the 15s `actionTimeout` from `playwright.config.js`, which wasn't enough for Stripe's Link popup to open on the block cart page. Passed an explicit 60s timeout.

## Testing instructions

- Setup the E2E environment: `npm run test:e2e-setup`
- Run the default E2E project: `npm run test:e2e`
- Verify all tests pass, and specifically all five tests in `link express checkout` pass:
  - inside the product page
  - inside the cart page (classic)
  - inside the checkout page (classic)
  - inside the cart page (block)
  - inside the checkout page (block)

---

### Changelog entry

<details>

#### Comment 

Link E2E test stability fixes only — no runtime behavior changes.

</details>

**Post merge**

### DIFF
--- a/tests/e2e/tests/express-checkout/utils.js
+++ b/tests/e2e/tests/express-checkout/utils.js
@@ -14,20 +14,16 @@ export const getLinkButton = async ( page, isBlockPage = false ) => {
 
 export const assertLinkModalLoads = async ( page, isBlockPage = false ) => {
 	const linkButton = await getLinkButton( page, isBlockPage );
-	await expect( linkButton ).toBeVisible();
+	await expect( linkButton ).toBeVisible( { timeout: 60 * 1000 } );
 	await expect( linkButton ).toBeEnabled();
 
 	const context = page.context();
 	const [ popup ] = await Promise.all( [
-		context.waitForEvent( 'page' ),
+		context.waitForEvent( 'page', { timeout: 60 * 1000 } ),
 		linkButton.click(),
 	] );
 
 	await popup.waitForLoadState();
 
-	await expect(
-		page.getByRole( 'button', {
-			name: 'Continue payment',
-		} )
-	).toBeVisible();
+	await expect( popup.getByTestId( 'pay-button' ) ).toBeVisible();
 };
--- a/tests/e2e/tests/express-checkout/utils.js
+++ b/tests/e2e/tests/express-checkout/utils.js
@@ -25,5 +25,7 @@ export const assertLinkModalLoads = async ( page, isBlockPage = false ) => {
 
 	await popup.waitForLoadState();
 
-	await expect( popup.getByTestId( 'pay-button' ) ).toBeVisible();
+	await expect( popup.getByTestId( 'pay-button' ) ).toBeVisible( {
+		timeout: 60 * 1000,
+	} );
 };


tests/e2e/tests/express-checkout/utils.js
```diff
@@ -14,20 +14,16 @@ export const getLinkButton = async ( page, isBlockPage = false ) => {
 
 export const assertLinkModalLoads = async ( page, isBlockPage = false ) => {
 	const linkButton = await getLinkButton( page, isBlockPage );
-	await expect( linkButton ).toBeVisible();
+	await expect( linkButton ).toBeVisible( { timeout: 60 * 1000 } );
```
malithsen, line 17:
```
After https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5100 deferred WC session creation on product pages, the ECE takes longer to mount on first load in a clean test context (no prior session from addToCart). Bumped expect(linkButton).toBeVisible() to 60s.
```

I doubt this reasoning is valid. If there was a performance impact, it should be minuscule (I'd expect it to be mili/micro seconds, not seconds). Also, this e2e is still flaky, so increasing the timeout doesn't seem to have fixed it entirely

malithsen (approved):
Thanks for the fixes. E2Es are passing 💯 
I don't think increasing the default timeout is the right approach, but it doesn't do any harm and is not a blocker.

tests/e2e/tests/express-checkout/utils.js
diegocurbelo: When I run the tests in debug locally, it takes a chunk of time — say 5s... but I agree that we should not hit the new 60s limit, maybe it's because we are using 5 workers and the GitHub CI only has 2 cores 🤷🏼

daledupreez: Not blocking, but it feels like it may be worth setting this to 30, 40, or 45 seconds for now. Given that many of the downstream flows have 2-3 retries for each test, that extra 15-30 seconds gets multiplied by 4 for each test case, and then across the suites. When things are broken, I'd rather learn sooner than have 20-30 minute e2e test runs due to unnecessarily long delays. Even 30 would double the current timeout.

But we can also ship this and monitor.

daledupreez: As a related note, it might be worth implementing a helper function to provide the extended timeout for actions so it's easier to apply blanket rules across the e2e tests. But that feels like it would be worth doing in a dedicated PR.

daledupreez (approved):
Let's get this shipped so we can resolve the current e2e test issues.

PR #5328, opened 2026-04-22
Extract Optimized Checkout into a dedicated gateway class

wjrosa:
Fixes STRIPE-1322

Introduce `WC_Stripe_OCS_Payment_Gateway` (extends `WC_Stripe_UPE_Payment_Gateway`) and have `WC_Stripe::get_main_stripe_gateway()` return it when Optimized Checkout is enabled. All OCS-aware behavior now lives on the OCS subclass; the base UPE gateway is unaware of OCS and handles the classic UPE flow.

**Why:** OCS logic was spread across ~90 LOC of `$this->oc_enabled && …` branches in the UPE gateway, token handler, and intent controller, so the two checkout paths couldn't be reasoned about independently. A polymorphic gateway picked at the boundary keeps each variant's behavior co-located; the base UPE gateway exposes `resolve_*` template-method hooks the OCS subclass overrides.

**What moved to the OCS subclass:** page/title gating (`is_valid_optimized_checkout_page`, `get_title`), `javascript_params` OCS keys, `payment_fields` OCS rendering, `get_enabled_payment_method_config` method restriction, `get_tokens` sub-gateway merging, and the `resolve_*` intent/token hooks. `get_main_stripe_gateway()` and `add_gateways()` select/filter based on the OCS gateway class, which is registered in the Composer classmap (autoloaded rather than `require`d).

**Compatibility:** `$oc_enabled` stays on UPE (default `false`) and `OPTIMIZED_CHECKOUT_DEFAULT_LAYOUT` is aliased as `WC_Stripe_OCS_Payment_Gateway::DEFAULT_LAYOUT`, so external consumers keep working. A few UPE methods moved `private` → `protected` so the subclass can override them.

**Subscriptions (why `trait-wc-stripe-subscriptions.php` changed):** two subclass-related guards that protect against the [#5325](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5325) duplicate-hook incident:
- `get_title()` uses the static `WC_Stripe_UPE_Payment_Method_OC::get_alternative_title()` instead of instantiating the OC payment method (whose constructor re-registers subscription hooks under the `stripe` id — the original cause of duplicate renewal charges).
- The trait's once-global hook guard matched the UPE gateway class exactly; widened to admit subclasses (via `is_subclass_of`) so the OCS gateway still registers subscription hooks (mandate-info injection, renewal/resubscribe meta cleanup, SCA redirect handling) while the OC payment method stays excluded.

## Testing instructions

Automated (Docker): `npm run phpstan` (0 errors), `npm run lint:php` (clean), `npm run test:php:parallel` (2,447 passing, incl. gateway-selection, token-merging, reusability-recheck, the OCS subscription-hook registration, and the order-title path).

Manual — confirm checkout behavior is unchanged in both variants:
1. OC **disabled**: complete a classic and a Blocks checkout; confirm standard UPE rendering and no OCS params.
2. OC **enabled**: complete a classic and a Blocks checkout; confirm the OC Element, testing instructions, adaptive-pricing selector, and correct payment-method title on thank-you/order-details (including the admin order screen).
3. Save a card under OCS, then visit **My Account → Payment Methods**; confirm sub-gateway tokens (SEPA/ACH/Amazon Pay) surface under the Stripe gateway.
4. Under OCS with a non-reusable method (e.g. iDEAL with save toggle off), confirm no token is persisted.
5. Run **Add Payment Method** and **Change Payment Method** (subscription) under OCS; confirm the OC Element is not shown.
6. With OCS enabled, complete a **WooCommerce Subscriptions** purchase and trigger a renewal; confirm exactly one renewal charge (no duplicate).
7. Confirm Express Checkout (Apple Pay / Google Pay / Link) is unchanged in both variants.

---

-   [ ] Tested on mobile (or does not apply) — backend-only refactor, no UI layout change

### Changelog entry

<details>

#### Comment

Dev - Extract Optimized Checkout into a dedicated payment gateway class so the classic UPE and Optimized Checkout flows are handled independently

</details>

**Post merge**

cc @woocommerce/quark


includes/class-wc-stripe.php
```diff
@@ -866,16 +867,38 @@ public function register_routes() {
...
+	 *
+	 * @return bool
+	 */
+	protected function is_optimized_checkout_enabled(): bool {
+		if ( ! WC_Stripe_Feature_Flags::is_oc_available() ) {
+			return false;
+		}
+		$settings = WC_Stripe_Helper::get_stripe_settings();
+		return isset( $settings['optimized_checkout_element'] ) && 'yes' === $settings['optimized_checkout_element'];
+	}
```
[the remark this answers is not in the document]
wjrosa: Done — `add_gateways()` now keys the OCS filtering off `$main_gateway instanceof WC_Stripe_OCS_Payment_Gateway` (the instantiated class) rather than the raw option, so filtering can't drift from the gateway `get_main_stripe_gateway()` actually selects. 5351831

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
```diff
@@ -0,0 +1,404 @@
...
+
+	/**
+	 * Whether the current page is one where the OCS Payment Element may render.
+	 *
+	 * @return bool
+	 */
+	public function is_valid_optimized_checkout_page(): bool {
+		return ! $this->is_on_add_payment_method_page()
+			&& ! $this->is_changing_payment_method_for_subscription();
+	}
```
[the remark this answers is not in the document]
wjrosa: Already separated in the latest revision: `is_optimized_checkout_active()` is the broad predicate used for non-render logic (token merging, title), while `is_valid_optimized_checkout_page()` now also requires `is_checkout()` to gate rendering. `javascript_params()` uses the renderable check; `get_tokens()` uses the broad one. Covered by `test_is_valid_optimized_checkout_page` (with an `is_checkout` column) and `test_is_optimized_checkout_active`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1129,6 +1023,58 @@ public function payment_fields() {
...
+	 * Whether to hide the store-level save-payment-method checkbox.
+	 *
+	 * Base UPE hides it when Link is enabled — Link handles save consent via the Payment
+	 * Element. Subclasses may override.
+	 *
+	 * @return bool
+	 */
+	protected function should_hide_save_payment_method_checkbox(): bool {
+		return WC_Stripe_UPE_Payment_Method_Link::is_link_enabled( $this );
+	}
```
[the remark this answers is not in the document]
wjrosa: This is pre-existing `develop` behavior, not new here: the old `$hide_for_link = ! $show_optimized_checkout && Link::is_link_enabled()` gate was form-wide, not per-method. The refactor relocates it unchanged into `should_hide_save_payment_method_checkbox()`, keeping this PR behavior-preserving. If the form-wide hide is itself undesirable, it's worth a separate fix outside this refactor.

wjrosa:
Picking up the three Major findings from the review that GitHub couldn't post inline:

**`get_title()` — order title on non-checkout order screens (OCS 49–53).** Valid bug — fixed in a0185f02e. The order-received / order-details title resolution was gated behind the checkout-only `is_valid_optimized_checkout_page()`, so the admin order-edit and My Account view-order screens (not checkout pages) fell through to the parent title. Now resolved via the broad `is_optimized_checkout_active()` check before the render gate, mirroring the `get_tokens()` fix. Added a regression test.

**`resolve_intent_payment_method_types()` — single-element type array (OCS 356–380).** Intentional and behavior-preserving: `develop`'s OC branch hard-coded `payment_method_types = [ $selected_payment_type ]` too (pre-merge). OCS deliberately sends a single payment_method_type rather than the base card+Link expansion — delegating to the base builder would change shipped OCS behavior. The empty-type edge predates this PR; happy to add a guard in a follow-up, but it's not a regression here.

**`__get()` / `__set()` deprecated-property proxies (UPE 4868–4913).** This code was added on `develop` in #5294 and isn't touched by this refactor — out of scope for the OCS extraction. The shadow-storage concern is real but pre-existing; best fixed in a separate PR against `develop`.

includes/class-wc-stripe.php
```diff
@@ -559,9 +559,12 @@ public function plugin_row_meta( $links, $file ) {
...
 	 */
 	public function add_gateways( $methods ) {
-		$main_gateway  = $this->get_main_stripe_gateway();
-		$methods[]     = $main_gateway;
-		$is_oc_enabled = 'yes' === $main_gateway->get_option( 'optimized_checkout_element', 'no' );
+		$main_gateway = $this->get_main_stripe_gateway();
+		$methods[]    = $main_gateway;
+		// Key the filtering off the instantiated gateway class so admin/editor filtering can't
+		// treat OCS as on while the classic gateway is actually selected (e.g. when the feature
+		// is unavailable but the setting is still 'yes'). Selection happens in get_main_stripe_gateway().
```
daledupreez: We can streamline this comment.
```suggestion
		// Use the type of the gateway instance to determine whether OCS is enabled.
```

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -551,7 +560,8 @@ public function test_payment_fields_renders_currency_selector_conditionally(
 	 * the test copy renders before the currency selector.
 	 */
 	public function test_payment_fields_renders_test_copy_before_currency_selector(): void {
-		$gateway = $this->getMockBuilder( WC_Stripe_UPE_Payment_Gateway::class )
+		// Currency-selector rendering under OCS lives on WC_Stripe_OCS_Payment_Gateway.
```
daledupreez: Shouldn't we move tests specifically aimed at the new class to a dedicated test class? We won't render the currency selector for the UPE gateway class, so I don't think the test should stay here.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -4062,14 +4079,19 @@ public function test_prepare_payment_information_oc_drops_save_flag_when_resolve
 		$order             = WC_Helper_Order::create_order();
 		$payment_method_id = 'pm_test_' . $resolved_type;
 
-		$this->mock_gateway->oc_enabled = true;
+		// The OCS type resolution and save-flag re-check live on the OCS gateway; the base
```
daledupreez: It feels like we may need some level of coverage that looks at both gateway types.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -802,101 +795,29 @@ protected function is_on_add_payment_method_page(): bool {
 	}
 
 	/**
-	 * Checks if we are on a page where the optimized checkout can be shown.
+	 * Whether Optimized Checkout is the active payment strategy for the current request.
 	 *
-	 * @return bool True if we are on a page where the optimized checkout can be shown, false otherwise.
-	 */
-	public function is_valid_optimized_checkout_page(): bool {
```
daledupreez, line 957: I think we need to keep `is_valid_optimized_checkout_page()` available but deprecated on the UPE gateway, as it was technically a public API. I want to make completely sure we discourage use before we remove it.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
daledupreez: @wjrosa, do we have a tracking issue or a dedicated PR for this concern? I think we should make sure we get this addressed given that it has been surfaced.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4704,64 +4584,96 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
...
 
-			// Skip if not available at checkout (e.g. currency not supported for this method).
-			if ( ! $this->is_enabled_at_checkout( $stripe_id ) ) {
-				continue;
-			}
+	/**
+	 * Resolves the payment-method instance used when creating a saved token from a setup intent.
+	 *
+	 * @param string $payment_method_type    Stripe payment method type key.
+	 * @param mixed  $payment_method_details Payment method details array or object.
```
daledupreez: Nit: Would it not be better to document this as `array|object`?
```suggestion
	 * @param array|object $payment_method_details Payment method details array or object.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4704,64 +4584,96 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
...
+	 * Resolves the selected payment type + payment method types sent to Stripe when creating an intent.
+	 *
+	 * @param string      $selected_payment_type   Currently selected payment type.
+	 * @param string|null $payment_method_id       Stripe payment method id (may be empty for express flows).
+	 * @param object      $payment_method_details  Payment method details object from Stripe.
+	 * @param WC_Order    $order                   WC order being paid.
+	 * @param string|null $express_payment_type         Express checkout type if applicable.
+	 * @param bool        $save_payment_method_to_store  Whether the payment method should be saved.
+	 *
+	 * @return array{ selected_payment_type: string, payment_method_types: string[], save_payment_method_to_store: bool }
```
daledupreez: Not blocking: should we use [WordPress array formatting](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#1-1-parameters-that-are-arrays) here and in the subclass? It feels like it would be a bit easier to see what the values are if we use that multiline approach.
```suggestion
	 * @return array{
	 *     @type string   $selected_payment_type        The selected payment type.
	 *     @type string[] $payment_method_types         The payment method types to send to Stripe.
	 *     @type bool     $save_payment_method_to_store Whether to save the payment method.
	 * }
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4704,64 +4584,96 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
...
+	 */
+	protected function resolve_intent_payment_method_types( string $selected_payment_type, $payment_method_id, $payment_method_details, $order, $express_payment_type, bool $save_payment_method_to_store = false ): array {
```
daledupreez: Can we use stricter types for the arguments? It feels a bit weird to have so many loose types blended in with two strict types.
```suggestion
	protected function resolve_intent_payment_method_types( string $selected_payment_type, ?string $payment_method_id, object $payment_method_details, WC_Order $order, ?string $express_payment_type, bool $save_payment_method_to_store = false ): array {
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3281,36 +3194,10 @@ protected function prepare_payment_information_from_request( WC_Order $order ) {
...
-			$payment_method_types = $this->get_payment_method_types_for_intent_creation(
-				$selected_payment_type,
-				$order->get_id(),
-				$this->get_express_payment_type_from_request()
-			);
-		}
+		$intent_selection             = $this->resolve_intent_payment_method_types( $selected_payment_type, $payment_method_id, $payment_method_details, $order, $express_payment_type, $save_payment_method_to_store );
+		$selected_payment_type        = $intent_selection['selected_payment_type'];
+		$payment_method_types         = $intent_selection['payment_method_types'];
+		$save_payment_method_to_store = $intent_selection['save_payment_method_to_store'];
```
daledupreez: Is it worth using a split assignment to make the extraction more explicit? (I am not sure which is going to be more readable, so this is a suggestion/question to consider, not an explicit recommendation or request.)
```suggestion
		[
			'selected_payment_type'        => $selected_payment_type,
			'payment_method_types'         => $payment_method_types,
			'save_payment_method_to_store' => $save_payment_method_to_store,
		] = $this->resolve_intent_payment_method_types( $selected_payment_type, $payment_method_id, $payment_method_details, $order, $express_payment_type, $save_payment_method_to_store );
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4704,64 +4584,96 @@ public function get_validated_option( string $field_key, $empty_value = null ) {
...
+	/**
+	 * Resolves the payment method type + instance used when saving a token from a successful payment.
+	 *
+	 * @param string $payment_method_type   Payment method type currently in scope.
+	 * @param object $payment_method_object Payment method object from Stripe.
+	 *
+	 * @return array{0: string, 1: WC_Stripe_UPE_Payment_Method}
+	 */
+	protected function resolve_payment_method_for_saved_token( string $payment_method_type, $payment_method_object ): array {
+		return [ $payment_method_type, $this->payment_methods[ $payment_method_type ] ];
```
daledupreez: Nit: Is `$this->payment_methods[ $payment_method_type ]` guaranteed to be a payment method instance?

This is a pre-existing issue, but I want to flag it regardless.

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
```diff
@@ -0,0 +1,488 @@
...
+			do_action( 'wc_stripe_payment_fields_' . $this->id, $this->id );
+		} catch ( Exception $e ) {
+			WC_Stripe_Logger::error( 'Error in OCS payment fields.', [ 'error_message' => $e->getMessage() ] );
+			?>
+			<div>
+				<?php echo esc_html__( 'An error was encountered when preparing the payment form. Please try again later.', 'woocommerce-gateway-stripe' ); ?>
+			</div>
+			<?php
+		}
+	}
```
daledupreez, line 263: I really think we should try to avoid duplicating this rendering logic as much as we can -- there are a lot of detailed elements that need to be present across both gateway instances, and replicating the core elements feels wrong to me. Are there any specific reasons we can't parameterise this further? It feels like adding a `get_testing_instructions_payment_method()` helper method and extending `get_testing_instructions_allowed_tags()` to include the additional tags for `wp_kses()` could do a lot of the lifting.

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
```diff
@@ -0,0 +1,488 @@
...
+ * classic UPE flow. {@see WC_Stripe::get_main_stripe_gateway()} picks the instance to use.
+ */
+class WC_Stripe_OCS_Payment_Gateway extends WC_Stripe_UPE_Payment_Gateway {
+
+	/**
+	 * Layout used when the store hasn't configured the OCS layout option.
+	 *
+	 * @var string
+	 */
+	public const DEFAULT_LAYOUT = 'accordion';
```
daledupreez: Two comments:
* I think we should use a more descriptive name for the constant, as it's not clear what layout this is referring to without knowing that OCS has a special layout property.
* Is there a reason to have this be public?

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
```diff
@@ -0,0 +1,488 @@
...
+		$stripe_params['shouldShowOptimizedCheckout']   = $should_show_optimized_checkout;
+		$stripe_params['shouldExpandOptimizedCheckout'] = $should_show_optimized_checkout && WC_Stripe_Feature_Flags::should_expand_ocs_in_legacy_checkout();
+		$stripe_params['isAdaptivePricingEnabled']      = $should_show_optimized_checkout && $this->is_adaptive_pricing_supported();
+
+		if ( $should_show_optimized_checkout ) {
+			$stripe_params['OCLayout']                      = $this->get_option( 'optimized_checkout_layout', self::DEFAULT_LAYOUT );
+			$stripe_params['paymentMethodConfigurationId']  = WC_Stripe_Payment_Method_Configurations::get_configuration_id();
+			$stripe_params['excludedPaymentMethodTypes']    = $this->get_excluded_payment_method_types();
+			$stripe_params['optimizedCheckoutClassicTitle'] = WC_Stripe_UPE_Payment_Method_OC::get_classic_title();
+		}
```
daledupreez, line 123: I am not sure if we have unit tests that cover these values, especially the values that should be included when OCS is enabled.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -84,8 +84,12 @@ public function maybe_init_subscriptions() {
 			return;
 		}
 		// Secondary check to skip registration for the OC payment method, which mimics the Stripe ID.
+		// Allow the UPE gateway and its subclasses (e.g. WC_Stripe_OCS_Payment_Gateway) through;
+		// the OC payment method is not a gateway, so it is correctly excluded. A string/get_class
+		// check (rather than instanceof) keeps PHPStan from narrowing $this across the many
+		// payment-method classes that also use this trait.
 		$current_class = get_class( $this );
-		if ( WC_Stripe_UPE_Payment_Gateway::class !== $current_class ) {
+		if ( WC_Stripe_UPE_Payment_Gateway::class !== $current_class && ! is_subclass_of( $current_class, WC_Stripe_UPE_Payment_Gateway::class ) ) {
```
daledupreez: FYI, I've just pushed up https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5540 to introduce a refactor here that looks like it would work with your changes.

daledupreez (commented):
I have feedback based on inspection, but haven't tested yet.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Done in 5d4c3a0e2 — moved the OCS-specific tests (currency selector, OC config builder, page predicates, get_title, subscription hooks, OCS save-flag resolution) into a new `WC_Stripe_OCS_Payment_Gateway_Test` class.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Added `test_prepare_payment_information_base_gateway_keeps_form_sent_type` in 5d4c3a0e2: same harness, but on the base gateway, asserting the form-sent type and save flag survive untouched when Stripe reports a different type. The OCS variant lives in the new OCS test class.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Restored in 5d4c3a0e2 as a deprecated shim (`wc_deprecated_function`, 10.9.0) with the original page-aware behavior, pointing callers at the OCS gateway method.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Applied in 5d4c3a0e2. One adjustment: `WC_Stripe_API::get_payment_method()` can yield null/non-object on failure (surfaced by the mocked-API tests), so the call site now normalizes the response to an object before passing it, keeping the `object` type honest.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Went with the destructured assignment in 5d4c3a0e2 — it drops the throwaway `$intent_selection` variable and the keys make the extraction self-documenting.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: It was not guaranteed, no — an unregistered type would have fataled. Now returns `?? null` and the docblock says `WC_Stripe_UPE_Payment_Method|null`; the only caller (`handle_saving_payment_method`) already bails on a falsy instance. In 5d4c3a0e2.

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
wjrosa: Agreed — the override is gone in 5d4c3a0e2. The base `payment_fields()` now renders instructions via `render_testing_instructions()` + `get_testing_instructions_payment_method()`, and the OCS class only swaps the instruction source (OC method) and extends `get_testing_instructions_allowed_tags()` with the `div` wrappers, exactly as you suggested. The OC instructions keep a small `render_testing_instructions()` override because the per-method `<div>` blocks cannot live inside the base `<p class="testmode-info">` wrapper. The fieldset/save-checkbox/error markup is no longer duplicated.

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
wjrosa: Better yet: the constant was redundant — the base gateway already exposes `OPTIMIZED_CHECKOUT_DEFAULT_LAYOUT` (used by the settings page), so 5d4c3a0e2 deletes `DEFAULT_LAYOUT` and uses the inherited one.

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
wjrosa: Added `test_javascript_params_exposes_ocs_keys` in the new OCS test class (5d4c3a0e2): asserts the OC flags mirror the render check and that the OCS-only keys (OCLayout, paymentMethodConfigurationId, excludedPaymentMethodTypes, optimizedCheckoutClassicTitle) are present only on a valid OC page.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Merged develop (including #5540) into this branch — the trait conflict resolved cleanly on top of your refactor: `maybe_register_plugin_hooks()`'s `instanceof` check already admits the OCS subclass, so my subclass guard became unnecessary.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Heads-up: PHPStan cannot parse the WP-style `@type` lines inside `array{}` (broke the build), so 5d4c3a0e2 converts it to the equivalent PHPStan array-shape syntax — still multiline with the same per-key info.

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
```diff
@@ -174,68 +175,56 @@ protected function get_enabled_payment_method_config(): array {
...
+	}
+
+	/**
+	 * The payment method whose testing instructions render in test mode.
+	 *
+	 * @return WC_Stripe_UPE_Payment_Method
+	 */
+	protected function get_testing_instructions_payment_method(): WC_Stripe_UPE_Payment_Method {
+		if ( $this->is_valid_optimized_checkout_page() ) {
+			return new WC_Stripe_UPE_Payment_Method_OC();
```
daledupreez, line 239: Are we keeping the OC payment method as well as the new gateway class? It feels like that may introduce confusion in future, so I think it would be well worth describing the thinking behind keeping the class and/or describing how you think it should (not) be used in future.

tests/phpunit/payment-methods/class-wc-stripe-ocs-payment-gateway-test.php
```diff
@@ -0,0 +1,647 @@
...
+				WC_Stripe_Payment_Methods::IDEAL,
+				WC_Stripe_Payment_Methods::SEPA_DEBIT,
+			]
+		);
+
+		$get_config = new ReflectionMethod( WC_Stripe_OCS_Payment_Gateway::class, 'get_enabled_payment_method_config' );
+		$get_config->setAccessible( true );
+		$config = $get_config->invoke( $gateway );
+
+		$by_method = $config[ WC_Stripe_Payment_Methods::CARD ]['showSaveOptionByMethod'];
```
daledupreez, line 236: Non-blocking nit: it may be worth confirming that this is a valid array before asserting a specific array shape.
```suggestion
		$this->assertIsArray( $config[ WC_Stripe_Payment_Methods::CARD ]['showSaveOptionByMethod'] ?? null );
		$by_method = $config[ WC_Stripe_Payment_Methods::CARD ]['showSaveOptionByMethod'];
```

tests/phpunit/payment-methods/class-wc-stripe-ocs-payment-gateway-test.php
```diff
@@ -0,0 +1,647 @@
...
+				'is_add_payment_method'      => false,
+				'is_changing_payment_method' => false,
+				'expected'                   => true,
+			],
+			'OCS disabled'                                => [
+				'oc_enabled'                 => false,
+				'is_add_payment_method'      => false,
+				'is_changing_payment_method' => false,
+				'expected'                   => false,
+			],
```
daledupreez, line 522: Just to be extra safe, it may be worth testing with OCS disabled _and_ the flags indicating that we would otherwise expect the expected flag to be true. That ensures that we won't get a regression where the downstream logic ignores the OCS disabled setting.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -4071,35 +4171,35 @@ public function test_prepare_payment_information_sets_klarna_preferred_locale_fo
...
-		// UPE gateway leaves the form-sent type untouched. Exercise the OCS variant here.
-		$gateway = $this->getMockBuilder( WC_Stripe_OCS_Payment_Gateway::class )
+		$gateway = $this->getMockBuilder( WC_Stripe_UPE_Payment_Gateway::class )
 			->setConstructorArgs( [] )
 			->onlyMethods( [ 'get_stripe_customer_id' ] )
 			->getMock();
+		$gateway->method( 'get_stripe_customer_id' )->willReturn( 'cus_mock' );
 
-		$resolved_method_stub = $this->getMockBuilder( WC_Stripe_UPE_Payment_Method::class )
+		$ideal_stub = $this->getMockBuilder( WC_Stripe_UPE_Payment_Method::class )
```
daledupreez: Nit: Why not use the iDEAL payment method directly?

phpstan-baseline.neon
```diff
@@ -4302,6 +4302,12 @@ parameters:
 			count: 1
 			path: includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
 
+		-
+			message: '#^Cannot access offset string on array\|object\|false\.$#'
```
daledupreez: Do we know where this new error is coming from?

includes/class-wc-stripe.php
wjrosa: Applied — the comment now reads exactly your suggestion (`includes/class-wc-stripe.php:569`).

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Documented as `array|object|false`: the setup-intent path can also yield `false` when no details are present, so the wider union keeps the type honest.

tests/phpunit/payment-methods/class-wc-stripe-ocs-payment-gateway-test.php
wjrosa: Applied via your committed suggestion (d2122a443) — the array shape is now asserted before we destructure it.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Done in b796dbde7 — swapped the abstract mock for a real `WC_Stripe_UPE_Payment_Method_Ideal`. `set_up()` enables `sepa_tokens_for_ideal` so it's reusable, and its retrievable type is SEPA, which is exactly the cross-type case this test exercises.

includes/payment-methods/class-wc-stripe-ocs-payment-gateway.php
wjrosa: Documented the rationale in the class docblock (b796dbde7): it's an internal pseudo-method the OCS gateway composes for the title, the aggregated per-method testing instructions and the retrievable type — not a real Stripe method and not a standalone selectable one. The docblock states it's owned by the OCS gateway and must not be used outside it.

phpstan-baseline.neon
wjrosa: It came from the `array|object|false` return type on `get_payment_method_data_from_intent()`: two callers do `$payment_method_details[ $type ]->generated_sepa_debit`, and offset access isn't valid on the `object|false` side of that union. Rather than baseline it, b796dbde7 guards both accesses with `is_array()` (the generated-SEPA token only exists on the array-cast charge details for redirect methods) and drops the baseline entry — `npm run phpstan` is clean without it.

tests/phpunit/payment-methods/class-wc-stripe-ocs-payment-gateway-test.php
wjrosa: This is already covered by the `OCS disabled` provider row: it sets the neutral page flags (both `false`) — the exact flags that return `true` on the OCS gateway — but routes through the base `WC_Stripe_UPE_Payment_Gateway`, whose `is_optimized_checkout_active()` is hard-wired to `false`. So it asserts the disabled gateway wins even when the page flags would otherwise read as active, which is the regression you're guarding against. Happy to rename the row if the intent isn't obvious.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
wjrosa: No tracking issue exists for it yet. Since it's pre-existing `develop` behavior (the form-wide Link hide) rather than something this refactor introduces, I'd keep it out of this PR's scope and file a separate follow-up to narrow it per-method — happy to open one so it isn't lost.

tests/phpunit/payment-methods/class-wc-stripe-ocs-payment-gateway-test.php
daledupreez: Just to be explicit, I think the cases that we should add are `OCS disabled, add payment method page` and `OCS disabled, change payment method`. While they are technically redundant, they ensure that we won't fall into this case in future. This outcome/failure mode is probably not possible under the current implementation, but the point is to protect against that via some additional test cases.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
daledupreez: That was my thinking, as I agree that it is not something for this PR. Can you create a tracking issue for this? I think you'll be able to capture the situation better.

tests/phpunit/payment-methods/class-wc-stripe-ocs-payment-gateway-test.php
wjrosa: Alright, makes sense. Added them in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5328/commits/2b134622a63473d6f5aa6a07972e84b201921936

Mayisha (commented):
Thanks for the refactor and addressing all the previous review comments. Did some functional testing and it looks good to me. I 

✅ Checkout with OCS disabled (card, Cash App)
✅ Checkout with OCS enabled (card, Cash App)
✅ Checkout with AP
✅ Checkout with Google Pay
✅ Save card and non-card method in My Account.
✅ Checkout with a saved payment method.
✅ Save payment method during checkout (checkbox visiblity is correct)
✅ Save iDeal as Sepa token checked, checkout with iDeal.
✅ Payment method title on order received page, order edit page and emails look good.
⚠️ Got checkout failure in one scenario for iDeal. I had Euro as store currency and on checkout page the selected address was in US. The dynamic element showed me iDeal for payment but checkout fails with the following error-

Image

Image

wjrosa:
Thanks for the thorough pass, Mayisha! 🙏

On the iDEAL case: I dug in, and it's a pre-existing behavior, not introduced by this refactor. I tested it against the `develop` branch. I am working on a fix, and will open another PR for it

Mayisha (approved):
Thanks for checking the iDeal issue @wjrosa. This looks good from my end 👍

daledupreez:
@wjrosa, as a thought: might it make sense to include this change in our August release? Given that it's a pretty fundamental change, I think it's worth giving ourselves a few weeks with the change in `develop` to help identify a larger set of possible bugs.

wjrosa:
Sure! Just updated the milestone here

PR #5330, opened 2026-04-22
[RSM - Diagnostics Mode] Implement frontend event recorder

malithsen:
Fixes RSM-628
p2: phcsdm-tK-p2

This PR implements a client-side Checkout Diagnostics Mode recorder. When `window.wcStripeDiag.active === true`, this code captures a  structured trace of a checkout session: Stripe Element lifecycle, Stripe SDK call timings, Express Checkout interactions, Blocks `onPaymentSetup` brackets, and parent-frame console warnings. Trace is sent via `navigator.sendBeacon` to a REST endpoint (not implemented yet)

All recorder code lives in the new `client/diagnostics/` directory. 

### Some architectural choices

- **Call-site brackets** Every Stripe SDK call is captured via `diagnostics.aroundStripeCall( method, () =>
  stripe.method(args) )`.  In an earlier iteration, I tried to wrap the Stripe instance itself, but it broke `@stripe/react-stripe-js`'s type validations. The current call-site approach doesn't have that risk
- **Active-gate at every helper entry.** Recorder code never loads its event subscriptions when `window.wcStripeDiag?.active` is false; pass-through is the dominant code path.
- **sessionStorage persistence for off-site round-trips.** Buffer is persisted on `pagehide` and replayed on next boot. Persisted copy is cleared after a successful pagehide-flush
- **Parent-frame console capture.**  Each call`console.warn` / `console.error` gets an event with the message truncated. **Known limitation:** Stripe's controller and Element iframes are cross-origin - their `console.warn`/`console.error` calls are not captured.
- **Per-event allowed fields.** Every event has an explicit allowed-data-keys list. 

## Testing instructions

**Set-up**
The recorder is dormant unless a `wcStripeDiag` global is set. Update `client/api/index.js` to add the following snippet

```
if ( typeof window !== 'undefined' && ! window.wcStripeDiag ) {
        window.wcStripeDiag = {
                active: true,
                sessionId: 'smoke-' + Math.random().toString( 36 ).slice( 2, 10 ),
                nonce: 'fake',
                endpoint: '/wp-json/wc/v3/wc_stripe/diagnostics/events',
        };
}
```

All POST /`wp-json/wc/v3/wc_stripe/diagnostics/events` requests will return 404. That's expected. Just inspecting the request body is enough.

- Go through Blocks/Classic checkout with successful/failed/3DS cards and notice requests to the above diagnostics endpoints with a list of `events` reflecting the shopper activity.
- Test express checkout buttons in checkout pages and notice the requests to the diagnostics endpoint.
- Remove the above `wcStripeDiag` stub and notice that no requests are sent when going through checkout

---

### Changelog entry

<details>

#### Comment 
This is being merged into a feature branch. We can add a single changelog entry when merging the feature branch to develop.

</details>

**Post merge**


client/diagnostics/recorder.js
```diff
@@ -0,0 +1,345 @@
...
+	 * @param {Object} element The Stripe Element instance.
+	 * @param {string} kind    The element kind (e.g. 'payment', 'card').
+	 * @param {string} surface The checkout surface (e.g. 'classic', 'blocks').
+	 */
+	attachAfterReady( element, kind, surface ) {
+		this.record( 'element.ready', { element_type: kind } );
+		this.attach( element, kind, surface );
+	}
+
+	attach( element, kind ) {
```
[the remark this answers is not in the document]
malithsen: Dropped this parameter in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5330/commits/40635634aa344718c252e72c44f5a683ccd32435. `surface` doesn't change per-event and it's better captured elsewhere.

client/diagnostics/recorder.js
```diff
@@ -0,0 +1,345 @@
...
+
+	flush() {
+		if ( ! this.config?.active || this.buffer.length === 0 ) {
+			return;
+		}
+		const payload = JSON.stringify( {
+			diag_session_id: this.config.sessionId,
+			events: this.buffer,
+		} );
+		navigator.sendBeacon( this.config.endpoint, payload );
```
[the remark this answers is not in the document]
malithsen: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5330/commits/c04717139381df6ce9972b5afa2b4e7609201bfe

client/api/index.js
```diff
@@ -352,7 +353,9 @@ export default class WCStripeAPI {
 
 		const confirmAction = isSetupIntent
 			? this.getStripe().confirmSetup( confirmArgs )
```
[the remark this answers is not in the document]
malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5330/commits/769584d936b2fb8a15f83494251c15a651f65d98

wjrosa (commented):
(Superseded by the inline-comment review below.)

client/diagnostics/recorder.js
```diff
@@ -0,0 +1,357 @@
...
+		const hadBuffer = this.buffer.length > 0;
+		const storageKey = STORAGE_KEY_PREFIX + this.config.sessionId;
+		if ( hadBuffer ) {
+			const existing = readStoredState( storageKey ) || {};
+			writeStoredState( storageKey, {
+				...existing,
+				bufferedEvents: [ ...this.buffer ],
+			} );
+		}
+		this.flush( 'pagehide' );
```
wjrosa: `flush()` accepts a `reason` arg that goes nowhere. Callers pass `'pagehide'` here, `'idle'` at line 278, and `'manual'` in tests, but the `flush()` signature on line 252 doesn't take a parameter and the reason is never recorded in the payload, the URL, or any event.

Either thread it into the payload (a `flush_reason` field would actually be useful for debugging trace boundaries server-side — "why did this batch arrive?") or drop the args from call sites. Current state is dead code that misleads the reader into thinking flush context is captured.

client/diagnostics/recorder.js
```diff
@@ -0,0 +1,357 @@
...
+	flush() {
+		if ( ! this.config?.active || this.buffer.length === 0 ) {
+			return;
+		}
+		const payload = JSON.stringify( {
+			diag_session_id: this.config.sessionId,
+			events: this.buffer,
+		} );
+		navigator.sendBeacon( buildIngestUrl( this.config ), payload );
+		this.buffer = [];
```
wjrosa: `sendBeacon` returns `false` when the browser refuses the request — most commonly when the payload exceeds the per-spec ~64KB ceiling. Here the buffer is cleared unconditionally, so on overflow events are silently dropped.

A long checkout session with a chatty card field (every keystroke is an `element.change` event) can plausibly hit this. Either:
- Track buffer bytes and force-flush when crossing a threshold (~32KB to leave headroom), or
- Check the `sendBeacon` return value and re-buffer / drop oldest on failure.

OK as a follow-up if the backend track has trace-size protection planned, but worth being explicit. Tests don't cover the `sendBeacon === false` path either.

client/diagnostics/recorder.js
```diff
@@ -0,0 +1,357 @@
...
+
+			const cleared = { ...persisted };
+			delete cleared.bufferedEvents;
+			writeStoredState( storageKey, cleared );
+		}
+
+		this.pagehideHandler = () => this._handlePagehide();
+		window.addEventListener( 'pagehide', this.pagehideHandler );
+
+		this._interceptConsole();
```
wjrosa, line 55: `_interceptConsole` installs wrappers in `boot()` but they're only restored in `destroy()` (line 271), which is never called automatically. If `wcStripeDiag.active` flips off mid-session the wrappers stay installed permanently for the page lifetime.

It's cheap (`record()` gates on `active`), so this is a minor concern — but worth knowing, especially for the case where someone might toggle diagnostics off via a debug flag without reloading.

client/diagnostics/recorder.js
```diff
@@ -0,0 +1,357 @@
...
+		}
+	}
+}
+
+let _singleton = null;
+
+export function getRecorder() {
+	if ( ! _singleton ) {
+		_singleton = new Recorder();
+		_singleton.boot();
```
wjrosa, line 312: `getRecorder()` caches a singleton from boot-time `wcStripeDiag`. A late-set or mutated `wcStripeDiag` (e.g. flipping `.active`, rotating `sessionId`, swapping `endpoint`) won't take effect — the snapshot is captured once in `boot()` at line 24.

Probably intended given the use case, but worth a one-liner in the JSDoc so callers don't expect dynamic config.

client/blocks/checkout-sessions/hooks.js
```diff
@@ -27,7 +28,9 @@ export const usePaymentSetupHandler = (
 ) => {
 	useEffect(
 		() =>
-			onPaymentSetup( () => {
+			onPaymentSetup( async () => {
+				const diagHandle =
+					diagnostics.blocksPaymentSetupStart( 'checkout_sessions' );
```
wjrosa, line 33: Related gap (line 248, outside the diff): `await checkout.confirm(confirmArgs)` in `useCheckoutSuccessHandler` is the actual Stripe SDK confirmation call for Custom Checkout / OC sessions, and it's not wrapped with `diagnostics.aroundStripeCall`.

The `blocks.payment_setup.start/end` bracket here only covers the data-prep handler (`handlePaymentProcessing`), which returns success metadata before any SDK call. So the trace ends with `blocks.payment_setup.end` and then has no visibility into the actual Stripe confirmation timing — exactly the thing you'd want to see when debugging a hung confirm.

Wrapping line 248 with `diagnostics.aroundStripeCall('confirm', () => checkout.confirm(confirmArgs))` would emit `stripe.confirm.invoke/resolve` and close the gap.

client/diagnostics/wiring.js
```diff
@@ -0,0 +1,119 @@
...
+	 * Used at every Stripe SDK call site rather than mutating the Stripe
+	 * instance — see recorder.js for why mutation is unsafe.
+	 *
+	 * @param {string}   method The Stripe method name (e.g. 'confirmPayment').
+	 * @param {Function} fn     A zero-arg function that performs the Stripe call and returns its Promise.
+	 * @return {Promise} The promise returned by fn(), with diagnostics events recorded when active.
+	 */
+	aroundStripeCall( method, fn ) {
+		if ( ! isActive() ) return fn();
+		return getRecorder().aroundStripeCall( method, fn );
```
wjrosa, line 93: Microscopic perf nit (ignore unless you're doing perf passes): `aroundStripeCall` in `recorder.js:127` uses `Promise.resolve().then(fn)` to coerce sync throws into rejections. `try { return Promise.resolve(fn()) } catch (e) { return Promise.reject(e) }` would skip the unnecessary microtask hop while still normalizing throws into rejections. Fires on every wrapped Stripe call when active.

wjrosa (commented):
Most findings are inlined below. Two items can't be inlined because the relevant lines aren't in the diff:

- **`client/express-checkout/payment-flow.js:141`** — `stripe.createPaymentMethod({ elements })` in `handleManualPaymentMethodFlow` is not wrapped with `diagnostics.aroundStripeCall`.
- **`client/express-checkout/payment-flow.js:210`** — `stripe.createConfirmationToken({ elements })` in `handleConfirmationTokenFlow` is not wrapped either.

These are the meat of the ECE payment path on classic checkout — without the wraps, traces will show `express.confirm` → silence → end-of-trace. Same `diagnostics.aroundStripeCall(method, () => …)` pattern as the others.

client/diagnostics/recorder.js
malithsen: I dropped this arg for now, to simplify the implementation. We can bring that back later as needed.

client/diagnostics/recorder.js
malithsen: Added a comment in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5330/commits/50060b5d93329cc703b465c9027f8c83aae14153

client/diagnostics/recorder.js
malithsen: I'd defer the payload splitting part. It'll be a bit complex to implement and that's something we can iterate on later. 

Added `false` response handling in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5330/commits/b80abd343a142aea5dc2a8294b92eec11e9a035c

client/diagnostics/recorder.js
malithsen: I'd leave this as-is. There's no way to flip `active` mid-session.

client/diagnostics/wiring.js
malithsen: I'd defer this one as well. Also, I find the promise flow to be more readable than `try/catch`

client/blocks/checkout-sessions/hooks.js
malithsen: Thanks, updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5330/commits/b2c3e6a4855c809bb532524548073e5ab18eabbb

client/diagnostics/recorder.js
```diff
@@ -0,0 +1,371 @@
...
+				...existing,
+				bufferedEvents: [ ...this.buffer ],
+			} );
+		}
+		this.flush();
+		// If we got here, the synchronous sendBeacon was queued. Clear the
+		// persisted copy so the next boot does not double-send. If the tab
+		// is torn down BEFORE this line runs, the persisted copy survives
+		// and the next-boot replay path acts as the safety net.
+		if ( hadBuffer ) {
```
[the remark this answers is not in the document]
malithsen: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5330/commits/0e71296ebe97656c2dbd5b67a6707ea641620bb6

wjrosa (approved):
Thanks for addressing my (and Claude's) points! Looks good to me and tests well 👍

PR #5331, opened 2026-04-22
Detect Agentic Commerce sessions via payment_intent.agent_details

### LINKED ISSUE #5340: Agentic Commerce checkout.session.completed webhooks skipped as non-agentic
### Description

`checkout.session.completed` webhooks for Agentic Commerce sessions are being dropped with the debug log `Checkout session is not agentic, skipping agentic processing`. As a result, orders originating from an ACP agent never reach the agentic processing path.

Root cause: `WC_Stripe_Agentic_Checkout_Session::is_agentic()` inspected line-item `price.external_reference` and tried to resolve it to a numeric WooCommerce product ID. Stripe actually sends SKU strings in `external_reference`, so the check always failed and every agentic webhook fell through to the non-agentic skip path.

### Steps to reproduce

1. Configure an Agentic Commerce integration so Stripe can create an agentic checkout session for this store.
2. Complete an agentic checkout so Stripe fires a `checkout.session.completed` webhook whose payment intent carries `agent_details.network_business_profile`.
3. Tail the Stripe webhook log (`wp-content/uploads/wc-logs/woocommerce-gateway-stripe-*.log`).
4. Observe the `Checkout session is not agentic, skipping agentic processing` line and note that no agentic order is created.

### Expected behavior

The session should be recognized as agentic and routed through the agentic processing path, producing the expected order. Non-agentic sessions should continue to be skipped.

### Additional context

- Detection should key off the canonical Stripe signal that the webhook handler already expands on retrieve: `payment_intent.agent_details.network_business_profile`.
- Fix in #5331.
- This issue is scoped to detection only. A separate follow-up covers SKU-based product lookup inside the agentic order mapper (tracked alongside #5317).

wjrosa:
Fixes STRIPE-1103
Fixes #5340

`WC_Stripe_Agentic_Checkout_Session::is_agentic()` now keys off the Stripe-native signal the webhook handler already expands on retrieve: `payment_intent.agent_details.network_business_profile`.

The previous implementation inspected line-item `price.external_reference` and tried to resolve it to a numeric WC product ID. Stripe actually sends SKU strings there, so real agentic `checkout.session.completed` webhooks fell through the check and were logged as `Checkout session is not agentic, skipping agentic processing`. Switching to `agent_details.network_business_profile` uses the canonical Stripe signal that identifies an originating agent, and it matches what the webhook handler already expands.

This PR only changes detection. Once a session is flagged as agentic, the order mapper's SKU resolution still needs the same treatment that the `finalize_checkout`/`customize_checkout` paths got in #5317 — that will ship in a follow-up so each change stays focused.

**How can this code break?**
- A future Stripe schema change that moves or renames `agent_details.network_business_profile` would silently return `false` and re-skip agentic sessions. Existing debug logging (`Checkout session is not agentic…`) remains the early warning.
- If the webhook handler stops expanding `payment_intent` on retrieve, `$session->payment_intent` will be a string and detection will correctly return `false`; this is the intended fail-closed behavior.

- Data-provider PHPUnit coverage exercises: nbp present, nbp empty string, `network_business_profile` missing, `agent_details` absent, and unexpanded (string) `payment_intent`.
- The webhook handler test fixture (`build_checkout_session_response`) now only populates `agent_details` when the `$agentic` flag is set, so the non-agentic path in `test_process_checkout_session_completed_skips_non_agentic` exercises the new skip condition rather than a stale external_reference branch.

## Testing instructions

1. Check out this branch and run `npm run test:php -- --filter WC_Stripe_Agentic_Checkout_Session_Test` — all data-provider cases for `is_agentic()` should pass.
2. Run `npm run test:php -- --filter WC_Stripe_Webhook_Handler_Agentic_Test` — `test_process_checkout_session_completed_skips_non_agentic` and `test_skips_session_with_empty_network_business_profile` should still skip (now via absent / empty `agent_details.network_business_profile`).
3. Trigger an agentic `checkout.session.completed` webhook whose payment intent includes `agent_details.network_business_profile` populated (e.g. via an ACP partner). Confirm the `Checkout session is not agentic, skipping agentic processing` log line no longer appears and the agentic order flow runs.
4. Trigger a normal (non-agentic) `checkout.session.completed` webhook and confirm the skip log line still appears and no agentic processing runs.

---

-   [x] Tested on mobile (or does not apply) — server-side webhook logic; no UI surface.

### Changelog entry

Added in `changelog.txt` under `10.7.0`:

> Fix - Detect Agentic Commerce sessions via payment_intent.agent_details so their checkout.session.completed webhooks aren't skipped

<details>

#### Comment

N/A — changelog entry included above.

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.7.0
+* Fix - Detect Agentic Commerce checkout sessions from the payment intent's agent_details.network_business_profile so checkout.session.completed webhooks are no longer skipped as non-agentic
 * Fix - Look up products by SKU in Agentic Commerce manual approval and tax calculation flows
 * Dev - Rename payment request references to express checkout
 * Fix - Store Stripe Terminal IPP channel metadata on orders so WooCommerce can identify POS payments and suppress standard transactional emails
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.7.0
-* Fix - Detect Agentic Commerce checkout sessions from the payment intent's agent_details.network_business_profile so checkout.session.completed webhooks are no longer skipped as non-agentic
+* Fix - Detect Agentic Commerce sessions via payment_intent.agent_details so their checkout.session.completed webhooks aren't skipped
 * Fix - Look up products by SKU in Agentic Commerce manual approval and tax calculation flows
 * Dev - Rename payment request references to express checkout
 * Fix - Store Stripe Terminal IPP channel metadata on orders so WooCommerce can identify POS payments and suppress standard transactional emails
--- a/includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
+++ b/includes/agentic-commerce/class-wc-stripe-agentic-checkout-session.php
@@ -285,19 +285,26 @@ class WC_Stripe_Agentic_Checkout_Session {
 	/**
 	 * Checks whether this checkout session originates from agentic commerce.
 	 *
-	 * A session is agentic when at least one line item has an
-	 * external_reference that resolves to a nonzero integer (product ID).
+	 * A session is agentic when its payment intent carries an
+	 * agent_details.network_business_profile, which Stripe populates to
+	 * identify the originating agent. The webhook handler expands
+	 * payment_intent.agent_details on retrieve for exactly this detection.
 	 *
 	 * @since 10.6.0
 	 * @return bool
 	 */
 	public function is_agentic(): bool {
-		foreach ( $this->get_line_items() as $line_item ) {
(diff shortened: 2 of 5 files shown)


malithsen (approved):
Changes look good on inspection. Approving.

One question on the testing instructions: how are we triggering the webhooks in steps 3 and 4? Is there a simulator we could use?

daledupreez (approved):
The changes LGTM based on inspection, and it seems this is a better heuristic than what we were using before.

wjrosa:
> One question on the testing instructions: how are we triggering the webhooks in steps 3 and 4? Is there a simulator we could use?

@malithsen, sorry for the lack of better instructions! After onboarding to agentic commerce inside the Stripe dashboard, you can navigate to the product catalog section and click the "Test" link on a product from the list. This will open an agentic purchase simulator. At the end of the simulation, it triggers the related webhooks on the store

PR #5333, opened 2026-04-22
Fall back to free-form shipping line when no WC rate matches

### LINKED ISSUE #5342: Agentic Commerce orders lost when session shipping rate doesn't match any configured WC method
### Description

Agentic Commerce `checkout.session.completed` webhooks silently lose the order and log `Failed to create agentic order from checkout session` whenever the Stripe session's chosen shipping rate can't be matched to a configured WooCommerce shipping method.

Two distinct failure modes both end up in this bucket:

1. **No matching WC rate.** `WC_Stripe_Agentic_Commerce_Order_Mapper::map_shipping()` hard-fails when the chosen `shipping_rate` carries neither a `metadata.wc_rate_id` (produced by the `customize_checkout` response) nor a `display_name` that matches a configured WC shipping method. This happens whenever Stripe or the agent supplies its own canned rate (e.g. `"Expedited"`) instead of one emitted by our `customize_checkout` response — for example if the `customize_checkout` webhook never reached the merchant, or if the agent platform has default rates that don't correspond to any WC zone method. The mapper's outer try/catch deletes the in-progress order and nothing is recorded. (Since `customize_checkout` itself previously failed on non-numeric SKUs — fixed in #5332 — real traffic is now landing on this path.)
2. **Null WC session under Action Scheduler / WP Cron.** `WC_Shipping::calculate_shipping_for_package()` reads per-package rate caches from `WC()->session`, which is `null` in Action Scheduler / WP Cron context because no HTTP request bootstraps it. The agentic checkout flow runs entirely under Action Scheduler, so `calculate_shipping()` throws `Call to a member function get() on null` before it can return any rates, and the outer try/catch then deletes the order.

### Steps to reproduce

**Failure mode 1 — unmatched shipping rate:**

1. Configure an Agentic Commerce integration (requires #5331 and #5332).
2. Complete an agentic checkout where Stripe delivers a `checkout.session.completed` whose `shipping_cost.shipping_rate` has no `metadata.wc_rate_id` and whose `display_name` doesn't match any configured WC shipping method (e.g. Stripe-side default `"Expedited"`).
3. Tail `wp-content/uploads/wc-logs/woocommerce-gateway-stripe-*.log`.
4. Observe `Failed to create agentic order from checkout session` and confirm no order exists.

**Failure mode 2 — null session under Action Scheduler:**

1. Complete an agentic checkout for a session whose chosen rate does carry a valid `wc_rate_id`.
2. Observe the webhook fail with `Call to a member function get() on null` originating from `WC_Shipping::calculate_shipping_for_package` when processed from Action Scheduler.
3. Confirm no order is created.

### Expected behavior

- When the WC session is null, the mapper should initialize it (`WC()->initialize_session()`) so `calculate_shipping()` can read/write from session caches as it expects.
- When shipping calculation throws for any other reason (third-party shipping method bugs, cron-unfriendly code elsewhere), the mapper should treat that as "no rates available" rather than destroying the order.
- When the session's chosen rate has no matching WC rate, the mapper should create the order anyway with a free-form shipping line (`method_id = stripe_agentic`, `method_title = display_name`, `total = amount_shipping`) and log a warning that includes the session ID, Stripe rate details, and the list of WC rates that were actually calculated — so the shipping-zone gap is visible without blocking the order.

### Additional context

- Fix in #5333 (covers both failure modes).
- Depends on #5331 (detection) and #5332 (line-item SKU lookup); without them the webhook never reaches `map_shipping`.
- The mapper's outer try/catch behavior (delete the partial order) is intentional for fatal mapping errors, but shipping-rate mismatch and AS/cron session nullity shouldn't qualify as fatal.

wjrosa:
Fixes STRIPE-1105
Fixes #5342

Two related failures in `WC_Stripe_Agentic_Commerce_Order_Mapper::map_shipping()` were silently losing Agentic Commerce orders. The PR makes both paths land an order and log the gap instead of triggering the outer try/catch that deletes the in-progress order.

**1. Free-form fallback when no WC rate matches.** The mapper used to throw whenever the chosen Stripe `shipping_rate` had neither a `metadata.wc_rate_id` nor a `display_name` matching a configured WC shipping method. That happens whenever Stripe/the agent supplies a canned rate (e.g. `customize_checkout` never reached the merchant, or the agent has defaults like `"Expedited"`). Now the mapper creates a `WC_Order_Item_Shipping` with `method_id = 'stripe_agentic'`, `method_title = display_name`, `total = convert_from_stripe_amount(total_details.amount_shipping, currency)`, and logs the session id, Stripe rate details, and the WC rates that *were* calculated.

**2. Session bootstrap + Throwable guard around `calculate_shipping()`.** Under Action Scheduler / WP Cron there's no HTTP request to bootstrap `WC()->session`, which `calculate_shipping_for_package()` reads from. The agentic flow runs entirely under AS, so `calculate_shipping()` was throwing `Call to a member function get() on null` and the mapper dropped the order. Two defenses:
- Call `WC()->initialize_session()` when the session is null.
- Wrap `calculate_shipping()` + `get_packages()` in `try/catch(Throwable)` so a broken third-party shipping method or null-session `Error` (which the outer Exception handler wouldn't catch) falls through to the free-form fallback above.

Since `customize_checkout` previously failed entirely on non-numeric SKUs (fixed in #5332), real traffic is now hitting these paths.

### How can this code break?
- Merchants who relied on unmatched shipping rates *blocking* order creation will now receive those orders with a free-form line. Mitigated by the warning log and the namespaced `stripe_agentic` `method_id` (filterable / auditable). Explicit tradeoff: lost orders are worse than mismatched shipping lines for agentic commerce's current rollout.
- A broken `WC()->initialize_session()` factory could raise here. The outer `try/catch(Throwable)` around `calculate_shipping()` keeps that from cascading into an order deletion.
- Missing/zero `total_details.amount_shipping` produces a zero-total shipping row, which WooCommerce handles natively.

## Testing instructions

1. `npm run test:php -- --filter 'WC_Stripe_Agentic_Commerce_Order_Mapper_Test'` — the new fallback + Throwable-recovery tests pass alongside the existing `wc_rate_id` / single-rate / display-name-match cases.
2. `npm run phpstan` — clean for files this PR touches.
3. Fire an agentic `checkout.session.completed` webhook where `shipping_cost.shipping_rate` has no `metadata.wc_rate_id` and a `display_name` no WC method matches. Order should land with `method_id = stripe_agentic`, `method_title = display_name`, `total = amount_shipping`, `verify_order_total` passing, and a warning log recorded.
4. Repeat (3) under Action Scheduler / WP Cron (so `WC()->session` is `null` at entry). Confirm no `get() on null` error and the order is created.
5. Sanity-check the happy path: a session whose chosen rate has a valid `wc_rate_id` should still use the real WC shipping line, not the fallback.

---

-   [x] Tested on mobile (or does not apply) — server-side webhook; no UI surface.

### Changelog entry

* Fix - Stop Agentic Commerce checkout.session.completed orders from failing on null WC sessions or shipping rates that don't match any configured WC method

**Post merge**


wjrosa:
Added a second commit ([f7cad1dc3](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5333/commits/f7cad1dc3)) that addresses the root cause surfaced while testing: `WC_Shipping::calculate_shipping_for_package` reads from `WC()->session`, which is `null` under Action Scheduler / WP Cron — where `checkout.session.completed` actually runs. The result was a PHP `Error` ("Call to a member function get() on null") that didn't reach the existing `catch (Exception)` clauses, so orders were silently left in `pending`.

Two defenses now live alongside the free-form fallback in `map_shipping()`:

1. `WC()->initialize_session()` if the session is `null`, so shipping methods get a working session and the happy path (matching a `wc_rate_id`) can actually return real rates.
2. `try/catch (Throwable)` around `calculate_shipping()` + `get_packages()`. Any failure now logs a warning and treats the rate list as empty, which flows naturally into the existing free-form fallback.

Together: B restores correctness when `customize_checkout` did populate `wc_rate_id`; A keeps the webhook from losing orders when third-party shipping/tax code misbehaves in a cron context.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -450,9 +450,34 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
+				'Agentic order mapper: WC shipping calculation failed; will use free-form shipping line.',
+				[
+					'session_id' => $session->get_id(),
+					'error'      => $e->getMessage(),
+					'exception'  => get_class( $e ),
+					'file'       => $e->getFile(),
+					'line'       => $e->getLine(),
+				]
+			);
+		}
```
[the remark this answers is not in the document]
wjrosa: Fixed in [7fef1aa5](https://github.com/woocommerce/woocommerce-gateway-stripe/commit/7fef1aa5). Moved `WC()->initialize_session()` (and its explanatory comment) inside the existing `try { ... } catch (Throwable $e)` block so any `Error`/`TypeError` from a broken session handler now falls through to the free-form shipping fallback instead of escaping past the outer `catch (Exception)` and leaving the order in pending.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -450,9 +450,34 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
-		$wc_shipping->calculate_shipping( [ $package ] );
-		$packages = $wc_shipping->get_packages();
-		$rates    = $packages[0]['rates'] ?? [];
+		// Under Action Scheduler / WP Cron there's no HTTP request to
+		// bootstrap WC()->session, which WC_Shipping::calculate_shipping_for_package
+		// reads from. Initialize it lazily so shipping methods can cache and
+		// return rates normally.
+		if ( null === WC()->session ) {
+			WC()->initialize_session();
+		}
```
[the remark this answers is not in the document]
wjrosa: Fixed in [7fef1aa5](https://github.com/woocommerce/woocommerce-gateway-stripe/commit/7fef1aa5) — moved the `WC()->initialize_session()` guard inside the `try { ... } catch (Throwable)` block. Good catch — a non-existent class from a filter would surface as an `Error`, not an `Exception`, and would have slipped past both handlers.

wjrosa:
Addressed CodeRabbit + Claude bot review feedback in [7fef1aa5](https://github.com/woocommerce/woocommerce-gateway-stripe/commit/7fef1aa5):

**Inline (Major):** `WC()->initialize_session()` is now inside the `try { ... } catch (Throwable $e)` block. If a broken `woocommerce_session_handler` filter (non-existent class, etc.) raises a PHP `Error`, it's caught and falls through to the free-form shipping fallback — instead of escaping past the outer `catch (Exception)` and leaving the order stuck in pending.

**Outside diff — stale `@throws` docblock:** Updated `map_shipping()` to document the new step 4 (free-form fallback) in the priority list, and corrected the `@throws` contract so it only claims to throw when `WC()->shipping()` is unavailable.

**Outside diff — missing test for `Throwable` recovery path:** Added `test_shipping_falls_back_to_free_form_line_when_calculation_throws`. It hooks `woocommerce_package_rates` with a callback that throws an `Error` (exercising the specific branch where `Exception` wouldn't suffice), then asserts the order is created with `method_id = 'stripe_agentic'`, `method_title = display_name`, and the Stripe amount. The filter is removed and shipping cache reset in `finally`.

Local results:
- `--filter WC_Stripe_Agentic_Commerce_Order_Mapper_Test` — new test passes, no regressions.
- `npm run phpstan` — clean.

changelog.txt
```diff
@@ -1,6 +1,8 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.7.0
+* Fix - Initialize the WooCommerce session and guard the shipping calculation in the Agentic Commerce order mapper so Action Scheduler / WP Cron runs of the checkout.session.completed webhook no longer fail on a null session or a misbehaving shipping method
+* Fix - Create Agentic Commerce orders with a free-form shipping line when the session's chosen shipping rate has no wc_rate_id metadata and its display_name does not match any configured WC shipping method, instead of hard-failing
```
malithsen: Do we need two entries? Personally, given that this feature is unreleased, I'd avoid the "Fix" entries altogether.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -477,21 +508,52 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
+						return [
+							'id'    => $rate->get_id(),
+							'label' => $rate->get_label(),
+							'cost'  => $rate->get_cost(),
+						];
+					},
+					$rates
+				),
+			]
+		);
```
daledupreez: Should we also add an order note so the details are shown in a somewhat contextual location relative to the order?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -477,21 +508,52 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
+			$order->add_item( $shipping_item );
+			return;
 		}
 
+		// No WC rate matched. This happens when Stripe/the agent supplies a
+		// shipping rate that did not originate from our customize_checkout
+		// response (so there is no wc_rate_id metadata), and the display name
+		// does not match any configured WC shipping method. Rather than hard
+		// failing and losing the order, fall back to a free-form shipping line
+		// using Stripe's display name and amount so downstream totals match.
```
daledupreez: I think we can make this comment shorter.
```suggestion
		// No WC rate matched. This can happen when Stripe/the agent supplies a
		// shipping rate that does not include matching wc_rate_id metadata and the display name
		// does not match any configured WC shipping method.
		// When this occurs, we create the order and use `stripe_agentic` as the shipping method.
```

daledupreez:
> 3. Trigger an agentic `checkout.session.completed` for a session whose `shipping_cost.shipping_rate` has no `metadata.wc_rate_id` and whose `display_name` doesn't match any configured WC shipping method.
> 4. Trigger the same webhook under Action Scheduler (i.e. process it via a queued WP Cron job rather than synchronously), ensuring WC()->session is null at entry. Confirm the mapper initializes the session, calculate_shipping() runs without the get() on null error, and the order is created.
> 5. Sanity-check a happy-path session whose chosen rate does carry a valid wc_rate_id matching a configured WC method — the order should still use the real WC shipping line, not the stripe_agentic fallback.

@wjrosa, can you please include more specific testing details? I don't know how to trigger any of these events, especially with the specificity mentioned in the instructions.

wjrosa:
Unit tests issues unrelated. Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5366

wjrosa:
@daledupreez Actually, I am not sure if it is possible to manually test this PR. We may need to rely on the new unit tests for it.

Basically, this PR fixed an issue with the agentic commerce demo store x Stripe demo site integration (for the Stripe Sessions event). Since the shipping methods on the Stripe demo site were not available in the demo store (the Woo instance), the hook event handler failed, resulting in orders not being created. With this change, if we cannot find a matching shipping method, we dynamically create it as an inline item.

We don't currently have a "shipping methods sync" feature (and I don't think it makes sense). And the provided CSV file did not contain the methods in an "importable" way.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -477,21 +502,49 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
+			$order->add_item( $shipping_item );
+			return;
 		}
 
+		// No WC rate matched (Stripe/agent supplied a rate from outside our
+		// customize_checkout response). Fall back to a free-form shipping line
+		// so the order still lands and downstream totals balance.
+		$stripe_amount = $session->get_shipping_amount();
+		$currency      = $session->get_currency() ?? '';
+		$total         = null !== $stripe_amount
```
malithsen, line 533: Question: Does `$stripe_amount` include tax? If so, would this result in double-taxed shipping taxes?

malithsen (commented):
>  this PR fixed an issue with the agentic commerce demo store x Stripe demo site integration (for the Stripe Sessions event). Since the shipping methods on the Stripe demo site were not available in the demo store (the Woo instance), the hook event handler failed, resulting in orders not being created.


Is this change meant to be a permanent fix or a temporary workaround to get the demo store working? If it's the latter, would it make sense to patch the demo site directly?

changelog.txt
wjrosa: Not really, but I fixed that by updating the branch with `develop`

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
wjrosa: No, `total_details.amount_shipping` is Stripe's pre-tax shipping subtotal — tax is rolled into `amount_tax` separately. The shipping item is set at the pre-tax total here, and `verify_order_total()`'s `$order->calculate_totals()` is the single point where WC's tax engine adds shipping tax on top. So no double-tax in this path.

(The bigger assumption — that local WC tax config matches Stripe's `amount_tax` — exists across the whole mapper, not just this fallback.)

wjrosa:
@malithsen This fallback should be permanent, as it avoids issues when we cannot find a matching shipping option and ensures orders are created correctly regardless.

The issue with Stripe's demo site is that they already had the catalog of products and shipping methods, and we had to match that on the store side. The CSV they provided doesn't follow a format that Woo recognizes by default, so it was difficult to replicate.

This should not happen in live environments when we launch the feature, as merchants would have full control of their products and shipping methods. And the flow would be true agentic, not using a mocked AI chat (the current demo).

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -477,21 +502,49 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
+			? WC_Stripe_Helper::convert_from_stripe_amount( $stripe_amount, $currency )
+			: 0;
+
+		WC_Stripe_Logger::warning(
```
daledupreez: I think we should make this an error to ensure it gets written to the log -- I believe warnings are only written to the log when logging is enabled.
```suggestion
		WC_Stripe_Logger::error(
```

daledupreez (commented):
The changes look OK to me based on inspection, but I think we also need to add an order note when this occurs so the exception is shown in context, not just in the Stripe logs. And I think we should upgrade the warning to an error for additional visibility.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
wjrosa: Done in 963573e70 — switched to `WC_Stripe_Logger::error`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
wjrosa: Done in 963573e70.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
wjrosa: Good call — added in 963573e70. Note records the unmatched rate's display name and amount, and a test asserts on it.

malithsen (approved):
Changes look good to me on inspection.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -489,22 +514,60 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
-		$shipping_item->set_total( $matched_rate->get_cost() );
+		$shipping_item->set_method_title( $display_name );
+		$shipping_item->set_method_id( 'stripe_agentic' );
+		$shipping_item->set_total( (string) $total );
 		$order->add_item( $shipping_item );
+
+		$order->add_order_note(
+			sprintf(
+				/* translators: 1: shipping rate label from Stripe, 2: formatted shipping amount */
+				__( 'Agentic Commerce: chosen shipping rate "%1$s" (%2$s) did not match any configured WooCommerce shipping method. Recorded as a free-form shipping line.', 'woocommerce-gateway-stripe' ),
```
daledupreez: We need to specify that this is Stripe agentic commerce.
```suggestion
				__( 'Stripe Agentic Commerce: chosen shipping rate "%1$s" (%2$s) did not match any configured WooCommerce shipping method. Recorded as a free-form shipping line.', 'woocommerce-gateway-stripe' ),
```

daledupreez (approved):
Changes look good to me based on inspection. I'll update the order note text, and then get this merged.

Issue #4801, opened 2025-11-14
Stripe API calls fail URL validation


daledupreez:
### Description

Some merchants have encountered the following error being returned from attempted API calls to Stripe, mostly as a result of reconnecting to Stripe:

* WP_Error code: `http_request_failed`
* WP_Error message: `A valid URL was not provided.` (though it may be translated)

This error is from WordPress core [here](<https://github.com/WordPress/wordpress-develop/blob/ddf80469127bbc32190ef5a6f4b150bcfbb55986/src/wp-includes/class-wp-http.php#L295>) when one of the calls to `wp_http_validate_url()` or `wp_kses_bad_protocol()` detects the URL as invalid. Given that we're fairly confident we're sending in a URL with `https` as the protocol, we believe that the issue is coming from `wp_http_validate_url()`. In looking at the implementation for `wp_http_validate_url()` (see below), we believe the most likely issues stem from [the `gethostbyname()` DNS lookup failing](<https://github.com/WordPress/wordpress-develop/blob/ddf80469127bbc32190ef5a6f4b150bcfbb55986/src/wp-includes/http.php#L595-L598>) *or* [the resolved IP address being detected as local](<https://github.com/WordPress/wordpress-develop/blob/ddf80469127bbc32190ef5a6f4b150bcfbb55986/src/wp-includes/http.php#L602-L620>), which should not be happening, but could occur if a host has configured host name resolution for `api.stripe.com` to some local IP address.

[https://github.com/WordPress/wordpress-develop/blob/ddf80469127bbc32190ef5a6f4b150bcfbb55986/src/wp-includes/http.php#L539-L653](<https://github.com/WordPress/wordpress-develop/blob/ddf80469127bbc32190ef5a6f4b150bcfbb55986/src/wp-includes/http.php#L539-L653>)

### Steps to reproduce

A handful of merchants have tripped over this issue, but we don't think there is anything in the software that is causing the issue, beyond our usage of `wp_safe_remote_(get|post)()` triggering this URL validation failure. It's worth noting that the `safe` versions of those two functions apply the URL validation, while the `wp_remote_(get|post)()` functions don't do so (though it is an option).

#### Workaround

It is possible to override the URL validation logic using some code like the following, but that is not a good long-term solution.

```php
/**
 * Filter to bypass the default Stripe Gateway logic that validates whether API calls to Stripe need to pass
[...]

PR #5337, opened 2026-04-23
Fix STRIPE-816: Use wp_remote_* for Stripe API calls

rakshitwesley:
## Summary

Switches the four Stripe API call sites from `wp_safe_remote_post/get()` to `wp_remote_post/get()` so that an intermittent DNS resolution failure on the host no longer blocks every request to `api.stripe.com` with `A valid URL was not provided.`.

Merchants have tripped over this error in support tickets (`10410812-zen`, `10386690-zen`, `10441971-zen`). It originates from `wp_http_validate_url()`, which calls `gethostbyname()` and rejects the URL if DNS fails or the resolved IP is treated as local. All four affected requests target the hardcoded Stripe endpoint, so the SSRF guard that `wp_safe_remote_*` adds has no practical value for this code path.

## Bug

[STRIPE-816: Stripe API calls fail URL validation](https://linear.app/a8c/issue/STRIPE-816/stripe-api-calls-fail-url-validation)

Related:
- PR #4764 added diagnostic logging that made this easier to observe (merged).
- GitHub issue [#4801](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4801)

## Changes

- `includes/class-wc-stripe-api.php` — `request()` and `retrieve()` now use `wp_remote_post`/`wp_remote_get` against `self::ENDPOINT . $api` (constant is `https://api.stripe.com/v1/`).
- `includes/admin/class-wc-rest-stripe-account-keys-controller.php` — `test_account_keys()` now uses `wp_remote_post`/`wp_remote_get` against the hardcoded `https://api.stripe.com/v1/tokens[/…]` URLs.
- `tests/phpunit/class-wc-stripe-api-test.php` — two regression tests asserting `reject_unsafe_urls` is not set on the POST and GET paths.
- `tests/phpunit/class-wc-rest-stripe-account-keys-controller-test.php` — regression test covering both calls in `test_account_keys()`.
- `changelog.txt` and `readme.txt` — 10.7.0 Fix entry.

The rest of the WordPress HTTP pipeline (timeouts, SSL verification, user filters on `http_request_args`) is unchanged because both the safe and non-safe wrappers share the same underlying `WP_Http::request()` code path — `wp_safe_remote_*` only adds `reject_unsafe_urls => true`, which is what triggers `wp_http_validate_url()`.

## Testing

- `./vendor/bin/phpunit --filter 'WC_Stripe_API_Test|WC_REST_Stripe_Account_Keys_Controller_Test'` — 55 tests, 148 assertions, all passing.
- `./vendor/bin/phpcs --standard=phpcs.xml.dist -n` on the changed files — clean.
- Manual reproduction of the original failure mode (DNS redirect of `api.stripe.com` to `127.0.0.1` per the instructions on #4764) was not re-run locally; CI and QIT should exercise the existing paths.

## Notes


includes/admin/class-wc-rest-stripe-account-keys-controller.php
```diff
@@ -381,7 +384,10 @@ public function test_account_keys( WP_REST_Request $request ) {
 		$response_data = json_decode( $response['body'] );
 		$token_id      = $response_data->id;
 
-		$response = wp_safe_remote_get(
+		// Intentionally uses wp_remote_get() (not wp_safe_remote_get()): the URL is the
+		// hardcoded Stripe endpoint, and wp_http_validate_url() can reject it on hosts with
+		// flaky DNS resolution. See STRIPE-816.
+		$response = wp_remote_get(
```
daledupreez: I have two comments here that also apply to other areas of the code:
* Overall, I think we should tighten up the comments to focus on the why.
* We should reference the public issue rather than the internal issue reference: https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4801
```suggestion
		// Use wp_remote_post() instead of wp_safe_remote_post() as we have a hard-coded URL
		// and the safe version fails when there are DNS resolution issues.
		// See https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4801
		$response = wp_remote_post(
'https://api.stripe.com/v1/tokens',
[
'method'  => 'POST',
@@ -381,7 +384,10 @@ public function test_account_keys( WP_REST_Request $request ) {
$response_data = json_decode( $response['body'] );
$token_id      = $response_data->id;

		// Use wp_remote_post() instead of wp_safe_remote_post() as we have a hard-coded URL
		// and the safe version fails when there are DNS resolution issues.
		// See https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4801
		$response = wp_remote_get(
```

tests/phpunit/class-wc-stripe-api-test.php
```diff
@@ -202,6 +202,57 @@ public function mock_unauthorized_response() {
 		];
 	}
 
+	/**
+	 * Regression test for STRIPE-816: the POST path must not use wp_safe_remote_*,
+	 * which would trigger wp_http_validate_url() and fail when the host's DNS is flaky.
```
daledupreez: I think we should reference that the regression test is for this PR.
```suggestion
	 * Regression test for https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5337
	 * We must not use wp_safe_remote_post(), as the calls can fail when the host's DNS resolution is flaky.
```

tests/phpunit/class-wc-stripe-api-test.php
```diff
@@ -202,6 +202,57 @@ public function mock_unauthorized_response() {
...
+
+		$this->assertIsArray( $captured_args );
+		$this->assertNotTrue(
+			$captured_args['reject_unsafe_urls'] ?? false,
+			'Stripe API POST requests must not set reject_unsafe_urls (see STRIPE-816).'
+		);
+	}
+
+	/**
+	 * Regression test for STRIPE-816: the GET path must not use wp_safe_remote_*.
```
daledupreez: As above RE the reference.
```suggestion
	 * Regression test for https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5337
	 * We must not use wp_safe_remote_get() as the calls can fail when the host's DNS resolution is flaky.
```

tests/phpunit/class-wc-stripe-api-test.php
```diff
@@ -202,6 +202,57 @@ public function mock_unauthorized_response() {
...
+		try {
+			WC_Stripe_API::request( [], 'test_endpoint', 'POST' );
+		} finally {
+			remove_filter( 'pre_http_request', $capture_filter, 10 );
+		}
+
+		$this->assertIsArray( $captured_args );
+		$this->assertNotTrue(
+			$captured_args['reject_unsafe_urls'] ?? false,
+			'Stripe API POST requests must not set reject_unsafe_urls (see STRIPE-816).'
```
daledupreez: No need for the additional reference here.
```suggestion
			'Stripe API POST requests must not set reject_unsafe_urls.'
```

tests/phpunit/class-wc-stripe-api-test.php
```diff
@@ -202,6 +202,57 @@ public function mock_unauthorized_response() {
...
+		try {
+			WC_Stripe_API::retrieve( 'test_endpoint' );
+		} finally {
+			remove_filter( 'pre_http_request', $capture_filter, 10 );
+		}
+
+		$this->assertIsArray( $captured_args );
+		$this->assertNotTrue(
+			$captured_args['reject_unsafe_urls'] ?? false,
+			'Stripe API GET requests must not set reject_unsafe_urls (see STRIPE-816).'
```
daledupreez: As above, we don't need the reference here.
```suggestion
			'Stripe API GET requests must not set reject_unsafe_urls.'
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.7.0
+* Fix - Stop Stripe API calls from failing when WordPress URL validation rejects api.stripe.com due to intermittent DNS resolution
```
daledupreez: I think we should simplify the changelog entry a bit. (Same should apply in `readme.txt`.)
```suggestion
* Fix - Prevent Stripe API calls from failing due to DNS resolution issues
```

daledupreez (commented):
@rakshitwesley, this is looking good from a technical perspective, but I do have a number of suggestions that are mostly focused on comments and the like.

Please let me know if you can take those on, or if I can step in to help, as I think it would be good to get this into the next plugin release.

thuautp:
Thanks for the suggestions @daledupreez! @rakshitwesley and I were working together on this PR during our Bug blitz session. I've pushed the changes based on your comments.

tests/phpunit/class-wc-rest-stripe-account-keys-controller-test.php
```diff
@@ -242,4 +242,56 @@ public function provide_test_validate_api_key(): array {
...
+			$captured_args_by_url['https://api.stripe.com/v1/tokens']['reject_unsafe_urls'] ?? false,
+			'POST to https://api.stripe.com/v1/tokens must not set reject_unsafe_urls (see STRIPE-816).'
+		);
+
+		$get_url = 'https://api.stripe.com/v1/tokens/tok_visa_test';
+		$this->assertArrayHasKey( $get_url, $captured_args_by_url );
+		$this->assertNotTrue(
+			$captured_args_by_url[ $get_url ]['reject_unsafe_urls'] ?? false,
+			'GET to https://api.stripe.com/v1/tokens/{id} must not set reject_unsafe_urls (see STRIPE-816).'
+		);
```
daledupreez, line 295: We can remove the Linear references in these tests as well.
```suggestion
		$this->assertArrayHasKey( 'https://api.stripe.com/v1/tokens', $captured_args_by_url );
		$this->assertNotTrue(
			$captured_args_by_url['https://api.stripe.com/v1/tokens']['reject_unsafe_urls'] ?? false,
			'POST to https://api.stripe.com/v1/tokens must not set reject_unsafe_urls'
		);

		$get_url = 'https://api.stripe.com/v1/tokens/tok_visa_test';
		$this->assertArrayHasKey( $get_url, $captured_args_by_url );
		$this->assertNotTrue(
			$captured_args_by_url[ $get_url ]['reject_unsafe_urls'] ?? false,
			'GET to https://api.stripe.com/v1/tokens/{id} must not set reject_unsafe_urls'
		);
```

daledupreez (approved):
Thanks for the updates, @thuautp and @rakshitwesley!

This is now looking good to me, though I have one non-blocking suggestion to remove two Linear references from the unit tests.

tests/phpunit/class-wc-rest-stripe-account-keys-controller-test.php
daledupreez: I ended up just applying this change -- there is no need for an additional update/review cycle for something this small. 😁

PR #5338, opened 2026-04-23
Fix STRIPE-1045: make tooltip link clickable and fix contrast

IliaDi:
## Summary

In the account status section, when payouts or payments are disabled, a help icon shows a tooltip with an \"Update now\" link pointing to the Stripe dashboard. Two issues were present:

1. **Clicking \"Update now\" did nothing** 
2. **Link was hard to read** 

**Screenshot**
Screenshot 2026-04-23 at 11 20 23

## Bug

[STRIPE-1045: Bug: Non-clickable link in account status warning and contrast issue](https://linear.app/a8c/issue/STRIPE-1045/bug-non-clickable-link-in-account-status-warning-and-contrast-issue)

## Changes

- `client/components/tooltip/index.js` — skip `preventDefault` (and toggle logic) in the button's `onClick` when the click originates from an anchor element inside the tooltip
- `client/components/tooltip/style.scss` — add explicit `color: white; text-decoration: underline` for links inside the tooltip, ensuring adequate contrast regardless of WordPress admin CSS overrides
- `client/components/tooltip/test/index.test.js` — add regression test verifying that clicking a link inside tooltip content does not close/toggle the tooltip

## Testing

- All existing tooltip tests pass; new regression test passes
- With manual testing this is 
Screenshot 2026-04-23 at 12 09 41
the result: 

## Notes

This PR was created during the [Woo Bug Blitz](https://woohappyp2.wordpress.com/2026/03/18/extensions-bug-blitz/) with Claude Code assistance by a Happiness Engineer who encounters this bug in support tickets.


IliaDi:
> client/components/tooltip/index.js (1)
> 
> > `16-25`: _⚠️ Potential issue_ | _🟡 Minor_
> > **Accessibility: nested `<a>` inside `<button>` is invalid HTML.**
> > The early-return correctly fixes the navigation/contrast bug, but note that the tooltip's outer element is a `<button>` (line 33) and the content now commonly contains an anchor. Nesting interactive elements is invalid per the HTML spec and can cause inconsistent focus/keyboard behavior across browsers and assistive tech (e.g., Tab order, Enter activation ambiguity). Not a blocker for this fix, but worth tracking as a follow-up — consider changing the trigger wrapper from `<button>` to a non-interactive element with appropriate ARIA, or rendering the link via a popover that is a sibling rather than a descendant of the button.
> > Also consider guarding against `event.target` not being an `Element` (e.g., defensive `typeof event.target.closest === 'function'`) — low priority since React synthetic click targets are always Elements in practice.

I've opened a sub-issue for this https://linear.app/a8c/issue/STRIPE-1101/a11y-tooltip-trigger-button-contains-interactive-descendants-invalid

daledupreez (approved):
Thanks for the fix, @IliaDi and @ivonabrajovic! The changes look much better. I'll get this merged shortly.

PR #5344, opened 2026-04-23
[RSM - Diagnostics Mode] Add diagnostics filters: wc_stripe_request_response and wc_stripe_localized_data

wjrosa:
Towards RSM-629

Companion to the frontend recorder in #5330.

Adds two foundational filters that later PRs in this track (trace store, recorder singleton, redaction engine, webhook correlation) will subscribe to. This PR is intentionally the smallest still-functional slice: pure additions, no behavior change, no new classes, no REST endpoints, no storage.

### 1. `wc_stripe_request_response`

New filter in `WC_Stripe_API::request()`, applied to the decoded response body before return.

```php
apply_filters( 'wc_stripe_request_response', $response_body, $api, $method, $request );
```

Args:
- `$response_body` — decoded response body (mixed: object, array, scalar)
- `$api` — Stripe endpoint (`'payment_intents'`, `'setup_intents'`, etc.)
- `$method` — HTTP method
- `$request` — the request body after `wc_stripe_request_body` has run

Applied in the single return path covering both `$with_headers = true` and `$with_headers = false` callers, so every outbound Stripe API response surfaces through it once.

### 2. `wc_stripe_localized_data`

New filter wrapping each `wp_localize_script` call site in the plugin's UPE and express-checkout enqueue paths.

```php
apply_filters( 'wc_stripe_localized_data', $data, $script_handle, $object_name );
```

Wrapped sites:
- `abstract-wc-stripe-payment-gateway.php` → `woocommerce_stripe` / `wc_stripe_params`
- `class-wc-stripe-upe-payment-gateway.php` → `wc-stripe-upe-classic` / `wc_stripe_upe_params`
- `class-wc-stripe-express-checkout-element.php` → `wc_stripe_express_checkout` / `wc_stripe_express_checkout_params` (plus the pay-for-order `wcStripeExpressCheckoutPayForOrderParams` site)

Subscribers can either mutate the data array in place, or use the filter as a hook point to attach a separate global (e.g., `wcStripeDiag`) to the same script handle via an additional `wp_localize_script` call.

### Note on Blocks

`includes/class-wc-stripe-blocks-support.php` does **not** use `wp_localize_script` — Blocks ships data via `get_payment_method_data()` consumed through the WC Blocks extensibility API. There is no `wp_localize_script` call site in that path, so this PR does not touch it. When the recorder lands in a follow-up, the Blocks surface can be covered either by extending the existing `wc_stripe_params` filter (already applied inside `get_gateway_javascript_params()`) or by adding a dedicated filter on `get_payment_method_data()`. Flagging this explicitly so reviewers don't expect a Blocks change here.

## Testing instructions

These filters are currently unused by any subscriber in the plugin. Verify by smoke-testing:

1. `npm run up`, go through classic and Blocks checkout, place an order. No observable behavior change.
2. Drop a temporary `add_filter( 'wc_stripe_request_response', fn( $r, $api ) => $r, 10, 2 )` into a must-use plugin and confirm via logging that the callback fires for each outbound Stripe API call.
3. Similarly add a `wc_stripe_localized_data` callback that logs `$script_handle` and `$object_name`, then load a product page with express checkout, the classic checkout, and the pay-for-order page — confirm the filter fires for each relevant handle.

## Test coverage

- New PHPUnit test `test_request_response_filter_fires_and_can_mutate_body` in `tests/phpunit/class-wc-stripe-api-test.php` asserts the new response filter fires with correct args and that its return value replaces the body.
- The `wc_stripe_localized_data` sites are mechanical filter-then-pass; exercised end-to-end by the recorder PR that subscribes to them.

Ran:
- `npm run lint:php` on touched files — clean
- `npm run phpstan` — `[OK] No errors`
- `./bin/run-tests.sh --filter 'WC_Stripe_API_Test'` — 23/23 pass

---

- [x] Tested on mobile (or does not apply) — N/A, no UI

### Changelog entry

Changelog added to `changelog.txt` and `readme.txt`:
> Dev - Add wc_stripe_request_response and wc_stripe_localized_data filters to pave the way for the upcoming diagnostics recorder

## Next steps in the RSM-629 track (not in this PR)

1. Trace store — file-backed writer (one JSON file per trace under `uploads/wc-stripe-diagnostics/`) with index + FIFO cap + per-trace size caps
2. Redaction engine — allow-list registry + redaction-audit test
3. `WC_Stripe_Diagnostics_Recorder` singleton — subscribes to the filters added here + `wc_stripe_request_headers`, `wc_stripe_request_body`, `wc_stripe_webhook_received`; injects `metadata[wc_diag_session_id]` on PI / SetupIntent / Customer create
4. Action Scheduler jobs — daily 7-day cleanup, 5-min `pending → abandoned` promotion

### DIFF
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -2230,11 +2230,24 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 
 		wp_register_script( 'woocommerce_stripe', plugins_url( 'assets/js/stripe' . $suffix . '.js', WC_STRIPE_MAIN_FILE ), [ 'jquery-payment', 'stripe' ], WC_STRIPE_VERSION, true );
 
-		wp_localize_script(
-			'woocommerce_stripe',
-			'wc_stripe_params',
-			apply_filters( 'wc_stripe_params', $this->javascript_params() )
-		);
+		$wc_stripe_params = apply_filters( 'wc_stripe_params', $this->javascript_params() );
+
+		/**
+		 * Filters the localized data passed to a Stripe frontend script.
+		 *
+		 * Fires for every `wp_localize_script` call site in the plugin's
+		 * UPE and express checkout enqueue paths. Subscribers can modify
+		 * the data, or attach additional globals to the same handle.
+		 *
+		 * @since 10.7.0
+		 *
+		 * @param array  $data          The localized data array.
+		 * @param string $script_handle The registered script handle.
+		 * @param string $object_name   The JS variable name the data will be assigned to.
+		 */
+		$wc_stripe_params = apply_filters( 'wc_stripe_localized_data', $wc_stripe_params, 'woocommerce_stripe', 'wc_stripe_params' );
+
+		wp_localize_script( 'woocommerce_stripe', 'wc_stripe_params', $wc_stripe_params );
 
 		$this->tokenization_script();
 		wp_enqueue_script( 'woocommerce_stripe' );
--- a/includes/class-wc-stripe-api.php
+++ b/includes/class-wc-stripe-api.php
@@ -282,7 +282,9 @@ class WC_Stripe_API {
 		);
 
 		/**
-		 * Filters the decoded response body returned from the Stripe API.
+		 * Fires after the Stripe API returns a successful response, before it is
+		 * returned to the caller. Observe-only: handlers must not mutate the
+		 * response. Used by diagnostics to capture API exchanges.
 		 *
 		 * @since 10.7.0
 		 *
@@ -291,7 +293,7 @@ class WC_Stripe_API {
 		 * @param string $method        The HTTP method.
 		 * @param array  $request       The request body sent to the Stripe API.
 		 */
-		$response_body = apply_filters( 'wc_stripe_request_response', $response_body, $api, $method, $request );
+		do_action( 'wc_stripe_api_response_received', $response_body, $api, $method, $request );
 
 		if ( $with_headers ) {
 			return [
--- a/includes/class-wc-stripe-api.php
+++ b/includes/class-wc-stripe-api.php
@@ -281,6 +281,18 @@ class WC_Stripe_API {
 			]
 		);
 
+		/**
+		 * Filters the decoded response body returned from the Stripe API.
+		 *
+		 * @since 10.7.0
+		 *
+		 * @param mixed  $response_body The decoded response body.
+		 * @param string $api           The Stripe API endpoint.
+		 * @param string $method        The HTTP method.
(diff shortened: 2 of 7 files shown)


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2230,11 +2230,24 @@ public function payment_scripts() {
...
+		 * UPE and express checkout enqueue paths. Subscribers can modify
+		 * the data, or attach additional globals to the same handle.
+		 *
+		 * @since 10.7.0
+		 *
+		 * @param array  $data          The localized data array.
+		 * @param string $script_handle The registered script handle.
+		 * @param string $object_name   The JS variable name the data will be assigned to.
+		 */
+		$wc_stripe_params = apply_filters( 'wc_stripe_localized_data', $wc_stripe_params, 'woocommerce_stripe', 'wc_stripe_params' );
```
Mayisha, line 2248: QQ: why do we need this new filter here? Couldn't we achieve the result using the existing `wc_stripe_params` filter?

Mayisha: Nvm. I understand now that in other files we have filters with other names and we need a unified filter for your use case across the files.

Mayisha (approved):
Looks good to me from code inspection 👍

includes/class-wc-stripe-api.php
```diff
@@ -281,6 +281,18 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
...
+		 * Filters the decoded response body returned from the Stripe API.
+		 *
+		 * @since 10.7.0
+		 *
+		 * @param mixed  $response_body The decoded response body.
+		 * @param string $api           The Stripe API endpoint.
+		 * @param string $method        The HTTP method.
+		 * @param array  $request       The request body sent to the Stripe API.
+		 */
+		$response_body = apply_filters( 'wc_stripe_request_response', $response_body, $api, $method, $request );
```
malithsen: Question on this new filter: This filter allows mutating the Stripe response. For our purposes, we only need the ability to observe the response. So I wanted to double-check this change. Would it be safer to use `do_action` here instead?

malithsen (commented):
For diagnostics, we only need to observe the data. The filters we are introducing also allow mutating the data. Is that an intentional decision?

includes/class-wc-stripe-api.php
wjrosa: Good call — diagnostics only needs to observe, and a filter would let third parties mutate Stripe responses (extra surface we don't need). Easier to promote action → filter later than to demote a public filter.

Switched to `do_action('wc_stripe_api_response_received', ...)` in a44d2f97a; renamed to make the observe-only intent explicit, and the test now asserts the response returned to the caller is unaffected by listeners.

malithsen (approved):
LGTM!

PR #5348, opened 2026-04-24
[RSM - Diagnostics Mode] Add WC_Stripe_Diagnostics_Redactor

wjrosa:
Towards RSM-629

Adds `WC_Stripe_Diagnostics_Redactor`: the pre-store scrubbing layer every diagnostics event passes through before it lands in the trace store. Pure class + tests + changelog. No subscribers or runtime effect yet.

The class lives under `includes/diagnostics/`, which is in composer's classmap (added in #5363), so no `require_once` wiring is needed.

### Two layers, fail-closed

1. **Allow-list registry per event kind.** Known kinds: `apiRequest`, `apiResponse`, `webhookReceived`, `consoleMessage`, `elementEvent`, `sdkCall`, `blocksEvent`, `expressCheckout`, `sessionEnd`. Each kind has an explicit list of allow-listed field paths (dot-notation for nested fields). Any field not on the list is dropped. Unknown kinds return an empty event.

2. **String scrubbers** run on every remaining value:
   - Emails → `[email]`
   - IPv4 and IPv6 (including `::`-compressed forms) → `[ip]`
   - Stripe-shaped secrets (`sk_live_…`, `rk_live_…`, `whsec_…`) → `[secret]`
   - URL query strings stripped (path kept, `?…` replaced with `?[redacted]`)
   - Strings truncated at 512 chars

The bias is intentionally conservative: it is far cheaper to drop a legitimate field than to ship PII in a support bundle.

### Redaction-audit test (per parent issue acceptance criterion)

`test_redaction_audit_no_pii_survives_for_any_allow_listed_field` iterates **every kind × every field × every PII pattern we know about**, injects the PII, and asserts the serialized post-redaction JSON does not contain the payload. This is the gating test for the parent issue's "zero PII patterns survive the recorder boundary" acceptance criterion.

### How can this code break?

- **A new event kind ships without an allow list:** unknown kinds fail closed (empty event), so the worst case is a missing event in a trace, not a leaked field.
- **A future scrubber regex over-matches:** the redaction-audit test catches this for every known PII pattern; new patterns should be added to the parameterized data provider so the same matrix protects them.
- **A nested allow-listed field contains a value type the scrubbers don't expect (e.g. binary):** scrubbers operate only on strings; non-string scalars pass through untouched, which is acceptable because they can't carry PII the regexes target.

## Testing instructions

No runtime effect yet — no subscribers. To verify:

1. Run the new PHPUnit suite:
   ```bash
   npm run test:php -- --filter 'WC_Stripe_Diagnostics_Redactor_Test'
   # → OK (10 tests, 344 assertions)
   ```
2. Smoke-test in a WP shell:
   ```php
   $r = new WC_Stripe_Diagnostics_Redactor();
   $r->redact( [ 'kind' => 'consoleMessage', 'level' => 'warn', 'message_truncated' => 'user shopper@example.com saw error' ] );
   // → ['kind' => 'consoleMessage', 'level' => 'warn', 'message_truncated' => 'user [email] saw error']
   ```
3. `npm run phpstan` → expect `[OK] No errors`
4. `npm run lint:php`

---

-   [x] Tested on mobile (or does not apply) — does not apply (pure PHP function, no UI)

### Changelog entry

* Dev - Add WC_Stripe_Diagnostics_Redactor (allow-list registry + PII scrubbers) for the upcoming diagnostics recorder

**Post merge**

---

## Stack

- ✅ #5344 (request/response + localized_data filters) — **merged**
- ✅ #5347 (file-backed trace store) — **merged**
- ✅ #5363 (composer-autoload `includes/diagnostics/`) — **merged into `feature/diagnostics`**
- 👉 **This PR** — adds `WC_Stripe_Diagnostics_Redactor`, base `feature/diagnostics`
- 🔄 #5349 (recorder) — open, depends on this
- 🔄 #5350 (REST controller) — open, depends on this and on #5349

## Assumption to flag

The allow list currently lives in `default_allow_list()` and matches `docs/diagnostics-contracts.md §5`. That doc doesn't exist yet — it will land with the recorder PR. For now the allow list reflects the frontend recorder's event shapes in #5330. Reviewers may want to reconcile both once the contracts doc lands.

### DIFF
--- a/includes/diagnostics/class-wc-stripe-diagnostics-redactor.php
+++ b/includes/diagnostics/class-wc-stripe-diagnostics-redactor.php
@@ -7,10 +7,10 @@ defined( 'ABSPATH' ) || exit;
  *
  * Two layers run against every event before it reaches the trace store:
  *
- * 1. An **allow-list registry** keyed by event kind (`apiRequest`,
- *    `apiResponse`, `webhookReceived`, `consoleMessage`, ...). Any field not
- *    explicitly listed for the event's kind is dropped. Unknown kinds produce
- *    an empty event (fail-closed).
+ * 1. An **allow-list registry** keyed by event kind (`element.ready`,
+ *    `console.warn`, `stripe.confirmPayment.invoke`, `webhook.received`, ...).
+ *    Any field not explicitly listed for the event's kind is dropped. Unknown
+ *    kinds produce an empty event (fail-closed).
  *
  * 2. A set of **string scrubbers** applied to every remaining value: emails
  *    become `[email]`, IPv4/IPv6 become `[ip]`, URL query strings are
@@ -97,27 +97,70 @@ class WC_Stripe_Diagnostics_Redactor {
 	 * @return array<string, string[]>
 	 */
 	public static function default_allow_list(): array {
-		return [
-			'apiRequest'      => [
+		$list = [
+			// Stripe Element lifecycle (frontend: client/diagnostics/recorder.js).
+			'element.ready'                 => [ 'element_type' ],
+			'element.focus'                 => [ 'element_type' ],
+			'element.blur'                  => [ 'element_type' ],
+			'element.change'                => [
+				'element_type',
+				'complete',
+				'empty',
+				'valid',
+				'brand',
+				'error_code',
+				'error_type',
+			],
+			'element.loaderror'             => [
+				'element_type',
+				'error_code',
+				'error_type',
+				'error_message',
+			],
+
+			// Blocks onPaymentSetup brackets.
+			'blocks.payment_setup.start'    => [ 'site' ],
+			'blocks.payment_setup.end'      => [
+				'site',
+				'duration_ms',
+				'result_type',
+				'error_message',
+			],
+
+			// Express Checkout Element events.
+			'express.click'                 => [ 'wallet_type' ],
+			'express.confirm'               => [ 'wallet_type', 'payment_method_type' ],
+			'express.cancel'                => [ 'wallet_type' ],
+			'express.shippingratechange'    => [ 'wallet_type' ],
+			'express.shippingaddresschange' => [ 'wallet_type', 'country' ],
+			'express.paymentmethod'         => [ 'wallet_type', 'payment_method_type' ],
+
+			// Parent-frame console interception.
+			'console.warn'                  => [ 'message', 'source' ],
+			'console.error'                 => [ 'message', 'source' ],
+
+			// Server-side: Stripe API request / response.
+			'stripe.api.request'            => [
 				'api',
 				'method',
-				'request_id',
-				'fields',
 				'has_metadata',
 				'metadata.order_id',
 				'metadata.wc_diag_session_id',
+				'fields',
 			],
-			'apiResponse'     => [
+			'stripe.api.response'           => [
 				'api',
 				'method',
 				'status',
 				'request_id',
+				'latency_ms',
 				'error.code',
 				'error.type',
 				'error.decline_code',
-				'latency_ms',
 			],
-			'webhookReceived' => [
+
+			// Server-side: webhook receipt (correlated via metadata session id).
+			'webhook.received'              => [
 				'type',
 				'status',
 				'intent_id',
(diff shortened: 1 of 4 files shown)


includes/diagnostics/class-wc-stripe-diagnostics-redactor.php
```diff
@@ -0,0 +1,252 @@
...
+		return $value;
+	}
+
+	/**
+	 * The default allow list. Keep this small — adding a field is an explicit
+	 * statement that it's safe to ship to a diagnostics bundle.
+	 *
+	 * @return array<string, string[]>
+	 */
+	public static function default_allow_list(): array {
```
malithsen, line 99: This list doesn't match what the frontend sends currently, so the events are dropped. 

Untested fix from Claude 👇 

```
 public static function default_allow_list(): array {
      $list = [
          // -------------------------------------------------------------------
          // Stripe Element lifecycle (frontend: client/diagnostics/recorder.js)
          // -------------------------------------------------------------------
          'element.ready'     => [ 'element_type' ],
          'element.focus'     => [ 'element_type' ],
          'element.blur'      => [ 'element_type' ],
          'element.change'    => [
              'element_type',
              'complete',
              'empty',
              'valid',
              'brand',
              'error_code',
              'error_type',
          ],
          'element.loaderror' => [
              'element_type',
              'error_code',
              'error_type',
              'error_message',
          ],

          // -------------------------------------------------------------------
          // Blocks onPaymentSetup brackets
          // -------------------------------------------------------------------
          'blocks.payment_setup.start' => [ 'site' ],
          'blocks.payment_setup.end'   => [
              'site',
              'duration_ms',
              'result_type',
              'error_message',
          ],

          // -------------------------------------------------------------------
          // Express Checkout Element events
          // -------------------------------------------------------------------
          'express.click'                  => [ 'wallet_type' ],
          'express.confirm'                => [ 'wallet_type', 'payment_method_type' ],
          'express.cancel'                 => [ 'wallet_type' ],
          'express.shippingratechange'     => [ 'wallet_type' ],
          'express.shippingaddresschange'  => [ 'wallet_type', 'country' ],
          'express.paymentmethod'          => [ 'wallet_type', 'payment_method_type' ],

          // -------------------------------------------------------------------
          // Parent-frame console interception
          // -------------------------------------------------------------------
          'console.warn'  => [ 'message', 'source' ],
          'console.error' => [ 'message', 'source' ],

          // -------------------------------------------------------------------
          // Server-side: Stripe API request / response
          // -------------------------------------------------------------------
          'stripe.api.request'  => [
              'api',
              'method',
              'has_metadata',
              'metadata.order_id',
              'metadata.wc_diag_session_id',
              'fields',
          ],
          'stripe.api.response' => [
              'api',
              'method',
              'status',
              'request_id',
              'latency_ms',
              'error.code',
              'error.type',
              'error.decline_code',
          ],

          // -------------------------------------------------------------------
          // Server-side: webhook receipt (correlated via metadata session id)
          // -------------------------------------------------------------------
          'webhook.received' => [
              'type',
              'status',
              'intent_id',
              'charge_id',
              'order_id',
              'session_id',
          ],
      ];

      // Stripe SDK call brackets emitted by the frontend recorder via
      // `diagnostics.aroundStripeCall( method, fn )`. Each method gets
      // .invoke / .resolve / .throw with method-agnostic shapes.
      $sdk_call_methods = [
          'confirmPayment',
          'confirmSetup',
          'confirmCashappSetup',
          'confirmCashappPayment',
          'createPaymentMethod',
          'confirm', // Custom Checkout Sessions (Blocks)
      ];
      foreach ( $sdk_call_methods as $method ) {
          $list[ "stripe.{$method}.invoke" ]  = [ 'method' ];
          $list[ "stripe.{$method}.resolve" ] = [
              'method',
              'intent_status',
              'payment_method_type',
              'has_error',
              'error_type',
              'error_code',
              'error_decline_code',
          ];
          $list[ "stripe.{$method}.throw" ] = [
              'method',
              'error_message',
          ];
      }

      return $list;
  }
```

malithsen (changes requested):
I think we should update the allow list

includes/diagnostics/class-wc-stripe-diagnostics-redactor.php
wjrosa: Good catch — the allow list pre-dated the recorder landing in #5330 and never got re-aligned, so every frontend event was failing closed. Applied your fix in 76d2608, verified each kind/field against `client/diagnostics/recorder.js` and the six methods used at `aroundStripeCall()` call sites. Tests updated, phpunit/phpcs/phpstan green.

includes/diagnostics/class-wc-stripe-diagnostics-redactor.php
```diff
@@ -0,0 +1,293 @@
...
+			'/\b(?:[A-Fa-f0-9]{0,4}:){2,7}[A-Fa-f0-9]{0,4}\b/'     => '[ip]',
+			// Stripe secret-ish tokens (sk_live_…, rk_live_…, sk_test_…, whsec_…).
+			'/\b(?:sk|rk|whsec)_(?:live|test)?_?[A-Za-z0-9]{10,}/' => '[secret]',
+		];
+		foreach ( $patterns as $pattern => $replacement ) {
+			$replaced = preg_replace( $pattern, $replacement, $value );
+			if ( is_string( $replaced ) ) {
+				$value = $replaced;
+			}
+		}
```
malithsen, line 233: Do we currently record any of these PII fields? 
Looking at the trace logs, it doesn't look like we do. So these might not be needed?

malithsen (approved):
Looks good on inspection. I'm not sure if scrubbing is necessary but won't hurt to have it.

includes/diagnostics/class-wc-stripe-diagnostics-redactor.php
wjrosa: Allow-listed free-form strings can carry these patterns even though happy-path traces don't:

- `console.warn` / `console.error` → `message` (anything passed to `console.error` in app code).
- `element.loaderror`, `blocks.payment_setup.end`, `stripe.{method}.throw` → `error_message` (Stripe error messages occasionally include emails or addresses they echoed back from a request).

The allow list can't be more selective on free-form text, so the scrubbers are the second-line defense — they kick in only when one of those error paths actually fires.

PR #5349, opened 2026-04-24
[RSM - Diagnostics Mode] Add WC_Stripe_Diagnostics_Recorder (server-side)

wjrosa:
Towards RSM-630
Fixes RSM-631

Adds `WC_Stripe_Diagnostics_Recorder`: the server-side subscriber that connects the filters + store + redactor into a functioning recorder. Pure class + tests + changelog. The class lives under `includes/diagnostics/` (autoloaded via composer's classmap from #5363), so no `require_once` plumbing is added; only an `init()` call from `WC_Stripe::init()` to register the hook subscriptions.

### Subscriptions

| Hook | Action |
|---|---|
| `wc_stripe_request_body` (filter) | Injects `metadata[wc_diag_session_id]` into `payment_intents` / `setup_intents` / `customers` requests when a session is active. Records an `apiRequest` event. |
| `wc_stripe_api_response_received` (action) | Records `apiResponse` with computed latency and the Stripe error shape when present. |
| `wc_stripe_webhook_received` (action) | Correlates the webhook back to its trace via `metadata.wc_diag_session_id`. If the webhook arrives before any other activity, a new trace is created (matches the parent issue's "first-wins decrement" rule). |
| `wc_stripe_localized_data` (filter) | Publishes `wcStripeDiag` (`{active, sessionId, nonce, endpoint}`) alongside the plugin's existing localized params so the frontend recorder in #5330 can boot. |

`wc_stripe_api_response_received` is the action `WC_Stripe_API::request()` actually fires (added in #5344). The earlier draft of this PR subscribed to a phantom `wc_stripe_request_response` hook; that's been corrected and `test_init_subscribes_to_real_api_hooks` now exercises the real `do_action` after `init()` so future drift between this class and the API hook surface fails in CI instead of silently dropping events.

### Session lifecycle

- `start_session( $id )` is how the REST endpoint (#5350) hands off the session id from the client. Rejects when the store is at the 200-trace cap.
- `current_session_id()` reads from an in-memory field first, then falls back to a 30-minute transient so subsequent requests in the same checkout resume the same session id.

### Dependency injection

The class has an explicit constructor accepting the trace store and redactor. Tests inject fakes; production code calls `get_instance()`, which constructs the real dependencies once.

### Metadata-injection scope

Only `payment_intents`, `setup_intents`, `customers`. These are the Stripe resources whose `metadata` survives in webhook payloads, which is what the correlation relies on.

### How can this code break?

- **Stripe API class introduces a new response hook name:** the wiring smoke test fails, surfacing the drift before the recorder silently stops capturing.
- **Two requests within the same session race:** the trace store serializes its writers via `flock(LOCK_EX)`, so concurrent `append_event()` calls don't clobber each other.
- **A webhook arrives 30+ minutes after the last request (transient expired):** `current_session_id()` returns `null`, no event is recorded — by design (the "session ended naturally" path).
- **Store is at 200-trace cap when `start_session()` runs:** the call returns `false` and the caller (REST endpoint) gets a chance to react instead of having an in-flight session silently fail later.

## Testing instructions

No runtime effect yet — the only call to `start_session()` lives in the upcoming REST endpoint (#5350). To verify the wiring on its own:

1. Run the recorder suite:
   ```bash
   npm run test:php -- --filter 'WC_Stripe_Diagnostics_Recorder_Test'
   # → OK (13 tests, 24 assertions)
   ```
2. Run the full diagnostics suite to confirm nothing else regressed:
   ```bash
   npm run test:php -- --filter 'WC_Stripe_Diagnostics'
   ```
3. End-to-end smoke test (requires this PR + #5350 + the frontend in #5330 cherry-picked):
   - In a WP shell:
     ```php
     WC_Stripe_Diagnostics_Recorder::get_instance()->start_session( 'smoke-' . bin2hex( random_bytes( 4 ) ) );
     ```
   - Reload the checkout page. The `wcStripeDiag` global should be present on `window`.
   - Complete a test checkout. Look in `wp-content/uploads/wc-stripe-diagnostics/<id>.json` — it should hold `apiRequest`, `apiResponse`, and `webhookReceived` events.
4. `npm run phpstan` → expect `[OK] No errors` (this PR is currently based on `add/diagnostics-redaction` so the redactor is present at build time; will retarget to `feature/diagnostics` once #5348 merges).

---

-   [x] Tested on mobile (or does not apply) — does not apply (server-only)

### Changelog entry

* Dev - Add WC_Stripe_Diagnostics_Recorder server-side singleton that injects wc_diag_session_id into Stripe metadata and records redacted request/response/webhook events

**Post merge**

---

## Stack

- ✅ #5344 (request/response + localized_data filters) — **merged**
- ✅ #5347 (file-backed trace store) — **merged**
- ✅ #5363 (composer-autoload `includes/diagnostics/`) — **merged into `feature/diagnostics`**
- 🔄 #5348 (redactor) — **open**, must merge first; this PR is temporarily based on `add/diagnostics-redaction` so CI has the redactor class available
- 👉 **This PR** — adds the server-side recorder, base `add/diagnostics-redaction` (will retarget to `feature/diagnostics` once #5348 merges)
- 🔄 #5350 (REST controller) — open, depends on this
- ⏭️ Admin toggle PR — flips `wc_stripe_diagnostics_enabled`

## Known follow-ups (not in this PR)

- `start_session()` exists but has no caller until the REST endpoint (#5350) lands, so nothing is actually recorded during checkouts on this PR alone.
- The `wcStripeDiag` nonce uses `wp_create_nonce( 'wc_stripe_diagnostics' )`. The REST endpoint verifies this nonce on the receiving side.

### DIFF
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -319,11 +319,15 @@ class WC_Stripe {
 	}
 
 	/**
-	 * Initialize diagnostics components. The recorder's `init()` is itself
-	 * gated on the diagnostics toggle, so this method is safe to call
-	 * unconditionally.
+	 * Initialize diagnostics components: the frontend loader (which emits
+	 * `wcStripeDiag` for the recorder.js boot) and the server-side recorder
+	 * (which subscribes to the Stripe API + webhook hooks). The recorder's
+	 * own `init()` is gated on the diagnostics toggle; the frontend loader
+	 * gates per-request inside `should_localize()`. Both are safe to call
+	 * unconditionally here.
 	 */
 	public function initialize_diagnostics(): void {
+		( new WC_Stripe_Diagnostics_Frontend_Loader() )->init();
 		WC_Stripe_Diagnostics_Recorder::get_instance()->init();
 	}
 
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -861,6 +861,9 @@ class WC_Stripe {
 
 		$exit_survey_controller = new WC_REST_Stripe_Exit_Survey_Controller();
 		$exit_survey_controller->register_routes();
+
+		$diagnostics_controller = new WC_REST_Stripe_Diagnostics_Controller();
+		$diagnostics_controller->register_routes();
 	}
 
 	/**
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -134,6 +134,7 @@ class WC_Stripe {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-database-cache.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-payment-method-configurations.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-database-cache-prefetch.php';
+		WC_Stripe_Diagnostics_Recorder::get_instance()->init();
 		include_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-api.php';
 		include_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-mode.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/compat/class-wc-stripe-subscriptions-helper.php';
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -134,7 +134,6 @@ class WC_Stripe {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-database-cache.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-payment-method-configurations.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-database-cache-prefetch.php';
-		WC_Stripe_Diagnostics_Recorder::get_instance()->init();
 		include_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-api.php';
 		include_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-mode.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/compat/class-wc-stripe-subscriptions-helper.php';
@@ -292,6 +291,8 @@ class WC_Stripe {
 
 		add_action( 'init', [ $this, 'initialize_apple_pay_registration' ] );
 
+		add_action( 'init', [ $this, 'initialize_diagnostics' ] );
+
 		// Initialize Agentic Commerce integration.
 		add_action( 'woocommerce_init', [ $this, 'initialize_agentic_commerce' ] );
 
@@ -315,6 +316,15 @@ class WC_Stripe {
 	 */
 	public function initialize_apple_pay_registration() {
 		new WC_Stripe_Apple_Pay_Registration();
+	}
+
+	/**
+	 * Initialize diagnostics components. The recorder's `init()` is itself
+	 * gated on the diagnostics toggle, so this method is safe to call
+	 * unconditionally.
+	 */
+	public function initialize_diagnostics(): void {
+		WC_Stripe_Diagnostics_Recorder::get_instance()->init();
 	}
 
 	/**
--- a/includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
+++ b/includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
@@ -1,0 +1,386 @@
+<?php
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Server-side diagnostics recorder.
+ *
+ * Subscribes to the plugin's existing request and webhook hooks. When a
+ * diagnostics session is active it:
+ *
+ * - Injects `metadata[wc_diag_session_id]` into outgoing payment-intent,
+ *   setup-intent, and customer create/update requests so webhooks can be
+ *   correlated back to the originating trace.
+ * - Records a redacted event for each Stripe API request/response round trip.
+ * - Records a redacted event for each webhook received.
+ * - Injects a `wcStripeDiag` global alongside the plugin's existing localized
+ *   script params so the frontend recorder can boot.
+ *
+ * The recorder is a singleton so the running session id can be shared across
+ * `wc_stripe_request_body` (request capture) and
+ * `wc_stripe_api_response_received` (response capture) without threading state
+ * through every call site.
+ */
+class WC_Stripe_Diagnostics_Recorder {
+
+	const SESSION_ID_META_KEY   = 'wc_diag_session_id';
+	const SESSION_OPTION        = 'wc_stripe_diag_active_session';
+	const SESSION_TTL           = 30 * MINUTE_IN_SECONDS;
+	const METADATA_ENABLED_APIS = [ 'payment_intents', 'setup_intents', 'customers' ];
+
+	/**
+	 * Singleton instance.
+	 *
+	 * @var self|null
+	 */
+	private static $instance;
+
+	/**
+	 * Trace store dependency.
+	 *
+	 * @var WC_Stripe_Diagnostics_Trace_Store
+	 */
+	private $store;
+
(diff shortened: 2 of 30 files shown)


includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
```diff
@@ -0,0 +1,351 @@
...
+		wp_localize_script(
+			(string) $script_handle,
+			'wcStripeDiag',
+			[
+				'active'    => true,
+				'sessionId' => $session_id,
+				'nonce'     => wp_create_nonce( 'wc_stripe_diagnostics' ),
+				'endpoint'  => esc_url_raw( rest_url( 'wc/v3/wc_stripe/diagnostics/events' ) ),
+			]
+		);
```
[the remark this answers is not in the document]
wjrosa: Good catch — fixed in d638dd6. Added a `has_localized_diag` flag that short-circuits subsequent invocations, plus `test_on_localized_data_attaches_wc_stripe_diag_only_to_first_invocation` (registers two script handles, fires the filter twice, asserts only the first handle ends up with `wcStripeDiag` in its localized data).

malithsen (commented):
@wjrosa What do you think about removing the redactor implementation from here, so we can test this PR in isolation? Redactor already has a separate PR #5348

wjrosa:
@malithsen Good call — done in e1d0b614. Dropped the redactor commit; this PR's diff is now just the recorder + its test.

Re-targeted the base from `feature/diagnostics` to `add/diagnostics-redaction` (the branch behind #5348) so CI runs against a base where `WC_Stripe_Diagnostics_Redactor` exists. Once #5348 merges, I'll retarget back to `feature/diagnostics` — diff stays the same.

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
```diff
@@ -0,0 +1,386 @@
...
+			$request['metadata'][ self::SESSION_ID_META_KEY ] = $session_id;
+		}
+
+		$this->request_start_times[ $this->request_key( $api ) ] = microtime( true );
+
+		$this->store->append_event(
+			$session_id,
+			$this->redactor->redact(
+				[
+					'kind'         => 'apiRequest',
```
malithsen: This and other `kind` fields should be updated to the new format in the redactor PR or update that PR, otherwise these will later get dropped

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
```diff
@@ -0,0 +1,386 @@
...
+		if ( null !== $this->current_session_id ) {
+			return $this->current_session_id;
+		}
+		$persisted = get_transient( self::SESSION_OPTION );
+		if ( is_string( $persisted ) && '' !== $persisted ) {
+			$this->current_session_id = $persisted;
+			return $persisted;
+		}
+		return null;
+	}
```
malithsen, line 162: I don't think this transient-based approach will work for concurrent shoppers. Since it's a global transient, two shoppers will share the same session. 

We should instead use WC Sessions `wc_stripe_diag_active_session` field introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5372/c

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
```diff
@@ -0,0 +1,386 @@
...
+	 *
+	 * @param array  $data          The existing localized data.
+	 * @param string $script_handle The script handle being localized.
+	 * @param string $object_name   The JS variable name.
+	 * @return array
+	 */
+	public function on_localized_data( $data, $script_handle, $object_name ) {
+		if ( $this->has_localized_diag ) {
+			return $data;
+		}
```
malithsen: I extracted this localization to a separate PR in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5372/, so it might make sense to remove it from here to keep this contained to the server-side recorder.

malithsen (commented):
Nice work. I haven't tested yet. Left some comments which we likely want to address first.

malithsen (changes requested):
Just to be explicit, requesting changes. (Posted my findings as comments previously)

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
wjrosa: Dropped `on_localized_data` + the `wc_stripe_localized_data` filter and field in 3da9dd1. #5372 owns it now.

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
wjrosa: Switched to WC Sessions (`wc_stripe_diag_session_id`, same key as #5372) in 3da9dd1. Per-shopper now.

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
wjrosa: Updated to the dotted form (`stripe.api.request` / `stripe.api.response` / `webhook.received`) in 3da9dd1.

includes/class-wc-stripe.php
```diff
@@ -134,6 +134,7 @@ public function init() {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-database-cache.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-payment-method-configurations.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-database-cache-prefetch.php';
+		WC_Stripe_Diagnostics_Recorder::get_instance()->init();
```
malithsen: We should move this under the initializer function introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5372/changes#diff-7ee2c0c14e54acfa1dffe5f2ee2500ff08309c80116eb86d692bb5af887c6d9aR294 and rename that function so that it's generic (which currently says `*frontend_loader`)

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
```diff
@@ -0,0 +1,348 @@
...
+	 */
+	public function __construct( WC_Stripe_Diagnostics_Trace_Store $store, WC_Stripe_Diagnostics_Redactor $redactor ) {
+		$this->store    = $store;
+		$this->redactor = $redactor;
+	}
+
+	/**
+	 * Wire up the filter subscriptions. Called once during plugin init.
+	 */
+	public function init(): void {
```
malithsen, line 106: We should gate the initialization so that we never initialize if the diagnostic toggle is turned off.

malithsen (changes requested):
Looks good and tests well. 
This PR and its parent are missing the latest changes merged onto `feature/diagnostics` ( I patched them locally to test the changes)

Requesting changes to gate initialization. Otherwise looks good to me.

includes/class-wc-stripe.php
wjrosa: Done in af0f2f5e0 — moved into a new `initialize_diagnostics()` method hooked on the `init` action (mirroring the Apple Pay registration / #5372 frontend loader pattern). When #5372 lands, the two methods can be consolidated into a single `initialize_diagnostics()` entry point.

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
wjrosa: Done in af0f2f5e0 — moved the gate inside `Recorder::init()` itself so any caller sees the same behavior. When the toggle is off, `init()` returns early and no Stripe API hooks register. Added a test asserting that no-hook-registered state.

includes/class-wc-stripe.php
malithsen: > When https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5372 lands, the two methods can be consolidated into a single initialize_diagnostics() entry point.

#5372 has already been merged. We should update this branch with the latest from `feature/diagnostics`

wjrosa: Already done in d8c43ba6b — merged `feature/diagnostics` (which carries #5372). Consolidated both into a single `initialize_diagnostics()` that initializes the frontend loader and the recorder.

malithsen (approved):
Looks good. Thanks for the changes.

PR #5351, opened 2026-04-24
[RSM - Diagnostics Mode] Add diagnostics Action Scheduler cleanup jobs

wjrosa:
Towards RSM-629

One recurring Action Scheduler job that keeps the diagnostics trace store bounded.

### Job

| Action | Cadence | What it does |
|---|---|---|
| `wc_stripe_diagnostics_daily_cleanup` | daily | Deletes traces with `created_at` older than 7 days. Hard guarantee that the store can't grow without bound regardless of shopper traffic. |

### Scheduling

- `ensure_scheduled()` runs on `init` priority 20 and checks:
  - `action_scheduler_init` has fired
  - `as_has_scheduled_action` / `as_schedule_recurring_action` exist (some embedded setups ship without Action Scheduler)
  - the job isn't already scheduled
- Idempotent: safe to call on every request; no duplicate schedules.
- Scheduled under the `woocommerce-gateway-stripe` group so it shows up in WP Admin → Tools → Scheduled Actions next to the existing plugin jobs.

### Action callback wrapping

`run_daily_cleanup()` returns the deleted count so tests can assert directly on it. The `init()` hook wraps it in a closure that discards the return value, because WordPress action callbacks are expected to return void. Separating "does the work" from "is an action callback" keeps the tests synchronous and readable.

## Testing instructions

1. **Unit tests:** `./bin/run-tests.sh --filter 'WC_Stripe_Diagnostics_Cleanup_Test'`.
2. **Manual:** in a WP shell, create a trace, backdate its `created_at`, then `do_action( 'wc_stripe_diagnostics_daily_cleanup' )` — the trace should be deleted.
3. **Scheduler verification:** visit WP Admin → Tools → Scheduled Actions, filter by group `woocommerce-gateway-stripe` — you should see the daily cleanup job scheduled.

### Test coverage

- Old traces are deleted, fresh traces kept.
- Edge: a trace inside the 7-day window is not deleted.

Local checks: `npm run lint:php` ✅ · `npm run phpstan` ✅ · cleanup test file ✅

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/diagnostics/class-wc-stripe-diagnostics-cleanup.php
```diff
@@ -0,0 +1,176 @@
...
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Action Scheduler jobs that keep the diagnostics trace store tidy.
+ *
+ * Two jobs run in the background:
+ *
+ * - **Daily 7-day sweep:** deletes traces older than seven days. Enforces the
+ *   "store never grows without bound" guarantee regardless of traffic.
+ * - **5-minute pending sweep:** promotes traces that have been stuck in
```
malithsen: I am wondering if we really need the 5-minute sweep. What are we losing if we get rid of it?

I am thinking that having some of the 'abandoned' traces in the `pending` state won't be so bad, compared to having a very frequent scheduled task just for the cleanup.

includes/class-wc-stripe.php
```diff
@@ -134,6 +134,7 @@ public function init() {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-database-cache.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-payment-method-configurations.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-database-cache-prefetch.php';
+		( new WC_Stripe_Diagnostics_Cleanup() )->init();
```
malithsen: We should move this to the initializer function as well.

changelog.txt
```diff
@@ -8,6 +8,7 @@ xxxx-xx-xx - version 10.7.0
 * Dev - Add wc_stripe_api_response_received action and wc_stripe_localized_data filter to pave the way for the upcoming diagnostics recorder
 * Dev - Add WC_Stripe_Diagnostics_Trace_Store (file-backed trace storage with FIFO eviction and size caps) for the upcoming diagnostics recorder
 * Dev - Add POST /wc/v3/wc_stripe/diagnostics/events REST endpoint that accepts batched client events and appends them to the file-backed trace store
+* Dev - Add Action Scheduler jobs for diagnostics trace cleanup (daily 7-day purge, 5-minute stale-pending promotion to abandoned)
```
malithsen: Super-nit: I've been ignoring changelog updates in my PRs with the plan of adding a single changelog entry when we merge the feature branch to `develop`. But that's not a huge deal either way, we can cleanup/keep/remove these entries just before the merge.

includes/diagnostics/class-wc-stripe-diagnostics-cleanup.php
wjrosa: Good call — dropped it in 5e7b182a9. `pending` aging out via the daily sweep is fine; the dashboard reading it as "still in flight" for a short window is acceptable.

includes/class-wc-stripe.php
wjrosa: Done in 5e7b182a9 — moved into `initialize_diagnostics()` next to the frontend loader and recorder.

changelog.txt
wjrosa: Dropped both entries in 5aa881c51.

includes/diagnostics/class-wc-stripe-diagnostics-cleanup.php
```diff
@@ -0,0 +1,124 @@
...
+
+	/**
+	 * Make sure both jobs are scheduled. Called late enough that
+	 * Action Scheduler is available.
+	 */
+	public function ensure_scheduled(): void {
+		if ( ! did_action( 'action_scheduler_init' ) ) {
+			return;
+		}
+		if ( ! function_exists( 'as_has_scheduled_action' ) || ! function_exists( 'as_schedule_recurring_action' ) ) {
```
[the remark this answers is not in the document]
wjrosa: Mostly addressed by 5e7b182a9 — the 5-minute `promote_abandoned` job was dropped per Malith's review, so the only remaining Action Scheduler entry is one daily cleanup. That one I'd rather leave unconditional: it's a cheap no-op on sites that never captured (empty index → skipped loop), and it's the safety net for sites that enable diagnostics briefly, capture traces, then disable the toggle — gating it would leave those traces orphaned on disk indefinitely.

malithsen (approved):
This looks good to me. I tested by backdating an existing trace and then manually running the scheduled action. 

As a side note, we should probably mention in the settings section that traces get auto-deleted after a week. We can handle that as a follow-up to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5412

Also, let's update the PR description to remove the 5-minute action that we no longer have.

wjrosa:
Done — dropped the `wc_stripe_diagnostics_promote_abandoned` row from the Jobs table, retitled the section as "Job" (singular), trimmed the "Action callback wrapping" blurb to just `run_daily_cleanup()`, and removed the two pending/abandoned promotion bullets from the test coverage list. Description now matches the code after 5e7b182a9.

PR #5352, opened 2026-04-24
[RSM - Diagnostics Mode] Add diagnostics admin toggle + auto-disable at N=20

wjrosa:
Part of RSM-629 — completes the backend work.

> ⚠️ Stacked on #5348 (Redactor) and #5349 (Recorder backend). The diff shrinks to just this PR's additions once those merge.

Final piece that turns the recorder + REST endpoint + store + redactor into a user-facing feature: an admin toggle with a 20-trace capture budget that auto-disables when spent.

### Admin REST endpoints

| Route | Auth | Body | Response |
|---|---|---|---|
| `GET /wc/v3/wc_stripe/diagnostics/state` | `manage_woocommerce` | — | `{ enabled, count, limit }` |
| `POST /wc/v3/wc_stripe/diagnostics/toggle` | `manage_woocommerce` | `{ enabled: bool }` | same as /state |

### Per-shopper session id generation

`WC_Stripe_Diagnostics_Recorder::ensure_shopper_session()` is called from the `wc_stripe_localized_data` filter. When the toggle is on and the budget isn't spent, it:

1. Returns the shopper's existing session id from WC session if present.
2. Otherwise generates `diag-<12-char-random>`, stores it in WC session, and returns it.

Per-shopper storage uses `WC()->session` (not a global transient) so different shoppers don't share a trace. When WC session isn't available (rare edge cases outside checkout), the method returns null — no harm done.

### Auto-disable at N=20

When the trace store already holds `CAPTURE_LIMIT` (20) traces, `ensure_shopper_session()`:
1. Flips `wc_stripe_diagnostics_enabled` to `no`.
2. Returns null — no further shoppers instrument.

The check runs **before** reading the existing WC session id, so mid-flight shoppers still stop instrumenting the moment the budget is spent — no "one extra trace" from an already-started session.

### Why the recorder owns `is_enabled()`

The events controller from #5350 previously owned `is_enabled()`. Moving it onto the recorder makes the recorder the single source of truth for "is the feature live?" — both controllers and the `ensure_shopper_session()` path now read from one method. The events controller keeps a thin shim so its public API is unchanged.

## Testing instructions

1. **Unit tests:**
   - `./bin/run-tests.sh --filter 'WC_REST_Stripe_Diagnostics_Toggle_Controller_Test'` — 4 tests
   - `./bin/run-tests.sh --filter 'WC_Stripe_Diagnostics_Recorder_Test'` — 15 tests
   - Full diagnostics suite: 48 tests, 635 assertions.

2. **End-to-end smoke test** (with the frontend #5330 cherry-picked):
   ```bash
   # Enable via REST
   NONCE=$(wp eval 'echo wp_create_nonce("wp_rest");')
   curl -X POST http://localhost:8072/wp-json/wc/v3/wc_stripe/diagnostics/toggle \
        -H "X-WP-Nonce: $NONCE" -H "Content-Type: application/json" \
        -d '{"enabled":true}' --cookie-jar /tmp/wpc.txt --cookie /tmp/wpc.txt
   ```
   Then load the storefront checkout: `wcStripeDiag` should be set to `{ active: true, sessionId: 'diag-xxxxx', ... }`. Complete a checkout. Repeat 20 times — on the 21st page load the toggle should have auto-flipped to off and `wcStripeDiag` should no longer be injected.

### Test coverage

- Admin-only access (administrator allowed, subscriber denied).
- State response shape and defaults.
- Toggle flips the option in both directions.
- Recorder returns null when disabled.
- Recorder generates a fresh id when enabled.
- Recorder auto-disables at the capture limit and returns null.

Local checks: `npm run lint:php` ✅ · `npm run phpstan` ✅ · full diagnostics suite ✅

## Known follow-ups (explicitly not in this PR)

- **Admin UI:** the toggle is wired as REST only. A visible UI in WooCommerce → Settings → Payments → Stripe → Diagnostics is a follow-up, since the parent issue didn't call it out. Merchants would flip it via the REST endpoint or a future admin screen.
- **Trace browser:** merchants have no way to see captured traces yet. That's the next track (the AI root-cause UI from the project description).
- **Filter-backed per-shopper id rotation:** right now the id lives in WC session for the whole session, which is fine for one checkout but stretches across multi-checkout visits. If needed, a timeout rotation can be added.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -10,6 +10,7 @@ xxxx-xx-xx - version 10.7.0
 * Dev - Add POST /wc/v3/wc_stripe/diagnostics/events REST endpoint that accepts batched client events and appends them to the file-backed trace store
 * Dev - Add WC_Stripe_Diagnostics_Redactor (allow-list registry + PII scrubbers) for the upcoming diagnostics recorder
 * Dev - Add WC_Stripe_Diagnostics_Recorder server-side singleton that injects wc_diag_session_id into Stripe metadata and records redacted request/response/webhook events
+* Dev - Add diagnostics admin toggle (REST endpoints at /wc/v3/wc_stripe/diagnostics/state and /toggle) with per-shopper session id generation and auto-disable at 20 captured traces
 
 2026-04-23 - version 10.6.1
 * Fix - Fix Optimized Checkout Suite bugs that could result in duplicate subscription charges and generic payment methods names
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,6 +9,7 @@ xxxx-xx-xx - version 10.7.0
 * Dev - Add WC_Stripe_Diagnostics_Trace_Store (file-backed trace storage with FIFO eviction and size caps) for the upcoming diagnostics recorder
 * Dev - Add POST /wc/v3/wc_stripe/diagnostics/events REST endpoint that accepts batched client events and appends them to the file-backed trace store
 * Dev - Add WC_Stripe_Diagnostics_Redactor (allow-list registry + PII scrubbers) for the upcoming diagnostics recorder
+* Dev - Add WC_Stripe_Diagnostics_Recorder server-side singleton that injects wc_diag_session_id into Stripe metadata and records redacted request/response/webhook events
 
 2026-04-23 - version 10.6.1
 * Fix - Fix Optimized Checkout Suite bugs that could result in duplicate subscription charges and generic payment methods names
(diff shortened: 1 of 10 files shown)


wjrosa:
Closing in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5371

PR #5354, opened 2026-04-24
[PoC] Add payouts tab to Stripe admin

daledupreez:

This PR explores adding a Payouts tab to the Stripe admin UI to show balance details and list current and past payouts. This is an initial AI-generated PoC that needs multiple improvements, but I wanted to get it shared so we can decide how to keep this moving.

### Approach

The PR adds basic REST APIs that forward requests to the Stripe v1 Payouts and Balance APIs, using local WooCommerce permissions checks.

The client code then renders the payouts data using a DataView, and the balance details in a `Card` from `@wordpress/components`.

### Dependency changes

To get the dataviews dependency handled correctly, I needed to install `@wordpress/dataviews` and update `@woocommerce/dependency-extraction-webpack-plugin` to a more recent version. In addition, I had to add and then remove `@wordpress/dependency-extraction-webpack-plugin` as an explicit dev dependency to "unstick" a dependency on an older version from the `@woocommerce/dependency-extraction-webpack-plugin` deps tree. Those changes are only visible in `package-lock.json`, but we needed that bump to get the built JS working.

### Outstanding items/work

The client code needs some closer inspection -- we definitely need to look at the currency and date formatting to see if we can re-use other logic (or maybe offer the date format as a customisable option).

We also need to work out how we should promote instant payouts. We could do that using the balance data, or we can look at using an embedded component.

We also need to investigate if we can access Stripe embedded components from this context -- I haven't done that yet.

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


daledupreez:
@millin-stripe, please take a look at this initial PoC within the plugin.

daledupreez:
The changes so far are _huge_ in terms of JS payload. I will try to dig in next week to see where that bump is coming from.

Issue #184, opened 2017-03-15
iDeal (Netherlands) implementation?


IkenHuub:
Hi,

Will iDeal support via Stripe in your plugin be added anytime soon?

Thank you!

roykho:
Close in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/issues/156

PR #5359, opened 2026-04-27
Fix Dependabot #184: bump serialize-javascript (GHSA-5c6j-r48x-rmvq)

wjrosa:
## Summary

Resolves [Dependabot alert #184](https://github.com/woocommerce/woocommerce-gateway-stripe/security/dependabot/184) — a high-severity RCE in `serialize-javascript` ([GHSA-5c6j-r48x-rmvq](https://github.com/advisories/GHSA-5c6j-r48x-rmvq)) via `RegExp.flags` and `Date.prototype.toISOString()`.

The lockfile carried a single deduped `6.0.2` copy, used transitively by `mocha`, `copy-webpack-plugin`, and `terser-webpack-plugin`. All dev-only; never reaches production builds.

### Fix

Single-line npm `override` in `package.json`:

```json
"serialize-javascript": "^7.0.5"
```

The 6.x → 7.x bump only drops support for Node < 20 — the public API is unchanged. The project already requires Node ≥ 20.18.1, so all three consumers work without modification.

### Verification

- `npm audit` no longer reports `serialize-javascript`.
- `mocha`, `copy-webpack-plugin`, `terser-webpack-plugin` all `require()` cleanly.
- `npm run test:js -- --testPathPattern='client/api'` passes.

### Changelog

No `changelog.txt` / `readme.txt` entry — per review feedback, transitive dev-only dependency bumps that don't ship in the built plugin don't belong in the merchant-facing changelog. The original entry was removed in a follow-up commit.

## Test plan

- [ ] CI green (lint, jest, build)
- [ ] `npm audit` shows no `serialize-javascript` advisories
- [ ] `npm run build:webpack` completes

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,7 @@ xxxx-xx-xx - version 10.7.0
 * Fix - Surface PHP Throwables from the Agentic Commerce checkout.session.completed flow so fatals are logged, the order rollback runs, and Action Scheduler marks the job failed
 * Fix - Look up products by SKU in Agentic Commerce manual approval and tax calculation flows
 * Dev - Rename payment request references to express checkout
+* Add - Add Agentic Commerce admin dashboard for monitoring product feed sync status, history, errors, and triggering manual syncs
 * Fix - Store Stripe Terminal IPP channel metadata on orders so WooCommerce can identify POS payments and suppress standard transactional emails
 
 2026-04-23 - version 10.6.1
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,7 @@ xxxx-xx-xx - version 10.7.0
 * Fix - Look up products by SKU in Agentic Commerce manual approval and tax calculation flows
 * Dev - Rename payment request references to express checkout
 * Fix - Store Stripe Terminal IPP channel metadata on orders so WooCommerce can identify POS payments and suppress standard transactional emails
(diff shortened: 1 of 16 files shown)


changelog.txt
```diff
@@ -5,6 +5,7 @@ xxxx-xx-xx - version 10.7.0
 * Fix - Look up products by SKU in Agentic Commerce manual approval and tax calculation flows
 * Dev - Rename payment request references to express checkout
 * Fix - Store Stripe Terminal IPP channel metadata on orders so WooCommerce can identify POS payments and suppress standard transactional emails
+* Dev - Bump transitive serialize-javascript dev dependency to resolve RCE GHSA-5c6j-r48x-rmvq
```
diegocurbelo: [nit] I personally don't find this transitive dependencies that do not affect the build code to be useful in the changelog; I think we should only add those that are runtime dependencies.

diegocurbelo (approved):
LGTM 👍🏼

changelog.txt
wjrosa: Good call — fixed in 7f1365f. `serialize-javascript` is only pulled in by `mocha`, `copy-webpack-plugin`, and `terser-webpack-plugin`, none of which ship in the built plugin, so the entry was noise for merchants. Dropped from both files and noted the rationale in the PR description.

malithsen (approved):
LGTM

PR #5360, opened 2026-04-27
[RSM - What's new popup] Add "What's new" modal after standalone plugin updates (RSM)

wjrosa:
See: phcsdm-2NN-p2
Fixes STRIPE-1116

Part of the **Radical Speed Month (RSM)** experiment, motivated by recurring wp.org reviews from merchants who feel "blindsided" by post-auto-update behavior changes.

When this plugin is updated **standalone** (i.e. the only target of the upgrader run), the next admin page load surfaces a small modal showing the changelog entries for the new version, sourced directly from `readme.txt`. Bulk plugin updates suppress the modal — the assumption being that a multi-plugin update is intentional and the merchant is already in a triage mindset.

Preview:
Screenshot 2026-04-29 at 17 39 20

**How it works**

- `WC_Stripe_Whats_New_Modal::maybe_flag_standalone_update()` listens on `upgrader_process_complete`. It only sets `wc_stripe_whats_new_modal_pending` when the upgrader's `$hook_extra` describes a single-plugin update whose target is `plugin_basename( WC_STRIPE_MAIN_FILE )`. Theme/core upgrades, installs, and bulk-with-others are all ignored.
- On `admin_enqueue_scripts`, the controller checks `manage_woocommerce` capability + transient + per-user dismissal meta, then ships a small React entrypoint (`build/whats-new-modal.*`) and localizes the parsed changelog as `wcStripeWhatsNewModalParams`.
- `parse_changelog_for_version()` reads `readme.txt`, locates the `= X.Y.Z - DATE =` block, and returns each `* Tag - Description` bullet as `{ tag, text }`. The readme path is overridable via a `protected get_readme_path()` seam so tests exercise the production regex.
- Dismissal posts to `wp_ajax_wc_stripe_dismiss_whats_new_modal`, which records the dismissed version in **per-user** meta (`wc_stripe_whats_new_modal_dismissed_version`). The site-wide pending transient is left alone so other admins still see the modal once. Each event also fires a Tracks event (`wcstripe_whats_new_modal_*`) so we can measure the experiment without waiting on review sentiment — see the linked PR comment for the event spec.

**How can this code break?**

- *Bulk-update detection relies on `$hook_extra`.* If a third-party upgrader plugin doesn't populate `plugins`/`plugin` correctly, we conservatively skip showing the modal. Better to under-show than to mis-classify a bulk update as standalone.
- *Changelog parser is regex-based.* If the `readme.txt` header format ever ships an RC/beta version (e.g. `= 10.7.0-rc.1 - … =`), the lookahead would currently merge that block into the previous one. The bundled readme doesn't ship pre-release headers today; a follow-up will widen the regex when this becomes relevant.
- *Telemetry is best-effort.* `WC_Tracks::record_event()` is guarded by `class_exists()` and the JS helper is wrapped in try/catch, so a missing/disabled Tracks library can never break the modal.
- *Disable switch:* `add_filter( 'wc_stripe_whats_new_modal_enabled', '__return_false' );` kills the experiment entirely.

- PHPUnit covers the parser, standalone-update detection via `@dataProvider`, per-user dismissal semantics, AJAX-context dismissal, the disable filter, and the capability gate.
- The admin-only surface area means classic, Blocks, OC, and ECE checkout paths are unaffected.

## Testing instructions

> Assume the tester has admin access to a WooCommerce store with this plugin active and does not have intimate knowledge of Stripe internals.

1. Install this branch and visit any wp-admin page once to bootstrap the version option.
2. **Quick visual preview (no real update needed)** — fastest way to just see the modal:
   - In `includes/admin/class-wc-stripe-whats-new-modal.php`, short-circuit `should_display()` to `return true;` immediately after the `manage_woocommerce` capability check. This bypasses the pending-transient and per-user-dismissal gates so the modal renders on every admin page load.
   - In the same file, swap the `parse_changelog_for_version( WC_STRIPE_VERSION )` call inside `maybe_enqueue_modal()` for `parse_changelog_for_version( '10.7.0' )` (or whichever next-shipping version's `readme.txt` block you want to preview) so you see the entries that will actually ship rather than the current installed version's.
   - Run `npm run build:webpack` and reload any wp-admin page. **Revert both edits before committing.**
3. **Standalone update path** — simulate a standalone update so the modal appears via the production gates:
   - Run in WP-CLI: `wp option delete wc_stripe_whats_new_modal_dismissed_version --user=<your admin id>; wp transient set wc_stripe_whats_new_modal_pending 1 --network=false`
   - Reload any wp-admin page. The "What's new in WooCommerce Stripe X.Y.Z" modal should appear with the changelog entries from `readme.txt` for the current version.
4. **Dismissal** — click "Got it". The modal closes, and a reload should not bring it back. Inspect user meta to confirm:
   - `wp user meta get <your admin id> wc_stripe_whats_new_modal_dismissed_version` should return the current version.
5. **Per-user dismissal** — log in as a *different* admin user and reload. The modal should still appear once for them (regression: it must not be suppressed by the first admin's dismissal).
6. **Bulk update suppression** — repeat step 3, then run a real bulk plugin update from the Plugins screen that includes another plugin alongside this one. The transient should not be set and the modal should not appear after the bulk update completes.
7. **Disable filter** — `add_filter( 'wc_stripe_whats_new_modal_enabled', '__return_false' );` in mu-plugins. Reload — the modal should not appear regardless of transient/meta state.
8. **Permissions** — log in as a Subscriber. Even with the transient set, the modal should not enqueue.
9. **AJAX security** — try POSTing to `admin-ajax.php?action=wc_stripe_dismiss_whats_new_modal` without a valid nonce; the response should be a 4xx, and your dismissal meta should remain unchanged.

---

-   [x] Tested on mobile (or does not apply) — admin-only feature, no checkout impact

### Changelog entry

* Add - Show a "What's new" modal to admins after a standalone update of this plugin, sourcing the entries from readme.txt (RSM experiment)

<details>

#### Comment

n/a — entry included above.

</details>

**Post merge**


wjrosa:
## Follow-up: Tracks instrumentation

Added the events needed to read the RSM experiment without waiting on wp.org review sentiment. All events carry `stripe_version` and `is_test_mode`.

| Event | Side | When | Extra props |
| --- | --- | --- | --- |
| `wcstripe_whats_new_modal_enqueued` | server | `WC_Stripe_Whats_New_Modal::maybe_enqueue_modal()` ships the asset | `entry_count` |
| `wcstripe_whats_new_modal_impression` | client | React component mounts | `entry_count` |
| `wcstripe_whats_new_modal_dismissed` | client **and** server (AJAX) | user closes the modal | `dwell_ms`, `source` (`primary_button` \| `close_icon`) |
| `wcstripe_whats_new_modal_changelog_click` | client | "View full changelog" link clicked | `entry_count` |

### Reading the experiment

- **Did anyone read it?** `impression / enqueued` ratio (rendered vs. shipped). `dismissed` events with `dwell_ms` < ~1500 ms ≈ blew past it.
- **Did it surface the right things?** `changelog_click / impression` — high CTR means the summary wasn't enough; low means we're communicating effectively in the modal itself.
- **Lagging metric still wp.org reviews.** Cohort by `stripe_version` and compare 1★ rate / "broke after update" theme rate vs. the previous 2-3 releases.

### Why both client and server for dismiss

Server-side fires whenever the AJAX handler runs (matches the option write), so we never lose a dismissal to a JS error or page navigation. Client-side carries `dwell_ms` and `source` and runs even if the AJAX dispatch is blocked. They should agree within a small margin; divergence means JS or auth issues worth investigating.

### Safety

`record_event()` no-ops when `WC_Tracks` is missing (older WC, test env). The JS helper bails when `window.wcTracks.isEnabled` is false or the API is missing, and is wrapped in try/catch so telemetry can never break the modal.

includes/admin/class-wc-stripe-whats-new-modal.php
```diff
@@ -0,0 +1,264 @@
...
+	 * @return void
+	 */
+	public function handle_dismiss() {
+		if ( ! current_user_can( 'manage_woocommerce' ) ) {
+			wp_send_json_error( [ 'message' => 'forbidden' ], 403 );
+		}
+
+		check_ajax_referer( self::DISMISS_AJAX_ACTION, 'nonce' );
+
+		update_option( self::DISMISSED_OPTION, WC_STRIPE_VERSION, false );
```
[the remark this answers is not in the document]
wjrosa: Good catch — fixed in 2d04fac1. Switched to `update_user_meta`/`get_user_meta` keyed by `get_current_user_id()` and added a regression test (`test_should_display_returns_true_when_a_different_user_dismissed`) that proves dismissal by one admin doesn't suppress the modal for other admins. Also kept `PENDING_TRANSIENT` site-wide and stopped clearing it on dismiss (was a bug under per-user semantics — would have hidden the modal from everyone after the first dismissal). Renamed the constant `DISMISSED_OPTION` → `DISMISSED_USER_META` to match the new storage location.

tests/phpunit/admin/class-wc-stripe-whats-new-modal-test.php
```diff
@@ -0,0 +1,225 @@
...
+						$line = ltrim( $line, "* \t" );
+						if ( preg_match( '/^([A-Za-z][A-Za-z ]*?)\s+-\s+(.+)$/', $line, $parts ) ) {
+							$items[] = [
+								'tag'  => trim( $parts[1] ),
+								'text' => trim( $parts[2] ),
+							];
+							continue;
+						}
+						$items[] = [
+							'tag'  => '',
```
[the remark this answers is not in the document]
wjrosa: Done in 2d04fac1. Added a `protected get_readme_path()` seam to the production class; the fixture test now uses an anonymous subclass that only overrides the path, so `parse_changelog_for_version`'s real regex/loop is what's under test:

```php
private function make_modal_with_readme( string $readme_path ): WC_Stripe_Whats_New_Modal {
    return new class( $readme_path ) extends WC_Stripe_Whats_New_Modal {
        private string $path;
        public function __construct( string $path ) { $this->path = $path; }
        protected function get_readme_path(): string { return $this->path; }
    };
}
```

Also wrapped the temp dir cleanup in try/finally so a failing assertion can't leak `/tmp/wc-stripe-whats-new-*` directories.

tests/phpunit/admin/class-wc-stripe-whats-new-modal-test.php
```diff
@@ -0,0 +1,225 @@
...
+							'text' => $line,
+						];
+					}
+					return $items;
+				}
+				return parent::parse_changelog_for_version( $version );
+			}
+		};
```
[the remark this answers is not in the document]
wjrosa: Same fix as the claude-bot thread above — addressed in 2d04fac1 by adding a `protected get_readme_path()` seam and replacing the duplicated parsing logic with a thin path-override subclass, so the regex/loop now under test is the production one.

tests/phpunit/admin/class-wc-stripe-whats-new-modal-test.php
```diff
@@ -0,0 +1,225 @@
...
+			null,
+			[
+				'action' => 'install',
+				'type'   => 'plugin',
+				'plugin' => plugin_basename( WC_STRIPE_MAIN_FILE ),
+			]
+		);
+
+		$this->assertFalse( get_transient( WC_Stripe_Whats_New_Modal::PENDING_TRANSIENT ) );
+	}
```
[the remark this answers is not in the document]
wjrosa: Consolidated in 2d04fac1. The six cases collapse into a single `test_maybe_flag_standalone_update` driven by `provide_flag_standalone_update_cases`, and I took the suggestion to add the missing edge cases too: `'empty plugins array ignored'` and `'missing action ignored'` are now covered explicitly.

includes/admin/class-wc-stripe-whats-new-modal.php
```diff
@@ -0,0 +1,321 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Shows a "What's new" modal to admins after a standalone update of this plugin.
+ *
+ * Part of the Radical Speed Month (RSM) experiment: when only this extension is
```
malithsen: nit: This line about RSM probably doesn't need to be documented here.

includes/admin/class-wc-stripe-whats-new-modal.php
```diff
@@ -0,0 +1,321 @@
...
+	/**
+	 * AJAX action used by the modal to persist dismissal.
+	 */
+	const DISMISS_AJAX_ACTION = 'wc_stripe_dismiss_whats_new_modal';
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		add_action( 'upgrader_process_complete', [ $this, 'maybe_flag_standalone_update' ], 10, 2 );
```
malithsen: Should we use `woocommerce_stripe_updated`?  It's more targeted and won't need the filtering logic we have now.

includes/admin/class-wc-stripe-whats-new-modal.php
```diff
@@ -0,0 +1,321 @@
...
+	 * Records a Tracks event, automatically tagging it with the plugin version
+	 * and current Stripe mode. Silently no-ops when WC_Tracks is unavailable
+	 * (e.g. older WooCommerce installs or test environments).
+	 *
+	 * @param string              $name  Event name.
+	 * @param array<string,mixed> $props Event properties.
+	 *
+	 * @return void
+	 */
+	protected function record_event( string $name, array $props = [] ) {
```
malithsen: Suggest renaming this method to something more specific (eg: `record_whats_new_modal_event`) so that it doesn't conflate with the main `record_event` function.

includes/admin/class-wc-stripe-whats-new-modal.php
```diff
@@ -0,0 +1,321 @@
...
+		if ( ! class_exists( 'WC_Tracks' ) ) {
+			return;
+		}
+
+		$props['stripe_version'] = WC_STRIPE_VERSION;
+		if ( class_exists( 'WC_Stripe_Mode' ) ) {
+			$props['is_test_mode'] = WC_Stripe_Mode::is_test() ? 'yes' : 'no';
+		}
+
+		WC_Tracks::record_event( $name, $props );
```
malithsen: We should probably use `wc_admin_record_tracks_event` instead, as these events are emitted from the admin area.

client/entrypoints/whats-new-modal/whats-new-modal.js
```diff
@@ -0,0 +1,157 @@
...
+	}, [] );
+
+	const dismiss = useCallback(
+		( source ) => {
+			const resolvedSource = source || dismissedSource.current;
+			const dwellMs = Date.now() - mountedAt.current;
+
+			setIsOpen( false );
+
+			recordEvent( 'wcstripe_whats_new_modal_dismissed', {
```
malithsen: This event is also fired on the PHP side in line 218 of the PHP script. It'll likely result in duplicate events.

malithsen (changes requested):
Left some feedback after code inspection

client/entrypoints/whats-new-modal/whats-new-modal.js
wjrosa: Dropped the JS-side dismissed event in ce82e7469. PHP `handle_dismiss` is now the only emitter.

includes/admin/class-wc-stripe-whats-new-modal.php
wjrosa: Trimmed in ce82e7469.

includes/admin/class-wc-stripe-whats-new-modal.php
wjrosa: Done in ce82e7469 — switched to `woocommerce_stripe_updated`. Drops 30+ lines of `$hook_extra` branching. Note: this also surfaces the modal on bulk updates (the previous code skipped those); flag if you'd rather keep the standalone-only behavior.

includes/admin/class-wc-stripe-whats-new-modal.php
wjrosa: Switched in ce82e7469.

includes/admin/class-wc-stripe-whats-new-modal.php
wjrosa: Renamed to `record_whats_new_modal_event` in ce82e7469.

client/entrypoints/whats-new-modal/whats-new-modal.js
```diff
@@ -0,0 +1,157 @@
...
+				) }
+				<Button
+					variant="primary"
+					onClick={ () => dismiss( 'primary_button' ) }
+				>
+					{ __( 'Got it', 'woocommerce-gateway-stripe' ) }
+				</Button>
+			</div>
+		</Modal>
+	);
```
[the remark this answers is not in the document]
wjrosa: Acknowledged. Tracking as a follow-up — Jest+RTL coverage for impression/dismiss/changelog-click paths is the right call, but it's a meaningful addition and out of scope for this iteration.

tests/phpunit/admin/class-wc-stripe-whats-new-modal-test.php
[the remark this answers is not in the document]
wjrosa: Obsolete — today's refactor (ce82e7469) dropped `maybe_flag_standalone_update` entirely in favor of the simpler `woocommerce_stripe_updated` listener, so there's no longer a multi-case test to consolidate.

tests/phpunit/admin/class-wc-stripe-whats-new-modal-test.php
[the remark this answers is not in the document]
wjrosa: Already addressed in an earlier commit — the production class exposes a `protected get_readme_path()` seam, and the test's anonymous subclass overrides only that method (not `parse_changelog_for_version`), so the production regex/loop is exercised.

includes/admin/class-wc-stripe-whats-new-modal.php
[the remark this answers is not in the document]
wjrosa: Done in 1d6a26be9 — wrapped `wc_admin_record_tracks_event` in try/catch (Throwable) so a failure in WC Admin internals can't break admin enqueue or the dismiss AJAX.

client/entrypoints/whats-new-modal/whats-new-modal.js
```diff
@@ -0,0 +1,155 @@
...
+		window.wcTracks.recordEvent;
+	if ( typeof fn !== 'function' ) {
+		return;
+	}
+	try {
+		fn( name, props );
+	} catch ( e ) {
+		// Tracks is best-effort; never let it break the modal.
+	}
+};
```
malithsen: Can we use the existing utility function https://github.com/woocommerce/woocommerce-gateway-stripe/blob/6271f3bba3376c781fcfa29d418fd1c3f2c17c54/client/tracking/index.js#L45

client/entrypoints/whats-new-modal/whats-new-modal.js
```diff
@@ -0,0 +1,155 @@
...
+		fullChangelogUrl,
+		dismissAjaxUrl,
+		dismissAjaxAction,
+		dismissNonce,
+	} = params;
+
+	const baseProps = {
+		stripe_version: version,
+		is_test_mode: isTestMode ? 'yes' : 'no',
+		entry_count: changes.length,
```
malithsen: Do we care about the `entry_count` property or can it be dropped? That would simplify the use of existing function I mentioned in the previous comment

readme.txt
```diff
@@ -148,6 +148,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.7.0 - xxxx-xx-xx =
+* Add - Show a "What's new" modal to admins after a standalone update of this plugin, sourcing the entries from readme.txt (RSM experiment)
```
malithsen: "RSM experiment" probably won't make sense to merchants. Suggest dropping it.

malithsen (commented):
When following the test instructions to force the notice, I can see it.

Image

However, when I try to upgrade from a previous version to a build from this branch, I can't get the notice to appear.

Does it appear for you when you're actually upgrading?

includes/admin/class-wc-stripe-whats-new-modal.php
wjrosa: Done in ce82e7469.

includes/admin/class-wc-stripe-whats-new-modal.php
wjrosa: Done in ce82e7469 — switched to `woocommerce_stripe_updated` and dropped the filtering logic.

includes/admin/class-wc-stripe-whats-new-modal.php
wjrosa: Done in ce82e7469 — renamed to `record_whats_new_modal_event`.

includes/admin/class-wc-stripe-whats-new-modal.php
wjrosa: Done in ce82e7469 — using `wc_admin_record_tracks_event`.

client/entrypoints/whats-new-modal/whats-new-modal.js
wjrosa: Done in ce82e7469 — JS no longer fires the dismiss event; PHP-side is the single source.

client/entrypoints/whats-new-modal/whats-new-modal.js
wjrosa: Done in a3e30108d — using `recordEvent` from `wcstripe/tracking`. Localized a minimal `wc_stripe_settings_params` from the modal's enqueue path so the utility's globals are present on non-Stripe admin pages too.

client/entrypoints/whats-new-modal/whats-new-modal.js
wjrosa: Dropped in a3e30108d (JS `baseProps` and the PHP `enqueued` event).

readme.txt
wjrosa: Dropped in a3e30108d.

wjrosa:
@malithsen Reproduced — but the cause is a version-match short-circuit, not an upgrade-flow gap.

This branch still carries `Version: 10.6.1` / `WC_STRIPE_VERSION = '10.6.1'` (we haven't bumped). WP.org's released version is also 10.6.1, so when you install from WP.org and then upload the branch zip on top, `WC_Stripe::install()` does:

```php
$previous_version = get_option( 'wc_stripe_version' );  // '10.6.1'
if ( WC_STRIPE_VERSION === $previous_version ) {        // '10.6.1' === '10.6.1' → true
    return;                                             // ← bails here
}
do_action( 'woocommerce_stripe_updated' );              // never reached
```

No `woocommerce_stripe_updated` → `flag_pending_modal()` never runs → no transient → no modal. The forced-display path works because it sets the transient directly, sidestepping `install()` entirely. Upload vs. "Update now" link makes no difference — both end the same way.

Quickest way to actually exercise the upgrade path locally: bump the version in `woocommerce-gateway-stripe.php` before zipping, or install an older WP.org version (10.6.0) first and upload 10.6.1 on top, or `wp option delete wc_stripe_version` between steps.

For the actual release this is moot — the next published version will be `> 10.6.1`, so the mismatch is real.

client/entrypoints/whats-new-modal/style.scss
```diff
@@ -0,0 +1,49 @@
...
+	}
+
+	&__tag {
+		flex: 0 0 auto;
+		display: inline-block;
+		min-width: 56px;
+		padding: 2px 8px;
+		border-radius: 2px;
+		background: #f0f0f1;
+		color: #1e1e1e;
```
malithsen: Not a blocker, but I wonder if we already have colors defined as variables somewhere.

malithsen (approved):
LGTM, works as described.

wjrosa:
cc @woocommerce/quark

client/entrypoints/whats-new-modal/style.scss
wjrosa: Good catch — done in 8b64796c. Switched to `styles.$wp-gray-2` and `styles.$gray-900` via the existing `abstracts/styles` aggregator.

wjrosa:
Closing in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5386

PR #5361, opened 2026-04-27
Cancel pending retry when Stripe Radar blocks a subscription renewal

wjrosa:
Fixes STRIPE-1110

PR #5170 (10.6.0) added a block intended to suspend the parent subscription when Stripe Radar blocked a renewal charge so WC Subscriptions wouldn't schedule further retries. In production that block is a no-op:

- `$renewal_order->update_status(FAILED)` runs first, fires `woocommerce_order_status_changed` synchronously.
- Core's `WC_Subscriptions_Renewal_Order::maybe_record_subscription_payment` (priority 10) catches the transition, calls `$subscription->payment_failed()`, which moves the subscription to `on-hold` and triggers `WCS_Retry_Manager::maybe_apply_retry_rule` to schedule the retry.
- When control returns to the trait, the subscription is already `on-hold` — which is in the PR's `$terminal_statuses` skip list — so `update_status('on-hold')` never runs. Even without the skip, `WC_Subscription::update_status` no-ops on same-status transitions. And `WCS_Retry_Manager::maybe_cancel_retry` only cancels a pending retry when the new subscription status differs from the rule's expected status — default rule 0 expects `on-hold`, so this is the one transition that does **not** cancel the retry.
- Retry fires on schedule → Stripe → Radar block → next rule scheduled. Cycle continues through ~7 days, exactly the block-rate inflation #5170 was meant to prevent.

The previous PHPUnit test passed only because WCS isn't bootstrapped in the unit test env: the mock subscription stayed `active`, the skip didn't trip, the test asserted that `update_status` flipped the mock — which it did, in isolation. The test never validated production integration.

### Fix

Drop the unreachable `update_status('on-hold')` loop and tear the retry down directly via `WCS_Retry_Manager`'s public API:

- Cancel the pending `WCS_Retry`: `last_retry->update_status('cancelled')`.
- Clear the subscription's `payment_retry` date.
- Attach the Radar note to both subscription and renewal order via `add_order_note()` (bypasses same-status `update_status` no-op).

The subscription still ends up `on-hold` via core's `payment_failed()` — this trait does not need to duplicate that.

### How can this break?

- WC Subscriptions API change to `WCS_Retry_Manager::store()->get_last_retry_for_order()` or `WCS_Retry::update_status()`. Mitigation: `class_exists`/`is_retry_enabled` guards already wrap the calls; failures are caught and logged.
- A future change inside core's `payment_failed()` flow that no longer schedules a retry by the time we reach this point. The fix degrades gracefully — `get_last_retry_for_order()` returns null, the cancel branch is skipped, and we still attach the note.

### What stops it from breaking?

- New PHPUnit test asserts post-conditions a regression would actually surface: pending retry → `cancelled`, `payment_retry` date cleared, Radar note attached to subscription. The mock now models the real WCS retry surface (`WCS_Retry_Manager::store()->get_last_retry_for_order()`) instead of relying on the bootstrap gap that masked the original bug.
- Reporter verified end-to-end on a live stack (WC Stripe 10.6.1 + WC Subscriptions 8.6.1, Radar rule blocking renewal) before and after: post-fix `_schedule_payment_retry` postmeta = `0`, `wp_wcs_payment_retries` row = `cancelled`, no pending Action Scheduler retry action, Radar notes on subscription and renewal order, no further retries fire.

## Testing instructions

Prereqs: Stripe Connect in **test mode**, WC Subscriptions ≥ 8.6 active, *Automatic Payment Retry* enabled in WooCommerce → Settings → Subscriptions.

1. Create a subscription product and place an order using Stripe's [Radar-blocked test card](https://docs.stripe.com/testing#radar) `4100 0000 0000 0019` (any future expiry, any CVC). The first charge succeeds — these test cards only trip Radar on subsequent charges.
2. In WooCommerce → Status → Scheduled Actions, find the next `woocommerce_scheduled_subscription_payment` action for the subscription and run it now.

✅ After the run, you should see:
- Renewal order status: **failed**.
- Subscription status: **on-hold** with a Radar note ("Stripe Radar blocked this payment …"). Same note on the renewal order.
- WooCommerce → Status → Scheduled Actions: **no pending** `woocommerce_scheduled_subscription_payment_retry` for this subscription. The absence of this action is the proof that no further charge will fire — Action Scheduler is what would fire it.

Regression check: repeat with the standard generic-decline test card `4000 0000 0000 0002`. Standard WCS retry behavior should be unchanged — a retry **is** scheduled (visible as a pending `woocommerce_scheduled_subscription_payment_retry` action) and fires per the configured rules.

---

-   [x] Covered with tests
-   [x] Tested on mobile (or does not apply) — backend-only change

### Changelog entry

<details>

#### Comment

Changelog entry added to `changelog.txt` and `readme.txt` for 10.7.0:

> Fix - Cancel pending WC Subscriptions retry when Stripe Radar blocks a renewal payment so further automatic retries do not re-trigger the same Radar block

</details>

**Post merge**


includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -324,6 +324,14 @@ public function process_change_subscription_payment_with_deferred_intent( $subsc
 					$subscription->save();
 				}
 
+				// Persist the express checkout type so it can be applied to the payment method
+				// title after the customer completes the 3DS/redirect confirmation, since the
+				// post-confirmation flow doesn't have access to the form's $_POST.
+				if ( ! empty( $_POST['express_checkout_type'] ) ) {
+					$subscription->update_meta_data( '_wc_stripe_express_checkout_type', wc_clean( wp_unslash( $_POST['express_checkout_type'] ) ) );
+					$subscription->save();
+				}
```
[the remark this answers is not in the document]
wjrosa: You're right — that block snuck in from another in-progress flow that was polluting my working tree (the corresponding read side in `class-wc-stripe-express-checkout-element.php` and the Jest test in `class-wc-stripe-express-checkout-element-test.php` are unstaged on this machine and belong to a separate feature). I tried to filter them out when staging but missed this one because it lives in the same trait file as the Radar fix.

Reverted in 728caa08b — the trait diff now contains only the Radar retry teardown. The express-checkout-type metadata feature will land in its own PR with its read side and tests.

tests/phpunit/helpers/class-wcs-retry-manager.php
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Mock for the static facade the trait uses.
+ */
+class WCS_Retry_Manager {
+	/**
+	 * Whether retries are enabled (toggle this in tests).
+	 *
+	 * @var bool
+	 */
+	public static $retry_enabled = true;
```
daledupreez: Not blocking, but it feels like it might be helpful to provide clearer guideposts as to which properties and methods are for mocks/stubs and which properties/methods are to match the underlying API. Maybe adding explicit `mock` strings to the names would help with that delineation.

includes/compat/trait-wc-stripe-subscriptions.php
Mayisha: Nit: We should update the notes as we are not putting the subscription on-hold but cancelling the retry.

Mayisha: AI suggestion: `Stripe Radar blocked this payment due to a custom Radar rule. The automatic retry has been cancelled to prevent further blocked payment attempts.`

We can add the `Stripe Radar blocked this payment due to a custom Radar rule.` part here and append the `The automatic retry has been cancelled to prevent further blocked payment attempts` part below if it goes through the cancellation path.

Mayisha (approved):
Tests as described for the blocked payment. 👍 

There is one step that I had to do differently from the steps in the description. Stripe's [Radar-blocked test card](https://docs.stripe.com/testing#radar) `4100 0000 0000 0019` was not working for the subscription purchase and it was failing. Steps I followed was-
- Purchased a subscription with a regular card.
- From My Account > Subscriptions > Subscription details page, I clicked the change payment method button.
-  Added Stripe's [Radar-blocked test card](https://docs.stripe.com/testing#radar) `4100 0000 0000 0019` as the new payment method.
- As an admin, found the next `woocommerce_scheduled_subscription_payment` action for this subscription in WooCommerce > Status > Scheduled Actions and ran it.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Done in 5fc650246 — split into a cause note (always added) and a cancellation suffix appended only when a pending retry was actually cancelled.

wjrosa: Fixed in 5fc650246 — dropped the "put on hold" phrasing.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -560,37 +560,56 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
 					case 'low_probability_of_authorization':
-						$radar_note = __( 'Stripe blocked this payment due to low probability of authorization. The subscription has been put on hold to prevent further blocked payment attempts.', 'woocommerce-gateway-stripe' );
+						$radar_cause = __( 'Stripe blocked this payment due to low probability of authorization.', 'woocommerce-gateway-stripe' );
 						break;
 					case 'highest_risk_level':
 					default:
-						$radar_note = __( 'Stripe Radar blocked this payment as high risk. The subscription has been put on hold to prevent further blocked payment attempts.', 'woocommerce-gateway-stripe' );
+						$radar_cause = __( 'Stripe Radar blocked this payment as high risk.', 'woocommerce-gateway-stripe' );
 						break;
 				}
```
daledupreez, line 586: `$radar_cause` needs to be defined outside this block, as the case where `false === $radar_reason` would never set `$radar_cause`.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -560,37 +560,56 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
-						if ( in_array( $subscription->get_status(), $terminal_statuses, true ) ) {
-							continue;
+
+					$retry_cancelled = false;
+					if ( class_exists( 'WCS_Retry_Manager' ) && WCS_Retry_Manager::is_retry_enabled() ) {
+						$last_retry = WCS_Retry_Manager::store()->get_last_retry_for_order( $renewal_order->get_id() );
+						if ( $last_retry && 'pending' === $last_retry->get_status() ) {
+							$last_retry->update_status( 'cancelled' );
+							$retry_cancelled = true;
+						}
```
daledupreez: Not blocking, but it may be worth checking that the functions we plan to call actually exist on `WCS_Retry_Manager`.

```suggestion
					if ( class_exists( 'WCS_Retry_Manager' ) && method_exists( 'WCS_Retry_Manager', 'is_retry_enabled' ) && WCS_Retry_Manager::is_retry_enabled() && method_exists( 'WCS_Retry_Manager', 'store' ) ) {
						$retry_store = WCS_Retry_Manager::store();
						$last_retry = method_exists( $retry_store, 'get_last_retry_for_order' ) ? $retry_store->get_last_retry_for_order( $renewal_order->get_id() ) : null;
						if ( $last_retry && 'pending' === $last_retry->get_status() ) {
							$last_retry->update_status( 'cancelled' );
							$retry_cancelled = true;
						}
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -560,37 +560,56 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+						: $radar_cause;
+
+					foreach ( $subscriptions as $subscription ) {
+						$subscription->add_order_note( $radar_note );
 					}
 					$renewal_order->add_order_note( $radar_note );
 				} catch ( Exception $radar_e ) {
 					WC_Stripe_Logger::error(
-						'Failed to put subscription on hold after Stripe Radar block: ' . $radar_e->getMessage(),
+						'Failed to cancel scheduled retry after Stripe Radar block: ' . $radar_e->getMessage(),
```
daledupreez: Not blocking, but I think we should note that a renewal was blocked.
```suggestion
						'Failed to cancel scheduled retry after Stripe Radar blocked subscription renewal: ' . $radar_e->getMessage(),
```

tests/phpunit/helpers/class-wc-subscription.php
```diff
@@ -159,4 +173,54 @@ public function update_status( $new_status, $note = '', $manual = false ) {
...
+	}
+
+	/**
+	 * Test helper to seed a date that get_date() will return.
+	 *
+	 * @param string $date_type Date type.
+	 * @param int    $timestamp Timestamp.
+	 * @return void
+	 */
+	public function set_date( $date_type, $timestamp ) {
```
daledupreez: Question: is `set_date()` part of the main `WC_Subscription` interface? If not, we should call this function something like `set_mock_date()` to clearly delineate which methods are part of our mock logic.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -560,37 +560,56 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+						$radar_cause = __( 'Stripe Radar blocked this payment due to a custom Radar rule.', 'woocommerce-gateway-stripe' );
 						break;
 					case 'low_probability_of_authorization':
-						$radar_note = __( 'Stripe blocked this payment due to low probability of authorization. The subscription has been put on hold to prevent further blocked payment attempts.', 'woocommerce-gateway-stripe' );
+						$radar_cause = __( 'Stripe blocked this payment due to low probability of authorization.', 'woocommerce-gateway-stripe' );
 						break;
 					case 'highest_risk_level':
 					default:
-						$radar_note = __( 'Stripe Radar blocked this payment as high risk. The subscription has been put on hold to prevent further blocked payment attempts.', 'woocommerce-gateway-stripe' );
+						$radar_cause = __( 'Stripe Radar blocked this payment as high risk.', 'woocommerce-gateway-stripe' );
```
daledupreez: I think we should improve this wording so it makes more sense in context, especially for subscriptions.
```suggestion
						$radar_cause = __( 'Stripe Radar blocked payment for the saved payment method due to a custom Radar rule.', 'woocommerce-gateway-stripe' );
break;
case 'low_probability_of_authorization':
						$radar_cause = __( 'Stripe blocked payment for the saved payment method due to low probability of authorization.', 'woocommerce-gateway-stripe' );
break;
case 'highest_risk_level':
default:
						$radar_cause = __( 'Stripe Radar blocked payment for the saved payment method as high risk.', 'woocommerce-gateway-stripe' );
```

daledupreez: For the default case, might it make sense to include the reason code in the message?

phpstan-baseline.neon
```diff
@@ -3709,8 +3709,8 @@ parameters:
 			path: includes/compat/class-wc-stripe-email-customer-failed-refund.php
 
 		-
-			message: '#^Call to method get_time\(\) on an unknown class WCS_Retry\.$#'
-			identifier: class.notFound
+			message: '#^Call to function method_exists\(\) with WCS_Retry and ''get_time'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
```
daledupreez: It is probably worth adding an ignore rule for this.

daledupreez (approved):
Overall, this is working as described, but I do think we should improve the wording to be a bit clearer.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Done in a9a77b6d — initialized `$radar_cause = ''` before the switch as a defensive belt-and-suspenders.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Done in a9a77b6d — adopted the "blocked payment for the saved payment method" wording across all three explicit cases.

wjrosa: Good call — done in a9a77b6d. Default branch now formats the reason code into the note via `sprintf`.

phpstan-baseline.neon
wjrosa: Done in a9a77b6d — moved to a narrow `ignoreErrors` rule in `phpstan.neon` (with a comment explaining the runtime guard). Also extended it to cover the equivalent always-true flags PHPStan now raises for `WCS_Retry_Manager::is_retry_enabled` / `store`, since the test mock satisfies their existence too.

tests/phpunit/helpers/class-wcs-retry-manager.php
wjrosa: Good suggestion — done in a9a77b6d. Renamed the test-only surfaces to `$mock_retry_enabled`, `$mock_last_retry`, and `mock_reset()`; left `is_retry_enabled()` / `store()` as-is since those mirror the real `WCS_Retry_Manager` API.

tests/phpunit/helpers/class-wc-subscription.php
wjrosa: You're right — `set_date()` isn't part of the real `WC_Subscription` interface (production uses `update_dates()`). Renamed to `set_mock_date()` in a9a77b6d.

phpstan.neon
```diff
@@ -22,3 +22,16 @@ parameters:
...
+		# method_exists() guards still have to run at runtime in case an older Subscriptions
+		# release is installed without the methods.
+		-
+			message: '#^Call to function method_exists\(\) with WCS_Retry and ''get_time'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
+			path: includes/compat/class-wc-stripe-email-failed-authentication-retry.php
+		-
+			message: '#^Call to function method_exists\(\) with ''WCS_Retry_Manager'' and ''(is_retry_enabled|store)'' will always evaluate to true\.$#'
+			identifier: function.alreadyNarrowedType
+			path: includes/compat/trait-wc-stripe-subscriptions.php
```
daledupreez, line 37: I think we should add inline ignores rather than adding them in this file -- keeping the details alongside the code feels like it should be a lot more sustainable in the long term.

Issue #4979, opened 2026-02-04
Pause defaulting on Optimized Checkout Suite for new installs


daledupreez:
We want to make some shopper UX improvements to Optimized Checkout Suite (OCS), as there are various situations where the current OCS shopper experience is fairly poor. We started enabling OCS by default for new installs in 10.2.0, but we want to pause that behaviour for the time being.

PR #5365, opened 2026-04-27
Enable OCS and Adaptive Pricing for new installs

diegocurbelo:
Fixes #<github_issue_id>

Re-enables the Optimized Checkout Suite (OCS) default-on flow for new installs (originally added in [#4811](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4811), reverted in [#4981](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4981) pending UX work tracked in [#4979](https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4979)) and bundles Adaptive Pricing into the same flow.

The existing two-step pattern is reused:

1. `WC_Stripe::install()` writes the `wc_stripe_optimized_checkout_default_on` marker option when `wc_stripe_version` is not yet set.
2. On the next OAuth Connect (`type=connect`) completion in `WC_Stripe_Connect::save_stripe_keys()`, the marker is consumed and **both** `optimized_checkout_element` and `adaptive_pricing` are written to `woocommerce_stripe_settings` as `'yes'`. The marker is deleted regardless of whether it was applied.

- The default is intentionally scoped to **fresh installs that connect via WooCommerce Connect OAuth**. Existing merchants (`wc_stripe_version` already populated) are untouched, and merchants connecting via Stripe App or manual API keys (`type=app`) are not affected.
- Adaptive Pricing is coupled to OCS because its UI is gated behind `WC_Stripe_Helper::is_adaptive_pricing_supported()`, which already requires Checkout Sessions / OCS and applies the existing region/currency restrictions (India/EEA exclusions still apply at runtime).

## Testing instructions

1. From a clean state (no `wc_stripe_version` option), activate the plugin. Confirm `wc_stripe_optimized_checkout_default_on` is set in `wp_options`.
2. Run the WooCommerce Connect OAuth flow (Stripe settings → **Connect or create a new test account**) and complete it.
3. **Expected** `woocommerce_stripe_settings['optimized_checkout_element'] = 'yes'` and `['adaptive_pricing'] = 'yes'`. The `wc_stripe_optimized_checkout_default_on` option is gone.
4. In **Stripe → Settings**, confirm both **Optimized Checkout Suite** and **Adaptive Pricing** checkboxes are enabled.
5. Verify the Optimized Checkout renders on classic checkout, Blocks checkout, and the Optimized Checkout view, and the Adaptive Pricing currency selector appears for supported regions.
6. **Negative case (upgrade)**: the easiest way is to create a JN site with version `10.6.1` and then upload a zip built from this branch. Then confirm the marker is **not** written and existing `woocommerce_stripe_settings` are unchanged.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/tests/phpunit/class-wc-stripe-connect-test.php
+++ b/tests/phpunit/class-wc-stripe-connect-test.php
@@ -54,7 +54,7 @@ class WC_Stripe_Connect_Test extends WC_Mock_Stripe_API_Unit_Test_Case {
 	 */
 	public function test_save_stripe_keys_defaults_ocs_and_adaptive_pricing_for_new_installs( bool $marker_set, string $type, string $expected_oc, string $expected_ap ): void {
 		if ( $marker_set ) {
-			update_option( 'wc_stripe_optimized_checkout_default_on', true );
+			update_option( 'wc_stripe_optimized_checkout_default_on', 'yes' );
 		}
 
 		$result                 = new stdClass();
--- a/tests/phpunit/class-wc-stripe-connect-test.php
+++ b/tests/phpunit/class-wc-stripe-connect-test.php
@@ -1,0 +1,105 @@
+<?php
+
+/**
+ * Tests for the WC_Stripe_Connect class.
+ *
+ * @package WooCommerce/Stripe/WC_Stripe_Connect
+ */
+class WC_Stripe_Connect_Test extends WC_Mock_Stripe_API_Unit_Test_Case {
+
+	/**
+	 * @var WC_Stripe_Connect
+	 */
+	private $connect;
+
+	/**
+	 * @inheritDoc
+	 */
+	public function set_up() {
+		parent::set_up();
+
+		// The settings-update filter merges defaults when saving settings, which would
+		// inject `optimized_checkout_element` / `adaptive_pricing` defaults and obscure
+		// the Connect consumer's explicit writes.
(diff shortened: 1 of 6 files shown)


tests/phpunit/class-wc-stripe-connect-test.php
```diff
@@ -0,0 +1,105 @@
...
+	 * @param string $type             Connection type ('connect' or 'app').
+	 * @param string $expected_oc      Expected value of optimized_checkout_element after save.
+	 * @param string $expected_ap      Expected value of adaptive_pricing after save.
+	 *
+	 * @dataProvider provide_save_stripe_keys_defaults_ocs_for_new_installs
+	 */
+	public function test_save_stripe_keys_defaults_ocs_and_adaptive_pricing_for_new_installs( bool $marker_set, string $type, string $expected_oc, string $expected_ap ): void {
+		if ( $marker_set ) {
+			update_option( 'wc_stripe_optimized_checkout_default_on', 'yes' );
+		}
```
daledupreez, line 60: Not blocking, but it may be worth removing the option when `$marker_set` is false. Or possibly having the argument be a nullable string so we test `null` (deleted), `'yes'`, and `'no'` values.
```suggestion
		if ( $marker_set ) {
			update_option( 'wc_stripe_optimized_checkout_default_on', 'yes' );
		} else {
			delete_option( 'wc_stripe_optimized_checkout_default_on' );
		}
```

daledupreez (approved):
The changes look good to me and test well.

PR #5230, opened 2026-04-06
Display a authorization notice inside the Pre-Authorized Debit (ACSS) payments element

diegocurbelo:
Fixes [STRIPE-567](https://linear.app/a8c/issue/STRIPE-567/globalstep-account-number-input-box-fails-to-appear-after-selecting)<br>Fixes [STRIPE-567](https://linear.app/a8c/issue/STRIPE-567/globalstep-account-number-input-box-fails-to-appear-after-selecting)

## Changes proposed in this Pull Request:

Adds a "After submission, you will need to authorize the payment with your bank" notice for Pre-Authorized Debit (ACSS) payments, similar to what other payment methods currently show.

This is needed because the Stripe Payment Element renders empty for the ACSS payment method.

* *Classic checkout*: Overrides payment_fields() in WC_Stripe_UPE_Payment_Method_ACSS to render the notice with an inline SVG icon below the payment element.
* *Blocks checkout*: Adds a new AcssMessageElement React component rendered conditionally when ACSS is the selected payment method.

**Pre-Authorized Debit (ACSS)**

<table>
<thead>
<tr>
<th>Blocks</th>
<th>Classic</th>
</tr>
</thead>
<tbody>
<tr>
<td>Screenshot 2026-04-06 at 16 42 17</td>
<td>Screenshot 2026-04-06 at 23 02 51</td>
</tr>
</tbody>
</table>

## Testing instructions

* Enable Pre-Authorized Debit (ACSS) as a payment method in Stripe settings
* Go to the classic checkout page and select ACSS — verify the "After submission, you will need to authorize the payment with your bank." message appears
* Go to the blocks checkout page and select ACSS — verify the same message appears
* Complete a purchase with ACSS and verify the payment flow works correctly
* Verify other payment methods (e.g. card, Klarna) are not affected and do not show the message (no double message)

---

- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

### Changelog entry

- [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details><summary>Changelog Entry Comment</summary>

#### Comment

</details>

**Post merge**

- [ ] Added testing instructions to the [Release Testing Instructions wiki page](<https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions>) (or does not apply)
- [ ] Added the [needs docs label](<https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs>) (or does not apply)
[...]

PR #5366, opened 2026-04-28
Floor PHP test matrix WP backlevel from L-2 to L-1

wjrosa:
Fixes the red WP=L-2 cells across every recent PR.

The PR PHP-tests matrix introduced in #5307 pins WP backlevel to L-2 (currently 6.7.5). Every released WC major in the matrix (`L` = 10.7.0 and `L-2` = 10.5.3) requires WP >= 6.8, so the WP=L-2 cells fail at WC plugin install:

```
Installing WooCommerce (10.5.3)
Warning: The package could not be installed.
        \"Your WordPress version is 6.7.5, however the uploaded plugin requires 6.8.\"
Plugin update failed.
Error: No plugins installed.
##[error]Process completed with exit code 1.
```

PHPUnit never runs in those cells, so the failures are environmental, not test-code. All six `PHP × WC` combinations against WP=L-2 fail identically.

This PR makes two surgical changes:

1. **Move WP backlevel from L-2 to L-1.** With L-1 = 6.8.x, every WC release in the matrix installs cleanly. We still test against one older WP. Adds an inline comment explaining the constraint so the next reviewer doesn't revert without checking WC's WP requirement first.

2. **Exclude PHP 8.5 × WP=L-1 specifically.** WP 6.8 ships pre-PHP-8.5 casts (`(double)` / `(boolean)`) in `wp-includes/IXR/class-IXR-message.php`; PHP 8.5 emits those as `E_DEPRECATED` and PHPUnit's default error handler promotes the notice to a test error on every test that loads core. WP 6.9 (= WP=L) patched the casts, so the PHP=8.5 × WP=L cells still exercise the upcoming-deprecation signal we wanted out of the 8.5 row. PHP 7.4 / 8.2 against WP=L-1 stay in the matrix.

Final shape: **10 jobs** instead of the original 12 — the 2 dropped cells are exactly the PHP=8.5 × WP=L-1 combinations that have no released WP patch.

### Why these specific moves

- **Couldn't keep WP=L-2.** Both WC versions in the matrix require WP >= 6.8; rolling WC further back than the loose-L-2 policy would let us reach a 6.7-compatible WC, but that breaks the WC L-2 contract more than dropping WP L-2 does.
- **Couldn't disable PHPUnit's deprecation-to-error promotion globally.** We want to keep catching deprecations in our own code.
- **Couldn't drop PHP 8.5 entirely.** It's the upcoming-deprecation signal the matrix was sized to give; WP=L (6.9) was already patched, so PHP 8.5 × WP=L still exercises it.
- **Couldn't pin a different WP 6.8 patch.** The IXR cast issue is unfixed across all WP 6.8.x releases.

### How can this code break?
- **Test coverage regression on WP 6.7?** Yes — we're consciously dropping it because no released WC version we'd ship with supports it. AGENTS.md's \"WordPress strict L-2\" rule technically still applies but is currently unenforceable. The comment makes the deferral explicit.
- **Test coverage regression on PHP 8.5 against WP 6.8?** Yes, but a backport of the IXR cast fix to WP 6.8.x would let us drop the exclude. Worth revisiting on the next 6.8 minor.
- **Future WC bumps WP minimum past 6.8?** L-1 will hit the same wall. We'd need to either change the policy text or roll WC further back than the \"loose L-2\" policy currently allows. Worth a follow-up doc PR if/when that happens.

## Testing instructions

This change is observable only in CI:
1. Open this PR — the workflow runs.
2. Confirm all 10 cells pass through to PHPUnit (no \"plugin requires 6.8\" install error, no IXR-cast deprecation explosions).
3. Confirm the matrix shows 10 jobs (the original 12 minus the two excluded PHP 8.5 × WP=L-1 cells).

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️) — workflow change verified by CI itself
-   [x] Tested on mobile (or does not apply) — does not apply

### Changelog entry

<details>

#### Comment

CI workflow change with no impact on shipped plugin behavior.

</details>

**Post merge**

### DIFF
--- a/.github/workflows/php-tests.yml
+++ b/.github/workflows/php-tests.yml
@@ -43,6 +43,13 @@ jobs:
         woocommerce_support_policy: ['L', 'L-2']
         wordpress_support_policy: ['L', 'L-1']
         php: ['7.4', '8.2', '8.5']
+        # Skip PHP 8.5 + WP L-1: WP 6.8 ships pre-PHP-8.5 casts in
+        # wp-includes/IXR/class-IXR-message.php (`(double)` / `(boolean)`),
+        # which PHP 8.5 emits as E_DEPRECATED and PHPUnit promotes to test
+        # errors on every test that loads core. WP 6.9 patched the casts.
+        exclude:
+          - php: '8.5'
+            wordpress_support_policy: 'L-1'
         include:
           # WooCommerce
           - woocommerce_support_policy: L
--- a/.github/workflows/php-tests.yml
+++ b/.github/workflows/php-tests.yml
@@ -39,8 +39,9 @@ jobs:
       fail-fast:    false
       max-parallel: 16
       # Try to keep basic matrix small to stay within our runner quota.
-      # The current matrix includes 12 combos (2 WC × 2 WP × 3 PHP),
-      # with coverage of PHP 7.4, 8.2, and 8.5.
+      # The current matrix includes 10 combos (2 WC × 2 WP × 3 PHP, minus
+      # the 2 PHP-8.5 × WP-L-1 exclusions below), with coverage of PHP
+      # 7.4, 8.2, and 8.5.
       #
       # WP backlevel is L-1, not L-2: every released WC major in our matrix
       # (the L and L-2 cells) requires WP >= 6.8, and WP L-2 is 6.7. Pinning
--- a/.github/workflows/php-tests.yml
+++ b/.github/workflows/php-tests.yml
@@ -54,6 +54,15 @@ jobs:
         woocommerce_support_policy: [ 'L', 'L-2' ]
         wordpress_support_policy:   [ 'L', 'L-1' ]
         php:                        [ '7.4', '8.2', '8.5' ]
+        # Skip PHP 8.5 + WP L-1: WP 6.8 ships pre-PHP-8.5 casts in
+        # wp-includes/IXR/class-IXR-message.php (`(double)` / `(boolean)`),
+        # which PHP 8.5 emits as E_DEPRECATED and PHPUnit promotes to test
+        # errors on every test that loads core. WP 6.9 patched the casts,
+        # so PHP 8.5 × WP=L still exercises the upcoming-deprecation
+        # signal we wanted out of the 8.5 row.
+        exclude:
+          - php: '8.5'
+            wordpress_support_policy: 'L-1'
         include:
           # WooCommerce
           - woocommerce_support_policy: L
--- a/.github/workflows/php-tests.yml
+++ b/.github/workflows/php-tests.yml
@@ -41,9 +41,18 @@ jobs:
       # Try to keep basic matrix small to stay within our runner quota.
       # The current matrix includes 12 combos (2 WC × 2 WP × 3 PHP),
       # with coverage of PHP 7.4, 8.2, and 8.5.
+      #
+      # WP backlevel is L-1, not L-2: every released WC major in our matrix
+      # (the L and L-2 cells) requires WP >= 6.8, and WP L-2 is 6.7. Pinning
+      # WP=L-2 here makes WC's installer reject the WP version and the
+      # whole job fails before PHPUnit runs. Once a future WC major lifts
+      # its WP minimum past 6.9 we can revisit, but until WC ships a
+      # release that supports WP L-2 again we have no way to honour the
+      # strict L-2 policy from AGENTS.md without rolling WC back further
+      # than the loose L-2 policy allows.
       matrix:
         woocommerce_support_policy: [ 'L', 'L-2' ]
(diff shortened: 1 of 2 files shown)


.github/workflows/php-tests.yml
```diff
@@ -41,10 +41,28 @@ jobs:
       # Try to keep basic matrix small to stay within our runner quota.
       # The current matrix includes 12 combos (2 WC × 2 WP × 3 PHP),
       # with coverage of PHP 7.4, 8.2, and 8.5.
```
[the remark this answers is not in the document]
wjrosa: Good catch — fixed in 6800657: comment now reads `10 combos (2 WC × 2 WP × 3 PHP, minus the 2 PHP-8.5 × WP-L-1 exclusions below)`.

.github/workflows/php-tests.yml
```diff
@@ -39,12 +39,31 @@ jobs:
...
+      # 7.4, 8.2, and 8.5.
+      #
+      # WP backlevel is L-1, not L-2: every released WC major in our matrix
+      # (the L and L-2 cells) requires WP >= 6.8, and WP L-2 is 6.7. Pinning
+      # WP=L-2 here makes WC's installer reject the WP version and the
+      # whole job fails before PHPUnit runs. Once a future WC major lifts
+      # its WP minimum past 6.9 we can revisit, but until WC ships a
+      # release that supports WP L-2 again we have no way to honour the
+      # strict L-2 policy from AGENTS.md without rolling WC back further
+      # than the loose L-2 policy allows.
```
diegocurbelo: This is really verbose... We should replace it with a short phrase like:

`WC support policy for WP is L and L-1 (https://woocommerce.com/support-policy/)`


And we should update the `AGENTS.md` accordingly, to reflect that the Stripe plugin needs to support `WC L, L-1 and L-2` and transitively `WP L and L-1`

.github/workflows/php-tests.yml
diegocurbelo: This should be 2 to align with the WC support policy

```suggestion
          version-count: 2
```

diegocurbelo (commented):
This works as expected (skips WP L-2 = `6.7`).

But I think we should update the comments (and the AGENTS.md) to reflect that the change is needed because the WC policy, WP L-2 is not supported by WC.

.github/workflows/php-tests.yml
```diff
@@ -39,12 +39,31 @@ jobs:
...
         php:                        [ '7.4', '8.2', '8.5' ]
+        # Skip PHP 8.5 + WP L-1: WP 6.8 ships pre-PHP-8.5 casts in
+        # wp-includes/IXR/class-IXR-message.php (`(double)` / `(boolean)`),
+        # which PHP 8.5 emits as E_DEPRECATED and PHPUnit promotes to test
+        # errors on every test that loads core. WP 6.9 patched the casts,
+        # so PHP 8.5 × WP=L still exercises the upcoming-deprecation
+        # signal we wanted out of the 8.5 row.
+        exclude:
+          - php: '8.5'
+            wordpress_support_policy: 'L-1'
```
diegocurbelo: This might not be necessary... for example, in [this PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5230/) only the `WP L-2` version checks fail, the `WP L-1` pass

.github/workflows/php-tests.yml
wjrosa: Done in 1e1f147f. Replaced the block with the one-liner pointing to WC's support policy and updated AGENTS.md to state WC L/L-1/L-2 → transitively WP L/L-1.

.github/workflows/php-tests.yml
wjrosa: Done in 1e1f147f.

.github/workflows/php-tests.yml
wjrosa: Good call — dropped the exclusion in 1e1f147f. PR #5230 shows WP L-1 passes on PHP 8.5, so the IXR-message concern was a 6.7-only issue.

diegocurbelo: My bad, the other PRs were already not using `L-1` that's why they pass.... we still need to exclude `PHP 8.5` + `WP L-1` .

Sorry for the confusion 😞

wjrosa: No worries! Looking at that PR more closely — its matrix was `[L, L-2]` (the pre-this-PR shape), so the passing column there was WP=L (6.9), not WP=L-1; L-1 wasn't exercised at all. The `(double)`/`(boolean)` casts in `wp-includes/IXR/class-IXR-message.php` are still present in 6.8 and only got patched in 6.9, so PHP 8.5 × WP=L-1 will fail. Restored the exclusion in 746ba925e.

diegocurbelo (approved):
LGTM 👍🏼

PR #5368, opened 2026-04-28
Pass Stripe auth headers through Playground CORS proxy

wjrosa:
> ⏸️ **Parked pending upstream Playground fix** — [WordPress/wordpress-playground#3559](https://github.com/WordPress/wordpress-playground/issues/3559).
>
> The mu-plugin in this PR is correct in isolation, but Playground's service-worker `fetchWithCorsProxy` only retries through the CORS proxy when the direct fetch *throws*. Stripe's permissive CORS makes the direct fetch resolve with status 401 (the browser strips JS-set `Authorization` because the request isn't `credentials: 'include'`), so the proxy retry never fires and the mu-plugin's opt-in header has no effect. Until upstream lands one of the three fixes proposed in the linked issue, account-bound Stripe flows can't work in Playground previews. End-to-end debugging trail in [#5368 (comment)](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5368#issuecomment-4339110774).

Fixes _N/A — follow-up to #5074 addressing the Playground CORS limitation surfaced in [#5074 (comment)](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5074#issuecomment-4309094827)._

Stacked on #5074. Lets the Stripe REST API actually work inside the Playground PR preview, so reviewers can exercise account connect / payment-method listing / checkout-against-Stripe instead of only admin UI shape.

**Root cause.** Playground runs PHP in a browser WASM sandbox; PHP HTTP calls (`wp_safe_remote_post`, cURL) are tunneled through the browser's `fetch()`. `api.stripe.com` doesn't return CORS headers permitting `playground.wordpress.net`, so the direct fetch fails. Since [WP/wp-playground#2076](https://github.com/WordPress/wordpress-playground/pull/2076) Playground retries through its CORS proxy at `cors.wordpress.net/proxy.php` — but [the proxy strips `Authorization` headers in both directions by default](https://github.com/WordPress/wordpress-playground/tree/trunk/packages/playground/php-cors-proxy) ("authentication headers are blocked in both directions … refuses to pass authentication credentials"). Stripe receives an unauthenticated request, returns 401, and the plugin surfaces it as *"test API keys we've saved for you are no longer valid"*. Payment methods don't list because `WC_Stripe_UPE_Payment_Gateway` gates method availability on `accounts/retrieve` capabilities, and that call has just been 401'd.

**Fix.** A Playground-only must-use plugin, written into the sandbox via a blueprint `writeFile` step (never shipped with the released plugin zip), hooks `http_request_args` for `api.stripe.com` URLs and adds the proxy's documented per-request opt-in:

```
X-Cors-Proxy-Allowed-Request-Headers: Authorization, Idempotency-Key, Stripe-Version, Stripe-Account
```

Webhook delivery still won't work in the sandbox (no inbound HTTP into the WASM runtime); the sticky comment is updated to reflect the new scope.

- New `.github/workflows/scripts/playground-mu-plugins/stripe-cors-headers.php`: the mu-plugin itself, URL-prefix-gated on `https://api.stripe.com/`.
- `.github/workflows/scripts/generate-playground-blueprint.js`: read the mu-plugin source at workflow runtime (`fs.readFileSync` + `__dirname`), prepend a `writeFile` step that drops it at `/wordpress/wp-content/mu-plugins/`, and update the sticky-comment scope copy.
- `.github/workflows/pr-build-live-branch.yml`: un-ignore `scripts/playground-mu-plugins/**` so edits to the mu-plugin retrigger the workflow.

**How can this code break?**
- Playground's CORS proxy could change its allowlist behavior or rename the opt-in header. Symptom is a regression to "test API keys are no longer valid" inside the preview — caught by the testing steps below.
- A future Stripe-side change adding new authentication-class headers would silently get stripped. The mu-plugin's allowlist is conservative; we extend it as new headers appear rather than allow-all (which would re-introduce abuse risk for the shared proxy).
- The mu-plugin runs on every Playground sandbox boot for this repo. Malformed PHP would break preview boot — caught by phpcs/PHPStan in CI plus the testing steps.

- The `http_request_args` filter is URL-prefix-gated on `https://api.stripe.com/` so it cannot affect non-Stripe outbound traffic in the sandbox.
- The mu-plugin lives only in `.github/workflows/scripts/` and is never copied into the released plugin zip — released-plugin behavior is unchanged.
- The blueprint generator reads the mu-plugin via `fs.readFileSync` at workflow runtime, so the file is reviewable and lintable as PHP rather than embedded as a JS string blob.
- Path-filter un-ignore on `scripts/playground-mu-plugins/**` ensures future edits trigger a rebuild instead of silently shipping a stale Playground bundle.

## Testing instructions

1. After this PR's CI run completes, click the freshly-updated Playground sticky comment link and let the sandbox boot.
2. Paste valid Stripe **test** keys into the Stripe Dev Tools settings screen.
3. Confirm the connection step succeeds (the connected-account block replaces the "test API keys we've saved for you are no longer valid" warning).
4. Navigate to **WooCommerce → Settings → Payments → Stripe** and confirm payment methods populate (Cards plus whatever methods the test account has capabilities for).
5. Open browser devtools → Network. Trigger any settings save or checkout that hits Stripe and confirm:
   - Requests are routed via `cors.wordpress.net/proxy.php?…api.stripe.com/…`
   - The request includes header `X-Cors-Proxy-Allowed-Request-Headers: Authorization, Idempotency-Key, Stripe-Version, Stripe-Account`
   - The proxy returns **200** (not 401)
   - The matching request appears in the Stripe dashboard logs for that test account
6. Add a product to cart, proceed to checkout, and confirm a `4242…` test card produces a successful Stripe `payment_intent` (visible in Stripe dashboard).
7. Edit `.github/workflows/scripts/playground-mu-plugins/stripe-cors-headers.php` (e.g., add a comment), push, and confirm the workflow re-runs (the path-filter un-ignore is what makes that work).
8. Negative case: spin up `npm run up` locally and confirm `wp-content/mu-plugins/stripe-cors-headers.php` is **not** present — the mu-plugin must only exist inside the Playground sandbox.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️) — CI/preview tooling change; coverage is the manual Playground walkthrough above.
-   [x] Tested on mobile (or does not apply) — preview tooling, no user-facing UI.

### Changelog entry

<details>

#### Comment

Internal CI/preview tooling only. The mu-plugin is written into the Playground sandbox via a blueprint step at workflow runtime and never ships in the released plugin zip, so there is no user-visible change.

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply) — does not apply (CI tooling).
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply) — does not apply.
-   [ ] Included this PR in the Release Thread scope (or does not apply) — does not apply.

### DIFF
--- a/.github/workflows/pr-build-live-branch.yml
+++ b/.github/workflows/pr-build-live-branch.yml
@@ -16,6 +16,7 @@ on:
             - '.github/**'
             - '!.github/workflows/pr-build-live-branch.yml'
             - '!.github/workflows/scripts/generate-playground-blueprint.js'
+            - '!.github/workflows/scripts/playground-mu-plugins/**'
             - '**/tests/**'
             - 'docker/**'
             - '*.md'
--- a/.github/workflows/scripts/generate-playground-blueprint.js
+++ b/.github/workflows/scripts/generate-playground-blueprint.js
@@ -121,7 +121,7 @@ The changes in this pull request can be previewed and tested using a [WordPress 
 
 [Test this pull request with WordPress Playground](${ url }).
 
-**Scope:** Playground runs PHP in a browser WASM sandbox. Outbound calls to \`api.stripe.com\` are routed through Playground's CORS proxy, which the bundled mu-plugin opts the Stripe \`Authorization\` header through — so account connect, payment-method listing, and checkout requests should work. Webhook delivery still won't (no inbound HTTP into the sandbox); for anything webhook-driven, use a real local environment (\`npm run up\`).
+**Scope:** Playground runs PHP in a browser WASM sandbox. Stripe REST API calls do **not** work end-to-end in this preview — Playground's service worker strips JS-set \`Authorization\` headers on direct fetches before falling back to its CORS proxy ([WordPress/wordpress-playground#3559](https://github.com/WordPress/wordpress-playground/issues/3559)), so account connect, payment-method listing, checkout-against-Stripe, and webhook delivery all fail. Use this preview to review admin UI, settings shape, and install/activation flow only; use a real local environment (\`npm run up\`) for anything that talks to Stripe.
 
 Note that this URL is valid for 30 days from when this comment was last updated. You can update it by closing/reopening the PR or pushing a new commit.
 `;
--- a/.github/workflows/scripts/generate-playground-blueprint.js
+++ b/.github/workflows/scripts/generate-playground-blueprint.js
@@ -1,5 +1,21 @@
+const fs = require( 'fs' );
+const path = require( 'path' );
+
+const stripeCorsMuPlugin = fs.readFileSync(
+	path.join( __dirname, 'playground-mu-plugins/stripe-cors-headers.php' ),
+	'utf8'
+);
+
 const generatePlaygroundBlueprint = ( runId, prNumber, devToolsAvailable ) => {
 	const steps = [
+		{
+			/* Playground's CORS proxy strips Authorization headers by default.
+			This mu-plugin opts Stripe's auth headers through the proxy so
+			account retrieval and payment-method listing work in the sandbox. */
+			step: 'writeFile',
+			path: '/wordpress/wp-content/mu-plugins/stripe-cors-headers.php',
+			data: stripeCorsMuPlugin,
+		},
 		{
 			step: 'installPlugin',
 			pluginData: {
@@ -98,7 +114,7 @@ The changes in this pull request can be previewed and tested using a [WordPress 
 
 [Test this pull request with WordPress Playground](${ url }).
 
-**Scope:** Playground runs PHP in a browser WASM sandbox, so outbound calls to \`api.stripe.com\` are subject to browser CORS. In practice, account connection, payment-method listing, checkout against Stripe, and webhook delivery will not work end-to-end here. Use the preview to review admin UI, settings shape, and install/activation flow; use a real local environment (\`npm run up\`) for anything that needs to talk to Stripe.
+**Scope:** Playground runs PHP in a browser WASM sandbox. Outbound calls to \`api.stripe.com\` are routed through Playground's CORS proxy, which the bundled mu-plugin opts the Stripe \`Authorization\` header through — so account connect, payment-method listing, and checkout requests should work. Webhook delivery still won't (no inbound HTTP into the sandbox); for anything webhook-driven, use a real local environment (\`npm run up\`).
 
 Note that this URL is valid for 30 days from when this comment was last updated. You can update it by closing/reopening the PR or pushing a new commit.
 `;
--- a/.github/workflows/scripts/generate-playground-blueprint.js
+++ b/.github/workflows/scripts/generate-playground-blueprint.js
@@ -105,9 +105,16 @@ async function run( { github, context, core } ) {
 		devToolsAvailable
 	);
 
-	const url = `https://playground.wordpress.net/#${ JSON.stringify(
-		blueprint
-	) }`;
+	// Base64-encode the blueprint so paren-, bracket-, and quote-containing
+	// payloads (e.g. PHP source written via the writeFile step) don't break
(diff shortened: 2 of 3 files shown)


wjrosa:
## Parking this PR pending an upstream Playground fix

After several rounds of debugging with @malithsen and @wjrosa testing in the Playground sandbox, the practical outcome is: **this PR's mu-plugin alone cannot make Stripe REST API calls work in Playground previews.** The bottleneck lives in Playground's service-worker layer, not in anything we control here.

### What the mu-plugin actually does — and why it's not enough

The mu-plugin correctly sets `X-Cors-Proxy-Allowed-Request-Headers: authorization, …` on every outbound `api.stripe.com` request. That opt-in header is the documented mechanism for letting the proxy forward `Authorization` to the target. Server-side (`packages/playground/php-cors-proxy/cors-proxy-functions.php`) it works exactly as intended.

The problem is in [`packages/php-wasm/web-service-worker/src/fetch-with-cors-proxy.ts`](https://github.com/WordPress/wordpress-playground/blob/trunk/packages/php-wasm/web-service-worker/src/fetch-with-cors-proxy.ts):

```ts
try {
    return await fetch(requestObject);   // direct attempt
} catch {
    // proxy retry with credentials: 'include' when opt-in header is set
}
```

The retry is reached only when the direct fetch *throws*. Stripe's CORS preflight is fully permissive — it echoes back any origin, any header, with `Access-Control-Allow-Credentials: true`. So the direct fetch from Playground to `api.stripe.com` *resolves* (no exception). It just resolves with **status 401**, because the browser strips the JS-set `Authorization` header on a CORS-tainted request that isn't `credentials: 'include'` — exactly what the spec requires ([Fetch §8.10.2](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch)).

The proxy retry never fires. The mu-plugin's opt-in header is set but unused.

End-to-end confirmation from a sandbox session:
- Stripe's response body, captured from the WC Stripe error log: `{"error":{"message":"You did not provide an API key. …","type":"invalid_request_error"}}`.
- Zero traffic to `cors.wordpress.net/proxy.php` in the browser Network tab.
- Plugin's own 5-401 circuit breaker (`WC_Stripe_API::INVALID_API_KEY_ERROR_COUNT_THRESHOLD`) trips within a single page load (UPE init calls `accounts/retrieve` once per payment method) and locks out new requests for 2 hours, so the symptom looks worse than the underlying cause.

### Upstream filed

I've filed [WordPress/wordpress-playground#3559](https://github.com/WordPress/wordpress-playground/issues/3559) with a minimal Stripe repro and three proposed fixes, in roughly conservative order:

1. Skip the direct fetch entirely when `X-Cors-Proxy-Allowed-Request-Headers` is present.
2. Treat 4xx as a retry trigger when the opt-in header is present.
3. Set `credentials: 'include'` on the direct fetch when the opt-in indicates `authorization` or `cookie`.

Any one of those would unblock this PR.

### What I'd suggest doing with this PR

- Keep it parked as a **draft**. The mu-plugin is technically correct — the moment the upstream fix lands, this becomes a one-line revisit. Closing and re-opening would lose the discussion thread.
- Land #5074 with the cautious sticky-comment scope copy ("Stripe API calls won't work end-to-end in this preview — UI/install-flow review only"). That's already its current state on the branch and accurately describes what the preview does today.
- Don't ship a mocked / fixture-based Stripe shim in Playground previews unless we explicitly want a different product. It's a real maintenance burden and doesn't catch regressions in real Stripe integration paths.

I'll keep an eye on the upstream issue and revisit this PR when there's movement.

PR #5369, opened 2026-04-28
[RSM - Diagnostics Mode] Wire frontend recorder with backend endpoint

malithsen:

This PR fixes a couple of issues that was preventing the events from the frontend recorder #5330 from being persisted via the endpoint in #5350 

1. The recorder posts via `navigator.sendBeacon` with a plain string, which sets Content-Type to `text/plain;charset=UTF-8`. WP's REST server only auto-decodes JSON bodies tagged `application/json`, so the controller's parameter lookup were failing. To fix that, this PR wraps the JSON payload in a `Blob` typed `application/json` before ahding to `sendBeacon`. 
2.  The controller's `args` required `sessionId`, but the recorder sends `diag_session_id`. The controller also did its own `wp_verify_nonce` against a custom `wc_stripe_diagnostics` action read from neither location the recorder actually populates, and a custom action that conflicts with WP's cookie-auth layer (which validates the same `_wpnonce` against `wp_rest`). This PR drops the controller's nonce check and trusts WP's cookie-auth layer for nonce validation. It's worth noting that the endpoint is unauthenticated by design. It's Abuse mitigation comes from the `is_enabled()` toggle and the trace store's caps (200 events/trace, FIFO trace-count eviction).

## Testing instructions
Since we don't yet have the merchant toggle to enable diagnostics, testing this requires a code snippet.

- Create `includes/diagnostics/class-wc-stripe-diagnostics-stub.php` with the following content: 3b18d-pb/#php
- Update `includes/class-wc-stripe.php` to import the above file in line 106 with `( new WC_Stripe_Diagnostics_Stub() )->init();`

1. Navigate to checkout and place an order.
2. In dev tools -> Network, notice POST requests to the `/events` endpoint
3. Confirm the response is `200 { written: <n> }` with `n` matching the number of events sent.
4. Confirm a trace file appears in `wp-content/uploads/wc-stripe-diagnostics/`.
5. Repeat as a logged-out shopper to confirm cookie-auth still gates guest traffic correctly.

---

### Changelog entry

<details>

#### Comment 
Fixes issues in an unreleased feature.
</details>

**Post merge**

### DIFF
--- a/client/diagnostics/recorder.js
+++ b/client/diagnostics/recorder.js
@@ -38,7 +38,7 @@ export class Recorder {
 		// Replay any buffer left over from a prior pageload (e.g. 3DS redirect)
 		// then clear it from sessionStorage to prevent a second replay on the next boot.
 		if ( persisted?.bufferedEvents?.length ) {
-			const payload = JSON.stringify( {
+			const payload = buildIngestBlob( {
 				diag_session_id: this.config.sessionId,
 				events: persisted.bufferedEvents,
 			} );
@@ -255,7 +255,7 @@ export class Recorder {
 		if ( ! this.config?.active || this.buffer.length === 0 ) {
 			return;
 		}
-		const payload = JSON.stringify( {
+		const payload = buildIngestBlob( {
 			diag_session_id: this.config.sessionId,
 			events: this.buffer,
 		} );
@@ -312,6 +312,15 @@ export function getRecorder() {
 		_singleton.boot();
 	}
 	return _singleton;
+}
+
+// Wrap the JSON payload in a Blob so sendBeacon sets Content-Type to
+// application/json. Without this, sendBeacon defaults to text/plain and
+// WP's REST server skips JSON body parsing.
+function buildIngestBlob( payload ) {
+	return new Blob( [ JSON.stringify( payload ) ], {
+		type: 'application/json',
+	} );
 }
 
 function buildIngestUrl( config ) {
--- a/client/diagnostics/recorder.js
+++ b/client/diagnostics/recorder.js
@@ -1,5 +1,9 @@
 const STORAGE_KEY_PREFIX = 'wc_stripe_diag_';
 const IDLE_FLUSH_MS = 5000;
+// Server caps a batch at 200 events; flush early so a noisy checkout
+// (e.g., intercepted console.warn/error) can't push past the cap and
+// have the controller silently truncate.
+const MAX_BUFFER_BEFORE_FLUSH = 200;
 const CONSOLE_MESSAGE_MAX_CHARS = 500;
 const CONSOLE_LEVELS = [ 'warn', 'error' ];
 
@@ -248,6 +252,10 @@ export class Recorder {
 			kind,
 			data,
 		} );
+		if ( this.buffer.length >= MAX_BUFFER_BEFORE_FLUSH ) {
+			this.flush();
+			return;
+		}
 		this._resetIdleTimer();
 	}
 
--- a/includes/admin/class-wc-rest-stripe-diagnostics-controller.php
+++ b/includes/admin/class-wc-rest-stripe-diagnostics-controller.php
@@ -57,8 +57,11 @@ class WC_REST_Stripe_Diagnostics_Controller extends WP_REST_Controller {
 	 * Permission callback.
 	 *
 	 * The endpoint is intentionally shopper-facing and unauthenticated.
-	 * Abuse mitigation comes from the `is_enabled()` toggle and the
-	 * trace store's caps (200 events/trace, FIFO trace-count eviction).
+	 * Abuse mitigation relies on the bounded debug window — the
+	 * `is_enabled()` toggle is meant to be on only briefly while a
+	 * merchant is actively debugging. Don't default it on: with the
+	 * toggle on, anyone can rotate `diag_session_id` to evict
+	 * legitimate traces from the FIFO trace store.
 	 *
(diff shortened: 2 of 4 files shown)


includes/admin/class-wc-rest-stripe-diagnostics-controller.php
```diff
@@ -40,24 +39,12 @@ public function register_routes(): void {
...
-					'sessionId' => [
-						'required' => true,
-						'type'     => 'string',
-					],
-					'events'    => [
-						'required' => true,
-						'type'     => 'array',
-					],
-				],
 			]
```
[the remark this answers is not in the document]
malithsen: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5369/commits/a717fdb0923f93ff967edbd09d24afb522f73e99

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
```diff
@@ -66,15 +53,6 @@ public function permissions_check( $request ) {
...
-
-		$nonce = $request->get_header( 'x_wp_nonce' );
-		if ( ! is_string( $nonce ) || '' === $nonce ) {
-			$nonce = (string) $request->get_param( 'nonce' );
-		}
-		if ( ! wp_verify_nonce( $nonce, self::NONCE_ACTION ) ) {
-			return new WP_Error( 'wc_stripe_diagnostics_bad_nonce', 'Invalid diagnostics nonce.', [ 'status' => 403 ] );
-		}
-
 		return true;
```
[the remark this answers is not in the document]
malithsen: Added a comment in https://github.com/woocommerce/woocommerce-gateway-stripe/commit/a717fdb0923f93ff967edbd09d24afb522f73e99

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
```diff
@@ -66,15 +67,6 @@ public function permissions_check( $request ) {
...
```
wjrosa, line 73: Docblock frames the trace-store caps as mitigation, but the FIFO trace-count eviction is the abuse *surface*: while the toggle is on, anyone can rotate `diag_session_id` to evict legitimate debug traces. The real mitigation is the bounded debug-window — worth saying that explicitly so the toggle doesn't get defaulted on later.

client/diagnostics/recorder.js
```diff
@@ -314,6 +314,15 @@ export function getRecorder() {
 	return _singleton;
 }
 
+// Wrap the JSON payload in a Blob so sendBeacon sets Content-Type to
+// application/json. Without this, sendBeacon defaults to text/plain and
+// WP's REST server skips JSON body parsing.
+function buildIngestBlob( payload ) {
+	return new Blob( [ JSON.stringify( payload ) ], {
+		type: 'application/json',
```
wjrosa, line 330: `application/json` isn't CORS-safelisted, so this Blob would trigger a preflight if the endpoint ever moves cross-origin (custom REST host, headless WP) — and `sendBeacon` can't preflight. Fine same-origin; worth a one-liner here so we don't re-debate it later.

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
```diff
@@ -85,9 +77,9 @@ public function permissions_check( $request ) {
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function ingest_events( $request ) {
-		$session_id = WC_Stripe_Diagnostics_Trace_Store::sanitize_id( (string) $request->get_param( 'sessionId' ) );
+		$session_id = WC_Stripe_Diagnostics_Trace_Store::sanitize_id( (string) $request->get_param( 'diag_session_id' ) );
```
wjrosa, line 83: Re: line 90 (`array_slice(..., 0, 200)`) — this is silent truncation. `flush()` clears the buffer on success and `sendBeacon` is fire-and-forget so the recorder never sees `written < sent`. And the recorder has no size cap on its buffer — `record()` resets the 5s idle timer, so a steady event stream just keeps growing the buffer until pagehide. A console-noisy checkout (intercepted warn/error) could realistically push past 200 in one batch.

Worth picking one of: (a) include a `truncated` count in the response, (b) return 413 here, or (c) add a size-cap-triggered flush in the recorder.

client/diagnostics/recorder.js
malithsen: Hm, this concern is generic to any other endpoint, we don't have such notes on those so might not make sense to add one here.

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
malithsen: Updated frontend recorder to flush at the cap size https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5369/commits/011f11c6089279228307171c25d2bafde6f3526c

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
malithsen: Good point. Updated docblock in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5369/commits/b875f091b3d3bb69ced6bda5ce720f2dc3c2853f

Also created https://linear.app/a8c/issue/RSM-1638 to implement a long-term solution

wjrosa (approved):
Code is good and works as expected:
Image

Thanks for addressing my points

PR #5378, opened 2026-05-03
[RSM - Diagnostics Mode] Instrument diagnostics mode usage

malithsen:
Fixes RSM-2093

Adds Tracks instrumentation for Diagnostics Mode so we can measure adoption .

- `wcadmin_wcstripe_diagnostics_mode_toggled`: Fired from the server when the diagnostic mode checkbox is saved.
- `wcadmin_wcstripe_diagnostics_copy_traces`: Fired from the client when the diagnostic traces are copied.

## Testing instructions

1. Make sure Usage tracking is enabled for the site.
2. Enable/Disable Diagnostics mode in Stripe Settings page.
3. In ~5 minutes, go to MC -> Tracks Live tracking page
4. Ensure there's an `wcstripe_diagnostics_mode_toggled` event with the correct `enabled` prop values
5. Click the "Copy failed traces for support" button.
6. In MC, there should be an `wcstripe_diagnostics_copy_traces` with correct props.

---

### Changelog entry

<details>

#### Comment 
This is an unreleased feature

</details>

**Post merge**

### DIFF
--- a/client/settings/advanced-settings-section/diagnostics-traces.js
+++ b/client/settings/advanced-settings-section/diagnostics-traces.js
@@ -3,6 +3,7 @@ import { __, _n, sprintf } from '@wordpress/i18n';
 import { Button, Notice } from '@wordpress/components';
 import apiFetch from '@wordpress/api-fetch';
 import { NAMESPACE } from 'wcstripe/data/constants';
+import { recordEvent } from 'wcstripe/tracking';
 
 // Clipboard writes silently truncate or fail above a few MB on most
 // browsers, and ticket forms strip large pastes long before that. Falling
@@ -95,6 +96,8 @@ const DiagnosticsTraces = () => {
 	].filter( Boolean );
 
 	const handleCopy = async ( statuses ) => {
+		// 'support' when filtering to SUPPORT_STATUSES (failed + abandoned), 'all' otherwise.
+		const scope = statuses ? 'support' : 'all';
 		setIsCopying( true );
 		setFeedback( null );
 		try {
@@ -106,6 +109,11 @@ const DiagnosticsTraces = () => {
 
 			if ( bytes > CLIPBOARD_BYTE_THRESHOLD ) {
 				downloadAsFile( json );
+				recordEvent( 'wcstripe_diagnostics_copy_traces', {
+					scope,
+					result: 'download',
+					count: response.count,
+				} );
 				setFeedback( {
 					status: 'success',
 					message: __(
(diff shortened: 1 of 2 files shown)


includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -566,7 +566,20 @@ private function update_is_diagnostics_enabled( WP_REST_Request $request ) {
...
+		$next     = (bool) $is_diagnostics_enabled;
+
+		$this->gateway->update_option( 'diagnostics', $next ? 'yes' : 'no' );
+
+		if ( $previous !== $next && function_exists( 'wc_admin_record_tracks_event' ) ) {
+			wc_admin_record_tracks_event(
+				'wcstripe_diagnostics_mode_toggled',
+				[
+					'enabled'   => $next ? 'yes' : 'no',
+					'test_mode' => WC_Stripe_Mode::is_test() ? 1 : 0,
```
wjrosa, line 579: Mixed types in props make downstream Tracks queries awkward — `enabled` is a `'yes'/'no'` string but `test_mode` is a `1/0` int. Worth picking one convention: either both strings (`'yes'/'no'`, matching the gateway-option style) or both bool-ish ints. Filtering charts with `WHERE enabled = 'yes' AND test_mode = 1` mid-string-mid-int is the kind of thing that bites later.

client/settings/advanced-settings-section/diagnostics-traces.js
```diff
@@ -131,6 +144,11 @@ const DiagnosticsTraces = () => {
 			}
 			refreshSummary();
 		} catch ( err ) {
+			recordEvent( 'wcstripe_diagnostics_copy_traces', {
+				scope,
+				result: 'error',
+				count: 0,
```
wjrosa: `count: 0` here reads in dashboards as "we successfully copied zero traces," which conflates with the empty-result case. The fetch never landed, so we don't actually know a count. Either drop the prop on this branch or use `null`/`undefined` so the error path is distinguishable from a 0-trace success.

wjrosa (commented):
Tight scope, fire-only-on-change gate is right. Two small things on the event property shapes — both about reducing downstream-query friction more than correctness.

client/settings/advanced-settings-section/diagnostics-traces.js
malithsen: Good point. Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5378/commits/7003f8da16e7c9631422ca94d6e961b6018bab55

includes/admin/class-wc-rest-stripe-settings-controller.php
malithsen: I prefer the text values here because they are easier to interpret, and these booleans are casted to strings in the Tracks backend anyway ('0' instead of 0). 

However, the existing pattern for `test_mode` is to use Booleans in this repo. So I've updated `enabled` to follow that pattern in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5378/commits/d607e01ad92ece915c9e8e92678e89de79e4796e

wjrosa (approved):
Looks good to me and works as intended! 👍 We can confirm the missing event arrives later, as we discussed

PR #5384, opened 2026-05-04
Fix subscription change-payment: saved card, wallet branding on My Account, and order note

### LINKED ISSUE #5382: Change-payment-method: misleading 'Credit Card' note and missing token association on non-ECE flows
Follow-up to [#5284](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5284), which addressed the equivalent issues for the new ECE change-payment-method path. Two related gaps remain on the broader change-payment flow.

## 1. Misleading "from X to Credit Card" subscription note

When changing the saved payment method for a subscription, `WC_Subscriptions_Change_Payment_Gateway::update_payment_method()` writes an order note using the WC gateway's `get_method_title()` value — which for the `stripe` gateway is `"Credit Card"`, regardless of the underlying card brand or wallet (Apple Pay / Google Pay / Link).

Visible result, observed by @daledupreez during PR #5284 testing:

> The subscription admin shows the Google Pay method, but it also shows some notes saying that the payment method was changed from Google Pay to Credit Card.

Our `update_subscription_payment_method_title()` hook (in `WC_Stripe_Express_Checkout_Element`) corrects the displayed title afterward, but the misleading note has already been written.

**Possible mitigation** (cheapest, no WCS changes): after applying the express title in `update_subscription_payment_method_title()` and `maybe_apply_express_title_after_confirmed_intent()`, append a corrective `add_order_note( sprintf( __( 'Payment method updated to %s.' ), \$express_label ) )` so the truth lives next to the misleading WCS line. A fuller fix would need WCS to expose a filter on the note content.

## 2. Token not associated with the subscription on non-ECE change-payment flows

The deferred-intent change-payment flow (`WC_Stripe_UPE_Payment_Gateway::process_change_subscription_payment_with_deferred_intent()`) updates `_stripe_source_id` and `_payment_method` on the subscription, but never calls `\$subscription->add_payment_token( \$token )`. PR #5284 patched this for ECE submissions only (gated on `\$_POST['express_checkout_type']` being present). The same gap likely affects \"enter a new card on the change-payment-method form\" submissions outside ECE.

Repro hypothesis (needs verification):

1. As a shopper, change a subscription's payment method by entering a brand-new card (no saved-token selection) on the change-payment-method form.
2. Observe that admin shows the new card's brand/last4 (because `set_payment_method_title_for_order()` runs), but **My Account → Subscriptions** keeps rendering the previously-saved token until you also use the new card on a regular checkout.

Root cause:

- `handle_saving_payment_method()` creates the new `WC_Payment_Token` against the **user**, then calls `maybe_update_source_on_subscription_order()`. That helper inspects `wcs_order_contains_subscription( \$order_id )` / `wcs_order_contains_renewal( \$order_id )` — both return `false` when `\$order` is the subscription itself (which is what the change-payment flow passes). So no subscription-level update happens there.
- `WC_Subscriptions_Change_Payment_Gateway::update_payment_method()` writes `_payment_method` and the post-meta entries supplied to it, but doesn't touch `_payment_tokens`.
- `WC_Order::get_payment_tokens()` reads `_payment_tokens` post meta, so My Account keeps showing the stale token.

**Suggested approach**: lift the per-ECE call to `WC_Stripe_Express_Checkout_Helper::replace_subscription_payment_token()` introduced in #5284 into a path the non-ECE deferred-intent change-payment branch also exercises (or generalize the helper out of the ECE class). Add PHPUnit coverage exercising both new-card-via-form and saved-token-selection paths so we don't regress them.

## Out of scope here

- Cascade-checkbox suppression for ECE — already handled in #5284.
- Title-override correctness for ECE on the change-payment page — already handled in #5284.

cc @daledupreez

wjrosa:
Fixes #5382

#5284 fixed the saved-card list and displayed title for the **express checkout** change-payment flow. Three related gaps remain on the broader change-payment surface; this PR closes them.

1. **Token association on non-ECE paths.** `process_change_subscription_payment_with_deferred_intent` updated `_stripe_source_id`/`_payment_method` but never rewrote `_payment_tokens`, so `get_payment_tokens()` kept returning the old token. Reuse the #5284 helper for the no-3DS branch and for `WC_Stripe_Intent_Controller::confirm_change_payment_from_setup_intent_ajax` (the classic 3DS/SCA path).
2. **Misleading order note.** WCS writes "Payment method changed from X to Credit Card", ignoring the wallet used. `apply_express_checkout_title_to_order` now appends a corrective "Payment method updated to <wallet> (Stripe)" note; both ECE handlers inherit it.
3. **Wallet brand on the My Account row.** `get_payment_method_to_display_for_payment_method` ignored `card->wallet->type`, so it still read "Via Visa card …" after a wallet change. It now surfaces the wallet brand (e.g. "Via Google Pay (Visa) ending in 4242"); `link` and manual cards stay bare. The saved-token `wallet_type` stays pinned per #5477 — the data provider asserts it's preserved across all nine card/Apple Pay/Google Pay transitions.

## Testing instructions

Prerequisite: WooCommerce Subscriptions + Stripe, with an active subscription originally paid by a saved Visa card. For each path, **My Account → Subscriptions →** select the subscription **→ Change Payment**, submit, then verify the new card replaces the old one on both the My Account row and the subscription admin.

- **Non-ECE, no 3DS** — enter `4242 4242 4242 4242`.
- **Non-ECE, 3DS via AJAX** — enter `4000 0027 6000 3184`, complete the SCA challenge.
- **ECE 3DS via redirect** (regression, #5284) — change via an Apple Pay/Google Pay button backed by a 3DS card; complete SCA. Verify the title (e.g. "Apple Pay (Stripe)"), the wallet brand on the My Account row, and that the admin notes contain **both** the WCS line and the new "Payment method updated to Apple Pay (Stripe)." line.
- **ECE, no 3DS** (regression, #5284) — same expectations without SCA.
- **Existing saved token** — select an existing saved token on the form (not a new card); verify it updates with no behaviour change vs. before this PR.

---

### Changelog entry

> Fix - When changing a subscription's payment method, reflect the new card and its Apple Pay/Google Pay branding on My Account and clarify the admin order note

**Post merge**


daledupreez:
@wjrosa, when you get a chance, could you resolve the conflicts in this PR? I am not sure what the right resolution should be for these two areas of the code.

wjrosa:
Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5384#issuecomment-4478452649

daledupreez (commented):
The two express checkout tests were not working for me -- in both cases the underlying Google Pay card was shown in the shopper-facing UI without any reference to Google Pay. The Google Pay details did show up in the subscription details, however, so I think this was limited to the Payment Methods and Subscription pages in My Account.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -1394,6 +1394,10 @@ public function confirm_change_payment_from_setup_intent_ajax() {
 			$gateway->set_customer_id_for_subscription( $subscription, $customer->get_id() );
 			$gateway->set_payment_method_id_for_subscription( $subscription, $token->get_token() );
 
+			// Mirror the no-3DS change-payment branch: associate the new WC token with
+			// the subscription so My Account → Subscriptions reflects the new card.
+			WC_Stripe_Express_Checkout_Helper::replace_subscription_payment_token( $subscription, $token->get_token() );
```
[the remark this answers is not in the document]
wjrosa: Done in cbb432ab7. Throws `subscription_token_replace_failed` on the false branch, lands in the existing catch which surfaces the failure to the user and logs the subscription id. Reused the `token_creation_failed` wording from a few lines up since the user-facing failure mode is the same ("the change couldn't be completed").

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -331,10 +331,7 @@ public function process_change_subscription_payment_with_deferred_intent( $subsc
 				);
 
 				// Link the new token to the subscription so My Account renders it.
-				// Scoped to ECE here; tracked for the broader paths in #5382.
-				if ( $is_express_checkout_submission ) {
-					WC_Stripe_Express_Checkout_Helper::replace_subscription_payment_token( $subscription, $payment_method_id );
-				}
+				WC_Stripe_Express_Checkout_Helper::replace_subscription_payment_token( $subscription, $payment_method_id );
```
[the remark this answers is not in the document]
wjrosa: Done in cbb432ab7. Same pattern — throws so the existing catch at line 391 renders the failure notice and logs the subscription id. Used the cleaner "We can't process your payment method change at this time" wording rather than the suggested message.

wjrosa:
Thanks for testing this, @daledupreez. That render path — saved card showing instead of "Google Pay" / "Apple Pay" on My Account → Payment Methods + Subscription pages — was a separate gap that #5439 covered. You tested at 562e9535a, just before I merged develop into this branch (614587b13), so #5439's wallet-branding fix wasn't in the build you exercised. The current head (cbb432ab7) includes it, plus the token-replace failure handling from CodeRabbit's review.

If you have a moment, re-running the two ECE testing steps should now surface the wallet branding correctly.

daledupreez:
The standard 3DS card flow is failing for me with the following log message:
> Error processing change subscription payment method with deferred intent for subscription: 62

That has the following `error_message` value: `subscription_token_replace_failed`, which is being logged from `includes/compat/trait-wc-stripe-subscriptions.php` line 398.

daledupreez:
Google Pay with a 3DS card is triggering the same error as the standard 3DS card.

Also, Google Pay with a non-3DS card is working, but the payment details in the subscription view are still showing the underlying card and not "Google Pay".

daledupreez (changes requested):

daledupreez:
For reference, I tested with a brand new site that has Optimized Checkout Suite and Adaptive Pricing enabled.

wjrosa:
@daledupreez Thanks — both reproduced and root-caused:

**1. 3DS `subscription_token_replace_failed`:** the token re-link ran in the deferred-intent handler *before* 3DS confirmation, where the confirmed token does not exist yet — so it threw and aborted the change before the modal even appeared. Moved it to the post-confirmation path and made it best-effort: a saved-card *display* refresh should not fail a change the shopper has already authenticated. Same path covers Google Pay 3DS.

**2. Google Pay non-3DS shows the underlying card:** a wallet over an already-saved card reuses that token by fingerprint, and we were not refreshing `wallet_type`, so My Account kept showing the bare card. Now refreshed on reuse (last-write-wins, so a manually-saved card will pick up wallet branding once reused via a wallet — flagging in case that is not what we want).

Pushed in 7d39234 (3DS) and 0498af6 (wallet_type), with regression tests for both. Ready to re-test on the OCS + Adaptive Pricing setup you used.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -322,19 +322,18 @@ public function process_change_subscription_payment_with_deferred_intent( $subsc
 			// Create a setup intent, or update an existing one associated with the order.
 			$payment_intent = $this->process_setup_intent_for_order( $subscription, $payment_information );
 
-			// Handle saving the payment method in the store.
-			if ( $payment_information['save_payment_method_to_store'] && $upe_payment_method && $upe_payment_method->get_id() === $upe_payment_method->get_retrievable_type() ) {
+			// Handle saving the payment method in the store. The token is linked to
+			// the subscription further down, in the no-confirmation branch only:
+			// when the intent still requires action (3DS), the confirmed token does
+			// not exist yet and is associated post-confirmation in
+			// WC_Stripe_Intent_Controller::confirm_change_payment_from_setup_intent_ajax().
```
daledupreez: I think this long comment is unnecessary, as a lot of the context is relevant to this branch's evolution.
```suggestion
			// Handle saving the payment method in the store.
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -370,6 +369,13 @@ public function process_change_subscription_payment_with_deferred_intent( $subsc
 				$this->set_payment_method_id_for_subscription( $subscription, $payment_method_id );
 				$this->set_customer_id_for_subscription( $subscription, $payment_information['customer'] );
 
+				// Link the saved token to the subscription so My Account renders the
+				// new card. Best-effort: this only refreshes the saved-card display,
+				// so a miss must not fail the payment-method change itself.
```
daledupreez: I think we can simplify the comment here as well.
```suggestion
				// If we saved a new token, link the saved token to the subscription for display purposes.
				// Intentionally ignore any failures, as the display update doesn't affect renewals.
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -370,6 +369,13 @@ public function process_change_subscription_payment_with_deferred_intent( $subsc
...
+				if ( $saved_payment_method_to_store && ! WC_Stripe_Express_Checkout_Helper::replace_subscription_payment_token( $subscription, $payment_method_id ) ) {
+					WC_Stripe_Logger::warning( 'Could not re-associate the saved token after change-payment for subscription: ' . $subscription_id );
```
daledupreez: Should this not be an error? These errors will be invisible on sites that don't have debugging enabled.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3630,9 +3630,14 @@ public function handle_saving_payment_method( WC_Order $order, stdClass $payment
...
 
 		if ( $found_token ) {
-			// `wallet_type` is intentionally not refreshed — it reflects how the
-			// token was created and must not flip when the same card is reused
-			// through a wallet sheet.
+			// Refresh `wallet_type` to match the latest tokenization so the saved-card
+			// display reflects the wallet actually used. Without this, paying with
+			// Apple/Google Pay over a card already saved manually reuses the existing
+			// token by fingerprint and keeps showing the bare card with no wallet
+			// branding (e.g. on My Account → Subscriptions after a change-payment).
```
daledupreez: This description is a bit confusing to me. Can we simplify it a bit? And please correct me if I misunderstood!
```suggestion
			// Update the saved token's wallet_type to match the payment method.
			// This will update the underlying token to the most recent usage, which may add or remove a wallet.
```

includes/class-wc-stripe-intent-controller.php
```diff
@@ -1394,6 +1394,14 @@ public function confirm_change_payment_from_setup_intent_ajax() {
 			$gateway->set_customer_id_for_subscription( $subscription, $customer->get_id() );
 			$gateway->set_payment_method_id_for_subscription( $subscription, $token->get_token() );
 
+			// Mirror the no-3DS change-payment branch: associate the new WC token with
+			// the subscription so My Account → Subscriptions reflects the new card.
+			// Best-effort: this only refreshes the saved-card display, so a miss must
+			// not fail the payment-method change the shopper has already authenticated.
```
daledupreez: Again, I think this context is too specific and we should simplify.
```suggestion
			// Update the saved token details to capture the current details for display purposes.
			// Errors are intentionally ignored as the changes are cosmetic and don't impact renewals.
```

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-element-test.php
```diff
@@ -409,6 +409,10 @@ function ( $id ) use ( $subscription ) {
 
 		$this->assertSame( $expected, $subscription->get_payment_method_title() );
 
+		// A clarifying note must be appended so the WCS-written "to Credit Card"
+		// log entry isn't the only record of the change.
```
daledupreez: I think this comment is unnecessary.
```suggestion
```

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-element-test.php
```diff
@@ -448,17 +452,28 @@ public function provide_test_update_subscription_payment_method_title() {
...
 		$subscription->set_payment_method_title( 'Credit Card (Stripe)' );
 		$subscription->update_meta_data( '_wc_stripe_express_checkout_type', $checkout_type );
+		$subscription->update_meta_data( '_wc_stripe_express_checkout_payment_method_id', 'pm_post3ds_card' );
 		$subscription->save();
 
 		$this->element->maybe_apply_express_title_after_confirmed_intent( $subscription );
 
 		$this->assertSame( $expected, $subscription->get_payment_method_title() );
-		// Meta should be cleaned up after applying.
+		// Both one-shot markers should be cleaned up after applying.
```
daledupreez: I think we can use a clearer term.
```suggestion
		// Both temporary meta fields should be cleaned up after applying.
```

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -1575,6 +1575,63 @@ public function provider_handle_saving_payment_method_respects_per_method_toggle
 		];
 	}
 
+	/**
+	 * handle_saving_payment_method() must refresh a reused token's wallet_type so the
+	 * saved-card display reflects the wallet just used. Paying with Apple/Google Pay
+	 * over a card the shopper already saved manually reuses the existing token by
+	 * fingerprint; without refreshing wallet_type the saved-card row (e.g. My Account
+	 * → Subscriptions after a change-payment) keeps showing the bare card.
```
daledupreez: This description feels unnecessary.
```suggestion
```

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -1575,6 +1575,63 @@ public function provider_handle_saving_payment_method_respects_per_method_toggle
...
+	 */
+	public function test_handle_saving_payment_method_refreshes_wallet_type_on_reused_token() {
```
daledupreez: I think we should use a data provider and test multiple cases:
 * Card => Apple Pay
 * Card => Google Pay
 * Card => Card
 * Apple Pay => Card
 * Apple Pay => Google Pay
 * Apple Pay => Apple Pay
 * Google Pay => Apple Pay
 * Google Pay => Card
 * Google Pay => Google Pay

The one I am most concerned about is making sure we add checks for going from a wallet back to a plain card.

tests/phpunit/class-wc-stripe-intent-controller-test.php
```diff
@@ -736,4 +736,58 @@ public function test_confirm_change_payment_allows_owner(): void {
 		WC_Subscriptions::set_wcs_get_subscription( null );
 		Ajax_Test_Helper::remove_hooks();
 	}
+
+	/**
+	 * Regression: a failed saved-token re-association only affects the My Account
+	 * saved-card display, so it must not fail a change-payment the shopper has
+	 * already authenticated via 3DS. Previously this threw
+	 * `subscription_token_replace_failed` and aborted the whole flow.
+	 */
```
daledupreez: I think we can drop the description here, as the regression was really on this branch, and not on develop.
```suggestion
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3630,9 +3630,14 @@ public function handle_saving_payment_method( WC_Order $order, stdClass $payment
...
+			if ( $found_token instanceof WC_Stripe_Payment_Token_CC ) {
+				$found_token->set_wallet_type( (string) ( $payment_method_object->card->wallet->type ?? '' ) );
+			}
```
diegocurbelo: I'm confused by this change... it's not explicitly mentioned in the PR description and seems to revert the fix merged in #5477 a couple days ago

daledupreez (commented):
I have some notes up from inspection last week, but will wait for clarity from Diego's review comments before (re)testing.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Intentional, not an accidental revert — it's the deliberate counterpart to #5477. That PR pinned `wallet_type` to stop a saved card flipping to a wallet brand (card→wallet); here the saved-card display needs to follow the latest tokenization after a change-payment, so it's now last-write-wins: a wallet payment adds the badge, reusing the card directly clears it. I'll call this out in the PR description. The new data provider in c2aabaef7 locks down all nine transitions, including wallet→plain card. Happy to fall back to #5477's pinning if you'd rather keep that.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Good catch — `warning()` is gated by `can_log()`, so it's invisible unless logging is enabled, while `error()` always writes. Switched this path and the 3DS one to `error()` in c2aabaef7.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Done in c2aabaef7 — converted to a data provider covering all nine card/Apple Pay/Google Pay transitions, with explicit rows for wallet→plain card (badge clears), since that was your main concern.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Re-added a one-line contract docblock here when moving to the data provider — repo convention is a short PHPDoc on each test describing what it guards. Kept it terse.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
diegocurbelo: I don't understand the deliberate counterpart... this changes `handle_saving_payment_method`, which is the general flow used for every UPE payment, not just change-payment, so it's literally re-introducing the exact same bug #5477 fixed; it can be replicated using the same repro steps from that PR:
- In a Stripe test account, save card 4000056655665556 in My Account → Payment Methods.
- Confirm the saved card shows as a plain Visa (no wallet branding).
    Screenshot 2026-06-02 at 00 30 32
- Purchase a subscription via Google Pay, selecting the same card from the Google Pay sheet.
- Reload My Account → Payment Methods.
- The saved card was renamed to "Google Pay (Visa)"
    Screenshot 2026-06-02 at 00 27 56

--

Also, the specific case for change-payment mentioned in the code comment is not working for me:
> Without this, paying with Apple/Google Pay over a card already saved manually reuses the existing token by fingerprint and keeps showing the bare card with no wallet branding (e.g. on My Account → Subscriptions after a change-payment).

https://github.com/user-attachments/assets/a43b9631-af42-48a9-ba5e-dc9c2037a062


--

Maybe it's a problem with my environment, are those 2 cases working for you @wjrosa?

wjrosa: Thanks for your review, and the careful repro! I was able to reproduce both locally today. I was trying to delegate the task of solving the issues on PR almost entirely to AI, but I gave up 😅.

- `handle_saving_payment_method` change: I reverted it in my last commit to keep the original behavior in place.

- The change-payment case: I was able to confirm it. And it's why the token had the wrong label. The card case now shows the wallet brand (e.g. "Via Google Pay (Visa) ending in 4242") via the existing `EXPRESS_METHODS_LABELS` map; `link` and manual cards stay the same, matching the token's own `get_wallet_brand_label()`.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -1576,21 +1570,24 @@ public function provider_handle_saving_payment_method_respects_per_method_toggle
...
+	 * #5477), so a manually-saved card can't flip to a wallet brand.
+	 *
+	 * @dataProvider provider_handle_saving_payment_method_preserves_wallet_type
+	 *
+	 * @param string $initial_wallet_type wallet_type already stored on the saved token.
+	 * @param string $new_wallet_type     wallet_type on the incoming payment method ('' for a plain card).
+	 * @param string $expected            wallet_type expected on the token after reuse (always the initial value).
 	 */
-	public function test_handle_saving_payment_method_refreshes_wallet_type_on_reused_token() {
+	public function test_handle_saving_payment_method_preserves_wallet_type_on_reused_token( $initial_wallet_type, $new_wallet_type, $expected ) {
```
daledupreez: Nit: It would be slightly more explicit if we used `$expected_wallet_type` for the expected value.

daledupreez (commented):
Overall, I think this is working as expected, but there is still one edge case where I am not clear on what the expected behaviour is or should be.

The specific case is as follows:
* Change the payment method to a plain, non-3DS card (e.g. the 4242 Visa)
* Change the payment method to Google Pay using the same underlying card
  - At this stage, the Subscription view shows the Google Pay details, while My Account -> Payment methods shows the underlying card still
* Change the payment method again to some other payment method
* Change the payment method again - you should see the Visa ending in 4242 as an option (with no wallet indicator)
* After changing the payment method, the subscription screen now shows the wallet details, not the underlying card

I think this ends up being somewhat separate from the changes in this PR, but it's an edge case that I want to flag.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Done in dfd5cb576 — renamed to `$expected_wallet_type`.

wjrosa:
@daledupreez thanks for the review and for flagging this issue! I opened a new PR to address this specific case in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5544

diegocurbelo (approved):
Thanks for the fixes @wjrosa! 

LGTM 👍🏼

daledupreez:
I am happy for this to get merged as things stand.

PR #5386, opened 2026-05-05
[RSM - What's new popup] Show changelog link in 'Updated!' message after manual plugin update

wjrosa:
See: phcsdm-2NN-p2
Fixes [STRIPE-1116](https://linear.app/a8c/issue/STRIPE-1116)

When a user manually updates the WooCommerce Stripe plugin from `wp-admin/plugins.php`, the existing **Updated!** notice now gets a "See what's new" link appended to it. The link reuses WordPress' built-in plugin information modal — same logic as the standard "View details" link — and opens it directly on the **Changelog** tab via `&section=changelog`. This gives users immediate access to the release notes for the version they just installed without leaving the Plugins page.

Implementation notes:

- `WC_Stripe_Plugins_Page_Controller` now ensures `thickbox` and `plugin-install` are enqueued on `plugins.php`, and localizes `plugin_slug` + `view_changelog_url` into `wcStripePluginsPageParams`.
- A new `client/entrypoints/plugins-page/append-changelog-link.js` listens for the WP `wp-plugin-update-success` jQuery event. If `response.slug` matches our plugin, it appends the link to the `Updated!` `<p>` inside the matching `tr.plugin-update-tr`. The handler is idempotent and a no-op when jQuery / params are missing.
- Coverage: 6 Jest specs (slug match, ignore other plugins, no double-insert, missing-jQuery / missing-params no-op, listener cleanup) and 2 PHPUnit specs (thickbox enqueued, localized params contract).

Preview:

Screenshot 2026-05-06 at 15 39 44

Screenshot 2026-05-05 at 10 13 35

### Quick UI check (browser console)

The fastest way to verify the link insertion and modal without doing a real plugin update. On `wp-admin/plugins.php`, open DevTools and paste:

```js
(() => {
    const slug = wcStripePluginsPageParams.plugin_slug;

    // Insert a fake "Updated!" row right after the Stripe plugin row.
    const pluginRow = document.querySelector(`tr[data-slug="${slug}"]`);
    if (!pluginRow) { console.warn('Stripe plugin row not found'); return; }

    // Mirror what `wp.updates.updatePluginSuccess` does to the plugin row,
    // otherwise the row keeps its `.active` bottom box-shadow and the preview
    // shows a stray 1px line above the green notice that doesn't appear in
    // real updates.
    pluginRow.classList.remove('update');
    pluginRow.classList.add('updated');

    const fakeRow = document.createElement('tr');
    fakeRow.className = 'plugin-update-tr active updated';
    fakeRow.dataset.slug = slug;
    fakeRow.innerHTML = `
        <td colspan="4" class="plugin-update colspanchange">
            <div class="update-message notice inline updated-message notice-success notice-alt">
                <p>Updated!</p>
            </div>
        </td>`;
    pluginRow.after(fakeRow);

    // Fire the same event WP fires after a successful update.
    jQuery(document).trigger('wp-plugin-update-success', [{ slug }]);
})();
```

Expected:

- The fake row reads `Updated! See what's new`.
- Clicking the link opens the WordPress plugin information modal directly on the **Changelog** tab (no page navigation).

### End-to-end (real update flow)

1. Make sure an update is available for this plugin on the Plugins page. The simplest way is to lower the `Version:` header in `woocommerce-gateway-stripe.php` (e.g. set it to `8.0.0`), then hit **Check again** on the Updates page so WordPress detects a newer version on wp.org.
2. Go to `wp-admin/plugins.php`.
3. In the WooCommerce Stripe Gateway row, click the inline **update now** link.
4. Wait for the row's status to flip to **Updated!**.
5. Verify the message now reads: `Updated! See what's new`.
6. Click the appended link — the modal should open on the **Changelog** tab.
7. Trigger an update for any other plugin and verify the link is **not** appended to its `Updated!` notice.

### Automated checks

- `npm run test:js -- --testPathPattern=plugins-page` (11 specs pass)
- `npm run test:php -- --filter WC_Stripe_Plugins_Page_Controller_Test` (5 specs pass)

---

-   [x] Tested on mobile (or does not apply) — admin-only feature, not surfaced on storefront mobile

### Changelog entry

Added under `10.7.0` in both `changelog.txt` and `readme.txt`:

> Add - Append a "what's new" changelog link to the Updated! message after manually updating the plugin from the Plugins page

**Post merge**


client/entrypoints/plugins-page/append-changelog-link.js
```diff
@@ -0,0 +1,73 @@
...
+ * opened directly on the changelog tab.
+ *
+ * @param {{plugin_slug: string, view_changelog_url: string}} params
+ * @return {() => void} Cleanup function that detaches the listener.
+ */
+export const initAppendChangelogLink = ( params ) => {
+	const { jQuery } = window;
+	if ( ! jQuery || ! params || ! params.plugin_slug ) {
+		return () => {};
+	}
```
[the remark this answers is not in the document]
wjrosa: Good catch — the URL crosses the PHP→JS boundary and the existing guard already validates `plugin_slug`, so extending it for `view_changelog_url` keeps the contract symmetric and avoids the `href="undefined"` failure mode if the payload is ever malformed. Added the clause and a regression test in 45692ddaa.

includes/admin/class-wc-stripe-plugins-page-controller.php
```diff
@@ -10,6 +10,12 @@
  */
 class WC_Stripe_Plugins_Page_Controller {
 
+	/**
+	 * Slug used to identify this plugin in WordPress update events
+	 * and in plugin information modal URLs.
+	 */
+	const PLUGIN_SLUG = 'woocommerce-gateway-stripe';
```
malithsen: Just double-checking if we can use an existing definition. Is this not already set somewhere?

wjrosa: Good call — switched to `basename( WC_STRIPE_PLUGIN_PATH )`. Derives the directory name from an existing constant and avoids introducing `PLUGIN_SLUG`.

The other repeated literals (textdomain, Action Scheduler group, `WC_Stripe_Logger::WC_LOG_FILENAME`) all carry different semantics, so reusing one of them for "WP plugin slug" felt wrong. `dirname( plugin_basename( WC_STRIPE_MAIN_FILE ) )` was the first thing I tried, but it's fragile when `WP_PLUGIN_DIR` doesn't match the path WP loaded the plugin from (broke the localization test in our container).

malithsen (changes requested):
The code looks good

Following Elizabeth's suggestion, I think we should remove the extra ~~margin~~ border in the notice

Image

Let's also update the test instructions so the snippet is up to date with the new changes.

wjrosa:
@malithsen 

> Following Elizabeth's suggestion, I think we should remove the extra margin in the notice

As we discussed, I think you are mentioning the border. This PR does not touch any styles; we are using the default styles from WordPress. So, that's not something we should change here.

> Let's also update the test instructions so the snippet is up to date with the new changes.

Done!

wjrosa:
@malithsen That line is the WC Stripe plugin row's `.active` bottom inset shadow (`list-tables.css:1280`), not a notice border. WP removes it on real updates by adding `.updated` to the row (which trips `.plugins .updated td { box-shadow: none }`), so it doesn't appear in the actual flow — the testing snippet just wasn't applying that class. Updated the snippet in the PR description; rerunning it shows the line gone, matching a real update.

malithsen (approved):
Tested by using a build with a back-dated version on a JN site. Looks great!

Image

includes/admin/class-wc-stripe-plugins-page-controller.php
malithsen: Now that I think about this more, I think hardcoding the slug is probably the right approach. WP.org listing is independent of the plugin name/path and doesn't need to be the same.

daledupreez (commented):
@wjrosa, just to triple-check: is the only way to trigger this additional notice by auto-updating the plugin? I just want to make 1000% sure that the link doesn't get added if someone updates to an older version.

wjrosa:
@daledupreez Correct. The link only attaches to `wp-plugin-update-success`. Manual zip upload, WP Rollback, WP-CLI and FTP don't fire that event, so they can't attach the link.

PR #5387, opened 2026-05-05
Improve UX for OCS and Adaptive Pricing settings

### LINKED ISSUE #5206: Optimized Checkout Suite and Adaptive Pricing settings are not discoverable
### Description

The UX for our current _Settings_ tab is sub-optimal when settings sync is disabled, or when Optimized Checkout is disabled, as we don't clearly communicate that a feature is disabled in the current state, so merchants can't tell that they are missing out on features.

#### State when settings sync is disabled

1. Set up a store with a connection to Stripe
2. Use WP CLI to disable settings sync: `wp option patch update woocommerce_stripe_settings pmc_enabled no`
3. Navigate to _Payments_ -> _Stripe_, click on the "Manage" button, and then the "Settings" tab
4. Scroll down to the "Advanced Settings" section, and notice that the only setting is for debugging

Image

5. Make sure to re-enable settings sync: `wp option patch update woocommerce_stripe_settings pmc_enabled yes`

#### State when Optimized Checkout Suite is not enabled

1. Set up a store with a connection to Stripe
2. Add the following filter to enable Adaptive Pricing:
```php
```
4. Navigate to _Payments_ -> _Stripe_, click on the "Manage" button, and then the "Settings" tab
5. Scroll down to the "Advanced Settings" section
6. Ensure that "Dynamically display the most relevant payment methods you've enabled" checkbox is disabled
7. Save the changes
8. Reload the page, and scroll down to the "Advanced Settings" section
9. Note that the Adaptive Pricing toggle is hidden

Image

There is also a related bug, where enabling Optimized Checkout Suite doesn't reveal the Adaptive Pricing checkbox -- the setting needs to be saved and the page reloaded for the second toggle to become visible (or hidden).

### Expected behavior

Even when the feature is not currently available to me, it would be good to be able to (1) discover that new features exist, and (2) learn more about what the feature is.

### LINKED ISSUE #5355: Adaptive Pricing: Improve checks for store and account settlement currency mismatches
Adaptive Pricing requires the displayed prices to use one of the settlement currencies for your account - [Adaptive Pricing Restrictions](https://docs.stripe.com/payments/currencies/localize-prices/adaptive-pricing?payment-ui=embedded-components#restrictions). We should improve the UX when we can detect that Adaptive Pricing won't be available to the merchant.

daledupreez:
Fixes STRIPE-1109
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5355
Fixes STRIPE-1017
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5206

This PR improves the settings UI for Optimized Checkout Suite (OCS) and Adaptive Pricing in a number of ways:
 * We always show the OCS and Adaptive Pricing settings, even when they are not available.
   - When OCS is not available, we show a warning notice _and_ disable the OCS checkbox
   - When Adaptive Pricing is not available, we disable the checkbox and add inline details to the help text in **bold**, which captures the following situations:
      * OCS is not available or not enabled
      * Adaptive Pricing is not available in the merchant's country (at present this is limited to accounts based in India)
      * We can't identify which settlement currencies are available in the merchant's account (this is only applied in live mode)
      * The merchant's settlement currencies don't include the store currency (only applied in live mode)
* With both checkboxes always visible, we now show the OCS layout fields immediately after the OCS checkbox and then the Adaptive Pricing checkbox, and we also remove the indent for the Adaptive Pricing checkbox
* The PR also removes a trailing `.` from the AP description to match other items, and tweaks the descriptive text for Adaptive Pricing to make it clearer what the feature does.
* The run-time checks for enabling Adaptive Pricing take into account the merchant's settlement currency when in live mode -- this ensures that we don't create a checkout session and try to load Adaptive Pricing when the merchant's account and store configuration don't allow for it.

Overall, these changes should make it much easier to understand what the state of the settings are, and what actions the merchant can take to enable the features, especially when Sync is disabled. (Though we need to work on that reconnect aspect as well.)

As things stand, the one gotcha with the new UX is that when the options are enabled, but one or both controls are disabled (e.g. Both settings are enabled, and then OCS is disabled), the disabled state is less visibly obvious. For this first pass, I think it's better to keep the option state visible, as I think it would be more confusing if we showed the option as disabled -- that would then mean that enabling OCS (for example), could suddenly result in Adaptive Pricing being toggled on.

The other concern here is that the notice is somewhat vague about the actions a merchant can take when Sync is disabled. @diegocurbelo was exploring a PR that could address that case.

#### Default state: Sync enabled and OCS disabled

Screenshot 2026-05-06 at 09 46 35

#### Sync disabled, so OCS not available

Screenshot 2026-05-06 at 09 49 15

#### OCS enabled

Screenshot 2026-05-06 at 09 46 56

## Testing instructions

* Set up a test site using the code from this branch.
* Connect to Stripe using a test account
* Navigate to the Stripe settings and confirm that Sync is enabled
* Confirm that both OCS and AP settings are visible
* If both are not enabled, enable them both
* Verify that no warnings are shown, and the OCS layout control is shown after the OCS checkbox
* Disable OCS
* Verify that the AP control is disabled and shows a note about OCS needing to be on
* Re-enable OCS
* Verify that the AP control is re-enabled and no longer shows the note
* Manually disable Sync: `wp option patch update woocommerce_stripe_settings pmc_enabled no`
* Reload the page
* Verify that the OCS warning is shown, and both the OCS and AP checkboxes are disabled - they may keep their previous state
* Re-enable Sync (or let it be auto-reenabled): `wp option patch delete woocommerce_stripe_settings pmc_enabled`
* Reload the page and re-enable OCS
* Change the WooCommerce store currency and reload the page
* Confirm that no warnings are shown
* Connect in live mode
* If needed, re-enable OCS and Adaptive Pricing
* If needed, update your store currency to be in a currency that you can not settle in, and then reload the page
* The Adaptive Pricing checkbox should be disabled, and you should see a warning about a currency mismatch
* Now add a simple product to your cart
* Open up the network tab of your developer tools
* Navigate to block checkout in one tab and classic checkout in a second tab
* Verify that there was no API call to the `wc_stripe_create_checkout_session` AJAX API in either tab
* Switch the store currency back to a supported settlement currency for your account
* Reload both block and classic checkout, and verify that you see `wc_stripe_create_checkout_session` AJAX calls in both cases

_Note: We use the AJAX calls as the failure mode for Adaptive Pricing in the currency mismatch case results does create a checkout session, but no currency selector is displayed. We want to make sure we are not trying to enable AP in this case._

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-helper.php
```diff
@@ -1116,8 +1116,45 @@ public static function has_cart_or_checkout_on_current_page() {
...
+		// Check that the store currency can be used for settlement.
+		$stripe_settlement_currencies = $stripe_account->get_supported_store_currencies();
+		if ( [] === $stripe_settlement_currencies ) {
+			return 'no-settlement-currencies';
+		}
+
+		$store_currency = get_woocommerce_currency();
+		if ( ! in_array( strtolower( $store_currency ), $stripe_settlement_currencies, true ) ) {
+			return 'store-currency-not-settlement-currency';
+		}
```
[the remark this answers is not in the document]
daledupreez: Normalised in a6a85168dc31760ebcf7c93effecb123d44873d7.

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -216,7 +216,14 @@ public function admin_scripts( $hook_suffix ) {
...
 
-		$is_ap_available_for_account = WC_Stripe_Helper::is_adaptive_pricing_available_for_account();
+		$is_checkout_sessions_available       = false;
+		$adaptive_pricing_availability_reason = 'disabled';
+		if ( WC_Stripe_Feature_Flags::is_checkout_sessions_available() ) {
+			$adaptive_pricing_availability_reason = WC_Stripe_Helper::get_adaptive_pricing_account_availability_reason();
+			if ( null === $adaptive_pricing_availability_reason ) {
+				$is_checkout_sessions_available = true;
+			}
+		}
```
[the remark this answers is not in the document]
daledupreez: Updated in 00e477a87606651ad63cb2de058783016e337f29.

client/settings/advanced-settings-section/optimized-checkout-feature.js
```diff
@@ -131,6 +222,21 @@ const OptimizedCheckoutFeature = () => {
...
+					adaptivePricingUnavailableReason
+				}
+				label={ __(
+					'Let customers pay in their local currency with Adaptive Pricing',
+					'woocommerce-gateway-stripe'
+				) }
+				help={ adaptivePricingHelp }
+				checked={ isAdaptivePricingEnabled }
+				onChange={ setIsAdaptivePricingEnabled }
+			/>
```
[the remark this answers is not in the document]
daledupreez: Handled via a `!== null` check in 84b100391ade4a320b5404f72b8abdb2e436b74e.

client/settings/advanced-settings-section/__tests__/index.test.js
```diff
@@ -65,14 +65,19 @@ describe( 'AdvancedSettings', () => {
...
 			screen.queryByText(
 				'Enable Optimized Checkout Suite (recommended)'
 			)
-		).not.toBeInTheDocument();
+		).toBeInTheDocument();
+		expect(
+			screen.queryAllByText(
+				/Optimized Checkout Suite is not currently available/
+			)
+		).not.toHaveLength( 0 );
```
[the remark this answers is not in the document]
daledupreez: This structure is needed due to `getByText()` failing -- the underlying notice component uses outputs the same text in two separate notes for a11y purposes. No plans to change this, but I've added an inline comment to explain the weird usage in ce63f877eedadc0ed8abc444e5f916ad6612729b.

includes/class-wc-stripe-helper.php
```diff
@@ -1116,8 +1116,45 @@ public static function has_cart_or_checkout_on_current_page() {
...
+
+		return null;
 	}
```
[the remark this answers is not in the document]
daledupreez: Unit tests added in 7a940f552ead1a0df1819ea2767e525ad2a8e61c.

includes/class-wc-stripe-helper.php
```diff
@@ -1120,8 +1120,47 @@ public static function has_cart_or_checkout_on_current_page() {
...
+
+		return false;
+	}
+
+	/**
+	 * Returns the reason why adaptive pricing is not available for the current Stripe account.
+	 *
+	 * @return string|null The reason why adaptive pricing is not available for the current Stripe account, or null if it is available.
+	 */
+	public static function get_adaptive_pricing_account_availability_reason(): ?string {
```
diegocurbelo: I find the name confusing, I'd expect it to be `get_adaptive_pricing_account_unavailability_reason()` to better align with what it returns.

includes/class-wc-stripe-helper.php
```diff
@@ -1120,8 +1120,47 @@ public static function has_cart_or_checkout_on_current_page() {
 	 * @return bool True if the account is not in supported countries.
```
diegocurbelo: This description/return is no longer accurate... we should either defer to the reason docblock or mention the currency check.

Also, the return block is inverted.

includes/class-wc-stripe-helper.php
```diff
@@ -1120,8 +1120,47 @@ public static function has_cart_or_checkout_on_current_page() {
...
 	 */
 	public static function is_adaptive_pricing_available_for_account(): bool {
-		$account_country = WC_Stripe::get_instance()->account->get_account_country();
-		return strtoupper( $account_country ) !== WC_Stripe_Country_Code::INDIA;
+		$reason = self::get_adaptive_pricing_account_availability_reason();
+		if ( null === $reason ) {
+			return true;
+		}
+
+		return false;
```
diegocurbelo: This can be simplified to one line:
```suggestion
		return null === self::get_adaptive_pricing_account_availability_reason();
```

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -234,7 +241,8 @@ public function admin_scripts( $hook_suffix ) {
 			'is_amazon_pay_available'               => WC_Stripe_Feature_Flags::is_amazon_pay_available(),
 			'is_oc_available'                       => WC_Stripe_Feature_Flags::is_oc_available(),
 			'is_oc_enabled'                         => $is_oc_enabled,
-			'is_cs_available'                       => WC_Stripe_Feature_Flags::is_checkout_sessions_available() && $is_ap_available_for_account,
+			'is_cs_available'                       => $is_checkout_sessions_available,
```
diegocurbelo, line 244: `is_cs_available` is no longer used, we should remove it; the only use was in `client/settings/advanced-settings-section/optimized-checkout-feature.js`, but this PR replaces the condition check with `adaptive_pricing_unavailable_reason`

diegocurbelo (approved):
It solution looks good to me, and it tests very well!

I left some comments, but they are non-blocking and can be tackled in a subsequent PR so we don't delay this one

includes/class-wc-stripe-helper.php
daledupreez: Updated in e320a20ebc9d5c7b9ee81daf1115856a4678ea2d, which also moved the link to `get_adaptive_pricing_account_unavailable_reason()`, and added a reference to that method in 13997d4350626bba2f4724385e9fd24c1fe0a371.

includes/class-wc-stripe-helper.php
daledupreez: I updated this in ce233e4500e4e515061279360deffffb92aef6f3 to `get_adaptive_pricing_account_unavailable_reason()` so we have a better name to start with, and don't add an API that then needs to be deprecated and eventually removed.

includes/class-wc-stripe-helper.php
daledupreez: We can tackle this in a follow-up PR.

includes/admin/class-wc-stripe-settings-controller.php
daledupreez: I was erring on the side of being conservative, as Adaptive Pricing and the related flag are technically part of our public API.

includes/class-wc-stripe-helper.php
```diff
@@ -1114,14 +1114,55 @@ public static function has_cart_or_checkout_on_current_page() {
...
+		// Check that the store currency can be used for settlement.
+		$stripe_settlement_currencies = $stripe_account->get_supported_store_currencies();
+		if ( [] === $stripe_settlement_currencies ) {
+			return 'no-settlement-currencies';
+		}
+
+		// Ensure we have lowercase currency codes.
+		$stripe_settlement_currencies = array_map( 'strtolower', $stripe_settlement_currencies );
+		$store_currency_lower         = strtolower( get_woocommerce_currency() );
+		if ( ! in_array( $store_currency_lower, $stripe_settlement_currencies, true ) ) {
```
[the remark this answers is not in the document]
daledupreez: `WC_Stripe_Account->get_supported_store_currencies()` has an explicit return type of `array`, so this is not a guard that we need.

PR #5388, opened 2026-05-05
Fix STRIPE-643: Return error response when PMC update fails in settings

thuautp:
## Summary

When the Stripe API rejects a Payment Method Configuration (PMC) update, the settings endpoint was silently swallowing the error and returning HTTP 200. This caused the frontend to display \"Settings saved.\" even though no payment method changes were actually persisted — customers see the success notice, refresh the page, and find their settings unchanged.

This PR threads a `WP_Error` return value through the three-layer call chain so the REST endpoint returns 502 on failure (the fault is the upstream Stripe call), allowing the frontend's existing error-notice handler to fire correctly. The failure does not block the other settings: they are applied first and the error is returned last, so the merchant can still, for example, disable the gateway while Stripe is rejecting PMC updates.

## Bug

[STRIPE-643: Display error in Payment Methods page when PMC update fails](https://linear.app/a8c/issue/STRIPE-643/display-error-in-payment-methods-page-when-pmc-update-fails)

## Changes

- `includes/class-wc-stripe-payment-method-configurations.php` — `update_payment_method_configuration()` now returns `WP_Error` on failure (no config found, or Stripe API error) and `true` on success, instead of `void`
- `includes/payment-methods/class-wc-stripe-upe-payment-gateway.php` — `update_enabled_payment_methods()` propagates the return value from the PMC layer
- `includes/admin/class-wc-rest-stripe-settings-controller.php` — `update_settings()` checks the result and returns HTTP 502 if the PMC update failed, after all other settings are applied (per review: an early return blocked every other setting, including disabling the gateway); `update_enabled_payment_methods()` returns the result instead of `void`; `get_payment_method_ids_to_enable()` guards against a non-array payload (returns `null` early so callers skip the PMC call cleanly)
- `tests/phpunit/admin/class-wc-rest-stripe-settings-controller-test.php` — regression tests asserting a 502 response when the Stripe API rejects the PMC update (and that the other settings, e.g. disabling the gateway, are still applied), and null return when `enabled_payment_method_ids` is absent

## Testing

- Regression test added: `test_update_settings_returns_502_when_pmc_update_fails` — mocks the Stripe API to return an error and asserts the settings endpoint returns 502 while the other settings still apply
- `test_get_payment_method_ids_to_enable_returns_null_when_not_array` — covers the null-guard path
- QIT automated tests cover the happy path (no regression to the normal save flow)
- To verify manually: on a PMC-enabled site, trigger a Stripe API error on save (e.g. temporarily use invalid credentials) and confirm the admin now shows \"Error saving settings.\" instead of \"Settings saved.\"

## Notes

This is a clean rebase of the work from #5191 onto the current `develop` HEAD — the previous PR accumulated unrelated develop changes during the Bug Blitz session.


thuautp:
@claude addressed — added `WC_Stripe_Logger::error()` at both call sites (`class-wc-stripe.php:1024` and `class-wc-rest-stripe-account-keys-controller.php:331`) in commit 2db6d622. Control flow is unchanged; failures are now logged rather than silently swallowed.

thuautp:
@zhongruige I've created this new PR [per your comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5191#pullrequestreview-4105230596). Let me know if anything still needs to be fixed. Thanks a lot!

zhongruige:
Thanks for getting this in @thuautp! Some feedback--the first one probably is worth a decision on how we best want to approach it, so curious on your thoughts:

1. Partial save on PMC failure (decision needed): Moving the PMC update to the top of `update_settings()` means if it fails, all other settings changes (test mode, manual capture, debug log, etc.) in the same request are silently dropped. Here's where I am curious on your thoughts: we need to decide whether this is acceptable UX, or whether the PMC update should stay at the end so unrelated settings are still persisted even when PMC fails. What do you think would be the most meaningful for the user?
2. Raw Stripe API error message forwarded to the browser: `$response->error->message` (a raw Stripe API string, English, untranslated) is passed directly into the 500 REST response body and ultimately shown to the admin. Consider using the generic translated fallback as the user-facing message, and keeping the Stripe message only in the log (where it already also appears).
3. Missing test docblock: `test_update_settings_skips_pmc_update_when_payment_method_ids_absent` has no PHPDoc, unlike its sibling test. Worth adding in one for consistency.
4. 500 test doesn't assert the error message: `test_update_settings_returns_500_when_pmc_update_fails` only asserts status code. Adding an assertion on `$response->get_data()['message']` would lock in the message-forwarding contract.

thuautp:
From Claude Code: Items 2, 3, and 4 are fixed in the latest push.

- **Item 2** — dropped `$response->error->message ??` from the `WP_Error` constructor; the Stripe message is already captured in the `WC_Stripe_Logger::error()` call above it, so the REST response now always uses the translated fallback.
- **Item 3** — added PHPDoc to `test_update_settings_skips_pmc_update_when_payment_method_ids_absent`.
- **Item 4** — added `$this->assertEquals( 'Unable to update payment method configuration.', $response->get_data()['message'] )` to `test_update_settings_returns_500_when_pmc_update_fails`.

**Item 1 (ordering decision):** The current order (PMC first) is intentional — it keeps the save atomic. If PMC fails after we've already written test mode or manual capture, the admin is left with settings in a partially-updated state that's hard to communicate and hard to undo. With fail-fast, they get a clear error, nothing has changed, and they can fix the issue and retry the whole save. I'd rather they retry once than need to figure out which settings actually persisted. Happy to discuss if you see a scenario where atomicity causes more harm than partial saves would.

------

@zhongruige based on this, I think PMC last could be harder to communicate and might confuse the user more. "Settings failed — please retry" is a better signal than "some things saved, payment methods didn't." Let me know what you think!

zhongruige:
I like that approach for the error messaging @thuautp! Looks good to me.

zhongruige (approved):
Overall looks good and tested great for me, happy to add my approval. We'll wait on @woocommerce/quark for final review and approval before merging.

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -284,14 +284,17 @@ public function get_settings() {
...
 	 */
 	public function update_settings( WP_REST_Request $request ) {
+		/* Settings > Payments accepted on checkout + Express checkouts */
+		$payment_method_ids_to_enable  = $this->get_payment_method_ids_to_enable( $request );
+		$is_upe_enabled                = $request->get_param( 'is_upe_enabled' );
+		$update_payment_methods_result = $this->update_enabled_payment_methods( $payment_method_ids_to_enable, $is_upe_enabled );
+		if ( is_wp_error( $update_payment_methods_result ) ) {
+			return new WP_REST_Response( [ 'message' => $update_payment_methods_result->get_error_message() ], 500 );
+		}
+
```
diegocurbelo, line 294: Moving this block to the top here change the behavior, now if the payment methods can not be saved, no settings are updated at all... and that includes the plugin enabled/disabled setting, so it can not be enabled/disabled (or the mode changed for example)

And while _we might want_ to prevent enabling the plugin if the `update_enabled_payment_methods` call fail, we still want the merchant to be able to disable the plugin.

I think we should move the block back to were it was... while that still shows the "Error saving settings" message it allows the plugin to be enabled/disabled.

diegocurbelo (changes requested):
The code looks good, and it tests well in general... except fo one use case that I think we should fix (update the  enable/disable setting even if the PMC call fails)

includes/admin/class-wc-rest-stripe-settings-controller.php
Mayisha: Agreed on this. We should not block other settings from updating.

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -284,14 +284,17 @@ public function get_settings() {
 	 * @return WP_REST_Response
 	 */
 	public function update_settings( WP_REST_Request $request ) {
+		/* Settings > Payments accepted on checkout + Express checkouts */
+		$payment_method_ids_to_enable  = $this->get_payment_method_ids_to_enable( $request );
+		$is_upe_enabled                = $request->get_param( 'is_upe_enabled' );
+		$update_payment_methods_result = $this->update_enabled_payment_methods( $payment_method_ids_to_enable, $is_upe_enabled );
+		if ( is_wp_error( $update_payment_methods_result ) ) {
+			return new WP_REST_Response( [ 'message' => $update_payment_methods_result->get_error_message() ], 500 );
```
Mayisha, line 292: I think the 500 here is not the right status. 500 Internal Server Error says our code faulted, but this failure is the Stripe PMC call coming back rejected or unreachable. That's an upstream problem, not an internal one.

includes/class-wc-stripe.php
```diff
@@ -1068,9 +1068,15 @@ public function maybe_toggle_payment_methods( WC_Payment_Gateways $gateways ) {
 			return;
 		}
 
-		$gateway->update_enabled_payment_methods(
+		$result = $gateway->update_enabled_payment_methods(
 			array_diff( $enabled_payment_methods, $payment_method_ids_to_disable )
 		);
+		if ( is_wp_error( $result ) ) {
+			WC_Stripe_Logger::error(
+				'Failed to update payment methods after deactivating incompatible plugins',
```
Mayisha, line 1076:
```suggestion
				'Failed to disable Stripe payment methods superseded by other active plugins',
```

This log message reads wrong to me. `maybe_toggle_payment_methods` doesn't deactivate any plugins, it disables our overlapping payment methods (Affirm/Klarna, etc.) precisely because the other official plugins are active. "Failed to update payment methods after deactivating incompatible plugins" implies we deactivated plugins, which we didn't.

Mayisha (changes requested):
Thanks for fixing this. Functionally it's working well. The code mostly looks good to me. I agree with Diego's feedback and I have a different opinion about the status code.

diegocurbelo (commented):
Hey @thuautp! Do you want me to take over this PR and implement the suggested changes?

includes/admin/class-wc-rest-stripe-settings-controller.php
daledupreez: Maybe a 502 is more appropriate for this specific error.

thuautp:
> Hey @thuautp! Do you want me to take over this PR and implement the suggested changes?

@diegocurbelo yes, please, if you could. Thank you!

wjrosa:
Hi @thuautp ! Diego is handling other things right now, so I will take this over instead. I will address all suggestions and update the PR

includes/admin/class-wc-rest-stripe-settings-controller.php
wjrosa: Fixed in a10362dd0: the failure response now comes last, after all other settings are applied, so the gateway can still be enabled or disabled. The test asserts the disable goes through.

includes/admin/class-wc-rest-stripe-settings-controller.php
wjrosa: Done in a10362dd0: the response is now 502, and the test asserts it.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -979,6 +979,7 @@ public function get_upe_available_payment_methods() {
 	 * Updates the enabled payment methods.
 	 *
 	 * @param string[] $payment_method_ids_to_enable
+	 * @return true|\WP_Error True on success, WP_Error on failure.
```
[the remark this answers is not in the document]
wjrosa: Both call sites log the WP_Error now (class-wc-stripe.php and the account keys controller), so those failures are no longer silent.

tests/phpunit/admin/class-wc-rest-stripe-settings-controller-test.php
```diff
@@ -136,6 +136,55 @@ public function test_update_stripe_payment_method_configurations_settings() {
...
+				]
+			);
+
+		$request = new WP_REST_Request( 'POST', self::SETTINGS_ROUTE );
+		$request->set_param( 'enabled_payment_method_ids', [ 'amazon_pay', 'card' ] );
+		$request->set_param( 'is_upe_enabled', true );
+
+		$response = $this->controller->update_settings( $request );
+		$this->assertEquals( 500, $response->get_status() );
+	}
```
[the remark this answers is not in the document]
wjrosa: The test asserts the response body message now, next to the status assertion.

includes/class-wc-stripe.php
```diff
@@ -1198,9 +1198,15 @@ public function maybe_toggle_payment_methods( WC_Payment_Gateways $gateways ) {
 			return;
 		}
 
-		$gateway->update_enabled_payment_methods(
+		$result = $gateway->update_enabled_payment_methods(
 			array_diff( $enabled_payment_methods, $payment_method_ids_to_disable )
 		);
+		if ( is_wp_error( $result ) ) {
+			WC_Stripe_Logger::error(
+				'Failed to disable Stripe payment methods superseded by other active plugins',
```
daledupreez: Non-blocking nit: I think we can use clearer language.
```suggestion
				'Failed to disable Stripe payment methods that should only be offered via other active plugins',
```

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -301,14 +301,14 @@ public function get_settings() {
 	 * @return WP_REST_Response
 	 */
 	public function update_settings( WP_REST_Request $request ) {
+		/* Settings > Payments accepted on checkout + Express checkouts */
+		$payment_method_ids_to_enable  = $this->get_payment_method_ids_to_enable( $request );
+		$is_upe_enabled                = $request->get_param( 'is_upe_enabled' );
+		$update_payment_methods_result = $this->update_enabled_payment_methods( $payment_method_ids_to_enable, $is_upe_enabled );
+
```
daledupreez: Non-blocking nit: Now that the error handling occurs at the end of the function, there is no longer any need for this code block to be moved in this PR. It feels like it would be simpler in the long run to move this back to its original location before we merge -- then the only diff would be the return value and small change in the whitespace.

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -329,6 +329,13 @@ public function update_settings( WP_REST_Request $request ) {
 		$this->update_is_debug_log_enabled( $request );
 		$this->update_oc_settings( $request );
 
+		// Surfaced last so a failed PMC update cannot block the other settings:
+		// the merchant must still be able to, for example, disable the gateway.
+		// 502, not 500: the failure is the upstream Stripe call, not our code.
```
daledupreez: Nit: I think we can improve on the comments.
```suggestion
		// Handle a PMC update error last so updates to other settings are saved.
		// Return an HTTP 502 error code because the error comes from Stripe.
```

tests/phpunit/admin/class-wc-rest-stripe-settings-controller-test.php
```diff
@@ -137,6 +137,71 @@ public function test_update_stripe_payment_method_configurations_settings() {
...
+	 * Test that a 502 response is returned when the PMC update fails.
+	 *
+	 * When the Stripe API rejects the payment method configuration update, the
+	 * settings endpoint must return an error (not 200) so the frontend can show
+	 * an error instead of the misleading "Settings saved." success notice. 502,
+	 * not 500, because the failure is the upstream Stripe call. The other
+	 * settings must still be applied, so the merchant can e.g. disable the
+	 * gateway while Stripe is rejecting PMC updates.
+	 */
+	public function test_update_settings_returns_502_when_pmc_update_fails() {
```
daledupreez, line 153: Nit: We can simplify the comments here as well.
```suggestion
	/**
	 * Test that a 502 response is returned when the PMC update fails.
	 *
	 * When the Stripe API doesn't accept a payment method configuration
	 * update, the settings endpoint must return an HTTP 502 error. The
	 * other settings must still be applied.
	 */
	public function test_update_settings_returns_502_when_pmc_update_fails() {
```

tests/phpunit/admin/class-wc-rest-stripe-settings-controller-test.php
```diff
@@ -137,6 +137,71 @@ public function test_update_stripe_payment_method_configurations_settings() {
...
+		// Set up initial state with only card enabled.
+		$this->mock_payment_method_configurations( [ 'card' ], [ 'amazon_pay' ] );
+
+		// Set pmc_enabled to yes to prevent migration.
+		$stripe_settings                = WC_Stripe_Helper::get_stripe_settings();
+		$stripe_settings['pmc_enabled'] = 'yes';
```
daledupreez, line 160: Just to be sure we are correctly testing the save behaviour, can you please ensure that `enabled` is set to `yes` as part of the setup?
```suggestion
		$stripe_settings['enabled']     = 'yes';
		$stripe_settings['pmc_enabled'] = 'yes';
```

tests/phpunit/admin/class-wc-rest-stripe-settings-controller-test.php
```diff
@@ -137,6 +137,71 @@ public function test_update_stripe_payment_method_configurations_settings() {
...
+		// still went through.
+		$this->assertSame( 'no', WC_Stripe_Helper::get_settings( null, 'enabled' ) );
+	}
+
+	/**
+	 * Test that the PMC update is skipped when enabled_payment_method_ids is absent.
+	 *
+	 * When the request does not include enabled_payment_method_ids, the settings
+	 * endpoint must not call the Stripe PMC API — the param is optional and its
+	 * absence means payment method selection was not part of this save.
```
daledupreez: Nit: Another comment we can simplify.
```suggestion
	 * When the request does not include enabled_payment_method_ids, the settings
	 * endpoint must not call the Stripe PMC API.
```

daledupreez (approved):
Thanks for the updates, @wjrosa, this is now looking pretty good to me. I do think we need to make the unit test tweak before merging, but my other suggestions are not blocking.

I also noticed that the error message is not used by the settings UI, but that should be handled in a follow-up PR once this PR has been merged.

includes/admin/class-wc-rest-stripe-settings-controller.php
wjrosa: Done in the latest commit: the block is back in its original position, so the diff is now just the captured result, the null guard, and the trailing 502.

wjrosa:
On the follow-up you mentioned (the error message not surfacing in the settings UI): it was small enough to do now, so I opened #5965, stacked on this branch. It reads the message from the REST error in the save notice and falls back to the generic text.

PR #5390, opened 2026-05-05
Fix compatibility with WooCommerce 10.8.0 beta

diegocurbelo:
Fixes STRIPE-1122
Fixes #<github_issue_id>

WC `10.8.0-beta.1` added a new guard at the start of `WC_Order::payment_complete()` — in the beta version it now bails out early (returning false, leaving status untouched) unless the order has checkout evidence, defined as either:

- `created_via` ∈ { checkout, store-api, rest-api, admin, pos-rest-api }
- a non-empty `cart_hash`

If neither is present, payment_complete() logs an error, adds an order note, fires `woocommerce_payment_complete_blocked`.

**What this PR changes:**
- Stamp Agentic Commerce orders with `created_via = 'stripe-agentic-commerce'`
- Register that value via the new `woocommerce_payment_complete_allowed_created_via_values` filter
- Update tests + helper to cover the new value
- Update PHPUnit Order Helper to add the `created_via = 'checkout'`

## Testing instructions

- Code review
- Make sure all checks pass, including the WC Beta tests:
    Screenshot 2026-05-06 at 00 32 08

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.7.0
+* Fix - Fix compatibility with WooCommerce 10.8 checkout-evidence guard (required created-via in orders)
```
daledupreez: Non-blocking nit: We could simplify the description.
```suggestion
* Fix - Fix compatibility with WooCommerce 10.8 payment handling, where created-via is required for orders
```

daledupreez (approved):
The changes look good to me!

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-integration-test.php
```diff
@@ -150,6 +150,14 @@ public function test_register_hooks() {
...
 			has_action( 'wc_stripe_agentic_commerce_sync_feed', [ $integration, 'sync_feed' ] )
 		);
+		$this->assertNotFalse(
+			has_filter( 'woocommerce_payment_complete_allowed_created_via_values', [ $integration, 'allow_agentic_payment_complete' ] )
+		);
+
+		$allowed = apply_filters( 'woocommerce_payment_complete_allowed_created_via_values', [], null );
+		$this->assertContains( \WC_Stripe_Agentic_Commerce_Order_Mapper::CREATED_VIA, $allowed );
+
+		remove_filter( 'woocommerce_payment_complete_allowed_created_via_values', [ $integration, 'allow_agentic_payment_complete' ] );
```
malithsen, line 160: Should we move this filter removal to a `tearDown` function just so that it doesn't remain if this test fails.

diegocurbelo: This inline `remove_filter` is here just symmetry with `test_is_enabled_when_filter_active` above. 

`WP_UnitTestCase::tear_down()` restores the hooks/filters, so the filter can't leak across tests if the assertion fails.

malithsen: TIL.
We seem to have a mix of both patterns in the repo. I see some not removing the filters and letting the parent `tear_down` restore them (like you suggested). Some use an explicit `tear_down`. In any case, this PR is good as is.

malithsen (approved):

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-integration-test.php
diegocurbelo: Yeah, Dale also mentioned that we use try/finally in some other cases... I personally prefer to keep the tests as minimal as possible, but we should choose and use **one** pattern.

PR #5392, opened 2026-05-06
Reclassify deprecated method removal as Update in changelog

wjrosa:
## Changes proposed in this Pull Request:

Reclassifies the `Remove deprecated backend methods that called wc_deprecated_function` entry from `Dev` to `Update` in both `changelog.txt` and `readme.txt` under `10.7.0`. Per [Dale's feedback on #5066](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5066#pullrequestreview-4234541739), removing these wrappers can affect merchants and extension authors who depended on the deprecated public surface, so the entry belongs under `Update` rather than `Dev`. Release-notes-only change; no code touched.

## Testing instructions

1. On this branch, open `changelog.txt` and confirm line 20 under `xxxx-xx-xx - version 10.7.0` reads:
   `* Update - Remove deprecated backend methods that called wc_deprecated_function`
2. Open `readme.txt` and confirm line 171 under `= 10.7.0 - xxxx-xx-xx =` reads:
   `* Update - Remove deprecated backend methods that called wc_deprecated_function`
3. Confirm no other lines were touched in either file.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️) — release-notes-only, no executable code changed
-   [x] Tested on mobile (or does not apply) — does not apply

### Changelog entry

-   [x] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment

This PR only edits the existing 10.7.0 changelog entry that landed in #5066 — it doesn't introduce any new behavior worth a separate entry.

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5394, opened 2026-05-06
Add Claude Code skills and repo review rules under .claude/

wjrosa:

Adds three repo-specific Claude Code skills and a `review-rules.md` checklist under `.claude/`. None of these touch plugin runtime — they are documentation files loaded by Claude Code (and any equivalent agent tooling) to provide repo-specific context.

### Files added

- `.claude/skills/changelog-entry/SKILL.md` — captures the dual-format trap (`changelog.txt` vs `readme.txt`), entry-type selection (Add/Update/Fix/Tweak/Dev), and the public-surface-removal classification rule (precedent: #5066 → #5392).
- `.claude/skills/payment-method-add/SKILL.md` — codifies the multi-step flow already documented in `AGENTS.md` (PHP class, constants, registration, icon, Blocks support) with an explicit four-surface validation matrix.
- `.claude/skills/phpstan-baseline-triage/SKILL.md` — encodes the project's "fix first, baseline last" rule and the "don't mix baseline churn with feature work" CRITICAL rule.
- `.claude/review-rules.md` — modelled on the equivalent file in the [WooPayments sibling repo](https://github.com/Automattic/woocommerce-payments/blob/develop/.claude/review-rules.md), adapted to this codebase: Stripe gateway hierarchy (`WC_Stripe_UPE_Payment_Gateway` / `WC_Stripe_UPE_Payment_Method`), webhook signature verification, the four-surface checkout validation requirement, WC L/L-1/L-2 support policy, and an Incident-Derived Rules section. The most important entry there is the **OCS duplicate-charge incident from #5325**: instantiating `WC_Stripe_UPE_Payment_Method_OC` to look up a display title also re-ran the constructor's hook registration, which caused subscription renewals to fire duplicate charges in 10.6.0. The section also covers the narrower scope `is_valid_optimized_checkout_page()` is meant to occupy (also #5325), the Stripe Radar + Subscriptions retry interaction (#5361), and the OCS + saved-tokens regression.
- `.claude/tmp/.gitignore` — scratch space convention from WooPayments (review notes, screenshots), gitignored except for the `.gitignore` itself.

### Discoverability

Added two lines to the `Documentation and Context Sources` section of `AGENTS.md` so contributors can find both the review rules and the skills directory.

### Why now

We had three reviewer-triggered learnings in flight this week (changelog classification on #5066, the PR-template gap on #5392, the link-text shortening on #5386) that should not be re-learned by every future contributor. The skills format mirrors the convention used in `Automattic/woocommerce-payments/.claude/skills/`, which is already in production use across that team.

This is opened as a **draft** to give the team room to review the wording before merge — particularly the Incident-Derived Rules section, where I'd rather have someone with first-hand context confirm the entries than ship a stale summary.

## Testing instructions

This PR adds documentation only — no executable code paths change.

1. From the branch, open `.claude/review-rules.md` and confirm the Architecture Compliance / Security / Performance / Test Coverage / Compatibility / Release Hygiene / Process / Incident-Derived Rules sections are accurate against current `AGENTS.md` and `includes/AGENTS.md`.
2. From the branch, open each `SKILL.md` under `.claude/skills/` and confirm the YAML frontmatter (`name`, `description`, `version`) and the workflow body match how the team actually does that task.
3. With Claude Code on this branch, type `/changelog-entry`, `/payment-method-add`, or `/phpstan-baseline-triage` — the skill content should load as agent context for that workflow.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️) — documentation files only; no executable code
-   [x] Tested on mobile (or does not apply) — does not apply

### Changelog entry

Added under `10.7.0` in both `changelog.txt` and `readme.txt`:

> Dev - Add Claude Code skills and review rules under .claude/ to capture repo-specific contributor guidance

**Post merge**


.claude/review-rules.md
```diff
@@ -0,0 +1,168 @@
...
+- [ ] Changes preserve WC L / L-1 / L-2 support (current major and previous two)
+- [ ] Changes preserve WP L / L-1 support (transitively constrained by WC's support policy)
+- [ ] PHP 7.4 syntax compatibility preserved (no PHP 8+ syntax in feature code)
+- [ ] WordPress / WooCommerce core files in `docker/wordpress/` and `docker/wordpress_xdebug/` are NEVER edited
+
+---
+
+## Release Hygiene
+
+- [ ] Version-impacting changes update `changelog.txt`, `readme.txt` stable tag and changelog, and the plugin file header *together*
```
malithsen: Do we ever update the stable tag directly? Shouldn't we leave that to the woorelease script?

.claude/skills/changelog-entry/SKILL.md
```diff
@@ -0,0 +1,83 @@
+---
+name: changelog-entry
+description: Use when adding or editing changelog entries for this plugin. Triggers include "add changelog", "changelog entry", "release notes", "update changelog.txt", "update readme.txt", or any PR that introduces user-visible behavior changes. Critical: changelog.txt and readme.txt use different version-header formats and must both be updated.
+version: 1.0.0
```
malithsen: `version` is not a valid frontmatter field: https://code.claude.com/docs/en/skills#frontmatter-reference

Applies to other SKILL.md files here as well

.claude/review-rules.md
wjrosa: Good catch — woorelease handles the stable tag at release time. Reworded the bullet to call out only `changelog.txt`, the `readme.txt` changelog section, and the plugin file header, with a note that the stable tag is woorelease's job. (9052ac7d5)

.claude/skills/changelog-entry/SKILL.md
wjrosa: Right — `version` isn't in the spec. Dropped it from all three SKILL.md files (`changelog-entry`, `payment-method-add`, `phpstan-baseline-triage`). (9052ac7d5)

.claude/skills/changelog-entry/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+2. Append `* Type - Description` to the bottom of that version's block.
+3. Locate the matching `= X.Y.Z - xxxx-xx-xx =` block in `readme.txt` under `== Changelog ==`.
+4. Append the *exact same* `* Type - Description` line.
+5. If the change is merchant-impacting, also add a bullet under `== Compatibility Notes ==` for that version in `readme.txt`.
+
+## Wording conventions
+
+- Describe behavior, not implementation.
+  - ❌ "Add `basename( WC_STRIPE_PLUGIN_PATH )` helper for slug derivation"
+  - ✅ "Append a 'See what's new' link to the plugin Updated! message after manual updates"
```
daledupreez: Two comments here:
 - Are the images going to be understood by the LLM? Maybe it could be better to use explicit "Incorrect:" and "Correct" labels.
 - There are a lot of `'` characters in the correct line
```suggestion
  - Incorrect: Add `basename( WC_STRIPE_PLUGIN_PATH )` helper for slug derivation
  - Correct: Append a "See what's new" link to the plugin Updated! message after manual updates
```

.claude/skills/changelog-entry/SKILL.md
```diff
@@ -0,0 +1,82 @@
...
+- Lead with the verb. ✅ "Cancel pending WC Subscriptions retry when Stripe Radar blocks a renewal payment"
```
daledupreez: As above, I think it will be clearer if we use explicit text.
```suggestion
- Lead with the verb: "Cancel pending WC Subscriptions retry when Stripe Radar blocks a renewal payment"
```

.claude/skills/payment-method-add/SKILL.md
```diff
@@ -0,0 +1,77 @@
...
+
+Add the method ID to `includes/constants/class-wc-stripe-payment-methods.php` using the existing `STRIPE_<METHOD>` constant convention. The constant is referenced from webhook handling, payment processing, and admin REST endpoints — leaving it out is a silent breakage.
+
+### 3. Registration
+
+Add the new class to the payment method classes array initialized in `WC_Stripe::init()`. The order in this array influences display order on checkout — match the design's intended position rather than appending blindly.
+
+### 4. Icon
+
+Add the icon asset under `client/payment-method-icons/` (SVG preferred). Register it in the icons map used by the payment-methods rendering layer so `get_icon()` resolves correctly.
```
daledupreez: Not blocking: would it be worth referencing the file that contains this icon map?

.claude/skills/payment-method-add/SKILL.md
```diff
@@ -0,0 +1,77 @@
...
+## Validation matrix (CRITICAL)
+
+Per the project's CRITICAL rule, any change to payment-method availability or rendering must be validated across all four checkout surfaces. A new method must be verified on:
+
+| Surface | Where to test |
+|---------|---------------|
+| Classic checkout | `/checkout` with Blocks disabled |
+| Blocks checkout | `/checkout` with Blocks enabled (default for new sites) |
+| Optimized Checkout (OCS) | `/checkout` with OCS enabled in plugin settings |
+| Express checkout (ECE) | If the method supports express, verify Apple Pay and Google Pay paths |
```
daledupreez: This instruction doesn't make a lot of sense to me -- should we not test the Product, Cart, and Checkout pages?

.claude/skills/payment-method-add/SKILL.md
```diff
@@ -0,0 +1,77 @@
...
+
+If the method intentionally does not support a surface, verifying that it's *cleanly hidden* — not erroring — is still a tested outcome.
+
+## Tests
+
+- **PHPUnit:** add `tests/phpunit/PaymentMethods/test-class-wc-stripe-upe-payment-method-<id>.php`. Use `@dataProvider` for capability matrix scenarios (subscriptions on/off, pre-orders on/off, currency, country).
```
daledupreez: Is this the correct file name? I thought we were adding `-test` as a suffix.
```suggestion
- **PHPUnit:** add `tests/phpunit/PaymentMethods/class-wc-stripe-upe-payment-method-<id>-test.php`. Use `@dataProvider` for capability matrix scenarios (subscriptions on/off, pre-orders on/off, currency, country).
```

.claude/skills/payment-method-add/SKILL.md
```diff
@@ -0,0 +1,77 @@
...
+- **Jest:** if the Blocks config or icon registration has logic beyond static configuration, add a Jest test under `tests/js/`.
```
daledupreez: This also seems wrong -- we tend to add tests `__tests__` folders alongside the code being tested.

.claude/skills/phpstan-baseline-triage/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+2. **For each new error, classify by category:**
+
+   | Category | Action |
+   |----------|--------|
+   | Real `null` / `false` handling missing | **Fix the code** |
+   | Wrong type annotation that lies about the actual return | **Fix the annotation** |
+   | Untyped WP/WC hook return (`mixed`, etc.) | **Add a defensive check at the boundary** |
+   | Stripe SDK type that genuinely allows multiple shapes | **Narrow with explicit assertion or fix the call site** |
+   | Genuine PHPStan limitation (intersection types, generic variance) | **Baseline with a `# Reason:` comment** |
```
daledupreez: Should we baseline or should we add an inline ignore rule? I think it will be much better to have an inline message about why something is OK rather than generating a separate baseline file that can't include any commentary.

.claude/review-rules.md
```diff
@@ -0,0 +1,168 @@
...
+- **Don't repurpose `is_valid_optimized_checkout_page()` as a generic "is OCS in play" check:** `WC_Stripe_UPE_Payment_Gateway::is_valid_optimized_checkout_page()` only answers "is this request the OCS-rendered checkout page itself?" Using it elsewhere (order received, order details, etc.) caused those pages to display "Stripe" as the payment method instead of the actual one. If you need a broader "OCS enabled / OCS could be in play" check, use a separate predicate or the relevant settings option directly. Precedent: PR #5325.
+- **Subscription renewals + Stripe Radar:** when Radar blocks a renewal, the WC Subscriptions retry queue must be cancelled rather than left pending — otherwise the next retry triggers the same Radar block (PR #5361).
+- **Saved tokens + OCS:** Optimized Checkout Suite must not hide saved payment tokens from the saved-cards list. Verify token retrieval works with OCS enabled and disabled when touching token logic.
```
daledupreez, line 168: I would add that new constructors should not include side effects whenever possible.

daledupreez (commented):
This is looking good overall, but I do have some suggestions and feedback.

.claude/skills/changelog-entry/SKILL.md
wjrosa: Applied — explicit `Incorrect`/`Correct` labels, kept the `"See what's new"` contraction inside double quotes per your suggestion.

.claude/skills/changelog-entry/SKILL.md
wjrosa: Done.

.claude/skills/payment-method-add/SKILL.md
wjrosa: Done — now points at `client/payment-method-icons/index.js`.

.claude/skills/payment-method-add/SKILL.md
wjrosa: You're right — the row was conflating wallet brand with placement. Rewrote it to list ECE's four placements: product, cart, checkout, and pay-for-order.

.claude/skills/payment-method-add/SKILL.md
wjrosa: Good catch on the suffix. Used the actual lowercase `tests/phpunit/payment-methods/` path; `includes/AGENTS.md` says `PaymentMethods/` but the filesystem disagrees — separate follow-up.

.claude/skills/payment-method-add/SKILL.md
wjrosa: Updated — points to `__tests__/` alongside the code (example: `client/blocks/upe/__tests__/`).

.claude/skills/phpstan-baseline-triage/SKILL.md
wjrosa: Agreed — inline reasons travel with the code and get reviewed in the same diff. Switched the table to recommend `@phpstan-ignore-next-line` with a `Reason:` comment, and demoted `phpstan:baseline` to a fallback for non-editable files.

.claude/review-rules.md
wjrosa: Added — folded onto the duplicate-charge bullet as a corollary: constructors should avoid side effects (hooks, network, option writes) so classes stay safe to instantiate for metadata reads. Side effects belong in an explicit `init()`/`register_hooks()` called from the bootstrap.

.claude/skills/payment-method-add/SKILL.md
```diff
@@ -0,0 +1,77 @@
...
+## Validation matrix (CRITICAL)
+
+Per the project's CRITICAL rule, any change to payment-method availability or rendering must be validated across all four checkout surfaces. A new method must be verified on:
+
+| Surface | Where to test |
+|---------|---------------|
+| Classic checkout | `/checkout` with Blocks disabled |
+| Blocks checkout | `/checkout` with Blocks enabled (default for new sites) |
+| Optimized Checkout (OCS) | `/checkout` with OCS enabled in plugin settings |
+| Express checkout (ECE) | Verify the Apple Pay / Google Pay / Link buttons on each ECE surface: product page, cart, checkout, and pay-for-order |
```
daledupreez: Nit: We should include Amazon Pay here.
```suggestion
| Express checkout (ECE) | Verify the Apple Pay / Google Pay / Link / Amazon Pay buttons on each ECE surface: product page, cart, checkout, and pay-for-order |
```

.claude/skills/phpstan-baseline-triage/SKILL.md
```diff
@@ -0,0 +1,88 @@
...
+
+2. **For each new error, classify by category:**
+
+   | Category | Action |
+   |----------|--------|
+   | Real `null` / `false` handling missing | **Fix the code** |
+   | Wrong type annotation that lies about the actual return | **Fix the annotation** |
+   | Untyped WP/WC hook return (`mixed`, etc.) | **Add a defensive check at the boundary** |
+   | Stripe SDK type that genuinely allows multiple shapes | **Narrow with explicit assertion or fix the call site** |
+   | Genuine PHPStan limitation (intersection types, generic variance) | **Add an inline `@phpstan-ignore-next-line` with a reason comment** |
```
daledupreez: Nit: I think this should be an inline `@phpstan-ignore` comment that specifies what rule(s) are being ignored.
```suggestion
   | Genuine PHPStan limitation (intersection types, generic variance) | **Add an inline `@phpstan-ignore` comment with a specific error type to ignore as well as descriptive text** |
```

.claude/skills/phpstan-baseline-triage/SKILL.md
```diff
@@ -0,0 +1,88 @@
...
+
+3. **Fix legitimate issues first.** Most level 8 errors indicate real null/false paths that production hasn't hit yet — the type checker is doing its job.
+
+4. **For genuine PHPStan limitations, prefer an inline ignore over the baseline:**
+
+   ```php
+   /** @phpstan-ignore-next-line Reason: PHPStan can't follow the intersection narrowing here. */
```
daledupreez: This is not the correct format, especially if we want specific issues ignored.
```suggestion
   // @phpstan-ignore function.alreadyNarrowedType (This is the output of a filter that may return other data types.)
```

.claude/skills/phpstan-baseline-triage/SKILL.md
```diff
@@ -0,0 +1,88 @@
...
+- Errors whose message mentions `null` or `false` — those are usually real.
+- Errors in critical paths: payment processing, webhook handling, intent controller, customer linking. A type error here is the kind of thing that becomes a fatal in production.
+
+## When baseline is the right answer
+
+- Genuine PHPStan limitations (intersection types it can't follow).
+- WP/WC core types that are wrong upstream and will be fixed there.
+- Stripe SDK return types that PHPStan over-narrowed.
+
+In all baseline-only cases, the `phpstan-baseline.neon` entry should carry a `# Reason:` comment so future reviewers can judge whether the suppression is still justified later. Prefer the inline `@phpstan-ignore-next-line` form whenever the file is editable — the reason lives next to the code instead of in a sibling file.
```
daledupreez: We can't add any commentary to the baseline as it is auto-generated. I think we should modify this instruction.
```suggestion
Prefer the inline `@phpstan-ignore` comment form whenever the file is editable — the reason lives next to the code instead of in a sibling file.
```

daledupreez (commented):
I think we still have some things to fix here, but I think we're super-close.

.claude/review-rules.md
```diff
@@ -0,0 +1,168 @@
...
+- [ ] Frontend behavior changes have Jest coverage under `tests/js/` or E2E coverage under `tests/e2e/`
+- [ ] Payment method availability/rendering changes are validated across classic, Blocks, OCS, and ECE
+- [ ] Recurring-payment-capable changes include subscription/pre-order regression coverage
+
+### Conventions
+
+- [ ] PHPUnit parameterized tests use `@dataProvider`
+- [ ] PHPUnit test methods include a PHPDoc describing the contract being guarded
+- [ ] E2E specs that mutate global store settings (currency, payment method availability) live in a dedicated Playwright project and CI matrix job, not in `default`
+- [ ] Tests live nearest to the code they cover (`tests/phpunit/Admin/`, `tests/phpunit/PaymentMethods/`, etc.)
```
malithsen:
```suggestion
- [ ] Tests live nearest to the code they cover (`tests/phpunit/admin/`, `tests/phpunit/payment-methods/`, etc.)
```

Actual directories use kebab-case

malithsen (approved):
Thanks, this looks good to me!

PR #5399, opened 2026-05-06
Add "Release Notes" link to the plugins page row

wjrosa:

Adds a **Release notes** link to the WooCommerce Stripe row meta on the Plugins admin screen (alongside Docs / Support). The link reuses WordPress' built-in plugin information modal, opened on the changelog tab via thickbox — the same modal that the parent PR (#5386) uses for the post-update "See what's new" link.

The idea for this entry-point came from @daledupreez, who suggested that surfacing the changelog from the plugin row (not only after a manual update) gives merchants a stable, always-available place to read what's new.

This is stacked on top of #5386 — please merge that one first. The controller now exposes a small `get_changelog_url()` helper so the JS-localized params and the new row link share one source of truth.

Preview:
Screenshot 2026-05-06 at 17 08 12

## Testing instructions

1. With this branch active, visit **WP Admin → Plugins**.
2. On the WooCommerce Stripe row, confirm a **Release notes** link appears among the row meta links (next to Docs / Support).
3. Click **Release notes** → the WordPress plugin information modal should open on the **Changelog** tab without leaving the page.
4. Confirm no equivalent link is added to other plugin rows.
5. Run `npm run test:php -- --filter=WC_Stripe_Plugins_Page_Controller_Test` and verify it passes.

---

### Changelog entry

Add - Add a "Release notes" link to the WooCommerce Stripe row on the Plugins page that opens the changelog modal

**Post merge**


daledupreez (commented):
Rather than requiring both PRs to be merged for this to get shipped, might it make sense to restructure this PR so it duplicates the server-side logic to compute the URL and enqueue the thickbox dependency, and then we can land this simpler PR quickly?

I am concerned that the other PR is a bit trickier to test, and don't want to merge too much just before we release.

wjrosa:
@daledupreez good call — restructured per your suggestion. Rebased onto `develop`, pulled the URL helper and `add_thickbox()` inline (kept the JS variant in #5386), and re-targeted the base. Single commit now: 4 files, 101 lines.

tests/phpunit/admin/class-wc-stripe-plugins-page-controller-test.php
```diff
@@ -65,4 +68,45 @@ public function provide_enqueue_scripts_hook_suffixes(): array {
 			'plugins.php registers and enqueues assets'    => [ 'plugins.php', true ],
 		];
 	}
+
+	/**
+	 * Tests that the "Updated!" changelog link relies on thickbox being enqueued
+	 * so the plugin information modal can open from the plugins.php page.
+	 *
```
malithsen: We should remove the "Updated!" reference as that change is not included in this PR.

includes/admin/class-wc-stripe-plugins-page-controller.php
```diff
@@ -73,10 +74,63 @@ public function enqueue_scripts( $hook_suffix = null ) {
...
+	}
+
+	/**
+	 * Builds the URL for the WordPress plugin information modal, opened on the changelog tab.
+	 *
+	 * @return string The thickbox-iframe URL.
+	 */
+	private function get_changelog_url(): string {
+		return self_admin_url(
+			'plugin-install.php?tab=plugin-information&plugin=' . $this->get_plugin_slug()
```
malithsen, line 106: It's possible for the plugin to be installed on a non-standard directory, in which case the URL lookup will fail to find the WP.org listing. I'd consider hard-coding the slug here.

malithsen (approved):
Looks good and works as described. Left a couple of optional suggestions.

daledupreez:
Question: how will the link work if a site is running an older version of the plugin? e.g. Site is running 10.7.0, and 10.8.0 has been released. Will we show the release notes for the currently installed version or the currently released version?

includes/admin/class-wc-stripe-plugins-page-controller.php
wjrosa: Good catch — switched to a hard-coded `'woocommerce-gateway-stripe'` slug in `get_plugin_slug()` so the modal still resolves when the directory is renamed (e.g. `woocommerce-gateway-stripe-1` from a re-upload). Pushed in cb26ceebd.

tests/phpunit/admin/class-wc-stripe-plugins-page-controller-test.php
wjrosa: Done — rewrote the docblock to describe the "Release notes" link being tested, since the "Updated!" link lives in #5386. Pushed in cb26ceebd.

wjrosa:
Good question — the modal pulls WP.org's current `readme.txt` changelog regardless of installed version, since it reuses the same `plugin-install.php?tab=plugin-information` endpoint core uses for "View details" on plugin rows. So a 10.7.0 site with 10.8.0 released would see 10.8.0 entries at the top.

To avoid surfacing release notes for a version that isn't installed, I pushed 1c24d1e23 to skip our link when WordPress has staged an update for the plugin (`update_plugins` site transient). In that case core already injects its own "View details" link into the row, so users on an older version still have a path to the changelog — they just go through core's link instead of ours, and we don't add a redundant entry.

includes/admin/class-wc-stripe-plugins-page-controller.php
```diff
@@ -76,28 +77,94 @@ public function enqueue_scripts( $hook_suffix = null ) {
...
+	 *
+	 * The link reuses WordPress' built-in plugin information modal, opened on the
+	 * changelog tab via thickbox (already enqueued by `enqueue_scripts`).
+	 *
+	 * @param array  $links Existing plugin row meta links.
+	 * @param string $file  Plugin file the row belongs to.
+	 * @return array Updated row meta links.
+	 */
+	public function add_release_notes_link( $links, $file ): array {
+		if ( plugin_basename( WC_STRIPE_MAIN_FILE ) !== $file ) {
```
daledupreez: Code Rabbit suggested in one of the comments that now seems to be buried that we should ensure `$links` is an array when we get it, otherwise we could trigger a fatal by not complying with our method signature.
```suggestion
		$links = (array) $links;

		if ( plugin_basename( WC_STRIPE_MAIN_FILE ) !== $file ) {
```

includes/admin/class-wc-stripe-plugins-page-controller.php
```diff
@@ -76,28 +77,94 @@ public function enqueue_scripts( $hook_suffix = null ) {
...
+			return (array) $links;
+		}
```
daledupreez, line 133: This function will be called for every installed plugin, so it feels like it may be worth storing the return value of `plugin_basename()` so we only compute it once.
```suggestion

	private string $stripe_plugin_basename = '';

	private function get_plugin_basename(): string {
		if ( '' === $this->stripe_plugin_basename ) {
			$this->stripe_plugin_basename = plugin_basename( WC_STRIPE_MAIN_FILE );
		}
		return $this->stripe_plugin_basename;
	}

	/**
	 * Appends a "Release Notes" link to the plugin row meta on the plugins admin page.
	 *
	 * The link reuses WordPress' built-in plugin information modal, opened on the
	 * changelog tab via thickbox (already enqueued by `enqueue_scripts`).
	 *
	 * @param array  $links Existing plugin row meta links.
	 * @param string $file  Plugin file the row belongs to.
	 * @return array Updated row meta links.
	 */
	public function add_release_notes_link( $links, $file ): array {
		if ( $this->get_plugin_basename() !== $file ) {
			return (array) $links;
		}
```

daledupreez (approved):
Overall this is looking good to me, but I think we need to compute the plugin basename only once and we need to guard against non-array values for the `$links` parameter, either by removing the `array` return type, or by casting to an array at the start of the method.

includes/admin/class-wc-stripe-plugins-page-controller.php
```diff
@@ -76,28 +77,103 @@ public function enqueue_scripts( $hook_suffix = null ) {
...
+		if ( $this->get_plugin_basename() !== $file ) {
+			return (array) $links;
+		}
+
+		// When an update is available, WordPress core already injects its own
+		// "View details" link into the plugin row pointing at the same modal.
+		// Skip ours to avoid redundancy and to avoid surfacing release notes
+		// for a version the user has not installed yet.
+		if ( $this->has_pending_update( $file ) ) {
+			return $links;
```
daledupreez, line 140: @wjrosa, in this code path, `$links` may not be an array. If it isn't, this method will fatal because it has an explicit `array` return type. Can you add defensive code to either cast `$links` to an array or remove the explicit `array` return type from the method?

wjrosa: Sorry, I missed this before the merge. Opened #5441: single `(array) $links` at the top of `add_release_notes_link()`.

PR #5400, opened 2026-05-06
Hide OCS first-position notice when methods above Stripe are disabled

Mayisha:
Fixes STRIPE-1112

The Optimized Checkout Suite (OCS) "Move Stripe to the top" notice was firing whenever a non-Stripe payment method sat above Stripe in `woocommerce_gateway_order` — even when those methods were disabled. In that case Stripe is already the first method merchants and shoppers actually see at checkout, so the notice is misleading.

The fix replaces the option-based scan in `WC_Stripe_Helper::is_stripe_in_position_one_in_woocommerce_gateway_order()` with a single iteration over `WC()->payment_gateways->payment_gateways`. WC core's `WC_Payment_Gateways::init()` already sorts that property by the same `woocommerce_gateway_order` option (then `ksort`s by position), so iterating it directly gives the effective checkout order. Disabled non-Stripe gateways are skipped because they don't render at checkout; Stripe itself counts as "reached" regardless of its own enabled state, preserving the prior intent of the helper.

A few side-effects of consolidating the data source:

- Stale entries in `woocommerce_gateway_order` (left behind by deactivated plugins) no longer count toward Stripe's effective position. They aren't loaded as gateway instances, so they don't render at checkout — silencing the notice in that case is the correct outcome.
- The `_wc_*` PES sentinel skip is no longer needed: those entries aren't real `WC_Payment_Gateway` instances, so they don't appear in the loaded property.
- One DB option read removed; the helper now does a single `foreach` over an in-memory array.

The single caller (`should_show_stripe_first_method_notice`) is untouched, and there are no frontend, Blocks, or checkout-flow changes — the React notice component just consumes the precomputed flag.

## Testing instructions

1. On a clean dev site, ensure the Stripe extension is installed and enabled, and Optimized Checkout Suite is **on** (`woocommerce_stripe_settings → optimized_checkout_element = yes`).
2. Go to **WooCommerce → Settings → Payments**. Drag a non-Stripe payment method (e.g. Cash on Delivery) above Stripe in the list. Leave it **disabled**. Disable any other non-Stripe methods that sit above Stripe.
3. Reload the page and open the Stripe section.
   - **Expected:** The "Optimized Checkout works best when Stripe is your first payment method. Move it to the top..." notice is **not** shown.
   - **Pre-fix (regression):** The notice was incorrectly displayed.
4. Enable Cash on Delivery while keeping it above Stripe. Reload.
   - **Expected:** The notice **is** shown (no regression of STRIPE-1030 happy path).
5. Click "Move to top" on the notice — the page should reload with Stripe in first position and the notice hidden.
6. Disable Cash on Delivery again and reload.
   - **Expected:** The notice goes away (Stripe is effectively first again).
7. **Settings page check:** Repeat step 2 and visit **WooCommerce → Settings → Stripe → Advanced settings** instead — the same notice rendering rules apply on that page.
8. **Dismissal:** Confirm the dismiss-X on the notice still hides it on subsequent loads (`wc_stripe_show_stripe_first_method_notice` option set to `no`).

---

-   [ ] Tested on mobile (or does not apply) — no UI changes; admin-only logic.

### Changelog entry

A changelog entry **is** included for this fix. See `changelog.txt` and `readme.txt` (10.7.0 unreleased block).

**Post merge**

-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply) — does not apply (admin notice fix, no merchant-facing doc change).


includes/class-wc-stripe-helper.php
```diff
@@ -645,30 +645,34 @@ public static function should_show_stripe_first_method_notice(): bool {
...
-		}
+		$loaded_gateways = WC()->payment_gateways->payment_gateways ?? [];
+
+		foreach ( $loaded_gateways as $gateway ) {
+			/**
+			 * Loaded payment gateway instance.
+			 *
+			 * @var WC_Payment_Gateway $gateway
+			 */
+			if ( WC_Stripe_UPE_Payment_Gateway::ID === $gateway->id || 0 === strpos( $gateway->id, 'stripe_' ) ) {
```
daledupreez, line 662: Is it worth adding a further check for `$gateway` being a Stripe payment method instance? Or is that redundant?

daledupreez (approved):
The changes LGTM and test well.

diegocurbelo (approved):
LGTM 👍🏼

PR #5402, opened 2026-05-06
Introduce WC_Stripe settings accessors as the canonical settings API (slice 1 of the helper-shim migration)

wjrosa:
Towards STRIPE-1321

First slice of the settings-ownership migration, reshaped per [review feedback](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5402#pullrequestreview-4746600923): this PR only introduces the new API; call-site migrations follow in small per-subsystem PRs, and the old helpers get deprecated last, once no in-tree callers remain.

The main `woocommerce_stripe_settings` option is owned by `WC_Stripe_Helper` — a kitchen-sink class — via `get_stripe_settings()` / `update_main_stripe_settings()`, called from ~60 production and ~140 test sites. This mirrors WooPayments' direction of accessing settings through a canonical owner instead.

- New canonical accessors on the plugin root: `WC_Stripe::get_instance()->get_settings()` / `update_settings()`, plus `WC_Stripe::read_method_settings_option( $method )` for per-method options and a `SETTINGS_OPTION_NAME` constant. Raw option semantics are preserved (`$settings['x'] ?? 'default'` keeps working; non-array option values normalize to `[]`).
- The `WC_Stripe_Helper` shims now delegate to the new accessors, so there is a single settings code path from day one — **no deprecation notices yet**, and no behavior change for the ~200 existing call sites.
- `WC_Stripe` itself migrates its internal reads/writes to the new methods.
- Unit tests for the new accessors; the existing helper suite (255 tests) exercises the delegating shims unchanged.

Follow-ups: per-subsystem call-site migration PRs — #5743 (admin), #5744 (payment methods/gateways), #5745 (core services), #5746 (support classes) — then a final PR adding `@deprecated` + `_deprecated_function()` to the shims.

**BC impact:** none — no public symbol is removed, renamed, or deprecated in this PR; the shims keep their exact signatures and semantics.

## Testing instructions

1. Run `npm run test:php` (or `--filter WC_Stripe_Test` / `WC_Stripe_Helper_Test`).
2. Smoke-check settings save on the Stripe settings page and a checkout — behavior is unchanged.

---

### Changelog entry

Changelog entries added to `changelog.txt` and `readme.txt` under 10.9.0 (Dev).

<details>

#### Comment

</details>

**Post merge**


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -35,6 +35,70 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
...
+	 * persisted value regardless of which write path ran.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @param array $settings The settings to persist.
+	 * @return bool Whether the option was actually written (matches `update_option`).
+	 */
+	public function update_settings( array $settings ): bool {
+		return WC_Stripe::write_settings_option( $settings );
+	}
```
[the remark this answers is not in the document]
wjrosa: Done in 105baabb9 — `update_settings()` now calls `refresh_settings_cache()` after the write so it no longer depends on the main gateway's hooks being active.

includes/class-wc-stripe.php
```diff
@@ -9,6 +9,61 @@
  */
 class WC_Stripe {
 
+	/**
+	 * The option name that stores the main Stripe gateway settings array.
+	 *
+	 * @since 10.9.0
+	 */
+	public const SETTINGS_OPTION_NAME = 'woocommerce_stripe_settings';
```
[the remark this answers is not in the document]
wjrosa: Done in 105baabb9 — both hook names are now built from `self::SETTINGS_OPTION_NAME` (same string, no external change).

tests/phpunit/admin/class-wc-stripe-express-checkout-controller-test.php
```diff
@@ -22,12 +22,12 @@ public function setUp(): void {
...
 
 		// Setup existing keys
-		$settings                         = WC_Stripe_Helper::get_stripe_settings();
+		$settings                         = WC_Stripe::read_settings_option();
 		$settings['publishable_key']      = 'original-live-key-9999';
 		$settings['secret_key']           = '';
 		$settings['test_publishable_key'] = 'original-test-key-9999';
 		$settings['test_secret_key']      = '';
-		WC_Stripe_Helper::update_main_stripe_settings( $settings );
+		WC_Stripe::write_settings_option( $settings );
```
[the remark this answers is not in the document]
wjrosa: Done in 105baabb9 — `test_admin_scripts` now asserts the seeded publishable key reaches the localized `wc_stripe_express_checkout_settings_params`.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -1519,9 +1519,9 @@ public function test_checkout_saves_sepa_generated_payment_method_to_order() {
 	 * @dataProvider provider_handle_saving_payment_method_respects_per_method_toggle
 	 */
 	public function test_handle_saving_payment_method_respects_per_method_toggle( string $sepa_tokens_for_ideal, bool $expected_save ) {
-		$stripe_settings                          = WC_Stripe_Helper::get_stripe_settings();
+		$stripe_settings                          = WC_Stripe::read_settings_option();
 		$stripe_settings['sepa_tokens_for_ideal'] = $sepa_tokens_for_ideal;
-		WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
+		WC_Stripe::write_settings_option( $stripe_settings );
```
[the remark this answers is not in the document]
wjrosa: Done in 105baabb9 — the test now reads and writes through `$this->mock_gateway->get_settings()`/`update_settings()`.

tests/phpunit/admin/migrations/class-wc-stripe-subscriptions-repairer-legacy-sepa-tokens-test.php
```diff
@@ -89,7 +89,7 @@ public function tear_down() {
 	 * We can't mock the check for WC_Subscriptions, so we'll test the rest of the conditions.
 	 */
 	public function test_updater_gets_scheduled_on_right_conditions() {
-		WC_Stripe_Helper::update_main_stripe_settings( [ 'upe_checkout_experience_enabled' => 'yes' ] );
+		WC_Stripe::write_settings_option( [ 'upe_checkout_experience_enabled' => 'yes' ] );
```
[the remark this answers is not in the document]
wjrosa: Skipping — the prior `update_main_stripe_settings()` also overwrote the whole option, so dropping `test_connection_type` is pre-existing behavior these tests already pass with; not a regression from this refactor.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -35,6 +35,72 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 	 */
 	protected $retry_interval = 1;
 
+	/**
+	 * Object-cache key/group for the request-scoped main settings cache.
+	 *
+	 * A single shared cache entry (rather than a per-instance property) so every
+	 * gateway instance reads the same value — no divergent in-memory copies.
+	 */
+	protected const SETTINGS_CACHE_GROUP = 'woocommerce_stripe';
```
Mayisha: AI feedback

abstract-wc-stripe-payment-gateway.php documents get_settings() as caching in a "request-scoped object-cache entry," but SETTINGS_CACHE_GROUP = 'woocommerce_stripe' is never passed to
  wp_cache_add_non_persistent_groups(). Under a persistent object cache (Redis/Memcached) the entry persists across requests — the docblock claim is false. Invalidation depends entirely on the
  add/update/delete_option_woocommerce_stripe_settings hooks, which are only registered when a UPE gateway is constructed in the writing request. A settings write in a context that doesn't construct the
  gateway (some cron/CLI/REST paths) would leave a stale entry across requests.

  Fix: register the group as non-persistent at bootstrap (wp_cache_add_non_persistent_groups( 'woocommerce_stripe' )) to match the stated intent and remove the cross-request staleness risk.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -355,7 +358,7 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 		 * Otherwise for new connections return 'yes' for `upe_checkout_experience_enabled` field.
 		 */
 		private function get_upe_checkout_experience_enabled() {
-			$existing_stripe_settings = WC_Stripe_Helper::get_stripe_settings();
+			$existing_stripe_settings = WC_Stripe::get_instance()->get_main_stripe_gateway()->get_settings();
```
Mayisha: AI feedback:

get_upe_checkout_experience_enabled() (connect.php:361) now reads through ...->get_main_stripe_gateway()->get_settings(), which constructs and memoizes the UPE gateway at line 274 — before the new keys are
  persisted at line 292. The author's own comment (lines 267–269, 333–335) explicitly intends first construction to happen at line 335, after keys are written. Practical impact is low (PMC migration at line
  339 and webhook config both read fresh state via read_settings_option() / WC_Stripe_API), but the gateway instance ends up caching empty secret_key/publishable_key and stale payment_methods for the rest
  of that request, and the documented ordering guarantee is silently broken.

  Fix: use WC_Stripe::read_settings_option() in get_upe_checkout_experience_enabled(), consistent with the write path right above it.

includes/connect/class-wc-stripe-connect.php
Mayisha, line 311: We are calling a deprecated method here.

Mayisha (commented):
Feedback from quick code inspection and AI reviews

- includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php still have a couple of `WC_Stripe_Helper::get_stripe_settings()` calls.
- Some of the tests are calling the deprecated method.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Good catch — fixed in 1ef647b6c. The group is now registered via `wp_cache_add_non_persistent_groups()` at bootstrap (`WC_Stripe::init()`), so the cache stays request-scoped under Redis/Memcached and the docblock claim holds.

includes/connect/class-wc-stripe-connect.php
wjrosa: Agreed — fixed in 1ef647b6c. `get_upe_checkout_experience_enabled()` now reads via `WC_Stripe::read_settings_option()`, so the gateway is no longer constructed/memoized before the new keys are persisted.

includes/connect/class-wc-stripe-connect.php
wjrosa: Fixed in 1ef647b6c — now uses `WC_Stripe::write_settings_option()`.

wjrosa:
Thanks @Mayisha! All addressed in 1ef647b6c — the OCS+AP migration and the remaining test call sites now use `WC_Stripe::read_settings_option()`/`write_settings_option()`. The only deprecated calls left are in `WC_Stripe_Helper` itself and the test that asserts the deprecation notices.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -35,6 +35,88 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
...
+	 *
+	 * Called once at plugin bootstrap (WC_Stripe::init()).
+	 *
+	 * @since 10.9.0
+	 *
+	 * @return void
+	 */
+	public static function register_settings_cache_group(): void {
+		wp_cache_add_non_persistent_groups( self::SETTINGS_CACHE_GROUP );
+	}
```
daledupreez: At a high level, I don't think it's worth the complication and overhead of managing object cache data if we are aiming for a per-request scope. I don't see what the `wp_cache_*()` approach buys us over a static variable.

Given that the gateway instance may depend on configuration, it feels like it might be better to have the settings logic live on `WC_Stripe` or possibly a dedicated class like `WC_Stripe_Settings` (which could also be a singleton). Our model with multiple possible gateway instances makes me lean towards that _not_ being the right place for the settings management, particularly because this approach is introducing the following call pattern:
```php
$settings = WC_Stripe::get_instance()->get_main_stripe_gateway()->get_settings();
```
It feels like it would be more direct to call `WC_Stripe::get_instance()->get_settings()` or `WC_Stripe_Settings::get_instance()->get_settings()`.

Based on the above, it also feels like having a shorthand _may_ be worth it, if only for making it easier to follow the calling code:
```php
public static function get_stripe_settings(): array {
	return WC_Stripe::get_instance()->get_settings();
}
```

includes/class-wc-stripe-apple-pay-registration.php
```diff
@@ -177,15 +177,17 @@ private function make_domain_registration_request( $secret_key ) {
...
 			$this->make_domain_registration_request( $secret_key );
 
 			// No errors to this point, registration success!
 			// Reload the settings, to avoid overwriting old, cached values.
-			$settings                              = WC_Stripe_Helper::get_stripe_settings();
+			$settings                              = $gateway->get_settings();
 			$settings['apple_pay_verified_domain'] = $this->domain_name;
 			$settings['apple_pay_domain_set']      = 'yes';
-			WC_Stripe_Helper::update_main_stripe_settings( $settings );
+			$gateway->update_settings( $settings );
```
daledupreez: This calling pattern is relatively common. It may be worth considering some sort of helper that allows specific keys to be updated, but that feels like it should be handled in a follow-up.

includes/class-wc-stripe-apple-pay-registration.php
```diff
@@ -177,15 +177,17 @@ private function make_domain_registration_request( $secret_key ) {
...
 
 			// Update cached settings.
 			$this->stripe_settings = $settings;
```
daledupreez, line 191: Not for this PR, but it would be _really_ good to remove local caches/copies of the settings data like this one.

includes/class-wc-stripe-feature-flags.php
```diff
@@ -196,7 +196,7 @@ public static function should_expand_ocs_in_legacy_checkout(): bool {
 	 * @return bool
 	 */
 	public static function is_oc_available() {
-		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
+		$stripe_settings = WC_Stripe::read_settings_option();
```
daledupreez: Why are we doing direct reads for these settings? Should we not want to prime the cache/local copy so we're avoiding later reads? Given that we're not making any updates here, it feels like any unrelated writes to the options should flow through.

includes/class-wc-stripe-helper.php
```diff
@@ -45,25 +45,29 @@ class WC_Stripe_Helper {
 	/**
 	 * Get the main Stripe settings option.
 	 *
+	 * @deprecated 10.9.0 Use `WC_Stripe::read_settings_option()` (or `WC_Stripe::read_method_settings_option( $method )`).
```
daledupreez: Why are we directing code to the lower-level API instead of the gateway API? Most external callers should not be calling this code to start with, and should be using the higher-level API for any unexpected use cases.

This applies to the other methods below as well.

includes/class-wc-stripe-mode.php
```diff
@@ -23,7 +23,7 @@ public static function is_live() {
 	 * @return bool Whether the plugin is in test mode.
 	 */
 	public static function is_test() {
-		$settings = WC_Stripe_Helper::get_stripe_settings();
+		$settings = WC_Stripe::read_settings_option();
 		return 'yes' === ( $settings['testmode'] ?? 'no' );
```
daledupreez, line 27: It feels like these reads should be via the cached option so we're avoid as much work as possible rather than reading the value from the option every time.

includes/class-wc-stripe.php
```diff
@@ -9,6 +9,61 @@
...
+	 * Static counterpart to read_settings_option(). The main gateway's
+	 * option-change hooks invalidate its cached copy after this write.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @param array $settings The settings to persist.
+	 * @return bool Whether the option was actually written (matches `update_option`).
+	 */
+	public static function write_settings_option( array $settings ): bool {
+		return update_option( self::SETTINGS_OPTION_NAME, $settings );
```
daledupreez, line 41: I am not seeing any hooks or actions that clear the higher-level cache when the option gets updated, and we're directing callers to these methods, which will always bypass the cache management at the higher-level. I think we need to make sure the implementation is tighter all round the possibility of a stale cache needs to be hugely reduced.

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -136,7 +136,7 @@ public function test_already_on_10_8_writes_only_ran_once_flag() {
 	 */
 	public function test_migration_scenarios( array $scenario ) {
 		update_option( self::STRIPE_VERSION_OPTION, $scenario['previous_version'] );
-		WC_Stripe_Helper::update_main_stripe_settings(
+		WC_Stripe::write_settings_option(
```
daledupreez: At a high level, it feels like all of our tests should be using the higher-level update helper to ensure that the top-level cache is correctly populated, given that most of our code should be calling the high-level helper to get the settings.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
daledupreez: After looking at more of the code, it feels like co-locating the two layers of code is going to be really important long-term, whether that's on `WC_Stripe` or a dedicated singleton. I say that because I can't tell which API I should be calling in which situations, and that will only get worse with time and code getting copy/pasta'ed. With the current split, the caching logic is also separate from the read/write code, which makes the knock-on effects of gaps or incorrect calls even more insidious and subtle.

.gitignore
```diff
@@ -50,3 +50,5 @@ phpunit-html/
 
 # PHPStan temporary files
 phpstan-tmp/
+
+.claude/scheduled_tasks.lock
```
daledupreez: I don't think this change should be included in this PR.

daledupreez (changes requested):

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Co-located on `WC_Stripe` as suggested: `get_settings()`/`update_settings()`/`refresh_settings_cache()` and the cache now live there, with the option-change invalidation hooks registered once at bootstrap. The gateway methods are thin delegates now, so there's one implementation and one entry point (`WC_Stripe::get_instance()->get_settings()`). `read_settings_option()`/`write_settings_option()` stay only as the low-level primitives the cached accessor wraps.

On wp_cache vs a static: the non-persistent group is what keeps it correct under PHPUnit — a DB rollback between tests restores the option but doesn't re-fire the option-change hooks, so a static property goes stale across tests; non-persistent wp_cache is flushed per test.

includes/class-wc-stripe.php
wjrosa: These hooks did exist, but were buried in the UPE gateway constructor (easy to miss). Moved them onto `WC_Stripe` and register them unconditionally at bootstrap, so any write to the option — ours, WooCommerce's admin save, or a raw `update_option()` — flushes the cache.

includes/class-wc-stripe-helper.php
wjrosa: Switched: the deprecated shims and all internal callers now point at the high-level cached `WC_Stripe::get_instance()->get_settings()`/`update_settings()`. `read_settings_option()` is now just the low-level primitive backing the cache.

includes/class-wc-stripe-feature-flags.php
wjrosa: Now routed through the cached `WC_Stripe::get_instance()->get_settings()`, so these prime/hit the shared cache instead of reading the option directly.

includes/class-wc-stripe-mode.php
wjrosa: Done — reads via the cached accessor now.

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
wjrosa: Swept all tests over to the high-level helper so the top-level cache stays populated.

includes/class-wc-stripe-apple-pay-registration.php
wjrosa: Agreed — leaving the targeted key-update helper as a follow-up.

includes/class-wc-stripe-apple-pay-registration.php
wjrosa: Agreed, out of scope here — follow-up.

.gitignore
wjrosa: Removed.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
daledupreez: My concern here is as much about overhead/latency as it is about correctness. By using the caching layer, we may be introducing network calls when a non-local object cache is in use or database calls when caching is via transient, and the benefit of that is not clear to me when options should already be cached. I think that it may be best to drop our caching altogether, as it mostly introduces complexity and I am not sure that it delivers many benefits.

includes/admin/class-wc-rest-stripe-account-keys-controller.php
```diff
@@ -280,7 +280,8 @@ public function validate_test_secret_key( $param, $request, $key ) {
 	 * @return WP_REST_Response
 	 */
 	public function set_account_keys( WP_REST_Request $request ) {
-		$settings       = WC_Stripe_Helper::get_stripe_settings();
+		$gateway        = WC_Stripe::get_instance()->get_main_stripe_gateway();
+		$settings       = $gateway->get_settings();
```
daledupreez: Why not call the getter on `WC_Stripe` directly, and then keep the `WC_Stripe` instance for the later update?
```suggestion
		$wc_stripe      = WC_Stripe::get_instance();
		$settings       = $wc_stripe->get_settings();
```

includes/admin/class-wc-rest-stripe-account-keys-controller.php
```diff
@@ -312,7 +313,7 @@ public function set_account_keys( WP_REST_Request $request ) {
 		// Before saving the settings, decommission any previously automatically configured webhook endpoint.
 		$settings = $this->decommission_configured_webhook_after_key_update( $settings, $current_account_keys );
 
-		WC_Stripe_Helper::update_main_stripe_settings( $settings );
+		$gateway->update_settings( $settings );
```
daledupreez: Linked to the comment above, it feels like we should either call `$wc_stripe->update_settings()` or `WC_Stripe::get_instance()->update_settings()`.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -259,13 +259,16 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
...
-			$current_options                            = WC_Stripe_Helper::get_stripe_settings();
-			$options                                    = array_merge( $default_options, is_array( $current_options ) ? $current_options : [] );
+			$publishable_key = $result->publishableKey; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+			$secret_key      = $result->secretKey; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+			$is_test         = 'live' !== $mode;
+			$prefix          = $is_test ? 'test_' : '';
+			$default_options = $this->get_default_stripe_config();
+			// Read/write the option directly (not via the gateway instance) so the
+			// gateway's first construction — and its PMC migration — still happens
+			// below, after the new keys are persisted.
```
daledupreez: I think this comment is stale - the code no longer instantiates the gateway.
```suggestion
```

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: That makes sense! I dropped the caching in 7973e975f. Thanks for pushing back into this

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -35,6 +35,46 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
...
+	 * Returns the main Stripe gateway settings array.
+	 *
+	 * Thin delegate to the co-located, cached accessor on WC_Stripe so callers
+	 * holding a gateway instance share the same single source of truth.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @return array
+	 */
+	public function get_settings(): array {
```
daledupreez: I think it would be cleaner if we didn't add these wrapper methods on the gateway, as I think it's cleaner to have one code path to make the right one clear to everyone. While these are convenient to have, I don't know that they are a good thing overall.

includes/admin/class-wc-rest-stripe-account-keys-controller.php
wjrosa: Updated the call below in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5402/commits/4849df161694938ad6c27b6beac230c7f467de88

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Agreed — removed both gateway wrappers; every caller now goes through `WC_Stripe::get_instance()->get_settings()/update_settings()`.

includes/admin/class-wc-rest-stripe-account-keys-controller.php
wjrosa: Done — now calls `$wc_stripe->update_settings()` directly.

includes/class-wc-stripe.php
```diff
@@ -9,6 +9,85 @@
...
+	 * @since 10.9.0
+	 *
+	 * @param string $method The payment method slug.
+	 * @return array
+	 */
+	public static function read_method_settings_option( string $method ): array {
+		$settings = get_option( 'woocommerce_stripe_' . $method . '_settings', [] );
+
+		return is_array( $settings ) ? $settings : [];
+	}
```
daledupreez: Given that we're no longer using our own cache, I think we should only add `get_settings()` and `update_settings()` -- the static functions don't provide much additional utility.
```suggestion
	/**
	 * Get the Stripe settings option. Always returns an array.
	 *
	 * @since 10.9.0
	 *
	 * @return array
	 */
	public function get_settings(): array {
		$settings = get_option( self::SETTINGS_OPTION_NAME, [] );

		return is_array( $settings ) ? $settings : [];
	}

	/**
	 * Persists the main Stripe settings.
	 *
	 * @since 10.9.0
	 *
	 * @param array $settings The settings to persist.
	 * @return bool Whether the option was actually written (matches `update_option`).
	 */
	public function update_settings( array $settings ): bool {
		return update_option( self::SETTINGS_OPTION_NAME, $settings );
	}

	/**
	 * Reads the raw per-method settings option directly (e.g. "boleto" →
	 * `woocommerce_stripe_boleto_settings`). Always returns an array.
	 *
	 * @since 10.9.0
	 *
	 * @param string $method The payment method slug.
	 * @return array
	 */
	public static function read_method_settings_option( string $method ): array {
		$settings = get_option( 'woocommerce_stripe_' . $method . '_settings', [] );

		return is_array( $settings ) ? $settings : [];
	}
```

daledupreez (commented):
Given how huge this PR is getting, I am wondering whether it may be worth getting the new APIs merged, and then tackling a number of smaller PRs that shift over to the new API, and then finally marking the old API as deprecated.

It may take us longer than getting this PR shipped, but it would reduce the risks and would allow us to use smaller sets of changes.

wjrosa:
Thanks, Dale! I agree: reshaped to that plan in 252ca302a. Follow-up PR is at https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5743

includes/class-wc-stripe.php
```diff
@@ -9,6 +9,53 @@
...
+	/**
+	 * Reads the raw per-method settings option directly (e.g. "boleto" →
+	 * `woocommerce_stripe_boleto_settings`). Always returns an array.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @param string $method The payment method slug.
+	 * @return array
+	 */
+	public static function read_method_settings_option( string $method ): array {
```
daledupreez: Not blocking, but is it worth using a similar method name to `get_settings()`? It feels like it would be clearer if we called this new function `get_payment_method_settings( string $payment_method )`? We might want to call the argument `$method_slug` or similar to ensure that we're describing the expected value better.
```suggestion
	public static function get_payment_method_settings( string $method_slug ): array {
```

includes/class-wc-stripe.php
```diff
@@ -906,13 +954,13 @@ public function register_routes() {
...
 
-		$settings_controller = new WC_REST_Stripe_Settings_Controller( $this->get_main_stripe_gateway() );
+		$settings_controller = new WC_REST_Stripe_Settings_Controller( $main_gateway );
 		$settings_controller->register_routes();
 
 		$stripe_account_keys_controller = new WC_REST_Stripe_Account_Keys_Controller( $this->account );
 		$stripe_account_keys_controller->register_routes();
 
-		$oc_setting_toggle_controller = new WC_Stripe_REST_OC_Setting_Toggle_Controller( $this->get_main_stripe_gateway() );
+		$oc_setting_toggle_controller = new WC_Stripe_REST_OC_Setting_Toggle_Controller( $main_gateway );
```
daledupreez: Not blocking, but this cleanup is not strictly related the main purpose of this PR, and I am not sure that there are any specific testing instructions for this. It may be worth splitting this cleanup out into a dedicated PR.

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -8,6 +8,74 @@
...
 class WC_Stripe_Test extends WC_Mock_Stripe_API_Unit_Test_Case {
+	/**
+	 * get_settings() returns the stored option as an array.
+	 *
+	 * @return void
+	 */
+	public function test_get_settings_returns_stored_option(): void {
+		update_option( WC_Stripe::SETTINGS_OPTION_NAME, [ 'enabled' => 'yes' ] );
+
+		$this->assertSame( 'yes', WC_Stripe::get_instance()->get_settings()['enabled'] );
```
daledupreez: Would it not be more explicit to assert that we get the same array structure back, and not just a single key? Ideally, we'd start with a richer array structure, as the single key value is a pretty simple test case.
```suggestion
		$this->assertSame( [ 'enabled' => 'yes' ], WC_Stripe::get_instance()->get_settings() );
```

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -8,6 +8,74 @@
...
+	 * update_settings() persists the option and subsequent reads reflect the new value.
+	 *
+	 * @return void
+	 */
+	public function test_update_settings_persists_option(): void {
+		WC_Stripe::get_instance()->update_settings( [ 'enabled' => 'yes' ] );
+		$this->assertSame( 'yes', WC_Stripe::get_instance()->get_settings()['enabled'] );
+
+		WC_Stripe::get_instance()->update_settings( [ 'enabled' => 'no' ] );
+		$this->assertSame( 'no', WC_Stripe::get_instance()->get_settings()['enabled'] );
```
daledupreez: Question: Is it worth calling `get_option()` to ensure we're testing the lower-level functionality? While `get_settings()` is not using a cache at present, it might in future.

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -8,6 +8,74 @@
...
+	 *
+	 * @return void
+	 */
+	public function test_get_settings_reflects_raw_option_write(): void {
+		WC_Stripe::get_instance()->update_settings( [ 'enabled' => 'yes' ] );
+		$this->assertSame( 'yes', WC_Stripe::get_instance()->get_settings()['enabled'] );
+
+		update_option( WC_Stripe::SETTINGS_OPTION_NAME, [ 'enabled' => 'no' ] );
+		$this->assertSame( 'no', WC_Stripe::get_instance()->get_settings()['enabled'] );
+	}
```
daledupreez, line 71: The PHPDoc feels a bit stale _and_ the implementation is exactly the same as `test_update_settings_persists_option()`. Do we need both tests? Or maybe we can keep one copy under a clearer name?

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -8,6 +8,74 @@
...
+	 * normalizes non-array values.
+	 *
+	 * @return void
+	 */
+	public function test_read_method_settings_option_reads_raw_option(): void {
+		update_option( 'woocommerce_stripe_boleto_settings', [ 'foo' => 'bar' ] );
+
+		$this->assertSame( 'bar', WC_Stripe::read_method_settings_option( 'boleto' )['foo'] );
+		$this->assertSame( [], WC_Stripe::read_method_settings_option( 'nonexistent_method' ) );
+	}
```
daledupreez: Can we drive this test via a data provider so we can cover more payment methods, not just Boleto?

includes/class-wc-stripe.php
daledupreez: Also, is it worth using some clearer or better-defined constant or type as the argument? I can easily see code supplying a value like `stripe_<method>` or use the wrong case.

includes/class-wc-stripe-helper.php
```diff
@@ -45,25 +45,27 @@ class WC_Stripe_Helper {
...
 	 * @param string $method (Optional) The payment method to get the settings from.
 	 * @return array $settings The Stripe settings.
 	 */
 	public static function get_stripe_settings( $method = null ) {
-		$settings = null === $method ? get_option( self::SETTINGS_OPTION, [] ) : get_option( 'woocommerce_stripe_' . $method . '_settings', [] );
-		if ( ! is_array( $settings ) ) {
-			$settings = [];
-		}
-		return $settings;
+		return null === $method ? WC_Stripe::get_instance()->get_settings() : WC_Stripe::read_method_settings_option( $method );
```
daledupreez: Not blocking nit: given that we're delegating to two different methods here, it might be clearer to use an early return rather than the conditional.
```suggestion
		if ( null === $method ) {
			return WC_Stripe::get_instance()->get_settings();
		}
		
		return WC_Stripe::read_method_settings_option( $method );
```

daledupreez (commented):
I have some more feedback from inspection. Thanks for slimming down the PR! I think that will be much easier to review and merge.

includes/class-wc-stripe.php
wjrosa: Applied — and completed the rename in 8041441eb (body/param, helper call site, test).

includes/class-wc-stripe.php
wjrosa: Reverted to keep the PR scoped (c81dc771b); will split it out separately if still worth it.

tests/phpunit/class-wc-stripe-test.php
wjrosa: Applied, with one adjustment (8041441eb): the pre_update_option filter merges defaults into the write, so the whole-array comparison is against the stored option, plus the key assertion.

tests/phpunit/class-wc-stripe-test.php
wjrosa: Good point — the update_settings() test now asserts via get_option() (c81dc771b), so it survives a future cache.

tests/phpunit/class-wc-stripe-test.php
wjrosa: De-duplicated in c81dc771b: the update test now asserts via get_option() (write path), the raw-write test only covers update_option → get_settings() (read path), with fresh PHPDocs.

tests/phpunit/class-wc-stripe-test.php
wjrosa: Done in c81dc771b — provider covers Boleto/Klarna/SEPA plus the normalization cases.

includes/class-wc-stripe.php
wjrosa: Documented the argument against the WC_Stripe_Payment_Methods constants and made the accessor normalize exactly the two mistakes you predicted — `stripe_`-prefixed gateway ids and wrong case (c81dc771b). A real value type would want PHP 8 enums, which 7.4 support rules out.

includes/class-wc-stripe-helper.php
wjrosa: Applied.

includes/class-wc-stripe-helper.php
```diff
@@ -45,25 +45,31 @@ class WC_Stripe_Helper {
 	/**
 	 * Get the main Stripe settings option.
 	 *
+	 * Delegates to the canonical accessors on WC_Stripe so there is a single
+	 * settings code path. Will be deprecated once in-tree callers migrate.
```
daledupreez: Is it worth adding an explicit `@deprecated` PHPDoc annotation in this PR? That will at least let IDEs flag the usages where they come up, but shouldn't impact unit tests or runtime uses.

includes/class-wc-stripe-helper.php
```diff
@@ -45,25 +45,31 @@ class WC_Stripe_Helper {
...
 	 *
+	 * Delegates to the canonical accessor on WC_Stripe so there is a single
+	 * settings code path. Will be deprecated once in-tree callers migrate.
+	 *
 	 * @param $options array The Stripe settings.
 	 * @return void
 	 */
 	public static function update_main_stripe_settings( $options ) {
-		update_option( self::SETTINGS_OPTION, $options );
+		WC_Stripe::get_instance()->update_settings( (array) $options );
```
daledupreez: Is it worth adding a more explicit guard against non-array values? It feels like the array cast could result in bad outcomes that don't match the previous behaviour. Prior to the cast, a scalar value would overwrite the option directly. With the new code and cast, we'd end up writing `[ $options ]` for any scalar value of `$options`. I would prefer that bad callers result in less destructive outcomes, so I prefer an explicit check for an incoming array.
```suggestion
		if ( is_array( $options ) ) {
			WC_Stripe::get_instance()->update_settings( (array) $options );
		}
```

includes/class-wc-stripe.php
```diff
@@ -9,6 +9,63 @@
...
+	 *                            constant. Gateway ids ("stripe_boleto") and mixed case are
+	 *                            normalized to the slug.
+	 * @return array
+	 */
+	public static function get_payment_method_settings( string $method_slug ): array {
+		$method_slug = strtolower( $method_slug );
+
+		// Gateway ids are the slug with a "stripe_" prefix and target the same
+		// option, so accept them rather than silently reading a wrong option.
+		if ( 0 === strpos( $method_slug, 'stripe_' ) ) {
```
daledupreez: Not blocking nit: I prefer `str_starts_with()` as an immediately readable version of this check.

includes/class-wc-stripe.php
```diff
@@ -9,6 +9,63 @@
...
+			$method_slug = substr( $method_slug, strlen( 'stripe_' ) );
```
daledupreez: Not blocking: is it worth guarding against `'' === $method_slug`?

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -8,6 +8,101 @@
...
+	public function test_get_settings_returns_stored_option(): void {
+		update_option( WC_Stripe::SETTINGS_OPTION_NAME, [ 'enabled' => 'yes' ] );
+
+		// Compare against the stored option rather than the literal written
+		// above: the pre_update_option filter merges defaults into the write.
+		$this->assertSame(
+			get_option( WC_Stripe::SETTINGS_OPTION_NAME ),
+			WC_Stripe::get_instance()->get_settings()
+		);
+		$this->assertSame( 'yes', WC_Stripe::get_instance()->get_settings()['enabled'] );
```
daledupreez: Not blocking nit: can we get the written data once and then assert against the value? It shouldn't change between calls, but it's clearer if we fetch once and then assert.
```suggestion
		// Compare against the stored option rather than the literal written
		// above: the pre_update_option filter merges defaults into the write.
		$updated_settings = WC_Stripe::get_instance()->get_settings();
		$this->assertSame(
			get_option( WC_Stripe::SETTINGS_OPTION_NAME ),
			$updated_settings
		);
		$this->assertSame( 'yes', $updated_settings['enabled'] );
```

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -8,6 +8,101 @@
...
+	 * The static get_payment_method_settings() reads the method's option,
+	 * normalizing gateway-id and mixed-case inputs to the slug.
+	 *
+	 * @dataProvider provide_payment_method_settings_inputs
+	 *
+	 * @param string $input       Value passed to get_payment_method_settings().
+	 * @param string $option_name Option the read should resolve to.
+	 * @return void
+	 */
+	public function test_get_payment_method_settings_reads_raw_option( string $input, string $option_name ): void {
```
daledupreez: Nit: I think we should rename `$input` to `$method_slug` so we are explicit.
```suggestion
	public function test_get_payment_method_settings_reads_raw_option( string $method_slug, string $option_name ): void {
```

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -8,6 +8,101 @@
...
+		update_option( $option_name, [ 'foo' => 'bar' ] );
+
+		$this->assertSame( 'bar', WC_Stripe::get_payment_method_settings( $input )['foo'] );
```
daledupreez: Is it worth asserting that the result has the same shape, not that it has the same value?
```suggestion
		$this->assertEquals( [ 'foo' => 'bar' ], WC_Stripe::get_payment_method_settings( $input ) );
```

daledupreez (approved):
The changes LGTM, though I have some non-blocking feedback.

includes/class-wc-stripe-helper.php
wjrosa: Done in 093787e1b — added `@deprecated 10.9.0` tags to both shims.

includes/class-wc-stripe.php
wjrosa: Agreed — switched to `str_starts_with()` in 093787e1b.

includes/class-wc-stripe.php
wjrosa: Good call — added the guard plus a test in 093787e1b.

includes/class-wc-stripe.php
```diff
@@ -9,6 +9,68 @@
...
+
+		// An empty slug would resolve to the nonsensical "woocommerce_stripe__settings" option.
+		if ( '' === $method_slug ) {
+			return [];
+		}
+
+		$settings = get_option( 'woocommerce_stripe_' . $method_slug . '_settings', [] );
+
+		return is_array( $settings ) ? $settings : [];
+	}
```
diegocurbelo: The per payment method settings were only used pre-upe (by the deprecated legacy gateway).
I don't think we should add this public method here, we can keep the `woocommerce_stripe_' . $method_slug` read with a deprecation notice in the original `WC_Stripe_Helper::get_stripe_settings()` instead of adding a new already deprecated method.

diegocurbelo (approved):
The code changes look good to me, and the smoke tests work as expected.

I left an optional suggestion for the per-method settings option wrapper.

includes/class-wc-stripe.php
wjrosa: Done in 932f9943d — removed `WC_Stripe::get_payment_method_settings()`; the per-method read stays inline in the deprecated `WC_Stripe_Helper::get_stripe_settings()`, whose notice now says the per-method form has no replacement since only the legacy gateways used those options.

PR #5404, opened 2026-05-07
Explore: route Stripe SDK calls through WordPress' HTTP API

wjrosa:

**Status: draft / exploratory.** Foundation for the upcoming `WC_Stripe_API` → SDK migration. Stacked on #5403 — merge that first.

The Stripe PHP SDK ships its own curl-based HTTP transport. That's normally fine, but in this codebase it gets in the way:

- The PHPUnit suite mocks Stripe responses by hooking `pre_http_request`. Curl requests bypass that filter entirely, so any test that exercises an SDK callsite either hits the live network or fails with `api_key cannot be the empty string`.
- WP's `wp_safe_remote_*` allowlist + any host-level HTTP filters (proxies, debug capture, third-party plugins) don't see SDK traffic.
- The bulk migration we attempted on top of #5403 produced **430 failures + 23 errors** for exactly this reason.

This PR introduces `WC_Stripe_SDK_WP_Http_Client` — an implementation of `\Stripe\HttpClient\ClientInterface` on top of `wp_safe_remote_request` — and registers it as the SDK's global HTTP client from `WC_Stripe_Client::get()` via `\Stripe\ApiRequestor::setHttpClient()`. After this lands, every SDK call goes through the WP HTTP API and inherits the same mock seam, allowlist, and filter chain the rest of the plugin uses.

### What the adapter does

- Maps the SDK's `(method, $absUrl, $headers, $params, $hasFile, $apiMode, $maxNetworkRetries)` contract onto `wp_safe_remote_request()`.
- POST + v1: passes `$params` as a form-encoded array body (WP handles encoding).
- POST + v2: JSON-encodes the body and adds `Content-Type: application/json`.
- GET / DELETE: appends params to the URL as a query string (mirrors `Stripe\HttpClient\CurlClient::constructUrlAndBody()`).
- Translates the SDK's `[ "Header: value", … ]` array into the assoc form WP expects, and back again on the response.
- Returns `[ body, status, headers ]` — the exact tuple the SDK contract requires.
- Surfaces `WP_Error` as `\Stripe\Exception\ApiConnectionException`.
- Throws `\Stripe\Exception\UnexpectedValueException` for file uploads (no callsite in this codebase needs them today; the next PR can wire that up if/when it does).
- Uses the same 70-second timeout `WC_Stripe_API::request()` carries.

### What it deliberately doesn't do

- **No retries.** `WC_Stripe_API::request()` doesn't retry today either, so we keep parity. Adding retries is a separate, scoped change with its own opinions about idempotency.
- **No request/response logging.** `WC_Stripe_Client::request()` (in #5403's follow-up scope) handles that at the wrapper layer; the adapter is a pure transport.

### Why this unblocks the migration

With the adapter in place, the bulk `WC_Stripe_API::` → `WC_Stripe_Client::` rename can land without rewriting tests: every existing `pre_http_request` mock keeps intercepting Stripe traffic. The follow-up PRs (per-resource migration → final cleanup that turns `WC_Stripe_API::request()` / `::retrieve()` / `::request_with_level3_data()` into thin shims) all build on this.

### Verification

- `npm run test:php` — full suite, **1922 tests / 4431 assertions, all green**.
- `npm run phpstan` — clean.
- `npm run lint:php` — clean.
- New file `tests/phpunit/class-wc-stripe-sdk-wp-http-client-test.php` covers 5 cases on the adapter directly: form-encoded v1 POST, JSON v2 POST, GET query-string handling, `WP_Error` → `ApiConnectionException`, and file-upload rejection.

## Testing instructions

This is foundation infrastructure with no user-visible behavior change. Smoke test:

1. Visit **WP Admin → WooCommerce → Settings → Payments → Stripe** and confirm the React UI still loads and the Stripe account banner is correct (this exercises a number of SDK / `WC_Stripe_API` paths).
2. Place a test order through classic + Blocks checkout — the customer-create flow already runs through the SDK after #5403.
3. `npm run test:php` (full suite).
4. `npm run phpstan`.

---

-   [x] Tested on mobile (or does not apply) — refactor only, admin/checkout-side.

### Changelog entry

<details>

#### Comment

Internal refactor / dep wiring; no user-visible behavior change. The user-visible entry already lives on #5403 (the parent PR introducing `stripe/stripe-php` and `WC_Stripe_Client`).

</details>

**Post merge**

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,8 @@
         "includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php",
         "includes/admin/class-wc-rest-stripe-exit-survey-controller.php",
         "includes/admin/class-wc-stripe-plugins-page-controller.php",
-        "includes/class-wc-stripe-client.php"
+        "includes/class-wc-stripe-client.php",
+        "includes/class-wc-stripe-sdk-wp-http-client.php"
       ]
     },
     "autoload-dev": {
--- a/includes/class-wc-stripe-client.php
+++ b/includes/class-wc-stripe-client.php
@@ -50,6 +50,11 @@ class WC_Stripe_Client {
 				self::PARTNER_ID
 			);
 
+			// Route SDK traffic through `wp_safe_remote_request` so existing
+			// `pre_http_request` mocks, third-party WP HTTP filters, and the
+			// `wp_safe_remote_*` allowlist all continue to apply.
+			\Stripe\ApiRequestor::setHttpClient( new WC_Stripe_SDK_WP_Http_Client() );
+
 			self::$client = new \Stripe\StripeClient(
 				[
 					'api_key'        => $secret_key,
--- a/includes/class-wc-stripe-sdk-wp-http-client.php
+++ b/includes/class-wc-stripe-sdk-wp-http-client.php
@@ -1,0 +1,156 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Stripe PHP SDK HTTP transport that routes through WordPress' HTTP API.
+ *
+ * The default `\Stripe\HttpClient\CurlClient` uses curl directly. That bypasses
+ * `wp_safe_remote_*`, the `pre_http_request` filter the WP test framework relies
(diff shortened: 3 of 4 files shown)


wjrosa:
Postponing this work for now.

PR #5405, opened 2026-05-07
Explore: add WC_Stripe_Client request/retrieve/request_with_level3_data wrappers

wjrosa:

**Status: draft / exploratory.** Stacked on #5404 (WP HTTP adapter) → #5403 (SDK + factory). Merge in that order.

Now that SDK calls go through `wp_safe_remote_request` (#5404), the SDK is a viable replacement transport. This PR adds the call-site-level entry points for migrating away from `WC_Stripe_API`:

- `WC_Stripe_Client::request( $request, $api, $method, $with_headers )` — drop-in for `WC_Stripe_API::request()`.
- `WC_Stripe_Client::retrieve( $api )` — drop-in for `WC_Stripe_API::retrieve()`, including the consecutive-401 invalid-API-key tracking the old helper carried.
- `WC_Stripe_Client::request_with_level3_data( $request, $api, $level3_data, $order )` — drop-in for the level-3 retry logic.

All three preserve the cross-cutting behavior the old helpers rely on:
- `wc_stripe_request_headers` / `wc_stripe_request_body` / `wc_stripe_idempotency_key` filters.
- Idempotency-key generation for `charges` POST and `payment_intents` POST.
- Request/response logging in the same format `WC_Stripe_Logger::debug()` produces today.
- Returning a `stdClass` with the same `->error` shape on API errors so existing callsites keep checking `! empty( $response->error )` without a try/catch.

### Adapter tweaks shaken loose by wiring this up

`WC_Stripe_SDK_WP_Http_Client` got two refinements (committed alongside the wrapper since they're functionally tied):

1. The `wc_stripe_request_headers` filter runs **inside the adapter**, on the SDK-built header set. The SDK doesn't allow an opaque `headers` key in request opts, so the wrapper can't pass them through; the adapter is the one place that sees the full header set just before it goes to WP HTTP.
2. The adapter now tolerates `pre_http_request` mocks that return `'response' => 200` as a bare int (which is what most tests in this repo do), not just the canonical `'response' => [ 'code' => 200, 'message' => 'OK' ]` shape. Without this, `wp_remote_retrieve_response_code()` returns `0`, the SDK rejects it as "Invalid response object", and a chunk of the test suite fails.

`WC_Stripe_Client::get()` also falls back to a placeholder secret key when none is configured, mirroring how `WC_Stripe_API::request()` let unkeyed requests go out and 401 — otherwise the SDK throws `api_key cannot be the empty string` at construction time, before any `pre_http_request` mock can fire.

### What's deliberately NOT in this PR

- **No callsite migrations.** All 62 in-repo `WC_Stripe_API::` callsites keep using the legacy helper. The follow-up resource PRs (customers, payment intents, charges/refunds, webhooks, terminal, admin REST) each migrate a coherent slice and accept the per-test reconciliation that goes with it.
- **No `WC_Stripe_API::request()` shim.** Turning it into a thin `WC_Stripe_Client::request()` shim is the *last* PR in the chain — once every callsite is off it.
- **`WC_Stripe_API_Test`** keeps testing the legacy implementation. It's the contract test for the deprecated path until that path goes away.

### Verification

- `npm run test:php` — full suite, **1927 tests / 4447 assertions, all pass**.
- `npm run phpstan` — clean.
- `npm run lint:php` — clean.

### Migration plan from here

Each follow-up PR migrates one resource family, re-running the full suite per PR:

1. **Customers** — `class-wc-stripe-customer.php` (most call density). Already has the parent PR's typed `->customers->retrieve()` example to lean on.
2. **Payment intents + setup intents** — `class-wc-stripe-intent-controller.php`, helpers, abstract gateway.
3. **Charges + refunds** — order handler, webhook handler, subscriptions trait.
4. **Webhook endpoints + admin REST** — account.php, webhook handler, REST controllers.
5. **Terminal** — locations + connection tokens controllers.
6. **Misc remaining** — checkout sessions, payment_method_configurations, voucher gateway.
7. **Cleanup** — turn `WC_Stripe_API::request()`, `::retrieve()`, `::request_with_level3_data()` into deprecation shims that delegate to `WC_Stripe_Client`.

## Testing instructions

This PR adds inert helpers — no in-repo callsite uses them yet, so production behavior is unchanged. Smoke test:

1. `npm run test:php` (full suite).
2. `npm run phpstan`.
3. Optional: in a dev install, drop a temporary `var_dump( WC_Stripe_Client::request( [], 'customers', 'GET' ) );` somewhere and confirm it returns an SDK object on success / a `stdClass` with `->error` on a missing-resource path.

---

-   [x] Tested on mobile (or does not apply) — admin/checkout-side refactor.

### Changelog entry

<details>

#### Comment

Internal infrastructure for the upcoming `WC_Stripe_API` migration; no user-visible behavior change. The user-visible entry already lives on #5403.

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-client.php
+++ b/includes/class-wc-stripe-client.php
@@ -42,6 +42,16 @@ class WC_Stripe_Client {
 	public static function get(): \Stripe\StripeClient {
 		$secret_key = WC_Stripe_API::get_secret_key();
 
+		// The SDK rejects an empty `api_key` at construction time, before the
+		// HTTP call ever happens. `WC_Stripe_API::request()` historically let
+		// unkeyed requests go out and 401 — which is what the invalid-key
+		// tracking in `WC_Stripe_Client::retrieve()` is designed to handle.
+		// Fall back to a placeholder so the SDK lets the request through and
+		// the WP HTTP layer (and any mocks) take over.
+		if ( '' === (string) $secret_key ) {
+			$secret_key = 'sk_test_placeholder';
+		}
+
 		if ( null === self::$client || self::$cached_secret_key !== $secret_key ) {
 			\Stripe\Stripe::setAppInfo(
 				'WooCommerce Stripe Gateway',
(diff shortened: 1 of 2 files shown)


wjrosa:
Postponing this work for now.

PR #5407, opened 2026-05-07
Detect Stripe API outages and surface an admin notice

wjrosa:
Fixes STRIPE-819

When Stripe's API is unreachable (network failure, timeout, or 5xx response), the plugin currently behaves badly:

- `WC_Stripe_API::request()` throws `WC_Stripe_Exception`, which can crash checkout if the caller doesn't catch it.
- `WC_Stripe_API::retrieve()` returns a generic `WP_Error('stripe_error')` and the wp-admin Stripe settings page silently shows empty/missing data — visually identical to a misconfigured account, with a misleading "we couldn't connect to your account, reconnect" notice telling merchants to rotate keys.

This PR adds outage classification at the API layer and surfaces a single, accurate admin notice so merchants understand what's happening and don't follow the wrong remediation steps.

**What's added**

- New `WC_Stripe_API_Outage_Status` helper that classifies a `wp_remote_*` response as an outage (`WP_Error` from network/timeout, or HTTP 500/502/503/504), records it in a transient (10 min TTL), and clears it on any non-outage response. 4xx is explicitly *not* an outage — Stripe is alive, just rejecting our request — so existing auth/validation error paths are unchanged.
- `WC_Stripe_API::request()` records/clears the outage flag and throws `WC_Stripe_Exception` with an outage-specific localized message when 5xx/network errors occur.
- `WC_Stripe_API::retrieve()` returns `WP_Error('stripe_api_outage', …)` (distinct from the existing `'stripe_error'` code) on outage, so callers can distinguish if needed.
- New `check_api_outage()` method in `WC_Stripe_Admin_Notices` renders a warning when the flag is set; the existing "couldn't connect to your account" notice is suppressed during an outage to avoid contradicting it.

**Why this approach**

A merchant during a Stripe outage today sees a red "your API keys may be invalid, reconnect" banner. Following that advice doesn't help (Stripe is down, not the keys) and risks them rotating credentials unnecessarily. An accurate outage notice is the minimum to avoid sending merchants down a wrong remediation path.

The classification is conservative on purpose: only signals that genuinely indicate "Stripe servers couldn't serve us" trip the flag. 4xx stays in the existing error handlers so plugin bugs and auth issues don't get hidden behind a generic "Stripe is down" message.

**How can this code break?**

- A short Stripe blip could flip the flag and confuse a merchant who reloads at the wrong moment. The 10-min transient TTL bounds the worst case, and any successful response clears it immediately.
- A non-Stripe outage (e.g. the merchant's host can't reach the internet) is also classified as outage. That's still accurate from the merchant's POV — Stripe is unreachable from this site — and the message ("temporarily unavailable, try again in a few minutes") doesn't blame Stripe specifically.
- The `wc_stripe_api_outage` transient name is new; no existing callers can read or write it.

**What's intentionally out of scope (follow-ups)**

- Wrapping additional checkout entry points that don't currently catch `WC_Stripe_Exception`. Existing throw behavior of `request()` is unchanged in this PR.
- Retry/backoff inside the API client.
- A React banner inside the Stripe settings UI (the wp-admin notice covers the same surface).

Notice preview:
Screenshot 2026-05-22 at 16 08 30

## Testing instructions

You'll simulate a Stripe outage with a small mu-plugin filter so you don't need to touch DNS or take Stripe down.

1. On a fresh test site with this plugin installed and connected to Stripe (test or live mode), confirm wp-admin shows no outage banner and the Stripe settings page (**WooCommerce → Settings → Payments → Stripe**) loads normally.
2. Drop the following file at `wp-content/mu-plugins/stripe-outage-sim.php` (create the `mu-plugins` folder if needed):

   ```php
   <?php
       if ( strpos( $url, 'api.stripe.com' ) === false ) {
           return $pre;
       }
       return [
           'response' => [ 'code' => 503, 'message' => 'Service Unavailable' ],
           'headers'  => [],
           'body'     => '{"error":"service unavailable"}',
       ];
   }, 10, 3 );
   ```
3. Reload the wp-admin dashboard. **Expected:** a yellow warning notice appears: *"Stripe is temporarily unavailable. We are unable to reach the Stripe API right now…"*. The Stripe settings page should load **without fatals**.
4. With the filter still active, confirm the previously-shown red **"Your customers cannot use Stripe on checkout, because we couldn't connect to your account"** notice is **not** displayed (it would be misleading).
5. Open a private/incognito window and visit the storefront checkout. The page should load without a white screen / fatal. (Submitting a payment will still fail because Stripe is "down" in the simulation — that's expected; we're only checking the page doesn't crash.)
6. Delete `wp-content/mu-plugins/stripe-outage-sim.php` and reload wp-admin. Within ~10 minutes the outage notice clears. Force-clearing is also fine: visit any wp-admin page and the next successful Stripe API call will clear it on the same request.
7. **Negative test (4xx is not an outage):** change the simulated `code` from `503` to `400` in the snippet, reload wp-admin. The outage notice should **not** appear (we treat 4xx as "Stripe is alive, just rejecting us").
8. **Network failure variant:** replace the `return [...]` line with `return new WP_Error( 'http_request_failed', 'Connection refused' );`, reload wp-admin. Expected: same outage notice as in step 3.

---

-   [x] Tested on mobile (or does not apply) — wp-admin notice only, no mobile-specific paths.

### Changelog entry

<details>

#### Comment

Changelog entry added in `changelog.txt` and `readme.txt` under 10.7.0:
> Add - Detect Stripe API outages (network failures, timeouts, 5xx responses) and surface a wp-admin notice instead of crashing or showing misleading "couldn't connect" messages

</details>

**Post merge**

cc @aheckler


includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -369,6 +375,30 @@ public function stripe_check_environment() {
...
+	 * @return void
+	 */
+	public function check_api_outage(): void {
+		if ( ! WC_Stripe_API_Outage_Status::is_in_outage() ) {
+			return;
+		}
+
+		$message = sprintf(
+			/* translators: 1) HTML strong open tag 2) HTML strong closing tag */
+			__( '%1$sStripe is temporarily unavailable.%2$s We are unable to reach the Stripe API right now. Payments and account details may not work until the service is restored. This notice will clear automatically once Stripe responds again.', 'woocommerce-gateway-stripe' ),
```
Mayisha: Nit suggestion:

**Stripe is temporarily unreachable.** Payments and account updates may not go through until the connection is restored. This notice will clear automatically once requests start succeeding again.

includes/class-wc-stripe-api-outage-status.php
```diff
@@ -0,0 +1,106 @@
...
+	 * Whether an outage is currently flagged.
+	 */
+	public static function is_in_outage(): bool {
+		return false !== get_transient( self::OUTAGE_TRANSIENT_KEY );
+	}
+
+	/**
+	 * Unix timestamp of when the current outage was first detected, or null.
+	 */
+	public static function get_outage_detected_at(): ?int {
```
Mayisha: This is currently unused.

includes/class-wc-stripe-api.php
```diff
@@ -337,6 +356,22 @@ public static function retrieve( $api ) {
 			]
 		);
 
+		if ( WC_Stripe_API_Outage_Status::is_outage_response( $response ) ) {
+			WC_Stripe_API_Outage_Status::record_outage();
+
+			$error_data = [
+				'stripe_api_key' => $masked_secret_key,
```
Mayisha: Is it necessary to log this masked key here?

Mayisha (approved):
Works as described. 
I see the admin notice when I add the `pre_http_request` filter in `mu_plugins` and clicked the `Refresh accoun data` option in the settings to trigger a Stripe API call. The notice disappeared when I removed the filter.

Image

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Adopted verbatim in c248aae0d — your wording is more accurate on both counts ("unreachable" vs "unavailable" and "requests start succeeding again" matches the actual recovery trigger). Thanks.

includes/class-wc-stripe-api-outage-status.php
wjrosa: Good catch — removed in c248aae0d, along with the two tests that were the only callers. The `detected_at` field still ships in the transient shape so a future "outage started X minutes ago" UI can read it back, but the dead public API goes.

includes/class-wc-stripe-api.php
wjrosa: No — stripped in c248aae0d. It was carry-over from the surrounding error paths; an outage is 5xx-level and Stripe-side, so the key identity adds no diagnostic value here.

PR #5408, opened 2026-05-07
Respect iDEAL/Bancontact save toggle when Optimized Checkout is on (STRIPE-1139)

wjrosa:
Fixes STRIPE-1139

With OCS on and global "Enable saved payment methods" on, checking the save box while paying with iDEAL/Wero or Bancontact added `setup_future_usage = off_session` even when the per-method save toggle was OFF — Stripe then generated and attached a SEPA Debit against the merchant's preference.

`prepare_payment_information_from_request()` computes the save flag against the OCS pseudo-method (`stripe` → `card`, always reusable) before resolving the real method (e.g. `ideal`) from the PaymentMethod API response. The frontend hides the checkbox container per sub-method via `showSaveOptionByMethod`, but the hide is CSS-only (`display: none` via a body class in `client/blocks/upe/styles.scss:81-83`) — the underlying `<input>` stays in the DOM and submits its `checked` state. So the backend is the only consistent enforcement point.

Fix:
1. In `prepare_payment_information_from_request()`: drop `save_payment_method_to_store` if `is_reusable()` is false on the resolved method.
2. In `handle_saving_payment_method()`: early-return on non-reusable methods. This is the single choke point for deferred-intent, confirmation-token, redirect-return, and the Adaptive Pricing webhook (which doesn't go through `prepare_payment_information_from_request()`).

The new null-check removes three pre-existing `method.nonObject` entries from `phpstan-baseline.neon`.

Downstream save paths already gated on `is_reusable()` / retrievable type, so the only observable change is that `setup_future_usage` is no longer added. Covered by data-provider tests for iDEAL/Bancontact (toggle on/off), a card regression, and a direct test of the choke-point gate.

## Testing instructions

Pre-reqs: test mode, OCS on, global save on, EUR store, NL (iDEAL) or BE (Bancontact) address.

**Repro note:** the save checkbox is only visible when OCS has `card` as the active sub-method. Switching the Payment Element to iDEAL/Bancontact CSS-hides the container but doesn't reset the input's `checked` state, so the bug needs a "check while card is active → switch sub-method → submit" sequence. Loading the page with iDEAL pre-selected won't reproduce it (and the frontend gating correctly hides the checkbox in that case — see step 6).

1. **Bug fix (iDEAL):** `Enable saved iDEAL | Wero payments for repeat payments` = OFF. On classic and Blocks, load OCS checkout, **check the save box while card is the active sub-method**, then switch the PE to iDEAL, then submit → no SEPA Debit on the Stripe customer or in My Account.
2. **Regression (iDEAL):** flip the per-method toggle ON, repeat the same sequence → SEPA Debit token saved as before.
3. Repeat 1 and 2 for Bancontact with a BE address.
4. **Card via OCS:** card order with save checked (no method switch) → token saved (unchanged).
5. **Adaptive Pricing:** OCS + AP on, iDEAL toggle OFF, iDEAL order → no token created. AP doesn't route through `prepare_payment_information_from_request()`, so this exercises the choke-point gate in `handle_saving_payment_method()`.
6. **UI:** per-method toggle OFF + iDEAL pre-selected inside OCS → save checkbox hidden from page load (unchanged frontend gating).

---

### Changelog entry

<details>

#### Comment

Changelog entry added to `changelog.txt` and `readme.txt`.

</details>

**Post merge**


malithsen (commented):
I might be misunderstanding the issue. I am not able to replicate this on `develop`

> Bug fix: Enable saved iDEAL | Wero payments for repeat payments = OFF. Place an iDEAL order on classic and Blocks with the save box checked → no SEPA Debit on the Stripe customer or in My Account.

When `Enable saved iDEAL | Wero payments for repeat payments = OFF`, I am not able to `Place an iDEAL order on classic and Blocks with the save box checked` because the checkbox never shows up

Image

I only see the save payment method checkbox when the "Enable saved iDEAL" option is enabled in settings.

wjrosa:
@malithsen good catch — the original step 1 was missing the trigger sequence. The checkbox is only visible while OCS has `card` as the active sub-method; switching the Payment Element to iDEAL toggles the `wc-stripe-hide-save-checkbox` body class in [`handle-display-of-saving-checkbox.js:81-87`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/fix/ocs-respect-ideal-bancontact-save-toggle/client/optimized-checkout/handle-display-of-saving-checkbox.js#L81-L87), and the matching rule in [`blocks/upe/styles.scss:81-83`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/fix/ocs-respect-ideal-bancontact-save-toggle/client/blocks/upe/styles.scss#L81-L83) hides the container with `display: none` — but it doesn't reset the underlying `<input>`'s `checked` state. So the repro is:

1. Load OCS checkout (card is the default sub-method, save checkbox visible).
2. **Check the save box.**
3. Switch the PE to iDEAL — checkbox CSS-hidden, but `checked = true` still posts.
4. Submit.

`prepare_payment_information_from_request()` then computes the save flag against the OCS pseudo-method (resolves to `card`, reusable), only learns it's iDEAL from the PaymentMethod API response, and without the new `is_reusable()` re-check at [`class-wc-stripe-upe-payment-gateway.php:3299-3306`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/fix/ocs-respect-ideal-bancontact-save-toggle/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L3299-L3306) keeps `setup_future_usage = off_session` on the intent. Stripe then creates the SEPA Debit.

Your screenshot is exactly what step 6 documents — pre-selecting iDEAL hides the checkbox from frame 0, which is the frontend gating working correctly. The bug only manifests via the check-then-switch sequence, where CSS hiding ≠ unmount.

Updated the PR body's testing instructions to spell this out, including the explicit "check while card is active → switch sub-method → submit" wording. Should reproduce now. Also worth noting: the choke-point gate in `handle_saving_payment_method()` covers the Adaptive Pricing path that bypasses `prepare_payment_information_from_request()` entirely — so even if the UI gating were airtight today, that gate is the right place to enforce the contract long-term.

malithsen (approved):
Thanks for the updated test instructions. I can confirm the issue in `develop`. In this branch `setup_future_usage` is correctly set to `null` when the frontend checkbox is not set.

I think it's also worth addressing this issue in the frontend, in addition to what's done here. The root cause is that the frontend is not clearing the toggle state properly, resulting in a `wc-stripe-new-payment-method: true` parameter in checkout requests when it should not be present. But that can be a separate follow-up.

wjrosa:
@malithsen good call — filed the frontend follow-up as #5461. It clears the checkbox state from a single helper called by both OC paths: `.checked = false` on the classic input + `__internalSetShouldSavePaymentMethod(false)` dispatch on the Blocks payment store. Branched off develop so either PR can ship independently.

PR #5409, opened 2026-05-07
Honor per-method save toggles when OCS is enabled (STRIPE-1139)

wjrosa:
## Summary

Fixes STRIPE-1139: when Optimized Checkout Suite is enabled, the per-method "Save iDEAL | Wero for repeat use" toggle (`sepa_tokens_for_ideal`, and the equivalent for Bancontact) was bypassed. Customers ended up with their iDEAL/Wero method saved/reusable on Stripe (and surfaced on subsequent checkouts) despite the merchant having disabled the toggle.

### Root cause
- `WC_Stripe_Payment_Methods::OC = 'card'` and `WC_Stripe_UPE_Payment_Method_OC` is hard-coded `is_reusable = true`. So under OCS, `should_save_payment_method_from_request()` evaluates against the OC pseudo-method (always reusable), and the resulting `save_payment_method_to_store` flag flows downstream regardless of the actual underlying method's per-method save toggle.
- `WC_Stripe_Intent_Controller::build_base_payment_intent_request_params()` then set `setup_future_usage = 'off_session'` on the PaymentIntent without re-checking the resolved method's reusability — so Stripe stored the SEPA debit even when the merchant had disabled the iDEAL toggle.
- The Adaptive Pricing / Checkout Sessions path additionally persisted a WC payment token via the `checkout.session.completed` webhook handler → `handle_saving_payment_method()`, again with no per-method gate.

### Fix
Two narrow gates added at the actual save decision points; both rely on the already-resolved actual payment method instance.

- **`includes/class-wc-stripe-intent-controller.php`** — `build_base_payment_intent_request_params()` now also gates `setup_future_usage` on the resolved method's `is_reusable()`, so Stripe is never asked to save a non-reusable method (iDEAL/Wero with the toggle off).
- **`includes/payment-methods/class-wc-stripe-upe-payment-gateway.php`** — `handle_saving_payment_method()` early-returns when the resolved method is non-reusable. This is the single choke point all save paths funnel through (deferred-intent, confirmation token, redirect-return, Adaptive Pricing webhook), so this enforces the per-method toggle end-to-end. The webhook handler delegates to this gate rather than duplicating the logic.
- **`phpstan-baseline.neon`** — three pre-existing `method.nonObject` baseline entries became unmatched because the new null-check in `handle_saving_payment_method()` and the extracted `$is_reusable` local in the intent controller resolve them. Removed/decremented those entries (no new baseline rows added).

## Test plan

Automated:
- [x] `WC_Stripe_Intent_Controller_Test::test_setup_future_usage_respects_per_method_toggle` — covers `sepa_tokens_for_ideal=no` (no `setup_future_usage` on the intent) and `sepa_tokens_for_ideal=yes` (intent opts into off-session).
- [x] `WC_Stripe_UPE_Payment_Gateway_Test::test_handle_saving_payment_method_respects_per_method_toggle` — covers OC-enabled save attempt against an iDEAL payment method object; asserts the `woocommerce_stripe_add_payment_method` action only fires when the toggle is on.
- [x] `npm run phpstan` clean.
- [x] `phpcs` clean on changed files.
- [x] Pre-existing checkout/save tests still pass on the gateway test file.

Manual (recommended before merging):
- [ ] OCS + Blocks checkout: with `Enable saved payment methods` ON and `Enable saved iDEAL | Wero payments for repeat payments` OFF, place an iDEAL order and verify that no saved iDEAL/SEPA token surfaces in WC and that the Stripe customer in dashboard does not list a saved SEPA debit.
- [ ] OCS + classic shortcode checkout: same as above.
- [ ] OCS + Adaptive Pricing (Checkout Session redirect path): same as above.
- [ ] Same flows with the iDEAL toggle ON should continue to save and reuse the SEPA debit.
- [ ] Sanity on Bancontact (uses the same SEPA tokenization path / `sepa_tokens_for_bancontact` toggle).

## Notes

- The frontend `wc-stripe-new-payment-method` checkbox state captured at Element init is the upstream source of the leak (it doesn't get unset when the customer switches to a non-reusable method inside the Payment Element). The server-side gates here are the authoritative fix; a follow-up could also clear `setupFutureUsage` client-side when the customer switches methods, to avoid even sending a save request for non-reusable selections.
- Issue tracker reference: `STRIPE-1139`. (No corresponding GitHub issue # was provided, so the PR doesn't auto-close one.)

### DIFF
--- a/includes/class-wc-stripe-intent-controller.php
+++ b/includes/class-wc-stripe-intent-controller.php
@@ -1137,8 +1137,13 @@ class WC_Stripe_Intent_Controller {
 		// Only exceptions are when using a confirmation token or manual renewal is required.
 		// For confirmations tokens, the setup_future_usage is set within the payment method.
 		$payment_method                 = WC_Stripe_UPE_Payment_Gateway::get_payment_method_instance( $selected_payment_type );
-		$has_auto_renewing_subscription = ! empty( $payment_information['has_subscription'] ) && ! $this->is_manual_renewal_required( $payment_method->is_reusable() );
-		if ( ! $is_using_confirmation_token && ( $payment_information['save_payment_method_to_store'] || $has_auto_renewing_subscription ) ) {
+		$is_reusable                    = $payment_method && $payment_method->is_reusable();
+		$has_auto_renewing_subscription = ! empty( $payment_information['has_subscription'] ) && ! $this->is_manual_renewal_required( $is_reusable );
+		// When OC is enabled, $payment_information['save_payment_method_to_store'] reflects the OC pseudo-method's
+		// reusability rather than the actual method the customer selected (e.g. iDEAL/Bancontact). Re-gate the
+		// `setup_future_usage` flag on the resolved method's `is_reusable()` so per-method save toggles such as
+		// `sepa_tokens_for_ideal` are honored end-to-end.
+		if ( ! $is_using_confirmation_token && $is_reusable && ( $payment_information['save_payment_method_to_store'] || $has_auto_renewing_subscription ) ) {
 			$request['setup_future_usage'] = 'off_session';
 		}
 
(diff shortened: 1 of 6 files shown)


wjrosa:
Closing as duplicate of #5408 (which was opened ~30 minutes earlier and addresses the same STRIPE-1139 bug). The defense-in-depth choke-point gate in `handle_saving_payment_method()` and the phpstan baseline cleanup from this PR have been folded into #5408.

PR #5410, opened 2026-05-08
fix: ECE compatibility for optimized MCCY solutions

frosso:
Fixes WOOPMNT-6137

When the shopper's currency is resolved on the client _after_ the page renders (country-based pricing like Price Based on Country, or a cache-optimized multi-currency setup), the Express Checkout Element can get initialized in the store's base currency rather than the visitor's.

A few changes to handle that:

- ECE now waits for the resolved currency before creating the Stripe Element, via a `wc-stripe.express-checkout.resolved-currency` filter. The Price Based on Country compat layer hooks into it, and other plugins can do the same.
- A server-side guard rejects the Store API checkout when the order currency doesn't match the one the Element was initialized with, so a mismatch fails cleanly instead of charging in the wrong currency.
- Amazon Pay is re-checked against the resolved currency (it only supports a fixed set per account), so it's dropped when the currency isn't supported instead of making Stripe reject the whole request, and kept when it is.
- If a multi-currency plugin flips the cart currency from the address picked inside an open Apple Pay / Google Pay sheet, the address is rejected up front with a clear message, rather than failing later at placement.

This stays Stripe-gateway specific. I didn't port WooPayments' own multi-currency adapter, since I don't think the two plugins are expected to run side by side.

## Testing instructions

You'll need a US-based Stripe account (Amazon Pay only supports USD there) and the [Price Based on Country](https://wordpress.org/plugins/woocommerce-product-price-based-on-countries/) plugin.

1. In your store, ensure that you have configured geolocation
Screenshot 2026-07-01 at 5 19 09 PM
Screenshot 2026-07-01 at 5 20 27 PM

2. Use an extension ([example](https://chromewebstore.google.com/detail/simple-modify-headers/gjgiipmpldkpbdfjkgofildhapegmmic)) to modify the headers sent to your store - ensure you're sending `X-Country-Code: DE`
Screenshot 2026-07-03 at 6 28 18 PM

3. In Price Based on Country, keep the base currency as USD and add EUR for an EU zone, including Germany
Screenshot 2026-07-01 at 5 10 17 PM

4. In Stripe settings, enable Express Checkout (Apple Pay / Google Pay) and Amazon Pay
Screenshot 2026-07-01 at 5 12 01 PM

5. In a fresh session (incognito works), open a product page as an EU visitor so prices show in EUR
6. Confirm the Apple Pay / Google Pay button shows, and Amazon Pay does _not_ (USD-only account)
Screenshot 2026-07-01 at 5 37 19 PM

6. Pay with Apple Pay / Google Pay, with a German address
7. Confirm the order is created in EUR end to end
8. Inside the wallet sheet, pick a shipping address in a country that maps to a different currency (e.g.: the US address), and confirm the address is rejected
9. Switch the product back to USD and confirm the button still works and Amazon Pay is back

---

### Changelog entry

fix: Express Checkout now boots in the shopper's resolved currency on country-based / multi-currency stores, and rejects currency mismatches instead of charging in the wrong currency.

**Post merge**

---


client/express-checkout/utils/compute-product-page-start-args.js
```diff
@@ -0,0 +1,77 @@
...
+		buttonContext: 'product',
+	} );
+
+	const resolvedCurrency = getResolvedCurrency( localizedCurrency );
+	const hasCurrencyChanged = resolvedCurrency !== localizedCurrency;
+
+	let total = getExpressCheckoutData( 'product' )?.total.amount;
+	let displayItems = getExpressCheckoutData( 'product' ).displayItems ?? [];
+	let requestShipping =
+		getExpressCheckoutData( 'product' )?.requestShipping ?? false;
```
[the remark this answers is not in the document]
frosso: Fixed in `4b0c1c4a1`.

client/entrypoints/express-checkout/index.js
```diff
@@ -968,25 +1018,29 @@ jQuery( function ( $ ) {
...
 										wcStripeECE.refreshTotals( response );
 										wcStripeECE.updateExpressCheckoutAmount(
 											response.total.amount
 										);
 									} else {
-										wcStripeECE.reInitExpressCheckoutElement(
+										await wcStripeECE.reInitExpressCheckoutElement(
 											response
 										);
 									}
```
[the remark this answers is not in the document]
frosso: Fixed in `4b0c1c4a1`.

client/blocks/express-checkout/express-checkout-container.js
```diff
@@ -8,6 +8,7 @@ import {
 	getPaymentMethodTypesForExpressMethod,
 	isManualPaymentMethodCreation,
 } from 'wcstripe/express-checkout/utils';
+import { setElementCurrency } from 'wcstripe/express-checkout/utils/element-currency-cache';
```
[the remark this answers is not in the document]
frosso: Fixed in `4b0c1c4a1`.

client/express-checkout/compatibility/wcpbc-currency.js
```diff
@@ -0,0 +1,56 @@
...
+		);
+		cleanups.push( () => clearTimeout( retriggerTimer ) );
+
+		// Hard watchdog: surrender rather than hang ECE forever.
+		const bailTimer = setTimeout(
+			async () => finish( await upstream ),
+			BAIL_AFTER_MS
+		);
+		cleanups.push( () => clearTimeout( bailTimer ) );
+	} );
```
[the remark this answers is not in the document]
frosso: Fixed in `4b0c1c4a1`.

client/api/index.js
```diff
@@ -625,6 +626,9 @@ export default class WCStripeAPI {
 				'X-WCSTRIPE-EXPRESS-CHECKOUT-NONCE':
 					getExpressCheckoutData( 'nonce' )
 						?.wc_store_api_express_checkout,
+				...( getElementCurrency() && {
+					'X-WCSTRIPE-PAYMENT-CURRENCY': getElementCurrency(),
+				} ),
```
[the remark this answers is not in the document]
frosso: Decided not to pursue.

client/express-checkout/compatibility/__tests__/wcpbc-currency.test.js
```diff
@@ -0,0 +1,151 @@
...
+
+	test( 'hard watchdog falls back to the upstream value', async () => {
+		global.wc_price_based_country_ajax_geo_params = {};
+		const piped = applyFilters( FILTER, Promise.resolve( 'usd' ), {} );
+
+		advanceTimers( 6000 );
+		await Promise.resolve();
+
+		await expect( piped ).resolves.toBe( 'usd' );
+	} );
```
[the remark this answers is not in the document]
frosso: Decided not to pursue.

client/express-checkout/utils/__tests__/compute-product-page-start-args.test.js
```diff
@@ -0,0 +1,174 @@
...
+			getSelectedProductData: jest.fn(),
+			transformLabeledDisplayItems: passthroughTransform,
+			useLegacyDisplayItems: false,
+		};
+
+		const result = await computeProductPageStartArgs( deps );
+
+		expect( result.requestPhone ).toBe( true );
+	} );
+} );
```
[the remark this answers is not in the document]
frosso: Decided not to pursue.

client/express-checkout/utils/resolved-currency-cache.js
```diff
@@ -0,0 +1,11 @@
...
+
+export const setResolvedCurrency = ( currency ) => {
+	resolvedCurrency = currency || null;
+};
+
+export const getResolvedCurrency = ( fallback ) => resolvedCurrency || fallback;
+
+export const __resetResolvedCurrencyForTests = () => {
+	resolvedCurrency = null;
+};
```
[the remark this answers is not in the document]
frosso: Decided not to pursue.

client/express-checkout/utils/compute-product-page-start-args.js
```diff
@@ -0,0 +1,77 @@
...
+	// let any registered resolver (e.g. WCPBC) settle the visitor's currency
+	// before we create the Stripe Element. fast-path when no resolver participates.
+	await resolveExpressCheckoutCurrency( localizedCurrency, {
+		buttonContext: 'product',
+	} );
+
+	const resolvedCurrency = getResolvedCurrency( localizedCurrency );
+	const hasCurrencyChanged = resolvedCurrency !== localizedCurrency;
+
+	let total = getExpressCheckoutData( 'product' )?.total.amount;
```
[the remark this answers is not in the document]
frosso: Alright, fine, fixed in 43df7e505

changelog.txt
```diff
@@ -55,6 +55,7 @@ xxxx-xx-xx - version 10.9.0
 * Fix - Prevent an error when creating the Adaptive Pricing checkout session for logged-in customers without a saved billing address
 * Tweak - Break down the Agentic Commerce feed preview's excluded product count by reason so subscriptions are no longer mistaken for a merchant-configured filter
 * Fix - Render the Express Checkout settings button preview background based on the button color
+* Update - fix: Express Checkout with optimized MCCY plugins
```
Mayisha: Nit. Can we make the statement a bit detailed. One suggestion is-
```suggestion
* Fix - Ensure Express Checkout charges in the shopper's currency on multi-currency stores
```

client/blocks/express-checkout/express-checkout-container.js
```diff
@@ -19,6 +20,12 @@ export const ExpressCheckoutContainer = ( props ) => {
 	);
 	const currency = billing.currency.code.toLowerCase();
 
+	// Record the currency the Element was created with only after commit, so the
+	// server-side mismatch guard reads the currency that's actually on screen.
+	useEffect( () => {
+		setElementCurrency( currency );
```
Mayisha, line 26: I didn't came accross any error scenarios but a thought reading the code: should we freeze the currency at ECE sheet open instead of following `billing.currency`?

client/express-checkout/utils/compute-product-page-start-args.js
```diff
@@ -0,0 +1,77 @@
...
+		getExpressCheckoutData( 'product' )?.currency || ''
+	).toLowerCase();
+
+	// let any registered resolver (e.g. WCPBC) settle the visitor's currency
+	// before we create the Stripe Element. fast-path when no resolver participates.
+	await resolveExpressCheckoutCurrency( localizedCurrency, {
+		buttonContext: 'product',
+	} );
+
+	const resolvedCurrency = getResolvedCurrency( localizedCurrency );
```
Mayisha: Nit: `await resolveExpressCheckoutCurrency` also returns the resolvedCurrency.

client/express-checkout/utils/compute-product-page-start-args.js
```diff
@@ -0,0 +1,77 @@
...
+			}
+		} catch ( e ) {
+			// fall back to the localized data. ECE may render in the wrong
+			// currency, but it won't fail silently at confirmation.
+		}
+	}
+
+	return {
+		total,
+		currency: resolvedCurrency,
```
Mayisha, line 93: AI feedback:

On a currency change it does `total = fresh.total.amount` and returns `{ total, currency: resolvedCurrency }` — but never checks that fresh actually came back in resolvedCurrency. The correctness of the amount depends on WCPBC having persisted the zone server-side (cookie) before `getSelectedProductData()` runs (line 55). If the zone isn't persisted yet, the AJAX returns base-currency totals while the label says the resolved currency → wrong amount on screen, caught only by the server guard at confirm. Worth either comparing `fresh.currency` to `resolvedCurrency` (bail/keep-localized on mismatch) or at minimum documenting the "zone must be persisted before this re-fetch" dependency.

client/express-checkout/compatibility/wcpbc-currency.js
```diff
@@ -0,0 +1,61 @@
...
+
+addFilter(
+	'wc-stripe.express-checkout.resolved-currency',
+	'automattic/wc-stripe/express-checkout/wcpbc',
+	( upstream ) =>
+		isWCPBCAjaxModeActive() ? waitForWCPBCCurrency( upstream ) : upstream
+);
```
Mayisha, line 72: QQ: Is this compatible with any version of the WCPBC plugin? If not, it would be good to mention the minimum compatible version in a code comment.

Mayisha (approved):
Thanks @frosso. This looks good to me and includes a solid coverage for the new additions. 

I tested Link and Google Pay with the [Price Based on Country](https://wordpress.org/plugins/woocommerce-product-price-based-on-countries/) plugin on product page; the behavior was correct based on the resolved currency.

Regression tests also went well. Tested with Google Pay on product,cart and checkout pages without the Price Based on Country plugin.

The changelog entry needs an update and I have a couple of nit suggestion/questions.

changelog.txt
frosso: Sure - reworded 👍

client/express-checkout/utils/compute-product-page-start-args.js
frosso: Thank you - done 👍

client/express-checkout/utils/compute-product-page-start-args.js
frosso: Guard added 👍

client/express-checkout/compatibility/wcpbc-currency.js
frosso: It's been a while - I had Claude find it's been available since 2020

client/blocks/express-checkout/express-checkout-container.js
frosso: My thought here was to keep track of the currency used to create the element. I'm not sure I'm following - why keeping track of the currency when the ECE payment sheet opens instead?

Mayisha: > My thought here was to keep track of the currency used to create the element

That makes sense. I was reading the event wrongly and thought it would change when the billing details are change on the ECE modal.

client/api/index.js
daledupreez: Not blocking, but I prefer stashing the result in a variable so we can be confident that we will only run the function code once. While it's likely negligible impact, the implementation may change in future, and this is a subtle, implicit expectation.
```suggestion
				// Earlier in function
				const elementCurrency = getElementCurrency();
				// ...
				...( elementCurrency && {
					'X-WCSTRIPE-PAYMENT-CURRENCY': elementCurrency,
				} ),
```

client/api/index.js
```diff
@@ -643,7 +647,15 @@ export default class WCStripeAPI {
...
 		const url = `/wc/store/v1/checkout/${ order }?key=${ key }&billing_email=${ billingEmail }`;
-		return this.postToStoreApi( url, paymentData );
+		return this.postToStoreApi( url, paymentData, {
+			'X-WCSTRIPE-EXPRESS-CHECKOUT': true,
+			'X-WCSTRIPE-EXPRESS-CHECKOUT-NONCE':
+				getExpressCheckoutData( 'nonce' )
+					?.wc_store_api_express_checkout,
+			...( getElementCurrency() && {
+				'X-WCSTRIPE-PAYMENT-CURRENCY': getElementCurrency(),
+			} ),
```
daledupreez: As above RE using a local variable.

client/entrypoints/express-checkout/index.js
```diff
@@ -245,6 +250,25 @@ jQuery( function ( $ ) {
...
+			).toLowerCase();
+
+			// is_amazon_pay_enabled() is currency-aware server-side, but on product
+			// pages it reflects the rendered base currency, not the one resolved
+			// after render. When the currency changed, re-check it against the
+			// supported set: a currency Amazon Pay can't take would make Stripe
+			// reject the whole element, while a supported one keeps it offered.
+			// Check the currency the element is actually built with (options.currency)
+			// rather than the resolved-currency global, which an overlapping init()
+			// could overwrite between here and element creation.
```
daledupreez: I think we can simplify this comment.
```suggestion
			// Add a client-side check to make sure the resolved customer-facing currency
			// is actually supported by Amazon Pay, as it may change in the client.
			// Note that we check `options.currency` to make sure we avoid race conditions
			// involving updates to global values that may not have propagated yet.
```

client/express-checkout/compatibility/wcpbc-currency.js
```diff
@@ -0,0 +1,65 @@
...
```
daledupreez, line 72: Is it worth running the check before adding the filter? As written, this filter will _always_ be hooked in, which feels somewhat unnecessary.

client/express-checkout/utils/compute-product-page-start-args.js
```diff
@@ -0,0 +1,91 @@
+/**
+ * Builds the args object passed to `startExpressCheckout` on a shortcode
+ * product page. Extracted from the entrypoint so the resolver-aware start path
+ * is testable without standing up the entire jQuery/Stripe.js shell.
```
daledupreez: Nit: I think we can simplify the comment here.
```suggestion
 * product page. Extracted into a dedicated file to support testing.
```

client/express-checkout/utils/compute-product-page-start-args.js
```diff
@@ -0,0 +1,91 @@
...
+			// under the resolved-currency label; the shopper falls back to the
+			// regular checkout.
+			return null;
+		}
+
+		// only trust the re-fetched payload when the server actually returned
+		// amounts in the resolved currency. an error, a missing total, or a
+		// currency mismatch (e.g. the zone wasn't persisted before this call)
+		// would mean a base-currency amount shown as the resolved currency, so
+		// bail instead of misleading the shopper.
```
daledupreez: Nit: Please review the comments and ensure they have correct capitalisation.

client/stripe-utils/utils.js
```diff
@@ -245,6 +245,24 @@ export const isAmazonPayEnabled = () => {
...
+ *
+ * @param {string} currency ISO currency code (case-insensitive).
+ * @return {boolean} True if Amazon Pay supports the currency; false otherwise.
+ */
+export const isAmazonPaySupportedForCurrency = ( currency ) => {
+	const supported =
+		// eslint-disable-next-line camelcase, no-undef
+		wc_stripe_express_checkout_params?.stripe
+			?.amazon_pay_supported_currencies ?? [];
+	return supported.includes( ( currency || '' ).toLowerCase() );
```
daledupreez: What is the right way to handle `currency === ''`? At a minimum, it feels like we should return early for that case, as that should be a short-circuit rather than checking whether the value is in the array.

includes/payment-methods/class-wc-stripe-express-checkout-currency-guard.php
```diff
@@ -0,0 +1,108 @@
...
+	/**
+	 * Hook the assertion onto Store API checkout order builds.
+	 *
+	 * @return void
+	 */
+	public function init() {
+		add_action(
+			'woocommerce_store_api_checkout_update_order_from_request',
+			[ $this, 'assert_currency_matches_element' ],
+			10,
```
daledupreez: Is it worth attaching using a later priority to ensure we're running after any hooks that use the default priority of 10?
```suggestion
			20,
```

includes/payment-methods/class-wc-stripe-express-checkout-currency-guard.php
```diff
@@ -0,0 +1,108 @@
...
+	 * non-ECE caller).
+	 *
+	 * @param WC_Order                              $order   The order being created.
+	 * @param WP_REST_Request<array<string, mixed>> $request The Store API request (unused).
+	 *
+	 * @return void
+	 *
+	 * @throws RouteException When the currencies disagree.
+	 */
+	public function assert_currency_matches_element( $order, $request ) {
```
daledupreez: Nit: Given that this function may or may not assert depending on context, and it only applies to express checkout, it may be worth making that more explicit.
```suggestion
	public function maybe_assert_order_currency_matches_express_checkout_currency( $order, $request ) {
```

includes/payment-methods/class-wc-stripe-express-checkout-currency-guard.php
```diff
@@ -0,0 +1,108 @@
...
+				'order_currency'   => $actual,
+			]
+		);
+
+		throw new RouteException(
+			'wc_stripe_express_checkout_currency_mismatch',
+			sprintf(
+				/* translators: 1: expected currency code, 2: actual currency code */
+				__(
+					'Your shipping address requires a different currency (%2$s) than this payment was started in (%1$s). You haven\'t been charged. Please reload the page and try again.',
```
daledupreez: Nit: I prefer using `"` quotes to ensure the full text string is searchable without having to worry about escaping, or for downstream translators to deal with the `\`.
```suggestion
					"Your shipping address requires a different currency (%2$s) than this payment was started in (%1$s). You haven't been charged. Please reload the page and try again.",
```

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -55,6 +55,9 @@ public function __construct( WC_Stripe_Express_Checkout_Ajax_Handler $express_ch
 		$this->express_checkout_helper       = $express_checkout_helper;
 		$this->express_checkout_ajax_handler = $express_checkout_ajax_handler;
 		$this->express_checkout_ajax_handler->init();
+
+		$currency_guard = new WC_Stripe_Express_Checkout_Currency_Guard( $this->express_checkout_helper );
+		$currency_guard->init();
```
daledupreez: Nit: We're generally trying to move away from side-effectful constructors. Can we move this logic outside of a constructor? Maybe into `WC_Stripe->init_express_checkout()`?

client/express-checkout/utils/resolved-currency-cache.js
```diff
@@ -0,0 +1,11 @@
+let resolvedCurrency = null;
+
+export const setResolvedCurrency = ( currency ) => {
+	resolvedCurrency = currency || null;
+};
+
+export const getResolvedCurrency = ( fallback ) => resolvedCurrency || fallback;
```
daledupreez: This function isn't used from any real code, only from tests. Is the cache even needed if we're not getting the value from the cache?

client/express-checkout/utils/resolve-currency.js
```diff
@@ -0,0 +1,32 @@
...
+		Promise.resolve( fallbackLower ),
+		ctx
+	);
+
+	let resolved = fallbackLower;
+	try {
+		const value = await piped;
+		if ( typeof value === 'string' && value ) {
+			resolved = value.toLowerCase();
+		}
```
daledupreez: There is no guarantee that the filter will return in any reasonable amount of time. Can we add a Promise.race() or similar to ensure we guard against long-running code?

client/express-checkout/utils/resolve-currency.js
```diff
@@ -0,0 +1,32 @@
...
+ * @param {Object} ctx      Passed through to filter callbacks.
+ * @return {Promise<string>} Lowercase ISO currency.
+ */
+export async function resolveExpressCheckoutCurrency( fallback, ctx ) {
+	const fallbackLower = ( fallback || '' ).toLowerCase();
+	const piped = applyFilters(
+		'wc-stripe.express-checkout.resolved-currency',
+		Promise.resolve( fallbackLower ),
+		ctx
+	);
```
daledupreez: Can we move this logic into the `try` block below as a protective measure? If this runs synchronously, the code could throw and not be caught.

client/express-checkout/compatibility/wcpbc-currency.js
```diff
@@ -0,0 +1,65 @@
+/* global wc_price_based_country_ajax_geo_params */
+
+import { addFilter } from '@wordpress/hooks';
+
+const RETRIGGER_AFTER_MS = 3000;
+const BAIL_AFTER_MS = 6000;
```
daledupreez, line 13: Should we consider shorter time windows here? We should document why these values are so long while we're at it. It's not clear why we chose 3s and 6s.

daledupreez (commented):
I have some light feedback from code inspection, but I also had an AI agent run some deeper checks, and it flagged the following findings (which may or may not be 100% valid):
* It seems that the Pay for Order context may now send in the express checkout request headers, and that enables additional server-side behaviour. We should verify that flow works for this PR. (I believe that path also checks the order currency, as we have that available at that point in time.)
* The server side hooks are in place for all express checkout calls, including the shortcode cart and checkout. We need to verify that express checkouts still work from the shortcode cart and checkout pages.
* We need to check whether the Amazon Pay logic runs correctly from `reInitExpressCheckoutElement()` - that code is writing `getExpressCheckoutData('product').currency = response.currency` before calling `init()`
* The filter name -- our other filters use `wcstripe`, not `wc-stripe`.
* There may be race conditions introduced by making `init()` async. In particular, re-initialisation may be triggered by variation selection or quantity changes

Not necessarily for you to focus on, but it feels like we should also keep the changes from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5709 in mind, as that PR also tackles express checkout initialisation.

client/express-checkout/utils/compute-product-page-start-args.js
```diff
@@ -0,0 +1,91 @@
...
+		if (
+			! fresh ||
+			fresh.error ||
+			fresh.currency !== resolvedCurrency ||
+			fresh.total?.amount === undefined
+		) {
+			return null;
```
wjrosa, line 83: When this returns null, the express buttons just don't render (no warning or log). Could we log the reason on each bail?

client/express-checkout/utils/compute-product-page-start-args.js
frosso: Done in b8635db13.

client/express-checkout/utils/compute-product-page-start-args.js
frosso: Ok, done in 17078efca.

client/stripe-utils/utils.js
frosso: Good point, done in 669d11016.

includes/payment-methods/class-wc-stripe-express-checkout-currency-guard.php
frosso: Good call, done in a5d7bbf35.

includes/payment-methods/class-wc-stripe-express-checkout-currency-guard.php
frosso: Ok, done in e5cc1fe8f.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
frosso: Agreed, done in a2df518ac.

client/express-checkout/utils/resolved-currency-cache.js
frosso: Thank you, good catch. Done in 2b7849c3b.

client/express-checkout/utils/resolve-currency.js
frosso: Thank you, done in 47f5adc4c.

client/express-checkout/utils/resolve-currency.js
frosso: Thank you, done in 4a8e45a37.

client/express-checkout/utils/compute-product-page-start-args.js
frosso: Good idea, done in 307e71381.

client/api/index.js
frosso: Yeah, although I feel like that's kinda speculative, and `getElementCurrency()` is already returning just a scoped variable with no additional logic - done in 52c916c9b

client/express-checkout/compatibility/wcpbc-currency.js
frosso: The problem is that `wc_price_based_country_ajax_geo_params` might not be defined when our JS runs (it might be defined later), so that would mean that once our code executes, it won't be possible to re-register this hook.

client/entrypoints/express-checkout/index.js
frosso: Thank you - done in 94e7142bc

includes/payment-methods/class-wc-stripe-express-checkout-currency-guard.php
frosso: The problem is that once we switch to double quotes, `$` becomes syntax for variable interpolation. I understand your point, tho, thank you for the callout. I changed `haven't` to `have not` in 085859ce0

client/express-checkout/compatibility/wcpbc-currency.js
frosso: Sure thing - done in e7900a962. FWIW, the timeouts will apply only on sites where WCPBC is enabled.

frosso:
@daledupreez @wjrosa thank you for taking the time with the reviews so far. Between the SIRTs/releases/etc, I haven't been able to give this PR as much attention as it deserved last week, so I haven't been able to re-test after the changes. I'll be AFK for the next couple of weeks, I'll re-test after I'm back. Thank you for your patience!

EDIT: I'm pausing this for mise-en-place in WC Core.

PR #5412, opened 2026-05-08
 [RSM - Diagnostics Mode] Render traces inline with toolbar and per-row actions

malithsen:
Fixes RSM-2301

This PR reworks **Advanced settings → Checkout diagnostics**

  - Splits the section into three cards: Debug mode, Checkout diagnostics, OC.
  - Replaces the diagnostics checkbox with a `ToggleControl`
  - Renders stored traces inline
  - Stored traces stay accessible when recording is off - only the indicator
    changes.
  - Adds `DELETE /wc/v3/wc_stripe/diagnostics/traces` to back the Clear action

Before
CleanShot 2026-05-08 at 09 24 24@2x

Now
CleanShot 2026-05-08 at 21 00 04@2x

## Testing instructions

1. Visit **WooCommerce → Settings → Payments → Stripe → Advanced settings**.
   Confirm two separate cards (Debug mode, Checkout diagnostics).
2. Toggle "Capture checkout diagnostics" -  indicator switches between pulsing red "Recording" and hollow grey "Not recording".
3. With recording on, drive a checkout failure (Stripe test card `4000 0000 0000 0002`). A row appears with a status dot, id, summary, duration.
4. Toolbar **Copy all** -> clipboard gets the JSON; success toast appears.
5. Toolbar **Clear** -> confirmation modal -> confirm -> list empties.
6. Toggle filter to **Failed only** and back to **All** - list filters in place.
7. Per-row **Copy** copies only that trace; per-row **View** opens a modal with pretty-printed JSON.
8. Flip the toggle off with traces stored: list, filter, Copy, Clear, per-row actions all remain functional.

---

### Changelog entry

<details>

#### Comment 

Unreleased feature, merging to a feature branch

</details>

**Post merge**


client/settings/advanced-settings-section/diagnostics-traces.js
```diff
@@ -44,165 +36,231 @@ const downloadAsFile = ( contents ) => {
...
+		apiFetch( { path: TRACES_PATH } )
+			.then( ( response ) => setTraces( response.traces || [] ) )
+			.catch( () => setTraces( [] ) );
 	}, [] );
 
 	useEffect( () => {
-		refreshSummary();
-	}, [ refreshSummary ] );
+		refreshTraces();
+	}, [ refreshTraces ] );
```
wjrosa, line 67: This only fires once on mount. If the merchant toggles recording on and then drives a failing checkout in another tab, the list won't update until they reload — the indicator switches to "Recording" but the row never appears. Consider polling on an interval while `isRecording`, or at least refetching when the toggle flips.

wjrosa (commented):
Nice cleanup — the inline list reads much better than the summary-then-fetch flow. A few small things below.

client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
```diff
@@ -0,0 +1,190 @@
...
+						totalCount
+					) }
+					{ failedCount > 0 && (
+						<>
+							{ ' · ' }
+							<span className="wc-stripe-diagnostics-toolbar__failed-count">
+								{ sprintf(
+									/* translators: %d: number of failed traces */
+									__(
+										'%d failed',
```
wjrosa, line 74: Use `_n( '%d failed', '%d failed', failedCount, ... )` here — `__` won't pluralize for locales with multiple plural forms.

client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
```diff
@@ -0,0 +1,190 @@
...
+						className={ classNames(
+							'wc-stripe-diagnostics-segmented__option',
+							{
+								'is-active': filter === FILTER_FAILED,
+							}
+						) }
+						onClick={ () => onFilterChange( FILTER_FAILED ) }
+					>
+						{ __( 'Failed only', 'woocommerce-gateway-stripe' ) }
+					</button>
```
wjrosa, line 127: The segmented filter is visual-only for assistive tech — screen readers can't tell which segment is active. Add `aria-pressed={ filter === FILTER_ALL }` / `aria-pressed={ filter === FILTER_FAILED }` to the two buttons.

client/settings/advanced-settings-section/diagnostics-trace-row.js
```diff
@@ -0,0 +1,163 @@
...
+			/>
+			<div className="wc-stripe-diagnostics-trace-row__main">
+				<div className="wc-stripe-diagnostics-trace-row__heading">
+					<span className="wc-stripe-diagnostics-trace-row__id">
+						{ trace.id }
+					</span>
+					<span className="wc-stripe-diagnostics-trace-row__when">
+						{ '· ' }
+						{ formatRelative(
+							trace.created_at || trace.updated_at || nowSeconds,
```
wjrosa: Falling through to `nowSeconds` when both timestamps are missing makes the row render as "just now", which is misleading. Returning `'—'` from `formatRelative` for the no-timestamp case would match what `formatDuration` already does.

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
```diff
@@ -177,6 +185,26 @@ public function get_traces( $request ) {
...
+	 * Wipe every stored trace. Powers the "Clear" action in the admin trace
+	 * list. Does not gate on the diagnostics-enabled toggle: a merchant who
+	 * disabled capture must still be able to clear out the bundle they
+	 * captured while it was on.
+	 *
+	 * @return WP_REST_Response
+	 */
+	public function delete_traces() {
+		$deleted = $this->store->count();
+		$this->store->delete_all();
```
wjrosa: `count()` then `delete_all()` isn't atomic — two concurrent DELETEs would each return the full pre-delete count. Cheap fix: have `delete_all()` return the rows it actually removed and use that as `deleted`.

client/settings/advanced-settings-section/diagnostics-traces.js
malithsen: I think updating this list only on page refresh is okay. Polling comes with a performance overhead. There's no expectation that the trace list should be updated in real time (In the same way, the WC orders list doesn't poll for new orders)

client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
malithsen: Thanks. Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5412/commits/da5dfb1638c52187f269d7c933c26710a4b3bd84

client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
malithsen: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5412/commits/966fabd3d92b847945824f352a5c876ca70e6efb

client/settings/advanced-settings-section/diagnostics-trace-row.js
malithsen: Handled in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5412/commits/64b7b09fd114e3fdb2ff3aca461026e9967219d1

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
malithsen: Concurrent delete scenario should be super rare in practice, since it is only exposed to admins. However, handling it is not that hard, so updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5412/commits/45111ad6b0095da961ae0359717a956bfb7309e8

wjrosa (approved):
Looks good to me! Thanks for addressing the suggestions 👍

PR #5413, opened 2026-05-08
Make subscription detached messages hideable (#5391)

### LINKED ISSUE #5391: No way to hide "payment method for this subscription has been detached" admin message
### Description

I have a bespoke setup using WC Stripe gateway as one of several other options. I also don't enable auto-renew. Customers often change the one they want to use. 

Due to the nature of my use, this "Subscription detached" message is showing on the admin side for pretty much all of them:

> The payment method for this subscription has been detached, preventing renewals. To fix this, either:
> 1) Share the payment method page link with the customer to update it: [Payment method page →](https://example.com/)
> or
> 2) Manually update the payment method in the subscription's billing details using a valid payment method from the customer's Stripe account: [Stripe customer page →](https://example.com/)
> 
> To list all your current subscriptions with payment methods detached, go to WooCommerce -> Status -> [Tools](https://example.com/) -> List Stripe subscriptions with detached payment method.

I notice a number of other messages in [`class-wc-stripe-admin-notices.php`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/5ccc3383ca8895a8cfd77323183847d87eaff4a5/includes/admin/class-wc-stripe-admin-notices.php) are implemented with the ability to either dismiss or disable them. This one is not.

### Steps to reproduce

1. Have a subscription product that was paid for once via Stripe.
2. Have it "detach" (to be honest, I'm not sure how this happens).

### Expected behavior

Either:

1. The notice should be made dismissible, because in some cases it's completely intended behaviour that a payment method has become detached.
2. Add an option to disable it entirely (I'd guess [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/5ccc3383ca8895a8cfd77323183847d87eaff4a5/includes/admin/class-wc-stripe-admin-notices.php#L594) and [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/5ccc3383ca8895a8cfd77323183847d87eaff4a5/includes/admin/class-wc-stripe-admin-notices.php#L455).

### Screenshots

Image

### System report

<details>

<details><summary>System Status Report</summary>
removed
</details>

</details>

### Additional context

I am happy to make this change and PR it, if it's something you'd want. I understand why the message is there but it's quite shocking it can't be dismissed or disabled.

ara303:
Closes #5391.

Please see the issue mentioned above for an explanation of this.

**To recap,** the Stripe gateway plugin adds a large notice to every subscription post's admin page if it has become detached. Although usually useful, in cases where it isn't (see my issue) it can't be hidden.

This PR changes only enough to make the message hideable if the user wants to.


includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -453,6 +453,10 @@ public function check_express_checkout_location(): void {
 	 * @return void
 	 */
 	public function subscription_check_detachment() {
+		if ( 'no' === get_option( 'wc_stripe_show_subscription_detached_notice' ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
ara303: Well, isn't that embarrassing. 😓 Obviously it's no good if my change globally changes all detached notices indefinitely, so I've changed it to store an array of the dismissed subscriptions and it will only hide those where the user's specifically opted to hide them.

tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
```diff
@@ -797,6 +797,86 @@ public function provide_test_subscription_check_detachment() {
...
+
+		remove_filter( 'pre_http_request', $test_request, 10, 3 );
+
+		$this->assertArrayNotHasKey( 'subscription_detached', $notices->notices );
+
+		unset( $_REQUEST );
+		$theorder = $original_order;
+		WC_Stripe_Database_Cache::delete( 'payment_method_for_source_src_123' );
+		delete_option( 'wc_stripe_show_subscription_detached_notice' );
+	}
```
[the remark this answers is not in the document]
ara303: I think Coderabbit might be making a mistake here. It thinks `has_status( [ 'active' ] )` is new code but it's not. However I noticed I accidentally deleted a comment here that I've restored in 69d5a34.

wjrosa:
Thanks for the contribution, @ara303 ! Is this PR ready for review? If so, would you mind adding the missing changelog entries? You can target the next version (10.8.0).

ara303:
Hi @wjrosa! Yes, I think this is ready for review now. 

Happy to add the missing changelog.txt and readme.txt entries for this, but I notice a heading for `10.8.0` doesn't exist yet. Please could I ask you to clarify you want me to make a new heading in both files for 10.8.0 (in this PR)? Like so:

```diff
+xxxx-xx-xx - version 10.8.0
+* Fix - Make subscription detached messages hideable


xxxx-xx-xx - version 10.7.0
* Fix - Hide the "move Stripe to the top" Optimized Checkout notice when all payment methods above Stripe are disabled
[... continued]
```

Sorry about this question. I haven't yet been able to contribute to big projects like this so I'm not sure of all the ins and outs.

wjrosa:
@ara303 no worries! It is fine the way you suggested

ara303:
Thanks, @wjrosa. I've added a new heading and my change (same title as this PR) to [`readme.txt`](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5413/changes#diff-9e6e4772050998a5c0dc3c61acf3dab0a7e594566171fa5746d6b62f9598efb6) and [`changelog.txt`](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5413/changes#diff-b40cd67182487ef24807d9c9268329d35fbd96aa2b0a9cae69e2e0d746b1c666) in commit 4a736bc.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -476,6 +476,12 @@ public function subscription_check_detachment() {
 			return;
 		}
 
+		$dismissed_subscriptions = (array) get_option( 'wc_stripe_show_subscription_detached_notice', [] );
```
wjrosa: Option name reads `show_subscription_detached_notice` but actually stores subscriptions where the notice should NOT show. Suggest renaming to `wc_stripe_dismissed_subscription_detached_notices` so the key matches the data.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -646,6 +652,20 @@ public function hide_notices() {
...
+					$subscription_id = 0;
+					if ( WC_Stripe_Woo_Compat_Utils::is_custom_orders_table_enabled() ) {
+						$subscription_id = absint( $query_params['id'] ?? 0 );
+					} elseif ( isset( $query_params['post'] ) ) {
+						$subscription_id = absint( $query_params['post'] );
+					}
+					$dismissed = (array) get_option( 'wc_stripe_show_subscription_detached_notice', [] );
+					if ( $subscription_id && ! in_array( $subscription_id, $dismissed, true ) ) {
+						$dismissed[] = $subscription_id;
+						update_option( 'wc_stripe_show_subscription_detached_notice', $dismissed );
```
wjrosa: Defaults to autoload. The list grows by one ID per dismissal and is read on every admin pageload — pass `false` as the third arg. Even cleaner: store as post meta on the subscription, so dismissals follow the entity and disappear when it's deleted.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -646,6 +652,20 @@ public function hide_notices() {
 						wp_safe_redirect( remove_query_arg( [ 'wc-stripe-hide-notice', '_wc_stripe_notice_nonce' ], esc_url_raw( wp_unslash( $_SERVER['REQUEST_URI'] ) ) ) );
 					}
 					break;
+				case 'subscription_detached':
+					$query_params  = wp_unslash( $_REQUEST ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
+					$subscription_id = 0;
+					if ( WC_Stripe_Woo_Compat_Utils::is_custom_orders_table_enabled() ) {
```
wjrosa: Resolution asymmetry: `subscription_check_detachment()` derives the subscription via `$theorder` / `$GLOBALS['post']`, but the dismiss path reads `$_REQUEST['id']` / `$_REQUEST['post']`. They line up only because the dismiss link is generated on the subscription edit page — consider a shared helper so both sides can't drift.

tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
```diff
@@ -797,6 +797,86 @@ public function provide_test_subscription_check_detachment() {
...
+		$notices = new WC_Stripe_Admin_Notices();
+		$notices->subscription_check_detachment();
+
+		remove_filter( 'pre_http_request', $test_request, 10, 3 );
+
+		$this->assertArrayNotHasKey( 'subscription_detached', $notices->notices );
+
+		unset( $_REQUEST );
+		$theorder = $original_order;
+		WC_Stripe_Database_Cache::delete( 'payment_method_for_source_src_123' );
```
wjrosa, line 852: `unset( $_REQUEST )` wipes the whole superglobal and can leak into sibling tests. Mirror the second test and only unset the keys this test set (`page`, `id`).

wjrosa (commented):
A few small things on top of the existing review thread.

ara303:
I've addressed all but one feedback point in [58fb1a5](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5413/commits/58fb1a5c4c69ed7eac554754ffba22b31fd4058b) and brought my `readme.txt` and `changelog.txt` changes up-to-date in [e2895af](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5413/commits/e2895af02ec92070543414c3db247994e67a4e7b). 

[Here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5413/#discussion_r3228588807) you asked me to pass `false` to `update_option` (done). You also mentioned:

> Even cleaner: store as post meta on the subscription, so dismissals follow the entity and disappear when it's deleted.

That'd be a fundamentally different approach so I may as well close this PR, rebase from main, and restart work using post meta instead of a WP option.

Can you please confirm you'd prefer that? If so, I don't mind doing it. I'll close this PR and open another, since the code will be completely different (different method to store data + different test).

wjrosa:
> Can you please confirm you'd prefer that? If so, I don't mind doing it. I'll close this PR and open another, since the code will be completely different (different method to store data + different test).

The post meta solution feels cleaner, but I am OK with this version too. What do you think, @Mayisha ?

ara303:
@wjrosa I have some commits with a version using postmeta and tests to accompany that if you or others want to see!

It honestly is a way simpler implementation.

Sorry about the failing checks. I can't for the life of me get your dev env going, but it's likely my fault (I use Windows with WSL2 for a Ubuntu VM and VS Code in remote mode to work on it, so there's probably some link in the chain that's broken there). I've got phpcs and phpstan working at least.

daledupreez:
@ara303, if you'd prefer to create a new PR, please feel free to do so! If it's much cleaner then it's much cleaner, and will be better for everyone in the long run.

Mayisha:
> The post meta solution feels cleaner, but I am OK with this version too. What do you think, @Mayisha ?

I also prefer the post meta solution.

wjrosa:
I will go ahead and close this one in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5434, ok?

PR #5414, opened 2026-05-08
[RSM - Diagnostics Mode] Capture order snapshot on trace finalization

malithsen:
Fixes RSM-2777

Captures a curated snapshot of the order data at the moment a diagnostics trace transitions to a terminal status

CleanShot 2026-05-08 at 13 40 29@2x

CleanShot 2026-05-08 at 13 41 04@2x

## Testing instructions

1. Toggle on **WooCommerce → Settings → Payments → Stripe → Advanced settings → Capture checkout diagnostics**.
2. Drive a guest checkout to a known failure (Stripe test card `4000 0000 0000 0002`).
3. In Advanced settings, click **View** on the failed trace row.
4. Confirm the JSON has a top-level `order_snapshot` with populates fields including `status`, `total`, `payment_method_title`, `line_items`, and `order_notes` 
5. Confirm the trace's `meta.order_id` is set to the failed order id.

---

### Changelog entry

<details>

#### Comment 
Unreleased. Merging to a feature branch.

</details>

**Post merge**


includes/diagnostics/class-wc-stripe-diagnostics-order-snapshotter.php
```diff
@@ -0,0 +1,266 @@
...
+			'payment_method_title' => (string) $order->get_payment_method_title(),
+			'transaction_id'       => (string) $order->get_transaction_id(),
+			'customer_id'          => (int) $order->get_customer_id(),
+			'created_at'           => $created_at ? (int) $created_at->getTimestamp() : null,
+			'date_paid'            => $date_paid ? (int) $date_paid->getTimestamp() : null,
+			'date_completed'       => $date_completed ? (int) $date_completed->getTimestamp() : null,
+			'order_key'            => (string) $order->get_order_key(),
+			'billing_country'      => (string) $order->get_billing_country(),
+			'shipping_country'     => (string) $order->get_shipping_country(),
+			'currency_locale'      => (string) get_locale(),
```
wjrosa: `get_locale()` returns the WP/site locale (e.g. `en_US`), not a currency or customer locale. Suggest renaming to `site_locale` to avoid future confusion.

includes/diagnostics/class-wc-stripe-diagnostics-order-snapshotter.php
```diff
@@ -0,0 +1,266 @@
...
+	/**
+	 * Action listener for `wc_stripe_diagnostics_trace_finalized`. Queues
+	 * the session for end-of-request snapshotting; the actual order read
+	 * happens in {@see self::flush_pending_snapshots()} during shutdown.
+	 *
+	 * @param string $session_id Trace session identifier.
+	 * @param string $status     New terminal status (unused, but part of the action contract).
+	 */
+	public function on_trace_finalized( $session_id, $status ): void {
+		unset( $status );
```
wjrosa: `unset( $status )` is cosmetic — PHP doesn't warn on unused params. Drop it; the docblock already explains the param.

includes/diagnostics/class-wc-stripe-diagnostics-trace-store.php
```diff
@@ -151,6 +151,97 @@ function () use ( $session_id, $status ) {
...
+
+		return (bool) $this->with_lock(
+			function () use ( $session_id, $snapshot ) {
+				$trace = $this->read_trace( $session_id );
+				if ( null === $trace ) {
+					return false;
+				}
+				$trace['order_snapshot'] = $snapshot;
+				$trace['updated_at']     = time();
+				return $this->write_trace( $session_id, $trace );
```
wjrosa, line 242: `set_order_snapshot` writes via `write_trace()` which doesn't enforce `MAX_TRACE_BYTES` — only `append_event` checks the cap. The snapshotter's per-field caps make a 100KB overflow unlikely in practice, but the class-docblock invariant ("the serialized trace blob is capped at MAX_TRACE_BYTES") no longer strictly holds. Either enforce here or downgrade the invariant doc.

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
```diff
@@ -272,6 +290,13 @@ public function on_webhook_received( $webhook_type, $notification, $resolved_ord
 		}
 		$this->store->create( $session_id, [ 'source' => 'webhook' ] );
 
+		// Pin the order id to the trace's meta. First-writer-wins, so a
+		// webhook arriving after the original API request flow doesn't
+		// clobber the order id captured there.
+		if ( is_object( $resolved_order ) && method_exists( $resolved_order, 'get_id' ) ) {
```
wjrosa: `$resolved_order` is typed `?WC_Order` upstream — prefer `$resolved_order instanceof WC_Order` over `is_object && method_exists` (the existing line 310 uses the same duck-typing; could extract a local `$order_id = ... instanceof WC_Order ? ... : null` and reuse for both pin + event).

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
```diff
@@ -201,6 +212,13 @@ public function on_request_body( $request, $api ) {
 			$request['metadata'][ self::SESSION_ID_META_KEY ] = $session_id;
 		}
 
+		// Pin the order id to the trace's meta so the order_snapshotter
+		// can read it deterministically when the trace finalizes.
+		// First-writer-wins, safe to call from any hook that observes one.
+		if ( isset( $request['metadata']['order_id'] ) ) {
+			$this->store->set_order_id( $session_id, (int) $request['metadata']['order_id'] );
+		}
```
wjrosa, line 219: Request-side pin only: today every order-bound Stripe call seeds `metadata.order_id`, but if a future flow omits it the snapshot will silently drop. Worth a one-line note in the comment so the gap reads as intentional rather than accidental.

tests/phpunit/diagnostics/class-wc-stripe-diagnostics-order-snapshotter-test.php
```diff
@@ -0,0 +1,155 @@
...
+
+	/**
+	 * No order_snapshot is appended when the trace has no order id pinned
+	 * to its meta — the common production case for traces that finalize
+	 * without ever observing an order (abandoned express, console-only
+	 * traces). Status still transitions normally.
+	 */
+	public function test_skips_when_order_unresolvable() {
+		$this->store->create( 'sess-unresolvable' );
+
```
wjrosa, line 129: Asserting on the literal `[email]` placeholder couples this test to the redactor's replacement string. If the placeholder ever changes the snapshotter test breaks for the wrong reason. Consider keeping just `assertStringNotContainsString( 'foo@bar.com', ... )` and moving the placeholder assertion to the redactor's own tests.

wjrosa (commented):
Nice work — solid docs, end-to-end test, sensible shutdown deferral. A few inline nits and one minor invariant question.

includes/diagnostics/class-wc-stripe-diagnostics-trace-store.php
malithsen: I don't think we should cap the order snapshot. In normal trace data, there's a security concern that requires this cap. Whereas in order snapshots, data comes from the DB and is bounded by default. So, prefer to keep as-is.

includes/diagnostics/class-wc-stripe-diagnostics-order-snapshotter.php
malithsen: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5414/commits/890a29ce04f01e0beea38156cf6b69733dd8454f

includes/diagnostics/class-wc-stripe-diagnostics-order-snapshotter.php
malithsen: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5414/commits/890a29ce04f01e0beea38156cf6b69733dd8454f

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
malithsen: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5414/commits/890a29ce04f01e0beea38156cf6b69733dd8454f

tests/phpunit/diagnostics/class-wc-stripe-diagnostics-order-snapshotter-test.php
malithsen: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5414/commits/890a29ce04f01e0beea38156cf6b69733dd8454f

includes/diagnostics/class-wc-stripe-diagnostics-recorder.php
malithsen: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5414/commits/42f0d5f41ddeb716c7b5fb754dfcb5ec0db0cd7b

wjrosa (approved):
Looks good to me! 👍

PR #5421, opened 2026-05-11
Guard against multiple WC_Stripe instances before plugin initialization

daledupreez:

This PR adds safeguards in `WC_Stripe::init()` to prevent duplicate initialisation logic and hook registrations from occurring when `WC_Stripe` is instantiated more than once due to code being called from within `WC_Stripe::init()` making calls to `WC_Stripe::get_instance()` before `WC_Stripe::$instance` is populated.

The specific change is to populate `self::$instance` within the constructor, and then ensuring that all downstream logic only triggers general initialization logic when `self::$instance === $this`, i.e. we are the first `WC_Stripe` instance to be registered. This ensures various objects are only instantiated once, and that hooks are only registered against the first `WC_Stripe` instance.

It's worth noting that this is a relatively messy or noisy change, but the driver for that was to minimise the risks from the change. In particular, we need to run all the `require_once`s during the first run through to ensure that we have the dependencies for the properties of the `WC_Stripe` object, and we were aiming to keep the various `new <something>()` calls inline so they were not at the end of various files which we eventually want to autoload. There is a _lot_ of cleanup that needs to happen in `init()`, but this PR is not the place to tackle that work.

The specific issue this addresses was an infinite loop that I encountered while testing 10.7.0. The trace for the loop was as follows:
* `do_action('plugins_loaded')`
* `WP_Hook->do_action`
* `WP_Hook->apply_filters`
* `woocommerce_gateway_stripe_init`
* `woocommerce_gateway_stripe`
* `WC_Stripe::get_instance`
* `WC_Stripe->__construct`
* `WC_Stripe->init`
* `WC_Stripe_Payment_Gateways_Controller->__construct`
* `WC_Stripe_Payment_Method_Configurations::get_upe_enabled_payment_method_ids`
* `WC_Stripe_Payment_Method_Configurations::maybe_migrate_payment_methods_from_db_to_pmc`
* `WC_Stripe_UPE_Payment_Method_Amazon_Pay::is_amazon_pay_available_for_account_country`
* `WC_Stripe::get_instance`
* `WC_Stripe->__construct`
* `WC_Stripe->init`
* `WC_Stripe_Payment_Gateways_Controller->__construct`
* `WC_Stripe_Payment_Method_Configurations::get_upe_enabled_payment_method_ids`
* ...

The specific trigger for this situation was the dedicated `wc_stripe_amazon_pay_default_on` option being set to `'yes'` (which is the case for recent installs), and the `pmc_enabled` sub-option was deleted. Any admin page visit would then 

## Testing instructions

* Check the code from inspection
* Verify that all unit and e2e tests pass
* Try to replicate the issue with a build from the `release/10.7.0` by doing the following:
  - Connect to Stripe -- I triggered the situation in test mode, so that should be fine
  - Ensure that `wc_stripe_amazon_pay_default_on` is set to `'yes'` -- `wp option update wc_stripe_amazon_pay_default_on yes`
  - Remove the `pmc_enabled` sub-option -- `wp option patch delete woocommerce_stripe_settings pmc_enabled`
  - Immediately load an admin page
  - In my site, that triggered an Out of Memory fatal due to the infinite loop, but the symptoms could be different
* Apply the changes from this PR to the site (or install the new build and replicate the settings above)
  - It's especially crucial to remove the `pmc_enabled` sub-setting

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe.php
+++ b/includes/class-wc-stripe.php
@@ -80,6 +80,13 @@ class WC_Stripe {
 	protected $stripe_gateway = null;
 
 	/**
+	 * Tracking initialization of various components to prevent repeated work.
+	 *
+	 * @var array<string, bool>
+	 */
+	protected static $initialization_tracking = [];
+
+	/**
 	 * Private clone method to prevent cloning of the instance of the
 	 * *Singleton* instance.
 	 *
@@ -115,7 +122,9 @@ class WC_Stripe {
 	 * @version 5.0.0
 	 */
 	public function init() {
-		if ( is_admin() ) {
+
+		if ( is_admin() && ! isset( self::$initialization_tracking['privacy'] ) ) {
+			self::$initialization_tracking['privacy'] = true;
 			require_once WC_STRIPE_PLUGIN_PATH . '/includes/admin/class-wc-stripe-privacy.php';
 			new WC_Stripe_Privacy();
 		}
@@ -146,7 +155,11 @@ class WC_Stripe {
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-action-scheduler-service.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-webhook-state.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/class-wc-stripe-webhook-handler.php';
-		new WC_Stripe_Webhook_Handler();
+
+		if ( ! isset( self::$initialization_tracking['webhook_handler'] ) ) {
+			self::$initialization_tracking['webhook_handler'] = true;
+			new WC_Stripe_Webhook_Handler();
+		}
 
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/payment-tokens/trait-wc-stripe-fingerprint.php';
 		require_once WC_STRIPE_PLUGIN_PATH . '/includes/payment-tokens/interface-wc-stripe-payment-method-comparison.php';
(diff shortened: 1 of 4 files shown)


includes/class-wc-stripe.php
```diff
@@ -202,111 +212,146 @@ public function init() {
...
 		}
 
 		// Modify emails emails.
-		add_filter( 'woocommerce_email_classes', [ $this, 'add_emails' ], 20 );
+		if ( self::$instance === $this ) {
+			add_filter( 'woocommerce_email_classes', [ $this, 'add_emails' ], 20 );
+		}
 
-		if ( version_compare( WC_VERSION, '3.4', '<' ) ) {
+		if ( version_compare( WC_VERSION, '3.4', '<' ) && self::$instance === $this ) {
```
diegocurbelo, line 320: The plugin requires `WC 9.9+` we can just remove this if block

daledupreez: TBH, I'd prefer to clean that up in a dedicated follow-up PR.

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -691,4 +691,214 @@ public function provide_test_maybe_reconfigure_webhooks_after_adaptive_pricing_e
...
+			);
+		}
+	}
+
+	/**
+	 * The frontend-only billing-fields filter must follow the same guard.
+	 */
+	public function test_checkout_update_email_field_priority_filter_registered_only_for_first_instance_in_frontend(): void {
+		$initial_screen = $GLOBALS['current_screen'] ?? null;
+		$current_screen = \WP_Screen::get( 'post.php' );
```
diegocurbelo: `$current_screen` is not used, maybe it was intended to be `$GLOBALS['current_screen']` (that is restored in the finally block?

diegocurbelo (approved):
Code changes look good, and it tests very well!

Left a small question for a test, but they all pass, so I'm happy leaving that for a future PR.

tests/phpunit/class-wc-stripe-test.php
daledupreez: Good catch -- fixed in 520a55e3b677d3acdb7e0de1cb08f31f46a0e6e9. I only partly changed from `global $current_screen` to `$GLOBALS['current_screen']` while trying to get this right.

PR #5427, opened 2026-05-12
Add PHP 8.5 as an option to Manual QIT and Manual E2E workflows

daledupreez:

This PR adds PHP 8.5 to the list of options for the manual e2e and QIT workflows, as both only included PHP 7.4 through 8.4.

## Testing instructions

Inspection should be sufficient.

---

### Changelog entry

<details>

#### Comment 
This is a small tweak to our GitHub workflows.

</details>

**Post merge**


.github/workflows/ci-manual.yml
```diff
@@ -22,6 +22,7 @@ on:
           - '8.1'
           - '8.3'
           - '8.4'
+          - '8.5'
```
wjrosa, line 25: Since you're touching both PHP lists, worth aligning them — ci-manual is `7.4, 8.0, 8.1, 8.3, 8.4, 8.5` (no 8.2) while manual-e2e is `7.4, 8.1, 8.2, 8.3, 8.4, 8.5` (no 8.0). Either gap is probably an oversight rather than intent; picking up 8.2 here (and deciding whether to keep EOL 8.0) avoids a "why isn't my version listed" question later.

wjrosa (commented):
One observation while you're in both files.

.github/workflows/ci-manual.yml
daledupreez: Updated in 995d62110558d50ff93e110417688bdabd47a389. I decided to update `ci-manual.yml` by dropping 8.0 and adding 8.2. I don't see much value in testing 8.0, especially given that we are still testing 7.4.

wjrosa (approved):
Looks good to me as well 👍

PR #5431, opened 2026-05-13
Adds an auto-off feature for diagnostics mode

malithsen:
Fixes RSM-2302

Adds an auto-off capture limit to the Checkout Diagnostics card so recording turns off automatically after a configurable number of captured checkouts.

**Trade-offs / decisions**
- Auto-off is mandatory; there is no "unlimited" option. 
- The flip happens *after* the new trace is stored, so the trace that pushes us to the limit is still captured. The next
  request gets 403'd by the existing diagnostics-disabled permission callback.
- Reducing the limit below the current stored count does not retroactively flip the toggle.

CleanShot 2026-05-13 at 08 56 33@2x

## Testing instructions

1. Enable Diagnostics mode via **Stripe → Settings → Advanced settings**.
2. Set the inline **Auto-off after** picker to `5`.
3. Run 5 test checkouts so 5 traces accumulate. Refresh the settings page between attempts to watch the counter and progress bar pdate (`1 of 5 captured`, `2 of 5 captured`, …).
4. After the 5th trace lands, refresh the page. Verify:
   - The diagnostics toggle is now **off**.
   - The trace list is still visible and the **Copy all** / **Clear** actions still work.
   - The inline limit picker is hidden .
5. Click **Clear**, confirm - the count resets to 0/5.
6. Toggle diagnostics back on, change the picker to `10`, click **Save changes**, run one more checkout. The counter should now read `1 of 10 captured`.

---

### Changelog entry

<details>

#### Comment 
Unreleased feature merging to a feature branch
</details>

**Post merge**

### DIFF
--- a/client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
+++ b/client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
@@ -7,12 +7,18 @@ import ConfirmationModal from 'wcstripe/components/confirmation-modal';
 const FILTER_ALL = 'all';
 const FILTER_FAILED = 'failed';
 
+// Mirrors PHP CAPTURE_LIMIT_PRESETS / DEFAULT_CAPTURE_LIMIT — keep in sync.
+const LIMIT_PRESETS = [ 5, 10, 25, 50 ];
+const DEFAULT_LIMIT = 10;
+
 const DiagnosticsTraceToolbar = ( {
 	totalCount,
 	failedCount,
 	filter,
 	onFilterChange,
 	isRecording,
+	captureLimit = DEFAULT_LIMIT,
+	onChangeCaptureLimit,
 	onCopy,
 	onClear,
 	isCopying,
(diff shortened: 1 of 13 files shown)


client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
```diff
@@ -7,12 +7,18 @@ import ConfirmationModal from 'wcstripe/components/confirmation-modal';
 const FILTER_ALL = 'all';
 const FILTER_FAILED = 'failed';
 
+// Mirrors PHP CAPTURE_LIMIT_PRESETS / DEFAULT_CAPTURE_LIMIT — keep in sync.
+const LIMIT_PRESETS = [ 5, 10, 25, 50 ];
```
wjrosa: These are the third place these literals live (PHP `CAPTURE_LIMIT_PRESETS`/`DEFAULT_CAPTURE_LIMIT`, the `useDiagnosticsCaptureLimit` default in `client/data/settings/hooks.js`, and here). Could the settings REST payload expose `diagnostics_capture_limit_presets` (and reuse the PHP default) so the JS side has no parallel source of truth? Otherwise a future preset change is a four-file edit waiting to drift.

client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
```diff
@@ -30,61 +36,103 @@ const DiagnosticsTraceToolbar = ( {
...
+								</span>
+							</>
+						) }
+					</span>
+				</div>
+				{ isRecording && onChangeCaptureLimit && (
+					<span className="wc-stripe-diagnostics-toolbar__limit">
+						<span>
+							{ __(
+								'Auto-off after',
```
wjrosa: Splitting the sentence into 'Auto-off after' + `<select>` + 'orders' bakes in English word order and breaks in many locales. Consider a single translatable string like `'Auto-off after %s orders'` rendered via `createInterpolateElement` (or similar) with the `<select>` as the interpolated node.

wjrosa (commented):
Auto-off flow looks solid and well tested. A few short notes inline.

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
```diff
@@ -285,4 +293,58 @@ public static function is_enabled(): bool {
...
+	 * Silently flips the diagnostics toggle to 'no' when the stored trace
+	 * count reaches the capture limit. Returns true when this call flipped.
+	 */
+	public static function enforce_capture_limit( ?WC_Stripe_Diagnostics_Trace_Store $store = null ): bool {
+		if ( ! self::is_enabled() ) {
+			return false;
+		}
+		$limit = self::capture_limit();
+		$store = $store ?? new WC_Stripe_Diagnostics_Trace_Store();
+		if ( $store->count() < $limit ) {
```
wjrosa, line 329: Two parallel `ingest_events` requests arriving at `count = limit - 1` can both pass this check, each append a trace (one slot over budget), and only one wins the toggle flip. Probably acceptable given the PR description's stated trade-off, but a one-line comment here would help future readers not mistake this for strict enforcement.

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
```diff
@@ -285,4 +293,58 @@ public static function is_enabled(): bool {
...
+			return false;
+		}
+
+		$settings = get_option( self::SETTINGS_OPTION, [] );
+		if ( ! is_array( $settings ) ) {
+			$settings = [];
+		}
+		$settings[ self::SETTINGS_KEY ] = 'no';
+		update_option( self::SETTINGS_OPTION, $settings );
```
wjrosa: `update_diagnostics_capture_limit` in the settings controller writes via `$this->gateway->update_option(...)`, which also keeps the gateway's in-memory `settings` cache consistent. A direct `update_option(...)` here skips that — code in the same request that reads `$gateway->get_option('diagnostics')` (rather than `self::is_enabled()`) will still see `yes` until the gateway is reinstantiated. Worth keeping the two write paths consistent.

client/settings/advanced-settings-section/style.scss
```diff
@@ -135,6 +143,38 @@ $diagnostics-bg-toolbar: styles.$wp-gray-0;
...
+
+	&__progress {
+		height: 3px;
+		background: $diagnostics-border;
+		overflow: hidden;
+	}
+
+	&__progress-bar {
+		height: 100%;
+		background: #2271b1;
```
wjrosa: Hardcoded blue. Every other color in this file is a SCSS variable (`$diagnostics-*`) or comes from `styles.$wp-*`. Consider extracting a `$diagnostics-progress-fg` (or reusing the WP blue var) so this stays themable alongside the rest of the surface.

client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
malithsen: Good idea. Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5431/commits/882d942bc528c73b14b1498557adddd3bc13765d

I opted to keep the frontend presets defined as a fallback, but the primary driver is the backend presets

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
malithsen: It's not meant to be a hard limit. Added a note in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5431/changes/3e10c3cc40b024fd6755520c29310caa06ad7369

client/settings/advanced-settings-section/style.scss
malithsen: Good catch! Replaced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5431/commits/617505c1639908ef670c3b7bf1be7bd6e853ced8

includes/admin/class-wc-rest-stripe-diagnostics-controller.php
malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5431/commits/337ca03127b0600594c8ba636605014c63ab578f

client/settings/advanced-settings-section/diagnostics-trace-toolbar.js
malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5431/commits/fb1bb7c6312e6131627633757245ab49540ff328

wjrosa (approved):
Code is good and works as expected! Thanks for addressing the review comments.

Image
Image
Image

PR #5439, opened 2026-05-14
Surface wallet branding on saved tokens (Apple Pay / Google Pay)

### LINKED ISSUE #5437: Saved payment methods don't show wallet branding (Apple Pay / Google Pay / Link) for wallet-paid cards
Surfaced during PR woocommerce/woocommerce-gateway-stripe#5284 testing on 2026-05-13 (see [this comment](<https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5284#issuecomment-4444726165>) from @daledupreez):

> For Apple Pay and Google Pay, the shopper-facing payment method shows up as the underlying card, and there is no mention of Apple Pay or Google Pay.
>
> I tested purchasing a new subscription with Google Pay, and the saved payment method also stores the card details without the payment type, so I think this is a pre-existing behaviour.

Confirmed pre-existing; reproduces on a brand-new wallet purchase too, independent of subscriptions or change-payment flows.

## Why it matters

The shopper cannot distinguish their saved tokens in **My Account → Payment Methods**: a Visa tokenized via Apple Pay renders identically to a Visa entered manually. The same gap leaks into the change-payment-method "from"-side note when the original purchase used Apple Pay / Google Pay / Mastercard-with-Link-signup (see woocommerce/woocommerce-gateway-stripe#5284 for the "to"-side fix and woocommerce/woocommerce-gateway-stripe#5382 for the related token-association gap).

## Root cause

When the Stripe gateway tokenizes a wallet payment, it stores only the underlying card data on the resulting `WC_Payment_Token_CC` (brand, last4, exp_month, exp_year). The wallet type — available from Stripe at `payment_method.card.wallet.type` (`apple_pay` / `google_pay` / `link`) — is dropped.

Stripe plugin code already consumes `card.wallet.type` elsewhere:

* `includes/payment-methods/class-wc-stripe-upe-payment-method-cc.php:65`
* `includes/payment-methods/class-wc-stripe-upe-payment-method-oc.php:49`
* Shared helper: `WC_Stripe_UPE_Payment_Method::get_card_wallet_type_title()` (`includes/payment-methods/class-wc-stripe-upe-payment-method.php:857`)

## Suggested approach

1. In the token-create paths in `WC_Stripe_Payment_Tokens` (`includes/payment-tokens/class-wc-stripe-payment-tokens.php`, around line 410), read `$payment_method->card->wallet->type` and persist it as a `_stripe_wallet_type` meta on the new `WC_Payment_Token_CC`.
2. Augment the existing `get_account_saved_payment_methods_list_item` filter (line 56 of the same file) to prepend the wallet label when the meta is set — e.g. "Apple Pay · Visa •••• 4242".
3. PHPUnit coverage:
   * Token from a non-wallet card: no meta, display unchanged.
   * Token from each wallet type: meta set, display includes wallet branding.
   * Legacy tokens predating the meta: graceful no-op.

## Out of scope

* Storing the email-with-Link badge UI for Link-stamped cards (bare Mastercard with the Link sign-in checkbox) — UX call requiring product/design input.
* Subscription `_payment_method_title` rendering — already handled.

cc @daledupreez

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

wjrosa:
Fixes #5437

- Persist Stripe's `card.wallet.type` (`apple_pay` / `google_pay` / `link`) on `WC_Stripe_Payment_Token_CC`.
- Render wallet-tokenized cards in **My Account → Payment Methods** as "Apple Pay (Visa) ending in 4242" instead of just the underlying card.

### How it works

1. `wallet_type` is set on every CC token-create path (`create_payment_token_for_user`, `add_token_to_user` new + duplicate-refresh branches, `WC_Stripe_Customer::add_source`). `handle_saving_payment_method` also backfills it on the `update_payment_token` reuse branch so pre-fix tokens upgrade on next use of the same card.
2. `get_account_saved_payment_methods_list_item` wraps the brand as `"Wallet Label (Card Brand)"`. Apple Pay / Google Pay surface; `link` is persisted but not displayed (the dedicated `WC_Payment_Token_Link` already covers pure-Link UI).
3. `normalize_payment_method_label` (already hooked on `woocommerce_get_credit_card_type_label`) re-looks-up the inner brand after the WC template's lowercase+ucwords pass — without that, `(visa)` survives uncapitalized because `(` isn't a `ucwords` word boundary.

## Testing instructions

1. Enable Apple Pay / Google Pay in Stripe settings.
2. Purchase a subscription via the Apple Pay or Google Pay button on standard checkout.
3. Visit **My Account → Payment Methods** and verify the saved token reads "Google Pay (Visa) ending in 4242".
4. Repeat with a manually entered card and confirm it still reads "Visa ending in 4242".
5. Repeat step 2 with the **same card** through the wallet to confirm the duplicate-refresh path backfills wallet branding on older tokens.

---

- [x] Covered with tests

### Changelog entry

- [x] Updated `changelog.txt` and `readme.txt`.

**Post merge**


daledupreez (commented):
@wjrosa, the PR is working as described, but the updated descriptions are _only_ displayed with the new text in _My account_ > _Payment methods_. I see only the card details in both classic and block checkout, which ends up being more confusing to me.

#### Apple Pay and Google Pay cards in classic checkout

Image

#### Apple Pay and Google Pay cards in block checkout

Image

I think we should aim to provide consistent descriptions as part of this PR.

wjrosa:
Pushed 5127eff06 - overrode `WC_Stripe_Payment_Token_CC::get_display_name()` so wallet-tokenized cards render as "Apple Pay (Visa) ending in 4242 (expires 03/27)" on classic + Blocks checkout (both call into the token's display name), matching the My Account label. 

Manual cards still go through the parent. Pulled the wallet-slug → label map onto the token (`get_wallet_brand_label()`) so the saved-methods filter and the display-name override read from one place.

includes/payment-tokens/class-wc-stripe-cc-payment-token.php
```diff
@@ -23,12 +23,74 @@ class WC_Stripe_Payment_Token_CC extends WC_Payment_Token_CC implements WC_Strip
...
+	 * @inheritDoc
+	 */
+	public function get_display_name( $deprecated = '' ) {
+		$wallet_label = $this->get_wallet_brand_label();
+		if ( '' === $wallet_label ) {
+			return parent::get_display_name( $deprecated );
+		}
+
+		return sprintf(
+			/* translators: 1: wallet brand 2: card brand 3: last 4 digits 4: expiry month 5: expiry year */
```
daledupreez: Not blocking: it may be worth including examples to make it easier for translators.
```suggestion
			/* translators: 1: wallet brand e.g. "Apple Pay", "Google Pay"; 2: card brand e.g. Visa; 3: last 4 digits; 4: expiry month; 5: expiry year */
```

daledupreez (approved):
This is testing as described, though the minor thing that was not great was that the shopper-facing subscription details page was not showing the wallet brand. I _think_ that's being addressed by one of the other PRs, however, so I am going to approve.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3623,6 +3623,13 @@ public function handle_saving_payment_method( WC_Order $order, stdClass $payment
 		if ( $found_token ) {
 			// Update the token with the new payment method ID.
 			$payment_method_instance->update_payment_token( $found_token, $payment_method_object->id );
+
+			// `update_payment_token` only refreshes the PM id, so backfill wallet_type
+			// when reusing a pre-existing token (e.g. same card paid via a wallet).
+			if ( $found_token instanceof WC_Stripe_Payment_Token_CC && isset( $payment_method_object->card ) ) {
+				$found_token->set_wallet_type( (string) ( $payment_method_object->card->wallet->type ?? '' ) );
+				$found_token->save();
```
Mayisha, line 3631: Not a major issue but I found one scenario when this backfill overwrites an existing type.

The scenario is-
- Save a card in My Account or purchase a subscription with the card (I used `4000056655665556`).
- Confirm that the card is present in My Account

Image

- Now purchase a subscription with Google Pay. Select the same card from the Google Pay modl.
- In My Account page, notice that the card type is changes to the wallet type

Image

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -85,9 +85,16 @@ public function normalize_payment_method_label( $label ) {
...
 			case 'sepa iban':
 				return 'SEPA IBAN';
-			default:
-				return $label;
 		}
+
+		// The My Account template re-applies wc_get_credit_card_type_label to our wrapped
+		// brand. Its ucwords pass leaves the inner card name lowercase because `(` isn't
+		// a word boundary — restore it here.
+		if ( preg_match( '/^(Apple Pay|Google Pay) \(([a-z][a-z ]*)\)$/', $label, $matches ) ) {
```
Mayisha, line 93: AI feedback: The `normalize_payment_method_label` regex ^(Apple Pay|Google Pay) \(([a-z][a-z ]*)\)$ assumes the inner brand arrives all lowercase. That only holds for single-word brands. (flagged `cartes_bancaires`)

Mayisha (commented):
Works as described 👍 

I flagged 2 issues worth checking but not a blocker.

Mayisha:
@wjrosa I submitted my comments after the merge. Please check if you would like any action items from them.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Confirmed locally — when a saved card without wallet branding is later used through a wallet sheet, the duplicate-match path overwrites `wallet_type` on the existing token. Same shape exists in two places: `WC_Stripe_UPE_Payment_Gateway::handle_saving_payment_method` (the order-save path) and `WC_Stripe_Payment_Tokens::add_token_to_user` (the token-sync path).

Taking the stricter contract: `wallet_type` reflects how the token was originally created and isn't refreshed on the duplicate-match path. Legacy tokens without `wallet_type` won't get backfilled by a subsequent wallet payment, but the alternative is silently re-badging the shopper's saved card — which is the worse failure mode for what's effectively a display field. Spinning up a follow-up PR off develop with the change in both spots and the regression test.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: Good catch. Traced it: WC core's `wc_get_credit_card_type_label` runs strtolower + ucwords on our wrapped "Apple Pay (Cartes Bancaires)", which becomes "Apple Pay (cartes Bancaires)" — `(` isn't a word boundary so the first inner letter stays lowercase, but the second word gets re-capitalized. The `[a-z][a-z ]*` group then rejects the capital B and the repair never fires. Widening the inner group to `[^)]+` and letting `wc_get_credit_card_type_label` normalize from there. Same follow-up PR as the wallet_type fix.

wjrosa:
@mayisha thanks for digging in after the merge — both are real. Spinning up a follow-up PR off develop now with the fix in both refresh paths plus the regex widening. Will tag you on it.

PR #5440, opened 2026-05-14
Add initial group of existing classes to autoloader

daledupreez:
Towards STRIPE-504

This PR adds a number of files to our Composer autoloader classmap. In most of these classes, there are no references to other classes in the plugin, so we can safely move these classes due to the lack of dependencies. The only minor exception is for the express checkout state mapping helpers, but we're moving both of the classes into the autoloader, so no major issue there.

In addition to adding the classes to the Composer classmap, this PR also implements a new `woocommerce_stripe_init_autoloader()` function in `woocommerce-gateway-stripe.php` that loads the autoload file. For now, this will simply fail if the autoload file is not present -- we may need to add a more graceful failure mode. This is so we can ensure the autoloader is present both during standard plugin loads and from the plugin deactivation code.

Finally, this PR also removes any requires and includes for the files that have been added to the autoloader.

Note that this PR and its follow-ups will eventually help us to clean up our plugin initialisation code, as we'll have more flexibility in refactoring our code.

## Testing instructions

* All tests should continue to pass, especially e2e tests

I think we need to discuss whether additional regression testing is needed for this PR, as the classes being switched are pretty isolated.
 

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -


woocommerce-gateway-stripe.php
```diff
@@ -76,16 +76,28 @@ function woocommerce_stripe_wc_not_supported() {
...
 }
 
+/**
+ * Initialize the autoloader for the plugin.
+ *
+ * @since 10.8.0
+ * @return void
+ */
+function woocommerce_stripe_init_autoloader(): void {
+	static $autoloader_initialized = false;
```
wjrosa: Nit: the static flag is redundant — `require_once` already short-circuits on repeat calls.

woocommerce-gateway-stripe.php
```diff
@@ -76,16 +76,28 @@ function woocommerce_stripe_wc_not_supported() {
...
+	if ( $autoloader_initialized ) {
+		return;
+	}
+
+	require_once WC_STRIPE_PLUGIN_PATH . '/vendor/autoload.php';
```
wjrosa: Worth tracking the graceful-failure follow-up you flagged in the description: `wcstripe_deactivated()` now also fatals if `vendor/` is missing, so a dev who clones without `composer install` cannot even cleanly deactivate.

wjrosa (commented):
Two small notes inline. Otherwise the move is mechanical and looks safe — `includes/constants/` as a directory entry picks up the lazy-evaluated `const STATES = WC_Stripe_Express_Checkout_Button_States::STATES` cross-reference fine via autoload.

woocommerce-gateway-stripe.php
daledupreez: I've made the deactivation path more resilient against the autoloader not being present in bdaee0591fc2822f71e6b483eb73a12da421f095.

woocommerce-gateway-stripe.php
daledupreez: As part of making deactivation more resilient in bdaee0591fc2822f71e6b483eb73a12da421f095, I've added some conditional logic here that means the static usage is more relevant.

wjrosa (approved):
Looks good to me! Thanks for addressing my points

PR #5441, opened 2026-05-14
Normalize $links input in add_release_notes_link()

wjrosa:
Follow-up to #5399 (Dale's [unaddressed review comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5399#discussion_r3243772839)).

`WC_Stripe_Plugins_Page_Controller::add_release_notes_link()` is wired to the WordPress `plugin_row_meta` filter and declares `: array` as its return type. The pending-update early-return and the fall-through `$links['wc_stripe_release_notes'] = …` assignment both assume `$links` is an array — but the filter is shared across callbacks, and an upstream callback can replace the value with anything (`null`, `false`, a scalar, an object). In that case the method would either fatal on offset-assign or violate its own `: array` contract.

Cast `$links` to `(array)` once at the top, drop the redundant cast on the basename-mismatch early return, and broaden the `@param` from `array` to `mixed` with a note explaining why we normalize. The return type stays strict.

## Testing instructions

1. Run `npm run test:php -- --filter=WC_Stripe_Plugins_Page_Controller_Test` and verify the new `test_add_release_notes_link_normalizes_non_array_links` data provider passes for `null`, `false`, `''`, `'unexpected'`, and `0`.
2. Optional smoke test: add a temporary `plugin_row_meta` filter that returns `null` (e.g. via a mu-plugin) and confirm the WooCommerce Stripe plugin row still renders the Release notes link.

### Automated checks

- `npm run test:php -- --filter=WC_Stripe_Plugins_Page_Controller_Test` — 12 specs pass
- `npm run phpstan` — no errors
- `./vendor/bin/phpcs <touched files>` — clean

---

-   [x] Tested on mobile (or does not apply) — admin-only

### Changelog entry

-   [x] This Pull Request does not require a changelog entry. Internal hardening of #5399 in the same unreleased version (10.8.0); no user-visible behavior change.

**Post merge**


includes/admin/class-wc-stripe-plugins-page-controller.php
```diff
@@ -123,13 +123,16 @@ private function get_plugin_basename(): string {
 	 * The link reuses WordPress' built-in plugin information modal, opened on the
 	 * changelog tab via thickbox (already enqueued by `enqueue_scripts`).
 	 *
-	 * @param array  $links Existing plugin row meta links.
+	 * @param mixed  $links Existing plugin row meta links. Normalized to an array because
+	 *                      other `plugin_row_meta` filter callbacks may return non-array values.
```
daledupreez: I don't think the normalization comment is required. If some other plugin implements a bad filter, we can't be held responsible if we clean it up come other way.
```suggestion
	 * @param mixed  $links Existing plugin row meta links.
```

tests/phpunit/admin/class-wc-stripe-plugins-page-controller-test.php
```diff
@@ -137,6 +137,43 @@ public function test_add_release_notes_link_skipped_when_update_is_pending(): vo
...
+	 *
+	 * @param mixed $links Value a misbehaving upstream filter could return.
+	 *
+	 * @return void
+	 */
+	public function test_add_release_notes_link_normalizes_non_array_links( $links ): void {
+		$controller = $this->get_mock_controller();
+
+		$other_plugin = $controller->add_release_notes_link( $links, 'some-other/some-other.php' );
+		$this->assertIsArray( $other_plugin );
```
daledupreez, line 155: Is it worth asserting that array doesn't include the `wc_stripe_release_notes` key?
```suggestion
		$this->assertIsArray( $other_plugin );
		$this->assertArrayNotHasKey( 'wc_stripe_release_notes', $other_plugin );
```

daledupreez (approved):
The changes look good to me, though I think we can tidy up a bit.

PR #5443, opened 2026-05-14
Refactor Docker setup to be worktree-friendly

malithsen:

Ports the Docker worktree-friendly refactor from [Automattic/woocommerce-payments#11291](https://github.com/Automattic/woocommerce-payments/pull/11291) so multiple `git worktree`s can run their own WordPress instances in parallel without port/database collisions.

**Architecture**

```
┌──────────────────────────── SHARED INFRASTRUCTURE ────────────────────────────┐
│                       (docker-compose.infra.yml, run once)                    │
│  ┌────────────────────────────┐    ┌────────────────────────────────────┐     │
│  │   MariaDB (wcstripe_db)    │    │  phpMyAdmin (wcstripe_phpmyadmin)  │     │
│  │   localhost:5668           │    │  http://localhost:8073             │     │
│  │   Databases:               │    └────────────────────────────────────┘     │
│  │   • wordpress (site data)  │                                               │
│  │   • wcstripe_tests_default │    Shared bind-mount volumes:                 │
│  │   • wcstripe_tests_<id>... │    wcstripe-{plugins,themes,uploads,          │
│  │     (one per worktree)     │                   mu-plugins}                 │
│  └────────────────────────────┘                                               │
└──────────────────────────────────┬────────────────────────────────────────────┘
                                   │ wcstripe-network
        ┌──────────────────────────┼──────────────────────────┐
        ▼                          ▼                          ▼
┌───────────────┐         ┌───────────────┐         ┌───────────────┐
│ Main checkout │         │ Worktree A    │         │ Worktree B    │
│ WP container  │         │ WP container  │         │ WP container  │
│  :8072        │         │  :8170        │         │  :8171        │
│ WORKTREE_ID=  │         │ WORKTREE_ID=  │         │ WORKTREE_ID=  │
│   default     │         │  feature_a    │         │  feature_b    │
└───────────────┘         └───────────────┘         └───────────────┘
```

Each worktree gets:
- Its own WP container (`wcstripe_wp_<id>`) on its own port
- Its own `wp-config.php` (under `./docker/wordpress/`)
- Its own test database (`wcstripe_tests_<id>`)
- The same `WORKTREE_ID` echoed back in a small admin-bar mu-plugin so parallel instances are easy to tell apart

Worktrees share:
- The MariaDB instance and the WP site database
- `wp-content/{plugins,themes,uploads,mu-plugins}` via four Docker bind volumes pointed at the main checkout's `docker/wordpress/wp-content/`

**Stripe-specific differences from the WooPayments PR**

- Naming prefix is `wcstripe-` (network/volumes) and `wcstripe_` (container names)
- Worktree port range is 8170–8189 (WCPay uses 8180–8199) to keep distance from this repo's main port 8072.
- Preserved Stripe-specific volumes that WCPay doesn't have: `node_modules`, `phpstan_stubs`, `release_directory`, `docker_directory`, the `./tests/phpunit` bind, the `dev-php.ini` / `override-php-error-logging.ini` PHP overlays, and `./docker/bin:/var/scripts`.
- `bin/run-tests.sh` and `bin/run-tests-parallel.sh` both pass `wcstripe_tests_${WORKTREE_ID}` to `install-wp-tests.sh`. WCPay only ports the parallel runner. We need both
- `bin/jurassic-tube-setup.sh` is **not** modified - this repo's JT setup doesn't grep `docker ps` for the port the way WCPay's does, so there was nothing to wire. JT remains a main-checkout-only flow for now.

**New npm scripts**

| Script | What it does |
| --- | --- |
| `npm run infra:up` | Start shared db + phpMyAdmin + bind volumes (run from main checkout). |
| `npm run infra:down` | Stop shared infra (preserves volumes). |
| `npm run worktree:setup` | Write `.env` with `WORKTREE_ID` and an unused `WORDPRESS_PORT`. |
| `npm run worktree:status` | List all worktrees with port, URL, container state; warn about orphan containers. |
| `npm run worktree:cleanup` | Stop this worktree's container, drop its test DB, remove `.env`. Run before `git worktree remove`. |
| `npm run up` | Bring up this worktree's WordPress container (idempotent). |
| `npm run up:recreate` | Force-recreate this worktree's WordPress container; auto-starts shared infra if needed. |
| `npm run down` | Stop this worktree's WordPress container only. |

`wp` and `listen` are rewired to read `.env` so they target this worktree's container/port.

### 1. Main checkout (first-run)

- From the main checkout, run `npm run up:recreate`.
- Verify `.env` is created with `WORKTREE_ID=default` and `WORDPRESS_PORT=8072`.
- Verify the site loads at `http://localhost:8072/wp-admin/` (login `admin` / `admin`).
- Verify the admin bar shows `[default]` (the worktree indicator).
- Verify `docker ps` shows `wcstripe_db`, `wcstripe_phpmyadmin`, and `wcstripe_wp_default` running.
- Verify phpMyAdmin loads at `http://localhost:8073`.

### 2. Create a second worktree

```
git worktree add ../wcstripe-feature-x dev/worktree-friendly-setup
cd ../wcstripe-feature-x
npm install
composer install --prefer-dist --ignore-platform-req=php
npm run up:recreate
```

- Verify `.env` in the new worktree has a `WORKTREE_ID=wcstripe_feature_x` and an auto-assigned `WORDPRESS_PORT` (first free in 8170–8189).
- Verify a second WP container `wcstripe_wp_wcstripe_feature_x` starts.
- Verify the new site loads at `http://localhost:<port>/wp-admin/`.
- Verify the admin bar there shows `[wcstripe_feature_x]`.
- Verify the shared db and phpMyAdmin were **not** restarted.

### 3. Worktree-aware tests

From each worktree:

```
npm run test:php -- --filter=WC_Stripe_Helper_Test
```

- Verify the install banner says `Using test database: wcstripe_tests_<id>` with the correct id per worktree.
- Verify the suite runs to completion (parallel runs across worktrees should not collide on the test DB).

### 4. Status + cleanup

From any worktree:

```
npm run worktree:status
```

- Verify two rows are shown (main + feature worktree), each with their port, URL, and `running` status.

In the test worktree:

```
npm run worktree:cleanup
cd ..
git worktree remove wcstripe-feature-x
```

- Verify the container `wcstripe_wp_wcstripe_feature_x` is stopped.
- Verify `wcstripe_tests_wcstripe_feature_x` is dropped from the shared db.
- Verify `.env` is removed.
- Verify `npm run worktree:status` now shows only the main checkout.
- Verify `git worktree remove` succeeds without errors.

### 5. Stop / restart

- `npm run down` from the main checkout stops only the main WP container.
- `npm run up` brings it back without recreating; site state is preserved.
- `npm run infra:down` stops the shared db + phpMyAdmin; volumes/network remain so `npm run up` reattaches cleanly afterwards.

### 6. PHPStan and JS lint (regression tests)

- Run `npm run phpstan` and `npm run lint:js` - neither should regress.
---

-   [ ] Tested on mobile (or does not apply) — does not apply (dev tooling).

### Changelog entry

<details>

#### Comment

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply) — does not apply (dev tooling, never executes in release).
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply) — does not apply (developer-facing only; covered by `docs/DOCKER.md` and `AGENTS.md` updates in this PR).
-   [ ] Included this PR in the Release Thread scope (or does not apply) — does not apply.


bin/docker-infra-up.sh
```diff
@@ -0,0 +1,93 @@
...
+if git rev-parse --is-inside-work-tree &>/dev/null; then
+    GIT_DIR=$(git rev-parse --git-dir)
+    GIT_COMMON_DIR=$(git rev-parse --git-common-dir)
+    if [[ "$GIT_DIR" != "$GIT_COMMON_DIR" ]]; then
+        echo ""
+        echo "WARNING: Running infra:up from a worktree."
+        echo "   If this is not intended, run infra:up from your main checkout instead."
+        echo ""
+        read -p "Do you want to continue? (y/N) " -r
+        if [[ ! ${REPLY,,} =~ ^y(es)?$ ]]; then
```
wjrosa: `${REPLY,,}` (lowercase expansion) is bash 4+. macOS ships `/bin/bash` 3.2 by default, so this will error out. Same pattern lives in `docker-worktree-cleanup.sh`. Could swap to `[[ ! $REPLY =~ ^[Yy] ]]`.

bin/docker-worktree-cleanup.sh
```diff
@@ -0,0 +1,41 @@
...
+
+set -e
+
+WORKTREE_ID="default"
+if [[ -f ".env" ]]; then
+    source .env
+else
+    echo "No .env file detected, which means you might not be on a worktree."
+    read -p "Do you want to continue? (y/N) " -r
+    if [[ ! ${REPLY,,} =~ ^y(es)?$ ]]; then
```
wjrosa: Same bash 3.2 compat issue with `${REPLY,,}` as `docker-infra-up.sh:17`.

bin/docker-worktree-cleanup.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+# bin/docker-worktree-cleanup.sh
+# Run this before `git worktree remove` to clean up Docker resources.
+
+set -e
+
+WORKTREE_ID="default"
+if [[ -f ".env" ]]; then
+    source .env
```
wjrosa, line 28: If this runs from the main checkout (which always has `.env` with `WORKTREE_ID=default`), it'll silently stop the main WP container, drop `wcstripe_tests_default`, and delete `.env` — no prompt, since the prompt only fires when `.env` is missing. Worth an `is_main_checkout` guard (or at least a confirmation) before the destructive steps.

bin/docker-port-setup.sh
```diff
@@ -0,0 +1,77 @@
...
+        return 0
+    fi
+    local git_dir git_common_dir
+    git_dir=$(git rev-parse --git-dir)
+    git_common_dir=$(git rev-parse --git-common-dir)
+    [[ "$git_dir" == "$git_common_dir" ]]
+}
+
+get_reserved_ports() {
+    git worktree list 2>/dev/null | cut -d' ' -f1 | while read -r dir; do
```
wjrosa: `git worktree list | cut -d' ' -f1` breaks if a worktree path contains spaces. `docker-worktree-status.sh` already uses `--porcelain` with a `^worktree ` regex — same pattern would be safer here.

changelog.txt
```diff
@@ -9,6 +9,7 @@ xxxx-xx-xx - version 10.8.0
 * Add - Append a "what's new" changelog link to the Updated! message after manually updating the plugin from the Plugins page
 * Add - Add a "Release notes" link to the WooCommerce Stripe row on the Plugins page that opens the changelog modal
 * Fix - Add guards against invalid values for webhook state timestamps
+* Dev - Refactor docker setup to be docker friendly
```
wjrosa: Tiny wording nit: "docker friendly" reads odd given the change *is* to docker setup. Looks like "worktree-friendly" (per the PR title) was intended.

readme.txt
```diff
@@ -160,5 +160,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Add - Append a "what's new" changelog link to the Updated! message after manually updating the plugin from the Plugins page
 * Add - Add a "Release notes" link to the WooCommerce Stripe row on the Plugins page that opens the changelog modal
 * Fix - Add guards against invalid values for webhook state timestamps
+* Dev - Refactor docker setup to be docker friendly
```
wjrosa: Same wording as `changelog.txt:12` — likely meant to be "worktree-friendly".

wjrosa (commented):
Nice port of the WCPay worktree setup — architecture and docs are clear. A few small things, mostly bash 3.2 compat and a safety nit on cleanup.

bin/docker-infra-up.sh
malithsen: Hm, I don't think this is correct. MacOS has been coming with zsh by default for a while now. While bash is available, bash 4.0 was released in 2009. So I don't think we should worry about older versions.

bin/docker-worktree-cleanup.sh
malithsen: ~I don't think this is an issue. See previous reply~ This is, in fact, an issue. Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/8c03b6db3a67fa331264f70541722d837c1244a7

bin/docker-infra-up.sh
malithsen: ah nvm, I stand corrected. User default shell (zsh) doesn't matter here. And apparently, `/bin/bash`, what the shebang in these scripts actually invokes is still `3.2.57` on every macOS for the past ~15 years 😮 I have a newer version of `bash` in `/usr/local/bin/bash` so my `bash --version` check wasn't showing the actual version the script is using.

I was also able to replicate the error. Will fix.

changelog.txt
malithsen: 🤦 
Thank you!
Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/8a12c090561af29555a30e3b7027edf5cbe67a50

readme.txt
malithsen: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/8a12c090561af29555a30e3b7027edf5cbe67a50

bin/docker-port-setup.sh
malithsen: I didn't know that worktrees supported spaces. Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/381e2204bedda2edc20855b85e4400e3a9e319b0

bin/docker-worktree-cleanup.sh
malithsen: Good catch. Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/4d24b4811c98009980013fdd7ad8c1201d9e56ea

bin/docker-infra-up.sh
malithsen: Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/8c03b6db3a67fa331264f70541722d837c1244a7

Tested it by
1. Creating a new worktree
2. `npm run infra:up`
3. Select 'N' when the warning is shown.
4. There should not be an error.

wjrosa (approved):
Code is good and works as expected! Thanks for addressing my points

bin/docker-worktree-status.sh
```diff
@@ -0,0 +1,111 @@
...
+
+echo ""
+echo -e "${BOLD}Worktree Status${NC}"
+echo ""
+printf "  ${BOLD}%-6s %-30s %-14s %s${NC}\n" "PORT" "URL" "STATUS" "NAME"
+
+orphan_containers=()
+
+for wt_path in "${worktrees[@]}"; do
+    wt_name=$(basename "$wt_path")
```
daledupreez: Nit: I think it would be clearer to use `worktree_path` and `worktree_name` instead of the `wt_*` names. (Not blocking.)

bin/docker-worktree-status.sh
```diff
@@ -0,0 +1,111 @@
...
+done
+
+if [[ ${#orphan_containers[@]} -gt 0 ]]; then
+    echo ""
+    echo -e "${YELLOW}Warnings:${NC}"
+    for orphan in "${orphan_containers[@]}"; do
+        echo "  - Orphan container: $orphan (no matching worktree)"
+    done
+    echo ""
+    echo "  To clean up: docker stop <name> && docker rm <name>"
```
daledupreez: Minor suggestion: might it be worth actually echoing a command to clean them up? If we want to get fancy, we could even try to generate a one-line command split by `;` to make each sub-command independent.
```suggestion
    orphan_cleanup=()
    for orphan in "${orphan_containers[@]}"; do
        echo "  - Orphan container: $orphan (no matching worktree)"
        orphan_cleanup+=("docker stop $orphan && docker rm $orphan")
    done
    echo ""
    echo "  To clean up: "
    printf "%s; " "${orphan_cleanup[@]}"
```

daledupreez:
I am getting the following error when running `npm run up:recreate`:
```
mariadb-check: Got error: 2005: Unknown server host 'wcstripe_db' (-2) when trying to connect
```

Not sure what the cause is, but wanted to flag it.

malithsen:
> mariadb-check: Got error: 2005: Unknown server host 'wcstripe_db' (-2) when trying to connect

@daledupreez To confirm, do you see this a few times and then setup proceeds, or does it loop indefinitely?

I too see the error, but it's transient. We [poll for db readiness](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/changes#diff-52801e42af3a1577bbe6903d592452deb0b349189d29e25e412124f6c402b4a2L28-L31) while the db container spins up, so it's expected to see this error a few times and is harmless

If it is getting stuck in a loop for you, that's a different problem, and I'd love to dig in.

daledupreez:
@malithsen, I just retried and it went on for a few minutes before I killed the process. That was after ~100 checks. It's possible this is an issue with my container engine, but it doesn't feel super likely.

bin/docker-infra-down.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# bin/docker-infra-down.sh
+# Stops shared infrastructure (database, phpMyAdmin).
+# Does NOT remove the named volumes or the wcstripe-network so per-worktree
+# WordPress containers can be restarted later without rebuilding everything.
+
+set -e
+
+echo "Stopping shared infrastructure..."
+docker compose -f docker-compose.infra.yml down
```
daledupreez, line 30: The YAML file expects `WCSTRIPE_SHARED_WP_PATH` and `WCSTRIPE_DB_DATA_PATH` to be defined, but we're not loading any files to ensure that the variable is defined. I think we should load `default.env` for this, but we may need more. (Remember that this command may be run from a different terminal from the location where the up command was run from.)

docker-compose.infra.yml
```diff
@@ -0,0 +1,60 @@
...
+    driver: local
+    driver_opts:
+      type: none
+      o: bind
+      device: ${WCSTRIPE_SHARED_WP_PATH}/wp-content/mu-plugins
+
+services:
+  db:
+    container_name: wcstripe_db
+    image: mariadb:10.5.8
```
daledupreez: This was the cause of my issues -- in the original docker file we were using `mariadb:latest`, and my instance was refusing to start because log files from a newer version were in `docker/data/`. Maybe we should shift this back to `mariadb:latest` to match the old behaviour and avoid downgrade issues like the one I tripped over.

bin/docker-setup.sh
```diff
@@ -14,24 +28,48 @@ redirect_output() {
...
+
+# Refresh WP core files to align with whatever version `wp core update` last pulled.
+echo "Refreshing WordPress core files..."
+cli wp core download --force --skip-content --path=/var/www/html --quiet 2>/dev/null || true
+
+# wp-config.php settings are per-worktree (file-based) — always set them
+# (each worktree's container has its own wp-config.php under ./docker/wordpress).
+echo "Configuring wp-config.php for this worktree..."
+cli wp config set DOCKER_HOST "\$_SERVER['HTTP_X_FORWARDED_HOST'] ?? \$_SERVER['HTTP_X_ORIGINAL_HOST'] ?? \$_SERVER['HTTP_HOST'] ?? 'localhost'" --raw
+docker exec $WP_CONTAINER bash -c "grep -q '\\\$_SERVER\[.HTTP_HOST.\] = DOCKER_HOST' /var/www/html/wp-config.php || sed -i \"/define.*'DOCKER_HOST'/a \\\\\\\$_SERVER['HTTP_HOST'] = DOCKER_HOST;\" /var/www/html/wp-config.php"
```
daledupreez: I think I am now tripping over a permissions issue on this line:
```
sed: couldn't open temporary file /var/www/html/sedkeNrOZ: Permission denied
```

bin/docker-setup.sh
```diff
@@ -14,24 +28,48 @@ redirect_output() {
...
+
+# Refresh WP core files to align with whatever version `wp core update` last pulled.
+echo "Refreshing WordPress core files..."
+cli wp core download --force --skip-content --path=/var/www/html --quiet 2>/dev/null || true
+
+# wp-config.php settings are per-worktree (file-based) — always set them
+# (each worktree's container has its own wp-config.php under ./docker/wordpress).
+echo "Configuring wp-config.php for this worktree..."
+cli wp config set DOCKER_HOST "\$_SERVER['HTTP_X_FORWARDED_HOST'] ?? \$_SERVER['HTTP_X_ORIGINAL_HOST'] ?? \$_SERVER['HTTP_HOST'] ?? 'localhost'" --raw
+docker exec $WP_CONTAINER bash -c 'CONFIG=/var/www/html/wp-config.php; grep -q "\$_SERVER\[.HTTP_HOST.\] = DOCKER_HOST" "$CONFIG" || { sed "/define.*'\''DOCKER_HOST'\''/a \\\$_SERVER['\''HTTP_HOST'\''] = DOCKER_HOST;" "$CONFIG" > /tmp/wp-config.new && cat /tmp/wp-config.new > "$CONFIG" && rm /tmp/wp-config.new; }'
```
daledupreez: I got this working locally with the following command that uses a variable instead of a temp file.
```suggestion
docker exec $WP_CONTAINER bash -c "WP_CONFIG_CONTENTS=\"\$( cat /var/www/html/wp-config.php )\" && echo \"\$WP_CONFIG_CONTENTS\" | grep -q '\\\$_SERVER\[.HTTP_HOST.\] = DOCKER_HOST' || ( echo \"\$WP_CONFIG_CONTENTS\" | sed \"/define.*'DOCKER_HOST'/a \\\\\\\$_SERVER['HTTP_HOST'] = DOCKER_HOST;\" > /var/www/html/wp-config.php )"
```

malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/b743ffa3ce6f85e9fdbede2a859d8dfbbe7e7186

bin/docker-setup.sh
malithsen: Switch to variables in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/b743ffa3ce6f85e9fdbede2a859d8dfbbe7e7186

bin/docker-infra-down.sh
daledupreez: I think this qualifies as a blocker for getting this merged. Trying to stop the infrastructure from another terminal fell over in a heap for me.

docker-compose.infra.yml
malithsen: I think having `mariadb:latest` is what sets us up for this failure mode. It's a floating tag, so if we pull a fresh image months later and get a different major, we'll run into the same issue again. So I've pinned it to the current major in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/353ed120b43881961ec94115577d0cc23e07963a
I also added a note to DOCKER.md

bin/docker-infra-down.sh
malithsen: Agreed. Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/changes/ec30300039c5589f4b77148fdc6ae9478443c238

bin/docker-worktree-status.sh
malithsen: Good idea. I included this in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/a7b01eeadff27156c93c5abe5b739768986cee21 along with variable renames.

bin/docker-worktree-status.sh
malithsen: Makes sense. Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/a7b01eeadff27156c93c5abe5b739768986cee21

bin/docker-infra-down.sh
```diff
@@ -0,0 +1,25 @@
...
+# Load default env so docker compose can interpolate ${WCSTRIPE_SHARED_WP_PATH}
+# and ${WCSTRIPE_DB_DATA_PATH} when stopping infra from any shell. `set -a`
+# exports each sourced variable so docker compose (a child process) sees them.
+set -a
+source default.env
+[[ -f ".env" ]] && source .env
+set +a
+
+echo "Stopping shared infrastructure..."
+docker compose -f docker-compose.infra.yml down
```
daledupreez, line 30: Follow-up nit: the `WCSTRIPE_SHARED_WP_PATH` and `WCSTRIPE_DB_DATA_PATH` values are relative, so `npm run infra:down` from a worktree doesn't stop the shared entities because we look at the worktree-local file paths. I think we need to mimic the resolution steps used in `bin/docker-infra-up.sh` for these two values.

docker-compose.infra.yml
daledupreez: No strong opinions on my side, but older logs are generally acceptable to newer versions (provided the version jump has not been too large), so picking up a newer version is mostly going to be OK. That said, pinning to a specific version should be more stable.

All that said, it looks like the 11.8 LTS release went GA about a year ago, so it may be better to use 11.8 here. (I also see that 12.3 went GA last week in 12.3.2, but we may want to wait before we switch to that version!)

daledupreez (approved):
Thanks for addressing the last few issues, this is now working well for me overall.

I do think we need to fix the infra-down issue, but that could be in this PR or a follow-up.

I also think we should strongly consider switching to MariaDB 11.8, but that can also be a follow-up.

docker-compose.infra.yml
malithsen: Pinned to 11.8 in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/b960fe34fb40bd74b8e411979555b2651d9214ed

bin/docker-infra-down.sh
malithsen: Thanks. Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5443/commits/b960fe34fb40bd74b8e411979555b2651d9214ed

PR #5444, opened 2026-05-14
Add foundation classes for Payment Plugins for Stripe settings migration

pierorocca:
## Summary

PR 1 of 2 — foundation classes for the PP→Woo Stripe settings migration triggered at OAuth completion. This PR is **non-load-bearing**: nothing here is wired in yet. PR 2 (`agent/stripe-1146-orchestrator`) adds the orchestrator and the one-line OAuth wiring on top.

Splitting into two PRs lets reviewers verify the data model (snapshot blob shape, ledger row shape, map row shape) and the per-class behavior independently before reviewing how the orchestrator uses them.

## What's in this PR

| Class | Purpose |
| --- | --- |
| `WC_Stripe_PP_Settings_Map` (abstract) | Contract for per-major-version mapping tables. |
| `WC_Stripe_PP_Settings_Map_3X` | Concrete map for PP v3.X — 9 AUTO + 5 TRANSFORM rows; DROP/INVESTIGATE/BUILD lists. |
| `WC_Stripe_PP_Settings_Map_4X` | Placeholder for PP 4.X; orchestrator falls back to 3.X best-effort. |
| `WC_Stripe_PP_Version_Detector` | Three-layer fallback: PP version option → plugin header → option-shape sniff. |
| `WC_Stripe_Pre_Migration_Snapshot` | Full snapshot of all Stripe-related options (PP + Woo Stripe) for forensic / rollback purposes. |
| `WC_Stripe_Settings_Migration_Ledger` | Append-only per-row decision log with secret-keyed value redaction. |

## Design notes

- **Pre-fill, not overwrite.** All AUTO/TRANSFORM rows use a `! empty( $dest[ $key ] )` guard so OAuth-set values and merchant overrides are preserved. The snapshot exists for rollback support, not because pre-fill is risky.
- **Canonical key names.** Destination keys verified against [`includes/admin/stripe-settings.php`](includes/admin/stripe-settings.php). Express checkout uses `express_checkout_*` (not the legacy `payment_request_button_*`); OCS layout uses `optimized_checkout_layout`. The plan's first draft had outdated names; corrected during implementation.
- **Snapshot↔orchestrator contract.** Snapshot hardcodes `COMPLETED_OPTION_NAME = 'woocommerce_stripe_pp_settings_migrated'` and the orchestrator (PR 2) sets the matching `COMPLETED_OPTION` constant. Documented in the snapshot class header so the contract is explicit.
- **Secret redaction by key-name pattern.** Keys matching `secret_key`, `webhook_secret`, `refresh_token`, `access_token`, or ending in `_secret`/`_token` have their values replaced with `__REDACTED__` in the ledger and log file. The row itself is still recorded so the audit trail proves the migration processed the field.

## Test plan

- [x] PHPStan level 8 clean — `npm run phpstan` (no new baseline entries)
- [x] PHPCS clean — `npm run lint:php` (60+ auto-fixes applied during dev)
- [x] PHPUnit clean — `npm run test:php -- --filter='WC_Stripe_PP_Version_Detector_Test|WC_Stripe_Pre_Migration_Snapshot_Test|WC_Stripe_Settings_Migration_Ledger_Test|WC_Stripe_PP_Settings_Map_3X_Test'` → 83 tests, 279 assertions, 0 failures
- [x] Pre-commit hook (`./vendor/bin/phpcs`) passed on every commit

### Coverage highlights

- Version detector: all three layers (option / plugin header / shape sniff), unknown-version path, version-option priority over shape sniff
- Snapshot: known + catch-all option capture, dedicated `woocommerce_gateway_stripe_%` pattern, recursion prevention, archive-on-rerun, autoload=false, missing-vs-empty sentinel handling
- Ledger: apply / skip / drop / error / revert row shapes, run_id grouping, find_by_source_key / find_by_run_id, secret redaction (6 patterns + 2 suffixes), append-only across writes
- 3.X map: row structure validation, canonical-key conformance, every transformer's full domain with @dataProvider, DROP/INVESTIGATE/BUILD disjointness

### Related PRs

- **PR 2 of 2:** `agent/stripe-1146-orchestrator` — adds the orchestrator and wires it into `WC_Stripe_Connect::save_stripe_keys()`. Review and merge this PR first.


pierorocca:
# Review Report — PR 5444 (Foundation)

**Branch:** `agent/stripe-1146-foundation` → `develop`
**Title:** Add foundation classes for Payment Plugins for Stripe settings migration
**Date:** 2026-05-14
**Scope:** 6 new PHP classes (version detector, snapshot, ledger, map abstract + 3.X + 4.X placeholder) + 4 test classes + changelog/readme.
**Blast radius:** Zero — no production code path calls these files. The orchestrator in PR 5445 is what wires them in.

## Overall Verdict

**APPROVE**

All four review dimensions pass. Foundation code is non-load-bearing on its own (nothing calls it), strictly additive, and follows existing migration patterns in `includes/migrations/`. Idempotency, redaction, and forensic-snapshot logic are all sound and ready for the orchestrator to consume.

## Architecture Compliance — PASS

- Class naming follows repo convention (`WC_Stripe_*` prefix, `class-wc-stripe-*.php` files).
- Subdirectory `pp-settings-maps/` mirrors patterns like `payment-methods/`, `payment-tokens/`.
- All classes stay within the migrations layer; no rendering, REST, or admin-UI dependencies.
- Snapshot is read-only by design (header comment line 13 documents the principle).
- Ledger is append-only; redaction occurs at append time so secrets are never persisted in plaintext.
- Version detector implements three-layer fallback (PP version option → plugin header → option shape sniff) for graceful degradation when PP has been uninstalled.

## Security — PASS

- **SQL injection (snapshot LIKE sweep):** `class-wc-stripe-pre-migration-snapshot.php:263-265` correctly parameterizes LIKE queries via `$wpdb->prepare()` with `%s`, with `$wpdb->esc_like()` escaping at line 232. No injection surface.
- **Secret redaction:** Ledger patterns (`secret_key`, `webhook_secret`, `refresh_token`, `access_token` substrings + `_secret`/`_token` suffixes) cover all known Stripe secrets stored by PP and Woo Stripe. Non-secret keys are correctly not redacted.
- **Snapshot raw values:** The snapshot itself is NOT redacted — it stores raw values for forensic use. Acceptable trade-off given `autoload=false` and admin-only access. Documented in class header.
- **No new public-facing endpoints.** Foundation classes are libraries; CSRF/auth surface is zero on this PR.
- **Plugin file path:** Version detector uses hardcoded `woo-stripe-payment/stripe-payments.php` resolved via `WP_PLUGIN_DIR`. No user input; symlink traversal is gated by `file_exists()`. Acceptable risk.

## Performance — PASS

- **Snapshot SQL footprint:** Exactly 4 LIKE queries per capture. No N+1 patterns. Estimated 20-50ms.
- **Autoload correctness:** Snapshot (line 194) and ledger (line 389) both explicitly `update_option(..., false)`. Critical because snapshot blobs can exceed 200KB on heavily-configured PP installs.
- **Ledger growth:** Append-only with predictable shape; ~150-200 bytes per row. After 5 reruns the blob is ~20-30KB. Bounded.
- **No work happens at load time** since nothing calls these classes from production paths in this PR.

## Test Coverage — PASS

- **70+ tests across 4 test files**, all public methods covered.
- Version detector: 11 tests, all 3 detection layers covered, edge cases (empty, malformed, null) handled.
- Snapshot: 16 tests including autoload=false enforcement, archive rotation, infrastructure-key exclusion (recursion prevention), missing-vs-empty-string distinction.
- Ledger: 14 tests including 6-pattern secret redaction, append-only behavior, run_id grouping, unique IDs per row.
- 3.X map: 14 tests covering row structure, destination key canonicality (verified against `stripe-settings.php`), full transformer input domains via `@dataProvider`, categorization disjointness.
- Strong tear-down hygiene; all touched `wp_options` cleaned between tests.

## Suggestions (Non-blocking)

1. **Document the secret redaction pattern list** in the settings-map guide so future maintainers know to update it when new sensitive keys appear.
2. **Periodically audit `KNOWN_OPTION_KEYS`** in `class-wc-stripe-pre-migration-snapshot.php` against new PP/Woo Stripe keys. (Catch-all LIKE sweep is the safety net, but explicit enumeration is faster.)
3. **Future: add a transformer-exception isolation test.** Plan §Test Strategy line 1266 called for a test where one transformer throws and others still run; current tests don't exercise this scenario. Add when the orchestrator lands.

## Required Changes

None.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-3x.php
```diff
@@ -0,0 +1,278 @@
...
+			'merchant_id',
+			'merchant_name',
+			'notice_enabled',
+			'all_browsers',
+			'order_status',
+			'icon',
+			'branded_type',
+			// UPM — managed via Stripe PMC, not configurable in WC admin.
+			'live_payment_method_configuration',
+			'test_payment_method_configuration',
```
daledupreez, line 191: These may belong in the `build` group/rows.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map.php
```diff
@@ -0,0 +1,66 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Defines the per-major-version mapping contract used by the PP settings
+ * migration orchestrator. Concrete subclasses (one per PP major version)
+ * return the AUTO/TRANSFORM/DROP/INVESTIGATE/BUILD row lists.
+ *
+ * @since 9.7.0
```
daledupreez: Nit: The since versions are all way off.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-3x.php
```diff
@@ -0,0 +1,278 @@
...
+	}
+
+	/**
+	 * PP `mode` ("test"/"live") → Woo Stripe `testmode` ("yes"/"no").
+	 *
+	 * @param mixed $value PP source value.
+	 * @return string
+	 */
+	public static function transform_mode_to_testmode( $value ): string {
+		return 'test' === $value ? 'yes' : 'no';
```
daledupreez, line 226: It feels like overkill to have dedicated callbacks for basic mappings like this. It feels like having an inline `map` array in the configuration could be plenty for these simple cases.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map.php
```diff
@@ -0,0 +1,66 @@
...
+	 * @return array<int, array<string, mixed>>
+	 */
+	abstract public function get_transform_rows(): array;
+
+	/**
+	 * DROP rows: documented no-ops. Returned for ledger/audit visibility; no code path executes for these.
+	 *
+	 * @return array<int, string> PP setting names that we explicitly do not migrate.
+	 */
+	abstract public function get_dropped_rows(): array;
```
daledupreez, line 51: Not blocking, but it feels like "skipped" might be a better description, as "drop" could imply being removed, especially for folks familiar with SQL.

includes/migrations/class-wc-stripe-pp-version-detector.php
```diff
@@ -0,0 +1,142 @@
...
+			$version = self::from_plugin_header();
+		}
+
+		if ( null === $version ) {
+			$version = self::from_option_shape();
+		}
+
+		if ( null === $version || '' === $version ) {
+			return self::VERSION_UNKNOWN;
+		}
```
daledupreez: Nit: it feels like it would be simpler to have this be in a private getter function where we could return early for the first non-null value. This fallthrough logic is hard to follow.

EDIT: Maybe we can just call `self::detect_full_version()`?
```suggestion
		$version = self::detect_full_version();

		if ( null === $version || '' === $version ) {
			return self::VERSION_UNKNOWN;
		}
```

includes/migrations/class-wc-stripe-pp-version-detector.php
```diff
@@ -0,0 +1,142 @@
...
+	 * @var string
+	 */
+	const PP_VERSION_OPTION = 'stripe_wc_version';
+
+	/**
+	 * Sentinel returned when no PP version can be determined.
+	 *
+	 * @var string
+	 */
+	const VERSION_UNKNOWN = 'unknown';
```
daledupreez: Nit: I think all of these `const`s should have explicit visibility.

includes/migrations/class-wc-stripe-pre-migration-snapshot.php
```diff
@@ -0,0 +1,314 @@
...
+			);
+
+			if ( ! is_array( $rows ) ) {
+				continue;
+			}
+
+			foreach ( $rows as $row ) {
+				$key = isset( $row['option_name'] ) ? (string) $row['option_name'] : '';
+
+				if ( '' === $key || isset( $captured[ $key ] ) ) {
```
daledupreez, line 245: Nit: this should check `self::KNOWN_OPTION_KEYS`, as we only set `$captured[ $key ]` when we got a value back.
```suggestion
				if ( '' === $key || isset( self::KNOWN_OPTION_KEYS[ $key ] ) ) {
```

includes/migrations/class-wc-stripe-pre-migration-snapshot.php
```diff
@@ -0,0 +1,314 @@
...
+ * never overwrites live state from the snapshot automatically.
+ *
+ * @package WooCommerce_Stripe/Migrations
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-wc-stripe-pp-version-detector.php';
```
daledupreez: I am very wary of storing multiple options in the options table. That's not a good thing for performance.

includes/migrations/class-wc-stripe-settings-migration-ledger.php
```diff
@@ -0,0 +1,429 @@
...
+ * @since 9.7.0
+ */
+class WC_Stripe_Settings_Migration_Ledger {
+
+	/**
+	 * WP options key for the ledger blob.
+	 *
+	 * @var string
+	 */
+	const OPTION_NAME = 'wc_stripe_settings_migration_ledger';
```
daledupreez: Using options for this level of logging and data is not a great idea.

includes/migrations/class-wc-stripe-settings-migration-ledger.php
```diff
@@ -0,0 +1,429 @@
...
+	 * @var array<int, string>
+	 */
+	const SECRET_SUFFIXES = [ '_secret', '_token' ];
+
+	/**
+	 * Sentinel string used in place of secret values.
+	 *
+	 * @var string
+	 */
+	const REDACTED = '__REDACTED__';
```
daledupreez: All of these `const`s should have explicit visibility.

daledupreez (commented):
I have some initial comments from inspection. I think we need to tweak some aspects of the approach, especially the heavy option usage.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-4x.php
```diff
@@ -0,0 +1,50 @@
...
+ *   3. Add tests in `tests/phpunit/admin/migrations/class-wc-stripe-pp-settings-map-4x-test.php`.
+ *
+ * @package WooCommerce_Stripe/Migrations
+ */
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map.php';
```
daledupreez: Also, we can remove all the `require_once` statements everywhere by adding the folders for the new classes to the classmap in `composer.json`.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map.php
pierorocca: Fixed — updated all five new classes from `@since 9.7.0` to `@since 10.8.0`.

includes/migrations/class-wc-stripe-pp-version-detector.php
pierorocca: Fixed — added `public` visibility to all constants in the version detector.

includes/migrations/class-wc-stripe-settings-migration-ledger.php
pierorocca: Fixed — added `public` visibility to all constants in the ledger class.

includes/migrations/class-wc-stripe-pp-version-detector.php
pierorocca: Fixed — `detect_major_version()` now delegates the three-layer walk to `detect_full_version()` and just extracts the major component from the result. Eliminates the duplicated fallback logic.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-4x.php
pierorocca: Fixed — added `includes/migrations/` and `includes/migrations/pp-settings-maps/` to the composer.json classmap. The `require_once` statements in the production files can be removed in a follow-up once the orchestrator PR is also updated to rely on autoload.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-3x.php
pierorocca: These are in DROP, not BUILD, intentionally. BUILD means "Woo Stripe has the destination feature but the migration can't populate it yet" (e.g., `button_radius` when Woo Stripe adds per-button radius). DROP means "there is no Woo Stripe equivalent or the setting is auto-handled." Country restriction settings (`allowed_countries`, `except_countries`, `specific_countries`) don't map to any Woo Stripe per-method setting — they're WordPress payment gateway base settings that WooCommerce core manages independently of our plugin. There's nothing to build toward; it's a genuine no-op.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-3x.php
pierorocca: The callable pattern is intentional here for two reasons: (1) each transformer has non-trivial logic (strip placeholders, truncate, invert boolean semantics) that deserves its own named function and isolated test — `transform_mode_to_testmode()` tested directly in `secret_key_provider` style data providers is cleaner than testing an inline closure; (2) the callable is stored as a PHP callable array `[ self::class, 'transform_...' ]` which is type-safe and statically analysable. A plain value map only works for simple 1:1 key substitutions; the moment a transformer needs to strip regex patterns or truncate strings it needs a function anyway. Happy to discuss further if you see a specific case where inline is clearly better.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map.php
pierorocca: The distinction is semantic and load-bearing. `skip` means the migration *tried* but couldn't proceed at runtime — destination already set, source missing, transformer threw. The same row might succeed on a different site. `drop` means the setting was *intentionally never migrated* as a design decision — it appears in the DROP/INVESTIGATE/BUILD categories of the settings map regardless of runtime state. Conflating the two would make it impossible to audit "which settings did we decide not to migrate" vs. "which settings failed to migrate on this site." The `OUTCOME_DROPPED` constant (added in a recent commit) makes this distinction explicit in the ledger schema.

includes/migrations/class-wc-stripe-pre-migration-snapshot.php
pierorocca: This PR adds exactly two new options: one for the snapshot blob and one for the ledger — both with `autoload=false`. They're one-time-per-site writes that are never read on any page load or REST request (only accessed via support/diagnostic tooling). The concern about options-table performance is real for autoloaded options or frequently-written options; neither applies here. The snapshot blob could grow to ~200KB on a heavily-configured site, which is why `autoload=false` is enforced at the write site and covered by a dedicated test.

includes/migrations/class-wc-stripe-settings-migration-ledger.php
pierorocca: Agreed that wp_options is not ideal for general-purpose logging, but for a one-time migration audit trail it's the right choice in this context: it requires no schema migration, no custom table, no Action Scheduler dependency, and survives plugin deactivation/reactivation. The ledger is not a log in the operational sense — it's a forensic record of a single irreversible event, sized at ~150 bytes/row × ~30 rows = ~5KB total. The human-readable log is mirrored to a WC log file (`woocommerce-gateway-stripe-pp-settings-migration-{date}.log`) for real-time tailing; the options-stored ledger is for structured querying after the fact.

includes/migrations/class-wc-stripe-pre-migration-snapshot.php
pierorocca: @daledupreez do you have an alternative you could recommend that suits the purpose and that's durable?

includes/migrations/class-wc-stripe-settings-migration-ledger.php
pierorocca: @daledupreez same comment as above do you have an alternative you could recommend that suits the purpose and that's durable?

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-3x.php
pierorocca: > A plain value map only works for simple 1:1 key substitutions

@daledupreez this may be the key comment to parse out. Do you ever see a scenario where there is more than the simple key substitution?

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-3x.php
pierorocca: @daledupreez could you elaborate?

includes/migrations/class-wc-stripe-pre-migration-snapshot.php
pierorocca: The current `isset($captured[$key])` is correct and handles more cases than the suggested `isset(self::KNOWN_OPTION_KEYS[$key])`. `$captured[$key]` is non-empty for any key already collected — either from Layer 1 (KNOWN_OPTION_KEYS) OR from earlier LIKE patterns that already matched it. The suggested check would only skip Layer-1 keys, potentially allowing a key matched by two different LIKE patterns to be captured twice. The current guard is the right dedup for the full two-layer capture.

On the width of the LIKE sweep more generally: the wide net is deliberate. The snapshot serves three purposes (see file header): rollback, forensic diagnostics, and migration runway for INVESTIGATE rows. Purpose 1 only needs the specific keys the migration writes — but purposes 2 and 3 benefit from capturing the full Stripe option namespace as it existed pre-migration. For a support case where a merchant reports "something changed after I connected to Woo Stripe," having the complete pre-migration state lets you diff against current state and immediately see what changed. For INVESTIGATE rows that become actionable in a future version, the snapshot already holds the merchant's value even if it wasn't in the map yet at migration time.

The snapshot is read-only infrastructure — nothing iterates over its `options` array to decide what to migrate. The migration's writes are entirely driven by the settings map rows. So a third-party plugin that happens to store options under a matching prefix only adds noise to the snapshot blob; it doesn't affect migration behavior.

includes/migrations/pp-settings-maps/class-wc-stripe-pp-settings-map-4x.php
pierorocca: Done — removed all `require_once` statements from the migration classes in both PRs. The only remaining `require_once` in the migrations directory is the conditional lazy-load of `wp-admin/includes/plugin.php` in the version detector, which loads a WordPress core file on demand and is unrelated to class autoloading.

includes/migrations/class-wc-stripe-settings-migration-ledger.php
pierorocca: Good call. The ledger no longer touches `wp_options` — it now writes to a dedicated `{$wpdb->prefix}wc_stripe_settings_migration_ledger` table created via `dbDelta()` in `install_table()` (gated on a `DB_VERSION` option). Each decision is a single `$wpdb->insert()` instead of the previous read-the-whole-blob-then-rewrite (which was O(n²) across a run), and `find_by_run_id()` / `find_by_source_key()` are now indexed lookups. The WC log-file mirror stays for human tailing, and the table is dropped under `WC_REMOVE_ALL_DATA` on uninstall. Done in 9fe490b74.

includes/migrations/class-wc-stripe-pre-migration-snapshot.php
pierorocca: Agreed. `capture()` no longer writes a new `wc_stripe_pre_migration_snapshot_archive_{timestamp}` option on each run (and `list_archives()` is gone) — it overwrites a single bounded `wc_stripe_pre_migration_snapshot` option in place. Since the migration is one-time, the most recent pre-migration state is all we retain. Done in 9fe490b74.

PR #5445, opened 2026-05-14
Add PP settings migration at OAuth completion

pierorocca:
## Summary

PR 2 of 2 — adds the orchestrator that ties together the foundation classes and wires it into the OAuth completion handler. The migration is functional end-to-end after this PR.

**Depends on [#5444](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5444) (foundation classes).** Targets the foundation branch so the diff shown here is only the new code on top.

## What's in this PR

| Change | Description |
| --- | --- |
| `WC_Stripe_PP_Settings_Migration` (orchestrator) | `maybe_run()` + AUTO/TRANSFORM application + per-method enable/order migrations + DROP/INVESTIGATE/BUILD recording. |
| `WC_Stripe_Connect::save_stripe_keys()` wiring | One-line addition between OAuth credential write (line 297) and PMC migration (line 333). |
| Orchestrator unit tests | 21 tests, 44 assertions covering idempotency, no-PP-data short-circuit, AUTO/TRANSFORM behavior, snapshot ordering, per-method enable UPM-gating, PP-only-as-BUILD, per-method order mapping, unknown-version fallback. |

## Order of operations inside `save_stripe_keys()`

The wiring is between two existing calls — ordering is load-bearing:

1. `update_main_stripe_settings($options)` — OAuth writes credentials + Woo Stripe defaults (line 297)
2. `clear_caches_after_key_update()` (line 304)
3. Connection-refresh scheduling
4. **`WC_Stripe_PP_Settings_Migration::maybe_run()`** ← NEW
5. `get_main_stripe_gateway()` (gateway instantiation)
6. `maybe_migrate_payment_methods_from_db_to_pmc(true)` — picks up our PP-derived enabled list (line ~333)
7. `configure_webhooks()`

Without ordering before step 6, merchants who switch from PP to Woo Stripe would land in PMC with Woo Stripe's empty default enabled list, losing the fidelity of their PP method selection.

## Design notes

- **Idempotent and self-detecting.** Gated by `woocommerce_stripe_pp_settings_migrated` option flag. Safe to call on every OAuth completion. If no PP data is present, marks itself complete with no writes.
- **Snapshot captured BEFORE any writes.** Load-bearing for rollback correctness. Verified by `test_captures_pre_migration_snapshot_before_writes`.
- **Per-method enable migration is architecture-gated.** When `woocommerce_stripe_upm_settings.enabled === 'yes'`, the merchant managed methods centrally via PMC, so per-gateway `enabled` flags in PP options are stale — we SKIP and record the decision.
- **Per-method order runs unconditionally.** If OCS is on, the stored order sits dormant (PMC handles ordering). If the merchant ever disables OCS, their historical PP order applies instead of Woo Stripe defaults.
- **PP-only methods recorded as BUILD ledger rows.** Billie, Konbini, MobilePay, PayNow, PromptPay, Revolut Pay, Swish, TWINT, Zip, Pay by Bank, FPX, GrabPay — none have Woo Stripe destinations. Audit trail captures the shopper-facing regression.
- **Don't write `stripe_legacy_method_order`.** The orchestrator only writes the modern `stripe_upe_payment_method_order` key. Writing the deprecated sibling would collide with Woo Stripe's internal pre-UPE → UPE bridge logic.

## Test plan

- [x] PHPStan level 8 clean — `npm run phpstan`
- [x] PHPCS clean — `npm run lint:php`
- [x] PHPUnit clean — `npm run test:php -- --filter='WC_Stripe_PP_Settings_Migration_Test'` → 21/21 tests, 44 assertions
- [x] Full migration suite + connect tests — 107/107 tests, 332 assertions
- [x] Pre-commit hook passed on every commit
- [ ] **Reviewers:** verify the `save_stripe_keys()` insertion point and ordering against `WC_Stripe_Payment_Method_Configurations::maybe_migrate_payment_methods_from_db_to_pmc()`
- [ ] **Reviewers:** verify PP gateway ID → UPE method ID mapping table against `WC_Stripe_Payment_Methods` constants
- [ ] **Manual QA (future):** install PP 3.X test site with seeded values, complete OAuth in Woo Stripe, verify pre-fill produces expected merchant-visible state in Settings → Payments → Stripe

### Coverage highlights

- Idempotency: completed flag short-circuits; running twice produces no additional ledger rows
- No-PP-data short-circuit: marks complete without snapshot or ledger writes
- Pre-fill behavior: applies when source present + dest empty; preserves merchant overrides; preserves OAuth-set values; records each decision with proper outcome
- Snapshot ordering: assertion proves snapshot reflects pre-call state, not post-call state
- Per-method enable: writes mapped UPE method IDs; skips when UPM enabled; records PP-only methods as BUILD
- Per-method order: maps PP gateway IDs to UPE methods, drops unmappable / non-Stripe entries, preserves existing Woo Stripe order
- Version fallback: unknown 5.X → 3.X best-effort with logged warning; 4.X placeholder → no writes but completed flag still set

## Follow-up tasks

- [ ] **Remove `require_once` statements from migration files** — `class-wc-stripe-pp-settings-migration.php` and its peers still have `require_once` guards that predate the composer classmap addition in #5444. Once this PR merges and the orchestrator relies on autoload, those guards can be deleted in a dedicated cleanup commit.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,6 +19,7 @@ xxxx-xx-xx - version 10.8.0
 * Dev - Add Claude Code skills and review rules under .claude/ to capture repo-specific contributor guidance
 * Dev - Move some independent classes into autoloader
 * Dev - Add foundation classes for the Payment Plugins for Stripe settings migration (version detector, pre-migration snapshot, structured ledger, and 3.X mapping table)
+* Add - Pre-fill Woo Stripe settings from Payment Plugins for Stripe at OAuth completion (idempotent, non-destructive, snapshot-backed for rollback)
 
 2026-05-12 - version 10.7.0
 * Fix - Hide the "move Stripe to the top" Optimized Checkout notice when all payment methods above Stripe are disabled
--- a/includes/connect/class-wc-stripe-connect.php
+++ b/includes/connect/class-wc-stripe-connect.php
@@ -324,6 +324,12 @@ if ( ! class_exists( 'WC_Stripe_Connect' ) ) {
 				$this->unschedule_connection_refresh();
 			}
 
+			// Pre-fill settings from Payment Plugins for Stripe when present. Idempotent and
+			// self-gated; safe to call on every OAuth completion. Runs BEFORE the PMC migration
+			// below so the existing PMC migration picks up our PP-derived enabled-methods list.
+			require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-wc-stripe-pp-settings-migration.php';
+			WC_Stripe_PP_Settings_Migration::maybe_run();
+
 			// For new installs the legacy gateway gets instantiated because there is no settings in the DB yet,
 			// so we need to instantiate the UPE gateway just for the PMC migration.
 			WC_Stripe::get_instance()->get_main_stripe_gateway();
--- a/includes/migrations/class-wc-stripe-pp-settings-migration.php
+++ b/includes/migrations/class-wc-stripe-pp-settings-migration.php
@@ -218,6 +218,9 @@ class WC_Stripe_PP_Settings_Migration {
 			);
 		}
 
+		// Persist all buffered ledger rows in a single DB write.
+		$ledger->flush();
+
 		update_option( self::COMPLETED_OPTION, 'yes' );
 
 		self::log_info(
@@ -283,6 +286,10 @@ class WC_Stripe_PP_Settings_Migration {
 				continue;
 			}
 
+			// Note: `! empty()` treats '0' as empty, so a future AUTO row targeting a numeric
+			// field where 0 is a meaningful merchant choice (e.g., button radius) would incorrectly
+			// allow an overwrite. Add a sentinel comment in the map row when such a case first
+			// appears so reviewers know to switch this guard to `null !== $dest_before`.
 			if ( ! empty( $dest_before ) ) {
 				$ledger->record_skip(
 					WC_Stripe_Settings_Migration_Ledger::CATEGORY_AUTO,
@@ -361,6 +368,7 @@ class WC_Stripe_PP_Settings_Migration {
 				continue;
 			}
 
+			// Note: see the same guard in apply_auto_rows() — same numeric-zero caveat applies here.
 			if ( ! empty( $dest_before ) ) {
 				$ledger->record_skip(
 					WC_Stripe_Settings_Migration_Ledger::CATEGORY_TRANSFORM,
@@ -460,9 +468,20 @@ class WC_Stripe_PP_Settings_Migration {
 		$dest_before  = $dest[ self::DEST_KEY_ENABLED_METHODS ] ?? [];
 		$enabled_list = is_array( $dest_before ) ? $dest_before : [];
 
+		// Pre-load all PP per-gateway settings in one pass to avoid 21+ individual get_option()
+		// calls inside the loop below.
+		$all_gateway_ids      = array_merge( array_keys( self::PP_GATEWAY_ID_TO_UPE_METHOD ), self::PP_ONLY_GATEWAY_IDS );
+		$gateway_settings_map = [];
+		foreach ( $all_gateway_ids as $gw_id ) {
+			$settings = get_option( 'woocommerce_' . $gw_id . '_settings', null );
+			if ( is_array( $settings ) ) {
+				$gateway_settings_map[ $gw_id ] = $settings;
+			}
+		}
+
 		foreach ( self::PP_GATEWAY_ID_TO_UPE_METHOD as $pp_gateway_id => $upe_method_id ) {
-			$gateway_settings = get_option( 'woocommerce_' . $pp_gateway_id . '_settings', null );
-			if ( ! is_array( $gateway_settings ) ) {
+			$gateway_settings = $gateway_settings_map[ $pp_gateway_id ] ?? null;
+			if ( null === $gateway_settings ) {
 				continue;
 			}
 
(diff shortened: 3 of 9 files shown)


pierorocca:
# Review Report — PR 5445 (Orchestrator + OAuth Wiring)

**Branch:** `agent/stripe-1146-orchestrator` → `agent/stripe-1146-foundation`
**Title:** Add PP settings migration at OAuth completion
**Date:** 2026-05-14
**Scope:** 1 new orchestrator class (`class-wc-stripe-pp-settings-migration.php`, 658 lines) + 6-line OAuth wiring in `class-wc-stripe-connect.php` + orchestrator tests + changelog.
**Blast radius:** Production code path — runs synchronously at every OAuth completion. Bounded by an idempotency flag (runs at most once per site) but worth careful review of the wiring.

## Overall Verdict

**APPROVE** (with one performance trade-off documented and three non-blocking optimization suggestions)

All four dimensions pass. The orchestrator is non-destructive (PP options never modified, Woo Stripe writes guarded by `! empty($dest[$key])`), idempotent (gated by `woocommerce_stripe_pp_settings_migrated`), and correctly ordered relative to the existing PMC migration. The +300-450ms OAuth-callback latency is documented and within the acceptable budget for a one-time-per-site operation.

## Architecture Compliance — PASS

- **Insertion point is correct and load-bearing.** `WC_Stripe_PP_Settings_Migration::maybe_run()` is called in `class-wc-stripe-connect.php:331` — after `WC_Stripe_Helper::update_main_stripe_settings($options)` at line 297 (OAuth credentials written) and OAuth stat updates at lines 300-302, and BEFORE `WC_Stripe_Payment_Method_Configurations::maybe_migrate_payment_methods_from_db_to_pmc(true)` at line 339. This ordering ensures the existing PMC migration picks up the PP-derived enabled-methods list our orchestrator writes.
- **Inline call (no `do_action` hook).** Consistent with how the existing PMC migration call is structured. One-time post-OAuth work with deterministic ordering.
- **Idempotency contract is consistent.** Orchestrator's `COMPLETED_OPTION = 'woocommerce_stripe_pp_settings_migrated'` matches Pre_Migration_Snapshot's `COMPLETED_OPTION_NAME` (PR 5444). Dependency direction is correct: orchestrator references the snapshot's constant, not vice versa. Both are excluded from the snapshot's catch-all sweep (recursion prevention).
- **Per-method enable migration is architecture-gated** on `woocommerce_stripe_upm_settings.enabled` — UPM merchants manage methods centrally via PMC, so per-gateway `enabled` flags would be unreliable. Correct skip with audit-trail row in the ledger.

## Security — PASS

- **Authorization inherited from OAuth handler.** The orchestrator is called from inside `save_stripe_keys()` which is private and only reachable through `connect_oauth()` → `maybe_handle_redirect()`. `maybe_handle_redirect()` gates on `current_user_can('manage_woocommerce')` plus a nonce check.
- **No new public-facing endpoints.** Static method `maybe_run()` is only invoked from already-protected OAuth flow.
- **No CSRF surface added.** Orchestrator accepts no user input; reads/writes `wp_options` only.
- **`! empty($dest[$key])` guard correctly preserves merchant values.** Edge cases evaluated: `''` → allows write (correct), `'no'` → preserves (correct, merchant explicitly disabled something), `'0'` → allows write (acceptable; no current rows use numeric-zero-meaningful semantics).

## Performance — PASS (with documented trade-off)

- **Estimated added OAuth-callback latency: 300-450ms worst-case** (typical: 200-300ms):
  - Snapshot capture: 4 LIKE queries + deserialization → ~50ms
  - AUTO/TRANSFORM applies: ~10 writes → ~100-150ms
  - Per-method enable loop: 21 `get_option()` calls + 1 `update_option()` → ~50-80ms
  - Per-method order migration: ~10ms
  - Ledger appends: ~30 appends with read+write each → ~100ms
- **Context:** OAuth callback baseline is already 2-3 seconds. Adding 15-20% to a one-time-per-site operation is acceptable.
- **No N+1 patterns, no autoload pressure** (snapshot and ledger both `autoload=false`).
- **Memory:** Snapshot held in PHP memory during capture; ~200KB worst case. Acceptable.

## Test Coverage — PASS

- **21 orchestrator tests, 44 assertions, all passing.**
- Covers: idempotency (completed flag short-circuit), unknown-version fallback, AUTO/TRANSFORM application with destination-set preservation, per-method enable migration including UPM-gating, per-method order migration with non-mappable gateway handling, snapshot-before-writes ordering (critical for rollback correctness), and PP 4.X placeholder handling.
- Uses `@dataProvider` extensively for parameterized cases.
- Tear-down properly resets `woocommerce_stripe_settings`, all PP option keys, snapshot, and ledger between tests.

## Suggestions (Non-blocking optimizations)

1. **Ledger batching could harvest ~100ms.** Currently each `record_*()` call does a `get_option()` + `update_option()`. Buffering rows in memory and flushing once at the end of `maybe_run()` would reduce ledger writes from ~30 to 1. Tracked for a follow-up PR.

2. **Pre-load PP gateway settings before the per-method enable loop.** `class-wc-stripe-pp-settings-migration.php` lines ~463-490 issue 21 separate `get_option()` calls inside the loop. Loading once into an array before the loop saves ~50ms.

3. **Integration test gap.** Orchestrator is unit-tested in isolation. No end-to-end test verifies "OAuth completes via `Connect::save_stripe_keys()` → orchestrator runs → settings appear." Acceptable for v1 (unit coverage is strong), but worth a follow-up.

4. **Future: numeric-zero-meaningful row sentinel.** The `! empty($dest[$key])` guard would incorrectly overwrite if a future row migrates a numeric field where `'0'` is a meaningful merchant choice (e.g., button radius of 0). Add a comment when such a row appears in the map.

## Required Changes

None.

PR #5447, opened 2026-05-15
[RSM - Diagnostics Mode] [DO NOT MERGE] Rate-limit the diagnostics /events endpoint

wjrosa:
Fixes RSM-1638

Adds a per-IP rate-limit gate to the unauthenticated `POST /wc/v3/wc_stripe/diagnostics/events` endpoint so a flooder can't fill the FIFO trace store, pad existing traces to their 200-event cap, or burn disk/CPU while diagnostics mode is on. The endpoint must stay unauthenticated (guest shoppers post traces), so a rate limit was the natural mitigation.

**Server**

- New `is_rate_limited()` / `arm_rate_limit()` helpers on `WC_REST_Stripe_Diagnostics_Controller` consult `WC_Rate_Limiter` (already used by `WC_Stripe_Intent_Controller`) keyed on `wp_hash` of the client IP. The rate-limit table never holds a raw IP.
- `permissions_check()` returns `WP_Error 429 wc_stripe_diagnostics_rate_limited` when the window is active, emits a `wcstripe_diagnostics_rate_limited` Tracks event (parallels the existing `wcstripe_diagnostics_auto_off`), and bumps a 24-hour transient counter.
- `ingest_events()` arms the next-request window up-front so malformed-event flooders are also bounded.
- Window is filterable via `wc_stripe_diagnostics_events_rate_limit` (default 2 seconds). Set to 0 to disable.
- `/diagnostics/summary` now returns `rate_limited_count` (rolling 24h).

**Client**

- Frontend loader passes the same window value (in ms) into `wcStripeDiag.rateLimitMs` so the recorder mirrors the server-side gate.
- `Recorder.flush()` coalesces flushes inside the window: a burst of `record()` calls schedules a single deferred flush instead of firing N `sendBeacon` calls the server will 429. `sendBeacon` is fire-and-forget so we can't observe 429s post-hoc; preempting them is the only honest path. Pagehide flushes force past the throttle.
- `DiagnosticsMode` renders an inline notice with the 24h 429 count and the filter name when the count is non-zero — operator-facing signal that the gate is biting on this store.

**Design decisions**

| Question | Decision | Rationale |
| --- | --- | --- |
| Per-IP vs per-session | Per-IP only | Attacker controls `diag_session_id` and will rotate it, so a per-session gate buys nothing against the actual abuse case (slot eviction). The 200-event-per-trace cap already bounds within-trace cost. |
| Default window | 2s | Loose enough that wallet bursts (Apple Pay / Google Pay) and shoppers behind a shared NAT egress don't 429 each other on a legitimate double-tap. Filter still scales per-store. |
| Storage | `WC_Rate_Limiter` (`wc_rate_limits` table) | Consistent with neighbour code. Bounded by the diagnostics-on window. |
| Observability | Tracks event + admin counter | Fleet visibility + per-store signal so operators can act without leaving the page. |
| IP resolver | `WC_Geolocation::get_ip_address()` | Same primitive WC core uses elsewhere. Spoofable behind a misconfigured proxy; trade-off documented in the docblock. |

### How can this code break?

- A merchant on a high-NAT egress (corporate proxy, mobile carrier) could see legitimate shoppers 429 each other on the same window. Mitigations: the 2s default is loose, the filter lets operators widen it, and the admin notice tells them when this is happening.
- `sendBeacon` is fire-and-forget so we can't actually retry a 429 — affected events drop silently when multi-tab / shared-NAT clients trip the gate. The client-side throttle eliminates the single-tab case; multi-tab is acceptable loss given diagnostics is a short debug window.
- The Stripe API still serves traffic — only the diagnostics ingest path is gated.

- 39 PHPUnit + 10 Jest tests covering: rate-limit trip, per-IP partitioning, no-IP fallback, filter-to-zero disable, Tracks counter increment, client-side throttle coalescing, deferred-flush behaviour, pagehide force, destroy cleanup, and the admin notice render.
- `npm run phpstan` and `npm run lint:php` clean.

## Testing instructions

1. Turn diagnostics mode on (**WooCommerce → Settings → Payments → Stripe → Advanced** → Checkout diagnostics).
2. From a single tab, open the cart/checkout and interact with the Stripe Elements normally — payments should still work end-to-end, with no client-visible delay.
3. From two browser windows on the same machine (same IP, different sessions), drive checkout in both within a one-second span. The second window's diagnostics requests should `429`; trace capture for that window may be incomplete.
4. Reload the **Stripe → Advanced** page. The diagnostics section should now show a notice like _"N diagnostics event requests were rate-limited in the last 24 hours…"_ referencing the `wc_stripe_diagnostics_events_rate_limit` filter.
5. Apply `add_filter( 'wc_stripe_diagnostics_events_rate_limit', '__return_zero' );` in a quick must-use plugin and confirm the gate stops firing for new requests. Remove the filter.
6. Confirm the existing `wcstripe_diagnostics_auto_off` flow is unaffected — turn diagnostics on, drive `capture_limit` traces to completion, and the toggle should still auto-off as before.

---

-   [x] Covered with tests

### Changelog entry

`Dev - Rate-limit the diagnostics /events endpoint per-IP (2s default, filterable) and emit a Tracks event when the limit trips` — added to `changelog.txt` and `readme.txt` under the 10.8.0 section.

**Post merge**

cc @malithsen

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@ xxxx-xx-xx - version 10.8.0
 * Dev - Exclude AGENTS.md and CLAUDE.md contributor-instruction files from the built plugin zip
 * Add - Append a "what's new" changelog link to the Updated! message after manually updating the plugin from the Plugins page
 * Add - Implement diagnostic mode to help debug checkout issues
+* Dev - Rate-limit the diagnostics /events endpoint per-IP (2s default, filterable) and emit a Tracks event when the limit trips
 * Fix - Add guards against invalid values for webhook state timestamps
 
 2026-05-12 - version 10.7.0
--- a/client/diagnostics/__tests__/recorder.test.js
+++ b/client/diagnostics/__tests__/recorder.test.js
@@ -286,6 +286,205 @@ describe( 'Recorder', () => {
 			);
 		} );
 
+		describe( 'client-side rate-limit throttle', () => {
+			// Wire a manual scheduler so the throttle's deferred flush is
+			// observable without a real timer: capture the (fn, ms) pair on
+			// setTimer and fire it manually after asserting on it.
+			function makeManualScheduler() {
+				const scheduled = [];
+				return {
+					setTimer: ( fn, ms ) => {
+						scheduled.push( { fn, ms } );
+						return scheduled.length;
+					},
+					clearTimer: () => {},
+					scheduled,
+				};
+			}
+
+			it( 'flushes immediately on the first record() when rateLimitMs is set', () => {
+				window.wcStripeDiag.rateLimitMs = 2000;
+				const mockTime = 0;
+
+				const recorder = makeRecorder( { now: () => mockTime } );
+				recorder.boot();
+				recorder.record( 'element.ready', { element_type: 'card' } );
+				recorder.flush();
+
+				expect( sendBeaconSpy ).toHaveBeenCalledTimes( 1 );
+			} );
+
+			it( 'defers a second flush within the rate window to a deferred-flush timer instead of firing sendBeacon', () => {
+				window.wcStripeDiag.rateLimitMs = 2000;
+				let mockTime = 0;
+				const scheduler = makeManualScheduler();
+
+				const recorder = makeRecorder( {
+					now: () => mockTime,
+					setTimer: scheduler.setTimer,
+					clearTimer: scheduler.clearTimer,
+				} );
+				recorder.boot();
+
+				// First flush goes through.
+				recorder.record( 'element.ready', { element_type: 'card' } );
+				recorder.flush();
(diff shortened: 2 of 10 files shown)


malithsen (commented):
I haven't tested this, but I wonder if we should table rate-limits until we ship/verify the current implementation.

Endpoint's risk surface is greatly reduced now that we auto turn off diagnostic mode after a limited number of attempts (and there's no way to keep it turned on indefinitely)
So the chance of someone abusing this endpoint is very limited. Diagnostic mode will be automatically disabled before rate limits kick in.

wjrosa:
You're right, Malith! I will convert this to a draft, and we can revisit this later if it makes sense

wjrosa:
Closing this for now. We can revisit this later if it makes sense

PR #5453, opened 2026-05-19
[Exploration] Require class const visibility

daledupreez:

Over the last week or two, I've made a lot of nit-picks requesting that class `const`s have explicit visibility, largely so we can make explicit decisions about our API surface area, especially because it's really hard to refactor class constants.

This PR explores the following changes:
 * Use PHPCS to require visibility of class `const`s via the `PSR12.Properties.ConstantVisibility.NotFound` sniff
 * Update all existing class `const`s that were present in the 10.7.0 release to have `public` constants
 * For new class `const`s added since 10.7.0, shift them to `protected`
    - I may have missed some
 * Adds a rule to our `AGENTS.md` that notes the new requirements. There is one noteworthy exception: for classes defined in `includes/constants/` we _intend_ for constant re-use, so that is one area where public constants make sense.

## Testing instructions

* Verify that the bulk of the changes involve setting the visibility for class `const`s to `public`
* For the handful of cases where the visibility was _not_ `public`, verify that these constants have been added after 10.7.0 was released.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa (commented):
Nice cleanup — the `public const` pass plus the PSR12 sniff is a clean way to lock this in. A few notes inline.

One more miss for the "shift new consts to protected" pass: `WC_Stripe_Agentic_Commerce_Feed_Validator::MAX_COLLECTED_ERRORS` (`includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-validator.php:38`) was added post-10.7.0 in #5319 as `public const`. Its only external consumer is the test file, so it would happily downshift to `protected` (test can use the same reflection helper pattern).

includes/constants/class-wc-stripe-payment-methods.php
```diff
@@ -9,54 +9,54 @@
...
-	const BACS_DEBIT_LABEL      = 'Bacs Direct Debit';
-	const GOOGLE_PAY_LABEL      = 'Google Pay';
-	const APPLE_PAY_LABEL       = 'Apple Pay';
-	const AMAZON_PAY_LABEL      = 'Amazon Pay';
-	const LINK_LABEL            = 'Link';
-	const PAYMENT_REQUEST_LABEL = 'Payment Request';
+	public const BACS_DEBIT_LABEL      = 'Bacs Direct Debit';
+	public const GOOGLE_PAY_LABEL      = 'Google Pay';
+	public const APPLE_PAY_LABEL       = 'Apple Pay';
+	public const AMAZON_PAY_LABEL      = 'Amazon Pay';
```
wjrosa, line 50: Per the PR rule, this one should be `protected` — `AMAZON_PAY_LABEL` was added in #5284 (May 18, post-10.7.0). It does have an external consumer (`class-wc-stripe-express-checkout-element.php:619`), so either route a getter or call out explicitly that we're keeping it `public` to match its sibling labels.

includes/AGENTS.md
```diff
@@ -58,6 +58,7 @@ Agentic Commerce feed integration:
 - Use data providers for parameterized PHPUnit coverage.
 - Define failure contracts for token/intent creation paths and handle failure explicitly at call sites.
 - For amount conversion paths, round before integer casting and test edge precision cases.
+- When adding `const` declarations to classes, they **MUST** have explicit visibility, and the visibility should default to `protected` or `private` with their values exposed via getter methods.
```
wjrosa: The "expose via getter methods" guidance won't fit every case — class consts are used in places that require constant expressions (default parameter values, attribute args, keys/values inside other `const` arrays like `EXPRESS_PAYMENT_METHODS` here). Worth softening to something like "where practical, expose via getters; otherwise keep `protected` for subclass-only use."

tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
```diff
@@ -1229,4 +1237,14 @@ public function test_hide_notices_dismisses_ece_location_notice() {
...
 		unset( $_GET['wc-stripe-hide-notice'], $_GET['_wc_stripe_notice_nonce'] );
 	}
+
+	/**
+	 * Get the meta key for the detached notice dismissed meta.
+	 *
+	 * @return string
+	 */
+	private function get_detached_notice_dismissed_meta_key(): string {
+		$constant = new ReflectionClassConstant( WC_Stripe_Admin_Notices::class, 'DETACHED_NOTICE_DISMISSED_META' );
```
wjrosa: Using `ReflectionClassConstant` to read a `protected` const works but defeats the encapsulation we just added — anyone outside the class can still grab the value. If we expect this pattern to recur (and it will, every time we move an existing const to `protected`), a small `WC_Stripe_Test_Helpers::get_protected_const()` would at least centralize the escape hatch. Or just expose a `protected static function get_*` and call it via reflection-of-method, which reads a bit more like intent.

includes/constants/class-wc-stripe-payment-methods.php
daledupreez: In general, I think that constants defined on classes in `includes/constants/` should be public by default, as the point of having them as constants is for re-use across the code base. So I don't think this one makes sense, but it also means that we need to be more specific in the rules. I've updated `includes/AGENTS.md` to incorporate your suggestion and to note that class constants in `includes/constants/` are generally expected to be public. That's in 459b6ae3d816757708b47d0f8bfe7541dd7b2ec8.

includes/AGENTS.md
daledupreez: I've relaxed this in 459b6ae3d816757708b47d0f8bfe7541dd7b2ec8, as well as incorporating that `includes/constants/` should get special treatment, as flagged in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5453#discussion_r3302505799).

tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
daledupreez: I've centralised this logic in a new `WC_Stripe_Test_Helper::get_protected_const_value()` helper, and updated `WC_Stripe_Admin_Notices_Test` to call the new helper in cd9bb35f5dd35b6e8eb26e2fda8d7ead7a90fe77. I also made `WC_Stripe_Agentic_Commerce_Feed_Validator::MAX_COLLECTED_ERRORS` protected in 7dfc6e084aedd7d4c9e3647414b3efd49ce8abe8, and then refactored to use the new test helper in 9c7f2d53bbe2d441491c7d071eaf0e67d755c8ef.

Thanks for flagging the consolidation and the missed constant!

daledupreez: Also, as a side note, reflection exists whether or not we encapsulate code. But it should be very clear to developers that using reflection to reach into our code is unlikely to be a stable approach! Our actual API surface becomes clearer if the constants are not directly accessible.

wjrosa (approved):
Looks good to me! 👍

malithsen (approved):
Looks good to me

PR #5454, opened 2026-05-19
Migrate webhook refund handler call sites to WC_Stripe_Refund

wjrosa:

Follow-up to #5298. Wires the new `WC_Stripe_Refund` domain wrapper into the four hand-rolled refund-handling sites in `WC_Stripe_Webhook_Handler` so the wrapper actually carries the weight it was written for.

- **`process_webhook_refund()`** — wraps the refund object once, replaces direct property reads with the wrapper's typed getters (`get_id()`, `get_balance_transaction_id()`), and uses `get_amount_decimal( $order->get_currency() )` so the decimal-currency branch lives in one place.
- **`process_webhook_refund_updated()`** — same pattern, plus `is_failed_or_canceled()` / `is_failed()` in place of inlined `in_array( $status, ['failed', 'canceled'], true )` predicates. The legacy `WC_Stripe_Helper::get_refund_reason_description()` call is retained at this site (rather than using the wrapper's `get_friendly_failure_reason()`) because the wrapper returns null when `failure_reason` is absent, while the helper falls through to its default branch — preserving the "Unknown reason" fallback that the existing order note relies on.
- **`get_refund_amount()`** — collapses the hand-rolled decimal-currency check to a single `get_amount_decimal( $notification->data->object->currency )` call.
- **Partial-capture path in `charge.succeeded`** — `$refund_object->balance_transaction` now goes through `get_balance_transaction_id()` with a null guard.

**Drive-by fix:** `WC_Stripe_Refund::get_amount_decimal()` was calling `WC_Stripe_Helper::no_decimal_currencies()`, which `develop` deprecated in 10.7.0 in favour of `WC_Stripe_Currency_Code::NO_DECIMAL_CURRENCY_CODES`. #5298 has no production call sites yet, so the deprecation never fired — this PR adds the call sites, which would surface the deprecation warning under `WP_DEBUG`. Switched the comparison to the non-deprecated constant (uppercase) and uppercased the currency before checking. The wrapper's unit tests still pass without changes because they pass lowercase currencies and the constructor's `strtolower()` / `strtoupper()` normalisation handles both shapes.

**PHPStan baseline.** Three suppressions drop because the property reads are now mediated by the typed wrapper:

- `object::$amount` (count 3 → 0)
- `object::$balance_transaction` (count 1 → 0)
- `object::$id` (count 8 → 2; remaining 2 are non-refund sites)

## Testing instructions

This is a behaviour-preserving refactor. Reviewer verification is the test suite:

- `npm run test:php -- --filter=WC_Stripe_Webhook_Handler_Test::test_process_webhook_refund_updated` — the existing data-provider parameterised case (invalid_status / failed / canceled) continues to exercise the migrated function end-to-end.
- `npm run test:php -- --filter=WC_Stripe_Refund_Test` — the wrapper's 33 PHPUnit cases still pass.
- `npm run phpstan` — no new errors at level 8; baseline shrinks.
- `npm run lint:php` on the modified files — clean.

End-to-end refund regression (Stripe Dashboard refund → `charge.refunded` webhook → order note; `refund.updated` with failed status → email + status revert) was not re-run locally — the test matrix is doing the lift.

---

### Changelog entry

<details>

#### Comment

Internal refactor of the webhook refund paths with no user-visible behaviour change. Lands in the same unreleased 10.8.0 cycle as its parent #5298, which already covers the new class.

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-refund.php
+++ b/includes/class-wc-stripe-refund.php
@@ -27,17 +27,17 @@ class WC_Stripe_Refund {
 	/**
 	 * The underlying Refund payload.
 	 *
-	 * @var stdClass
+	 * @var object
 	 */
-	private stdClass $refund;
+	private object $refund;
 
 	/**
 	 * Constructor.
 	 *
 	 * @since 10.8.0
-	 * @param stdClass $refund The Stripe refund payload (as returned by `json_decode`).
+	 * @param object $refund The Stripe refund payload (typically stdClass produced by `json_decode`).
 	 */
-	public function __construct( stdClass $refund ) {
+	public function __construct( object $refund ) {
 		$this->refund = $refund;
 	}
 
@@ -48,7 +48,7 @@ class WC_Stripe_Refund {
 	 *
 	 * @since 10.8.0
 	 */
-	public function raw(): stdClass {
+	public function raw(): object {
 		return $this->refund;
 	}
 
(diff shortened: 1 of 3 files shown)


wjrosa:
Postponing this work for now

PR #5455, opened 2026-05-19
Add tools to support autoload rollout

daledupreez:
Towards STRIPE-504

This PR adds two scripts to make it easier to roll out autoload support across the plugin:
 * `bin/build-class-dependency-map.php` - scan the files in the `includes/` directory to identify how each of the classes interrelate, what types of dependencies exist, and whether each class is already part of the autoloader
    - Usage: `php bin/build-class-dependency-map.php [--path=<dir>] [--format=json|mermaid] [--output=<file>] [--exclude-autoloaded] [--no-tier-dedupe]`
       * `--path=<directory>` - directory to scan; defaults to `includes`
       * `--format=(json|mermaid)` - output format; defaults to `json` -- note that no MermaidJS renderer is included
       * `--output=<file>` - file to write to; defaults to stdout rather than a file
       * `--exclude-autoloaded` - skip classes that are already being autoloaded via the classmap in `composer.json`; this also ignores them as dependencies
       * `--no-tier-dedupe` - skip the deduplication that removes lower tier dependencies when a higher tier dependency exists; the hierarchy is compile-time > runtime > PHPdoc
 * `bin/find-autoload-candidates.php` - call the tool above to build a dependency map and identify the next set of classes that make sense for autoloading.
   - Usage: `php bin/find-autoload-candidates.php [--path=<dir>] [--max-group-size=<n>] [--deep] [--types=<csv>]`
     * `--path=<directory>` - Directory to scan; defaults to `includes`
     * `--max-group-size=<size>` - Largest permissible group size when trying to identify groups of interrelated classes that could be added as a group.
     * `--deep` - Always run the deep checks for groups of interrelated classes. Without this flag, the tool will prompt the user.
     * `[--types=<dependency_types>]` - The dependency types to check. Defaults to `all` or `compile-time,runtime,doc`, but can also be any comma-separated combination of `compile-time`, `runtime`, and `doc`.
        - Allows for control over what types of dependencies we want to use for our checks -- for our early waves of checks, we're going to want to use `compile-time,runtime`

The tools depend on existing dev dependencies, and were 100% put together by Claude. The base-level scans seem to be working, but we'll likely need to get the initial files autoloaded before we can get a better sense for how well the tool works for identifying groups of interrelated classes.

## Testing instructions

* Run `composer install`
* Run both tools, especially `find-autoloader-candidates` in multiple ways and spot check the output.

---

### Changelog entry

<details>

#### Comment 
This is an internal-only tooling change that will have no merchant-facing impact. It doesn't warrant a changelog entry.
</details>

**Post merge**

### DIFF
--- a/bin/build-class-dependency-map.php
+++ b/bin/build-class-dependency-map.php
@@ -1,0 +1,636 @@
+<?php // phpcs:ignore WordPress.Files.FileName.InvalidClassFileName
+
+/**
+ * AST-based class dependency map for includes/.
+ *
+ * Walks every PHP file under --path (default: includes/) and emits a map of
+ * each internal class/interface/trait's dependencies on other internal
+ * classes. Internal-only by design — external symbols (WP/WC/SPL/
+ * Automattic\WooCommerce\*) are filtered out.
+ *
+ * Two edge buckets:
+ *   - strict: explicit declared relationships (extends, implements, use trait,
+ *     and native param/return/property type declarations).
+ *   - loose: named runtime references (new X, X::y, instanceof X, catch X,
+ *     and docblock types in @var/@param/@return/@throws/@property).
+ */
+
+declare( strict_types=1 );
+
+require __DIR__ . '/../vendor/autoload.php';
+
+use PhpParser\Node;
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\NameResolver;
+use PhpParser\NodeVisitorAbstract;
+use PhpParser\ParserFactory;
+
+// ---------------------------------------------------------------------------
+// CLI parsing
+// ---------------------------------------------------------------------------
+
+$cli_options = getopt( '', [ 'path::', 'format::', 'output::', 'include-external', 'exclude-autoloaded', 'help' ] );
+
+if ( isset( $cli_options['help'] ) ) {
+	// phpcs:ignore WordPress.WP.AlternativeFunctions.file_system_operations_fwrite
+	fwrite(
+		STDERR,
+		"Usage: php bin/build-class-dependency-map.php [--path=<dir>] [--format=json|mermaid] [--output=<file>] [--exclude-autoloaded]\n"
+		. "  --path=<dir>           Directory to scan (default: includes)\n"
+		. "  --format=json|mermaid  Output format (default: json)\n"
+		. "  --output=<file>        Write to file instead of stdout\n"
+		. "  --exclude-autoloaded   Drop classes covered by composer.json's autoload.classmap,\n"
+		. "                         and strip references to them from remaining classes' edges.\n"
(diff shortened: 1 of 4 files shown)


bin/build-class-dependency-map.php
```diff
@@ -0,0 +1,747 @@
...
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\NameResolver;
+use PhpParser\NodeVisitorAbstract;
+use PhpParser\ParserFactory;
+
+// ---------------------------------------------------------------------------
+// CLI parsing
+// ---------------------------------------------------------------------------
+
+$cli_options = getopt( '', [ 'path::', 'format::', 'output::', 'include-external', 'exclude-autoloaded', 'no-tier-dedupe', 'help' ] );
```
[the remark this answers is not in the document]
daledupreez: Oops. Removed in 0d38306de4112802b2f339202ed8353deee52c26.

malithsen (approved):
Nice tool. I spot checked a couple of subdirectories, and the output appears to be correct. Running without any path restrictions generates a massive mermaid diagram that my viewer isn't capable of rendering 😅 

Just one observation. ` --strict-only` flag, as mentioned in the PR description, isn't really implemented.

daledupreez:
@malithsen, thanks for the quick review. I made some changes earlier today that use a richer compile-time | runtime | docs classification for dependencies, but I forgot to update the PR description. Could you take another look?

daledupreez:
Also, if you want to visualise things, you can use the [Mermaid CLI](https://github.com/mermaid-js/mermaid-cli) to generate an image, but you need to increase the default character limit of 50,000 to get a render. (Though it's not super-useful given the size!)

bin/build-class-dependency-map.php
```diff
@@ -0,0 +1,747 @@
...
+
+declare( strict_types=1 );
+
+require __DIR__ . '/../vendor/autoload.php';
+
+use PhpParser\Node;
+use PhpParser\NodeTraverser;
+use PhpParser\NodeVisitor\NameResolver;
+use PhpParser\NodeVisitorAbstract;
+use PhpParser\ParserFactory;
```
malithsen, line 26: nit: Looks like php-parser comes transitively via phpunit/phpstan. Would it be worth including it as an explicit dev dependency in composer?

malithsen (approved):
Tested with the `--types` parameter, and still looks good to me.

bin/build-class-dependency-map.php
daledupreez: I think there are multiple indirect dependencies, but I've gone ahead and added a Composer dev dependency in 83de324eb8bb2db2c5f2507a644c1158e93211de.

PR #5457, opened 2026-05-20
Fix: tooltip button intercepts clicks from portal-rendered content

Snaacks:
Fixes STRIPE-1101

React bubbles synthetic events through the **component tree**, not the DOM tree. This means clicks on portal-rendered content (e.g. an `<ExternalLink>` inside a tooltip popup) bubble up to the `<button>` wrapper in `Tooltip`, where `handleMouseClick` was calling `event.preventDefault()` unconditionally blocking link navigation.

The tooltip popup is already rendered via `createPortal` outside the button's DOM subtree, so there is no HTML spec violation. The problem was the cross-portal event bubbling reaching the button handler.

**Fix:** added a DOM containment guard in `handleMouseClick`. If `event.target` is not within the button's own DOM subtree, the click originated from the portal and should be ignored.

```js
if ( ! event.currentTarget.contains( event.target ) ) {
    return;
}
```

This is a one-line guard that stops portal clicks from being treated as trigger activations without affecting any existing hover/click/touch behavior.

> This fix was developed with AI assistance (Claude Sonnet 4.6 / Claude Code).

## Testing instructions

1. Go to **WooCommerce → Settings → Payments → Stripe**
2. Find an account with payouts disabled the "Payout" row in the account details section shows an info icon with an **"Update now"** link inside the tooltip (see `client/settings/account-details/index.js`)
3. Click the info icon to open the tooltip
4. Click the **"Update now"** link it should navigate to the Stripe dashboard (previously: click was swallowed by `preventDefault`, link did nothing)

Unit test added: `does not close the tooltip when clicking interactive content inside it` verifies that a click on a link inside the tooltip popup does not toggle the `isClicked` state of the button wrapper.

---

### Changelog entry

<details>

Fix is scoped to internal tooltip interaction behavior; no user-visible change in the common case. A changelog entry can be added if the team prefers.
</details>


Snaacks:
**Conflict resolved: rebase against develop**

`develop` had a partial fix already in place:

```js
if ( event.target.closest( 'a' ) ) {
    return;
}
```

This guards against clicks on `<a>` elements specifically, but misses other interactive elements that could appear in tooltip content (buttons, inputs, etc.).

Our fix uses a DOM containment check instead:

```js
if ( ! event.currentTarget.contains( event.target ) ) {
    return;
}
```

This seems to be a more general fix; it doesn't care what *type* of element was clicked, only whether the click originated from inside the button's own DOM subtree. Any click from the portal (which is always outside the button in the DOM) is ignored. Both approaches solve the link case; ours covers the broader pattern.

> This fix was developed with AI assistance (Claude Sonnet 4.6 / Claude Code).

malithsen (commented):
> Click the "Update now" link it should navigate to the Stripe dashboard (previously: click was swallowed by preventDefault, link did nothing)

@Snaacks Thanks for the PR! However, I am having trouble replicating the issue on `develop` branch. It seems this was already fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5338

If this PR is an improvement over 5338, can you explain what it fixes?

daledupreez:
@Snaacks, just to follow up on this PR, can you confirm whether you still plan to work on the changes? Malith's [last comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5457#pullrequestreview-4430680637) flagged that it's not 100% clear how this is an improvement over #5338, so I want to follow up and check whether you plan to keep working on this. If you do plan to pick up work fairly soon, we should keep the PR open, but if you don't I think it makes sense to close the PR for the time being, and you can then start a new PR when you are in a position to tackle the issue.

daledupreez:
Noting that I am going to close this PR for the time being, as we need to make sure we're improving on the current markup.

PR #5458, opened 2026-05-20
Add wc_stripe_agentic_commerce_should_sync_product filter for catalog visibility adapters

wjrosa:
Towards STRIPE-1168

Introduces `wc_stripe_agentic_commerce_should_sync_product` — a per-product visibility gate that external plugins can hook to exclude products from every Agentic Commerce sync surface. Primary intended consumer is [`Automattic/woocommerce-ai`](https://github.com/Automattic/woocommerce-ai), which already owns merchant-facing product-selection UI (`all` / `by_taxonomy` / `selected` modes plus a per-product meta-box override). Today that UX only governs WAS's UCP surfaces; the Stripe Files API feed pushes every published product regardless, so a merchant who has explicitly hidden a product from AI agents still sees it appear in Stripe-routed agent conversations.

Rather than have the Stripe plugin learn about WAS's settings directly, this PR opens the seam so the adapter can live in WAS and the Stripe plugin stays ignorant of any specific consumer.

**Applied at three per-product entry points:**

- `WC_Stripe_Agentic_Commerce_Product_Mapper::map_product()` returns an empty row when the filter votes false. The existing validator's required-field check then rejects it, so the walker skips the entry. No walker changes.
- `WC_Stripe_Agentic_Commerce_Inventory_Tracker::track_stock_change()` early-returns so excluded products don't queue inventory deltas Stripe wouldn't be able to match to a SKU.
- `WC_Stripe_Agentic_Commerce_Inventory_Tracker::track_product_archive()` early-returns so deleted or trashed products that were never on Stripe don't push `out_of_stock` entries for SKUs Stripe never received.

**Companion query-level scoping.** The existing `wc_stripe_agentic_commerce_product_query_args` filter (added in 10.5.0) lets adapters scope the full-feed query — `post__in`, `post__not_in`, `tax_query`. Adapters should hook both: query args for the bulk feed path (cheap), `should_sync_product` for the per-product event surfaces (necessary). The docblock on the new method points adapters at the existing filter to avoid validator log spam from large exclusion sets.

**Default behaviour unchanged.** With no filter registered, every product flows through to the feed exactly as today.

## Testing instructions

PHPUnit only — no merchant-facing UX change without an adapter installed.

- `npm run test:php -- --filter='WC_Stripe_Agentic_Commerce_Product_Mapper_Test|WC_Stripe_Agentic_Commerce_Inventory_Tracker_Test'` — 77 tests pass, including the four new cases that guard the filter contract on both classes.
- `npm run phpstan` — clean.
- `npm run lint:php` — clean.

To exercise end-to-end with an adapter, add a temporary callback in `wp_loaded`:

```php
add_filter( 'wc_stripe_agentic_commerce_should_sync_product', function ( $sync, $product ) {
    return $product->get_id() !== 123; // Exclude product 123 from every sync surface.
}, 10, 2 );
```

Then trigger a feed sync (`wp wc stripe agentic-commerce sync`), change product 123's stock, and delete product 123 — none of the three actions should produce a Stripe-side entry for it.

## Related

- WAS adapter follow-up will be filed separately on `Automattic/woocommerce-ai`.

---

### Changelog entry

Changelog entry added to `changelog.txt` and `readme.txt`.

**Post merge**


includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
```diff
@@ -103,6 +103,10 @@ public function register_hooks(): void {
 	 * @return void
 	 */
 	public function track_stock_change( \WC_Product $product ): void {
+		if ( ! WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product( $product ) ) {
+			return;
+		}
```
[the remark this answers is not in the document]
wjrosa: Fixed in e3dac4c5a. Extracted a private `evict_pending_entries()` helper that clears both `PENDING_UPDATES_OPTION` and `PENDING_ARCHIVES_OPTION` for the product ID, called from both early-return paths. New test in each method pre-seeds both queues, flips the filter, asserts both options end empty.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
```diff
@@ -927,4 +927,101 @@ public function test_delete_field_always_returns_null_for_active_products() {
...
+		$callback = static fn() => false;
+		add_filter( 'wc_stripe_agentic_commerce_should_sync_product', $callback );
+
+		try {
+			$this->assertFalse( \WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product( $product ) );
+		} finally {
+			remove_filter( 'wc_stripe_agentic_commerce_should_sync_product', $callback );
+			$product->delete( true );
+		}
+	}
```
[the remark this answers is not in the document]
wjrosa: Fixed in e3dac4c5a. Added `provide_falsy_filter_values` data provider (`false`, `0`, `''`, `null`) and renamed the test to `test_should_sync_product_treats_any_falsy_filter_return_as_exclude` so the name matches the contract.

TateLyman:
Source-level read only; I did not run the plugin or touch any merchant/Stripe surface.

The new per-product visibility seam looks useful, especially with the stale pending-entry eviction from the follow-up commit. One lifecycle edge to confirm with the WAS adapter: what removes a product from Stripe if it was already synced while visible, then the merchant later hides it from agentic commerce?

Right now the excluded branch in `track_stock_change()` / `track_product_archive()` evicts local pending entries and returns early. That is correct for products that were never exported, but for a product that already exists in the Stripe Files/API feed, returning early on later stock/archive events means no tombstone/out-of-stock update is emitted from this plugin. If the adapter always triggers a full feed resync immediately when visibility changes, that may be enough. If not, the sequence can leave a stale product visible to Stripe-routed agent conversations until the next full sync:

1. product A is visible and full-feed sync exports it;
2. merchant hides product A in WAS / adapter starts returning false;
3. product A changes stock or is trashed before a full feed replacement runs;
4. tracker evicts pending entries and returns, so Stripe never receives an explicit removal for A.

Suggested guard: document that adapters must schedule a full feed replacement when the filter outcome changes, or expose a small helper/event that lets the adapter queue an archive/tombstone for products that were previously exported and are now excluded.

wjrosa:
Thanks for reviewing this, @TateLyman! Your concern is fair — there's no `out_of_stock` tombstone for the "previously exported, now excluded, then trashed" sequence until the next full sync. Going with the lighter mitigation you suggested: pushed a follow-up commit on this branch that adds a `wc_stripe_agentic_commerce_schedule_full_resync` action adapters fire when their filter outcome changes (idempotent — no-op while a sync is already pending), and the filter docblock now spells out the lifecycle contract.

TateLyman:
Thanks for turning that into an adapter-triggered full-resync hook. That is the right contract boundary.

One small follow-up to confirm before merge: the “resync now” queue should not be blocked by the recurring full-feed schedule using the same hook. If the recurring Action Scheduler occurrence is already pending, a broad `as_has_scheduled_action( self::SCHEDULED_ACTION )` guard can make the adapter-triggered action a no-op exactly when the merchant changes visibility.

The test I would want here is the collision case: recurring full-feed sync already scheduled, adapter fires `wc_stripe_agentic_commerce_schedule_full_resync`, and an immediate one-off resync is still queued once, without stacking duplicates. A separate Action Scheduler group for the one-off resync, or an explicit query for an identical one-off action, would keep that contract clear.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -169,13 +169,43 @@ public function get_id(): string {
...
+	public function schedule_full_resync_now(): void {
+		if ( ! function_exists( 'as_has_scheduled_action' ) || ! function_exists( 'as_enqueue_async_action' ) ) {
+			return;
+		}
+
+		if ( as_has_scheduled_action( self::SCHEDULED_ACTION ) ) {
+			return;
+		}
+
+		as_enqueue_async_action( self::SCHEDULED_ACTION, [], 'wc-stripe' );
```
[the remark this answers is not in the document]
wjrosa: Confirmed — `as_has_scheduled_action` with an empty $group drops the group filter entirely (`ActionScheduler_DBStore::query_actions` line 476: `if ( ! empty( $query['group'] ) )`), so the existing guard matched the recurring `'wc-stripe'` pending occurrence and made the adapter hook a permanent no-op after the first `activate()`.

Applied the suggestion: new `ASYNC_RESYNC_GROUP = 'wc-stripe-agentic-resync'` constant, both check and enqueue scoped to it, and `deactivate()` clears both groups. Added `test_schedule_full_resync_now_enqueues_when_recurring_is_pending` that pre-schedules a recurring action in `'wc-stripe'` before firing the adapter hook and asserts the one-off lands in the resync group and dedups on repeat.

wjrosa:
Confirmed and addressed — see [the inline thread](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5458#discussion_r3288831118) for the API detail (empty `$group` drops the group filter entirely, so the guard was matching the recurring occurrence). New `ASYNC_RESYNC_GROUP = 'wc-stripe-agentic-resync'` keeps the one-off check + enqueue independent of the recurring schedule, plus a collision test that pre-schedules the recurring action and asserts the adapter-fired one-off still lands without stacking.

TateLyman:
I re-read the current PR head from GitHub (`a09f924d022d8e559a1e37da6dd44b0eaa69edc3`) after the inline reply. I may be looking before a push landed, but the separate resync group does not appear to be in the branch yet.

Current raw file at head still shows:

```php
if ( as_has_scheduled_action( self::SCHEDULED_ACTION ) ) {
	return;
}

as_enqueue_async_action( self::SCHEDULED_ACTION, [], wc-stripe );\n```\n\nI also do not see `ASYNC_RESYNC_GROUP` / `wc-stripe-agentic-resync` in the current PR diff. So the collision case looks still open in the pushed code, even though the inline thread says it was addressed.\n\nThe intended fix still looks right: check/enqueue the adapter-fired one-off in its own Action Scheduler group, and leave the recurring `wc-stripe` schedule independent. Once that commit is actually on the branch, this concern should be closed.

**Comment by wjrosa** [comment]
Sorry — you were right, my earlier reply described work that never actually got committed. Pushed it now in 271df01a2:

- `ASYNC_RESYNC_GROUP = 'wc-stripe-agentic-resync'` constant; `schedule_full_resync_now()` checks and enqueues in that group, leaving the recurring `wc-stripe` schedule independent.
- `deactivate()` tears down both groups so plugin removal still leaves a clean queue.
- New `test_schedule_full_resync_now_still_enqueues_when_recurring_is_pending` covers the collision: pre-schedules the recurring occurrence in `wc-stripe`, fires the adapter hook, asserts the one-off lands in `wc-stripe-agentic-resync` without stacking on repeated fires.

PHPStan + lint clean, integration test class 56/56.

File: includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php [hunk 4]
```diff
@@ -179,6 +187,39 @@ public function maybe_cancel_pending_archive( int $post_id ): void {
...
+	 *
+	 * Called from the track_* methods when the visibility filter votes false so
+	 * a product that was queued before the filter flipped doesn't still flush
+	 * to Stripe on the next batch.
+	 *
+	 * @since 10.8.0
+	 * @param int $product_id Product ID to evict.
+	 * @return void
+	 */
+	private function evict_pending_entries( int $product_id ): void {
```
Mayisha, line 201: I thought of one scenario but I do not have a very good solution to it, except that we walk through all the products before sync. I do not recommend it at all, but sharing the scenario here anyway. 
The current `should_sync_product()` enforcement is event-driven: the `track_*` early-returns prevent enqueuing an already-excluded product, and `evict_pending_entries()` cleans up an entry whose filter outcome flipped after it was queued. That covers the common cases, but when no follow-up event arrives, we do not exclude the hidden product.
 1. Product A is visible → a stock change queues an inventory delta.
 2. Merchant hides A (filter flips to false).
 3. No further stock change or trash fires for A before the batch window elapses.
 4. sync_inventory() → generate_inventory_feed() reads the pending option and ships A to Stripe — the flush path never re-checks should_sync_product().

Mayisha (approved):
Changes look good to me and the tests are clean 👍

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
wjrosa: Good catch — fixed in baf708105. Added a flush-time `prune_stale_pending_entries()` that re-runs `should_sync_product()` for each pending row and persists the filtered list, so an entry whose visibility flipped during the batch window gets dropped before the upload. Runs after the `MAX_PENDING_UPDATES` threshold check so a runaway queue still defers cleanly without paying the per-row `wc_get_product` cost. For permanently-deleted products: inventory entries are dropped (the SKU no longer exists on Stripe's side), archive entries are kept (the row was mapped at archive time, Stripe still has it to mark out_of_stock). Four new tests cover both prune-on-filter-flip and the missing-product policies.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker-test.php
```diff
@@ -576,6 +628,102 @@ function () {
...
+			]
+		);
+
+		$this->sut->sync_inventory();
+
+		// The deleted product was pruned, leaving nothing to upload — the
+		// option should have been cleared by the prune helper.
+		$pending = get_option( WC_Stripe_Agentic_Commerce_Inventory_Tracker::PENDING_UPDATES_OPTION, [] );
+		$this->assertEmpty( $pending );
+	}
```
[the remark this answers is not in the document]
wjrosa: Fair point — fixed in ef2f01d4. The Files API pre-request filter gets the real CSV path, so the tests now read the file and assert which IDs reached the upload: kept SKU present, flipped SKU absent. The deleted-product inventory test asserts the upload never fires; the deleted-product archive test asserts the captured row IS shipped.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker-test.php
```diff
@@ -956,6 +1157,88 @@ function ( $pre, $args, $url ) {
...
+				'status' => 'pending',
+			]
+		);
+
+		$this->sut->sync_archives();
+
+		// Upload completed, pending list cleared on success.
+		$pending = get_option( WC_Stripe_Agentic_Commerce_Inventory_Tracker::PENDING_ARCHIVES_OPTION, [] );
+		$this->assertEmpty( $pending );
+	}
```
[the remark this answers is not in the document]
wjrosa: Same fix in ef2f01d4 — archive tests now parse the uploaded CSV via the pre-request filter and assert the kept id is present, the flipped id is absent, and (for the deleted-product case) the captured row is still shipped.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-integration-test.php
```diff
@@ -212,16 +212,140 @@ public function test_register_hooks() {
...
+		$this->assertSame(
+			count( $pending_before ),
+			count( $pending_after ),
+			'Repeated calls while the resync is pending must not stack queue entries.'
+		);
+
+		remove_action( 'wc_stripe_agentic_commerce_schedule_full_resync', [ $integration, 'schedule_full_resync_now' ] );
+		as_unschedule_all_actions( \WC_Stripe_Agentic_Commerce_Integration::SCHEDULED_ACTION, [], 'wc-stripe' );
+		as_unschedule_all_actions( \WC_Stripe_Agentic_Commerce_Integration::SCHEDULED_ACTION, [], 'wc-stripe-agentic-resync' );
+	}
```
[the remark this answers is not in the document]
wjrosa: Added in a44a9257 — `test_deactivate_clears_both_scheduled_action_groups` seeds a recurring action in `wc-stripe` and a one-off resync in `wc-stripe-agentic-resync`, calls `deactivate()`, and asserts both groups are empty.

wjrosa:
Thanks for the review, @Mayisha ! Following your comment [here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5458#discussion_r3298670901), I did some major changes to the PR. I'd appreciate another quick look

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
```diff
@@ -179,6 +187,92 @@ public function maybe_cancel_pending_archive( int $post_id ): void {
...
```
Mayisha, line 201: Is this `evict_pending_entries` calls needed anymore now that you are checking `prune_stale_pending_entries` in the `sync_*` methods?

Mayisha (approved):
The latest commits also look good to me 👍

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -965,4 +974,43 @@ private function get_cached_shipping_zones(): array {
...
+		 * products. Adapters that want immediate convergence when their filter
+		 * outcome changes (e.g. a merchant flips a visibility setting) MUST fire
+		 * `do_action( 'wc_stripe_agentic_commerce_schedule_full_resync' )` to
+		 * enqueue an immediate full-catalog sync.
+		 *
+		 * @since 10.8.0
+		 * @param bool        $should_sync Whether to include the product. Default true.
+		 * @param \WC_Product $product     Product being evaluated.
+		 */
+		return (bool) apply_filters( 'wc_stripe_agentic_commerce_should_sync_product', true, $product );
```
diegocurbelo: We could use `wp_validate_boolean()` to prevent `'false'` being parsed as `true`

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -44,6 +44,21 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
...
+	 * guard in {@see self::schedule_full_resync_now()} can ask "is another
+	 * adapter-fired resync already pending?" without matching the recurring
+	 * full-feed occurrence — `as_has_scheduled_action()` filters by group,
+	 * so a shared group would make the one-off a no-op whenever the cron
+	 * tick is queued.
+	 *
+	 * @var string
+	 * @since 10.8.0
+	 */
+	private const ASYNC_RESYNC_GROUP = 'wc-stripe-agentic-resync';
```
diegocurbelo, line 60: A manual sync from the UI doesn't clear this new group...  `trigger_sync()` in `class-wc-rest-stripe-agentic-commerce-controller.php` only unschedules the `wc-stripe group`

diegocurbelo (changes requested):
In general the code looks good, and work as described

However, excluded products pollute the logs and result in every sync being labeled as `Partial success`:

`map_product()` returns `[]` for excluded products, so `validate_required_fields()` emits one error per required field (~10 errors):

```php
add_filter( 'wc_stripe_agentic_commerce_should_sync_product', function ( $sync, $product ) {
    return $product->get_id() !== 30;
}, 10, 2 );
```

Image


Image

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
wjrosa: Mostly — but not for archives. `prune` can't re-check the filter for a permanently-deleted product (no `WC_Product` to load), so it keeps those rows. `evict` runs inside `track_product_archive()` with a live product, so it's the only place the filter is honored for a product about to be deleted. Without it, `included → archived → excluded → permanently deleted` re-ships `out_of_stock` via prune's keep-policy — the leak this filter exists to close.

Added `test_sync_archives_does_not_ship_excluded_product_after_permanent_delete` to lock that path; it fails if I drop the `evict` call. (On the inventory queue you're right that prune fully subsumes it.)

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Good call — fixed in 70bd51f44. Switched the wrapper to `wp_validate_boolean( apply_filters( … ) )` so an adapter returning the string `'false'` is read as false instead of as a truthy non-empty string. It still normalises `null` / `0` / `''` to false like the old cast.

Also extended `provide_falsy_filter_values` with `'false'` and `'False'` (mixed-case) — those two data sets fail under the old `(bool)` cast and pass now, so the contract is locked.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Right — the manual sync only reset the `wc-stripe` group, so a pending adapter-fired one-off in `wc-stripe-agentic-resync` would still fire right after, redoing the upload the manual sync just completed.

Fixed in fdf394df8: added `WC_Stripe_Agentic_Commerce_Integration::cancel_pending_full_resync()` (clears only the resync group) and call it from `trigger_sync()` after a successful sync. Scoped to that group so it leaves the recurring `wc-stripe` reschedule the manual sync just set untouched.

Covered at both levels: a unit test that the cancel clears the resync group but spares the recurring occurrence, and an end-to-end REST test that a manual `/sync` drops a queued one-off (both fail without the wiring).

TateLyman:
Re-read the current head after `4206e27`, `70bd51f`, `fdf394d`, and the develop merge. The two edge cases I was watching now look covered from the diff: adapter-returned string falsy values go through `wp_validate_boolean()`, and manual sync clears the pending adapter-fired resync group without touching the recurring `wc-stripe` schedule.

I do not see another contract issue from the public diff. Remaining blocker looks like the requested maintainer review / queued CI rather than the lifecycle points above.

wjrosa:
Thanks for reporting it, @diegocurbelo. I have fixed in 1e1e371b3.

The walker only has one skip signal (a non-empty `validate_entry()` return), so an excluded product — which arrives as an empty row from `map_product()` — was indistinguishable from a validation failure: ~10 missing-field errors per product, and the `iterated − synced` skip count flipped the sync to `succeeded_with_errors`.

The fix classifies the skip instead of conflating it:
- `validate_entry()` drops an excluded empty row without recording per-field errors and tallies it in a separate `excluded_count` (`get_excluded_count()`).
- `sync_feed()` subtracts that from the dropped total, so only genuine validation failures warn and drive the Partial success badge. Exclusions are logged as a plain `excluded_products` count.

So a hidden product no longer spams the log or downgrades the status, while a real schema failure still does both.

Verified with 3 tests (validator silent-skip, empty-row-without-exclusion-is-still-a-failure, and an end-to-end sync asserting `skipped_products = 0` / status not `succeeded_with_errors` for a pure exclusion). Reverting the classification fails the guard tests with `skipped_products = 1`, reproducing the reported symptom.

diegocurbelo (approved):
Thanks for the changes @wjrosa!

It looks good to me, the feed status is now `Success` when skipping products 👍🏼

PR #5459, opened 2026-05-20
Ensure that upgrade handlers are called explicitly

daledupreez:

This PR introduces a new `WC_Stripe_Update_Manager` class and refactors how we trigger logic that is intended to run when we detect that a new version of the plugin has been installed.

With the new approach, the code in `WC_Stripe->install()` is updated to call `WC_Stripe_Update_Manager::run_update_checks()` instead of calling `do_action( 'woocommerce_stripe_updated' )`. `WC_Stripe_Update_Manager::run_update_checks()` then calls any methods that are responsible for performing any upgrade-related checks or data migrations, before then calling `do_action( 'woocommerce_stripe_updated' )` to allow external hooks to run. The intent here is to make it explicit as to what code will run when handling version changes. In turn, the changes also reduce the amount of code that was _always_ running in `WC_Stripe->init()`, and they remove the need for the migration classes to hook into the `woocommerce_stripe_updated` action from their constructors. This should be a cleaner, simpler pattern for upgrade-time logic in future.

It's also worth noting that the `require_once`s are moved from `WC_Stripe->init()` to `WC_Stripe_Update_Manager::run_update_checks()`, which should also reduce the overhead slightly for the majority of incoming requests.

## Testing instructions

This is up for discussion, as the refactor is pretty simple and straightforward. Some of the individual migrations have unit tests, and the only change is in when the files are loaded, and how they are called.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -710,8 +709,20 @@ public function get_setting_link() {
 	 * @return void
 	 */
 	public function stripe_updated() {
-		$previous_version = get_option( 'wc_stripe_version' );
+		wc_deprecated_function( __METHOD__, '10.8.0', 'WC_Stripe_Admin_Notice::check_update_notices' );
```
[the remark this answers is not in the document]
daledupreez: Fixed in 85728b507d0e5bad800cd9175b0d26baaa246151.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -710,8 +709,20 @@ public function get_setting_link() {
...
+		self::check_update_notices( get_option( 'wc_stripe_version' ) );
```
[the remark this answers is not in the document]
daledupreez: Fixed in 85728b507d0e5bad800cd9175b0d26baaa246151.

includes/class-wc-stripe-update-manager.php
```diff
@@ -0,0 +1,92 @@
...
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-allowed-payment-request-button-types-update.php';
+		$functions[] = [ new Allowed_Payment_Request_Button_Types_Update(), 'maybe_migrate' ];
+
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-migrate-payment-request-data-to-express-checkout-data.php';
+		$functions[] = [ new Migrate_Payment_Request_Data_To_Express_Checkout_Data(), 'maybe_migrate' ];
+
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-sepa-tokens-for-other-methods-settings-update.php';
+		$functions[] = [ new Sepa_Tokens_For_Other_Methods_Settings_Update(), 'maybe_migrate' ];
+
+		return $functions;
```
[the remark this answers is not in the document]
daledupreez: Oops! Added back in f6ea486b89ccc6752adb855f8d6be81b8f6b6a9d.

includes/class-wc-stripe-update-manager.php
```diff
@@ -0,0 +1,44 @@
...
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-sepa-tokens-for-other-methods-settings-update.php';
+		( new Sepa_Tokens_For_Other_Methods_Settings_Update() )->maybe_migrate();
+
+		( new WC_Stripe_Express_Checkout_Add_Change_Payment_Method_Location_Update() )->maybe_migrate();
+
+		/**
+		 * Action triggered when the plugin is updated.
+		 */
+		do_action( 'woocommerce_stripe_updated' );
+	}
```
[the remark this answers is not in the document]
daledupreez: Basic tests added in afa74af6f9ac9823c38f650d132bb92be8fdd3f4; test for the action added in 21361c572bf76b876b6664880ea51083712d8078.

tests/phpunit/class-wc-stripe-upgrade-manager-test.php
```diff
@@ -0,0 +1,118 @@
+<?php
+
+/**
+ * These tests make assertions against the WC_Stripe_Checkout_Sessions_Ajax_Handler class.
```
diegocurbelo:
```suggestion
 * These tests make assertions against the WC_Stripe_Update_Manager class.
```

includes/class-wc-stripe-update-manager.php
```diff
@@ -0,0 +1,94 @@
...
+	/**
+	 * Run update checks when a new version of the plugin is being installed.
+	 * This can include updates, downgrades, and fresh installs.
+	 *
+	 * @param string $previous_version The previous version of the plugin.
+	 * @param string $current_version  The current version of the plugin.
+	 * @return void
+	 * @since 10.8.0
+	 */
+	public static function run_update_checks( $previous_version, $current_version ): void {
```
diegocurbelo: `$current_version` is not passed down to the migrations (not used in this method), and if a migration wants it it can use the WC_STRIPE_VERSION constant.

includes/class-wc-stripe-update-manager.php
```diff
@@ -0,0 +1,94 @@
...
+		do_action( 'woocommerce_stripe_updated' );
+	}
+
+	/**
+	 * Run the update functions.
+	 *
+	 * @param string $previous_version The previous version of the plugin.
+	 * @return void
+	 */
+	protected function run_upgrade_functions( $previous_version ): void {
```
diegocurbelo: This should be `run_update_functions()` to keep it consistent with the Manager class name

tests/phpunit/class-wc-stripe-upgrade-manager-test.php
```diff
@@ -0,0 +1,118 @@
...
+ */
+class WC_Stripe_Upgrade_Manager_Test extends WP_UnitTestCase {
```
diegocurbelo: The new Manager class is Update... we should rename this to `WC_Stripe_Update_Manager_Test` (and also the file)

diegocurbelo (commented):
The implementation looks solid to me.. and it tests well!

I just think we should address the naming inconsistency "Update" vs "Upgrade" before merging though

tests/phpunit/class-wc-stripe-upgrade-manager-test.php
daledupreez: 🤦 Fixed in 2f820b707a4d00436db84bb61379644a2c250f35. Thanks for catching this!

tests/phpunit/class-wc-stripe-upgrade-manager-test.php
daledupreez: Good catch. Fixed in 2f820b707a4d00436db84bb61379644a2c250f35.

includes/class-wc-stripe-update-manager.php
daledupreez: I've removed the parameter in 66f79f9f763596502b2eb8ee42e50c6beced3737.

includes/class-wc-stripe-update-manager.php
daledupreez: Renamed in 326f382df7c758c37126d748c664fd3aeb11ee31.

diegocurbelo (approved):
Thanks for the changes @daledupreez!

LGTM 👍🏼

PR #5460, opened 2026-05-20
Fix STRIPE-1164 Amazon Pay billing last name fallback

ryanr14:
## Summary
- Adds a placeholder billing last name when an express checkout event only provides a single billing name.
- Uses only the provider-supplied `billingDetails.name` for billing first/last name normalization, avoiding customer-store last-name reuse.
- Uses `-` instead of borrowing the shipping name when the billing name has no last-name segment, so Amazon Pay shipping recipients do not get mixed into billing details.
- Updates the 10.8.0 changelog entries.

## Testing
- `npm run test:js -- client/express-checkout/utils/__tests__/normalize.test.js --runInBand`
- `npx eslint client/express-checkout/utils/normalize.js client/express-checkout/utils/__tests__/normalize.test.js`
- `git diff --check`
- `rg -n "getBillingLastNameFallback|getExistingBillingName|existingBillingLastName" client/express-checkout/utils`

ESLint still reports the existing Browserslist freshness warning.

Fixes STRIPE-1164


daledupreez (approved):
Thanks for pushing this change through, @ryanr14! I actually had an Amazon sandbox account that was perfect for testing this -- I had originally set it up with only a first name by mistake!

But I confirmed that the behaviour was as described for the flow where no customer details exist, as well as the case where the name matches for an existing customer and the last name is joined in. I can also confirm that this failed previously -- I had to stay away from using this account!

ryanr14:
Thanks for checking and testing! I'll merge now, woo!

daledupreez:
@ryanr14, as a quick follow-up on this: I have some concerns about defaulting in the last name. I think we should ship the PR without that fallback logic, and we can work out how risky it would be to combine that data via a dedicated, smaller change. I'd be happy to work on splitting that change out.

And to be clear, my concern is that we'd be combining data from two different sources in our system of record, which feels like it is not something we should be doing without thinking that through carefully.

ryanr14:
Ah ok, I can work to pull out that logic and then think through the best way to solve that in another issue. 🙂

ryanr14:
Pushed `e12a0871` to remove the stored customer billing-address last-name reuse from this PR.

The billing name normalization now only uses the provider-supplied `billingDetails.name`; for single-name provider data, it still sends the `-` placeholder so STRIPE-1164 remains covered. Any future policy for combining stored customer data with provider billing data can happen in a separate, smaller follow-up.

Is this what you meant, @daledupreez, and if so, is this ok to merge now? We can then keep discussing some sort of smarter fallback logic in place of the current solution.

daledupreez:
Thanks for the updates, @ryanr14! I've just retested and everything is looking good. I'll get this merged once the tests pass in the next little bit.

ryanr14:
Woo, thanks!

PR #5461, opened 2026-05-20
Clear save-checkbox state when OC hides it for a non-reusable sub-method

wjrosa:
Follow-up to #5408 — addresses [Malith's review suggestion](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5408#pullrequestreview-4330490194) to also fix the root cause on the frontend.

`handleDisplayOfSavingCheckbox()` currently flips a body class so the save-checkbox container disappears via CSS, but the underlying state stays untouched. Result: a checkbox the shopper checked while `card` was the active OC sub-method stays checked when the PE switches to iDEAL / Bancontact / etc. — and the request body still carries the save signal (`wc-stripe-new-payment-method=1` on classic, `shouldSavePaymentMethod=true` on Blocks).

#5408's backend gate drops `setup_future_usage` regardless, so this PR is **defence-in-depth**: keep request bodies honest so any future server-side code that reads those fields can trust them, and surface the right thing to integrations / logs / order-attribution tooling that may inspect the payload.

Both OC paths (block-checkout and classic-with-OC) now route through a single `clearSaveCheckboxState()` helper called when the body class is toggled to "hide":

- **Classic** — `.checked = false` on `#wc-stripe-new-payment-method`.
- **Blocks** — dispatch `__internalSetShouldSavePaymentMethod( false )` on `wc/store/payment` via `wp.data`. That's the React store WC Blocks reads `shouldSavePaymentMethod` from; clearing the DOM input alone wouldn't work because the input is React-controlled. Wrapped in feature detection + `try`/`catch` so classic-only contexts that don't ship `wp.data` no-op cleanly.

The clearing is **only on transitions to hide**. When a method stays reusable and the checkbox stays visible, we never touch the state — the shopper's value is preserved.

**Update — Adaptive Pricing path:** when Adaptive Pricing is enabled, Blocks OC renders the checkout-sessions `CheckoutForm` (its own Payment Element) instead of `PaymentProcessor`, and that path never called `handleDisplayOfSavingCheckbox`. As a result the checkbox stayed visible and checked for non-reusable sub-methods on Adaptive Pricing stores. The shared helper is now wired into `CheckoutForm`'s `onSelectedPaymentMethodChange` plus a mount effect (so the initial method is evaluated too), mirroring the `PaymentProcessor` path.

## Testing instructions

Pre-reqs: test mode, OCS on, global save on, EUR store, NL (iDEAL) or BE (Bancontact) address. Per-method save toggle OFF.

1. **Classic OC, iDEAL:** Load OCS checkout, **check the save box** while card is the active sub-method, then switch the PE to iDEAL. Open dev tools and inspect `#wc-stripe-new-payment-method` — its `.checked` should be `false`. Submit; the request body's `wc-stripe-new-payment-method` should not be `1`.
2. **Blocks OC, iDEAL:** Same sequence on Blocks checkout. Submit; the `POST /wc/store/v1/checkout` payload's `payment_data` should not include `wc-stripe-new-payment-method=true`, and the Stripe API call should not include `setup_future_usage`.
3. **Repeat for Bancontact** with a BE address.
4. **Regression — reusable method:** With `Enable saved iDEAL` ON, repeat (1) and (2) — the checkbox stays visible and the shopper's `checked` value must round-trip (we must not silently clear a value the shopper set).
5. **Regression — card path:** Card order with save checked, no method switch → token still saved.
6. **Adaptive Pricing (Blocks OC):** With Adaptive Pricing enabled, repeat (2)/(3). Confirm the checkbox hides for iDEAL/Bancontact (and for card when Link is enabled) and isn't carried on submit. Regression: with `Enable saved iDEAL` ON, the checkbox stays visible.
7. **Jest:** `npm run test:js -- --testPathPattern handle-display-of-saving-checkbox` (26 tests) and `--testPathPattern checkout-sessions/__tests__/checkout-form` (5 tests, +2 new for the Adaptive Pricing path).

## Why not stacked on #5408

Branched off `develop` rather than #5408's branch so either PR can ship independently. The backend gate stands alone; this PR is purely additive on the request-body shape. If #5408 lands first, this no-ops harmlessly on top; if this lands first, #5408 still has the same effect.

---

### Changelog entry

<details>

#### Comment

Changelog entry included for 10.8.0: "Ensure the save-payment-method checkbox is unchecked when a non-reusable payment method is selected in Optimized Checkout." It's worded generically across OC render paths, so it also covers the Adaptive Pricing fix added here.

</details>

**Post merge**


client/optimized-checkout/handle-display-of-saving-checkbox.js
```diff
@@ -68,6 +68,49 @@ const shouldHideSaveCheckboxFromConfig = ( method, paymentMethodsConfig ) => {
...
+ *
+ * The body-class toggle hides the container visually but doesn't reset the
+ * underlying state. Without this, a checkbox the shopper checked while a
+ * reusable sub-method was active inside the OC Payment Element stays
+ * checked when the PE switches to a non-reusable method (iDEAL, Bancontact,
+ * etc.) — and the request body still carries `wc-stripe-new-payment-method=1`
+ * (classic) or `shouldSavePaymentMethod=true` (Blocks). The backend gate
+ * added in 10.8.0 drops `setup_future_usage` regardless, so this is
+ * defence-in-depth: request bodies stay clean and any future server-side
+ * code that reads the field can trust it.
```
daledupreez: Let's simplify this comment -- this is a LOT of text to wade through.
```suggestion
 * This ensures that we don't include a stale value for the "save payment method"
 * input field when the field has been hidden due to a payment method change.
```

client/optimized-checkout/handle-display-of-saving-checkbox.js
```diff
@@ -68,6 +68,49 @@ const shouldHideSaveCheckboxFromConfig = ( method, paymentMethodsConfig ) => {
...
+ */
+const clearSaveCheckboxState = () => {
+	// Classic: the form input WC reads on submit.
+	const classicInput = document.getElementById(
+		'wc-stripe-new-payment-method'
+	);
+	if ( classicInput instanceof HTMLInputElement ) {
+		classicInput.checked = false;
```
daledupreez, line 84: Nit: if we know we're in classic mode, can we not return early?
```suggestion
		classicInput.checked = false;
		return;
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.8.0
+* Dev - Clear the save-payment-method checkbox state when Optimized Checkout hides it for a non-reusable sub-method, so the request body matches the UI
```
daledupreez: Nit: I think this should qualify as a `Fix`, and it may be worth rephrasing this slightly.
```suggestion
* Fix - Ensure the save-payment-method checkbox is unchecked when a non-reusable payment method is selected in Optimized Checkout
```

daledupreez (commented):
I am not sure that this is working as described. Things look OK for classic checkout. However, in block checkout, the checkbox remains enabled when switching between payment methods, regardless of whether they are reusable or not. As far as I can tell, the code to handle these changes doesn't seem to be triggered.

Image
Image

wjrosa:
Thanks @daledupreez! On the latest (cfd47b10), Blocks OC hides and clears the Bancontact checkbox for me. I did some debugging locally with the help of Claude. Can you try rebuilding and giving it another try, please?

daledupreez:
@wjrosa, I've just retested in https://deliciously-various.jurassic.ninja/ and I am still hitting the issue after making the first purchase and then coming back to block checkout -- the checkbox is staying visible and selected regardless of which payment method I pick. I'll send credentials via DM shortly.

wjrosa:
Thanks @daledupreez — reproduced and root-caused it. The trigger turned out to be **Adaptive Pricing**, not the saved card: with it enabled, Blocks OC renders the checkout-sessions `CheckoutForm` (its own Payment Element) instead of `PaymentProcessor`, and that path never called `handleDisplayOfSavingCheckbox`, so the checkbox stayed visible/checked for non-reusable methods. ("After first purchase" just lined up with the checkout session loading on return.)

Just pushed a fix that wires the shared helper into `CheckoutForm` (on method change + on mount). Could you rebuild and retest on the same site? It should now hide and clear for iDEAL/Bancontact while still showing for reusable methods.

daledupreez (approved):
I've just retested the block checkout + Adaptive Pricing case that was failing before and it is now working as it should.

I also think it's worth looking at the test issue raised by Code Rabbit in [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5461#pullrequestreview-4404584646), but I don't think that should be considered a major blocker.

PR #5462, opened 2026-05-20
[RSM - remote config] Send additional parameters to the remote-config server

malithsen:

Expand the remote-config request identity beyond `mode` and `plugin_version` so the WPCOM endpoint can target flag responses on additional store-shape signals. 

- WC version
- Account country
- Store currency
- Subscriptions enabled?
- Pre orders enabled?

I'm open to introducing more parameters if needed.

## Testing instructions
Code inspection should be enough. To test the request without the WPCOM component:

1. Set a breakpoint [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/8f4b7b6a411cefd6d78d6051ba498cde9cb1ef50/includes/remote-config/class-wc-stripe-remote-config-client.php#L47)
2. Go to Tools -> Scheduled Actions
3. Search for the `wc_stripe_remote_config_sync` action and manually run it.
4. Inspect the request shape when the breakpoint is reached.

---

### Changelog entry

<details>

#### Comment 
Modifies an unreleased feature

</details>

**Post merge**


includes/remote-config/class-wc-stripe-remote-config-client.php
```diff
@@ -90,4 +87,41 @@ public function fetch( string $mode ) {
...
+	private function build_query_args( string $mode ): array {
+		return [
+			'mode'                  => $mode,
+			'plugin_version'        => WC_STRIPE_VERSION,
+			'wc_version'            => defined( 'WC_VERSION' ) ? WC_VERSION : '',
+			'account_country'       => $this->get_account_country( $mode ),
+			'store_currency'        => function_exists( 'get_woocommerce_currency' ) ? get_woocommerce_currency() : '',
+			'subscriptions_enabled' => $this->bool_param( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() ),
+			'pre_orders_enabled'    => $this->bool_param( class_exists( 'WC_Pre_Orders' ) ),
+		];
```
daledupreez: Should we remove values that are `''`? It's not a big thing, but dropping unknown values from the query feels like it would reduce the overhead.

includes/remote-config/class-wc-stripe-remote-config-client.php
```diff
@@ -90,4 +87,41 @@ public function fetch( string $mode ) {
...
+	 * @return array<string, string>
+	 */
+	private function build_query_args( string $mode ): array {
+		return [
+			'mode'                  => $mode,
+			'plugin_version'        => WC_STRIPE_VERSION,
+			'wc_version'            => defined( 'WC_VERSION' ) ? WC_VERSION : '',
+			'account_country'       => $this->get_account_country( $mode ),
+			'store_currency'        => function_exists( 'get_woocommerce_currency' ) ? get_woocommerce_currency() : '',
+			'subscriptions_enabled' => $this->bool_param( WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled() ),
```
wjrosa, line 114: `account_country` collapses to `''` on a cache miss (cold install, post-flush, account not yet connected). WPCOM can't tell that apart from a real country signal. Worth either omitting the key entirely when empty (via `array_filter`), or documenting that empty == unknown so the server side knows not to target against it.

includes/remote-config/class-wc-stripe-remote-config-client.php
```diff
@@ -90,4 +87,41 @@ public function fetch( string $mode ) {
...
+			'pre_orders_enabled'    => $this->bool_param( class_exists( 'WC_Pre_Orders' ) ),
+		];
+	}
```
wjrosa, line 124: `pre_orders_enabled` is a shallower signal than the name implies — `class_exists( 'WC_Pre_Orders' )` only tells you the plugin is loaded, not that the merchant has any pre-orderable products. Contrast with `subscriptions_enabled` which goes through `WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled()`. Either rename the param (e.g. `pre_orders_plugin_active`) or push the check deeper, otherwise WPCOM will target on a noisy signal.

includes/remote-config/class-wc-stripe-remote-config-client.php
```diff
@@ -90,4 +87,41 @@ public function fetch( string $mode ) {
...
+	 * @param string $mode 'live' or 'test'.
+	 *
+	 * @return array<string, string>
+	 */
+	private function build_query_args( string $mode ): array {
+		return [
+			'mode'                  => $mode,
+			'plugin_version'        => WC_STRIPE_VERSION,
+			'wc_version'            => defined( 'WC_VERSION' ) ? WC_VERSION : '',
+			'account_country'       => $this->get_account_country( $mode ),
```
wjrosa: `defined( 'WC_VERSION' )` is defensive but the plugin's WC dependency means this branch is unreachable in practice — `WC_VERSION` is always defined when this method runs. Consistent with how the rest of the codebase treats WC as a hard requirement. Safe to drop the guard.

includes/remote-config/class-wc-stripe-remote-config-client.php
```diff
@@ -90,4 +87,41 @@ public function fetch( string $mode ) {
...
+	 *
+	 * @param string $mode 'live' or 'test'.
+	 *
+	 * @return array<string, string>
+	 */
+	private function build_query_args( string $mode ): array {
+		return [
+			'mode'                  => $mode,
+			'plugin_version'        => WC_STRIPE_VERSION,
+			'wc_version'            => defined( 'WC_VERSION' ) ? WC_VERSION : '',
```
wjrosa: The existing `test_fetch_request_shape_and_decoded_body` only asserts on `mode` and `plugin_version`. Adding 5 new args without extending that test (or adding a sibling) means a future change can silently drop one and CI won't catch it. A few extra `assertStringContainsString` lines or a dedicated test would close this.

wjrosa (commented):
Five new identity signals, single file, clean. A few questions on edge cases and signal fidelity.

includes/remote-config/class-wc-stripe-remote-config-client.php
malithsen: Added tests in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5462/commits/2b7456528886c2ebdef09ed425b05f38329a39b2

includes/remote-config/class-wc-stripe-remote-config-client.php
malithsen: Makes sense, we don't do `defined` for `WC_VERSION` in other places in this repo either. Dropped in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5462/commits/a82c8678c4f77a1d0e1fe1f98eb1a8ff4782a82e

includes/remote-config/class-wc-stripe-remote-config-client.php
malithsen: Makes sense, filtering out keys with empty values in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5462/commits/3d7f7dda86a7e2c4fcebfc24da125477f286cdc8

includes/remote-config/class-wc-stripe-remote-config-client.php
malithsen: Makes sense, filtering out keys with empty values in https://github.com/woocommerce/woocommerce-gateway-stripe/commit/3d7f7dda86a7e2c4fcebfc24da125477f286cdc8

includes/remote-config/class-wc-stripe-remote-config-client.php
malithsen: >  Contrast with subscriptions_enabled which goes through WC_Stripe_Subscriptions_Helper::is_subscriptions_enabled()

This contrast doesn't look accurate. [is_subscriptions_enabled](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/2b7456528886c2ebdef09ed425b05f38329a39b2/includes/compat/class-wc-stripe-subscriptions-helper.php#L43) does a simple class exists check too. These parameters are also named as `*_enabled`, which doesn't necessarily indicate whether pre-order/subscription products exist in the store.

wjrosa (approved):
Looks good to me 👍

diegocurbelo (approved):
The change looks good to me, and it tests well 👍🏼

daledupreez:
@wjrosa, this looks clean and ready to merge. Is there anything blocking us from getting this merged into the feature branch?

wjrosa:
@daledupreez, the main PR/feature branch has been in review for a long time, so I was thinking of merging it first to keep the original scope and make it easier for reviewers to re-review. But happy to merge it there if you think that's better

daledupreez:
Ah, that makes sense, @wjrosa! I'll try to expedite review of the main PR.

PR #5463, opened 2026-05-20
Update and refactor supported countries logic

daledupreez:
Related to #5053

This PR refactors how we represent supported account countries and supported billing countries for payment methods, as our logic was not consistent across payment methods, with the `supported_countries` property being used for both checks in different contexts. (See [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5053#pullrequestreview-4317483201) and some of the other history on #5053 for more.)

The main changes are to split out the supported values into distinct `supported_account_countries`, `unsupported_account_countries`, and `supported_billing_countries` properties and `is_available_for_account_country()` and `is_available_for_billing_country()` methods, as well as introducing new `SUPPORTED_ACCOUNT_COUNTRIES`, `SUPPORTED_BILLING_COUNTRIES`, and `UNSUPPORTED_ACCOUNT_COUNTRIES` constants on the base `WC_Stripe_UPE_Payment_Method` class that can be overridden by subclasses. Note that the main use case for `UNSUPPORTED_ACCOUNT_COUNTRIES` is for payment methods that are supported across most countries, but are unsupported in only a few countries.

I have also introduced a `__get()` magic method on the `WC_Stripe_UPE_Payment_Method` to log a deprecation message for any access of the `supported_countries` property, and to return the value of `supported_account_countries`.

In addition to the refactor, I've reviewed all the current Stripe documentation for the payment methods to ensure we're current across the various billing and account restrictions.

## Testing instructions

Code inspection is going to be important, though you can also point your friendly LLM to Stripe's markdown docs at (https://docs.stripe.com/payments/payment-methods/overview.md) and see what it comes up with. You may also need to scrape the full set of supported countries from (https://stripe.com/global) after excluding the _Preview_ and _Extended network_ countries.

After that, we need to verify that the logic for `is_available_for_account_country()` and `is_available_for_billing_country()` make sense, and are set up as expected, especially for the payment methods that rely on `unsupported_account_countries` and `UNSUPPORTED_ACCOUNT_COUNTRIES`.

Finally e2e tests should still be clean -- there should be no functional changes except for some edge cases around when specific payment methods are available.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -137,6 +177,22 @@ public function __construct() {
...
+	 * @param string $property The property name.
+	 * @return mixed
+	 */
+	public function __get( $property ) {
+		if ( 'supported_countries' === $property ) {
+			wc_doing_it_wrong( get_class( $this ) . '->supported_countries', 'Use supported_account_countries or supported_billing_countries instead.', '10.8.0' );
+			return $this->supported_account_countries;
+		}
+
+		throw new Exception( 'Property ' . $property . ' is not defined.' );
```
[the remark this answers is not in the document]
daledupreez: Addressed in ecdcb19822807ea8e398c57b8fe1a0959bec2574.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -345,22 +401,39 @@ public function is_enabled_at_checkout( $order_id = null, $account_domestic_curr
 	 * @return array Supported customer locations.
 	 */
 	public function get_available_billing_countries() {
-		$account         = WC_Stripe::get_instance()->account->get_cached_account_data();
-		$account_country = isset( $account['country'] ) ? strtoupper( $account['country'] ) : '';
+		if ( $this->has_domestic_transactions_restrictions() ) {
+			$account         = WC_Stripe::get_instance()->account->get_cached_account_data();
+			$account_country = isset( $account['country'] ) ? strtoupper( $account['country'] ) : '';
+			return [ $account_country ];
```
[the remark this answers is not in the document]
daledupreez: Partially addressed in 35b76ad6fd5af4df76724fbd1883eb7d8cb878df -- I added a comment to note that this is intentional, as returning `[]` implies that all billing countries are supported, which is not the case.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -345,22 +401,39 @@ public function is_enabled_at_checkout( $order_id = null, $account_domestic_curr
...
+	 * Whether the given billing country for a shopper is permitted to use this payment method.
+	 *
+	 * @param string $country_code Two-letter ISO country code.
+	 *
+	 * @return bool True when no restriction is set or the country is in the supported list.
+	 */
+	public function is_available_for_billing_country( $country_code ): bool {
+		if ( ! empty( $this->supported_billing_countries ) ) {
+			return in_array( $country_code, $this->supported_billing_countries, true );
 		}
```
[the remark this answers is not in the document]
daledupreez: Partially addressed in c861d2a2e777136610fc0f459976cfc518bb9e71 for `is_available_for_billing_country()`.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -137,6 +177,29 @@ public function __construct() {
...
+	 * Magic method to get properties.
+	 * Used for backwards compatibility with deprecated properties.
+	 *
+	 * @param string $property The property name.
+	 * @return mixed
+	 */
+	public function __get( $property ) {
+		if ( 'supported_countries' === $property ) {
+			wc_deprecated_function( get_class( $this ) . '->supported_countries', 'Use supported_account_countries or supported_billing_countries instead.', '10.8.0' );
+			return $this->supported_account_countries;
```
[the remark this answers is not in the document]
daledupreez: Switched back to calling `wc_doing_it_wrong()` in 60de05b026841155fceb139b969a1b60d39163ff.

diegocurbelo (commented):
The scope is large but the refactor is mechanical, and in general looks good to me.

Should be add also `__isset()` to `class-wc-stripe-upe-payment-method.php`? In case some custom code uses `isset()` or `empty()` on `supported_countries`

includes/payment-methods/class-wc-stripe-upe-payment-method-acss.php
```diff
@@ -14,23 +14,42 @@ class WC_Stripe_UPE_Payment_Method_ACSS extends WC_Stripe_UPE_Payment_Method {
...
-		$this->supported_currencies     = [ WC_Stripe_Currency_Code::CANADIAN_DOLLAR ]; // The US dollar is also supported, but has a high risk of failure since only a few Canadian bank accounts support it.
-		$this->supported_countries      = [ WC_Stripe_Country_Code::CANADA ];
-		$this->label                    = __( 'Pre-Authorized Debit', 'woocommerce-gateway-stripe' );
-		$this->description              = __(
+		$this->stripe_id                     = self::STRIPE_ID;
+		$this->title                         = __( 'Pre-Authorized Debit', 'woocommerce-gateway-stripe' );
+		$this->is_reusable                   = true;
+		$this->supported_currencies          = [ WC_Stripe_Currency_Code::CANADIAN_DOLLAR ]; // The US dollar is also supported, but has a high risk of failure since only a few Canadian bank accounts support it.
+		$this->supported_account_countries   = self::SUPPORTED_ACCOUNT_COUNTRIES;
+		$this->unsupported_account_countries = self::UNSUPPORTED_ACCOUNT_COUNTRIES;
```
Mayisha: Nit: the unsupported assignment seems redundant here and in some other classes, as there is no 'UNSUPPORTED_ACCOUNT_COUNTRIES' listed in the corresponding classes.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -78,12 +102,28 @@ abstract class WC_Stripe_UPE_Payment_Method extends WC_Payment_Gateway {
...
+	 */
+	protected $unsupported_account_countries = [];
+
+	/**
+	 * Customer billing countries permitted to use this payment method at checkout.
+	 * Empty means no restriction (all billing countries permitted).
+	 *
+	 * @var string[]
+	 */
+	protected $supported_billing_countries = [];
```
Mayisha, line 126: Nit.
I see in the payment method sub-classes we have the following assignments in the constructors. We could do this in the base class constructor.

```
$this->supported_account_countries   = self::SUPPORTED_ACCOUNT_COUNTRIES;
$this->unsupported_account_countries = self::UNSUPPORTED_ACCOUNT_COUNTRIES;
$this->supported_billing_countries   = self::SUPPORTED_BILLING_COUNTRIES;
```

Mayisha (approved):
Code changes look good to me. On the checkout pages (block/shortcode) payment methods are shown and hidden correctly when I change the billing address (i,e. Klarna, ACH are hidden when I select BD as the billing country and re-appears when I switch the billing country to US)

I also agree with Diego's suggestion of adding the `__isset()`  for the deprecated property.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
daledupreez: I moved the property initialisation to the parent class in 9c7ce5d71e71f6ce20c73289f10418aa0be4b578, but the logic needs to use `static::*` to ensure that we're looking at the overridden constant values and not the values defined in the parent class.

includes/payment-methods/class-wc-stripe-upe-payment-method-acss.php
daledupreez: I removed all the initialisation logic into the parent constructor in 9c7ce5d71e71f6ce20c73289f10418aa0be4b578 -- that way child classes don't need to do anything, and we ensure that all three properties are always initialised, and we don't miss any values in children moving forward.

daledupreez:
I've also implemented `__isset()` in 94da9bf181f49f12d3c4432e14e7b76dd9e6ca18.

daledupreez:
@diegocurbelo, would you mind taking another look at this PR?

diegocurbelo (approved):
Thanks for the changes @daledupreez!

LGTM 👍🏼

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -135,6 +175,54 @@ public function __construct() {
...
+	}
+
+	/**
+	 * Magic method to check if properties are set.
+	 * Used for backwards compatibility with deprecated properties.
+	 *
+	 * @param string $property The property name.
+	 * @return bool
+	 */
+	public function __isset( $property ): bool {
```
Mayisha, line 219: Optional suggestion: Do you plan to remove the `__get` and `__isset` after a few releases? In that case, we could add a comment in the method that this would be deleted in the future.

Mayisha (approved):
Looks good to me 💯

includes/payment-methods/class-wc-stripe-upe-payment-method.php
daledupreez: We may deprecate more fields, so the methods may linger for a while. That said, I added `@internal` entries to both sets of PHPDoc so we can try to prevent direct access in the meantime.

PR #5466, opened 2026-05-22
Add ECE preconnect header and bundle preload

diegocurbelo:
Fixes STRIPE-1079

Two low-risk network hints to cut wallet-button paint time on every page where the Express Checkout Element (ECE) renders — product, cart, checkout, pay-for-order, change-payment-method.

  - **Preconnect to Stripe hosts.** Filters `wp_resource_hints` to add `preconnect` entries for `js.stripe.com`, `m.stripe.network`, and `q.stripe.com`. WordPress core already auto-emits
  `dns-prefetch` for `js.stripe.com` via `wp_dependencies_unique_hosts()`, but DNS resolution alone doesn't cover the TCP + TLS handshake. The two sub-hosts are fetched at runtime by
  Stripe.js itself, so core can't infer them — we add them explicitly.
  - **Preload the Express Checkout bundle.** Filters `wp_preload_resources` to add a `<link rel="preload" as="script">` for `build/express-checkout.js`. Core handles the `<link>` emission, escaping, and dedup; the hint lands in `<head>` ahead of the footer enqueue, so the browser's preload scanner overlaps the bundle fetch with the rest of `<head>` parsing.

  Both callbacks short-circuit on pages where ECE wouldn't render anyway, reusing the existing `is_page_supported() && should_show_express_checkout_button()` guards from `scripts()`.
  Registration is unconditional in `init()`; per-page gating lives inside the callbacks (same pattern as the existing always-on filters at the top of `init()`).

  Expected impact: ~100–300 ms faster first paint of Apple Pay / Google Pay / Link / Amazon Pay buttons. Preconnect/preload are advisory — if a host is wrong, browsers silently skip, so blast radius is effectively zero.

## Testing instructions

Automated:

  ```
  npm run test:php:parallel -- --filter=WC_Stripe_Express_Checkout_Element_Test
  ```

  Manual:

  1. Connect Stripe in test mode with Apple Pay (and Google Pay / Link if available) enabled.
  2. Open a product page in Chrome → View Source. Confirm the following `<link>` tags appear in `<head>`:
     - `<link rel='preconnect' href='https://js.stripe.com' crossorigin='anonymous'>`
     - `<link rel='preconnect' href='https://m.stripe.network' crossorigin='anonymous'>`
     - `<link rel='preconnect' href='https://q.stripe.com'>`
     - `<link rel="preload" as="script" href=".../build/express-checkout.js?ver=...">`
  3. Repeat on cart, checkout (classic + Blocks), pay-for-order, change-payment-method. Tags should be present on all.
  4. Visit a page where ECE doesn't render (e.g., a CMS page, or the About page). Confirm none of the four `<link>` tags appear.
  5. DevTools → Network: connections to `js.stripe.com` / `m.stripe.network` are initiated earlier than before; wallet buttons paint sooner.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -428,6 +435,84 @@ private function register_express_checkout_script() {
...
+			return $urls;
+		}
+
+		if ( ! $this->express_checkout_helper->is_page_supported() ) {
+			return $urls;
+		}
+
+		if ( ! $this->express_checkout_helper->should_show_express_checkout_button() ) {
+			return $urls;
+		}
```
daledupreez, line 455: Related thought: how many times do we call these helper methods on each request? Might it make sense to have each helper use a static variable to avoid double-work? It might need to be a static variable on the helper class to make it possible to reset for tests etc.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -84,6 +84,13 @@ public function init() {
 		// form submission, regardless of which page hook bootstrapped this class.
 		add_filter( 'woocommerce_subscription_note_new_payment_method_title', [ $this, 'filter_change_payment_method_note_title' ], 10, 3 );
 
+		// Network hints for the wallet-button paint path. The callbacks short-circuit on
+		// pages where ECE isn't going to render, so registering unconditionally is safe.
+		// `wp_head` priority 1 emits the preload before WP core's resource_hints (priority 2),
+		// keeping the bundle fetch ahead of the browser's preload scanner cutoff.
+		add_filter( 'wp_resource_hints', [ $this, 'add_resource_hints' ], 10, 2 );
+		add_action( 'wp_head', [ $this, 'print_ece_bundle_preload' ], 1 );
```
daledupreez: I think we can use the `wp_preload_resources` filter rather than rolling our own.

https://github.com/WordPress/wordpress-develop/blob/bc730ed53034c4cdfa8995dbb644ce555a75f1bf/src/wp-includes/general-template.php#L3763-L3787

includes/payment-methods/class-wc-stripe-express-checkout-element.php
daledupreez, line 72: Should we also check whether express checkout should be shown in the current context? If we know we don't have any express checkout methods enabled in the current page, we probably don't need most of these hooks. We should also check whether this condition is only looking at Apple Pay/Google Pay or all express checkouts.

changelog.txt
```diff
@@ -21,6 +21,7 @@ xxxx-xx-xx - version 10.8.0
 * Dev - Move some independent classes into autoloader
 * Dev - Add Jest tests for ECE payment flow
 * Fix - Respect the iDEAL/Wero and Bancontact save toggle when Optimized Checkout is enabled
+* Tweak - Preconnect to Stripe hosts and preload the Express Checkout bundle so wallet buttons paint sooner
```
daledupreez: I think this qualifies as an update, and should be communicated as such. 😁 
```suggestion
* Update - Improve express checkout load times by preloading the resources
```

includes/payment-methods/class-wc-stripe-express-checkout-element.php
diegocurbelo: There is another draft PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5467 to improve that

includes/payment-methods/class-wc-stripe-express-checkout-element.php
diegocurbelo: The hooks are registered after the early return at line 72, so we already skip when ECE is not enabled disabled. We don't skip in a per-page case where ECE is enabled but the current page isn't supported — those still call into `should_show_express_checkout_button()`, in any case the PR #5467 also fixes this.

includes/payment-methods/class-wc-stripe-express-checkout-element.php
diegocurbelo: Refactored in [6967987](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5466/commits/6967987fdc0e2c9c52a90987b114fa0bf205a24b)

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -84,6 +84,11 @@ public function init() {
 		// form submission, regardless of which page hook bootstrapped this class.
 		add_filter( 'woocommerce_subscription_note_new_payment_method_title', [ $this, 'filter_change_payment_method_note_title' ], 10, 3 );
 
+		// Network hints for the wallet-button paint path. The callbacks short-circuit on
+		// pages where ECE isn't going to render, so registering unconditionally is safe.
```
daledupreez: I think we can improve the inline comment here.
```suggestion
		// Add network and preload hints for express checkout resources.
		// The hooks check whether express checkout will be shown, and are no-ops in other contexts.
```

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -428,6 +433,87 @@ private function register_express_checkout_script() {
...
+	/**
+	 * Append preconnect URLs for Stripe-owned hosts when the current page will render ECE.
+	 *
+	 * WordPress core already emits a `dns-prefetch` for `js.stripe.com` via
+	 * `wp_dependencies_unique_hosts()` (because we register the Stripe.js script). DNS
+	 * resolution is only one step of the connection; `preconnect` also performs the TCP +
+	 * TLS handshake, saving ~100–200 ms on the wallet-button paint. The two sub-hosts
+	 * (`m.stripe.network`, `q.stripe.com`) are fetched at runtime by Stripe.js itself, so
+	 * core can't infer them and we add them explicitly.
+	 *
```
daledupreez: We can remove this from the main method doc. It feels like noise to me, especially given that we are still adding `js.stripe.com` to ensure we also start a TLS handshake.

```suggestion
```

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -428,6 +433,87 @@ private function register_express_checkout_script() {
...
+		$asset_data = $this->get_asset_data();
+
+		$preload_resources[] = [
+			'href' => add_query_arg(
+				'ver',
+				$asset_data['version'],
+				WC_STRIPE_PLUGIN_URL . '/build/express-checkout.js'
+			),
+			'as'   => 'script',
+		];
```
daledupreez: Might it be worth adding a defensive check to make sure we have the fields we expect in `$asset_data`?
```suggestion
		if ( is_array( $asset_data ) && isset( $asset_data['version'] ) ) {
			$preload_resources[] = [
				'href' => add_query_arg(
					'ver',
					$asset_data['version'],
					WC_STRIPE_PLUGIN_URL . '/build/express-checkout.js'
				),
				'as'   => 'script',
			];
		}
```

daledupreez (approved):
Overall, this is looking good and testing well.

Please update the PR description to reflect that we're using supported WordPress filters and not the custom logic documented in the PR for the **Preload the Express Checkout bundle** section.

It may also be worth noting that these changes should add minimal overhead once #5467 gets merged -- without that change, this PR would result in two additional calls to `should_show_express_checkout_button()` and would need to run the checks multiple times.

In testing, I also noticed that a number of secondary requests are fetching resources from `b.stripecdn.com`. Might it be worth preconnecting to that domain as well?

And as one further thought, might it be worth using a per-request cache for `is_page_supported()`? I don't think that belongs in this PR, but it might be useful given that we're calling that helper at least 2 more times per request as a result of this PR.

PR #5467, opened 2026-05-22
Memoize should_show_express_checkout_button() per request

diegocurbelo:
Fixes STRIPE-1079

`WC_Stripe_Express_Checkout_Helper::should_show_express_checkout_button()` is the 16-guard check that decides whether the Express Checkout Element renders on the current page. It's called 3 times per request from independent callers:
  - `WC_Stripe_Blocks_Support::get_payment_method_data()`
  - `WC_Stripe_Express_Checkout_Element::scripts()`
  - `WC_Stripe_Express_Checkout_Element::display_express_checkout_button_html()`

This PR wraps the public method with a request-scoped cache and moves the existing body into a new private `compute_should_show_express_checkout_button()`. No guard logic changes; the inner method is identical to today's body.

## Testing instructions

Code review, and verify all checks pass.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -42,6 +42,17 @@ class WC_Stripe_Express_Checkout_Helper {
 	 */
 	private $gateway;
 
+	/**
+	 * Request-scoped cache for `should_show_express_checkout_button()`.
+	 *
+	 * Keyed by whether we're inside the `woocommerce_after_add_to_cart_form` action,
+	 * because the One Page Checkout branch in `compute_should_show_express_checkout_button()`
+	 * consults `doing_action()` and would otherwise return stale answers across call sites.
```
daledupreez: I think the PHPDoc here should be less specific to this specific situation. That level of documentation makes more sense close to the code that computes the separate cache key.
```suggestion
	 * Keys in the array capture specific contexts that may apply within a request.
	 * Once such case is for handling One Page Checkout, which can affect when we are in checkout via runtime hooks.
```

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -745,6 +756,23 @@ public function should_show_ece_on_change_payment_method_location() {
 	 * @return  boolean  True if express checkout elements are supported on current page, false otherwise
 	 */
 	public function should_show_express_checkout_button() {
+		$cache_key = doing_action( 'woocommerce_after_add_to_cart_form' ) ? 'in_atc' : 'default';
```
daledupreez: As noted above, it's worth adding explicit commentary here as to why this is happening.

I also think we can use a better cache key so it's more self-descriptive.

```suggestion
		// One Page Checkout can make a page act as a checkout page via runtime hooks.
		// We need a separate cache key for that case to ensure we enable express checkout.
		$cache_key = doing_action( 'woocommerce_after_add_to_cart_form' ) ? 'after_add_to_cart' : 'default';
```

daledupreez (approved):
The changes look good to me, though I have some minor suggestions on the doc and the cache key naming that are not blocking.

Thanks for getting this shipped.

PR #5472, opened 2026-05-22
Add per-product exclude toggle for Agentic Commerce sync

wjrosa:
Towards STRIPE-1168

Depends on #5598, which adds the WooCommerce-core-prefixed `woocommerce_agentic_commerce_should_sync_product` filter and deprecates the Stripe-prefixed `wc_stripe_agentic_commerce_should_sync_product`. This PR adds the per-product UI and storage on top of that filter, and is targeted at #5598's branch. **Merge #5598 first.**

Adds an "Agentic Commerce" exclude checkbox to the WooCommerce product editor's Inventory tab. When checked, the product is dropped from every Agentic Commerce surface: the full-feed walker, the inventory-change tracker, and the archive tracker.

The work is split across two classes:

- **`WC_Stripe_Agentic_Commerce_Product_Exclusion`** — owns the flag storage (`is_excluded()` / `set_excluded()`) and registers the `woocommerce_agentic_commerce_should_sync_product` filter, voting `false` when the per-product meta is `'yes'`. The filter registers in every context (cron/CLI included), independent of admin.
- **`WC_Stripe_Agentic_Commerce_Product_Meta_Box`** — the admin-only editor UI on top of that storage.

No new code path inside the sync pipeline; the filter does all the work.

Storage contract: underscore-prefixed hidden post meta (`_wc_stripe_agentic_commerce_exclude`), `'yes'`/`'no'` string values, rendered via `woocommerce_wp_checkbox()` for visual consistency with the surrounding Inventory checkboxes, with a static `is_excluded()` reader as the single source of truth.

The checkbox renders only for `simple` and `variable` products (the only types the feed syncs) and only while the merchant has Agentic Commerce enabled in settings. Variations inherit the parent's flag — the checkbox surfaces on the parent only, so reading a variation's own meta would always be `'no'` and let variants leak past the parent's setting during sync.

Preview:
Screenshot 2026-05-22 at 11 25 52

## Testing instructions

1. **Enable the Agentic Commerce developer feature flag.** The toggle registers inside `WC_Stripe::initialize_agentic_commerce()`, which short-circuits when `WC_Stripe_Feature_Flags::is_agentic_commerce_enabled()` returns false. Flip it on locally: `wp option update _wcstripe_feature_agentic_commerce yes` (or hook the `wc_stripe_is_agentic_commerce_enabled` filter to return true). Also make sure the merchant-facing toggle is enabled in **WooCommerce → Settings → Payments → Stripe → Agentic Commerce**.
2. Edit a published product → **Inventory** tab → tick **"Agentic Commerce"** (description "Exclude from the Stripe Agentic Commerce catalog sync") → update.
3. Trigger a full sync (`wp stripe agentic-commerce sync` or the admin button) → confirm the product does not appear in the uploaded CSV / Stripe Files API payload.
4. Untick the checkbox → save → re-run the sync → confirm the product reappears.
5. Change the product's stock quantity → confirm the inventory tracker queues an update only when the product is NOT excluded; when excluded, no entry is queued and any stale queued entry for that product is evicted on the next save.
6. Create a variable product, tick the checkbox on the parent → confirm all variations are excluded from the feed.
7. Disable the merchant toggle in settings → reload the product editor → confirm the **Agentic Commerce** checkbox is no longer rendered.

---

-   [x] Tested on mobile (or does not apply) — admin-only UI, n/a

### Changelog entry

<details>

#### Comment

Entry added: `Add - Per-product "Agentic Commerce" exclude toggle on the product editor's Inventory tab`.

</details>

**Post merge**


TateLyman:
I think the immediate-resync path is currently masked by the normal recurring feed sync.

`save_meta()` fires `wc_stripe_agentic_commerce_schedule_full_resync` on a real value change, but `schedule_full_resync_now()` returns early whenever `as_has_scheduled_action( self::SCHEDULED_ACTION )` is true. In the normal feature-enabled state, `activate()` has already scheduled the recurring `self::SCHEDULED_ACTION`, so the toggle change can save the meta but enqueue no immediate one-off resync. That leaves the previously exported product in Stripe until the normal recurring full-feed run, which is the stale-catalog gap this PR is trying to close.

The new test covers the idle case after `as_unschedule_all_actions()`, but it does not cover the active-recurring case. I would add a test that first schedules the recurring action, then fires `wc_stripe_agentic_commerce_schedule_full_resync`, and asserts that a separate immediate sync is still queued.

Patch direction: dedupe only an already-pending immediate resync, not the recurring cadence. A distinct immediate hook handled by `sync_feed()`, or a distinguishable immediate-only action signature, would avoid the recurring action suppressing the catalog-convergence job.

I only did a public diff read plus PHP syntax check on the new meta-box class/test; no store credentials or production calls involved.

wjrosa:
@TateLyman Good catch — confirmed. The recurring `SCHEDULED_ACTION` always keeps a pending occurrence in the enabled state, so the `as_has_scheduled_action()` guard early-returned every time and the immediate resync was never enqueued.

Fixed in 0507759f3: the immediate resync now uses a distinct `IMMEDIATE_SYNC_ACTION` hook (also wired to `sync_feed()`), so it dedupes only against a pending *immediate* job and the recurring cadence no longer masks it. Also unscheduled on deactivate, and added the active-recurring regression test you suggested.

includes/class-wc-stripe.php
```diff
@@ -994,6 +994,13 @@ public function initialize_agentic_commerce() {
 			WP_CLI::add_command( 'stripe agentic-commerce', 'WC_Stripe_Agentic_Commerce_CLI' );
 		}
 
+		// Per-product exclude toggle. WC AI Storefront owns this surface when active —
+		// its product-selection UI funnels into the same should-sync filter, so
+		// rendering both at once would let merchants set conflicting expectations.
+		if ( ! defined( 'WC_AI_STOREFRONT_VERSION' ) && class_exists( 'WC_Stripe_Agentic_Commerce_Product_Meta_Box' ) ) {
```
malithsen: Shouldn't we also gate this behind the "Enable agentic commerce" checkbox in settings? Right now the checkbox in Inventory tab shows up even when agentic commerce is disabled in Stripe

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * Render the "Agentic Commerce" exclude checkbox.
+	 *
+	 * Uses WC's `woocommerce_wp_checkbox()` helper so the visual treatment
+	 * matches the surrounding Inventory checkboxes (Manage stock, Sold
+	 * individually, etc.). The description renders inline next to the
+	 * checkbox (no tooltip) so the opt-out behavior is visible at a glance.
+	 */
+	public function render_checkbox(): void {
```
malithsen, line 68: Does agentic mode support syncing all product types? I haven't tested, but my recollection is that it only supports simple and variable products. Right now I see the checkbox under subscription products.

malithsen (commented):
I have a couple of questions after testing.

includes/class-wc-stripe.php
wjrosa: Good catch — fixed. The meta box only gated on the developer feature flag, not the merchant `ENABLED_OPTION`. `render_checkbox()` and `save_meta()` now bail on `is_merchant_enabled()`, so the toggle disappears when Agentic Commerce is off in settings (and a hidden, unposted checkbox can't clobber a stored opt-out).

Drafted with assistance from Claude Code; reviewed and posted by me.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: Right — the feed query only walks `simple` + `variation` (so parents are `simple`/`variable`). The checkbox now renders only when `$product_object->get_type()` is `simple` or `variable`, so it no longer shows on subscription/grouped/external products. Added tests for both this and the enable gate.

Drafted with assistance from Claude Code; reviewed and posted by me.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,202 @@
...
+			// schedules a resync.
+			$previous_value = 'no';
+		}
+
+		update_post_meta( $product_id, self::META_KEY, $value );
+
+		// On a real change, kick the catalog so Stripe converges immediately rather
+		// than waiting for the next scheduled full sync to drop / re-add this product.
+		if ( $previous_value !== $value ) {
+			do_action( 'wc_stripe_agentic_commerce_schedule_full_resync' );
```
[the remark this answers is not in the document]
wjrosa: Fixed in d4ce7e35e — added the same `SUPPORTED_TYPES` gate to `save_meta()` via `wc_get_product()->get_type()`, so a type switch (e.g. simple → grouped) no longer clobbers a stored `yes` or fires a spurious resync. Added a regression test for that case.

changelog.txt
```diff
@@ -28,6 +28,8 @@ xxxx-xx-xx - version 10.8.0
 * Dev - Require explicit visibility for class constants
 * Dev - Add payment token classes to autoloader
 * Remove - Remove deprecated code up to version 10.1.0
+* Dev - Add wc_stripe_agentic_commerce_should_sync_product filter so adapters can exclude products from the Agentic Commerce catalog, inventory, and archive syncs
+* Add - Per-product "Agentic Commerce" exclude toggle on the product editor's Inventory tab (hidden when WC AI Storefront is active)
```
malithsen: Do we need two entries?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -45,19 +45,19 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
...
+	 * cadence keeps a pending occurrence at all times, so reusing that hook for
+	 * the on-demand resync would make `as_has_scheduled_action()` always report
+	 * "already pending" and silently drop every immediate convergence request.
+	 * A separate hook lets us dedupe immediate resyncs against each other
+	 * without the recurring action masking them.
 	 *
 	 * @var string
 	 * @since 10.8.0
 	 */
-	private const ASYNC_RESYNC_GROUP = 'wc-stripe-agentic-resync';
```
malithsen, line 60: This was introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5458 and removed in this PR. Is this intentional?

malithsen (commented):
This PR seems to alter/remove changes introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5458

It's not obvious to me if those are intentional. PR description only indicates the per-product toggle. I suspect we need some cleanup following the last rebase

changelog.txt
wjrosa: No — dropped the duplicate in de9e429a6. The `should_sync_product` filter entry was already shipped by #5458, so only the new per-product toggle entry belongs here. (changelog.txt + readme.txt.)

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Intentional. #5458's separate-group dedup still let the always-pending recurring action mask the immediate resync (TateLyman flagged it above), so `ASYNC_RESYNC_GROUP` is replaced by a distinct `IMMEDIATE_SYNC_ACTION` hook. No remaining references to the old const.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -45,19 +45,14 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
...
-	 * tick is queued.
+	 * Action Scheduler hook for an immediate, one-off full resync. Kept distinct
+	 * from {@see self::SCHEDULED_ACTION} — the recurring action is always pending,
+	 * so sharing it would make `as_has_scheduled_action()` drop every immediate resync.
 	 *
 	 * @var string
 	 * @since 10.8.0
 	 */
-	private const ASYNC_RESYNC_GROUP = 'wc-stripe-agentic-resync';
+	public const IMMEDIATE_SYNC_ACTION = 'wc_stripe_agentic_commerce_sync_feed_now';
```
daledupreez: Nit: Can we make this constant protected or private? The only usage is in this class and the related tests, where we can use `WC_Stripe_Test_Helper::get_const_value()` to get the value in the tests.
```suggestion
	protected const IMMEDIATE_SYNC_ACTION = 'wc_stripe_agentic_commerce_sync_feed_now';
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,167 @@
+<?php
+/**
+ * Per-product "Agentic Commerce" exclude meta box.
+ *
+ * Adds an Inventory-tab checkbox to opt a product out of the Stripe Agentic
+ * Commerce sync. Hidden when WC AI Storefront is active, since that plugin owns
```
daledupreez: My understanding is that relevant features from that project will be folded into WooCommerce AI, so we may need to be careful about how we approach the wording and the checks over the next few weeks. Maybe we can mark this as internal in some way as a starting point?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		return ! self::is_excluded( $product->get_id() );
+	}
+
+	/**
+	 * Product types the feed walks, and the only types the toggle renders for.
+	 * The feed syncs `simple` and `variation`; a variation's parent is `variable`.
+	 */
+	private const SUPPORTED_TYPES = [ 'simple', 'variable' ];
```
daledupreez: Two nits:
 * I think this const should be declared towards the top of the file and not in the middle like this.
 * We could optionally use the Woo constants for the two product types.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,167 @@
...
+		$previous_value = get_post_meta( $product_id, self::META_KEY, true );
+		if ( '' === $previous_value ) {
+			$previous_value = 'no';
+		}
+
+		update_post_meta( $product_id, self::META_KEY, $value );
+
+		// Converge Stripe now instead of waiting for the next scheduled sync.
+		if ( $previous_value !== $value ) {
+			do_action( 'wc_stripe_agentic_commerce_schedule_full_resync' );
```
daledupreez: Nit: Can we make this an explicit API call rather than a hook?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,167 @@
...
+	 * @param \WC_Product $product     Product being evaluated.
+	 * @return bool
+	 */
+	public function filter_should_sync_product( bool $should_sync, \WC_Product $product ): bool {
+		if ( ! $should_sync ) {
+			return false;
+		}
+
+		return ! self::is_excluded( $product->get_id() );
+	}
```
daledupreez: I don't think this logic should live within the metabox code class, as this feels like it's way more general and separate from the admin UI implementation (which is what the main metabox code should be responsible for). With that in mind, I think it would be better to have a dedicated class that is responsible for storing the flag and returning the value, as that should allow for a more loosely coupled implementation.

daledupreez (commented):
I have some feedback from inspection.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Done — it's `protected` now (in the prior commit), and the tests read it through `WC_Stripe_Test_Helper::get_class_const_value()` instead of touching the constant directly.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: Marked both the meta box and the new storage class `@internal` so we keep room to rework the wording/checks as this folds into WooCommerce AI.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: Moved the const to the top of the class and switched the values to `ProductType::SIMPLE`/`ProductType::VARIABLE`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: `save_meta()` now calls `( new …_Integration() )->schedule_full_resync_now()` directly. Left the `wc_stripe_agentic_commerce_schedule_full_resync` action registered since it's the documented adapter extension point — our own code just no longer routes through it.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: Pulled the flag storage (`is_excluded()`/`set_excluded()`) and the should-sync filter into a dedicated `WC_Stripe_Agentic_Commerce_Product_Exclusion` class. The meta box is now just the editor UI on top of it, and the filter registers in every context (cron/CLI) independent of admin. Split the tests to match.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
```diff
@@ -0,0 +1,114 @@
...
+	 * @return bool
+	 */
+	public function filter_should_sync_product( $should_sync, $product ): bool {
+		if ( ! $should_sync ) {
+			return false;
+		}
+
+		if ( ! $product instanceof WC_Product ) {
+			return (bool) $should_sync;
+		}
```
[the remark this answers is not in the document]
wjrosa: Good catch, fixed in 70bb2017e. The vote is now normalized with `wp_validate_boolean()` before branching, and the non-product test covers mixed `'false'`/`'0'` inputs via a data provider.

malithsen (approved):
This looks good to me and works as described

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
```diff
@@ -0,0 +1,117 @@
...
+ * Per-product Agentic Commerce sync exclusion flag.
+ *
+ * Owns storage and retrieval of the per-product exclude flag, and votes in the
+ * `wc_stripe_agentic_commerce_should_sync_product` filter so the feed honors it
+ * in every context (cron/CLI included). Kept separate from the admin meta box so
+ * the flag's read/write contract stays decoupled from the editor UI.
+ *
+ * @internal Not part of the plugin's public API; may change without notice.
+ * @package WooCommerce_Stripe
+ * @since 10.8.0
```
daledupreez: Two nits:
* The since tag should be updated.
* The context about the UI and all contexts is mostly relevant on this branch, but can be assumed in general.
```suggestion
 * Owns storage and retrieval of the per-product exclude flag, and ensures the exclusion flag
 * is applied through the `wc_stripe_agentic_commerce_should_sync_product` filter.
 *
 * @internal Not part of the plugin's public API; may change without notice.
 * @package WooCommerce_Stripe
 * @since 10.9.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
```diff
@@ -0,0 +1,117 @@
...
+		// Normalize before branching: a raw truthiness check would treat a prior
+		// callback's 'false'/'0' string vote as true and resurrect the product.
+		$should_sync = wp_validate_boolean( $should_sync );
+		if ( ! $should_sync ) {
+			return false;
+		}
+
+		if ( ! $product instanceof WC_Product ) {
+			return true;
+		}
```
daledupreez, line 126: Why are we ignoring a non-product instance? If we don't have a valid `WC_Product` instance, should we not exclude it by default?
```suggestion
		if ( ! $product instanceof WC_Product ) {
			return false;
		}
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
```diff
@@ -0,0 +1,117 @@
...
+	 * @since 10.8.0
+	 * @param int $product_id Product post ID.
+	 * @return bool
+	 */
+	public static function is_excluded( int $product_id ): bool {
+		if ( $product_id <= 0 ) {
+			return false;
+		}
+
+		$product = wc_get_product( $product_id );
```
daledupreez: I have a concern that we are (re)calling `wc_get_product()` here, which we would be doing for every product in the product catalog. Given that we already have the product instance in `filter_should_sync_product()`, this feels like it's unnecessary work. Can we not loosen the API to allow either a `WC_Product` instance or an integer?

At a higher level, it feels like we should aim for logic more like my proposed changes in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5513, where we're applying the exclusions as part of the primary filter for identifying products as part of the main database query logic. With the current per-product approach, we may trigger a query to fetch post meta for every product, which could apply unnecessary query load on busy sites.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,130 @@
...
+	 * @return void
+	 */
+	public function render_checkbox(): void {
+		if ( ! function_exists( 'woocommerce_wp_checkbox' ) ) {
+			return;
+		}
+
+		if ( ! WC_Stripe_Agentic_Commerce_Integration::is_merchant_enabled() ) {
+			return;
+		}
```
daledupreez, line 77: Might it be worth running this from `init()` so we don't hook in when we know Agentic Commerce is not enabled?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,130 @@
...
+
+		global $product_object;
+		$type = $product_object instanceof WC_Product ? $product_object->get_type() : '';
+		if ( ! in_array( $type, self::SUPPORTED_TYPES, true ) ) {
+			return;
+		}
+
+		woocommerce_wp_checkbox(
+			[
+				'id'          => WC_Stripe_Agentic_Commerce_Product_Exclusion::META_KEY,
```
daledupreez: Not blocking: is it worth exposing this value via a getter method rather than a constant?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,130 @@
...
+ * @since 10.8.0
+ */
+class WC_Stripe_Agentic_Commerce_Product_Meta_Box {
+
+	/**
+	 * Product types the toggle renders for, and the only types `save_meta()`
+	 * persists. The feed syncs `simple` and `variation`; a variation's parent is
+	 * `variable`, which is where the checkbox lives.
+	 */
+	private const SUPPORTED_TYPES = [ ProductType::SIMPLE, ProductType::VARIABLE ];
```
daledupreez, line 37: Not something we need to tackle in this PR, but we should consider how best to centralise the eligible product type logic, both for queries and for this product editor logic. It will be _really_ easy for the two to drift apart as things are written.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
```diff
@@ -0,0 +1,130 @@
...
+			return;
+		}
+
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing -- Nonce verified above.
+		$posted_value = isset( $_POST[ WC_Stripe_Agentic_Commerce_Product_Exclusion::META_KEY ] ) ? sanitize_text_field( wp_unslash( $_POST[ WC_Stripe_Agentic_Commerce_Product_Exclusion::META_KEY ] ) ) : '';
+		$excluded     = 'yes' === $posted_value;
+
+		$changed = WC_Stripe_Agentic_Commerce_Product_Exclusion::set_excluded( $product_id, $excluded );
+
+		// Converge Stripe now instead of waiting for the next scheduled sync.
```
daledupreez: I think we can use simpler language -- it took me two reads to understand this!
```suggestion
		// Request immediate sync with Stripe instead of waiting for the next scheduled sync.
```

includes/class-wc-stripe.php
```diff
@@ -1026,6 +1026,19 @@ public function initialize_agentic_commerce() {
 			WP_CLI::add_command( 'stripe agentic-commerce', 'WC_Stripe_Agentic_Commerce_CLI' );
 		}
 
+		// Per-product exclude toggle. Skipped when WC AI Storefront is active — it
+		// owns product selection through the same should-sync filter. The exclusion
+		// storage registers the feed filter in every context; the meta box is the
+		// admin-only editor UI on top of it.
+		if ( ! defined( 'WC_AI_STOREFRONT_VERSION' ) ) {
```
daledupreez: I am not sure we should add this exclusion -- I don't know if that plugin is or will be released to the .org plugin directory in its current shape.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
wjrosa: I'd keep `return true` here. The only place this filter runs from already requires a `\WC_Product`, so a non-product means someone called it incorrectly. At this point, `$should_sync` is already true (the guard above handles the false case), and this callback's only job is to check the exclude flag, which it can't read without a product, so it shouldn't be the one to say no. It also lines up with `is_excluded()`, which treats a non-product as "not excluded."

And the two risks aren't equal: returning `false` would quietly drop a real product from the feed, while returning `true` can't accidentally sync anything — the actual sync downstream still needs a real `WC_Product`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
wjrosa: Updated in 253215bd6 — `@since` is now 10.9.0 across the class (the inner tags were stale at 10.8.0), and the docblock is trimmed to your wording.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
wjrosa: Good call — loosened `is_excluded()` to accept a `WC_Product|int`, so `filter_should_sync_product()` passes the instance it already holds instead of re-fetching it per product (253215bd6). Agreed the query-args approach in #5513 is the better long-term shape for avoiding per-product meta lookups entirely; kept this PR to the local win.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: Done in 253215bd6 — moved the `is_merchant_enabled()` gate into `init()`, so the editor hooks aren't registered at all when Agentic Commerce is off. Kept the per-callback checks as defensive since both are public callbacks.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: Done — `META_KEY` is `protected` now with a `get_meta_key()` getter, matching the constant convention (253215bd6).

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: Agreed — the render-time type gate and the feed query can easily drift. Left it out of this PR as you suggested; worth a dedicated follow-up to centralize the eligible-type list so both read from one source.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-meta-box.php
wjrosa: Applied — the comment now reads "Request immediate sync with Stripe instead of waiting for the next scheduled sync."

includes/class-wc-stripe.php
wjrosa: Removed the guard in 253215bd6 — agreed that keying off `WC_AI_STOREFRONT_VERSION` (an unreleased plugin's constant) is fragile. Our exclusion now always registers; both plugins voting on the same `wc_stripe_agentic_commerce_should_sync_product` filter is harmless, so there's no need to suppress ours.

wjrosa:
@daledupreez @malithsen — base-branch change worth a re-look:

This PR is now targeted at #5598 instead of `develop`. #5598 adds a WooCommerce-core-prefixed `woocommerce_agentic_commerce_should_sync_product` filter and deprecates the Stripe-prefixed `wc_stripe_agentic_commerce_should_sync_product`, so the hook can be shared with other (non-Stripe) plugins.

What changed in this PR:
- `WC_Stripe_Agentic_Commerce_Product_Exclusion` now hooks the new `woocommerce_`-prefixed filter instead of the deprecated one.
- The filter definition/deprecation lives in #5598, so it no longer appears in this diff — which should clear up @malithsen's earlier note about #5458 changes showing up here after the rebase.

No behavior change to the toggle itself.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -214,22 +210,18 @@ public function schedule_full_resync_now(): void {
 			return;
 		}
 
-		if ( as_has_scheduled_action( self::SCHEDULED_ACTION, [], self::ASYNC_RESYNC_GROUP ) ) {
+		if ( as_has_scheduled_action( self::IMMEDIATE_SYNC_ACTION ) ) {
```
daledupreez: We should use the same group argument in the query as we add below.
```suggestion
		if ( as_has_scheduled_action( self::IMMEDIATE_SYNC_ACTION, [], 'wc-stripe' ) ) {
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
```diff
@@ -0,0 +1,132 @@
...
+	 * @since 10.9.0
+	 * @param mixed $should_sync Whether to include the product. Hook input — coerced to bool.
+	 * @param mixed $product     Product being evaluated. Hook input — validated below.
+	 * @return bool
+	 */
+	public function filter_should_sync_product( $should_sync, $product ): bool {
+		// Normalize before branching: a raw truthiness check would treat a prior
+		// callback's 'false'/'0' string vote as true and resurrect the product.
+		$should_sync = wp_validate_boolean( $should_sync );
+		if ( ! $should_sync ) {
```
daledupreez: Nit: It feels like we should use better baseline PHPDoc for the function's expected inputs.
```suggestion
	 * @param bool       $should_sync Whether to include the product.
	 * @param WC_Product $product     Product being evaluated.
	 * @return bool
	 */
	public function filter_should_sync_product( $should_sync, $product ): bool {
		// Use broad check for $should_sync to handle other filter outputs.
		$should_sync = wp_validate_boolean( $should_sync );
		if ( ! $should_sync ) {
```

daledupreez (commented):
This is looking OK based on inspection, though I am wondering whether it's worth pulling any of the changes from #5513 into this branch, or at least mapping out a path from the current code to the new code.

wjrosa:
Thanks for the additional review, @daledupreez!

> This is looking OK based on inspection, though I am wondering whether it's worth pulling any of the changes from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5513 into this branch, or at least mapping out a path from the current code to the new code.

I actually think they are complementary, and it makes sense to keep both. There isn't really a path that replaces the current code; it's additive: keep the filter, add the `meta_query` on top, sharing `get_meta_key()`, and either solve the variable-parent query or leave that case to the filter.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -45,19 +45,14 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
...
-	 * so a shared group would make the one-off a no-op whenever the cron
-	 * tick is queued.
+	 * Action Scheduler hook for an immediate, one-off full resync. Kept distinct
+	 * from {@see self::SCHEDULED_ACTION} — the recurring action is always pending,
+	 * so sharing it would make `as_has_scheduled_action()` drop every immediate resync.
 	 *
 	 * @var string
 	 * @since 10.8.0
 	 */
-	private const ASYNC_RESYNC_GROUP = 'wc-stripe-agentic-resync';
```
Mayisha, line 60: AI feedback:

`test_reschedule_next_feed_sync_schedules_recurring_action_when_async_resync_is_pending` green-passes without testing anything — it queues into the dead wc-stripe-agentic-resync group that this very PR deleted.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
```diff
@@ -0,0 +1,131 @@
...
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Stores the per-product exclude flag and filters Agentic Commerce sync eligibility.
+ *
+ * @internal
+ * @since 10.8.0
```
Mayisha: Nit: Remember to update this and all other `@since` tags before finally merging.

Mayisha (approved):
I can see the per product exclude toggle in eligible products when agentice is enabled. Toggling the checkbox immediately schedules a feed sync. Feed preview also renders the info correctly 👍

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Good catch — the test was still staging its one-off action in the removed `wc-stripe-agentic-resync` group, so the pending action could never collide with the recurring-schedule check and the test passed vacuously. Reworked in eeebb98d3: it now queues the resync through the real `schedule_full_resync_now()` path (`IMMEDIATE_SYNC_ACTION` in the `wc-stripe` group) and, since both actions now share a group, also asserts the pending resync survives `reschedule_next_feed_sync()` — which only unschedules the recurring hook.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-exclusion.php
wjrosa: Audited every `@since` this PR adds: this class-level `10.8.0` was the only stray (fixed to `10.9.0` in eeebb98d3).

PR #5477, opened 2026-05-25
Preserve wallet_type on saved tokens; widen wallet brand label regex

wjrosa:
Follow-up to #5439 addressing two issues @mayisha caught post-merge.

**1. Stop refreshing `wallet_type` on the duplicate-match token paths.**

`wallet_type` was being overwritten in two spots:
- `WC_Stripe_UPE_Payment_Gateway::handle_saving_payment_method` (the order-save path)
- `WC_Stripe_Payment_Tokens::add_token_to_user` (the token-sync path)

Repro: save a card directly in My Account (no wallet), then pay a subscription via Google Pay using the same card. Stripe returns a new PM with `wallet.type = google_pay`; the duplicate-match path matched by fingerprint and re-saved the existing token with the new wallet type, re-badging the shopper's saved Visa as "Google Pay (Visa)".

Treating `wallet_type` as create-time state and not refreshing it on subsequent uses. Other mutable card metadata (expiry, last4, card_type) still refreshes — only `wallet_type` is now pinned. Trade-off: legacy tokens that pre-date the `wallet_type` field don't get backfilled by a later wallet payment. The alternative (silently flipping the shopper's saved card to a wallet brand) is the worse failure mode for what's essentially a display field.

**2. Widen `normalize_payment_method_label`'s wallet brand regex to handle multi-word brands.**

The repair regex was `^(Apple Pay|Google Pay) \(([a-z][a-z ]*)\)$`. WC core's `wc_get_credit_card_type_label` runs `ucwords` on our wrapped value, and for multi-word brands the inner string comes out half-capitalized — e.g. `"Apple Pay (Cartes Bancaires)"` becomes `"Apple Pay (cartes Bancaires)"` because `(` isn't a word boundary (first letter stays lowercase) but the space before the second word is (so `Bancaires` gets re-capitalized). The original lowercase-only regex rejected the capital B and the repair never fired, leaving the mangled label on the page.

Widened the inner group to `[^)]+` and let `wc_get_credit_card_type_label` handle the case normalization — it does its own `strtolower` + label lookup. Single-word brands still work; `cartes_bancaires`, `american_express`, `unionpay` now do too.

## Testing instructions

**Fix 1 (`wallet_type` preservation):**
1. In a Stripe test account, save card `4000056655665556` in My Account → Payment Methods.
2. Confirm the saved card shows as a plain Visa (no wallet branding).
3. Purchase a subscription via Google Pay, selecting the same card from the Google Pay sheet.
4. Reload My Account → Payment Methods.
5. **Expected:** the saved card still shows as plain Visa. Without this fix, it flips to "Google Pay (Visa)".

**Fix 2 (regex for multi-word brands):**
This is hard to reproduce manually without a `cartes_bancaires`-branded test card. The new unit-test data provider rows feed `normalize_payment_method_label` the exact post-`ucwords` string the filter sees in production (`"Apple Pay (cartes Bancaires)"`) and assert it's re-labelled to `"Apple Pay (Cartes Bancaires)"`. Run `npm run test:php -- --filter WC_Stripe_Payment_Tokens_Test`.

---

- [x] Tested on mobile (or does not apply)

### Changelog entry

<details>

#### Comment

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.8.0
-* Fix - Address wallet branding edge cases on saved cards: stop re-badging a card when it's later used via Apple Pay or Google Pay, and render multi-word card brands (e.g. Cartes Bancaires) correctly
+* Fix - Preserve saved card branding when the same card is later used via a wallet, and render multi-word brands (e.g. Cartes Bancaires) correctly
 * Add - Show Apple Pay / Google Pay branding on saved card tokens in My Account → Payment Methods and at checkout
 * Add - Add a setting to control whether Express Checkout is shown on the WooCommerce Subscriptions change payment method page
 * Add - Allow shoppers to change a subscription payment method using Express Checkout (Apple Pay, Google Pay, Link)
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.8.0
+* Fix - Address wallet branding edge cases on saved cards: stop re-badging a card when it's later used via Apple Pay or Google Pay, and render multi-word card brands (e.g. Cartes Bancaires) correctly
 * Add - Show Apple Pay / Google Pay branding on saved card tokens in My Account → Payment Methods and at checkout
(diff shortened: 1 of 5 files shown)


Mayisha (approved):
Thanks for the fix @wjrosa.

Issue #5362, opened 2026-04-27
Deferred webhook handler silently abandons captured Stripe charge when order completed by another gateway during deferral window


mkalina:
### Description

When a customer initiates a Stripe (UPE / Stripe Link) PaymentIntent that requires async authentication (e.g., 3DS), abandons the flow, and successfully completes checkout on the same order via a different gateway (in our case Klarna Payments) within the \~2-minute window before the deferred webhook handler runs, the customer is charged twice: once by Stripe (intent succeeds asynchronously) and once by the second gateway. The plugin's deferred webhook handler does detect the conflict but silently bails without refunding or voiding the now-orphaned Stripe charge, leaving the merchant to discover the duplicate manually.

**Verified against current trunk:** the silent-bail code path at `class-wc-stripe-webhook-handler.php` lines 1334-1337 is unchanged in trunk as of the date of filing. The 10.6.0 fixes (`#5190` for the deferred-webhook TypeError, idempotency keys for intent creation) do not address the scenario described here. Reproducible on any version from 10.5.x through current trunk.

### Steps to reproduce

1. Customer adds a product to the cart and proceeds to checkout.
2. Customer selects a Stripe payment method that triggers `requires_action` (e.g., a card requiring 3DS, or Stripe Link with 3DS step-up). The PaymentIntent is created, `_stripe_upe_waiting_for_redirect` is set on the order, and the deferred webhook handler is scheduled via Action Scheduler with a `time() + 2 * MINUTE_IN_SECONDS` delay.
3. Customer dismisses the Stripe modal / closes the 3DS challenge before completing it on the merchant side, but the underlying authentication completes asynchronously (typical for Stripe Link, where authentication can finalise via push/email out-of-band).
4. Within the next \~2 minutes, customer goes back to checkout and pays the same order using a non-Stripe gateway (e.g., Klarna).
5. Both payments succeed:
   * Stripe webhook `payment_intent.succeeded` arrives \~30s after intent creation; the charge is captured automatically.
   * The other gateway captures shortly after.
6. The deferred webhook job fires after the 2-minute delay, sees the order is no longer `pending`/`failed`, logs `"Skipped processing deferred webhook for Stripe PaymentIntent X for order Y - payment already complete"`, and exits without acting on the captured Stripe charge.

### Expected behavior

When the deferred handler detects that the order has been completed by a different payment method, it should not silently bail. Either:

[...]

Issue #5268, opened 2026-04-12
payment_complete() race condition with Store API: payment meta persists but status transition lost, order auto-cancelled


teknovision:
## Description

A race condition between the Store API checkout handler and the Stripe webhook causes `payment_complete()` to be called concurrently on the same order. The payment metadata (`_date_paid`, `_stripe_charge_captured`) is written successfully, but the status transition from `pending` to `processing` is lost. The order remains at `pending` despite being fully paid.

Because WooCommerce's `wc_cancel_unpaid_orders()` runs on a schedule (controlled by `woocommerce_hold_stock_minutes`), the paid order is then auto-cancelled as "unpaid" since its status is still `pending`.

This appears related to #2536 (acknowledged as a stop-gap fix needing deeper refactoring) and #3154 (paid orders stuck at pending). The underlying race between the return-URL handler and the webhook handler may not be fully resolved for the Store API checkout path.

## Environment

- WordPress: 6.8
- WooCommerce: 10.6.2
- WooCommerce Stripe Gateway: 10.5.3
- Checkout method: Store API (block checkout)
- PHP: 8.2
- HPOS: disabled (legacy post storage)

## Steps to reproduce

Difficult to reproduce on demand as this is a timing-dependent race condition. Observed once in production:

1. Customer completes checkout via Store API
2. Store API handler calls `payment_complete()` on the order
3. Stripe webhook (`charge.succeeded` or `payment_intent.succeeded`) arrives and also calls `payment_complete()` on the same order
4. Both calls write payment meta (`_date_paid`, `_stripe_charge_captured = yes`)
5. Only one status transition succeeds; the other is silently lost
6. Order remains at `pending` with valid payment metadata
7. After `hold_stock_minutes` elapses (60 min in our case), `wc_cancel_unpaid_orders()` cancels the order

## Observed state (post-incident)

```
post_status:              pending (later auto-cancelled)
_date_paid:               1775940497 (set - payment was captured)
_stripe_charge_captured:  yes
_paid_date:               2026-04-11 22:48:17
_transaction_id:          (empty)
```

Payment was captured by Stripe. The charge exists on the Stripe dashboard. But WooCommerce treated the order as unpaid because the status never moved from `pending`.

## Expected behavior

If `payment_complete()` is called and payment meta is written, the status transition to `processing` should be guaranteed. At minimum, `wc_cancel_unpaid_orders()` should check `_date_paid` before cancelling an order that has confirmed payment.

[...]

PR #5479, opened 2026-05-25
Recover missing Stripe charge ID so affected card orders stay refundable

### LINKED ISSUE #5478: Refunds permanently fail on card orders missing the stored charge ID (`_transaction_id`)
### Summary

On some **card** orders, refunds fail permanently because the Stripe charge ID (`_transaction_id`) is missing from the order, even though the payment succeeded.

* **Refunding from wp-admin** returns: *"An error occurred while attempting to create the refund using the payment gateway API."*
* **Refunds initiated from the Stripe Dashboard** do not sync back to WooCommerce.
* The webhook log shows:

  ```
  Could not find order via refund ID (re_xxx) or charge ID (ch_xxx)
  ```

Affected orders still have their payment intent recorded (`_stripe_intent_id` and the intent order note are present) and no checkout-failure note — only the charge ID is missing.

### Root cause

For card (synchronous) payments, the charge ID is written to the order **only** during the synchronous checkout request (`process_response()` → `payment_complete()`). There is no webhook that back-fills it afterwards:

* `WC_Stripe_Webhook_Handler::process_webhook_charge_succeeded()` returns early for synchronous methods, including `card`.
* `WC_Stripe_Webhook_Handler::process_payment_intent()` (handling `payment_intent.succeeded`) only acts on orders in `pending`/`failed` status; an already-paid card order is `processing`/`completed`, so it no-ops.

If that single synchronous write is lost or overwritten (e.g. a concurrent/stale order save during Store API + HPOS checkout — see woocommerce/woocommerce-gateway-stripe#5268 / woocommerce/woocommerce-gateway-stripe#5362), the order is left paid but with no charge ID, and:

* `process_refund()` returns `false` immediately because `$order->get_transaction_id()` is empty (it bails before ever consulting the stored intent), and
* `process_webhook_refund()` cannot match the order by refund ID or charge ID.

The payment intent is saved *earlier* in checkout (with its own `save()`), which is why the intent survives but the charge ID does not.

### Impact

Card orders that hit this become permanently un-refundable from both wp-admin and the Stripe Dashboard. Observed on a high-volume, card-only store running HPOS (sync off); intermittent (not every order).

### Proposed fix

1. `process_refund()`: when `_transaction_id` is missing, recover the charge ID from the order's stored payment intent, reconcile the captured flag, and persist it back so the order self-heals.
2. `process_webhook_refund()`: add a payment-intent-ID fallback to locate the order when the refund/charge lookups miss, back-filling the charge from the notification so Dashboard refunds sync.

This restores refunds on already-affected orders without requiring a re-charge. The underlying lost-write during checkout is tracked separately (woocommerce/woocommerce-gateway-stripe#5268 / woocommerce/woocommerce-gateway-stripe#5362).

wjrosa:
Fixes #5478
Fixes STRIPE-1099

Fixes refunds permanently failing on **card** orders that are missing the stored Stripe charge ID (`_transaction_id`), even though the payment succeeded.

For card (synchronous) payments the charge ID is written to the order **only** during the synchronous checkout request — no webhook back-fills it afterwards (`process_webhook_charge_succeeded()` returns early for synchronous methods, and `process_payment_intent()` only acts on `pending`/`failed` orders). If that single write is lost or overwritten (e.g. a concurrent/stale order save under Store API + HPOS — see #5268 / #5362), the order is left paid but un-refundable:

- `process_refund()` returns `false` because `get_transaction_id()` is empty (it bails before consulting the stored intent), producing *"An error occurred while attempting to create the refund using the payment gateway API."*
- `process_webhook_refund()` can't match the order by refund ID or charge ID, so Dashboard-initiated refunds don't sync back.

The payment intent is saved earlier in checkout, which is why affected orders still have `_stripe_intent_id` (and the intent note) but no charge ID.

This PR:

1. **`process_refund()`** — when the charge ID is missing, recover it from the order's stored payment intent, reconcile the captured flag, and persist it back so the order self-heals.
2. **`process_webhook_refund()`** — add a payment-intent-ID fallback to locate the order and back-fill the charge from the notification, so Dashboard refunds sync.

This restores refunds on already-affected orders without requiring a re-charge. The underlying lost-write during checkout is tracked separately (#5268 / #5362).

### Testing instructions

1. Create a card order, then remove `_transaction_id` from it while keeping `_stripe_intent_id` (simulating the lost write); ensure the order is `processing`/`completed`.
2. Attempt a refund from **wp-admin** → it now succeeds, and the order's charge ID is back-filled (an order note records the recovery).
3. Issue a refund from the **Stripe Dashboard** for an affected order → the refund now syncs to WooCommerce and the charge ID is back-filled.
4. Confirm normal orders (charge ID present) refund exactly as before.

Automated coverage:

- `tests/phpunit/class-wc-stripe-payment-gateway-test.php` — `test_process_refund_recovers_missing_charge_id_from_intent`, `test_process_refund_returns_false_when_charge_id_unrecoverable`
- `tests/phpunit/class-wc-stripe-webhook-handler-test.php` — `test_process_webhook_refund_finds_order_via_intent_when_charge_id_missing`

- [x] Covered with tests (PHPUnit; see note on local execution)
- [ ] Changelog and readme entries — added under 10.7.0


includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -1210,13 +1210,23 @@ public function process_refund( $order_id, $amount = null, $reason = '' ) {
...
-		$captured       = $order_helper->is_stripe_charge_captured( $order );
 		$charge_id      = $order->get_transaction_id();
 
+		// Some orders end up paid but without a stored charge ID (for example, a lost write
+		// during checkout). For card payments no webhook back-fills it afterwards, leaving the
+		// order permanently un-refundable. If the payment intent is still stored on the order,
+		// recover the charge ID from it so the order becomes refundable again.
+		if ( ! $charge_id && $order instanceof WC_Order ) {
+			$charge_id = $this->recover_charge_id_from_intent( $order );
+		}
```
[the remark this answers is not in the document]
wjrosa: Fixed in 0596afaa8 — wrapped the charge fetch in try/catch (logs and returns ``) and added an `is_object($charge)` guard.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -746,6 +746,26 @@ public function process_webhook_refund( $notification ) {
...
+
+			if ( $order instanceof WC_Order && ! $order->get_transaction_id() ) {
+				$order->set_transaction_id( $notification->data->object->id );
+
+				if ( isset( $notification->data->object->captured ) ) {
+					WC_Stripe_Order_Helper::get_instance()->set_stripe_charge_captured( $order, (bool) $notification->data->object->captured );
+				}
+
+				$order->save();
+			}
```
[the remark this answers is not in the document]
wjrosa: Fixed in 0596afaa8 — the intent-fallback back-fill is now gated on `is_stripe_gateway_order()`, and `$order_helper` is reused.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -1346,6 +1353,58 @@ public function process_refund( $order_id, $amount = null, $reason = '' ) {
 		}
 	}
 
+	/**
+	 * Recovers a missing charge ID from the order's stored payment intent and persists it.
+	 *
+	 * @param WC_Order $order The order to recover the charge ID for.
+	 * @return string The recovered charge ID, or an empty string if it could not be recovered.
+	 */
+	private function recover_charge_id_from_intent( WC_Order $order ) {
```
Mayisha: Nit
```suggestion
	private function recover_charge_id_from_intent( WC_Order $order ): string {
```

tests/phpunit/class-wc-stripe-payment-gateway-test.php
```diff
@@ -593,6 +610,78 @@ public function test_process_refund_success() {
...
+		$reloaded = wc_get_order( $order_id );
+		$this->assertSame( 'ch_123', $reloaded->get_transaction_id() );
+
+		remove_filter( 'pre_http_request', $callback );
+	}
+
+	/**
+	 * Tests that process_refund returns false when the charge ID can't be recovered.
+	 */
+	public function test_process_refund_returns_false_when_charge_id_unrecoverable() {
```
Mayisha, line 675: AI feedback: test_process_refund_returns_false_when_charge_id_unrecoverable only covers the no-intent case. It doesn't exercise the path where the
  intent exists but get_latest_charge_from_intent() throws → returns '' → process_refund returns false.

Mayisha (approved):
Works as described 👍 

Image

diegocurbelo (approved):
The code changes look good, and it tests well!

> Create a card order, then remove _transaction_id from it while keeping _stripe_intent_id (simulating the lost write); ensure the order is processing/completed.

The PR description mentions HPOS, so I nulled the `transaction_id` in the `wc_orders` table instead, and after the successful refund the transaction_id column was updated with the charge_id.

tests/phpunit/class-wc-stripe-payment-gateway-test.php
wjrosa: Good catch — added `test_process_refund_returns_false_when_intent_charge_lookup_throws` in cd130f4f. It seeds a payment intent, has `payment_intents/pi_456` return successfully but `charges/ch_456` return a Stripe error, so `get_latest_charge_from_intent` throws, the recover path catches and returns '', and `process_refund` falls through to false. Also asserts the transaction ID stays empty.

Issue #270, opened 2017-05-30
Refunds arent beign processed because of the transaction ID


Lunatikzx:
when we are trying to refund our customers on stripe the automated refund is not working. 
I found the cause, this is due to the transaction id not being inserted in the transaction, 

The plugin should process the refund automatically.

An alert box apperread with the error in the screenshot
![capture d ecran 2017-05-29 a 09 34 25](https://cloud.githubusercontent.com/assets/1518503/26584731/d61620ee-4517-11e7-8273-fbd0d6dc5b6a.png)

- Buy a product
- Pay with stripe
- Try to refund it automatically

Personnal note:
The bug is linked with the transaction id that does not get added to the order when the order is paid with stripe. 

So when in the process_refund method, the get_transaction id will never work.

-------------------

- [ ] Issue assigned to next milestone.
- [ ] Issue assigned a priority (will be assessed by maintainers).

roykho:
Already fixed in master and will be released.

Issue #269, opened 2017-05-26
Can't refund


WillBrubaker:
#### Affected ticket(s)
ticket: 574646-zd-woothemes

#### What I expected
To be able to issue a refund

#### What happened instead
When attempting to issue a refund, get a message similar to "An error occurred while attempting to create the refund using the payment gateway API."

#### Steps to reproduce the issue

1. Stripe is set to authorize instead of capture
2. Go through the checkout process
3. Order is set to on-hold as expected
4. Change the order to processing
5. Initiate a refund, see message as noted above because the transaction id wasn't set when capturing the charge and so process_refund returns false here https://github.com/woocommerce/woocommerce-gateway-stripe/blob/3.1.9/includes/class-wc-gateway-stripe.php#L937

@mattyza 

-------------------

- [ ] Issue assigned to next milestone.
- [ x] Issue assigned a priority (will be assessed by maintainers).

room34:
It looks to me like this change will solve the problem on any _future_ orders placed after this code is in place, but it won't resolve the issue for earlier orders. I suppose there's not a way to obtain the `_transaction_id` value after the fact, so maybe there's nothing else that can be done.

roykho:
This only affected authorized first capture later setups. Also yes you can still get the ID, it is under the meta name `Stripe Payment ID`.

room34:
My client's site is set to capture later. They're running into this same error message when trying to process a refund. The orders in question have `Stripe Payment ID` meta data, but do not have `_transaction_id`. My understanding of this ticket is that it's describing the same scenario I'm seeing, and without `_transaction_id` the refund can't be processed. Is that incorrect?

roykho:
That is correct. But again you can backfill the `_transaction_id` with the value found in the `Stripe Payment ID` and you will have a working auto refund. Note that you can still manually refund even without any id, it just becomes a two step process instead of one.

room34:
Thanks. I will manually set that value and see if it resolves things. Still I'm wondering if there couldn't be a way to automate filling in `_transaction_id` if it's blank and `Stripe Payment ID` isn't, so this fix would work retroactively. (I started poking around the code for potential hooks but didn't see anything.)

jessepearson:
@roykho @room34 From what I understand in these last comments, even if a store admin installs this new version, they are not able to refund older orders, correct?

PR #5481, opened 2026-05-25
Store charge ID as transaction ID for express checkout (confirmation token) orders

### LINKED ISSUE #5480: Express checkout (confirmation token) orders don't store the charge ID as `_transaction_id`, breaking refunds
### Summary

Refunds from WooCommerce admin fail with *"An error occurred while attempting to create the refund using the payment gateway API"* for orders paid via **express checkout (Apple Pay / Google Pay)**, which use the confirmation token flow (`process_payment_with_confirmation_token`).

The charge ID is shown in the order notes (e.g. "Stripe charge complete (Charge ID: ch\_...)") and the PaymentIntent succeeded in Stripe, but WooCommerce's native `_transaction_id` is empty:

```
_stripe_charge_captured = yes
_stripe_intent_id       = pi_...
_stripe_upe_payment_type = card
_transaction_id         = (empty)
```

`process_refund()` returns early when `get_transaction_id()` is empty, so no refund request is ever sent to Stripe.

### Root cause

In `process_payment_with_confirmation_token()` the charge ID is persisted only through `process_response()`, which calls WooCommerce core's `payment_complete( $charge_id )`. Core only calls `set_transaction_id()` when the order is in a `PAYMENT_COMPLETE_STATUSES` status (pending/on-hold/failed/cancelled/checkout-draft):

```php
// WC_Order::payment_complete()
if ( $this->has_status( $valid_completed_statuses ) ) {
    if ( ! empty( $transaction_id ) ) {
        $this->set_transaction_id( $transaction_id );
    }
    ...
}
```

In the express checkout flow the order is frequently already in a paid status (`processing`/`completed`) by the time `process_response()` runs, so `set_transaction_id()` is skipped — while `process_response()` still adds the "Stripe charge complete" note, which is why the charge ID appears in the notes but not in `_transaction_id`.

This is the same class of issue previously fixed for the legacy flow ([#269](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/269>), [#270](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/270>)) but not for the confirmation token flow.

### Expected behavior

`set_transaction_id()` should be called with the charge ID during the confirmation token payment flow, regardless of the order's current status.

### Proposed fix

After `process_response()` in `process_payment_with_confirmation_token()`, persist the charge ID explicitly when it isn't already set.

### Related

Refund-side recovery for orders that are already missing the charge ID is handled separately in woocommerce/woocommerce-gateway-stripe#5478 / woocommerce/woocommerce-gateway-stripe#5479.

wjrosa:
Fixes #5480

Fixes refunds failing for orders whose charge ID was never persisted to `_transaction_id` at checkout time.

The charge ID is persisted through `process_response()` → WooCommerce core `payment_complete( $charge_id )`. Core only calls `set_transaction_id()` when the order is in a `PAYMENT_COMPLETE_STATUSES` status (pending/on-hold/failed/cancelled/checkout-draft). When the order is already in a paid status (`processing`/`completed`) by the time `process_response()` runs, `set_transaction_id()` is skipped — yet the "Stripe charge complete (Charge ID: …)" note is still added. The order ends up with the charge ID in its notes but an empty `_transaction_id`, so `process_refund()` returns early and refunds fail with *"An error occurred while attempting to create the refund using the payment gateway API."*

This most commonly surfaces with **express checkout (Apple Pay / Google Pay)** orders, which are frequently already paid by the time `process_response()` runs, but it can affect any flow — for example a race condition between checkout and the webhook.

The fix lives in `process_response()` (the shared completion path), so it backfills the charge ID for **any flow that completes a capture**, not just express checkout: after `payment_complete()`, if `_transaction_id` isn't already set, we persist the charge ID explicitly. Same class of fix as #269 / #270 for the legacy flow.

> [!NOTE]
> The refund-side recovery for orders that are *already* missing the charge ID is handled separately in #5479 (issue #5478). This PR fixes the root cause so the ID is stored at checkout time.

### Testing instructions

1. Pay for an order with Apple Pay or Google Pay (express checkout).
2. Confirm the order's `_transaction_id` is set to the Stripe charge ID (`ch_...`).
3. Refund the order from WooCommerce admin → it succeeds and the refund appears in Stripe.

Automated coverage:

- `tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php` — `test_process_payment_with_confirmation_token_backfills_missing_transaction_id` (order already in a paid status; asserts `_transaction_id` is persisted).

- [x] Covered with tests
- [ ] Changelog and readme entries — added under 10.8.0


diegocurbelo (approved):
The code change looks good, and it tests well!

One thought though, did we consider putting this if block inside `process_response()`, right after `$order->payment_complete( $response->id )` (line ~661)? That would also cover non-confirmation-token paths.

daledupreez (approved):
The changes are also testing well for me, though I am interested to see your response to @diegocurbelo's question about a possible alternative location for setting the transaction ID.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -3944,6 +3944,69 @@ function ( $payment_information ) use ( $payment_method_id, $express_payment_met
...
+		// Before processing the order has no transaction ID.
+		$this->assertSame( '', $order->get_transaction_id() );
+
+		$response = $this->mock_gateway->process_payment( $order_id );
+
+		$this->assertEquals( 'success', $response['result'] );
+
+		$reloaded = wc_get_order( $order_id );
+		$this->assertSame( 'ch_mock1234567890', $reloaded->get_transaction_id() );
+	}
```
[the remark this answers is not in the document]
wjrosa: Done in 86df8caca — added `provide_paid_statuses_for_confirmation_token_backfill` covering both `processing` and `completed`. Both cases pass.

wjrosa:
@diegocurbelo good call — done in a7568e845. `payment_complete()` skips `set_transaction_id()` for *any* already-paid order, so I moved the backfill into `process_response()` (right after `payment_complete()`, guarded by `! $order->get_transaction_id()`). It now covers the non-CT express paths and checkout/webhook races, reuses the method's existing `save()`, and never overwrites an existing ID. The parameterized test still guards the original repro.

wjrosa:
Re-requesting review since the last commit (a7568e845) moved the backfill out of the confirmation-token flow into the shared `process_response()`, so it now runs on every charge path including webhooks. @diegocurbelo this is your suggested placement; @daledupreez flagging you since it's now in payment-critical shared code rather than the CT-only method you approved earlier. Guarded by `! $order->get_transaction_id()` so it only fills a genuine gap.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -660,6 +660,13 @@ public function process_response( $response, $order ) {
 				} else {
 					$order->payment_complete( $response->id );
 
+					// payment_complete() skips set_transaction_id() when the order is already in a
+					// paid status (e.g. express checkout orders, or a checkout/webhook race). Persist
+					// it explicitly so refunds, which rely on _transaction_id, work for these orders.
```
daledupreez: I think we can rephrase this slightly, as I needed to read it more than once to make sure I understood it.
```suggestion
					// $order->payment_complete() does not call $order->set_transaction_id() when the order
					// is already in a paid status. This can occur for express checkout orders, or in situations
					// where there is a race condition between checkout and the webhook.
					// We ensure we have a transaction ID so downstream actions like refunds will work.
```

readme.txt
```diff
@@ -152,6 +152,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.8.0 - xxxx-xx-xx =
+* Fix - Store the Stripe charge ID as the transaction ID for orders already marked paid when the charge completes (e.g. express checkout), so they remain refundable
```
daledupreez: I think we should tighten this description up -- the "e.g." feels a bit weird to me.
```suggestion
* Fix - Store transaction IDs for orders when we get charges to ensure we can refund correctly
```

daledupreez (approved):
I've tested the updated change and express checkout is still working and storing the transaction ID. I also think the check for an existing transaction ID means we should only fill in the transaction ID when it was missing, and this specific code is only running when the charge succeeded, so I think that's sufficiently guarded.

I still think it's worth getting @diegocurbelo to take another look as well -- I am honestly not clear on what edge cases we may want to look at more closely.

diegocurbelo (approved):
Thanks for the fix @wjrosa, it looks good to me 👍🏼 

We might want to update the PR description before merging, the fix now affects any flow calling `process_response()`:
- the fix lives in `process_response()` and therefore backfills the charge ID for any flow that completes a capture — including express checkout

wjrosa:
Thanks for the review, Diego! I have updated the description 👍

Issue #2536, opened 2023-01-28
UPE payments and webhooks have a race condition


reykjalin:
## Describe the bug

When processing payments made via UPE (the new checkout experience) we do some initial processing, [set the order status to `pending`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/cb47c373ee53a5378227b25c637bc2b455226308/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L576) and then redirect to the order received page. Once the customer is on the order received page we finalize the payment in [WC_Stripe_UPE_Payment_Gateway::process_order_for_confirmed_intent](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/cb47c373ee53a5378227b25c637bc2b455226308/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L889-L947).

Independently from this we receive webhooks from Stripe and the `payment_intent.success` webhook calls [WC_Stripe_Payment_Gateway::process_response](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/cb47c373ee53a5378227b25c637bc2b455226308/includes/abstracts/abstract-wc-stripe-payment-gateway.php#L504-L566) where we call [WC_Order::payment_complete](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/cb47c373ee53a5378227b25c637bc2b455226308/includes/abstracts/abstract-wc-stripe-payment-gateway.php#L535-L536) if the intent status is `succeeded`.

Usually this isn't a problem, but when the webhook is processed **before the customer loads the Order Received page** we never finish processing the payment in the UPE gateway because the order status is wrong; it's been set to `processing` by the call to `WC_Order::payment_complete` when processing the webhook.

This usually doesn't cause a problem since the payment is completed and the order is marked as such. However, when it's necessary to save a card -- such as in the case of subscriptions -- the payment method ID is never saved to the order, causing subscription renewals to fail. This also makes it so that the card is never saved to the customer.

- [ ] Patch the issue https://github.com/woocommerce/woocommerce-gateway-stripe/pull/2540.
- [ ] Introduce a proper fix, perhaps by refactoring the `UPE::process_payment` method, perhaps by refactoring the webhook processing.

## To Reproduce

> **Warning**
> You **must** be using UPE on your test store to reproduce the issue.

> **Warning**
> You **must** have webhooks set up correctly on your test store to reproduce the issue.

[...]

PR #5486, opened 2026-05-25
Don't auto-cancel Stripe orders that have already been paid

### LINKED ISSUE #5485: Paid Stripe orders auto-cancelled as unpaid when a checkout/webhook race leaves them stuck at pending
### Summary

A paid order can be auto-cancelled as "unpaid". A race between the Store API (block) checkout handler and the Stripe webhook causes `payment_complete()` to run concurrently on the same order: the payment meta (`_date_paid`, `_stripe_charge_captured`) is written, but the `pending → processing` status transition is lost (a stale order-object save reverts `post_status` to `pending` on legacy/CPT storage). The order then sits at `pending` despite being paid, and `wc_cancel_unpaid_orders()` cancels it after `hold_stock_minutes`.

Observed post-incident: `post_status: pending`, `_date_paid` set, `_stripe_charge_captured: yes`, `_transaction_id` empty. The charge exists in Stripe.

Reported in woocommerce/woocommerce-gateway-stripe#5268. Related: [#2536](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2536>), [#3154](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/3154>). Same concurrency family as woocommerce/woocommerce-gateway-stripe#5362.

### Scope of this fix

The underlying concurrent-save race needs a deeper fix (atomic locking around `payment_complete` + the post-checkout save), tracked under [#2536](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/2536>). This PR addresses the **customer-visible damage**: a paid order should never be cancelled as unpaid.

The plugin already hooks `woocommerce_cancel_unpaid_order` (`prevent_cancelling_orders_awaiting_action`), but only bails for orders awaiting action — not for orders that have already been paid.

### Proposed fix

In the `woocommerce_cancel_unpaid_order` guard, also bail when `$order->get_date_paid()` is set for Stripe orders. This prevents paid-but-pending orders from being cancelled regardless of the race.

wjrosa:
Fixes #5485
Fixes STRIPE-1076
Reported in #5268

Prevents paid Stripe orders from being auto-cancelled as "unpaid".

A race between the Store API (block) checkout handler and the Stripe webhook can call `payment_complete()` concurrently on the same order. The payment meta (`_date_paid`, `_stripe_charge_captured`) is written, but the `pending → processing` status transition is lost — on legacy/CPT storage a stale order-object save reverts `post_status` to `pending`. `wc_cancel_unpaid_orders()` then cancels the paid order after `hold_stock_minutes`.

The plugin already hooks `woocommerce_cancel_unpaid_order` via `prevent_cancelling_orders_awaiting_action()`, but it only bailed for orders *awaiting action*. This PR also bails when the order has a `date_paid` set (for Stripe orders), so a paid order is never cancelled as unpaid regardless of the race.

This addresses the customer-visible damage. The deeper fix for the concurrent-save race itself (atomic locking around `payment_complete` and the post-checkout save) is broader and tracked under #2536 / the #5362 lineage.

### Testing instructions

1. Create a Stripe order and put it in the paid-but-pending state (set `_date_paid`, keep status `pending`) — or reproduce via the checkout/webhook race.
2. Run `wc_cancel_unpaid_orders()` (or wait for the scheduled `woocommerce_cancel_unpaid_orders` action).
3. The order is **not** cancelled (previously it would be).
4. Confirm a genuinely unpaid Stripe order (no `date_paid`) past the hold time is still cancelled as before.

Automated coverage:

- `tests/phpunit/class-wc-stripe-order-handler-test.php` — `test_prevent_cancelling_orders_that_have_been_paid` (paid Stripe order → not cancelled, without fetching the intent); existing `test_prevent_cancelling_orders_awaiting_action` still passes.

- [x] Covered with tests
- [ ] Changelog and readme entries — added under 10.8.0

### DIFF
--- a/includes/class-wc-stripe-order-handler.php
+++ b/includes/class-wc-stripe-order-handler.php
@@ -474,8 +474,20 @@ class WC_Stripe_Order_Handler extends WC_Stripe_Payment_Gateway {
 
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
 
-		// Bail if payment method is not stripe or `stripe_{apm_method}` or doesn't have an intent yet.
-		if ( ! $order_helper->is_stripe_gateway_order( $order ) || ! $this->get_intent_from_order( $order ) ) {
+		// Bail if the payment method is not stripe or `stripe_{apm_method}`.
+		if ( ! $order_helper->is_stripe_gateway_order( $order ) ) {
+			return $cancel_order;
+		}
+
+		// Never cancel an order that has already been paid. A race between the Store API checkout
+		// and the Stripe webhook can leave a paid order stuck at `pending` (payment meta written,
+		// status transition lost), which wc_cancel_unpaid_orders() would otherwise cancel.
+		if ( $order->get_date_paid( 'edit' ) ) {
+			return false;
+		}
+
+		// Bail if the order doesn't have an intent yet.
+		if ( ! $this->get_intent_from_order( $order ) ) {
 			return $cancel_order;
 		}
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.8.0
+* Fix - Don't auto-cancel Stripe orders that have already been paid, preventing paid orders left at pending (e.g. by a checkout/webhook race) from being cancelled as unpaid
 * Fix - Preserve saved card branding when the same card is later used via a wallet, and render multi-word brands (e.g. Cartes Bancaires) correctly
 * Add - Show Apple Pay / Google Pay branding on saved card tokens in My Account → Payment Methods and at checkout
 * Add - Add a setting to control whether Express Checkout is shown on the WooCommerce Subscriptions change payment method page
--- a/readme.txt
+++ b/readme.txt
@@ -152,6 +152,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.8.0 - xxxx-xx-xx =
+* Fix - Don't auto-cancel Stripe orders that have already been paid, preventing paid orders left at pending (e.g. by a checkout/webhook race) from being cancelled as unpaid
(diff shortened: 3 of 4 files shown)


includes/class-wc-stripe-order-handler.php
```diff
@@ -474,8 +474,20 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+		// Bail if the payment method is not stripe or `stripe_{apm_method}`.
+		if ( ! $order_helper->is_stripe_gateway_order( $order ) ) {
+			return $cancel_order;
+		}
+
+		// Never cancel an order that has already been paid. A race between the Store API checkout
+		// and the Stripe webhook can leave a paid order stuck at `pending` (payment meta written,
+		// status transition lost), which wc_cancel_unpaid_orders() would otherwise cancel.
+		if ( $order->get_date_paid( 'edit' ) ) {
+			return false;
```
daledupreez, line 486: Is it worth triggering some sort of notice or event when we detect this situation? While we're preventing cancellation of the order with this PR, it still feels like there should be some notification or notice regarding the order. That should probably be handled in a follow-up, but I think we should map that out to make sure we surface this situation in some way.

daledupreez (approved):
This is testing as described and looks like it should be an effective guard against this situation when it occurs.

includes/class-wc-stripe-order-handler.php
wjrosa: Agreed. A paid order stuck at `pending` is itself the signal that the checkout/webhook save raced, so it deserves to be observable rather than silently corrected here. I'd keep this PR scoped to the cancellation guard and map a follow-up that, when we hit this branch, adds an order note and fires a `do_action` hook so stores can alert on it. That lines up with the deeper race fix tracked under #2536 / #5362.

wjrosa: Mapped it out in #5496 (stacked on this branch): when we block the cancellation, we now add an order note and fire a `wc_stripe_paid_order_cancellation_prevented` action, guarded by a meta flag so it surfaces once rather than on every scheduled pass.

PR #5487, opened 2026-05-26
Fix unexpected "Address Line 2" field in the Payment Element when the field is disabled in WooCommerce

Mayisha:
Towards STRIPE-1182

 When a store disables the **Address Line 2** billing field in WooCommerce (common on B2B/German stores), Stripe renders an extra "Address Line 2" field *inside* the Payment Element iframe. Because it comes from within the iframe, merchants can't hide it via CSS, checkout field editors, or translation overrides. It's most visible with SEPA Direct Debit — one merchant reported SEPA adoption dropping from 15% to 3% due to the confusing extra field.

**Root cause:** `getHiddenBillingFields()` in `client/stripe-utils/utils.js` maps each WooCommerce billing field to a Stripe collection mode — `never` when the field is enabled in WooCommerce (WC collects it), and `auto` otherwise (so Stripe can collect fields a payment method may require, e.g. country/email). When a store disables Address Line 2, `billing_address_2` drops out of `enabledBillingFields`, so `line2` fell back to `auto` and Stripe rendered its own field.

**Fix:** Force `line2` to `never` unconditionally. Unlike the other fields, Address Line 2 is always optional and is never required by Stripe to create a payment method, so opting out can't trigger a "value required at confirmation" error (this is why `phone`, which can be required, stays `auto`). WooCommerce already owns the address form, so Stripe should never collect this field.

## Testing instructions
Code review should be sufficient.

For manual testing-
1. Disable Address Line 2 in your checkout
    - Blocks checkout: edit the Checkout page → Checkout block → Address block → set Address line 2 to Hidden; or
    - Classic checkout: add a snippet that sets the Address line 2 as hidden

```
add_filter( 'woocommerce_default_address_fields', function ( $fields ) {
      $fields['address_2']['enabled'] = false;
      return $fields;
  } );
```
2. Set your store currency to Euro.
4. Enable SEPA from the payment methods list.
5. Add a product to your cart and go to the checkout page.
6. In `develop`, select SEPA and notice that there is an Address line 2 field in the payment element.

Screenshot 2026-05-26 at 12 47 15 PM

7. In this branch, verify that the Address line 2 field is not present in the payment element.

Screenshot 2026-05-26 at 12 40 10 PM

8. Complete a payment to confirm there are no Stripe confirmation errors.
9. Verify across classic, Blocks, and Optimized checkout.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (approved):
The change looks good, and it tests well! 👍🏼

client/stripe-utils/utils.js
```diff
@@ -463,9 +463,13 @@ export const getHiddenBillingFields = ( enabledBillingFields ) => {
...
-			line2: enabledBillingFields.includes( 'billing_address_2' )
-				? 'never'
-				: 'auto',
+			// The other fields fall back to "auto" when not enabled in WooCommerce so
+			// Stripe can collect them if a payment method needs them. Line 2 is the
+			// exception: it's always optional and never needed by Stripe, so forcing
+			// "never" is safe (unlike phone, it can't cause a confirmation error) and
+			// avoids stores that disable the field getting an unwanted "Address Line
+			// 2" rendered inside the Payment Element.
+			line2: 'never',
```
daledupreez: I think we can simplify this comment as it is quite verbose.
```suggestion
			// Line 2 should never be collected by Stripe.
			// It is not _required_ in any situations, and the are various cases where line 2
			// is hidden in WooCommerce. When the WooCommerce field has been hidden
			// by merchants, we don't want Stripe to collect it within the payment element.
			line2: 'never',
```

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.8.0
+* Fix - Prevent Stripe from rendering an unexpected "Address Line 2" field inside the Payment Element when the field is disabled in WooCommerce checkout
```
daledupreez: Given the implementation, we can remove the edge case details from the description, as we are always preventing line 2 from being displayed.
```suggestion
* Fix - Prevent Stripe from rendering an unexpected "Address Line 2" field inside the Payment Element
```

daledupreez (approved):
Changes LGTM based on inspection.

changelog.txt
Mayisha: Done.

PR #5488, opened 2026-05-26
Require a live connection before disabling test mode

Mayisha:
Fixes STRIPE-1044

Turning off "test mode" switched the Stripe gateway to **live** mode even when no live account was connected (no live API keys). With no live keys, the gateway runs in an invalid state that breaks checkout and the account screens.

Until the test to live transition UX is final, I am removing the ability to uncheck the test mode checkbox when live connection is absent.

This PR guards against that on two layers:
  
- **Backend (authoritative):** In `WC_REST_Stripe_Settings_Controller::update_is_test_mode_enabled()`,
    a request to turn test mode off is ignored when `WC_Stripe_Helper::is_connected( 'live' )`
    is false — test mode stays enabled. This is a soft clamp: the rest of the
    settings still save and the endpoint still returns `200` (no hard rejection).
- **Frontend (UX):** The "Enable test mode" checkbox reads the live connection
    state from `useAccount()`. When test mode is on and no live account is
    connected, the checkbox is shown checked and **disabled**, and the help text
    gains a second line: _"Connect a live Stripe account before turning off test
    mode."_ The original test-card-numbers help text is always shown.

## Testing instructions
- In the Stripe settings, connect a **test** account only (no live keys / live account).
- Confirm **Test mode** is enabled.
- Observe the "Enable test mode" checkbox is **checked**, with the help text showing _"Connect a live Stripe account before turning off test mode."_ on its own line below the usual help text. Also, you can not disable the checkbox.

Screenshot 2026-05-26 at 2 48 48 PM

- (Backend safety net) POST `is_test_mode_enabled: false` to `wc_stripe/settings` — the request returns `200` but `testmode` remains `yes`.
- Now connect a **live** account. The checkbox becomes enabled to change and can be unchecked to switch to live mode; saving persists `testmode = no`.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/settings/payment-settings/test-mode-checkbox.js
```diff
@@ -3,37 +3,62 @@ import interpolateComponents from '@automattic/interpolate-components';
 import { __ } from '@wordpress/i18n';
 import { CheckboxControl } from '@wordpress/components';
 import { useTestMode } from 'wcstripe/data';
+import { useAccount } from 'wcstripe/data/account';
 
 const TestModeCheckbox = () => {
 	const [ isTestModeEnabled, setTestMode ] = useTestMode();
+	const { data } = useAccount();
+	const isLiveAccountConnected = Boolean(
+		data?.oauth_connections?.live?.connected
```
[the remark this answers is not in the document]
Mayisha: Manual key entry is the legacy path, so i don't think it's worth adding a `has_live_keys` REST field just to make the lock handle this edge case for manual-key users.
Also a manual-keys merchant will see the checkbox locked and that should be enough.

client/settings/payment-settings/test-mode-checkbox.js
```diff
@@ -3,37 +3,62 @@ import interpolateComponents from '@automattic/interpolate-components';
...
-					},
-				} ) }
+				help={
+					<>
+						{ helpText }
+						{ isLockedToTestMode && (
+							<>
+								<br />
+								{ __(
+									'Connect a live Stripe account before turning off test mode.',
```
daledupreez: I have two notes for this:
* I think we should switch things around to document that test mode cannot be disabled.
* I think we should consider using bolder text and/or italics to emphasis this message. I think it's too easy to miss otherwise.
```suggestion
									'Live mode cannot be enabled before you have connected a live Stripe account.',
```
We could also use "Test mode cannot be disabled", but I think making it focus on live mode may be better. (Happy to take feedback on this.)

daledupreez (commented):
This is working as described.

However, it feels like we should guard against switching in both directions. And by that I mean:
 * If the Stripe plugin is in live mode, we should prevent switching to test mode if we don't have test keys
 * If the Stripe plugin is in test mode, we should prevent switching to live mode if we don't have live keys

The current implementation is only handling the one direction, and I am not sure why we would not guard against the other situation as well.

diegocurbelo (approved):
The code looks good and it tests well!.

I agree with @daledupreez that we should add the guard in both directions, but I'm ok doing it in a follow-up pr.

client/settings/payment-settings/test-mode-checkbox.js
Mayisha: I like the suggestion. Addressed your feedback in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5488/commits/4476f9be4355120659932b449a7f63c8bf0bb93c

daledupreez (approved):
Changes are looking good to me overall and test well. My only minor comment is that it would be a bit easier to see the restriction notice if it was slightly further away from the other text. With it being so close, it doesn't draw my eye. Adding some top padding should provide that space.

Image

aheckler:
@Mayisha Is the checkbox supposed to receive different styling when disabled? To my eye, it looks exactly the same:

Screenshot taken on 2026-06-16 at 08 36 42 UTC@2x

... and so the fact that I'm not supposed to be able to uncheck it isn't hinted at in the UI itself.

IMHO it could use `opacity: 0.5` or something in CSS, like so:

Screenshot taken on 2026-06-16 at 08 39 10 UTC@2x

PR #5492, opened 2026-05-26
Default OCS and Adaptive Pricing on for 10.8 with audience-targeted banners

malithsen:

This PR defaults Optimized Checkout Suite and Adaptive Pricing **on** for back-book merchants on upgrade and surfaces one of two in-product banners announcing the change or no banner at all, depending on the merchant's pre-upgrade state.

At a high level this PR:

1. Reads pre-flip OC/AP gateway settings, Stripe account country, and Stripe account creation date.
2. Decides whether to flip OC and AP on.
3. Decides which (if any) banner to surface.
4. Writes the banner-visibility options (`wc_stripe_show_ocs_ap_banner`, `wc_stripe_show_ap_only_banner`) and sets a flag that the migration was complete (`wc_stripe_ocs_ap_default_on_migration_ran`)

Migration bails out under the following conditions
- If the Stripe account is based in India
- If the merchant likely had the 10.7 frontbook default-on applied AND the feature is currently off.

There's no marker to directly identify the second condition above, so this PR tries to infer it from two signals.
The plugin version they were on before this upgrade (must include 10.7 or later) AND the date their Stripe account was created (must fall on or after the 10.7.0 release).

> [!NOTE]
> During testing, I noticed that we don't get the Stripe Account creation date (`account.created`) for standard accounts (we only seem to get it for connected accounts). In this case, we consider the merchant as a back-book merchant and flips the flags on.

### Banners

When both OCS and AP are defaulted on
CleanShot 2026-05-28 at 12 39 57@2x

When only AP is defaulted on
CleanShot 2026-05-28 at 12 43 18@2x

When only OCS is defaulted on (**This banner copy is pending approval**)
CleanShot 2026-06-04 at 11 35 16@2x

## Testing instructions

Update the version strings [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/0c4d8b63c53c2233b00c217f9987bbafd77fd4f1/woocommerce-gateway-stripe.php#L25) and [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/0c4d8b63c53c2233b00c217f9987bbafd77fd4f1/woocommerce-gateway-stripe.php#L8) to 10.8.0, then generate a build from this branch.

Please use [this Google Sheet](https://docs.google.com/spreadsheets/d/1VEVQVUPdqzJIqETEUk6nV_oOBd832XWLVXsLHz-mUM4/edit?usp=sharing) for testing. Feel free to duplicate the existing tabs if needed.

High-level test flow for a scenario is as follows:
1. Set up the site with a certain configuration ("Pre-conditions" in the spreadsheet)
2. Upgrade the Stripe plugin with the version you generated earlier
3. Make sure banners and check boxes are set (or not) as expected ("Post-conditions" in the spreadsheet)

For all scenarios except H, you can use the same site, but after each scenario, make sure to delete the migration flags with
```
wp option delete wc_stripe_ocs_ap_default_on_migration_ran wc_stripe_show_ocs_ap_banner wc_stripe_show_ap_only_banner
```

If needed, debug logs should show what branch the migration went through.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	/**
+	 * Entry point hooked to `woocommerce_stripe_updated`.
+	 *
+	 * @return void
+	 */
+	public function maybe_run(): void {
+		if ( 'yes' === get_option( self::MIGRATION_FLAG_OPTION ) ) {
+			WC_Stripe_Logger::info( '[OCS+AP 10.8] Skipping: migration already ran on this site.' );
```
malithsen, line 47: This and other logs in this file are mostly for debugging. Can be removed before merging.

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,269 @@
...
+			'OC-only frontbook-10.7 disabled-AP'    => [ '10.7.0', 'yes', 'no', 'US', $new_ts, 'no', 'no', 'yes', 'no' ],
+			'both-off backbook old-account'         => [ '10.7.0', 'no', 'no', 'US', $old_ts, 'yes', 'no', 'yes', 'yes' ],
+			'both-off frontbook-10.7 disabled-both' => [ '10.7.0', 'no', 'no', 'US', $new_ts, 'no', 'no', 'no', 'no' ],
+			'AP-only-on disabled-OC'                => [ '10.7.0', 'no', 'yes', 'US', $new_ts, 'no', 'no', 'no', 'yes' ],
+			'previous 10.6 both-off recent-account' => [ '10.6.0', 'no', 'no', 'US', $new_ts, 'yes', 'no', 'yes', 'yes' ],
+			'previous 10.6 OC-only recent-account'  => [ '10.6.0', 'yes', 'no', 'US', $new_ts, 'no', 'yes', 'yes', 'yes' ],
+			'India backbook both-off'               => [ '10.7.0', 'no', 'no', 'IN', $old_ts, 'no', 'no', 'no', 'no' ],
+			'India backbook OC-only'                => [ '10.7.0', 'yes', 'no', 'IN', $old_ts, 'no', 'no', 'yes', 'no' ],
+			'unavailable-country backbook both-off' => [ '10.7.0', 'no', 'no', '', $old_ts, 'yes', 'no', 'yes', 'yes' ],
+		];
```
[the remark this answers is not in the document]
malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/1da779d24c65fa7b1520f30861393544fb5278fe

malithsen:
@daledupreez @pierorocca Opening this for review. As a first step, I'm looking for feedback on the [test matrix](https://docs.google.com/spreadsheets/d/1VEVQVUPdqzJIqETEUk6nV_oOBd832XWLVXsLHz-mUM4/edit?usp=sharing), whether it captures all the scenarios we want to handle.

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,270 @@
...
+			'both-off backbook old-account'          => [ '10.7.0', 'no', 'no', 'US', $old_ts, 'yes', 'no', 'yes', 'yes' ],
+			'both-off standard-account null-created' => [ '10.7.0', 'no', 'no', 'US', null, 'yes', 'no', 'yes', 'yes' ],
+			'both-off frontbook-10.7 disabled-both'  => [ '10.7.0', 'no', 'no', 'US', $new_ts, 'no', 'no', 'no', 'no' ],
+			'AP-only-on disabled-OC'                 => [ '10.7.0', 'no', 'yes', 'US', $new_ts, 'no', 'no', 'no', 'yes' ],
+			'previous 10.6 both-off recent-account'  => [ '10.6.0', 'no', 'no', 'US', $new_ts, 'yes', 'no', 'yes', 'yes' ],
+			'previous 10.6 OC-only recent-account'   => [ '10.6.0', 'yes', 'no', 'US', $new_ts, 'no', 'yes', 'yes', 'yes' ],
+			'India backbook both-off'                => [ '10.7.0', 'no', 'no', 'IN', $old_ts, 'no', 'no', 'no', 'no' ],
+			'India backbook OC-only'                 => [ '10.7.0', 'yes', 'no', 'IN', $old_ts, 'no', 'no', 'yes', 'no' ],
+			'unavailable-country backbook both-off'  => [ '10.7.0', 'no', 'no', '', $old_ts, 'yes', 'no', 'yes', 'yes' ],
+		];
```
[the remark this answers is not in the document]
malithsen: Stripe plugin does not release RC or beta versions.

pierorocca:
Thanks for the ping @malithsen. I added a bunch of edge cases into the spreadsheet. I've culled a few out and at the end of my day so my head isn't able to concentrate. I'll take another look tomorrow and feel free to scrutinize that list and delete or update as needed.

client/settings/payment-settings/promotional-banner/ap-only-banner.js
```diff
@@ -0,0 +1,85 @@
...
+const Intro = styled.p`
+	line-height: 20px;
+`;
+
+const Footnote = styled.p`
+	font-size: 12px;
+	color: #757575;
+`;
+
+export const APOnlyBanner = ( { setShowPromotionalBanner } ) => {
```
daledupreez, line 20: Non-blocking nit: It feels like it might be better to call this something like `APEnabledBanner` and the other `OCandAPEnabledBanner`. Capturing the "enabled" aspect could make the use cases and intentions a bit clearer.

client/settings/payment-settings/promotional-banner/get-promotional-banner-type.js
```diff
@@ -1,18 +1,23 @@
...
 	STRIPE_TAX_BANNER,
 } from 'wcstripe/settings/payment-settings/constants';
 import { BNPL_METHODS } from 'wcstripe/stripe-utils/constants';
 
 /**
- * Returns the type of promotional banner to display based on the current extension state.
+ * Returns the type of promotional banner to display based on the current
+ * extension state. The server has already AND-gated the visibility flags
+ * against current OC/AP state and the India geo-exclusion, so this function
+ * only resolves precedence among the candidates.
```
daledupreez: Nit: The addition to the comments here is really contextual to the current change and not particularly general.
```suggestion
 * Returns the type of promotional banner to display based on the current extension state.
```

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -215,6 +215,19 @@ public function admin_scripts( $hook_suffix ) {
...
+
+		$show_ocs_ap_banner = get_option( 'wc_stripe_show_ocs_ap_banner', 'yes' ) === 'yes'
+			&& $is_oc_enabled
+			&& $is_ap_enabled
+			&& ! $is_india_account;
+
+		$show_ap_only_banner = get_option( 'wc_stripe_show_ap_only_banner', 'yes' ) === 'yes'
+			&& $is_oc_enabled
+			&& $is_ap_enabled
+			&& ! $is_india_account;
```
daledupreez: Non-blocking nit: I prefer doing the cheap local variable checks before triggering more expensive `get_option()` calls. Also, Yoda conditions, but I am less bothered by that! 🙃 
```suggestion
		$show_ocs_ap_banner = $is_oc_enabled
			&& $is_ap_enabled
			&& ! $is_india_account
			&& 'yes' === get_option( 'wc_stripe_show_ocs_ap_banner', 'yes' );

		$show_ap_only_banner = $is_oc_enabled
			&& $is_ap_enabled
			&& ! $is_india_account;
			&& 'yes' === get_option( 'wc_stripe_show_ap_only_banner', 'yes' )
```

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,209 @@
...
+	 *
+	 * @return void
+	 */
+	public function maybe_migrate(): void {
+		if ( 'yes' === get_option( self::MIGRATION_FLAG_OPTION ) ) {
+			WC_Stripe_Logger::info( '[OCS+AP 10.8] Skipping: migration already ran on this site.' );
+			return;
+		}
+
+		$previous_version = get_option( 'wc_stripe_version' );
```
daledupreez: Nit: The previous version is supplied as the first argument to the function. It may be best to accept the argument just in case we change the upstream code to remove or update the version option before calling this code.

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,209 @@
...
+	/**
+	 * Server-side visibility flags consumed by the settings-controller gate.
+	 */
+	private const SHOW_OCS_AP_BANNER_OPTION  = 'wc_stripe_show_ocs_ap_banner';
+	private const SHOW_AP_ONLY_BANNER_OPTION = 'wc_stripe_show_ap_only_banner';
+
+	/**
+	 * Epoch time for 2026-05-14 09:30 UTC.
+	 * Stripe accounts created on or after this timestamp are treated as
+	 * likely-10.7 frontbook signups for the audience-exclusion heuristic.
```
daledupreez: This is a really complicated comment. I think we should simplify.
```suggestion
	 * Stripe accounts created on or after this timestamp are treated as
	 * creating their account after 10.7 was released.
```

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,209 @@
...
+	 */
+	private const OC_AP_DEFAULT_ON_RELEASE_TS = 1778751000;
+
+	/**
+	 * ISO country code for the banner geo-exclusion.
+	 */
+	private const EXCLUDED_COUNTRY = 'IN';
```
daledupreez: I think we should improve the name and the documentation for this constant.
```suggestion
	/**
	 * ISO country code for India, where Adaptive Pricing is not available.
	 */
	private const EXCLUDED_AP_COUNTRY = 'IN';
```

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,209 @@
...
+
+		if ( $flip_oc ) {
+			$stripe_settings['optimized_checkout_element'] = 'yes';
+		}
+		if ( $flip_ap ) {
+			$stripe_settings['adaptive_pricing'] = 'yes';
+		}
+		if ( $flip_oc || $flip_ap ) {
+			WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
+		}
```
daledupreez, line 123: I think this code needs to be made more cohesive, because we're doing two separate computations in all of this logic:
 * In `$this->decide_banner_visibility()`, we are computing what banner we will show,
 * But this logic is separately computing what flags we are going to toggle.

It feels like it would be way too easy for the logic here to drift. It feels like it would be much clearer to call something like `[ $should_enable_ocs, $should_enable_ap ] = $this->decide_upgrade_actions( $oc_pre, $ap_pre, $is_frontbook, $is_india );`. We could then use the output variables to update the settings and decide on which banner to show in response to the specified actions.

Separately, I think we should use more self-descriptive variable names.
```suggestion
		[ $show_ocs_ap_banner, $show_ap_banner ] = $this->decide_banner_visibility( $oc_pre, $ap_pre, $is_frontbook, $is_india );

		WC_Stripe_Logger::info( sprintf( '[OCS+AP 10.8] Banner decision: wc_stripe_show_ocs_ap_banner=%s, wc_stripe_show_ap_only_banner=%s.', $show_ocs_ap_banner ? 'yes' : 'no', $show_ap_banner ? 'yes' : 'no' ) );

		update_option( self::SHOW_OCS_AP_BANNER_OPTION, $show_ocs_ap_banner ? 'yes' : 'no' );
		update_option( self::SHOW_AP_ONLY_BANNER_OPTION, $show_ap_banner ? 'yes' : 'no' );

		// - India merchants are excluded from backbook defaults
		// - Skip the flip for any feature the merchant likely
		//   explicitly disabled after frontbook default on in 10.7
		$enable_oc = ! $is_india && ! ( $is_frontbook && ! $oc_pre );
		$enable_ap = ! $is_india && ! ( $is_frontbook && ! $ap_pre );

		if ( $enable_oc ) {
			$stripe_settings['optimized_checkout_element'] = 'yes';
		}
		if ( $enable_ap ) {
			$stripe_settings['adaptive_pricing'] = 'yes';
		}
		if ( $enable_oc || $enable_ap ) {
			WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
		}
```

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,209 @@
...
+	protected function is_likely_frontbook_10_7( string $previous_version ): bool {
+		if ( version_compare( $previous_version, '10.7.0', '<' ) ) {
+			WC_Stripe_Logger::info( sprintf( '[OCS+AP 10.8] Frontbook heuristic: previous_version=%s < 10.7.0 -> not frontbook.', $previous_version ) );
+			return false;
+		}
+		$created = $this->get_account_created_ts();
+		if ( null === $created ) {
+			WC_Stripe_Logger::info( '[OCS+AP 10.8] Frontbook heuristic: account.created unavailable -> not frontbook (conservative).' );
+			return false;
+		}
```
daledupreez, line 154: I think this may indicate two distinct possibilities:
1. We don't have valid credentials, so we don't have any account details
2. We have valid credentials, but this field is not populated -- this is likely the case for merchants who are still connected via API keys and are not connected to our Stripe platform account.

For 1, we would want to set the flag. For 2, the conservative option would be not to enable the feature, as we may not be able to enable settings sync.

In both cases, it may be worth looking at the current account data and `pmc_enabled` (sub)setting to see if more details are available: if we don't have account data, we likely have invalid credentials. Otherwise we can see whether the merchant had settings sync enabled, which may guide us to whether we want to try and enable OCS -- if `pmc_enabled` was `no`, it may not be worth the attempt. (Side note: `pmc_enabled` is reset on upgrade after this code runs.)

includes/class-wc-stripe-update-manager.php
```diff
@@ -81,6 +81,9 @@ protected function get_update_functions(): array {
 			[ new WC_Stripe_Express_Checkout_Add_Change_Payment_Method_Location_Update(), 'maybe_migrate' ],
 		];
 
+		require_once WC_STRIPE_PLUGIN_PATH . '/includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php';
+		$functions[] = [ new WC_Stripe_OCS_AP_Default_On_Update(), 'maybe_migrate' ];
```
daledupreez: It looks like this class has been added to the autoloader classmap, so this `require_once` is unnecessary.
```suggestion
			[ new WC_Stripe_Express_Checkout_Add_Change_Payment_Method_Location_Update(), 'maybe_migrate' ],
			[ new WC_Stripe_OCS_AP_Default_On_Update(), 'maybe_migrate' ],
		];
```

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,270 @@
...
+	 * @dataProvider audience_matrix_provider
+	 */
+	public function test_audience_decision_matrix(
+		string $previous_version,
+		string $oc_pre,
+		string $ap_pre,
+		string $country,
+		?int $account_created,
+		string $expected_show_a,
+		string $expected_show_b,
```
daledupreez: I think we should use better variable names here.
```suggestion
		string $expected_show_ocs_ap_banner,
		string $expected_show_ap_banner,
```

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,270 @@
...
+			'AP-only-on disabled-OC'                 => [ '10.7.0', 'no', 'yes', 'US', $new_ts, 'no', 'no', 'no', 'yes' ],
+			'previous 10.6 both-off recent-account'  => [ '10.6.0', 'no', 'no', 'US', $new_ts, 'yes', 'no', 'yes', 'yes' ],
+			'previous 10.6 OC-only recent-account'   => [ '10.6.0', 'yes', 'no', 'US', $new_ts, 'no', 'yes', 'yes', 'yes' ],
+			'India backbook both-off'                => [ '10.7.0', 'no', 'no', 'IN', $old_ts, 'no', 'no', 'no', 'no' ],
+			'India backbook OC-only'                 => [ '10.7.0', 'yes', 'no', 'IN', $old_ts, 'no', 'no', 'yes', 'no' ],
+			'unavailable-country backbook both-off'  => [ '10.7.0', 'no', 'no', '', $old_ts, 'yes', 'no', 'yes', 'yes' ],
+		];
+	}
+
+	public function test_flip_writes_yes_for_back_book_merchant() {
```
daledupreez: Nit: I think we should use "migration" instead of "flip", as flip implies there can be two directions for the change, which is not correct.
```suggestion
	public function test_migration_writes_yes_for_back_book_merchant() {
```

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,270 @@
...
+		$this->assertSame( 'yes', get_option( self::MIGRATION_FLAG_OPTION ) );
+	}
+
+	/**
+	 * Full audience-decision matrix. Each row drives one execution of maybe_migrate()
+	 * and asserts both banner-visibility options and post-flip gateway state.
+	 *
+	 * @dataProvider audience_matrix_provider
+	 */
+	public function test_audience_decision_matrix(
```
daledupreez: I think there are a number of redundant tests here between the data provider and the individual tests. I think we can remove many of the other tests. If we do that, we should probably improve the naming for this test method too.
```suggestion
	public function test_migration_scenarios(
```

daledupreez (commented):
I have plenty of feedback from inspection.

includes/class-wc-stripe-update-manager.php
malithsen: Yeah, removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/e8cfeb621cde604cbea2209b8c677c444ab29693

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
malithsen: Renamed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/3bd1a732b9fc7f67dba8e3cc0b8350405a8f333c

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
malithsen: Renamed as a part of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/3bd1a732b9fc7f67dba8e3cc0b8350405a8f333c

includes/admin/class-wc-stripe-settings-controller.php
malithsen: Re-ordered the condition and switched to Yoda in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/changes/55e5f6a66c1bc111f55a2a71f3eb4f03669ed59c

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
malithsen: This variable was removed as part of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/c84fe90178342cea91dc55515aef457df28e23cb

We now rely on the existing `get_adaptive_pricing_account_unavailable_reason` function, which checks the country restriction.

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
malithsen: Makes sense. Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/4b891739befb6b3f08a4d9170852bf4329384273

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
malithsen: That's a fair concern. I ended up combining/refactoring these two functions in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/changes/c84fe90178342cea91dc55515aef457df28e23cb while introducing the OC-only banner.

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,216 @@
...
+			WC_Stripe_Logger::info( '[OCS+AP 10.8] Skipping: new install (no previous_version recorded). Frontbook 10.8 path will be handled at OAuth time.' );
+			update_option( self::MIGRATION_FLAG_OPTION, 'yes' );
+			return;
+		}
+
+		if ( version_compare( (string) $previous_version, '10.8.0', '>=' ) ) {
+			WC_Stripe_Logger::info( sprintf( '[OCS+AP 10.8] Skipping: previous_version=%s already >= 10.8.0.', $previous_version ) );
+			update_option( self::MIGRATION_FLAG_OPTION, 'yes' );
+			return;
+		}
```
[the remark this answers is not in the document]
malithsen: The Stripe plugin does not release beta/rc versions.

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
malithsen: Thanks for the breakdown. 

I made this check more precise in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/changes/1577627ef02c0d2c217a47541bc5ac31f9242660

Also introduced an additional check to only enable AP when OCS is eligible. We already have runtime checks for that, but was missing it in the migration previously https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/a0726c880db5de48c7024259dfaee70c86713490

client/settings/payment-settings/promotional-banner/ap-only-banner.js
```diff
@@ -39,24 +28,24 @@ export const APOnlyBanner = ( { setShowPromotionalBanner } ) => {
 		<CardBody>
 			<CardInner>
 				<CardColumn>
-					<Title>
+					<BannerTitle>
```
daledupreez, line 31: Definitely not something for this PR, but it feels like we should eventually move this to a more declarative API. This is a _lot_ of local componentry to compose each relatively simple banner.

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -218,15 +218,19 @@ public function admin_scripts( $hook_suffix ) {
...
+			&& ! $is_india_account
+			&& 'yes' === get_option( 'wc_stripe_show_ocs_ap_banner', 'yes' );
 
-		$show_ap_only_banner = get_option( 'wc_stripe_show_ap_only_banner', 'yes' ) === 'yes'
-			&& $is_oc_enabled
+		$show_ap_only_banner = $is_oc_enabled
 			&& $is_ap_enabled
-			&& ! $is_india_account;
+			&& ! $is_india_account
+			&& 'yes' === get_option( 'wc_stripe_show_ap_only_banner', 'yes' );
```
daledupreez: I am concerned that we may end up showing both banners to a single merchant due to the fact that we're using the same underlying conditions to compute these two flags with the only difference being the option we're looking at for each flag. I worry that dismissing the first banner would then fall through to show the second banner.

Update: the banner options are being used to manage the options that get displayed, so I think my point is somewhat moot.

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -24,35 +24,31 @@ class WC_Stripe_OCS_AP_Default_On_Update {
...
+	public function maybe_migrate( $previous_version = false ): void {
 		if ( 'yes' === get_option( self::MIGRATION_FLAG_OPTION ) ) {
 			WC_Stripe_Logger::info( '[OCS+AP 10.8] Skipping: migration already ran on this site.' );
 			return;
 		}
 
-		$previous_version = get_option( 'wc_stripe_version' );
-
-		// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_var_export -- distinguishing `false` (option unset) from string versions in the log message.
+		// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_var_export -- distinguishing `false` (new install) from string versions in the log message.
```
daledupreez: Side comment: `json_encode( $previous_version )` would also work and generally doesn't require a PHPCS exception.

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,194 @@
...
+		}
+
+		$stripe_settings = WC_Stripe_Helper::get_stripe_settings();
+		$oc_pre          = ( $stripe_settings['optimized_checkout_element'] ?? 'no' ) === 'yes';
+		$ap_pre          = ( $stripe_settings['adaptive_pricing'] ?? 'no' ) === 'yes';
+
+		WC_Stripe_Logger::info( sprintf( '[OCS+AP 10.8] Pre-flip gateway state: optimized_checkout_element=%s, adaptive_pricing=%s.', $oc_pre ? 'yes' : 'no', $ap_pre ? 'yes' : 'no' ) );
+
+		$ap_unavailable_reason = $this->get_ap_unavailable_reason();
+		$ap_unavailable        = null !== $ap_unavailable_reason;
```
daledupreez: Nit: It may be simpler to use `$ap_available` for the variable, as we are using `!==` here and we are then negating the value later in the code. Switching this around simplifies this line and the subsequent usage.
```suggestion
		$ap_available.         = null === $ap_unavailable_reason;
```

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
```diff
@@ -0,0 +1,194 @@
...
+				$show_ap_only ? 'yes' : 'no',
+				$show_ocs_only ? 'yes' : 'no'
+			)
+		);
+
+		update_option( self::SHOW_OCS_AP_BANNER_OPTION, $show_ocs_ap ? 'yes' : 'no' );
+		update_option( self::SHOW_AP_ONLY_BANNER_OPTION, $show_ap_only ? 'yes' : 'no' );
+		update_option( self::SHOW_OCS_ONLY_BANNER_OPTION, $show_ocs_only ? 'yes' : 'no' );
+
+		if ( $enable_oc ) {
```
daledupreez, line 115: Thought: is it worth re-fetching the Stripe settings here? I don't think they could have been invalidated by anything in-process, so we may end up using locally cached data regardless, but there is a lot of logic running between the time that we load the data and when we try to update the option.

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,227 @@
...
+	 * @return MockObject|WC_Stripe_OCS_AP_Default_On_Update
+	 */
+	private function build_migration( ?string $ap_unavailable_reason = null, ?int $created = null, bool $has_account_data = true ) {
+		$migration = $this->getMockBuilder( WC_Stripe_OCS_AP_Default_On_Update::class )
+							->disableOriginalConstructor()
+							->onlyMethods( [ 'get_ap_unavailable_reason', 'get_account_created_ts', 'get_account_data' ] )
+							->getMock();
+		$migration->method( 'get_ap_unavailable_reason' )->willReturn( $ap_unavailable_reason );
+		$migration->method( 'get_account_created_ts' )->willReturn( $created );
+		$migration->method( 'get_account_data' )->willReturn( $has_account_data ? [ 'id' => 'acct_test' ] : [] );
```
daledupreez: Not blocking: There is a lot packed into these lines. It may be worth expanding the lines a bit to make it a bit easier to see what the code is doing.
```suggestion
		$migration = $this->getMockBuilder( WC_Stripe_OCS_AP_Default_On_Update::class )
							->disableOriginalConstructor()
							->onlyMethods(
								[
									'get_ap_unavailable_reason',
									'get_account_created_ts',
									'get_account_data',
								]
							)
							->getMock();

		$migration->method( 'get_ap_unavailable_reason' )
			->willReturn( $ap_unavailable_reason );

		$migration->method( 'get_account_created_ts' )
			->willReturn( $created );

		$migration->method( 'get_account_data' )
			->willReturn( $has_account_data ? [ 'id' => 'acct_test' ] : [] );
```

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,227 @@
...
+			]
+		);
+
+		$this->build_migration()->maybe_migrate();
+
+		$stored = WC_Stripe_Helper::get_stripe_settings();
+		$this->assertSame( 'no', $stored['optimized_checkout_element'], 'Flip must not run when guard is set.' );
+		$this->assertSame( 'no', $stored['adaptive_pricing'], 'Flip must not run when guard is set.' );
+		$this->assertFalse( get_option( self::SHOW_OCS_AP_BANNER_OPTION ), 'Banner A option must not be set.' );
+		$this->assertFalse( get_option( self::SHOW_AP_ONLY_BANNER_OPTION ), 'Banner B option must not be set.' );
```
daledupreez: I think we should improve the test assertion text values here.
```suggestion
		$this->assertSame( 'no', $stored['optimized_checkout_element'], 'OCS must remain disabled when guard is set.' );
		$this->assertSame( 'no', $stored['adaptive_pricing'], 'Adaptive Pricing must remain disabled when guard is set.' );
		$this->assertFalse( get_option( self::SHOW_OCS_AP_BANNER_OPTION ), 'OCS and AP banner option must not be set.' );
		$this->assertFalse( get_option( self::SHOW_AP_ONLY_BANNER_OPTION ), 'AP only banner option must not be set.' );
```

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,227 @@
...
+				'optimized_checkout_element' => 'no',
+				'adaptive_pricing'           => 'no',
+			]
+		);
+
+		$this->build_migration()->maybe_migrate();
+
+		$stored = WC_Stripe_Helper::get_stripe_settings();
+		$this->assertSame( 'no', $stored['optimized_checkout_element'], 'New install must not flip OC.' );
+		$this->assertSame( 'no', $stored['adaptive_pricing'], 'New install must not flip AP.' );
```
daledupreez: As above, I think we should use "enable" instead of "flip".
```suggestion
		$this->assertSame( 'no', $stored['optimized_checkout_element'], 'New install must not enable OC.' );
		$this->assertSame( 'no', $stored['adaptive_pricing'], 'New install must not enable AP.' );
```

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,227 @@
...
+				'optimized_checkout_element' => 'no',
+				'adaptive_pricing'           => 'no',
+			]
+		);
+
+		$this->build_migration()->maybe_migrate( '10.8.0' );
+
+		$stored = WC_Stripe_Helper::get_stripe_settings();
+		$this->assertSame( 'no', $stored['optimized_checkout_element'], 'Already-10.8 must not flip OC.' );
+		$this->assertSame( 'no', $stored['adaptive_pricing'], 'Already-10.8 must not flip AP.' );
```
daledupreez: More flip => enable
```suggestion
		$this->assertSame( 'no', $stored['optimized_checkout_element'], 'Already-10.8 must not enable OCS.' );
		$this->assertSame( 'no', $stored['adaptive_pricing'], 'Already-10.8 must not enable AP.' );
```

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	public function audience_matrix_provider(): array {
+		$old_ts = 1747008000;   // 2025-05-12 — pre-10.7 era.
+		$new_ts = 1779148800;   // 2026-05-19 — post-10.7 era.
+
+		return [
+			// prev_ver, oc_pre, ap_pre, ap_unavailable_reason, created, pmc_enabled, has_account_data, show_ocs_ap, show_ap_only, show_ocs_only, oc_after, ap_after.
+			// Platform-connected merchants (pmc_enabled='yes') - OCS can function.
+			'both-on backbook'                          => [ '10.7.0', 'yes', 'yes', null, $new_ts, 'yes', true, 'no', 'no', 'no', 'yes', 'yes' ],
```
daledupreez: This is really hard to parse. Might it be worth using keyed arrays here? It will be a lot more verbose, but with 12 parameters, it's really hard to see what is going on with positional arrays.

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
```diff
@@ -0,0 +1,227 @@
...
+				'pmc_enabled'                => 'yes',
+				'statement_descriptor'       => 'My Store',
+			]
+		);
+
+		$this->build_migration( null, 1747008000 )->maybe_migrate( '10.7.0' );
+
+		$stored = WC_Stripe_Helper::get_stripe_settings();
+		$this->assertSame( 'yes', $stored['optimized_checkout_element'], 'OC must be flipped to yes.' );
+		$this->assertSame( 'yes', $stored['adaptive_pricing'], 'AP must be flipped to yes.' );
```
daledupreez: Nit: flip => enable
```suggestion
		$this->assertSame( 'yes', $stored['optimized_checkout_element'], 'OCS must be enabled.' );
		$this->assertSame( 'yes', $stored['adaptive_pricing'], 'AP must be enabled.' );
```

daledupreez (commented):
I have some more feedback based on inspection, most of which is not blocking. Will pick up testing tomorrow.

includes/admin/class-wc-stripe-settings-controller.php
malithsen: I think having `yes` as the default is problematic. I've switched to falsey defaults in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/91dc6b29eecc5ebdb9173d2712d1581e80b19401, which also prevents these notices from showing up on fresh installs.

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
malithsen: Addressed it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/be36fb912843cbe8b6ed4afc38aa99fa122f9e7f

I went with `ap_unavailable` as it matches the downstream helper function name, but looking back double negation is confusing, so updated it.

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
malithsen: I don't have a strong opinion about this. We don't mutate `woocommerce_stripe_settings` between the read and the write, so don't really see how the settings could get stale. But, I also don't see any harm in rereading the settings either.

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
malithsen: Good idea, addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/0fea4c911647e01d72270367f5aeab0e923b9b1f

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
malithsen: Thanks. Addressed this and the other test readability suggestions in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/56ccd89e918cc221c06306931b92c026bbc64fdc

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
malithsen: https://github.com/woocommerce/woocommerce-gateway-stripe/commit/56ccd89e918cc221c06306931b92c026bbc64fdc

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
malithsen: https://github.com/woocommerce/woocommerce-gateway-stripe/commit/56ccd89e918cc221c06306931b92c026bbc64fdc

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
malithsen: https://github.com/woocommerce/woocommerce-gateway-stripe/commit/56ccd89e918cc221c06306931b92c026bbc64fdc

tests/phpunit/admin/migrations/class-wc-stripe-ocs-ap-default-on-update-test.php
malithsen: https://github.com/woocommerce/woocommerce-gateway-stripe/commit/56ccd89e918cc221c06306931b92c026bbc64fdc

includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php
malithsen: Thanks! Went with `wp_json_encode` to prevent triggering another sniff

https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492/commits/e5e26b97603891bed788a35a7eabd1a765bef15f

client/settings/payment-settings/promotional-banner/ap-only-banner.js
malithsen: I went back and forth on this. I prefer to leave this as-is. Mostly because we use similar naming throughout this PR for constants, option keys, and file names. If we are to rename, I think we should rename all of those and that feels like a lot of work for a marginal gain.

daledupreez (commented):
The changes are looking good to me based on inspection, but I will pick up deeper testing tomorrow, as I didn't get to it today. 😐

malithsen:
Merging this now. I will address feedback (if any) on the OC-only banner in a follow-up.

PR #5493, opened 2026-05-27
Add order notice/action for unexpected paymentintent confirmation

### LINKED ISSUE #5362: Deferred webhook handler silently abandons captured Stripe charge when order completed by another gateway during deferral window

diegocurbelo:
Fixes STRIPE-1113
Fixes #5362

When a Stripe PaymentIntent succeeds asynchronously (e.g. Stripe Link finishing 3DS out-of-band) for an order that has already been paid via a different gateway, the webhook handler used to log a debug line and exit silently, leaving the merchant unaware of the unexpected Stripe charge.

This PR keeps the silent skip for the normal case (order paid by Stripe) but, when the order's current `payment_method` is **not** a Stripe gateway (i.e. neither `stripe` nor `stripe_*`), it now:

- Adds an order note: amount, PaymentIntent ID, charge ID (when available), the gateway that actually paid the order, and a link to the Stripe dashboard so the merchant can refund manually.
    Screenshot 2026-05-27 at 12 01 42

- Fires `do_action( 'wc_stripe_unexpected_charge_detected', $order, $charge, $webhook_type )` so site-specific integrations can react (e.g. auto-refund).

**No auto-refund is performed: Stripe fees are not returned on refunds, so manual merchant intervention is the safer default.**

## Testing instructions

The production race is hard to reproduce on demand. The most reliable manual path:

0. Make sure the store receives webhooks (e.g. have `npm run listen` running)
1. Add a snippet for the new action:
    ```
    add_action( 'wc_stripe_unexpected_charge_detected', function ( $order, $charge, $type ) {
		WC_Stripe_Logger::debug( "Unexpected charge detected via charge.succeeded on order {$order->get_id()} for charge $charge->id" );
    }, 10, 4 );
    ```
2. Enable the "Cash on delivery" gateway.
3. Place an order using an action-required payment method: like AliPay or CashApp
4. Before confirming the payment, open a new tab and navigate to `My account > Orders`
5. Chose the "Pending payment" order and pay it using "Cash on delivery"
6. Back in the original order (step 2) complete the payment
7. Verify that in the Order page the payment method is "Cash on delivery"
8. Go to the Order details page and check that a new order note was added
    Screenshot 2026-05-27 at 12 01 42
9. Go to the plugin logs, and verify that error entry is present
    Screenshot 2026-05-27 at 12 01 49

Regression check: place a normal Stripe order (3DS card with successful authentication). 2 minutes later, the scheduled job runs but **no** unexpected charge note is added and the hook does **not** fire

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1428,6 +1467,133 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+	 *
+	 * @param WC_Order    $order        The order to evaluate.
+	 * @param string      $intent_id    The Stripe PaymentIntent ID.
+	 * @param object|null $intent       The Stripe PaymentIntent object, when available.
+	 * @param object|null $charge       The Stripe Charge object, when available.
+	 * @param string      $webhook_type The Stripe webhook event type (e.g. 'charge.succeeded' or 'charge.captured').
+	 *
+	 * @return bool True when an orphan note was added, false when skipped.
+	 */
+	protected function maybe_flag_orphaned_payment_intent_on_order( $order, $intent_id, $intent, $charge, $webhook_type ): bool {
```
malithsen: Both callers of this function pass `null` to the `$intent`
Is this parameter really needed?

diegocurbelo: Good point, removed in [5d921ab](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5493/commits/5d921ab0fc0e076800e3b342295a852cc58716a7)

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -588,6 +606,27 @@ public function process_webhook_charge_succeeded( $notification ) {
...
+			$intent_order = WC_Stripe_Helper::get_order_by_intent_id( (string) $charge->payment_intent );
+			if ( $intent_order instanceof WC_Order
+				&& $this->maybe_flag_orphaned_payment_intent_on_order(
+					$intent_order,
+					(string) $charge->payment_intent,
+					$charge,
+					'charge.succeeded'
+				)
+			) {
+				$this->resolved_order = $intent_order;
```
daledupreez: Note that setting `$this->resolved_order` both here and in the other situation could have some subtle side-effects due to the fact that we will now start triggering the `'wc_stripe_webhook_received'` action for these orders where we were not doing that before. It feels like we should have some way to opt in to that behaviour rather than having the existing hooks pick up this new behaviour. And that's specifically because the shape of the hook suggests that we received a successful charge for an order.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -524,11 +524,29 @@ public function process_webhook_dispute_closed( $notification ) {
...
-		$order        = WC_Stripe_Helper::get_order_by_charge_id( $notification->data->object->id );
+		$charge       = $notification->data->object;
+		$order        = WC_Stripe_Helper::get_order_by_charge_id( $charge->id );
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
 
 		if ( ! $order ) {
-			WC_Stripe_Logger::warning( 'Could not find order via charge ID: ' . $notification->data->object->id );
+			// Orphan detection: when the captured charge isn't on any order, fall back to the parent
+			// PaymentIntent and flag if that order was settled via a different gateway.
+			if ( ! empty( $charge->payment_intent ) ) {
```
daledupreez: The "orphan" term is confusing and a bit muddled to me, particularly because I don't know if the order or the charge is meant to be the orphan. (Edit: It looks like the payment intent is considered the "orphan" based on the helper method.)

Might it make sense to use a term like "delayed charge", or "payment intent without charge", or maybe "unexpected charge"?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -588,6 +606,27 @@ public function process_webhook_charge_succeeded( $notification ) {
 		// https://docs.stripe.com/api/events/types#event_types-charge.succeeded
 		$charge = $notification->data->object;
 
+		$order = WC_Stripe_Helper::get_order_by_charge_id( $charge->id );
+
+		// Orphan detection runs before the synchronous-methods filter below so card/3DS orphans
+		// (e.g. shopper abandoned the Stripe flow after authentication and later settled the
+		// order via a different gateway) are also caught. The order will not carry this charge
+		// as its transaction_id, so we fall back to the parent PaymentIntent.
```
daledupreez: As above, I think we could use a clearer term, because I find this comment hard to follow.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1428,6 +1465,82 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+
+		$message = sprintf(
+			/* translators: 1) formatted amount with currency, 2) Stripe PaymentIntent ID, 3) Stripe charge ID, 4) order payment method title, 5) opening anchor tag for the Stripe dashboard link, 6) closing anchor tag. */
+			__( 'Stripe captured a charge of %1$s (PaymentIntent %2$s, charge %3$s) after this order was already paid via %4$s. This orphaned charge needs to be refunded manually from the %5$sStripe dashboard%6$s.', 'woocommerce-gateway-stripe' ),
+			wp_kses_post( $formatted_price ),
+			esc_html( $intent_id ),
+			esc_html( (string) $charge->id ),
+			esc_html( $order->get_payment_method_title() ),
+			'<a href="' . esc_url( $dashboard_url ) . '" target="_blank" rel="noopener noreferrer">',
+			'</a>'
```
daledupreez: Given that the order state should include the actual payment method details, it feels like we're not adding relevant information by specifying that payment method, especially given that any overrides or hooks that modify the title likely won't be triggered in this context.
```suggestion
			/* translators: 1) formatted amount with currency, 2) Stripe PaymentIntent ID, 3) Stripe charge ID, 4) opening anchor tag for the Stripe dashboard link, 5) closing anchor tag. */
			__( 'Stripe captured a charge of %1$s (PaymentIntent %2$s, charge %3$s) after this order was already paid. This delayed charge needs to be refunded manually from the %4$sStripe dashboard%5$s.', 'woocommerce-gateway-stripe' ),
			wp_kses_post( $formatted_price ),
			esc_html( $intent_id ),
			esc_html( (string) $charge->id ),
			esc_html( $order->get_payment_method_title() ),
			'<a href="' . esc_url( $dashboard_url ) . '" target="_blank" rel="noopener noreferrer">',
			'</a>'
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1428,6 +1465,82 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+		}
+		$dedup_meta_key = '_stripe_orphan_charge_flagged_' . $intent_id;
+		if ( '' !== (string) $order->get_meta( $dedup_meta_key ) ) {
+			return false;
+		}
+
+		$currency        = strtoupper( (string) $charge->currency );
+		$decimal_amount  = WC_Stripe_Helper::convert_from_stripe_amount( (int) $charge->amount, $currency );
+		$formatted_price = wc_price( $decimal_amount, [ 'currency' => $currency ] );
+		$dashboard_url   = sprintf( WC_Stripe_Helper::get_transaction_url( $this->testmode ), $intent_id );
```
daledupreez: Question: does the notification include information about whether its for test mode or live mode? It feels wrong to use the general flag here.

Also, it feels like this `sprintf()` should be in a helper method rather than replicated.

daledupreez (commented):
I have some comments based on inspection.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1428,6 +1467,133 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+	 *
+	 * @param WC_Order $order
+	 */
+	protected function is_order_paid_via_non_stripe_gateway( $order ): bool {
+		$payment_method = (string) $order->get_payment_method();
+		if ( '' === $payment_method ) {
+			return false;
+		}
+		return 'stripe' !== $payment_method && 0 !== strpos( $payment_method, 'stripe_' );
+	}
```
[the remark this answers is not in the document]
diegocurbelo: Intentionally not gating on `is_paid()`, for exmaple, BACS/Cheque orders stay `on-hold` until reconciled.
The detection is keyed on the order's *recorded gateway* being non-Stripe, which already means a deliberate switch to another gateway occurred (and not another payment method from the Stripe Gateway).

Also, renamed the helper to `order_uses_non_stripe_gateway()` so the intent is clear.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -524,11 +524,30 @@ public function process_webhook_dispute_closed( $notification ) {
...
+						$charge,
+						'charge.captured'
+					)
+				) {
+					$this->resolved_order = $intent_order;
+					return;
+				}
+			}
+
+			WC_Stripe_Logger::warning( 'Could not find order via charge ID: ' . $charge->id );
```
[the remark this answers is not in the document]
diegocurbelo: Payload validation is a pre-existing, file-wide pattern (every sibling handler accesses `$notification->data->object` directly) and is out of scope here.

includes/class-wc-stripe-webhook-handler.php
diegocurbelo: Good catch!. Rather than adding an opt-in flag, I dropped the `$this->resolved_order = $intent_order` in [03a83b3](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5493/commits/03a83b3710afc128a5ca8bbda1ea557b496b5063). That restores the pre-PR behavior.

includes/class-wc-stripe-webhook-handler.php
diegocurbelo: Fair point, replaced `orphan` with `unexpected` in [53dfbd3](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5493/commits/53dfbd34a243f2c1fa24ba373ae4746bddd3d90a) (it's now consistently named `unexpected charge`)

includes/class-wc-stripe-webhook-handler.php
diegocurbelo: Replaced orphan with unexpected in [53dfbd3](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5493/commits/53dfbd34a243f2c1fa24ba373ae4746bddd3d90a)

includes/class-wc-stripe-webhook-handler.php
diegocurbelo: Good point, dropped the payment method from the note in [2c3695d](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5493/commits/2c3695d5e690d339e87a15abff7390d44e0045e2)

includes/class-wc-stripe-webhook-handler.php
diegocurbelo: Good question — the charge object does include the mode, and I updated the dashboard link to derive its mode from the charge instead of the gateway's flag in [d91f8dd](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5493/commits/d91f8dd08e88f9a668a40d81969e4f5f61e64dc2).

As a side note, the webhook signature is validated against the current gateway mode keys; so the webhook would only be processed if the charge's mode matches the gateway's mode. But this change makes sense, to keep things consistent

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1428,6 +1465,81 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+	 * settled via a different payment method).
+	 *
+	 * @param WC_Order $order
+	 */
+	protected function order_uses_non_stripe_gateway( $order ): bool {
+		$payment_method = (string) $order->get_payment_method();
+		if ( '' === $payment_method ) {
+			return false;
+		}
+		return 'stripe' !== $payment_method && 0 !== strpos( $payment_method, 'stripe_' );
```
daledupreez: Not blocking, but I prefer the use of `str_starts_with()` so the code is more self-descriptive.
```suggestion
		return 'stripe' !== $payment_method && ! str_starts_with( $payment_method, 'stripe_' );
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1428,6 +1465,81 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+	 *
+	 * @param WC_Order $order        The order to evaluate.
+	 * @param string   $intent_id    The Stripe PaymentIntent ID.
+	 * @param object   $charge       The Stripe Charge object.
+	 * @param string   $webhook_type The Stripe webhook event type (e.g. 'charge.succeeded' or 'charge.captured').
+	 *
+	 * @return bool True when an unexpected-charge note was added, false when skipped.
+	 */
+	protected function maybe_flag_unexpected_charge_on_order( $order, $intent_id, $charge, $webhook_type ): bool {
+		if ( ! $this->order_uses_non_stripe_gateway( $order ) ) {
```
daledupreez: As constructed, this uses a double-negative via the `!` and the `uses_non_stripe_gateway()`. It may be a bit easier to read (and implement) if we use `$this->order_uses_stripe_gateway()`.

includes/class-wc-stripe-webhook-handler.php
daledupreez: Also not blocking, but it feels like we might want to check that this is a supported payment method, e.g. by checking whether the payment method key exists in `WC_Stripe_UPE_Payment_Gateway::UPE_AVAILABLE_METHODS`.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1428,6 +1465,81 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+	 * path that can detect the unexpected charge (charge.succeeded fallback, charge.captured fallback).
+	 *
+	 * @param WC_Order $order        The order to evaluate.
+	 * @param string   $intent_id    The Stripe PaymentIntent ID.
+	 * @param object   $charge       The Stripe Charge object.
+	 * @param string   $webhook_type The Stripe webhook event type (e.g. 'charge.succeeded' or 'charge.captured').
+	 *
+	 * @return bool True when an unexpected-charge note was added, false when skipped.
+	 */
+	protected function maybe_flag_unexpected_charge_on_order( $order, $intent_id, $charge, $webhook_type ): bool {
```
daledupreez: Also not blocking, but it feels like we could specify explicit types for the arguments.
```suggestion
	protected function maybe_flag_unexpected_charge_on_order( WC_Order $order, string $intent_id, string $charge, string $webhook_type ): bool {
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1428,6 +1465,81 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+		}
+
+		$currency        = strtoupper( (string) $charge->currency );
+		$decimal_amount  = WC_Stripe_Helper::convert_from_stripe_amount( (int) $charge->amount, $currency );
+		$formatted_price = wc_price( $decimal_amount, [ 'currency' => $currency ] );
+		$dashboard_url   = WC_Stripe_Helper::get_transaction_url_for_id( $intent_id, empty( $charge->livemode ) );
+
+		$message = sprintf(
+			/* translators: 1) formatted amount with currency, 2) Stripe PaymentIntent ID, 3) Stripe charge ID, 4) opening anchor tag for the Stripe dashboard link, 5) closing anchor tag. */
+			__( 'Stripe captured a charge of %1$s (PaymentIntent %2$s, charge %3$s) after this order was already paid. This unexpected charge needs to be refunded manually from the %4$sStripe dashboard%5$s.', 'woocommerce-gateway-stripe' ),
```
daledupreez: Is it worth noting that payment was via some other gateway/mechanism?
```suggestion
			__( 'Stripe captured a charge of %1$s (PaymentIntent %2$s, charge %3$s) after this order was already paid by another gateway. This unexpected charge needs to be refunded manually from the %4$sStripe dashboard%5$s.', 'woocommerce-gateway-stripe' ),
```

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -339,6 +339,390 @@ function ( $passed_order ) use ( $order ) {
...
+	}
+
+	/**
+	 * Repeat unexpected-charge detections for the same PaymentIntent on the same order must be no-ops: only one
+	 * note is added and the action fires exactly once, even when multiple webhook paths detect it.
+	 */
+	public function test_flag_unexpected_charge_is_idempotent_per_intent() {
+		$intent_id = 'pi_dedup';
+
+		$order = WC_Helper_Order::create_order();
```
daledupreez: Not blocking: there is a lot of duplicated code across these tests. It may be worth looking into a data provider for some of the test cases, or possibly some helper methods for the setup aspects.

daledupreez (approved):
The changes LGTM and test well.

I have some non-blocking comments, but we can ship this as needed.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
diegocurbelo: Refactored in [a174d93](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5493/commits/a174d933d824e666c40100c5480e93f610d8734b)

includes/class-wc-stripe-webhook-handler.php
diegocurbelo: Fair point, refactored in [02f0934](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5493/commits/02f09349a891481e06f3a5576d53c6f10d58105b)

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -569,36 +572,18 @@ public function test_process_webhook_capture_flags_unexpected_charge_via_payment
 	public function test_process_webhook_capture_does_not_flag_when_paid_via_stripe() {
```
daledupreez, line 572: Not blocking: is it worth adding multiple combinations of payment methods to this test so we have deeper coverage? While the current implementation is mostly agnostic to the payment methods involved, it feels like it would be good to make sure this works across payment methods.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -509,58 +542,28 @@ public function test_process_webhook_charge_succeeded_flags_unexpected_card_char
...
-			'id'                     => 'ch_capture_unexpected',
-			'captured'               => true,
-			'payment_intent'         => $intent_id,
-			'amount'                 => 8000,
-			'currency'               => 'eur',
-			'payment_method_details' => (object) [ 'type' => 'card' ],
-		];
+		$order        = $this->create_order_with_intent( 'klarna_payments', $intent_id );
+		$notification = $this->build_charge_notification(
+			'charge.captured',
```
daledupreez, line 547: Might it be worth adding coverage for all payment methods that support dedicated capture events?

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -458,48 +514,25 @@ public function test_process_webhook_charge_succeeded_fallback_skips_uncaptured_
...
-			'payment_method_details' => (object) [ 'type' => 'card' ],
-		];
+		$order        = $this->create_order_with_intent( 'klarna_payments', $intent_id );
+		$notification = $this->build_charge_notification(
+			'charge.succeeded',
+			$intent_id,
+			[
+				'id'                     => 'ch_card_unexpected',
+				'amount'                 => 5000,
+				'payment_method_details' => (object) [ 'type' => 'card' ],
```
daledupreez, line 524: Might it be worth adding coverage for additional payment methods that trigger charge.succeeded events? And possibly additional underlying payment methods, too?

daledupreez (approved):
The changes are looking good to me, though we can consider some additional test cases in follow-up PRs.

PR #5496, opened 2026-05-27
Surface paid Stripe orders whose unpaid-cancellation is prevented

wjrosa:
Fixes STRIPE-1076 (follow-up)
Refs #5485, #5486

> [!NOTE]
> Stacked on #5486 — this PR targets the `fix/stripe-1076-prevent-cancelling-paid-orders` branch. The diff will collapse to just these two files once #5486 merges and the base is retargeted to `develop`.

Follow-up to #5486, addressing [@daledupreez's review](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5486#discussion_r3310239840): a paid order stuck at `pending` is a symptom of the checkout/webhook race, so silently keeping it alive isn't enough — the situation should be observable.

When `prevent_cancelling_orders_awaiting_action()` blocks the unpaid-cancellation of a paid Stripe order, it now:

- Adds an **order note** so the stuck status is visible in the admin.
- Fires a `wc_stripe_paid_order_cancellation_prevented` **action** (passing the `WC_Order`) so stores can hook in to log, alert, or reconcile the status.

`wc_cancel_unpaid_orders()` re-runs on a schedule and would re-match the same stuck order on every pass, so a `_stripe_paid_order_cancellation_prevented` meta flag keeps the note + action **idempotent** — surfaced exactly once per order.

This does not change the cancellation decision itself (that remains #5486's job); it only makes the prevented case detectable. The deeper fix for the concurrent-save race is still tracked under #2536 / the #5362 lineage.

## Testing instructions

1. Create a Stripe order and put it in the paid-but-pending state (set `_date_paid`, keep status `pending`).
2. Run `wc_cancel_unpaid_orders()` (or wait for the scheduled `woocommerce_cancel_unpaid_orders` action).
3. The order is **not** cancelled, **and** it now carries an order note explaining that cancellation was prevented because it was already paid.
4. Run the cancellation pass again — no duplicate note is added.
5. (Optional) Hook `wc_stripe_paid_order_cancellation_prevented` and confirm it fires once with the order.

Automated coverage:

- `tests/phpunit/class-wc-stripe-order-handler-test.php` — `test_surfaces_prevented_paid_order_cancellation_only_once` asserts the action fires once, the meta flag is set, and exactly one order note is added across two cancellation passes. Existing `test_prevent_cancelling_orders_awaiting_action` and `test_prevent_cancelling_orders_that_have_been_paid` still pass.

---

### Changelog entry

<details>

#### Comment

This is a hardening follow-up to #5486, which lands in the same unreleased version (10.8.0). The user-facing fix is already described by #5486's changelog entry; the order note and developer action added here enhance that same fix, so no separate entry is needed.

</details>

**Post merge**


includes/class-wc-stripe-order-handler.php
```diff
@@ -498,4 +499,31 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+	 * @since 10.8.0
+	 *
+	 * @param WC_Order $order The paid order whose cancellation was prevented.
+	 */
+	private function surface_prevented_paid_order_cancellation( $order ): void {
+		if ( $order->get_meta( '_stripe_paid_order_cancellation_prevented' ) ) {
+			return;
+		}
+
+		$order->add_order_note( __( 'This order has already been paid, but its status is still pending. Stripe prevented it from being auto-cancelled as unpaid; please review and update the status.', 'woocommerce-gateway-stripe' ) );
```
daledupreez: Nit: We should specify more context to make it clear that this is local to the site and not Stripe-the-company preventing the autocancellation.
```suggestion
		$order->add_order_note( __( 'This order has already been paid, but its status is still pending. The Stripe gateway prevented it from being auto-cancelled as unpaid; please review and update the status.', 'woocommerce-gateway-stripe' ) );
```

More generally, should the merchant be checking the _order status_ or the _Stripe payment status_? It's not clear to me what the merchant should be reviewing based on this wording.

includes/class-wc-stripe-order-handler.php
```diff
@@ -498,4 +499,31 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+	/**
+	 * Surfaces a paid-but-pending Stripe order whose auto-cancellation we just blocked via an order
+	 * note and an action hook. A meta flag keeps it idempotent across wc_cancel_unpaid_orders() retries.
+	 *
+	 * @since 10.8.0
+	 *
+	 * @param WC_Order $order The paid order whose cancellation was prevented.
+	 */
+	private function surface_prevented_paid_order_cancellation( $order ): void {
+		if ( $order->get_meta( '_stripe_paid_order_cancellation_prevented' ) ) {
```
daledupreez: Should we not check that the value is `'yes`'? That would also make it clearer that this is our meta flag for preventing repeat notifications.
```suggestion
		if ( 'yes' === $order->get_meta( '_stripe_paid_order_cancellation_prevented' ) ) {
```

tests/phpunit/class-wc-stripe-order-handler-test.php
```diff
@@ -68,4 +68,47 @@ public function test_prevent_cancelling_orders_that_have_been_paid() {
...
+		add_action(
+			'wc_stripe_paid_order_cancellation_prevented',
+			function () use ( &$fired ) {
+				$fired++;
+			}
+		);
+
+		// Two scheduled passes over the same stuck order.
+		$this->order_handler->prevent_cancelling_orders_awaiting_action( true, $order );
+		$this->order_handler->prevent_cancelling_orders_awaiting_action( true, wc_get_order( $order->get_id() ) );
```
daledupreez, line 101: Is it worth also testing that the meta field is set and `$fired` is 1 after the first pass? And then that the second call doesn't change either value?

It feels like that more clearly tests the order of operations than the current test, which would still pass if the second call performed all the actions.

daledupreez (commented):
I have some feedback from inspection.

tests/phpunit/class-wc-stripe-order-handler-test.php
wjrosa: Done — now asserts state between the two passes: pass 1 sets the flag, fires the action, and adds the note; pass 2 leaves all three unchanged. The old end-state check couldn't tell that apart from "first pass no-ops, second does everything." Pulled the note lookup into a helper since it's asserted twice.

daledupreez (approved):
This is working as described for me, and definitely is an improvement over the current state. However, it feels like we should consider a louder notification as a follow-up. But I think we should get this merged in the meantime.

includes/class-wc-stripe-order-handler.php
```diff
@@ -498,4 +499,31 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+	 * @since 10.8.0
+	 *
+	 * @param WC_Order $order The paid order whose cancellation was prevented.
+	 */
+	private function surface_prevented_paid_order_cancellation( $order ): void {
+		if ( 'yes' === $order->get_meta( '_stripe_paid_order_cancellation_prevented' ) ) {
+			return;
+		}
+
+		$order->add_order_note( __( 'This order has already been paid, but its status is still pending. The Stripe gateway prevented it from being auto-cancelled as unpaid; please review and update the status.', 'woocommerce-gateway-stripe' ) );
```
daledupreez: As flagged by CodeRabbit, I think we should specify that we're talking about the order status.
```suggestion
		$order->add_order_note( __( 'This order has already been paid, but its status is still pending. The Stripe gateway prevented it from being auto-cancelled as unpaid; please review and update the order status.', 'woocommerce-gateway-stripe' ) );
```

PR #5497, opened 2026-05-27
Initial infrastructure to support filtering for agentic feed

daledupreez:

This PR implements the core functionality needed to support general filtering of products for the agentic feed. The approach here is not to build the UI, but to build the code that implements the core filtering logic in a way that will integrate with the existing feed interfaces and infrastructure, and allow us to build a clean UI in a follow-up change.

The basic approach is to build a new `WC_Stripe_Agentic_Commerce_Product_Filter` class that uses data stored in the `wc_stripe_agentic_commerce_product_filters` option and/or injected via the `wc_stripe_agentic_commerce_product_filter` filter to build the query that we will then use to identify products from the existing `ProductWalker` logic.

One complex issue is that due to us putting together only one set of query arguments that will be passed to `wc_get_products()`, we are limited in terms of what filters can be combined. This PR implements support for the following filters:
 * Only product IDs - a set of product IDs can be used to specify simple or variation products that will be included in the feed.
 * Only variable product IDs - a set of variable product IDs can be used to specify a set of variable product IDs where we will include their child variations in the feed.
 * ORed taxonomy conditions - any combination of product categories, tags, or brands that will be ORed together.

While it would be more merchant-friendly to offer more flexibility, we don't have that ability without adding some serious downsides to how we approach the logic here, as we'd need to run one or more pre-queries to identify matching products, and then use the resulting product IDs when querying the database, which would not scale well for large product catalogs.

The other issue here is that we may unintentionally be excluding variation products when taxonomy-based queries are used: my testing showed that taxonomy relationships are created for the variable product, but not for the child variation, which may mean that taxonomy-based queries may not include variations. :-/

## Testing instructions

 * Set up a store using this branch and ensure that you are connected to Stripe
 * Ensure that the Agentic Commerce feature flag is enabled: `wp option update _wcstripe_feature_agentic_commerce yes`
 * Ensure you have multiple products in the store and a variety of products, product categories, and product tags
 * Navigate to the Stripe admin UI and go to the Advanced settings tab
 * Ensure that Agentic Commerce is enabled
 * Either wait for the initial product sync to complete, or trigger a manual product sync, and confirm it is successful - make a note of the number of products that were synced
 * Run the following option updates to specify different criteria, and trigger a re-sync after each modification -- ensure that the expected number of products are uploaded
   - `wp option update wc_stripe_agentic_commerce_product_filters --format=json '{"product_ids":[20,30]}'` - replace `20` and `30` with two or more valid product IDs
   - `wp option update wc_stripe_agentic_commerce_product_filters --format=json '{"category_ids":[22]}'` - replace 22 with a valid product category ID
   - Feel free to try similar changes for tags and/or product brands

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -240,23 +240,38 @@ public function deactivate(): void {
...
+			'type'   => [
+				\Automattic\WooCommerce\Enums\ProductType::SIMPLE,
+				\Automattic\WooCommerce\Enums\ProductType::VARIATION,
+			],
+			'status' => [ \Automattic\WooCommerce\Enums\ProductStatus::PUBLISH ],
+		];
+
+		$filter = new WC_Stripe_Agentic_Commerce_Product_Filter();
+		if ( $filter->has_filters() ) {
+			$args['include'] = $filter->get_filtered_product_ids();
```
wjrosa: **Approach: `include` pre-resolution fights the streaming walker.** This flattens the entire match set to an ID array, but `ProductWalker` paginates (100/batch) specifically to stream large catalogs — and it carries this `include` into *every* batch query as a potentially huge `post__in`, after `get_filtered_product_ids()` already materialized all IDs via `limit => -1`. On a 100k catalog that's the opposite of what the walker buys us.

Worth considering: contribute `tax_query` clauses that WP_Query resolves *during* pagination instead of pre-flattening to IDs. The variable→variation expansion is the tricky part there, but it's off by default, so the common path may not need pre-resolution at all.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -240,23 +240,38 @@ public function deactivate(): void {
...
 		 * @param array $args WP_Query arguments.
 		 */
-		return apply_filters(
-			'wc_stripe_agentic_commerce_product_query_args',
-			[
-				'type'   => [ 'simple', 'variation' ],
-				'status' => [ 'publish' ],
-			]
-		);
+		return apply_filters( 'wc_stripe_agentic_commerce_product_query_args', $args );
```
wjrosa: **Collides with #5458's documented `post__in` contract.** `include` maps to `post__in` in the product data store (`WC_Product_Data_Store_CPT`: `'include' => 'post__in'`). #5458's docblock tells adapters (WAS) to scope the feed via `post__in`/`post__not_in` on this *same* `wc_stripe_agentic_commerce_product_query_args` filter — so once both land, an adapter setting `post__in` and this code setting `include` write the same WP_Query var and one clobbers the other depending on filter order. `post__not_in` + `include` also produces an AND that may not match intent.

Suggest picking one canonical key, documenting precedence, and intersecting with any pre-existing `post__in` rather than overwriting. A test for the `include` + `post__not_in` combination would be worth having given three PRs touch this filter.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -240,23 +240,38 @@ public function deactivate(): void {
...
+		$args = [
+			'type'   => [
+				\Automattic\WooCommerce\Enums\ProductType::SIMPLE,
+				\Automattic\WooCommerce\Enums\ProductType::VARIATION,
+			],
+			'status' => [ \Automattic\WooCommerce\Enums\ProductStatus::PUBLISH ],
+		];
+
+		$filter = new WC_Stripe_Agentic_Commerce_Product_Filter();
+		if ( $filter->has_filters() ) {
```
wjrosa, line 330: Minor: `has_filters()` calls `get_filters()`, then `get_filtered_product_ids()` calls it again — so `get_filters()` runs twice and the `wc_stripe_agentic_commerce_product_filter` override hook fires twice per sync. Harmless for pure callbacks, but a side-effecting/expensive adapter callback runs 2×. Resolving once and threading the result through avoids it.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -240,23 +240,38 @@ public function deactivate(): void {
 	/**
 	 * Get product feed query arguments.
 	 *
+	 * Constructs the `wc_get_products()` argument set passed to the walker.
+	 * When a {@see WC_Stripe_Agentic_Commerce_Product_Filter} has been
+	 * configured (via stored options or the
+	 * `wc_stripe_agentic_commerce_product_filters` filter), the set of
+	 * filtered product IDs is used the base set of IDs for this query.
```
wjrosa: Nit: "is used the base set of IDs" → "is used **as** the base set of IDs".

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
```diff
@@ -0,0 +1,392 @@
...
+	 *
+	 * @since 10.8.0
+	 * @param array  $values   Array of taxonomy IDs and/or slugs.
+	 * @param string $taxonomy Taxonomy slug.
+	 * @return int[]
+	 */
+	private function normalize_taxonomy_ids( array $values, string $taxonomy ): array {
+		$out = [];
+		foreach ( $values as $value ) {
+			if ( ctype_digit( $value ) ) {
```
wjrosa: **Bug:** `ctype_digit( 123 )` returns `false` — with an int argument it's read as an ASCII codepoint, not the digits (and it's deprecated on PHP 8.1+). So integer term IDs hit neither this branch nor the `is_string()` branch below and get silently dropped. The tests pass only because they also pass the slug, which resolves to the same term and masks it. Mirror `normalize_ids()` and use `is_numeric()` — this will bite once the settings UI POSTs integer term IDs.

```
$ php -r 'var_dump(ctype_digit(123), ctype_digit("123"));'
bool(false)
bool(true)
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
```diff
@@ -0,0 +1,392 @@
...
+
+	/**
+	 * Value returned by {@see self::get_filtered_product_ids()} to indicate no matching products.
+	 * Intentionally not [] as that is treated as "no filter" by the Agentic feed logic.
+	 * Uses 0 as the product ID to indicate a non-existent product ID should be matched.
+	 *
+	 * @var int[]
+	 * @since 10.8.0
+	 */
+	private const EMPTY_RESOLUTION = [ 0 ];
```
wjrosa: The `[0]` sentinel relies on every downstream consumer of `wc_stripe_agentic_commerce_product_query_args` leaving the `0` intact. An adapter that does `array_filter()` / `array_merge()` + de-dupe on `include` drops it and silently flips "match nothing" → "match everything," leaking the deliberately filtered-out catalog. Consider a less truthy-fragile signal, or document the invariant loudly.

Related semantics question: when filters resolve to zero products, `sync_feed()` deletes the file and returns false, so Stripe keeps the *previously-uploaded* full catalog rather than being emptied — intended, or should it push an empty feed?

wjrosa (commented):
Some early feedback on the API and approach — mostly about how this composes with the surfaces in #5458/#5472, plus one concrete bug. The normalization layer and test coverage look solid.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
daledupreez: Agreed that this is a problem. 😐 

I tested this out today and assigned a category to a variable product, and then generated variations for that product. Unfortunately, the variation products don't have the category assigned, so we can't run a dedicated query for variations that match a taxonomy query -- I think we need to get the matching variable product IDs in one pass and then the variations in a follow-up. 😬 

I am going to explore whether we can override the `ProductLoader` class to support this kind of dual query case somewhat transparently, but I don't know how complex that might end up being.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
daledupreez: I think this gets at a downstream question: when running these queries, how does the downstream code handle products in non-published statuses? I've limited the set of products to published products, but it feels like we'd want deleted products to be picked up so their deletion (if that is recent) gets propagated through to Stripe. But I can't recall whether including the products in the feed is the right way to handle deletion.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
daledupreez: I got some exploration in and realised a few minutes ago that `ProductWalker` has a private constructor. I've just pushed up code that would likely allow us to build a custom ProductLoader that combines multiple queries, but we don't have a way to inject that loader instance into the Product Walker. Back to the drawing board for this one...

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter-test.php
```diff
@@ -0,0 +1,1117 @@
...
+								],
+								...[
+									$product_brand_exists ? [
+										'taxonomy'         => 'product_brand',
+										'field'            => 'term_id',
+										'terms'            => [ 42, 9999 ],
+										'operator'         => 'IN',
+										'include_children' => false,
+									] : null,
+								],
```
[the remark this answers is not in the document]
daledupreez: Fixed in 7247c7823b508780e3e58c9060098366060a2676.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -240,23 +240,48 @@ public function deactivate(): void {
 	/**
 	 * Get product feed query arguments.
 	 *
+	 * Constructs the `wc_get_products()` argument set passed to the walker.
+	 * When a {@see WC_Stripe_Agentic_Commerce_Product_Filter} has been
+	 * configured (via stored options or the
+	 * `wc_stripe_agentic_commerce_product_filters` filter), the query arguments
+	 * are built from those criteria.
```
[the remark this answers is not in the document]
daledupreez: Oops. Fixed in 92af85db3621235014ae23565bb634e41236f1b0.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter-test.php
```diff
@@ -0,0 +1,1117 @@
...
+	public function test_get_effective_filter_types( array $stored_filters, array $expected_types ): void {
+		update_option( $this->option_name, $stored_filters );
+
+		$filter = new WC_Stripe_Agentic_Commerce_Product_Filter();
+
+		$actual_types = $filter->get_effective_filter_types();
+		$actual_types = sort( $actual_types );
+
+		$expected_types = sort( $expected_types );
+		$this->assertEqualsCanonicalizing( $expected_types, $actual_types );
```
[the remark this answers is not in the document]
daledupreez: Addressed via 660995c074391677a6ad27a01a92c4618b0c3538, which removes all local sorting and relies on  `assertEqualsCanonicalizing()` to allow ordering to be ignored.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-loader-test.php
```diff
@@ -0,0 +1,361 @@
...
+					\Automattic\WooCommerce\Enums\ProductType::VARIABLE,
+				],
+				'limit'  => -1,
+				'return' => 'ids',
+			]
+		);
+		foreach ( $existing as $product_id ) {
+			WC_Helper_Product::delete_product( $product_id );
+		}
+	}
```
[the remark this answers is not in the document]
daledupreez: Removed in 41eaf1e65f848c3bf28af7dce99ae6e32a13552e.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-loader-test.php
```diff
@@ -0,0 +1,361 @@
...
+
+		$loader = new WC_Stripe_Agentic_Commerce_Product_Loader();
+		$args   = $this->paginated_args( [], $variable_ids, 1, 3 );
+
+		$result = $loader->get_products( $args );
+
+		$this->assertSame( $variable_ids, $this->ids( $result->products ) );
+		$this->assertSame( 2, $result->total );
+		$this->assertSame( 1, $result->max_num_pages );
+	}
```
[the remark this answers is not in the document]
daledupreez: This file was removed in ee95b921e9521a0acafda5323b03e85606d89c25, so the issue is moot.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-loader-test.php
```diff
@@ -0,0 +1,361 @@
...
+	/**
+	 * @dataProvider provider_absent_or_empty_custom_filter
+	 */
+	public function test_returns_parent_result_when_custom_filter_absent_or_empty( $value ): void {
+		$simple_ids = $this->create_simple_products( 3 );
+
+		$args = $this->parent_args( $simple_ids );
+		if ( null !== $value ) {
+			$args[ WC_Stripe_Agentic_Commerce_Product_Loader::CUSTOM_FILTER_KEY ] = $value;
+		}
```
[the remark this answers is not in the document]
daledupreez: I removed the test file in ee95b921e9521a0acafda5323b03e85606d89c25, so this issue is moot.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter-test.php
```diff
@@ -0,0 +1,625 @@
...
+		}
+		$this->created_products = [];
+
+		foreach ( $this->created_terms as [$term_id, $taxonomy] ) {
+			wp_delete_term( $term_id, $taxonomy );
+		}
+		$this->created_terms = [];
+
+		parent::tearDown();
+	}
```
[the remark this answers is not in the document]
daledupreez: 889fae27843ccaf47f55a1dee3eb38192daee4c0 ensures that we don't try to create the taxonomy when it doesn't exist yet, so the cleanup is no longer needed.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
daledupreez: @wjrosa, as a quick note here, I don't see any `post__in` references in that PR. Were they removed?

daledupreez: Ah, I see they are in the main PR description, not in the actual code or PHPDoc. I've added a general warning to the docs for the `wc_stripe_agentic_commerce_product_query_args` filter in 9806a52dbe64e5bd72484f100882c41326cb289d.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
daledupreez: I added an instance variable to cache this across some of the commits I made on Friday.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
daledupreez: No longer relevant; marking as resolved.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
daledupreez: The sentinel has been removed, so I am resolving this thread.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
daledupreez: This comment was rephrased in light of me changing my approach. Marking as resolved.

daledupreez:
@coderabbitai resume

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: The new approach (return query args directly, single `wc_get_products()` call) sidesteps the original concern — walker streams natively, no `include` flattening. LGTM.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: The generic warning at line 339 reads fine — adapters can opt in carefully. 👍

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Verified — the `private ?array $filters` cache on the instance does the trick. `has_filters()` and `get_query_args()` both hit it within a single call. Thanks!

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
wjrosa: The `ctype_digit` is still there in `normalize_taxonomy_ids()` at line 308. A REST/admin caller passing `categories: [123, 456]` (ints) trips it — on PHP 7.4 / 8.0, `ctype_digit(123)` reads 123 as ASCII codepoint '{' and returns false, then `is_string($value)` is false too, so integer term IDs in the 2–47 / 58–255 range silently drop. Tests miss it because they pass slugs alongside the IDs and both resolve to the same term. `is_numeric($value)` (mirroring `normalize_ids()`) would fix it.

daledupreez:
@wjrosa, this PR is ready for another round of review. Note that I also have a draft PR in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5519 that builds a UI for the filtering, but that's not ready for testing yet.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
```diff
@@ -0,0 +1,454 @@
+<?php
+/**
+ * Product filter for the Agentic Commerce feed.
+ *
+ * @package WooCommerce_Stripe
+ * @since 10.8.0
```
malithsen:
```suggestion
 * @since 10.9.0
```

We should update these tags

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
```diff
@@ -0,0 +1,454 @@
...
+	 * @return bool True when the option was written successfully.
+	 */
+	public function save_filters( array $filters ): bool {
+		$product_ids = [];
+		if ( is_array( $filters['product_ids'] ?? null ) ) {
+			$product_ids = $this->validate_product_ids(
+				$filters['product_ids'],
+				[
+					\Automattic\WooCommerce\Enums\ProductType::SIMPLE,
+					\Automattic\WooCommerce\Enums\ProductType::VARIATION,
```
malithsen, line 108: nit: maybe import these at the top of the file to make this more concise
```suggestion
					ProductType::SIMPLE,
					ProductType::VARIATION,
```

malithsen (approved):
Looks good and tests well.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
daledupreez: Updated in 56d24ced59d5e0aec694b1c7206495fe1fd9bb7c.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-filter.php
daledupreez: I have a strong preference for using fully-qualified references as much as possible, mostly because it makes the reference 100% explicit inline. I find that most useful for diffs, as it means I know exactly what the reference is, and where it is coming from (as the `use` statements are rarely visible in diffs). It also means that any time I am reading the code, even in an IDE, I don't need to spend any mental effort working out what something refers to, as it's 100% inline.

This may be worth a more dedicated discussion, TBH.

malithsen: I guess this comes down to preference/setup and I don't have a strong opinion here and perfectly fine with having these FQNs. For me, it's the opposite. I find the long paths (especially inside arrays) breaks the flow of reading. When I do want to find where things are defined it's just a `gd` away.

PR #5500, opened 2026-05-28
Skip PR CI workflows when only docs/release-notes change

wjrosa:

Skip the PR-triggered build, test, lint, bundle-size, and Jurassic Ninja workflows when a PR only touches docs/release-notes files. Concretely:

- Adds a shared `detect-code-changes` composite action under `.github/actions/` that returns `code=true|false` based on whether any changed file falls outside the docs allow-list (`**/*.md`, `docs/**`, `.github/ISSUE_TEMPLATE/**`, `LICENSE.txt`, `changelog.txt`, `readme.txt`).
- Wires the action in as a gating `changes` job in `pr-build-live-branch.yml` (Jurassic Ninja), `e2e-tests.yml`, `php-tests.yml`, `compatibility.yml`, `js-tests.yml`, `linting.yml`, and `bundle-size.yml`.
- Downstream jobs (and any of their transitive `needs:` dependents) become `needs: changes` + `if: needs.changes.outputs.code == 'true'`, so docs-only PRs cause them to skip.

A skipped job still posts a status check that counts as a passing required check for branch protection, so docs-only PRs remain mergeable even if these workflows are configured as required checks — which is why this is a job-level skip rather than `paths-ignore` at the workflow trigger.

`phpstan.yml` already has a `paths:` allow-list (PHP files only) and didn't need a change. `validate-changelog.yml` and `pr-require-milestone.yml` are intentionally left running on every PR — those *are* the docs-PR gates.

## Testing instructions

1. After this PR merges, open a docs-only follow-up PR (e.g. tweak a `.md` file). Confirm each of the seven workflows reports its `changes` job as success and the downstream build/test jobs as `Skipped`.
2. Open or push to a PR that touches a non-docs file. Confirm the same workflows run their downstream jobs as usual.
3. If any of these workflows are configured as required status checks on `develop`, confirm a docs-only PR is mergeable — the skipped check should satisfy branch protection.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️) — CI-config change; verification is observing real PRs (see testing instructions).
-   [x] Tested on mobile (or does not apply) — does not apply.

### Changelog entry

<details>

#### Comment

CI-only change; no user-visible behavior.

</details>

**Post merge**


wjrosa:
@coderabbitai not applicable — the convention in this repo is unpinned `actions/checkout@v{N}` (and other first-party `actions/*`); SHA pinning is reserved for third-party actions. The SHA-pinned checkout in `bundle-size.yml` is the outlier, not the rule. Pinning only the new lines would create the inverse inconsistency; flipping the convention is a repo-wide change and out of scope here.

.github/actions/detect-code-changes/action.yml
```diff
@@ -0,0 +1,40 @@
...
+    - id: filter
+      shell: bash
+      env:
+        GH_TOKEN: ${{ github.token }}
+        PR_NUMBER: ${{ github.event.pull_request.number }}
+        REPO: ${{ github.repository }}
+      run: |
+        if [ -z "$PR_NUMBER" ]; then
+          # Non-PR event (workflow_dispatch, push, etc.) — don't gate, just run.
+          echo "code=true" >> "$GITHUB_OUTPUT"
```
daledupreez, line 26: Is it worth adding a note when this happens so we know why we're in this branch?
```suggestion
          echo "No PR number defined; assuming we have code to check"
          echo "code=true" >> "$GITHUB_OUTPUT"
```

.github/workflows/js-tests.yml
```diff
@@ -1,11 +1,26 @@
 name: JS tests
 
 on:
-  pull_request
+  pull_request:
```
daledupreez: Is this trailing `:` necessary or expected?

daledupreez (approved):
This looks OK to me, and the workflows work for this branch. I think we can merge the PR and take on iterations in future PRs. If we have trouble merging any PRs, we can revert this PR and keep moving.

.github/workflows/js-tests.yml
wjrosa: Not necessary — `pull_request` and `pull_request:` are equivalent (the colon just makes an empty mapping); leftover from a planned `paths:` filter. Reverted here and in the other five. Left `bundle-size.yml` as-is since its `on:` is a mapping, so the colon's required there.

Drafted with assistance from Claude Code; reviewed and posted by me.

PR #5505, opened 2026-06-01
 Fix payment method title on Blocks order-received page with Adaptive Pricing

malithsen:
Fixes STRIPE-1095

With Adaptive Pricing on, the Blocks order-received page briefly shows a generic "Stripe" label as the payment method instead of the actual method the customer paid with (e.g. "Credit / Debit Card"). A refresh after the `checkout.session.completed` webhook lands corrects it.

 The bug comes from a constant collision: the Optimized Checkout pseudo-method shares the same underlying type value as the card method, so a defensive swap meant to support OC-style title resolution kicks in for every card payment when OC is enabled. On the eager title write (before the webhook has the full payment-method object) that swap routes the title through a generic  "Stripe" fallback. 

This PR removes the override. The card method handles plain cards and wallet brands correctly on its own, and the other call sites either pass the Stripe payment-method object or have already resolved the concrete type before they get here. So the override doesn't seem necessary in the first place.

## Testing instructions

1. Settings -> Payments -> Stripe: turn on Optimized Checkout and Adaptive Pricing.
2. Open a Blocks checkout, pay with a test card.
3. On the order-received page, confirm "Payment: Credit / Debit Card" renders immediately (no refresh needed).
4. Place a second order with a non-card method (e.g. Klarna) and confirm its title still renders correctly.
5. Open the order in WP Admin and confirm the stored payment method title matches.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa (commented):
Diff is tight, root cause (`WC_Stripe_Payment_Methods::OC === ::CARD === 'card'`) confirmed. One ask inline.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2789,11 +2789,6 @@ protected function is_adaptive_pricing_supported(): bool {
 	public function set_payment_method_title_for_order( $order, $payment_method_type, $stripe_payment_method = false ) {
```
wjrosa, line 2789: Per AGENTS.md ("Bugfixes for fatals, checkout failures, and payment regressions MUST include or update targeted automated tests"), this should land with a unit test pinning the new behaviour — e.g., `oc_enabled = true` + `payment_method_type = 'card'` writes the CC method's title, not `'Stripe'`. Without it, a future change that re-introduces an OC override on this path could silently regress the same `OC === CARD` collision.

Mayisha (approved):
Works as described 👍 

Image
Image

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
malithsen: Added a test in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5505/commits/70a6d9ceb43eae9a2a3e109bc0ef6d6ba47f1b71

PR #5507, opened 2026-06-02
Fix empty payment method on classic checkout when the UPE element re-mounts

Mayisha:
Fixes STRIPE-1186

On classic (shortcode) checkout, WooCommerce re-renders the payment box on `updated_checkout`, e.g. when an address change recalculates shipping — which tears down and **asynchronously** re-mounts the Stripe Payment Element. If the customer hits "Place Order" during that re-mount window, the submit runs against a not-yet-mounted element and `wc-stripe-payment-method` posts empty. The order fails with *"The payment could not be initialized. Please refresh the page and try again."* and the shopper has to refresh to recover — a real source of abandoned orders on stores that trigger `update_checkout`  frequently (custom address autocomplete, shipping-zone recalcs, etc.).

The submit path never waited for the in-flight re-mount, so this PR makes it do exactly that:

- **Track the full (re)mount as an awaitable `component.mountPromise`** — element creation → `mount()` → listener wiring → (Adaptive Pricing) `loadActions`. It's cleared in a `finally`, guarded so a newer remount's promise isn't wiped out by an older one settling.
- **Add `ensureUPEElementMounted()`** — awaits any in-flight remount, and if the element was torn down but not re-mounted yet, mounts it. No-ops when there's no element on the page (e.g. a 100%-off coupon removed the payment box).
- **Await it at the top of `processPayment`'s submit flow**, before reading the Elements instance. The form is already blocked at that point, so the wait sits under the existing spinner.

### Approach notes / trade-offs

I deliberately left the `! gatewayUPEComponents[ paymentMethodType ]` early-return in `processPayment` returning `undefined` (not `false`). The `#order_review` (Pay-for-Order) submit handler is bound to the raw `submit` event, so forcing `false` there would block non-Stripe gateways. Gating on the awaited remount avoids that regression while still closing the race. This is also why the fix lives on the submit path rather than disabling the Place Order button — the await covers both the "mid-remount" and "torn-down, not-yet-remounted" states without new UI state to manage.
  
Scope is limited to classic checkout (`client/classic/upe/`). Blocks uses a separate flow and is unaffected.

#### How can this break / what guards against it
  
- An await on submit could hang if a mount never settles — `mountPromise` always settles and is cleared in `finally`; a failed mount sets `hasLoadError`, which surfaces an inline error rather than hanging.
- Adaptive Pricing / Checkout Session submit branch is unchanged; covered by the existing tests in this file, which still pass.

  ## Testing instructions

The race is timing-sensitive and only occurs when WooCommerce actually re-renders the payment box (the payment fragment HTML must change between updates), so this is difficult to reproduce. 

Regression checks:

- Normal classic checkout with a new card completes as before.
- Pay-for-Order page with a non-Stripe gateway selected still submits.
- If you run Optimized Checkout / Adaptive Pricing, a standard checkout still completes.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha:
@coderabbitai review

wjrosa (commented):
Left a few short inline comments: one stale-DOM-node edge case in `ensureUPEElementMounted`, plus two smaller notes.

client/classic/upe/payment-processing.js
```diff
@@ -638,102 +650,165 @@ export async function mountStripePaymentElement( api, domElement ) {
...
-		showErrorPaymentMethod( loadActionsError, domElement );
+	// A (re)mount is already in flight — wait for it to settle.
+	if ( component.mountPromise ) {
+		await component.mountPromise;
 	}
 
-	return component;
+	// The element was torn down (its iframe removed) but a re-mount hasn't been
+	// kicked off yet. Mount it now and wait, so we don't submit an empty field.
+	if ( domElement.children.length === 0 ) {
```
wjrosa: `domElement` was captured before this await. If another `updated_checkout` re-render lands while we wait, the captured node is detached, this check passes (`children.length === 0`), and `upeElement.mount()` would pull the iframe out of the live on-page element into the detached node. Suggest re-querying the element after the await (and maybe looping while a newer `mountPromise` shows up).

client/classic/upe/payment-processing.js
```diff
@@ -638,102 +650,165 @@ export async function mountStripePaymentElement( api, domElement ) {
...
+ */
+export async function ensureUPEElementMounted( api, paymentMethodType ) {
+	const component = gatewayUPEComponents[ paymentMethodType ];
+	if ( ! component ) {
+		return;
+	}
+
+	const domElement = document.querySelector(
+		`.wc-stripe-upe-element[data-payment-method-type="${ paymentMethodType }"]`
 	);
```
wjrosa: Nit: this selector + `children.length` "is mounted" heuristic duplicates `getMountedUPEComponent` below — worth extracting a tiny shared helper so the two can't drift.

client/classic/upe/payment-processing.js
```diff
@@ -854,6 +929,10 @@ export const processPayment = (
 
 	( async () => {
 		try {
+			// The element may be mid-remount; reading its Elements instance now
+			// would post an empty payment method field and fail the payment.
+			await ensureUPEElementMounted( api, paymentMethodType );
```
wjrosa, line 1016: The new tests exercise `ensureUPEElementMounted` directly, but nothing covers this wiring — a test asserting `processPayment` doesn't read `elements`/submit until an in-flight `mountPromise` settles would lock in the actual regression being fixed.

Mayisha: Added in [e4b93a1](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5507/commits/e4b93a17b1f5c79b4f1f5dd56136d91950958c1d)

client/classic/upe/payment-processing.js
```diff
@@ -638,102 +650,165 @@ export async function mountStripePaymentElement( api, domElement ) {
 			upeElementPromise;
 	}
 
-	const upeElement = await upeElementPromise;
+	// Expose the full (re)mount as a single awaitable promise so processPayment
+	// can wait for a re-mounting element instead of posting an empty field.
+	const mountPromise = ( async () => {
```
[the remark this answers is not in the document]
Mayisha: Addressed in [e4b93a1](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5507/commits/e4b93a17b1f5c79b4f1f5dd56136d91950958c1d)

client/classic/upe/payment-processing.js
Mayisha: Done in [7f4faea](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5507/commits/7f4faea02d9f7c67ec6f8693935f99a46f69f497)

client/classic/upe/payment-processing.js
Mayisha: Done in [7f4faea](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5507/commits/7f4faea02d9f7c67ec6f8693935f99a46f69f497)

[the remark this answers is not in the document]
Mayisha: Addressed in [7f4faea](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5507/commits/7f4faea02d9f7c67ec6f8693935f99a46f69f497)

Mayisha:
@wjrosa I have addressed your feedback. Could I have another round of review on this? 👀

wjrosa (approved):
Looks good to me! Thanks for addressing the suggestions 👍

PR #5512, opened 2026-06-02
Fix Stripe checkout failing when the order is submitted while the Payment Element re-mounts (classic + Blocks)

### LINKED ISSUE #5490: Update_checkout reinitializes Stripe UPE widget - payment_method field empty on submission
Description
When update_checkout is triggered during checkout (e.g. after shipping address validation or zone recalculation), the Stripe UPE widget is destroyed and reinitialized. During this reinitialization window, if the customer clicks "Place Order", the wc-stripe-payment-method-upe field arrives empty in $_POST, causing the payment to fail.
The customer sees the message "The payment could not be initialized. Please refresh the page and try again." and must refresh the page to complete their order. This leads to abandoned orders.
This issue was originally reported in 2019 (#1093) for the classic Stripe card fields. The fix at the time (PR #927) handled iframe repopulation for classic fields but does not apply to the current UPE widget. The bug is therefore still present in 2026.
Steps to reproduce

Add a product to cart and go to the checkout page
Fill in the shipping address field — this triggers a shipping zone recalculation via update_checkout
Immediately after the address is validated (while the UPE widget is reinitializing), click "Place Order"
The wc-stripe-payment-method-upe field is empty in $_POST — payment fails

Reproduced on: Safari iOS 18.7, Chrome Android 148, Samsung Browser 29, Chrome desktop 148, Firefox 151
Expected behavior
The Stripe UPE widget should either persist its state through update_checkout re-renders without requiring full reinitialization, or prevent form submission until the widget has fully reinitialized and the payment_method ID is available.
Screenshots
POST data captured server-side during the bug:
{"payment_method":"stripe","wc-stripe-payment-method-upe":"","wc_stripe_selected_upe_payment_type":""}
System report
<details>
### WordPress Environment ###

WordPress address (URL): [Redacted]
Site address (URL): [Redacted]
WC Version: 10.7.0
Legacy REST API Package Version: Legacy REST API plugin is not installed on this site.
WP Version: 7.0
WP Memory Limit: 512 MB
WP Debug Mode: ✔
WP Cron: ✔
Language: fr_FR

### Server Environment ###

Server Info: Apache
PHP Version: 8.4.3
MySQL Version: 10.6.21-MariaDB

### Active Plugins (16) ###

Advanced Custom Fields: 6.8.1
Code Snippets: 3.9.6
LiteSpeed Cache: 7.8.1
OneSignal Push Notifications: 3.9.0
Super Progressive Web Apps: 2.2.45
TKB Chamon'Eat: 3.3.7
UpdraftPlus: 1.26.4
User Role Editor: 4.65
Modern Image Formats: 2.6.1
WooCommerce Stripe Gateway: 10.7.0
WooCommerce: 10.7.0
Wordfence Security: 8.2.2
WP Mail SMTP: 4.8.0
WPForms Lite: 1.10.0.5
WPS Hide Login: 1.9.18
Yoast Duplicate Post: 4.6

### Must Use Plugins ###

TKB Stripe Fix - payment_method vide: 1.3.6

### Theme ###

Parent Theme: Divi 4.27.6
Child Theme: Chamon'eat 1.0

Version: 10.7.0
Account ID: [REDACTED]
Test Mode Enabled: No
OAuth Connected: Yes
Optimized Checkout Enabled: No
Enabled Payment Methods: card
</details>
Additional context

Problem started appearing after updating to WooCommerce Stripe Gateway 10.5.x and persists through 10.7.0
Affects approximately 30-50% of mobile orders (Safari iOS, Chrome Android) and also reproducible on desktop
Workaround implemented server-side to recover orders for customers with saved payment tokens, but guest customers and customers entering a new card must refresh the page to complete their order
This is a timing/race condition: the UPE widget reinitializes after update_checkout but the customer can submit the form before reinitialization completes
The custom checkout includes a Google Maps address autocomplete that triggers update_checkout for shipping zone calculation — this makes the race condition more frequent than on standard checkouts
WordPress 7.0, PHP 8.4.3

wjrosa:
Fixes #5490
Fixes STRIPE-1202

> #5507 (classic, STRIPE-1186) is already merged into `develop`, so this PR's base is `develop` and the diff below is the delta on top of it. This consolidates the two classic guards that addressed the same race into a single mechanism, extends that coverage to the Adaptive Pricing pre-step, and adds the same protection to Blocks checkout. The classic files in the diff are an intentional refactor of #5507's guards — not leftovers.

## Background

When the Stripe Payment Element re-mounts after a checkout update (shipping / coupon / address change), a submission landing in that window can fail:

- **Classic (shortcode) checkout** — WooCommerce tears the payment box down via jQuery and re-mounts it asynchronously, so a submit reads a *detached* element and posts an **empty** `wc-stripe-payment-method`. The order fails and the shopper has to refresh.
- **Blocks checkout** — the element is React-managed and already blocks an empty-PM submission via `isPaymentElementComplete`, but during a brief re-mount (cart amount crossing the setup/payment `mode` threshold, or an Optimized Checkout payment-method-config change) a submit fails immediately with *"Your payment information is incomplete"* instead of waiting.

This PR closes the window on both surfaces by making submission wait for an in-flight (re)mount to settle.

### Classic checkout — consolidate the two guards into one mechanism

- **Single in-flight tracker (`mountInProgress` + `trackMountInProgress`)** — a module-scoped promise that composes overlapping `updated_checkout` cycles, so a later/faster re-mount can't release submission while an earlier one is still detaching/mounting.
- **`deferred-intent.js` registers the whole `updated_checkout` → re-mount chain**, so the wait also covers the Adaptive Pricing checkout-session update pre-step.
- **`ensureUPEElementMounted` rewired to the global tracker** — awaits `mountInProgress` **before** inspecting the DOM (the box can be mid-swap), then actively re-mounts an element torn down but not yet re-mounted. `processPayment` calls this once before reading the Elements instance.
- Removed the redundant per-component `mountPromise` bookkeeping from the merge.

### Blocks checkout — extend the protection

- Mirror the element's completion into a **live ref** (the `onPaymentSetup` closure otherwise captures a stale value) and, on submit, **wait briefly for an in-flight (re)mount to settle** before failing.
- Applied to **both Blocks render paths** — the deferred-intent `PaymentProcessor` and the Adaptive Pricing `CheckoutForm` / `usePaymentSetupHandler`.
- Shared, unit-tested helper `waitForPaymentElementCompletion` bounds the wait with a timeout, so a genuinely incomplete element still errors (no hang). BLIK (not driven by the Payment Element) is unaffected.

## Testing instructions

**Classic checkout**
1. Classic (shortcode) checkout with the Stripe card (UPE) method.
2. Add a shippable product, fill the card field, change the shipping address/country (triggers `update_checkout`), and immediately click **Place Order** while the payment area re-renders.
3. Before: intermittent *"The payment could not be initialized…"* with an empty PM server-side. After: the order submits cleanly once the element finishes re-mounting (wait sits under the existing spinner).

**Blocks checkout**
4. Blocks checkout with the Stripe card method; trigger a cart update that re-mounts the Payment Element (e.g. amount crossing the mode threshold, or an OC config change) and submit during the re-render.
5. Before: *"Your payment information is incomplete"*. After: the submit waits briefly and proceeds.

**Regression**
6. Normal orders (no churn) still complete on classic, Blocks, OC, and Adaptive Pricing; Pay-for-Order with a non-Stripe gateway still submits.

- [x] Covered with tests (Jest for classic and Blocks; classic E2E spec)

### Changelog entry

> Fix - Prevent classic and Blocks checkout submissions from failing while the Stripe Payment Element is re-mounting after a checkout update


client/classic/upe/payment-processing.js
```diff
@@ -54,6 +54,36 @@ import { handleDisplayOfSavingCheckbox } from 'wcstripe/optimized-checkout/handl
...
+/**
+ * Registers a promise that resolves when an in-flight Payment Element
+ * (re)mount completes. Submissions await the most recently registered promise.
+ *
+ * @param {Promise<*>} promise The mount promise to track.
+ */
+export function trackMountInProgress( promise ) {
+	// Swallow rejections here so awaiting it in processPayment never throws;
+	// mount errors are surfaced through their own error handling.
+	mountInProgress = Promise.resolve( promise ).catch( () => {} );
```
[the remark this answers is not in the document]
wjrosa: Fixed in 1e8fd8d — `trackMountInProgress` now composes with the existing tracker via `Promise.allSettled`, so submission waits for every active re-mount.

client/classic/upe/__tests__/payment-processing.test.js
```diff
@@ -265,6 +265,47 @@ describe( 'payment-processing', () => {
...
+				// Re-mount completes; submission proceeds normally.
+				resolveMount();
+				await flushPromises();
+
+				expect( api._standardElements.submit ).toHaveBeenCalled();
+				expect(
+					stripeUtils.appendPaymentMethodIdToForm
+				).toHaveBeenCalledWith( form, 'pm_test_123' );
+				expect( form.trigger ).toHaveBeenCalledWith( 'submit' );
+			} );
```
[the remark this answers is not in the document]
wjrosa: Added: new test resolves the newer of two overlapping re-mounts first and asserts submission stays held until the older one settles.

tests/phpunit/bootstrap.php
```diff
@@ -57,6 +57,10 @@ function _manually_load_plugin() {
 
 tests_add_filter( 'muplugins_loaded', '_manually_load_plugin' );
 
+// WooCommerce 10.9's EmailLogger adds an 'Email "..." sent.' order note after each
+// transactional email, breaking tests that assert on the latest note. Suppress it.
+tests_add_filter( 'woocommerce_email_log_add_order_note', '__return_false' );
```
malithsen: This is no longer needed after https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5546

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -1657,12 +1657,12 @@ public function provide_test_process_webhook_refund_updated() {
...
+				'expected note'       => '/Refund failed for <span class="woocommerce-Price-amount amount"><bdi( class="woocommerce-Price-bidi")?><span class="woocommerce-Price-currencySymbol"( translate="no")?>&#36;<\/span>10.00<\/bdi><\/span> - Refund ID: refund_123 - Reason: Unknown reason Order status changed from Pending payment to Processing\./',
 			],
 			'canceled refund'       => [
 				'notification status' => 'canceled',
 				'email triggered'     => true,
-				'expected note'       => '/Refund canceled for <span class="woocommerce-Price-amount amount"><bdi( class="woocommerce-Price-bidi")?><span class="woocommerce-Price-currencySymbol">&#36;<\/span>10.00<\/bdi><\/span> - Refund ID: refund_123 - Reason: Unknown reason Order status changed from Pending payment to Processing\./',
+				'expected note'       => '/Refund canceled for <span class="woocommerce-Price-amount amount"><bdi( class="woocommerce-Price-bidi")?><span class="woocommerce-Price-currencySymbol"( translate="no")?>&#36;<\/span>10.00<\/bdi><\/span> - Refund ID: refund_123 - Reason: Unknown reason Order status changed from Pending payment to Processing\./',
```
malithsen, line 1933: These are also handled in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5546

malithsen (commented):
I am not able to replicate this issue in develop branch.

In both classic/blocks, while the payment element re-mounts, "Place order" button is disabled, so there's no way to submit the order.

tests/phpunit/bootstrap.php
wjrosa: Removed in cee747f2b — the bootstrap is back to develop's state.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
wjrosa: Yep — after merging the base branch these lines converged with #5546's version, so they no longer show in the diff.

wjrosa:
@malithsen, the overlay and the re-mount don't overlap: WC core unblocks the payment box *before* `updated_checkout` fires, and the re-mount starts on that event, so the whole async mount runs with "Place Order" clickable. Locally, that window may be fast, which is why it's hard to hit (#5490 has production POST captures of the empty field). 

To repro on `develop`: 
- Run in console:
```
jQuery(document.body).on('updated_checkout', () => jQuery('#place_order').trigger('click'));
```
- Then change the shipping address to one with a different rate (the payment part must actually change, or the element is never destroyed).

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -113,6 +114,10 @@ const PaymentProcessor = ( {
 
 	const hasLoadErrorRef = useRef( false );
 
+	// Live mirror of isPaymentElementComplete so an in-flight onPaymentSetup
+	// callback sees updates, not its stale closure value. See #5490.
```
malithsen: 5490 is an issue. Maybe it'll be more helpful to comment on the exact context that will be helpful for the reader. If not, suggest dropping the issue reference altogether.

malithsen (approved):
@wjrosa Are you able to replicate the issue locally?

I still cannot replicate it. But from the Linear discussion, it seems like a build of this was shared with the merchant, and that seems to have fixed the issue for them. So I'm okay with merging this (after resolving conflicts) as the code makes sense.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
wjrosa: Good call — dropped the `#5490` ref and put the reasoning inline: `onPaymentSetup` registers its callback once, so it closes over the stale initial `isPaymentElementComplete`; the ref feeds it the live value. Updated in a6247fc6c.

wjrosa:
Thanks for the review, @malithsen! Not really 😅 . I just kept this work for the same reason: it solves the issue for the merchant. Still, I will wait for an additional review just to make sure

Mayisha (commented):
I am also unable to reproduce the original issue locally. I will go ahead with the code review.

Mayisha:
@wjrosa, quick question. Are the classic related changes expected here? Asking to confirm as the PR description says it is stacked on top of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5507 which addressed the issue in classic checkout.

wjrosa:
@Mayisha yes, the classic changes are intentional, not a leftover. #5507 has already been merged into `develop`, so this PR's base is `develop`, and the diff you see is the sum of both. I will update the PR title/description to avoid confusion

client/blocks/__tests__/wait-for-payment-element-completion.test.js
```diff
@@ -0,0 +1,58 @@
+import { waitForPaymentElementCompletion } from 'wcstripe/blocks/wait-for-payment-element-completion';
+
+// #5490: a re-mount resets the Payment Element's completion state; the helper
```
Mayisha: Nit: I would prefer not to keep the Linear ID in the comment.

Mayisha (approved):
Thanks for the fix. Checkout on classic and block checkout pages both are working well 👍 

Note: I was not able to reproduce the original issue locally.

client/blocks/__tests__/wait-for-payment-element-completion.test.js
wjrosa: Good call — dropped the issue ID from the comment here and from the other `#5490` references in the re-mount race tests (in acad6f936).

wjrosa: Removed it https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5512/changes/acad6f93611b63be3f043fe254a55801d1dccd22

PR #5515, opened 2026-06-03
Show express checkout appearance override notice only when an override is in effect

Mayisha:
Fixes STRIPE-1053

The "Some appearance settings may be overridden by the express payment section of the Cart & Checkout blocks" notice was rendered unconditionally on the **Express Checkout** (Apple Pay / Google Pay) and **Amazon Pay** settings pages. Merchants saw it even when nothing was actually being overridden, which is just noise.

That notice is only relevant when the Cart & Checkout blocks' Express Checkout section has the **"Apply uniform style"** option enabled.

This PR makes the notice conditional on a real override:
- Adds `WC_Stripe_Helper::is_express_checkout_button_style_overridden()`, which reads the Express Checkout block attributes from the Cart and Checkout pages (via WooCommerce's own `CartCheckoutUtils::find_express_checkout_attributes()`) and returns `true` only when `showButtonStyles` is enabled on either page.
- Passes the result to the two settings screens through their existing localized params (`is_button_style_overridden`).
- Gates the notice in both React sections on that flag and displays the notice only when the flag is true.

## Testing instructions
1. Go to **WooCommerce → Settings → Payments → Apple Pay / Google Pay** (and **Amazon Pay**). Confirm the appearance override notice is **not** shown by default.
2. Edit your **Block Checkout** page, select the **Express Checkout** block, and enable **"Apply uniform style"**. Save.

Screenshot 2026-06-03 at 1 52 16 PM

3. Reload the Express Checkout and Amazon Pay settings pages and verify that the notice **now appears**.

Screenshot 2026-06-03 at 2 57 17 PM

4. Uncheck **"Apply uniform style"** on **both** the Checkout and Cart pages and save. Reload the settings pages, the notice should be **gone**

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (approved):
The changes look good to me, and it tests well!

Just one minor detail, the Amazon Pay message has an extra: "Follow the instructions there and check back soon.", and I think we should change them to use the same text.

**Google Pay/Apple Pay**
Image

**Amazon Pay**
Image

includes/class-wc-stripe-helper.php
```diff
@@ -1113,6 +1114,35 @@ public static function has_cart_or_checkout_on_current_page() {
 		return is_cart() || is_checkout() || has_block( 'woocommerce/cart' ) || has_block( 'woocommerce/checkout' );
 	}
 
+	/**
+	 * Whether the express checkout button styles are overridden by the "Apply uniform style"
+	 * option of the Cart & Checkout blocks' Express Checkout section (`showButtonStyles`).
+	 *
+	 * @since 10.8.0
```
malithsen: If we are not including this in the 3.8.0 release, as the current milestone suggests, we'd want to bump this up

malithsen (approved):
Looks good!

PR #5517, opened 2026-06-03
Fix console errors on the Blocks checkout page editor

Mayisha:
Opening the Blocks checkout page in the editor logged two console errors. This PR fixes both at their source.

1. **`wc.wcSettings` accessed without declaring `wc-settings`**
The Stripe blocks scripts read `wc.wcSettings.getSetting('stripe_data')` through the raw `wc` global, so webpack never registered `wc-settings` as a script dependency. WooCommerce flagged this at runtime. Switched both `getBlocksConfiguration()` (`client/blocks/utils.js`) and `getStripeServerData()` (`client/stripe-utils/utils.js`) to import `getSetting` from `@woocommerce/settings` — the dependency-extraction plugin now adds `wc-settings` to the asset automatically.

Screenshot 2026-06-03 at 4 37 16 PM

2. **`ReferenceError` on `wc_stripe_upe_params`**
`getPermittedFontDomains()` accessed `wc_stripe_upe_params` directly, which throws when the global isn't defined (e.g. in the editor). Added an `undefined` guard before reading it.

Screenshot 2026-06-03 at 4 50 41 PM

## Testing instructions
- In wp-admin, go to the block checkout editor.
- Verify that the above two console errors/warnings are not present.
- As a shopper, verify that the Stripe payment element is rendered correctly on block and classic checkout pages.
- Complete a checkout and verify that there is no regression.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 xxxx-xx-xx - version 10.8.0
 * Tweak - Drop redundant "Test mode:" label from test payment instructions on Blocks checkout, which already shows a Test Mode badge
+* Fix - Remove console errors on the Blocks checkout page editor caused by accessing wc.wcSettings without declaring the wc-settings dependency and an undefined wc_stripe_upe_params global
 * Fix - Prevent Stripe from rendering an unexpected "Address Line 2" field inside the Payment Element
 * Update - Ensure payment method restrictions based on account and shopper countries are up to date
 * Fix - Store transaction IDs for orders when we get charges to ensure we can refund correctly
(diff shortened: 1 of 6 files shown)


changelog.txt
```diff
@@ -2,6 +2,7 @@
 
 xxxx-xx-xx - version 10.8.0
 * Tweak - Drop redundant "Test mode:" label from test payment instructions on Blocks checkout, which already shows a Test Mode badge
+* Fix - Remove console errors on the Blocks checkout page editor caused by accessing wc.wcSettings without declaring the wc-settings dependency and an undefined wc_stripe_upe_params global
```
malithsen: Nit: This feels a bit technical. Perhaps something like 
```suggestion
* Fix - Resolve console errors shown when editing the Blocks checkout page"
```

malithsen (approved):
Thanks for the fix. Tests well.

PR #5520, opened 2026-06-03
Preserve cached account data on a failed account refresh

diegocurbelo:
Fixes #<github_issue_id>

_NOTE:_ Extracted from #5383 so the account-cache resilience can land independently.

Reworks "Refresh account details" so a transient Stripe API failure (network error / outage, 5xx) leaves the previously cached account data intact instead of wiping it. Invalid credentials (401) still clear the cache and surface the "reconnect" notice.

## Testing instructions

1. Connect your Stripe test account.
2. Navigate to `WooCommerce → Settings → Payments → Stripe` and check that the Account details card shows real data.
3. Add this snippet to simulate a Stripe API outage:
    ```
        if ( false !== strpos( $url, 'api.stripe.com' ) ) {
            return new WP_Error( 'http_request_failed', 'Simulated Stripe outage' );
        }
        return $pre;
    }, 10, 3 );
    ```
4. Click the "3 dots" button and then `Refresh account details`
5. Verify the card keeps showing the previously loaded account data and not a "we couldn't connect" error.
6. Remove the snippet added in step 3
7. Replace the secret key with an invalid value:
     `wp option patch update woocommerce_stripe_settings test_secret_key 'sk_test_invalid'`
8. Click the "3 dots" button and then `Refresh account details`
9. Verify the card now shows the "we couldn't connect" error message:
    Screenshot 2026-06-16 at 21 18 18

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha (commented):
> 5. Verify the card keeps showing the previously loaded account data and not a "we couldn't connect" error.

At step 5, I see the "we couldn't connect" message. It is not working as described.

Image

daledupreez:
@diegocurbelo, could you move this PR back to draft if you have paused work on it?

diegocurbelo:
> > 5. Verify the card keeps showing the previously loaded account data and not a "we couldn't connect" error.
> 
> At step 5, I see the "we couldn't connect" message. It is not working as described.

Sorry about that @Mayisha, I extracted the testing instructions from the original PR (#5383), but this one has 2 test cases; keep the cache account data for api errors except for 401 errors, for those we still need to clear the cache so the connection notice is shown. I update the testing instruction to cover both flows.

I also added 2 new tests in [d2af73e](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5520/commits/d2af73e53bb4abbebd289b9dd724dacc61bae90c).

Mayisha (approved):
Thanks for the clarification @diegocurbelo. Looks good to me 👍

includes/admin/class-wc-rest-stripe-account-controller.php
```diff
@@ -158,12 +158,18 @@ public function get_webhook_status_message() {
...
+	 *
+	 * Forces a fresh fetch from Stripe and overwrites the cache only on success, so a transient
+	 * API failure leaves the previously cached account data intact instead of wiping it.
 	 *
 	 * @return WP_REST_Response
 	 */
 	public function refresh_account() {
-		$this->account->clear_cache();
+		delete_transient( WC_Stripe_Account::LIVE_WEBHOOK_STATUS_OPTION );
+		delete_transient( WC_Stripe_Account::TEST_WEBHOOK_STATUS_OPTION );
```
daledupreez: I am not a huge fan of the code replication for deleting these two transient values, especially given that the other code location is in `WC_Stripe_Account->clear_cache()`, so it would be really easy for the two blocks of code to drift apart. Might it be worth exposing a dedicated `clear_webhook_status_cache()` helper in `WC_Stripe_Account` so both locations can call the same code?

Separately, we may want to push these transients into the database cache, as we may be making unnecessary Stripe API calls for some merchants with sub-par transient implementations.

tests/phpunit/class-wc-stripe-account-test.php
```diff
@@ -96,6 +96,41 @@ public function test_clear_cache() {
...
+			]
+		);
+
+		// An invalid API key makes retrieve() return null (Stripe responds with a 401); the stale
+		// data must not survive so the UI can surface the "reconnect" prompt.
+		WC_Helper_Stripe_Api::$retrieve_response = null;
+
+		$this->assertEmpty( $this->account->get_cached_account_data( null, true ) );
+		$this->assertEmpty( $this->account->get_cached_account_data() );
+	}
```
daledupreez, line 132: We're not adding any tests that explicitly verify that the transients are cleared as part of calls to `refresh_account()` (and I don't see any existing unit tests for `refresh_account()`, so I think it may be worth adding tests for the code that we're changing in this PR.

daledupreez (commented):
I have some basic feedback from code inspection. I think it's worth working out how we should make changes before I test manually.

tests/phpunit/class-wc-stripe-account-test.php
diegocurbelo: Added new tests for the account controller in [69cc372](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5520/commits/69cc37202bf5cb5278dac416fde31cb078707c32)

includes/admin/class-wc-rest-stripe-account-controller.php
diegocurbelo: Makes sense, refactored in [69cc372](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5520/commits/69cc37202bf5cb5278dac416fde31cb078707c32)

Moving from transients to the cache is a good idea, but it's a bigger change, I'll open a follow-up PR as it requires changes in the webhook handler and settings page.

daledupreez (approved):
The changes look good to me and test as described.

PR #5523, opened 2026-06-04
Fix the Adaptive Pricing checkbox's disabled state when toggling Optimized Checkout Suite

Mayisha:
Similar to STRIPE-1090

This PR addresses two small UX issues with the adaptive pricing checkbox.

 1. **AP stayed disabled after enabling OCS (until refresh).** The AP "unavailable
     reason" was derived from `is_checkout_sessions_available()`, which returns
     `false` whenever OCS is off — so it was baked in as `'disabled'` at page load
     and never re-evaluated when OCS was ticked. It now derives only from
     account-level availability (`get_adaptive_pricing_account_unavailable_reason()`).
     The OCS on/off gating stays the React app's job (it already reacts to
     `isOCEnabled`), so the AP checkbox enables the moment OCS is enabled. No change
     to the shared `is_checkout_sessions_available()` feature flag and no
     checkout-time impact.

  2. **AP was stuck "checked but disabled" after disabling OCS.** When OCS was
     unchecked with AP already on, the AP checkbox went disabled but stayed
     visually checked — looking "on" with no way to click it off. The displayed
     `checked` state is now derived from the same condition that disables the
     checkbox, so a disabled AP checkbox always reads as unchecked. The saved value
     is intentionally left untouched (it's inert at checkout, since AP runtime-gates
     on OCS), so re-enabling OCS immediately restores the merchant's previous AP
     choice.

## Testing instructions
- At first make sure OCS and AP both checkboxes are disabled (refresh the page if you made these changes just now)
- Now enable the OCS checkbox.
- In `develop` notice that the AP checkbox is still disabled and you can not check/uncheck it. Refresh the page and then you'll be able to toggle the AP checkbox.

Screenshot 2026-06-04 at 4 54 25 PM

- In this branch, the AP checkbox should be unblocked to toggle.
- Now enable OCS and AP both and save the settings. 
- Now uncheck the OCS checkbox. 
- In `develop`, notice that AP checkbox is still checked but you can not toggle it anymore.

Screenshot 2026-06-04 at 4 55 03 PM

- In this branch, the AP checkbox should be unblocked to toggle.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -51,6 +51,7 @@ xxxx-xx-xx - version 10.8.0
 * Dev - Move additional classes to autoloader, including email and migration classes
 * Add - Register Stripe gateway capabilities with the WordPress Abilities API for agent access (default off; opt in via the `wc_stripe_abilities_enabled` filter)
 * Dev - Add e2e tests for subscriptions and Optimized Checkout Suite and support stricter payment confirmation checks
+ * Fix - Keep the Adaptive Pricing setting in sync with the Optimized Checkout Suite toggle without requiring a page refresh
 
 2026-05-12 - version 10.7.0
 * Fix - Hide the "move Stripe to the top" Optimized Checkout notice when all payment methods above Stripe are disabled
--- a/client/settings/advanced-settings-section/__tests__/optimized-checkout-feature.test.js
+++ b/client/settings/advanced-settings-section/__tests__/optimized-checkout-feature.test.js
@@ -116,10 +116,10 @@ describe( 'Optimized Checkout Element feature setting', () => {
 			'Let customers pay in their local currency with Adaptive Pricing'
 		);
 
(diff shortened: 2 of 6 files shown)


daledupreez:
@Mayisha, I think we should look at making the disabled state clearer rather than this approach. I intentionally avoided this approach because it makes it less clear what toggling OCS will do, because the Adaptive Pricing box toggles on and off without any clear reason.

Mayisha:
> I intentionally avoided this approach because it makes it less clear what toggling OCS will do, because the Adaptive Pricing box toggles on and off without any clear reason.

@daledupreez I am confused about this part 🤔 
We are not toggling AP checkbox here but changing the `disabled` state of the checkbox based on the OCS checkbox value. Shouldn't we allow enabling AP as soon as OCS is enabled instead of requiring a refresh? 🤔

PR #5524, opened 2026-06-04
Fix local Jest runs by ignoring the docker directory during module resolution

Mayisha:
 Running the JS unit suite locally (`npm run test:js`) crashed for every suite that imports `@wordpress/components`, with:
 ```
TypeError: (0 , _data.combineReducers) is not a function
```

Screenshot 2026-06-04 at 8 24 48 PM

`docker/` is already excluded from test discovery via `testPathIgnorePatterns`, but that does not affect module resolution. This PR adds the module-map equivalent:

  ```js
  modulePathIgnorePatterns: [ '<rootDir>/docker/' ],
```

so Jest only ever resolves modules from the project's own node_modules.

Test-tooling only — no plugin/runtime code changes.

## Testing instructions
- Code review.
- `npm run test:js` should pass all the tests locally.

### Changelog entry

<details>

#### Comment 

Internal test-tooling fix (Jest module resolution). No user-facing or runtime behavior change.

</details>

**Post merge**


daledupreez:
@Mayisha, I am not reproducing the issue in `develop`, though I can confirm that `docker/wordpress/wp-includes/js/dist/components.js` and `docker/wordpress/wp-includes/js/dist/components.min.js` exist in my local directory. Is there something else going on that's causing the issues with Jest?

Mayisha:
> @Mayisha, I am not reproducing the issue in develop, though I can confirm that docker/wordpress/wp-includes/js/dist/components.js and docker/wordpress/wp-includes/js/dist/components.min.js exist in my local directory. Is there something else going on that's causing the issues with Jest?

@daledupreez Good question — those components.js/components.min.js files in wp-includes/js/dist aren't actually the trigger.

I ran the issue through Claude and got this root cause for my case-
The real cause is that docker/ is gitignored, so its contents are populated per-machine and vary between us. In my local `docker/wordpress/wp-content/plugins/` there's a second plugin `woocommerce-gateway-paypal-powered-by-braintree` as I tend to use my Stripe local site for other plugins' testing. That carries a full node_modules tree with  `@wordpress/*` / `@woocommerce/*` packages. jest-haste-map crawls all of rootDir (including docker/) and indexes every package.json by name, so those collide with our own deps and Jest can resolve the wrong copy — hence combineReducers is not a function in my case.

Normal `node_modules` resolution walks up the tree and never descends into docker/, so the haste map is the only thing pulling those in. That's why testPathIgnorePatterns (test discovery only) doesn't fix it but modulePathIgnorePatterns (the module/haste map) does.

You can likely repro by checking your own checkout for find docker -path '*node_modules/@wordpress*' -name package.json | head, then clearing the Jest cache (npx jest --clearCache) before re-running — if your docker/ has no such plugin node_modules, the collision never forms.

daledupreez (approved):
Thanks for explaining the details. That makes sense and I think we can get this shipped as module resolution should _never_ be looking in the `docker` directory.

Mayisha:
Seems like I mistakenly shipped this change with [my another PR](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5515/changes#diff-75fdba962fd4ca17964b94eca4d8116a8364b30e9f21fadb2a89cc56f554ce88) 🤦‍♀️ 

Now this PR shows 0 file changes.

Screenshot 2026-06-09 at 8 40 23 PM

PR #5456, opened 2026-05-19
Add payment token classes to autoloader

daledupreez:
Towards STRIPE-504

## Changes proposed in this Pull Request:

This PR adds the payment token implementations in `includes/payment-tokens/` to the autoloader classmap, as their only dependencies are already in the classloader: `WC_Stripe_Fingerprint_Trait`, `WC_Stripe_Payment_Method_Comparison_Interface`, and `WC_Stripe_Payment_Methods`.

At a practical level, this PR adds the file names to the class map in `composer.json`, and removes the related `require_once`s for the files in `WC_Stripe->init()`.

While I was really hoping to get all the classes in this directory into the autoloader in this pass, the `WC_Stripe_Payment_Tokens` class depends on all the payment method classes via constant references from the `UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD` const and has some explicitly typed parameters, as well as a large number of runtime references to other classes.

## Testing instructions

Code inspection plus unit tests and e2e tests should be sufficient, especially given that our e2e tests include coverage of saving payment methods.

---

-   [N/A] Covered with tests (or have a good reason not to test in description ☝️)
-   [N/A] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5528, opened 2026-06-05
Fix duplicate payment tokens on My Account → Payment methods

diegocurbelo:
Fixes #<github_issue_id>

Loading `My Account → Payment methods` duplicates saved Link, Cash App Pay, SEPA, Amazon Pay and Bacs Direct Debit tokens in the `wp_woocommerce_payment_tokens` table (on every load).

https://github.com/user-attachments/assets/da3b9ec4-bee7-4dc0-b89a-37361e3faf7d

Root cause:
- `WC_Payment_Tokens::get()` builds each token's class as `WC_Payment_Token_{type}` from the lowercase Stripe type (e.g. `WC_Payment_Token_link`) and silently drops the token if `class_exists()` is false. 
- Our classes are in the case-sensitive Composer classmap under mixed-case names (`WC_Payment_Token_Link`, `..._CashApp`, `..._SEPA`, etc) and aren't loaded at bootstrap, so the lookup misses and WooCommerce drops the tokens from `get_tokens()`.
- The sync (`sync_and_retrieve_customer_payment_tokens`) then can't see them and recreates them on each load. 

`woocommerce_payment_token_class(...)` already maps CC/ACH/ACSS/BECS/Klarna; this PR adds the missing mapping for the five that were missing.

_NOTE:_ This stops **new duplicates** but doesn't remove rows already accumulated. A one-time dedupe migration (keep one row per user_id+token preserving any subscription-referenced token) will be handled in a follow-up PR.

## Testing instructions

1. Enable Cash App Pay
2. As a customer save a Cash App Pay payment method (either by manually adding it from My Account or by purchasing a subscription with it)
3. Navigate to `My account > Payment methods` and verify that the new payment method is shown
4. Reload the page a couple times
5. Check the table `wp_woocommerce_payment_tokens ` in the DB (the newest rows), in `develop` several duplicates are present (same `token`); but with this branch there is only one row for the newly added payment method:
    Screenshot 2026-06-05 at 03 31 31

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-tokens/class-wc-stripe-payment-tokens.php
+++ b/includes/payment-tokens/class-wc-stripe-payment-tokens.php
@@ -1060,37 +1060,28 @@ class WC_Stripe_Payment_Tokens {
 	 * @return string
 	 */
 	public function woocommerce_payment_token_class( $class, $type ) {
-		// WooCommerce derives the token class from the lowercase Stripe type ('WC_Payment_Token_' . $type).
-		// The case-sensitive Composer classmap can't resolve that against these mixed-case class names, so
-		// WC_Payment_Tokens::get() drops the tokens (and the sync then recreates them on every load). Map
-		// each type to its concrete class so the tokens resolve instead of being dropped.
+		// Replace WooCommerce's core credit card token with our own subclass.
 		if ( WC_Payment_Token_CC::class === $class ) {
 			return WC_Stripe_Payment_Token_CC::class;
 		}
-		if ( WC_Stripe_UPE_Payment_Method_ACH::STRIPE_ID === $type ) {
-			return WC_Payment_Token_ACH::class;
+
+		// WooCommerce derives the token class from the lowercase Stripe type ('WC_Payment_Token_' . $type),
+		// which the case-sensitive Composer classmap can't resolve for these mixed-case classes. Without an
+		// explicit mapping, WC_Payment_Tokens::get() drops the tokens and the sync recreates them on every load.
+		$token_classes = [
+			WC_Stripe_Payment_Methods::ACH         => WC_Payment_Token_ACH::class,
+			WC_Stripe_Payment_Methods::ACSS_DEBIT  => WC_Payment_Token_ACSS::class,
+			WC_Stripe_Payment_Methods::BECS_DEBIT  => WC_Payment_Token_Becs_Debit::class,
+			WC_Stripe_Payment_Methods::BACS_DEBIT  => WC_Payment_Token_Bacs_Debit::class,
+			WC_Stripe_Payment_Methods::LINK        => WC_Payment_Token_Link::class,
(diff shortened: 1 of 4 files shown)


daledupreez:
As a side note, the payment token classes were added to the autoloader classmap last week in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5456, so no migration should be necessary, as it only applies to development builds from the last week.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -1060,18 +1060,28 @@ public static function get_duplicate_token( $payment_method, $user_id, $gateway_
...
+		$token_classes = [
+			WC_Stripe_Payment_Methods::ACH         => WC_Payment_Token_ACH::class,
+			WC_Stripe_Payment_Methods::ACSS_DEBIT  => WC_Payment_Token_ACSS::class,
+			WC_Stripe_Payment_Methods::BECS_DEBIT  => WC_Payment_Token_Becs_Debit::class,
+			WC_Stripe_Payment_Methods::BACS_DEBIT  => WC_Payment_Token_Bacs_Debit::class,
+			WC_Stripe_Payment_Methods::LINK        => WC_Payment_Token_Link::class,
+			WC_Stripe_Payment_Methods::CASHAPP_PAY => WC_Payment_Token_CashApp::class,
+			WC_Stripe_Payment_Methods::SEPA        => WC_Payment_Token_SEPA::class,
+			WC_Stripe_Payment_Methods::AMAZON_PAY  => WC_Payment_Token_Amazon_Pay::class,
+		];
```
daledupreez, line 1080: I just took a look at `WC_Stripe_Payment_Tokens::UPE_REUSABLE_GATEWAYS_BY_PAYMENT_METHOD` and 
it feels like we may need a similar declarative mapping for Stripe ID => token class at a similar level so the two are more likely to stay in sync. (We may still need special handling for Klarna in this code, but that's relatively straightforward to handle.)

tests/phpunit/payment-tokens/class-wc-stripe-payment-tokens-test.php
```diff
@@ -487,6 +487,33 @@ public function provide_test_woocommerce_payment_token_class() {
 				'expected' => WC_Payment_Token_Becs_Debit::class,
 				'type'     => WC_Stripe_UPE_Payment_Method_Becs_Debit::STRIPE_ID,
 			],
+			// The cases below pass the lowercase class name WooCommerce derives from the token type
```
daledupreez, line 490: If we add the constant mapping for the Stripe ID => token class, we should update this test to iterate over that mapping, but I think that makes sense for a follow-up PR.

daledupreez (approved):
The changes look good and test well for me.

Mayisha (approved):
Changes look good to me. 💯

PR #5529, opened 2026-06-05
Don't default Adaptive Pricing on for India-based accounts at Connect OAuth

Mayisha:
Towards STRIPE-1176

We are defaulting Optimized Checkout and Adaptive Pricing on for new installs at Connect OAuth time. However, Adaptive Pricing is [not supported for India-based Stripe accounts](https://docs.stripe.com/payments/currencies/localize-prices/adaptive-pricing?payment-ui=embedded-components#restrictions), and the OAuth path enabled it unconditionally — so India-based accounts also got `adaptive_pricing = 'yes'`. 

This PR moves the default-on logic in `WC_Stripe_Connect::save_stripe_keys()` to after the new keys are saved — the account country can only be fetched with the new keys — and skips the Adaptive Pricing flip when the connected account is India-based. The account data is force-refreshed before the check so a stale cache from a previous connection can't influence the decision.

## Testing instructions
- Create a new JN site.
- Build the plugin from `release/10.8.0` and install the plugin.
- Connect to a IN Stripe account.
- Notice that the OCS and AP both got auto-enabled but the AP checkbox can not be toggled.

India frontbook 10 7

- Now build the plugin from this branch.
- Create another new JN site and install the build of this branch.
- Connect to a IN Stripe account.
- Notice that the OCS is auto-enabled and AP is disabled with the correct message.

Screenshot 2026-06-05 at 5 10 42 PM

- Regression test: ensure that the default-on feature in working correctly for non-IN accounts.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -47,7 +47,7 @@ xxxx-xx-xx - version 10.8.0
 * Update - Improve express checkout load times by preloading the resources
 * Tweak - Reduce JS bundle sizes by using native browser features instead of polyfills
 * Dev - Memoize the Express Checkout button visibility check within a request
-* Fix - Don't default Adaptive Pricing on when connecting a Stripe account it is unavailable for (e.g. India-based accounts)
+* Fix - Don't default Adaptive Pricing on when first connecting a Stripe account that is ineligible for Adaptive Pricing
 * Fix - Fix admin banner dismissal and display logic
 * Add - Detect Stripe API outages (network failures, timeouts, 5xx responses) and surface a wp-admin notice instead of crashing or showing misleading "couldn't connect" messages
 * Fix - Require a connected account for the target mode before switching between test and live
--- a/changelog.txt
+++ b/changelog.txt
@@ -47,6 +47,7 @@ xxxx-xx-xx - version 10.8.0
 * Update - Improve express checkout load times by preloading the resources
 * Tweak - Reduce JS bundle sizes by using native browser features instead of polyfills
 * Dev - Memoize the Express Checkout button visibility check within a request
+* Fix - Don't default Adaptive Pricing on when connecting a Stripe account it is unavailable for (e.g. India-based accounts)
 * Fix - Fix admin banner dismissal and display logic
 * Add - Detect Stripe API outages (network failures, timeouts, 5xx responses) and surface a wp-admin notice instead of crashing or showing misleading "couldn't connect" messages
 * Fix - Require a connected account for the target mode before switching between test and live
--- a/includes/connect/class-wc-stripe-connect.php
+++ b/includes/connect/class-wc-stripe-connect.php
@@ -215,13 +215,16 @@ if ( ! class_exists( 'WC_Stripe_Connect' ) ) {
 		}
 
 		/**
-		 * Helper function to clear some important PMC caches after a key update.
+		 * Helper function to clear some important caches after a key update.
 		 */
 		public function clear_caches_after_key_update(): void {
 			// Note that we also need to update the fallback PMC details, but we can't simply wipe that data.
 
 			// Clear PMC cache after key updates.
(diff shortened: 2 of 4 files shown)


includes/connect/class-wc-stripe-connect.php
```diff
@@ -296,6 +292,21 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 
 			$this->clear_caches_after_key_update();
 
+			// Runs after the keys are saved: the Adaptive Pricing decision needs the account country.
+			if ( 'connect' === $type && $should_default_optimized_checkout_on ) {
+				$account = WC_Stripe::get_instance()->account;
+				// Force refresh to avoid stale data from a previous connection.
+				$account->get_cached_account_data( $mode, true );
```
daledupreez: Rather than forcing a refresh here, might it not make sense to clear the account cache from `clear_caches_after_key_update()`? I think we have a related issue to do just that STRIPE-1176.

Mayisha: Good call @daledupreez. Updated in [b2a7de8](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5529/commits/b2a7de80f4d6500e07315870b64615bc519da7c4)

changelog.txt
```diff
@@ -47,6 +47,7 @@ xxxx-xx-xx - version 10.8.0
 * Update - Improve express checkout load times by preloading the resources
 * Tweak - Reduce JS bundle sizes by using native browser features instead of polyfills
 * Dev - Memoize the Express Checkout button visibility check within a request
+* Fix - Don't default Adaptive Pricing on when connecting a Stripe account it is unavailable for (e.g. India-based accounts)
```
daledupreez: I think we can simplify this text, and specify that it only applies to first connections/new accounts.
```suggestion
* Fix - Don't default Adaptive Pricing on when first connecting a Stripe account that is ineligible for Adaptive Pricing
```

daledupreez: Alternatively, we could be explicit about the check for India.
```suggestion
* Fix - Don't default Adaptive Pricing on when first connecting a Stripe account from India
```

includes/connect/class-wc-stripe-connect.php
```diff
@@ -296,6 +295,19 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 
 			$this->clear_caches_after_key_update();
 
+			// Runs after the keys are saved: the Adaptive Pricing decision needs the account country.
+			if ( 'connect' === $type && $should_default_optimized_checkout_on ) {
+				$account = WC_Stripe::get_instance()->account;
+
+				$options['optimized_checkout_element'] = 'yes';
+				if ( WC_Stripe_Country_Code::INDIA !== strtoupper( $account->get_account_country() ) ) {
```
malithsen: Should we use the existing [is_adaptive_pricing_available_for_account](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/82c3fb92a4e1dff166f1f061df7b98ff153aff39/includes/class-wc-stripe-helper.php#L1122) function here? It also verifies whether the settlement currency is compatible in live mode.

malithsen (commented):
Had a question after inspection.

I've not been able to create an IN account for testing, as it's restricted to invite-only.

daledupreez (approved):
The changes are looking good to me and test well.

I think we should tweak the changelog entry, and I also think we should add a `Towards STRIPE-1176` reference in the PR description to capture the related work.

includes/connect/class-wc-stripe-connect.php
Mayisha: Done.

changelog.txt
Mayisha: The first suggestion is more suitable now that we are checking `is_adaptive_pricing_available_for_account`

PR #5533, opened 2026-06-05
Update the order description and metadata on the payment intent after an Adaptive Pricing payment completes

Mayisha:
Adaptive Pricing routes through Stripe Checkout Sessions instead of a directly-created PaymentIntent. The session is created in includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php:26, and its build_payment_intent_data() (line 167) only passes metadata, no description. 
So, on the Stripe transaction page, the payment intent ID is set as the description. 
When we pay with the payment intent API, the description is set as `WooCommerce Stripe Dev - Order <order_id>`. But during the checkout session creation we do not have the order ID or any other order meta.

When we receive a checkout session completed event, we update the checkout session metadata with the order meta but the checkout session API does not take the `description` field. Only way to update the description is the payment intent API.

In this PR, I have changed the API call to use payment intents API instead of checkout session API to update the metadata when the checkout session completed event is received. Now we are able to update the description in the same API call.

## Testing instructions
- Enable OCS and Adaptive Pricing.
- As a shopper, place an order using card in your local currency.
- Wait for a while and check the order as an admin.
- Order should be in processing status.
- Check the corresponding transaction in Stripe dashboard.
- The transaction should include the description and all the metadata.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1860,6 +1860,8 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 						],
 					]
 				);
+			} else {
+				WC_Stripe_Logger::error( 'Empty intent ID, so cannot add order details and metadata.' );
 			}
 		} catch ( Exception $e ) {
 			WC_Stripe_Logger::error(
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1860,14 +1860,16 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 
 			// Schedule a job to store the order description and remaining metadata on the payment intent.
 			// The session is created from the cart before the order exists, so neither could be set at creation.
-			$this->action_scheduler_service->schedule_job(
-				time() + $this->process_payment_intent_metadata_delay,
-				$this->process_payment_intent_metadata_action,
-				[
-					'payment_intent_id' => $intent->id,
-					'order_id'          => $order->get_id(),
-				]
-			);
+			if ( ! empty( $intent_id ) ) {
+				$this->action_scheduler_service->schedule_job(
+					time() + $this->process_payment_intent_metadata_delay,
+					$this->process_payment_intent_metadata_action,
+					[
+						'payment_intent_id' => $intent_id,
+						'order_id'          => $order->get_id(),
+					]
+				);
+			}
 		} catch ( Exception $e ) {
 			WC_Stripe_Logger::error(
 				'Error processing checkout session for order: ' . $order->get_id(),
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -65,7 +65,7 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 
 	/**
 	 * The legacy Action Scheduler hook that updated checkout session metadata after a webhook.
-	 * Kept registered so jobs queued before the switch to {@see $process_payment_intent_metadata_action} still drain.
+	 * Kept registered so jobs queued before the switch to {@see $process_payment_intent_metadata_action} still run to completion.
 	 *
 	 * @deprecated 10.8.0 Replaced by {@see $process_payment_intent_metadata_action}.
 	 *
@@ -1572,15 +1572,15 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	/**
 	 * Processes the checkout session metadata update event to store additional metadata on the checkout session object.
 	 *
-	 * @deprecated 10.8.0 Replaced by {@see process_payment_intent_metadata()}; kept so jobs queued before the switch still drain.
+	 * @deprecated 10.8.0 Replaced by {@see process_payment_intent_metadata()}; kept so jobs queued before the switch still run to completion.
 	 *
 	 * @param string $checkout_session_id The checkout session ID.
 	 * @param array $metadata The metadata from the checkout session.
 	 * @return void
 	 */
 	public function process_checkout_session_metadata( string $checkout_session_id, array $metadata ): void {
-		wc_deprecated_function( __METHOD__, '10.8.0', 'process_payment_intent_metadata' );
-
+		// No runtime deprecation notice yet: legacy jobs queued before the 10.8.0 switch can still run in this
+		// release, and we don't want to emit notices for those. The notice will be added in a future release.
 		try {
 			$response = WC_Stripe_API::request( [ 'metadata' => $metadata ], 'checkout/sessions/' . $checkout_session_id, 'POST' );
 			if ( ! empty( $response->error->message ) ) {
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -50,14 +50,24 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	protected $deferred_webhook_action = 'wc_stripe_deferred_webhook';
 
 	/**
-	 * How long to wait before processing checkout session metadata after a webhook.
+	 * How long to wait before updating the payment intent description and metadata after a webhook.
 	 *
 	 * @var int
 	 */
-	protected $process_checkout_session_metadata_delay = 2 * MINUTE_IN_SECONDS;
+	protected $process_payment_intent_metadata_delay = 2 * MINUTE_IN_SECONDS;
 
 	/**
-	 * The Action Scheduler hook to use when processing checkout session metadata after a webhook.
+	 * The Action Scheduler hook to use when updating the payment intent description and metadata after a webhook.
+	 *
+	 * @var string
+	 */
+	protected $process_payment_intent_metadata_action = 'wc_stripe_process_payment_intent_metadata';
+
+	/**
+	 * The legacy Action Scheduler hook that updated checkout session metadata after a webhook.
+	 * Kept registered so jobs queued before the switch to {@see $process_payment_intent_metadata_action} still drain.
+	 *
+	 * @deprecated 10.8.0 Replaced by {@see $process_payment_intent_metadata_action}.
 	 *
 	 * @var string
 	 */
@@ -93,6 +103,7 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 		WC_Stripe_Webhook_State::get_monitoring_began_at();
 
 		add_action( $this->deferred_webhook_action, [ $this, 'process_deferred_webhook' ], 10, 3 );
+		add_action( $this->process_payment_intent_metadata_action, [ $this, 'process_payment_intent_metadata' ], 10, 2 );
 		add_action( $this->process_checkout_session_metadata_action, [ $this, 'process_checkout_session_metadata' ], 10, 2 );
 	}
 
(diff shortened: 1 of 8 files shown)


Mayisha:
@coderabbitai review

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -50,14 +50,24 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	protected $deferred_webhook_action = 'wc_stripe_deferred_webhook';
 
 	/**
-	 * How long to wait before processing checkout session metadata after a webhook.
+	 * How long to wait before updating the payment intent description and metadata after a webhook.
 	 *
 	 * @var int
 	 */
-	protected $process_checkout_session_metadata_delay = 2 * MINUTE_IN_SECONDS;
+	protected $process_payment_intent_metadata_delay = 2 * MINUTE_IN_SECONDS;
```
malithsen, line 57: Why is the delay necessary here, as opposed to immediately making the request?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1535,14 +1546,58 @@ protected function maybe_flag_unexpected_charge_on_order( WC_Order $order, strin
...
+		$request = [
+			/* translators: 1) blog name 2) order number */
+			'description' => sprintf( __( '%1$s - Order %2$s', 'woocommerce-gateway-stripe' ), wp_specialchars_decode( get_bloginfo( 'name' ), ENT_QUOTES ), $order->get_order_number() ),
+			'metadata'    => [
+				'order_id'   => $order->get_order_number(),
+				'order_key'  => $order->get_order_key(),
+				'signature'  => $this->get_order_signature( $order ),
+				'tax_amount' => WC_Stripe_Helper::get_stripe_amount( $order->get_total_tax(), strtolower( $order->get_currency() ) ),
+			],
+		];
```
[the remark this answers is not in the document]
Mayisha: Done.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1803,20 +1841,27 @@ protected function handle_checkout_session_success( object $notification ): void
...
+			// The session is created from the cart before the order exists, so neither could be set at creation.
+			if ( ! empty( $intent_id ) ) {
+				$this->action_scheduler_service->schedule_job(
+					time() + $this->process_payment_intent_metadata_delay,
+					$this->process_payment_intent_metadata_action,
+					[
+						'payment_intent_id' => $intent_id,
+						'request'           => [
+							/* translators: 1) blog name 2) order number */
+							'description' => sprintf( __( '%1$s - Order %2$s', 'woocommerce-gateway-stripe' ), wp_specialchars_decode( get_bloginfo( 'name' ), ENT_QUOTES ), $order->get_order_number() ),
```
malithsen: non-blocking nit: This description template exists in 2 other places [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/c104a2d3e6bfbda8225289a8cb16979b7e7f234c/includes/class-wc-stripe-intent-controller.php#L897) and [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/c104a2d3e6bfbda8225289a8cb16979b7e7f234c/includes/class-wc-stripe-intent-controller.php#L513). Would it be worth extracting it into a helper function? Something like `WC_Stripe_Helper::get_payment_intent_description( $order )` ?

malithsen (commented):
This works as expected. Stripe receives the payment intent requests with additional data, and the transaction gets updated. Left a question and a suggestion in the comments 

Image

includes/class-wc-stripe-webhook-handler.php
Mayisha: Done.

includes/class-wc-stripe-webhook-handler.php
Mayisha: I actually copied the value from the existing code. 
I asked AI about the benefit of this delay and got the following answer, also suggested to keep it-

`schedule_job()` unschedules identical pending jobs before scheduling, so duplicate webhook deliveries for the same session (retries, or async_payment_succeeded arriving after completed) within that window coalesce into a single API call. And since the payload is now snapshotted at scheduling time, the delay no longer costs anything in data freshness.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -50,14 +50,24 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
...
-	 * The Action Scheduler hook to use when processing checkout session metadata after a webhook.
+	 * The Action Scheduler hook to use when updating the payment intent description and metadata after a webhook.
+	 *
+	 * @var string
+	 */
+	protected $process_payment_intent_metadata_action = 'wc_stripe_process_payment_intent_metadata';
+
+	/**
+	 * The legacy Action Scheduler hook that updated checkout session metadata after a webhook.
+	 * Kept registered so jobs queued before the switch to {@see $process_payment_intent_metadata_action} still drain.
```
daledupreez: Nit: I think "drain" is not a very clear term here. I would prefer "run to completion" or "get picked up and run" instead.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1535,14 +1546,41 @@ protected function maybe_flag_unexpected_charge_on_order( WC_Order $order, strin
...
 	 * Processes the checkout session metadata update event to store additional metadata on the checkout session object.
 	 *
+	 * @deprecated 10.8.0 Replaced by {@see process_payment_intent_metadata()}; kept so jobs queued before the switch still drain.
+	 *
 	 * @param string $checkout_session_id The checkout session ID.
 	 * @param array $metadata The metadata from the checkout session.
 	 * @return void
 	 */
 	public function process_checkout_session_metadata( string $checkout_session_id, array $metadata ): void {
+		wc_deprecated_function( __METHOD__, '10.8.0', 'process_payment_intent_metadata' );
```
daledupreez: We are expecting this to run in some stores as part of this release, so I think we should avoid notices for that case, and only add the deprecation notice in the next release. (But I don't think this is blocking.)

includes/class-wc-stripe-webhook-handler.php
daledupreez: @Mayisha, it doesn't look like you pushed these changes up. Can you do get them pushed up for review?

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -1764,11 +1768,104 @@ public function test_process_checkout_session_metadata_api_error_response(): voi
...
+
+		$request_captured = false;
+		$pre_http_filter  = function ( $return_value, $parsed_args, $url ) use ( $payment_intent_id, $request, &$request_captured ) {
+			$expected_url = WC_Stripe_API::ENDPOINT . 'payment_intents/' . $payment_intent_id;
+			if ( $url !== $expected_url ) {
+				return $return_value;
+			}
+			$request_captured = true;
+			$this->assertEquals( 'POST', $parsed_args['method'] );
+			$this->assertEquals( $request['description'], $parsed_args['body']['description'] );
```
daledupreez, line 1795: Nit: Is it worth asserting that we have array values in `$parsed_args`?
```suggestion
			$this->assertIsArray( $parsed_args['body'] );
			$this->assertArrayKeyExists( $parsed_args['body']['description'] );
			$this->assertEquals( $request['description'], $parsed_args['body']['description'] );
```

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -2016,10 +2110,10 @@ public function test_process_async_checkout_session_success_for_on_hold_order():
 			->method( 'schedule_job' )
 			->with(
 				$this->isType( 'int' ),
-				'wc_stripe_process_checkout_session_metadata',
+				'wc_stripe_process_payment_intent_metadata',
 				$this->callback(
-					function ( $args ) use ( $checkout_session_id ) {
-						return isset( $args['checkout_session_id'] ) && $checkout_session_id === $args['checkout_session_id'];
+					function ( $args ) {
+						return isset( $args['payment_intent_id'] ) && isset( $args['request'] );
```
daledupreez: Nit: I think we should also ensure that `$args['request']` is an array.
```suggestion
						return isset( $args['payment_intent_id'] ) && isset( $args['request'] ) && is_array( $args['request'] );
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1850,8 +1850,7 @@ protected function handle_checkout_session_success( object $notification ): void
 					[
 						'payment_intent_id' => $intent_id,
 						'request'           => [
-							/* translators: 1) blog name 2) order number */
-							'description' => sprintf( __( '%1$s - Order %2$s', 'woocommerce-gateway-stripe' ), wp_specialchars_decode( get_bloginfo( 'name' ), ENT_QUOTES ), $order->get_order_number() ),
+							'description' => WC_Stripe_Helper::get_payment_intent_description( $order ),
 							'metadata'    => [
```
daledupreez, line 1854: Not for this PR, but it also feels like it would be worth adding a helper method to define the default metadata fields given the number of code locations that are computing the description. It feels like it would be _way_ too easy for the fields to drift apart.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -1796,29 +1892,28 @@ public function test_process_checkout_session_schedules_metadata_job(): void {
...
 		$scheduled_args = null;
 		$mock_scheduler->expects( $this->once() )
 			->method( 'schedule_job' )
 			->with(
 				$this->callback(
-					function ( $timestamp ) {
+					function ( $timestamp ) use ( $scheduled_at ) {
 						$this->assertIsInt( $timestamp, 'Expected timestamp to be an integer.' );
-						$this->assertGreaterThanOrEqual( time() + 2 * MINUTE_IN_SECONDS, $timestamp, 'Expected timestamp to be in the future.' );
+						$this->assertGreaterThanOrEqual( $scheduled_at + 2 * MINUTE_IN_SECONDS, $timestamp, 'Expected timestamp to be in the future.' );
```
daledupreez: Not blocking, but it may be worth using slightly more accurate variable names and require a ceiling on the time to ensure that is actually two minutes in the future and not some time after that.
```suggestion
						$current_time = time();
						$this->assertGreaterThanOrEqual( $test_start_time + 2 * MINUTE_IN_SECONDS, $timestamp, 'Expected timestamp to be at least 2 minutes in the future.' );
						$this->assertLessThanOrEqual( $current_time + 2 * MINUTE_IN_SECONDS, $timestamp, 'Expected timestamp to be at most 2 minutes in the future.' );
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1803,20 +1841,26 @@ protected function handle_checkout_session_success( object $notification ): void
...
-						'order_id'   => $order->get_order_number(),
-						'order_key'  => $order->get_order_key(),
-						'signature'  => $this->get_order_signature( $order ),
-						'tax_amount' => WC_Stripe_Helper::get_stripe_amount( $order->get_total_tax(), strtolower( $order->get_currency() ) ),
-					],
-				]
-			);
+			// Schedule a job to store the order description and remaining metadata on the payment intent.
+			// The session is created from the cart before the order exists, so neither could be set at creation.
+			if ( ! empty( $intent_id ) ) {
```
daledupreez, line 1846: When might `$intent_id` be empty? The previous logic would run in all cases, but we're now making the update conditional on the intent ID being known and non-empty. If the intent ID is empty, should we fall back on the old logic?

Mayisha: I added this as an additional safeguard. But if the code came to this point, the intent ID should always be present.

tests/phpunit/class-wc-stripe-webhook-handler-test.php
Mayisha: Done.

includes/class-wc-stripe-webhook-handler.php
daledupreez: Can we at least throw an exception or log an error when `$intent_id` is empty? While we don't think it should ever be empty, we should make that obvious and not fail silently.
```suggestion
			if ( empty( $intent_id ) ) {
				WC_Stripe_Logger::error( 'Empty intent ID, so cannot add order details and metadata.' ); 
			}
```

includes/class-wc-stripe-webhook-handler.php
Mayisha: Follow-up PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5538

daledupreez (approved):
This is working as described, though I think we should at least add an error or similar should we encounter an empty intent ID. And we should add a comment to mention that this is not something we should encounter.

Issue #4780, opened 2025-11-11
WooCommerce Stripe Gateway update not fully compatible with Quantity Min Max Control PRO plugin


goodtosew:
### Description

With the upgrade of WooCommerce Stripe Gateway to version 10.0.1 a conflict developed with what appears to be the Min Max Control PRO plugin. An order with decimal increments can be put in the cart, but on entering the address and postcode, the cart rounds off the order to a whole number. Therefore 0.25m of fabric etc can not be purchased when the Stripe plugin is activated. I currently deactivate the Stripe plugin and my website will still run but only using Paypal. I would like to get back to using Stripe and Paypal with the ability for customers to order decimals.

I have gone through Woo Support who recommended logging the error here. They indicated this error has occurred for others in the following thread, which depicts it much more clearly than I can.

[https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4030](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/4030>)

### Steps to reproduce

So I activated the following plugins designed for this:

* Min Max Control PRO
* Min Max Control – Min Max Quantity Step Control for Woocommerce
* Quantity plus minus button for woocommerce<br>However doing this meant I could add an increment of 0.25m to the cart but then it rounded to 1m once it hit the cart. This was earlier than previously but I will show you below.

This recording shows the issue in troubleshooting mode where only the plugins listed above are active:<br>[https://www.loom.com/share/4face9de0b9544eda12a9741684ac339](<https://www.loom.com/share/4face9de0b9544eda12a9741684ac339>)

This recording shows all plugins active in the live environment (the original error occurs here):<br>[https://www.loom.com/share/5977b6952fa04a9289fe6e11b04217d6](<https://www.loom.com/share/5977b6952fa04a9289fe6e11b04217d6>)

This is all my plugins activated except Woocommerce for Stripe. This is my workaround at the moment as customers can still pay for increments via paypal but I would love stripe back up and running on my site.<br>[https://www.loom.com/share/ac689e445c7f4bdd92853ae68723bee4](<https://www.loom.com/share/ac689e445c7f4bdd92853ae68723bee4>)

I suspect the Woocommerce Stripe plugin as I can activate all plugins on my live site and the error occurs, but if I just deactivate the Woocommerce Stripe plugin it is fixed.

I have gone through Woo Support who recommended logging the error here. They indicated this error has occurred for others in the following thread:

[...]

PR #5534, opened 2026-06-06
Preserve decimal quantities in legacy express-checkout add-to-cart (variable products)

kamlesh08:
## What this changes

Stripe's express checkout (Apple Pay / Google Pay) has two AJAX handlers that add a product to the cart. Both parsed the posted quantity with `absint()`, which quietly drops the decimal part — so on stores that allow fractional quantities (via the `woocommerce_stock_amount` filter), a quantity like `0.25` became `0`.

This updates both handlers to parse the quantity the way WooCommerce core normalises it: run the sanitized value (`wc_clean( wp_unslash( … ) )`) through `wc_format_decimal()` (so a localised decimal separator like `0,25` isn't lost) and then `wc_stock_amount()`, clamping the result to a non-negative value. In the product-data handler the clamp is re-applied after the `woocommerce_add_to_cart_quantity` filter so a third-party callback can't produce a negative preview total. Whole-number stores are unaffected.

> **Note:** an earlier revision of this branch also scaled the product-data tax preview by the selected quantity. That change has been split out into #5585 so this PR stays focused on quantity parsing — the tax-preview lines are unchanged here.

Files touched:
- `includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php` — `ajax_add_to_cart()` and `ajax_get_selected_product_data()`
- Regression tests — decimal preserved and non-positive clamped, for both handlers
- `changelog.txt` / `readme.txt`

## Scope — a focused, partial fix

The decimal-quantity fix is meaningful for **variable products**, which use the **legacy** add-to-cart path (`useLegacyCartEndpoints`) and can be configured for fractional quantities. The legacy path also serves bookings, but bookings don't use decimal quantities, so the change is a safe no-op for them. Simple products add to the cart through the WooCommerce **Store API** (`/wc/store/v1/cart/add-item`), which this PR intentionally leaves alone.

## How this relates to #4780

I found this while looking into #4780, but to be clear: **this PR is not a fix for that report.** That case involves a simple product, which adds to the cart through the WooCommerce **Store API** rather than this extension's code. I verified on WooCommerce 10.8.1 that the Store API preserves decimal quantities when the product's quantity step is exposed as a decimal (via the `woocommerce_store_api_product_quantity_multiple_of` filter), so the rounding seen there originates in the store's quantity-control plugin rather than in Stripe. This PR stands on its own as a small correctness fix for the legacy variable-product path.

## Testing

- **Unit tests:** four regression tests covering both handlers — decimal quantity preserved on a decimal-enabled store, and a non-positive quantity clamped to zero (not added to the cart / preview shows $0). The full `WC_Stripe_Express_Checkout_Ajax_Handler_Test` class passes locally; `phpcs` and `phpstan` are clean.
- **Manual:** on WordPress 7.0 / WooCommerce 10.8.1 with a decimal-quantity plugin active, calling `ajax_add_to_cart()` with `0.25` on a $10 product returns success with a $2.50 (250 minor-unit) total. Before the change, the same request left the cart empty.
- **Localized decimals:** with a comma decimal separator, `0,25` is normalised to `0.25` via `wc_format_decimal()` (a bare `(float)` cast would have zeroed it).

Thanks for reviewing!


includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -261,8 +273,20 @@ public function ajax_get_selected_product_data() {
...
+			$qty         = 1;
+			$cleaned_qty = isset( $_POST['qty'] ) ? wc_clean( wp_unslash( $_POST['qty'] ) ) : 1;
+			if ( is_string( $cleaned_qty ) ) {
+				$qty = max( 0, wc_stock_amount( (float) wc_format_decimal( $cleaned_qty ) ) );
+			}
+
+			// Re-clamp after the add-to-cart-quantity filter (a third-party callback could
+			// return a negative or non-numeric value) so the preview total can't go negative.
+			$filtered_qty = apply_filters( 'woocommerce_add_to_cart_quantity', $qty, $product_id );
+			$qty          = is_numeric( $filtered_qty ) ? max( 0, wc_stock_amount( (float) $filtered_qty ) ) : 0;
```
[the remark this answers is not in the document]
kamlesh08: I considered this but opted not to add it. The value passed into `woocommerce_add_to_cart_quantity` here is already normalized to a float (via `wc_format_decimal()` + `wc_stock_amount()` just above), and the filter's contract is to return a numeric quantity - WooCommerce core itself uses the filtered result directly as a quantity in `WC_Cart::add_to_cart()`. 

A callback returning a localized display string like "0,25" would be putting presentation formatting into an internal quantity filter, which core doesn't defend against either. The current handling (numeric preserved; negative/non-numeric clamped to 0) keeps the preview safe without over-fitting to a malformed filter return. 

Happy to revisit if a real-world filter does this.

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-ajax-handler-test.php
```diff
@@ -196,4 +196,228 @@ public function test_ajax_add_to_cart_returns_success_for_simple_product() {
...
+		// filter for a float one. This mirrors what such plugins actually do
+		// (remove_filter intval / add_filter floatval).
+		remove_filter( 'woocommerce_stock_amount', 'intval' );
+		add_filter( 'woocommerce_stock_amount', 'floatval' );
+
+		$product = WC_Helper_Product::create_simple_product();
+
+		$this->express_checkout_helper->expects( $this->once() )
+			->method( 'supported_product_types' )
+			->willReturn( [ ProductType::SIMPLE ] );
```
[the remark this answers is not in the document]
kamlesh08: In this handler, the quantity is parsed, normalized, and clamped before the simple/variable branch - the variable branch only resolves the variation via `find_matching_product_variation()` and swaps `$product`; it doesn't touch `$qty`. 

So the decimal handling is branch-agnostic and is already covered by the existing test, and a variable-product test here would mainly exercise WooCommerce's variation matching rather than the decimal logic this change touches. 

I kept the simple product coverage for that reason. 
If the team would prefer an explicit variable-product case for parity with the PR title, I'm glad to add one.

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -91,8 +91,18 @@ public function ajax_add_to_cart() {
 		try {
 
 			$product_id = isset( $_POST['product_id'] ) ? absint( $_POST['product_id'] ) : 0;
-			$qty        = ! isset( $_POST['qty'] ) ? 1 : absint( $_POST['qty'] );
-			$product    = wc_get_product( $product_id );
+			// Use wc_stock_amount() rather than absint() so stores that allow decimal
+			// quantities (e.g. selling fabric by the metre) keep fractional values like
+			// 0.25 instead of having them truncated to an integer. wc_format_decimal()
+			// first normalises localised decimal separators (e.g. "0,25") so the float
+			// cast can't zero them; max() keeps the result non-negative.
```
Mayisha:
```suggestion
			// wc_stock_amount() respects the store's decimal-quantity setting; wc_format_decimal()
			// normalises localised separators ("0,25") before the cast so fractions survive.
```
We can keep the code comment less verbose.

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -118,15 +128,17 @@ public function ajax_add_to_cart() {
 				}
 			}
 
+			// $qty can be a float on decimal-quantity stores. WC_Cart::add_to_cart() accepts
+			// fractional quantities at runtime even though its stub types $quantity as int.
```
Mayisha:
```suggestion
			// On the add_to_cart() calls below: $qty can be a float on decimal-quantity stores. 
			// WC_Cart::add_to_cart() accepts fractional quantities at runtime even though its stub types $quantity as int.
			// (hence the @phpstan-ignore).
```
I was confused at first how this code comment relates to the following if block. I think it would be better to make it clear that it is the reason of the phpstan ignore.

readme.txt
```diff
@@ -157,6 +157,8 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Add - Default Optimized Checkout Suite and Adaptive Pricing on for all eligible stores, with in-app notices announcing the change
 * Fix - Prevent saved Link, Cash App Pay, SEPA, Amazon Pay and Bacs Direct Debit payment methods from being duplicated when viewing My Account → Payment methods
 * Fix - Ensure the save-payment-method checkbox is unchecked when a non-reusable payment method is selected in Optimized Checkout
+* Fix - Preserve decimal product quantities for variable products when adding to the cart through Apple Pay / Google Pay express checkout
```
Mayisha:
```suggestion
* Fix - Preserve decimal product quantities for products when adding to the cart through Apple Pay / Google Pay express checkout
```

readme.txt
```diff
@@ -157,6 +157,8 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
...
+* Fix - Calculate the Apple Pay / Google Pay product tax preview on the selected quantity instead of a single unit
```
Mayisha: Can we move the tax related changes to it's own PR?

Mayisha (changes requested):

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
kamlesh08: Thank you for the feedback.

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
kamlesh08: Thank you for the feedback.

readme.txt
kamlesh08: Thank you.

readme.txt
kamlesh08: Sure, I'll create a new PR for this. Thank you.

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -202,15 +209,18 @@ public function ajax_add_to_cart() {
...
 				$attributes = wc_clean( wp_unslash( $_POST['attributes'] ) );
 
 				$data_store   = WC_Data_Store::load( 'product' );
 				$variation_id = $data_store->find_matching_product_variation( $product, $attributes );
 
-				WC()->cart->add_to_cart( $product->get_id(), $qty, $variation_id, $attributes );
+				WC()->cart->add_to_cart( $product->get_id(), $qty, $variation_id, $attributes ); // @phpstan-ignore argument.type
 			} elseif ( in_array( $product_type, $this->express_checkout_helper->supported_product_types(), true ) ) {
-				WC()->cart->add_to_cart( $product->get_id(), $qty );
+				WC()->cart->add_to_cart( $product->get_id(), $qty ); // @phpstan-ignore argument.type
```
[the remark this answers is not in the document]
kamlesh08: The identifier form here is intentional. This repo pins PHPStan 2.1.33, which supports `@phpstan-ignore <identifier>` - that's been the recommended form since 1.10, and it's narrower than `@phpstan-ignore-line`: it suppresses only the `argument.type` error on that line instead of silencing every error on the line.

It's also already an established pattern in the codebase, e.g., `@phpstan-ignore nullCoalesce.offset` in `class-wc-stripe-logger.php`, `@phpstan-ignore property.notFound` in `class-wc-stripe-account.php`, and `@phpstan-ignore return.void` in `class-wc-stripe-upe-payment-gateway.php`.

On "PHPStan may still report": the PHPStan Analysis check is green on this PR, so the suppression is taking effect as intended.

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
```diff
@@ -351,8 +361,20 @@ public function ajax_get_selected_product_data() {
...
+			$cleaned_qty = isset( $_POST['qty'] ) ? wc_clean( wp_unslash( $_POST['qty'] ) ) : 1;
+			if ( is_string( $cleaned_qty ) ) {
+				$qty = max( 0, wc_stock_amount( (float) wc_format_decimal( $cleaned_qty ) ) );
+			}
+
+			// Re-clamp after the add-to-cart-quantity filter (a third-party callback could
+			// return a negative or non-numeric value) so the preview total can't go negative.
+			$filtered_qty = apply_filters( 'woocommerce_add_to_cart_quantity', $qty, $product_id );
+			$qty          = is_numeric( $filtered_qty ) ? max( 0, wc_stock_amount( (float) $filtered_qty ) ) : 0;
+
```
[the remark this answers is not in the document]
kamlesh08: Correct, and the PR description was stale on this point, so I've updated it. An earlier revision of this branch did scale the tax preview by quantity, but that change was reverted here to keep this PR focused on quantity parsing; the diff now leaves `get_taxes_like_cart( $product, $price )` unchanged.

The tax-preview fix has moved to #5585 (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5585), which passes the quantity-adjusted line subtotal (`$qty * $price`) into `get_taxes_like_cart()`. 

It corrects both directions: for qty > 1, the single-unit base understated tax, and for a fractional qty like 0.25, it overstated it (a full unit's tax on a partial line).

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-ajax-handler-test.php
```diff
@@ -197,6 +197,223 @@ public function test_ajax_add_to_cart_returns_success_for_simple_product() {
...
+		remove_filter( 'woocommerce_stock_amount', 'intval' );
+		add_filter( 'woocommerce_stock_amount', 'floatval' );
+
+		$product = WC_Helper_Product::create_simple_product();
+
+		$this->express_checkout_helper->method( 'is_invalid_subscription_product' )->willReturn( false );
+		$this->express_checkout_helper->method( 'get_product_price' )->willReturn( 10.0 );
+		$this->express_checkout_helper->method( 'get_taxes_like_cart' )->willReturn( [] );
+		$this->express_checkout_helper->method( 'get_total_label' )->willReturn( 'Total' );
+
```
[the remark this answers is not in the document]
kamlesh08: In this PR the `get_taxes_like_cart()` stub returns `[]` on purpose: #5534 doesn't change tax behavior, so these tests assert only that the decimal quantity survives into the price math (the `displayItems` amount).

The test lives in #5585 (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5585) `test_ajax_get_selected_product_data_scales_tax_by_quantity`.

It stubs a non-zero tax (`[6.0]`), asserts `get_taxes_like_cart()` is called with the quantity-adjusted subtotal (`30.0` for qty 3 × $10), and asserts the previewed tax line (`600`) and total (`3600`) scale with `$qty`.

readme.txt
kamlesh08: I have created a separate PR for tax-related changes now.

kamlesh08:
PR for tax-related changes here: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5585.

@Mayisha, can you please re-review this and let me know if any changes are needed?

Mayisha (approved):
Thank you @kamlesh08 for fixing this. Changes looks good to me 👍 

**Before with 1.5 quantity**

Image

**After with 1.5 quantity**
Image

PR #5522, opened 2026-06-04
Add Stripe benchmark fixture bootstrap

chubes4:
## Summary
- Adds a reusable `tests/benchmarks/fixture-bootstrap.php` helper for plain WordPress/WP Codebox benchmark workloads outside PHPUnit.
- Bootstraps WooCommerce install state, cart/checkout pages, flat-rate shipping, a purchasable product, cart/session state, and Stripe test/ECE settings before benchmark timing starts.
- Adds preflight assertions with fixture-specific failure messages plus a short usage README.

Fixes #5521.

## Follow-up use case
- #5530 uses this fixture to reproduce and measure product-page Express Checkout Element network fan-out while working on #1439.

## Testing
- `php -l tests/benchmarks/fixture-bootstrap.php`
- `./vendor/bin/phpcs --standard=phpcs.xml.dist -n tests/benchmarks/fixture-bootstrap.php`
- `./vendor/bin/phpstan analyse --configuration=phpstan.neon --memory-limit=2G tests/benchmarks/fixture-bootstrap.php`

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.5)
- **Used for:** Drafting the benchmark fixture helper and documentation; Chris remains responsible for review and testing.

PR #5535, opened 2026-06-06
Reserve product Express Checkout space before render

### LINKED ISSUE #4378: Express Checkout is bad for CLS (Cumulative Layout Shift)
### Description

The way we have express checkout configured at the moment, we are increasing the [Cumulative Layout Shift](https://developer.mozilla.org/en-US/docs/Glossary/CLS) for pages that include express checkout. The core issue at the moment is that we dynamically load the express checkout code in the client, but we don't allocate any space to the elements, so when the elements eventually load, they move the rest of the page down. This is a poor UX, and is treated as such in Core Web Vitals.

### Steps to reproduce

1. Enable Express Checkout on your site with at least one accepted payment method (i.e. Link, Apple Pay, or Google Pay)
2. Navigate to a product page where express checkout should be shown
3. Note the page layout shifts when the express checkout buttons are displayed
4. Add the product to your cart and proceed to checkout
5. Note the checkout page

### Expected behavior

We should aim to set aside some placeholder space for express checkout, or possibly show some set of loading elements. Mostly we don't want to shift the page layout unnecessarily.

#### Product page

https://github.com/user-attachments/assets/565cdde1-285a-4227-bf8f-4dc3bf944efa

#### Checkout

https://github.com/user-attachments/assets/b7d88f6b-8870-4aae-b1cc-f87d09c288b3

### Environment

- WordPress Version - 6.8.1
- WooCommerce Version - 9.8.5
- Stripe Plugin Version - 9.5.2 (with some development changes for 9.6.0)

chubes4:
## Summary
- Reserves product-page Express Checkout space before Stripe finishes rendering dynamic buttons.
- Clears the reserved space again if Stripe reports no available payment methods or the element load fails.
- Adds a minimum slot height matching the default Express Checkout button height so visible content does not jump when the button appears.

Closes #4378.

## Verification
- `PATH="$HOME/.nvm/versions/node/v20.18.1/bin:$PATH" npx eslint client/entrypoints/express-checkout/index.js`
- `PATH="$HOME/.nvm/versions/node/v20.18.1/bin:$PATH" npm run test:js -- client/express-checkout/utils/__tests__/index.test.js --runInBand`
- `PATH="$HOME/.nvm/versions/node/v20.18.1/bin:$PATH" npm run build:webpack`
- Push hook: `npm run phpstan`

Existing warnings only:
- Browserslist/caniuse-lite staleness warning.
- Existing webpack asset/entrypoint size warnings.

## Layout-shift evidence
Homeboy Rigs deterministic CLS profiles reproduced the reported mechanism before this production fix:

| Profile | CLS | Layout shifts | Visible button |
| --- | ---: | ---: | --- |
| `simulated-cls` | `0.0351` | `1` | yes |
| `simulated-cls-reserved` | `0` | `0` | yes |

Those traces prove the underlying issue shape from #4378: inserting a visible Express Checkout button without reserved space causes CLS, while reserving the same final button space prevents the layout shift.

Note: running the Homeboy rig directly against this branch currently requires the benchmark fixture from #5522, which is still open. The production change here stays focused on the CLS fix rather than bundling the benchmark fixture into this PR.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (openai/gpt-5.5)
- **Used for:** Drafted the product-page reserved-space implementation, ran targeted checks, and summarized the deterministic CLS evidence; Chris remains responsible for review and acceptance.

### DIFF
--- a/client/entrypoints/express-checkout/styles.scss
+++ b/client/entrypoints/express-checkout/styles.scss
@@ -9,8 +9,13 @@
 	flex-wrap: wrap;
 	justify-content: center;
 
+	&.wc-stripe-express-checkout-element--reserved-space {
+		min-height: 48px;
+	}
+
 	> div {
 		flex: 1;
 		min-width: 260px;
+		min-height: 48px;
 	}
 }
--- a/client/entrypoints/express-checkout/index.js
+++ b/client/entrypoints/express-checkout/index.js
@@ -70,6 +70,8 @@ jQuery( function ( $ ) {
 	);
 
 	let wcStripeECEError = '';
+	const productPageReservedSpaceClass =
+		'wc-stripe-express-checkout-element--reserved-space';
 	const defaultErrorMessage = __(
 		'There was an error getting the product information.',
 		'woocommerce-gateway-stripe'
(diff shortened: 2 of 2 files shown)


client/entrypoints/express-checkout/styles.scss
```diff
@@ -11,8 +9,13 @@
 	flex-wrap: wrap;
 	justify-content: center;
 
+	&.wc-stripe-express-checkout-element--reserved-space {
+		min-height: 48px;
```
diegocurbelo, line 13: The configured button height can be `40px`, `48px`, or `55px` after client-side clamping. A fixed `48px` reservation still shifts for the large size and adds empty space for the small size. 

Also, the > div rule affects the cart and checkout

daledupreez: Might it be worth using a CSS variable for the height? It feels like that's something we could inject into the page when needed. e.g. this would become something like `min-height: var( --wc-stripe-express-checkout-button-height );`

chubes4: Thanks y'all. About to address the merge conflicts and get this PR refreshed and re-tested.

PR #5539, opened 2026-06-08
Try add admin notices for OCS and AP default-on

daledupreez:
Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492

This PR explores using general admin notices for alerting merchants to the fact that we have defaulted on Adaptive Pricing and/or Optimized Checkout Suite. The code in this PR builds on the machinery introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5492, but uses admin notices across WooCommerce to present the same copy that was introduced in the original PR, and removes the Javascript data that was used to show the React banners. In addition, the PR hides the "Review settings" link/CTA when we are already on the Stripe settings page.

In support of the above changes, this PR adds the Stripe SVG logo to the `assets/images/` folder to allow the image to be loaded from wp-admin.

In addition to the new banner rendering logic, this PR also adds some implementation details to the code responsible for rendering admin notices:
 * An image may be specified as part of the notice `message` field, and that `img` tag may specify `src`, `alt`, and `style` attributes.
 * A new `css_rules` array may be specified for a notice -- this array of strings will be injected within a `style` element inline just before the wrapping notice `div` is written to the page

### Optimized Checkout Suite and Adaptive Pricing notice

Screenshot 2026-06-08 at 19 55 50

### Only Adaptive Pricing notice

Screenshot 2026-06-08 at 19 55 10

### Only Optimized Checkout Suite notice

Screenshot 2026-06-08 at 19 54 08

## Testing instructions

Run this build and set up the Stripe plugin. For each of the following commands, verify that you see the relevant banners, can dismiss the item, and that :
 * Mimic showing the OCS and AP notice by running `wp option update wc_stripe_show_ocs_ap_banner yes` and ensuring that both OCS and AP are enabled - visiting any WooCommerce admin screen should then show the notice
    - Click on the "Learn more" button - verify that you're taken to the WooCommerce page for 10.8.0 in a new tab/window
    - Click on the "Review settings" button - verify that you're taken to the settings page, and that the notice is displayed without the "Review settings" button
    - Dismiss the notice from the UI, and verify that it is hidden
 * Show the AP-only notice by running `wp option update wc_stripe_show_ap_only_banner yes`, and ensuring that OCS and AP are enabled - visiting any WooCommerce admin screen should then show the notice
    - Dismiss the notice from the UI, and verify that it is hidden
 * Show the OCS-only notice by running `wp option update wc_stripe_show_ocs_only_banner yes`, and ensuring that OCS is enabled and AP is disabled - visiting any WooCommerce admin screen should then show the notice
    - Dismiss the notice from the UI, and verify that it is hidden

---

### Changelog entry

<details>

#### Comment 
This PR reworks the banners introduced in an unreleased PR. No new changelog necessary.
</details>

**Post merge**


tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
```diff
@@ -1306,4 +1306,242 @@ public function test_couldnt_connect_notice_is_suppressed_during_outage() {
...
+			'priority: OCS+AP wins over AP-only'      => [
+				true,
+				'yes',
+				'US',
+				[
+					'wc_stripe_show_ocs_ap_banner'  => 'yes',
+					'wc_stripe_show_ap_only_banner' => 'yes',
+				],
+				'ocs_ap_banner',
+			],
```
malithsen, line 1465: nit: This case shouldn't be possible. Flags should be mutually exclusive, but it doesn't hurt to test priority.

malithsen (commented):
We should either remove _old_ Settings page promotional banners or suppress these admin notices in the settings page to avoid duplicates

Image

tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
daledupreez: Yes, this is just to tripe-check.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -483,6 +510,87 @@ public function check_express_checkout_location(): void {
...
+
+		$review_action = sprintf(
+			'<a href="%s" class="button button-primary" style="margin-top:1em;margin-right:2em;">%s</a>',
+			$this->get_setting_link(),
+			esc_html__( 'Review settings', 'woocommerce-gateway-stripe' )
+		);
+
+		$actions = [ $review_action, $learn_more_action ];
+
+		if ( $is_oc_enabled && $is_ap_enabled && ! $is_india && 'yes' === get_option( self::SHOW_OCS_AP_BANNER_OPTION, 'no' ) ) {
```
malithsen, line 574: Do we need the country check here? It's already captured within the migration flags.

If a runtime check is needed, should we use the existing [helper method](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/3ff5aa48e4b4dff294f69adb9be62ce8d06a8bdc/includes/class-wc-stripe-helper.php#L1168), which already captures the country requirement + currency requirements?

malithsen (commented):
Can we align the dismiss icon with the title? Similar to other admin notices.

Image

includes/admin/class-wc-stripe-admin-notices.php
malithsen: nvm, we use the exact same checks for React banners. So these checks are consistent with the existing implementation

malithsen:
> Can we align the dismiss icon with the title? Similar to other admin notices.

This is a result of the styling [here](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/36fe8dd1fa874a918f9de85de56d94d1e3fce0ad/includes/admin/class-wc-stripe-admin-notices.php#L132), which we are adding for all Stripe originated admin notices, so the issue is outside the scope of the PR.

But I wonder if we should still fix it, given the high visibility of these notices.

malithsen:
Can we hide the "Review settings" button when on the Settings page, as it'll be a self-link?

CleanShot 2026-06-08 at 09 52 52@2x

malithsen (approved):
This looks good to me and tests well. My previous comments on the notices (dismiss button alignment, "Review settings" button visibility) are non-blocking nits. I ran through a subset of the test matrix, and this works as expected. We'd still want to clean up the failing and outdated unit tests before the merge.

PR #5540, opened 2026-06-09
Use shared hook manager to prevent duplicate subscription hooks

daledupreez:
Fixes STRIPE-1132
Fixes STRIPE-1129

Subscription hooks were previously guarded by a `static` flag on `WC_Stripe_Subscriptions_Trait`. However, because the trait is composed into every downstream payment method class as well as the main UPE gateway, and the Optimized Checkout payment method also uses the same `stripe` ID as the UPE gateway, the static variable didn't apply across downstream classes. This had the knock-on effect of making duplicate registrations possible, and was a direct cause for the duplicate subscriptions renewal incident in 10.6.0, which were initially addressed and worked around in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5325.

This PR refactors the registration code in three ways:
 * Registration of hooks specific to each payment method has been moved into a new `maybe_register_payment_method_hooks()` function
 * Registration of hooks that are general to the plugin has been moved into a new `maybe_register_plugin_hooks()` method on the trait, but the underlying code has actually been moved to a new `WC_Stripe_UPE_Payment_Gateway->maybe_register_gateway_subscription_hooks()` function, as that code should be the responsibility of the gateway, not the trait, especially given that the trait is grafted onto payment method classes.
 * Both methods use a shared `WC_Stripe_Hook_Manager` singleton that manages shared registration state for both payment method hooks and plugin hooks via `are_plugin_hooks_registered()`, `register_plugin_hooks()`, `are_payment_method_hooks_registered()`, and `register_payment_method_hooks()` methods.
   - It's worth noting that I've introduced a new `WC_Stripe_Hook_Categories` constant class and a `SUBSCRIPTIONS` constant, as the manager accepts a category argument to allow groups of hooks to be handled independently.
      * We can definitely consider using helper methods rather than constants for this aspect -- this was just closer to code we already have
   - It's also worth noting that the approach involves two key elements that are worth being explicit about:
     * The first registration wins, so we need to be careful about that for the `stripe` payment method ID
     * There is no wrapped code or indirection -- we _want_ explicit `add_filter()` and `add_action()` calls to be findable, so this code allows for simple guards around that registration

**How can this break?** If a non-gateway instance (e.g. the OC payment method) registers the shared `stripe` renewal hook before the gateway, things could behave unexpectedly. The gateway is constructed during `WC_Stripe::init()` and the OC method only lazily afterwards, so the gateway should always win.

## Testing instructions

1. Enable WooCommerce Subscriptions and the Stripe gateway (test mode), and ensure that Optimized Checkout Suite is enabled.
2. Create a subscription product.
3. Buy a subscription product with a test card.
4. Trigger a renewal for that subscription via wp-admin (WooCommerce → Subscriptions → open the subscription → **Process renewal** in the "Order actions" box and then hit the "Update" button).
5. Confirm the renewal is processed exactly once and the subscription status updates correctly.
6. Inspect the renewal order, and verify that the total for the order is correct, including the Stripe fees and the order notes. (If there was a double renewal, the Stripe amount would be too high.)
7. Disable the Optimized Checkout Suite and confirm that you can purchase a test subscription and renew via wp-admin as above.

---

### Changelog entry

**Post merge**


includes/class-wc-stripe-hook-manager.php
```diff
@@ -0,0 +1,134 @@
...
+
+	/**
+	 * Check if the payment method ID is valid.
+	 *
+	 * @param string $payment_method_id The payment method ID to check.
+	 * @return bool True if the payment method ID is valid, false otherwise.
+	 */
+	public function is_valid_payment_method_id( string $payment_method_id ): bool {
+		return 'stripe' === $payment_method_id || str_starts_with( $payment_method_id, 'stripe_' );
+	}
```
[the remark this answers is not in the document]
daledupreez: This is a non-issue: WordPress includes a shim for `str_starts_with()`, so it can be used safely on PHP 7.4.

includes/constants/class-wc-stripe-hook-categories.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+/**
+ * Class WC_Stripe_Hook_Categories
+ */
```
diegocurbelo, line 11:
```suggestion
/**
 * Class WC_Stripe_Hook_Categories
 * 
 * @since 10.9.0
 */
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -50,19 +41,43 @@ public function maybe_init_subscriptions() {
...
+	 * These are registered once per payment method.
+	 *
+	 * @param string $payment_method_id The payment method ID to register the hooks for.
+	 * @return void
+	 */
+	private function maybe_register_payment_method_hooks( string $payment_method_id ): void {
+		// Ensure we register the subscription hooks only once per payment method.
+		$hook_manager = WC_Stripe_Hook_Manager::get_instance();
+		if ( ! $hook_manager->is_valid_payment_method_id( $payment_method_id ) ) {
 			return;
```
diegocurbelo, line 66: We might want to lo an error if for some reason (maybe merchant custom code?) the trait is used in an unexpected payment method class (non `stripe`/`stripe_*`):
```suggestion
			WC_Stripe_Logger::error(
						'Skipping subscription hook registration: payment method ID is not a valid Stripe gateway payment method. Renewals will not be processed for this method.',
						[
									'payment_method_id' => $payment_method_id,
									'class'             => get_class( $this ),
						]
			);
			return;
```

diegocurbelo (approved):
The changes look good to me, and it tests well 👍🏼 

Added two suggestions, but they are nits not blockers.

includes/constants/class-wc-stripe-hook-categories.php
daledupreez: I added `@since` tags to both the class and new constant in a6d1c30bf6983410124c9df0d7278da37dcd54a4.

includes/compat/trait-wc-stripe-subscriptions.php
daledupreez: That's a really good idea -- it means we should see unexpected situations like this report explicit errors rather than fail silently as a result of the refactor.

Added in 9376964068dfaad775689b25dc6854250245b825.

PR #5542, opened 2026-06-09
Send the billing address to Stripe for "Pay for Order" orders

malithsen:
Fixes STRIPE-952

On the Pay for Order page, the billing address was not sent to Stripe when creating the PaymentMethod. This can result in failed orders if there are Radar rules that check for billing address data. This PR fixes this by addressing the following two underlying causes.

- In the frontend, billing details were only attached to `createPaymentMethod` when it's a checkout form. The Pay for Order form is `#order_review`, so no `billing_details` were sent. The Payment Element's `defaultValues` were pre-filled, but Stripe's card element never collects the hidden address fields (line1/city/state), so they never reached the PaymentMethod.
- The billing data localized to the page was always sourced from the customer profile, which is empty for a **guest** paying a merchant-created order, so the address the merchant entered on the order never reached the page. In this PR for pay-for-order orders, we try to source the billing address from the order data. If that's not available, we fallback to the customer data.

## Testing instructions

1. In the Stripe Dashboard (test mode) -> **Radar -> Rules**, add a **Block** rule: `Block if :address_line1_check: in ('fail', 'not_provided')`.
2. In WooCommerce admin, create a **manual order**: add a product, set a customer billing address 
3. Copy the **Customer payment page** link
4. Open that link in a logged-out / incognito window and place the order
5. **Expected:** the payment succeeds. Radar should not block this.
6. **Before this fix** the same payment is blocked by the Radar rule.
7. Regression - confirm standard checkout still works: place a normal order from the cart and verify billing details are still sent and payment succeeds.
8. Regression - confirm a logged-in customer paying for their own order via Pay for Order still succeeds with the address sent. In this case, the address set in the customer account should be used.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/stripe-utils/utils.js
+++ b/client/stripe-utils/utils.js
@@ -520,15 +520,6 @@ export const appendCheckoutSessionIdToForm = ( form, checkoutSessionId ) => {
 };
 
 /**
- * Craft the defaultValues parameter, used to pre-fill
- * user email and phone number for Link in the Payment Element.
- * On order pay and change payment method pages, also preloads all billing details
- * from the customer billing data passed from the server.
- *
- * @param {boolean} forCheckoutSession Whether the default values are for a Checkout Session.
- * @return {Object} The defaultValues object for the Payment Element.
- */
-/**
  * Returns true when the current page is one of the deferred-payment flows
  * (order pay, change payment method, or add payment method).
  *
--- a/client/stripe-utils/utils.js
+++ b/client/stripe-utils/utils.js
@@ -528,69 +528,94 @@ export const appendCheckoutSessionIdToForm = ( form, checkoutSessionId ) => {
  * @param {boolean} forCheckoutSession Whether the default values are for a Checkout Session.
  * @return {Object} The defaultValues object for the Payment Element.
  */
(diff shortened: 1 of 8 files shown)


client/stripe-utils/utils.js
```diff
@@ -528,69 +528,94 @@ export const appendCheckoutSessionIdToForm = ( form, checkoutSessionId ) => {
  * @param {boolean} forCheckoutSession Whether the default values are for a Checkout Session.
  * @return {Object} The defaultValues object for the Payment Element.
  */
-export const getDefaultValues = ( forCheckoutSession = false ) => {
```
diegocurbelo, line 531: We should also remove the orphaned docblock above.

diegocurbelo (approved):
The code changes look good, and it tests well 👍🏼 

|develop| this branch|
|-|-|
|Image Image|Image|

client/stripe-utils/utils.js
malithsen: Thanks! Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5542/commits/bdf16c86482b0a8304820cee22a2e3bdfb04c58c

PR #5544, opened 2026-06-09
Sync subscription payment-method row with saved-card wallet branding

### LINKED ISSUE #5543: Subscription payment-method row and saved-card list disagree on wallet badge after card+wallet reuse
### Description

When a single underlying card has been used both as a manually-entered card **and** through a wallet (Apple Pay / Google Pay), the subscription "Payment: Via …" row and the My Account → Payment methods saved-card list can disagree about whether to show a wallet badge.

The two surfaces read different sources of truth:

- **Saved-token list** shows the token's pinned create-time `wallet_type` (intentionally pinned in #5477 so a manually-saved card is never re-badged as a wallet on later reuse).
- **Subscription "Via …" row** (`get_payment_method_to_display_for_payment_method`) does a live Stripe fetch of the subscription's `_stripe_source_id` and reads `card->wallet->type` directly (added in #5384).

Fingerprint dedup in `WC_Stripe_Payment_Tokens::add_token_to_user` repoints an existing card token at the newer (wallet-flavored) Stripe PaymentMethod via `set_token()` while deliberately leaving `wallet_type` pinned. Selecting that token on change-payment then sets `_stripe_source_id` to a wallet PM, so the subscription row reads "Google Pay" even though the token list shows a bare card.

### Steps to reproduce

1. Change a subscription's payment method to a plain, non-3DS card (e.g. the `4242` Visa).
2. Change it again to Google Pay using the same underlying card.
   - Subscription view shows Google Pay; My Account → Payment methods still shows the bare card (expected per #5477).
3. Change it to some other payment method.
4. Change it again and select the saved "Visa ending in 4242" option (no wallet indicator).
5. The subscription screen now shows the wallet (Google Pay) details rather than the bare card.

### Expected behavior

The subscription "Via …" row and the saved-token list should agree. The row should reflect what the shopper actually selected: when a saved card token (pinned to a bare card) is chosen, the row should show the bare card, not the wallet carried by the underlying Stripe PaymentMethod.

### Additional context

Flagged by @daledupreez during review of #5384 (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5384#pullrequestreview-4418984027). Kept out of #5384 to keep that PR scoped; tracked here as a follow-up.

Likely fix: when rendering the subscription row, prefer the matching saved token's pinned `wallet_type` over the live PaymentMethod's `card->wallet->type`, falling back to the live wallet only when no matching token exists (e.g. guest / parent-order lookups).

wjrosa:
Fixes #5543
Fixes STRIPE-1203

Follow-up to the edge case @daledupreez flagged on #5384 (now merged). When one underlying card has been used both as a manually-entered card and through a wallet, the subscription "Payment: Via …" row and the My Account saved-card list could disagree about the wallet badge.

The two surfaces read different sources of truth: the saved-token list uses the token's pinned `wallet_type` (#5477), while the subscription row does a live Stripe fetch and reads `card->wallet->type` (#5384). Fingerprint dedup in `WC_Stripe_Payment_Tokens::add_token_to_user` repoints a card token at a newer, wallet-flavored PaymentMethod via `set_token()` while leaving `wallet_type` pinned — so selecting that token set `_stripe_source_id` to a wallet PM and the row read "Google Pay" even though the token list showed a bare card.

`maybe_render_subscription_payment_method` now looks up the saved token backing the subscription's `_stripe_source_id` and passes its pinned `wallet_type` to `get_payment_method_to_display_for_payment_method`, where it wins over the live PaymentMethod's wallet. A `null` override (no matching token, e.g. guest / parent-order lookups) preserves the previous live-wallet behaviour. Both surfaces now read the same source of truth.

## Testing instructions

1. Change a subscription's payment method to a plain, non-3DS card (e.g. the `4242` Visa).
2. Change it again to Google Pay using the same underlying card.
3. Change it to some other payment method, then change it again and select the saved "Visa ending in 4242" option (no wallet indicator).
4. Verify the subscription "Via …" row now shows the bare Visa (matching the saved-card list), **not** Google Pay.
5. Regression: change to a genuine wallet (with no matching saved card token, e.g. a fresh wallet PM) and verify the row still shows the wallet brand.

---

### Changelog entry

> Fix - Keep the subscription payment-method row in sync with the saved-card list when a card has been used both directly and through Apple Pay/Google Pay

**Post merge**


includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -1222,13 +1223,46 @@ public function maybe_render_subscription_payment_method( $payment_method_to_dis
 		return __( 'N/A', 'woocommerce-gateway-stripe' );
 	}
 
+	/**
+	 * Returns the pinned `wallet_type` of the saved token backing a Stripe PaymentMethod.
+	 *
+	 * Fingerprint dedup can repoint a saved card token at a newer (wallet-flavored)
+	 * PaymentMethod while leaving its `wallet_type` pinned (see #5477). The subscription
```
malithsen: nit: PR references in this doc block and in tests don't add much value imo. I'd rather remove them

malithsen (approved):
Code makes sense. However, I am not able to replicate the issue. Google Pay (Or any express payment methods) doesn't show up for me on the change payment method page.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Makes sense! Updated it in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5544/changes/0be18744135abf35f46b451f9d8b31fd86e8f039

wjrosa:
@malithsen double-check if the ECE methods are enabled, and that the page is set up in their settings, like:

Screenshot 2026-06-16 at 10 03 47

And that your store is publicly accessible

Mayisha (approved):
Changes look good to me and tests well.

Tests performed-
- Saved `4242 ..` Visa card in my account.
- Purchased a subscription with Google Pay.
- Subscription shows `Google Pay MasterCard ..` as payment method.
- On the change payment method page, selected another card from Google Pay.
- Subscription shows `Google Pay MasterCard ..` as payment method (the newly chosen card)
- On the change payment method page, selected the Visa `4242 ..` card from Google Pay
- Subscription correctly shows `Via Visa card ending in 4242` as payment method.

PR #5545, opened 2026-06-09
Verify webhook event's Stripe account matches the connected store

wjrosa:
Fixes STRIPE-1194

The webhook handler processed every signature-valid event without confirming it originated from the Stripe account this store is connected to. A store's connected account can change, or an endpoint can receive events from a different/stale account; acting on those could update the wrong store's orders or trigger mismatched payment/charge lookups.

This PR adds an account-match guard:

- New `event_belongs_to_connected_account()` compares the event's originating account against the connected account ID (resolved via the cached account data — the same context used to pick the webhook secret).
- The originating account is read from `event.account` for Connect events **or** `event.context` for agentic `v1.delegated_checkout.*` events (which carry the account in `context`, not `account`).
- `check_for_webhook()` runs the guard after signature validation and **before** dispatching to either the standard or agentic processing path. On mismatch it acknowledges the event with HTTP `200` (so Stripe stops retrying an event meant for another account) and logs a redacted reason (event id + type only — no payload dump).
- The check **fails open** so legitimate events are never dropped:
  - Event carries no account field (`account`/`context` absent) → allowed.
  - Connected account ID is unknown (cold/empty cache or lookup failure) → allowed.

Note: the Stripe account ID (`acct_…`) is identical across test and live mode for a given account, so the comparison is mode-agnostic.

**How can this break?** A false mismatch would silently drop a legitimate event. Mitigated by failing open whenever the event account or connected account can't be determined, and by comparing with `hash_equals`.

**What keeps it from breaking?** Targeted unit coverage for Connect (`account`) and agentic (`context`) sources across both modes (below), plus the existing agentic suite confirming that path is unaffected.

### Automated

- `npm run test:php -- --filter WC_Stripe_Webhook_Handler_Test` — passes, including the match/mismatch matrix (Connect `account`, agentic `context`, missing/empty account, unknown connected account) and a test that replays the committed sample event.
- `npm run test:php -- --filter WC_Stripe_Webhook_Handler_Agentic_Test` — passes.
- `npm run lint:php` and `npm run phpstan` — clean.

### Manual — replay the sample event directly

A real `v1.delegated_checkout.customize_checkout` event body is committed at `tests/phpunit/dummy-data/agentic_customize_checkout_event.json` (its account is in the top-level `context` field). You don't need to craft a signature — run the decision against it with wp-cli:

```bash
wp eval '
$e = json_decode( file_get_contents( WC_STRIPE_PLUGIN_PATH . "/tests/phpunit/dummy-data/agentic_customize_checkout_event.json" ) );
$h = new WC_Stripe_Webhook_Handler();
$m = new ReflectionMethod( $h, "event_belongs_to_connected_account" ); $m->setAccessible( true );
$connected = WC_Stripe::get_instance()->account->get_cached_account_data()["id"] ?? "(none)";
echo "connected account: {$connected}\n";
echo "mismatch — context={$e->context} -> " . var_export( $m->invoke( $h, $e ), true ) . "\n";
$e->context = $connected; // rewrite to this store account
echo "match    — context={$e->context} -> " . var_export( $m->invoke( $h, $e ), true ) . "\n";
'
```

Expected: the first line (the fixture's foreign account) prints `false` (event would be skipped); after rewriting `context` to your store's connected account it prints `true` (event would be processed).

### Manual — full webhook flow (optional)

To exercise the whole `check_for_webhook()` path, set `E2E_TESTING=true` in test mode (this bypasses signature verification — `validate_request()`) and `curl` the fixture body to the webhook URL (`?wc-api=wc_stripe`). With `WC_Stripe_Logger` logging on, a foreign `context` logs "Webhook ignored: the event's Stripe account does not match the connected account" and the event is not processed.

---

### Changelog entry

Added to the 10.9.0 block in both `changelog.txt` and `readme.txt`:

> Fix - Ignore incoming webhook events whose Stripe account does not match the connected account to avoid acting on another account's data

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -224,7 +224,7 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 			return true;
 		}
 
-		return hash_equals( $connected_account, $event_account );
+		return $event_account === $connected_account;
 	}
 
 	/**
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -178,8 +178,10 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 			WC_Stripe_Logger::error(
 				'Webhook ignored: the event\'s Stripe account does not match the connected account.',
 				[
-					'event_id'   => $event->id ?? null,
-					'event_type' => $event_type,
+					'event_id'          => $event->id ?? null,
+					'event_type'        => $event_type,
+					'event_account'     => $this->get_event_account_id( $event ),
+					'connected_account' => $this->get_connected_account_id(),
 				]
 			);
 
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -202,16 +202,9 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	}
 
 	/**
-	 * Determines whether a webhook event originated from the Stripe account this store is connected to.
+	 * Whether the event's Stripe account matches the connected account.
 	 *
-	 * Connect events carry a top-level `account` field identifying the originating account. When it is
-	 * present, it must match the account connected for the current mode; a mismatch means the event
-	 * belongs to a different or stale account and must not be processed against this store's data.
-	 *
-	 * The check intentionally fails open in two cases so legitimate events are never dropped:
-	 * - The event has no `account` field (typical for direct/keyed-account events), so it cannot be
-	 *   attributed to a specific account from the payload.
-	 * - The connected account ID is unknown (e.g. the account cache is empty or the lookup failed).
+	 * Fails open when the event has no `account` field or the connected account is unknown.
 	 *
 	 * @param object $event The decoded webhook event.
 	 * @return bool True when the event may be processed, false when it must be skipped.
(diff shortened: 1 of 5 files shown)


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -184,6 +201,65 @@ public function check_for_webhook() {
...
+		}
+
+		$connected_account = $this->get_connected_account_id();
+
+		// Connected account is unknown: avoid dropping legitimate events.
+		if ( '' === $connected_account ) {
+			return true;
+		}
+
+		return hash_equals( $connected_account, $event_account );
```
malithsen: non-blocking-nit: It doesn't seem like `hash_equals` is necessary here as opposed to a `===` check. Account IDs aren't really a secret.

malithsen (approved):
Looks good on inspection.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -172,6 +172,23 @@ public function check_for_webhook() {
 			exit;
 		}
 
+		// Ignore events that belong to a different Stripe account than the one this store is connected to.
+		// Acting on them could update the wrong orders or trigger mismatched payment/charge lookups.
+		if ( ! $this->event_belongs_to_connected_account( $event ) ) {
+			WC_Stripe_Logger::error(
+				'Webhook ignored: the event\'s Stripe account does not match the connected account.',
```
diegocurbelo, line 179: It would be useful to log the conencted account and the event's account, so the merchant can take action.

diegocurbelo (approved):
The code looks good, and the sample event replay tests as described 👍🏼

includes/class-wc-stripe-webhook-handler.php
wjrosa: You're right! Simplified the logic in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5545/commits/d48d5ab4ef44007a9ebeae3dc6b4106c0b15279a

includes/class-wc-stripe-webhook-handler.php
wjrosa: Log added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5545/changes/1cfe09faf5686f24dd54508338e1da211afa2b2b

PR #5547, opened 2026-06-09
Attribute agentic orders to the producing site when stores share a Stripe account

wjrosa:
Fixes STRIPE-968

When multiple WooCommerce stores connect to the **same Stripe account**, the order-creating `checkout.session.completed` event is a standard, broadcast webhook delivered to **every** store's endpoint on that account — so each store could create an order for the same agentic checkout, producing duplicate or wrong-site orders.

Stripe allows only **one** agentic-commerce endpoint per account for the synchronous `customize_checkout` / `finalize_checkout` hooks, and confirmed that **every agentic checkout fires one of those hooks before `checkout.session.completed`, with a matching `checkout_session` id**. This PR uses that as a per-site ownership signal:

- `claim_agentic_session()` — records the `cs_…` id seen in the sync hooks (only the producing site receives them).
- `is_agentic_session_claimed()` — gates `handle_agentic_checkout_session()`; an agentic `checkout.session.completed` whose session this site never claimed is a **logged no-op** (no order side effects).
- Claims are stored in a mode-scoped `WC_Stripe_Database_Cache` entry with a 1-day TTL (comfortably covers the gap between the sync hook and a pending checkout's completion).

## Testing instructions

Automated:

- `npm run test:php -- --filter WC_Stripe_Webhook_Handler_Agentic_Test` — 17/17 pass, including:
  - `test_process_checkout_session_completed_skips_unclaimed_session` — an otherwise-valid agentic session this site never claimed creates **no** order (the core guard).
  - `test_agentic_session_claim_roundtrip` — claim/read-back/empty-id behavior.
  - Existing order-creation tests now claim the session first (reflecting the real sync-hook → completion flow).
- `npm run lint:php` and `npm run phpstan` — clean.

Manual (optional, two sites, one Stripe account):

1. Connect Site A and Site B to the same Stripe account; configure Site A as the agentic endpoint.
2. Complete an agentic checkout produced from Site A's catalog.
3. Confirm the order is created on **Site A only**. With `WC_Stripe_Logger` on, Site B logs "Agentic checkout session was not claimed by this site; skipping order creation…" and creates nothing.

---

### Changelog entry

Added to the 10.9.0 block in `changelog.txt` and `readme.txt`:

> Fix - Only create an agentic commerce order on the site that produced the checkout, preventing duplicate or wrong-site orders when multiple stores share one Stripe account

**Post merge**


diegocurbelo (approved):
The code changes look good, and it tests well!

Until we have a way to read/update the agentic url and signature via API this claim seems like the best solution 👍🏼

PR #5550, opened 2026-06-10
Prevent fatal when saving redirect APMs as SEPA tokens in Adaptive Pricing webhook

wjrosa:
Fixes STRIPE-1205

With Optimized Checkout + Adaptive Pricing enabled (defaults since 10.7.0) and "save as SEPA token" enabled for Bancontact, saving a Bancontact payment through the Checkout Sessions flow triggered a fatal `TypeError` in the `checkout.session.completed` webhook. The order was left stuck at **Pending payment** and Stripe kept retrying the same fatal. The same applies to iDEAL and Sofort.

**Why it broke:** the AP webhook save path passed the raw Bancontact PaymentMethod (`type=bancontact`, no `sepa_debit` child) into the base SEPA token path, which dereferenced `$pm->sepa_debit->fingerprint` → `null`; PHP 8's strict typing on `set_fingerprint( string $fingerprint )` fataled. The deferred-intent / subscription flow already handled this by detecting `get_id() !== get_retrievable_type()` and fetching the charge's `generated_sepa_debit`; the Checkout Sessions path was missing that conversion.

**The fix:**
- Webhook handler: convert the PaymentMethod via a new gateway helper before saving; skip the save when no reusable mandate exists; clear the save flag unconditionally so webhook retries don't loop.
- UPE gateway: new `get_reusable_payment_method_for_saving()` mirroring the deferred-intent conversion.
- Base payment method: `create_payment_token_for_user()` now throws a catchable exception instead of fataling with a `TypeError` on a non-SEPA-shaped PaymentMethod.

**Known limitation (out of scope):** with the Checkout Sessions (Adaptive Pricing) flow, Stripe currently provides no way to mint the reusable SEPA mandate for these methods, so ticking the save checkbox does not produce a token. Verified against the Stripe test API (pinned version `2025-09-30.clover`):

- Session **update** with `payment_method_options` → `parameter_unknown` (this is what an earlier revision of this PR attempted; it never worked).
- Session **creation** with `payment_method_options[bancontact][setup_future_usage]=off_session` → `Invalid …: must be none`.
- Session **creation** with `payment_intent_data[setup_future_usage]=off_session` → ``unsupported for payment method `bancontact` ``.
- Client-side `checkout.confirm( { savePaymentMethod } )` only supports cards per [Stripe's docs](https://docs.stripe.com/payments/checkout/save-during-payment?payment-ui=checkout-form).

A follow-up PR hides the save checkbox for these methods in the Checkout Sessions flow so shoppers aren't offered a save that can't happen.

**How can this code break / what prevents it:** the webhook conversion safely skips saving when `generated_sepa_debit` is `null`, so the fatal cannot recur even when no mandate is generated. Behavior change is scoped to the three SEPA-tokenized redirect APMs; methods saved under their own type (card, SEPA, etc.) are unaffected.

## Testing instructions

Prerequisites:
1. WooCommerce > Settings > Payments > Stripe > Advanced: turn **Optimized Checkout Suite** ON and **Adaptive Pricing** ON.
2. Enable **Bancontact**, and enable "Save Bancontact payments as SEPA Direct Debit tokens".
3. Store currency **EUR**; use the **Block** checkout.

Steps:
1. Log in as an existing customer, add a non-subscription product to the cart, go to checkout.
2. Select **Bancontact**, tick **"Save payment information to my account for future purchases."**, place the order, and complete the Stripe-hosted authorization.
3. Confirm the order moves to **Processing/Completed** (it must NOT get stuck at "Pending payment", and there must be no fatal in the logs). Per the limitation above, no SEPA token is created — the logs should show the save being skipped, not a fatal.
4. Repeat without ticking the save box: the order completes normally and no token is saved.
5. Regression: a card payment through the same AP/OC checkout still saves normally; a Bancontact payment with the SEPA-tokens setting OFF completes normally and saves nothing.

---

### Changelog entry

* Fix - Prevent a fatal error when a Bancontact, iDEAL or Sofort payment is placed with SEPA token saving enabled through the Adaptive Pricing checkout, which left the order unpaid

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1812,7 +1812,7 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 				$order_helper->update_stripe_source_id( $order, $payment_method_id );
 			}
 
-			// Fetch the charge once; reused below to process the response and to resolve any reusable SEPA mandate.
+			// Fetch the charge once; reused below.
 			$charge = $this->get_latest_charge_from_intent( $intent );
 
 			// Save payment method to store if the customer requested it during checkout.
@@ -1821,17 +1821,15 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 
 				$payment_method_object = is_object( $intent->payment_method ) ? $intent->payment_method : WC_Stripe_API::retrieve( 'payment_methods/' . $payment_method_id );
 				if ( $upe_gateway instanceof WC_Stripe_UPE_Payment_Gateway && ! is_wp_error( $payment_method_object ) && empty( $payment_method_object->error ) && ! empty( $payment_method_object ) ) {
-					// Redirect-based APMs (Bancontact, iDEAL, Sofort) are returned as their own PaymentMethod
-					// type but are saved as SEPA tokens. Convert to the generated SEPA debit PM before saving,
-					// mirroring the deferred-intent flow. Returns null when Stripe did not generate a reusable
-					// mandate, in which case there is nothing to save (and we avoid a fatal on set_fingerprint()).
+					// Redirect APMs saved as SEPA tokens need converting to their generated SEPA debit PM;
+					// null means no reusable mandate was generated, so there is nothing to save.
 					$payment_method_to_save = $upe_gateway->get_reusable_payment_method_for_saving( $payment_method_object, $charge );
 
 					if ( $payment_method_to_save instanceof stdClass && ! empty( $payment_method_to_save->type ) ) {
 						$upe_gateway->handle_saving_payment_method( $order, $payment_method_to_save, $payment_method_to_save->type );
 					}
 
-					// Clear the flag so it does not run again on webhook retries (even when nothing was saved).
+					// Clear the flag so retries don't re-run this, even when nothing was saved.
 					$order_helper->delete_should_save_stripe_payment_method( $order );
 				}
 			}
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1821,11 +1821,11 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 
 				$payment_method_object = is_object( $intent->payment_method ) ? $intent->payment_method : WC_Stripe_API::retrieve( 'payment_methods/' . $payment_method_id );
 				if ( $upe_gateway instanceof WC_Stripe_UPE_Payment_Gateway && ! is_wp_error( $payment_method_object ) && empty( $payment_method_object->error ) && ! empty( $payment_method_object ) ) {
-					// Redirect APMs saved as SEPA tokens need converting to their generated SEPA debit PM;
-					// null means no reusable mandate was generated, so there is nothing to save.
+					// Get the payment method details that should be saved. That may be different from the
+					// original payment method, e.g. for Bancontact and iDEAL/Wero, which are saved as SEPA.
 					$payment_method_to_save = $upe_gateway->get_reusable_payment_method_for_saving( $payment_method_object, $charge );
 
-					if ( $payment_method_to_save instanceof stdClass && ! empty( $payment_method_to_save->type ) ) {
+					if ( is_object( $payment_method_to_save ) && ! empty( $payment_method_to_save->type ) ) {
 						$upe_gateway->handle_saving_payment_method( $order, $payment_method_to_save, $payment_method_to_save->type );
 					}
 
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1812,15 +1812,26 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 				$order_helper->update_stripe_source_id( $order, $payment_method_id );
 			}
 
+			// Fetch the charge once; reused below to process the response and to resolve any reusable SEPA mandate.
+			$charge = $this->get_latest_charge_from_intent( $intent );
+
 			// Save payment method to store if the customer requested it during checkout.
 			if ( $order_helper->get_should_save_stripe_payment_method( $order ) && ! empty( $payment_method_id ) ) {
 				$upe_gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
 
 				$payment_method_object = is_object( $intent->payment_method ) ? $intent->payment_method : WC_Stripe_API::retrieve( 'payment_methods/' . $payment_method_id );
-				if ( ! is_wp_error( $payment_method_object ) && empty( $payment_method_object->error ) && ! empty( $payment_method_object ) ) {
-					$upe_gateway->handle_saving_payment_method( $order, $payment_method_object, $payment_method_object->type );
+				if ( $upe_gateway instanceof WC_Stripe_UPE_Payment_Gateway && ! is_wp_error( $payment_method_object ) && empty( $payment_method_object->error ) && ! empty( $payment_method_object ) ) {
+					// Redirect-based APMs (Bancontact, iDEAL, Sofort) are returned as their own PaymentMethod
+					// type but are saved as SEPA tokens. Convert to the generated SEPA debit PM before saving,
+					// mirroring the deferred-intent flow. Returns null when Stripe did not generate a reusable
+					// mandate, in which case there is nothing to save (and we avoid a fatal on set_fingerprint()).
+					$payment_method_to_save = $upe_gateway->get_reusable_payment_method_for_saving( $payment_method_object, $charge );
 
-					// Clear the flag so it does not run again on webhook retries.
+					if ( $payment_method_to_save instanceof stdClass && ! empty( $payment_method_to_save->type ) ) {
+						$upe_gateway->handle_saving_payment_method( $order, $payment_method_to_save, $payment_method_to_save->type );
+					}
+
+					// Clear the flag so it does not run again on webhook retries (even when nothing was saved).
 					$order_helper->delete_should_save_stripe_payment_method( $order );
 				}
 			}
@@ -1835,8 +1846,6 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 			}
 
 			$order->save();
-
-			$charge = $this->get_latest_charge_from_intent( $intent );
 
 			$charge->is_webhook_response = true;
 			$this->process_response( $charge, $order );
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -1525,9 +1525,10 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 	}
 
 	/**
-	 * Sets setup_future_usage on the session so Stripe mints a reusable SEPA mandate for redirect-based
-	 * APMs (Bancontact, iDEAL, Sofort). Best-effort — failures are logged and swallowed so they cannot
-	 * block checkout.
+	 * Sets `setup_future_usage` to `off_session` for the checkout session, which will
+	 * ensure the payment method is saved. This is necessary for payment methods that "convert"
+	 * to SEPA debits when saved, including Bancontact and iDEAL/Wero.
+	 * Note that failures are logged and ignored so we don't prevent any other processing.
 	 *
 	 * @param string $checkout_session_id The Stripe Checkout Session ID.
 	 * @param string $payment_method_type The Stripe payment method type (e.g. 'bancontact').
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -3778,6 +3778,11 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 			return null;
 		}
 
+		/**
+		 * Decoded Stripe API JSON objects are always stdClass.
+		 *
+		 * @var stdClass $generated_payment_method
+		 */
 		return $generated_payment_method;
 	}
 
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -1438,29 +1438,40 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
 		$order_helper->update_stripe_checkout_session_id( $order, $checkout_session_id );
 
-		// Persist the 'save-payment-method' flag so the webhook handler can create the token once payment is completed.
-		$upe_payment_method = $this->payment_methods[ $selected_payment_type ] ?? null;
-		if ( $save_payment_method && $upe_payment_method && $upe_payment_method->get_id() === $upe_payment_method->get_retrievable_type() ) {
-			$order_helper->update_should_save_stripe_payment_method( $order, true );
-		}
-
-		// Eagerly set the payment method title from the customer's selection so the order
-		// reflects the right method as soon as it lands on the order-received page, even if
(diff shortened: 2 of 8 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3666,6 +3666,48 @@ public function handle_saving_payment_method( WC_Order $order, stdClass $payment
...
+		if ( ! $instance instanceof WC_Stripe_UPE_Payment_Method || $instance->get_id() === $instance->get_retrievable_type() ) {
+			return $payment_method_object;
+		}
+
+		// Conversion required: locate the generated SEPA debit PM on the charge. If Stripe did not
+		// generate one (e.g. the Checkout Session never set setup_future_usage), there is nothing to save.
+		$generated_payment_method_id = is_object( $charge ) ? ( $charge->payment_method_details->{$type}->generated_sepa_debit ?? null ) : null;
+		if ( empty( $generated_payment_method_id ) ) {
+			return null;
+		}
```
daledupreez: There is a major implicit assumption that `'sepa_debit' === $instance->get_retrievable_type()`. We should only run this branch if we know we are looking for a SEPA token.
```suggestion
		$generated_payment_method_id = null;
		if ( WC_Stripe_UPE_Payment_Method_Sepa::STRIPE_ID == $instance->get_retrievable_type() ) {
			$generated_payment_method_id = is_object( $charge ) ? ( $charge->payment_method_details->{$type}->generated_sepa_debit ?? null ) : null;
		}

		if ( empty( $generated_payment_method_id ) ) {
			return null;
		}
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3666,6 +3666,48 @@ public function handle_saving_payment_method( WC_Order $order, stdClass $payment
...
+	 * in the deferred-intent flow (see process_upe_redirect_payment) so the Adaptive Pricing /
+	 * Checkout Sessions webhook path does not pass a non-SEPA-shaped object to a SEPA token.
+	 *
+	 * @param stdClass             $payment_method_object The PaymentMethod object used for the payment.
+	 * @param object|string|null   $charge                The latest charge from the intent, if available.
+	 *
+	 * @return stdClass|null The PaymentMethod to save: the generated SEPA debit PM when a conversion
+	 *                       is required and a mandate was generated; the original object when no
+	 *                       conversion is needed; or null when a conversion is required but Stripe did
+	 *                       not generate a reusable mandate (nothing to save).
```
daledupreez: I think we should simplify the commentary here.
```suggestion
	 * Payment methods that are saved as secondary types need special logic to get the saved 
	 * payment method details. This primarily occurs for Bancontact and iDEAL/Wero, which are 
	 * saved as SEPA tokens.
	 *
	 * @param stdClass             $payment_method_object The PaymentMethod object used for the payment.
	 * @param object|string|null   $charge                The latest charge from the intent, if available.
	 *
	 * @return stdClass|null The PaymentMethod to save: the payment method to save or null,
	 *                       where the payment method may need to be resolved for payment methods
	 *                       that have been converted to SEPA or similar.
```

includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -544,6 +544,17 @@ public function get_retrievable_type() {
 	 * @return WC_Payment_Token_SEPA
 	 */
 	public function create_payment_token_for_user( $user_id, $payment_method ) {
+		// Guard against non-SEPA-shaped PaymentMethods (e.g. a raw Bancontact/iDEAL/Sofort object reaching
+		// this SEPA token path). Without a SEPA fingerprint there is nothing reusable to tokenize, and
+		// set_fingerprint() would fatal on null. Fail with a catchable exception instead of a TypeError.
+		$sepa_debit = $payment_method->sepa_debit ?? null;
```
daledupreez, line 549: Not blocking: Having SEPA-specific code on this payment method feels super-messy, but that's a pre-existing issue.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1812,15 +1812,26 @@ protected function handle_checkout_session_success( object $notification ): void
...
 				$upe_gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
 
 				$payment_method_object = is_object( $intent->payment_method ) ? $intent->payment_method : WC_Stripe_API::retrieve( 'payment_methods/' . $payment_method_id );
-				if ( ! is_wp_error( $payment_method_object ) && empty( $payment_method_object->error ) && ! empty( $payment_method_object ) ) {
-					$upe_gateway->handle_saving_payment_method( $order, $payment_method_object, $payment_method_object->type );
+				if ( $upe_gateway instanceof WC_Stripe_UPE_Payment_Gateway && ! is_wp_error( $payment_method_object ) && empty( $payment_method_object->error ) && ! empty( $payment_method_object ) ) {
+					// Redirect-based APMs (Bancontact, iDEAL, Sofort) are returned as their own PaymentMethod
+					// type but are saved as SEPA tokens. Convert to the generated SEPA debit PM before saving,
+					// mirroring the deferred-intent flow. Returns null when Stripe did not generate a reusable
+					// mandate, in which case there is nothing to save (and we avoid a fatal on set_fingerprint()).
```
daledupreez: We can cut down on the commentary here.
```suggestion
					// Get the payment methods details that should be saved.
					// That may be different from the original payment method,
					// e.g. for Bancontact and iDEAL/Wero, which are saved as SEPA payment methods.
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1812,15 +1812,26 @@ protected function handle_checkout_session_success( object $notification ): void
...
+					$payment_method_to_save = $upe_gateway->get_reusable_payment_method_for_saving( $payment_method_object, $charge );
+
+					if ( $payment_method_to_save instanceof stdClass && ! empty( $payment_method_to_save->type ) ) {
```
daledupreez: Is it worth using `is_object()` here instead of `instanceof stdClass`?
```suggestion
					if ( is_object( $payment_method_to_save ) && ! empty( $payment_method_to_save->type ) ) {
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1438,29 +1438,40 @@ private function process_payment_with_checkout_session( int $order_id, string $c
...
+		// Persist the 'save-payment-method' flag so the webhook handler can create the token once payment is completed.
+		if ( $save_payment_method && $resolved_payment_method && $resolved_payment_method->is_reusable() ) {
+			$order_helper->update_should_save_stripe_payment_method( $order, true );
+
+			// Redirect-based APMs that are tokenized as SEPA (Bancontact, iDEAL, Sofort) only generate a
+			// reusable SEPA mandate when the Checkout Session asks for it via setup_future_usage. The save
+			// checkbox alone (saved_payment_method_options.payment_method_save) does not mint a
+			// `generated_sepa_debit` for these methods, so the save flag would otherwise be meaningless.
+			// Request it on the session before the client confirms it. Best-effort: a failure must not block
+			// checkout — the order still completes, and the webhook safely skips saving when no mandate exists.
```
daledupreez: We should reduce the commentary here.
```suggestion
			// Payment methods that are saved as a different payment method need to have
			// `setup_future_usage` specified to request that the payment method be saved.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1438,29 +1438,40 @@ private function process_payment_with_checkout_session( int $order_id, string $c
...
+			if ( $resolved_payment_method->get_id() !== $resolved_payment_method->get_retrievable_type() ) {
+				$this->set_setup_future_usage_on_checkout_session( $checkout_session_id, $resolved_payment_method->get_id() );
+			}
+		}
+
+		// Eagerly set the payment method title from the customer's selection so the order reflects the
+		// right method as soon as it lands on the order-received page, even if the
+		// checkout.session.completed webhook is delayed. The webhook handler will reaffirm/refine this
+		// once the actual payment method type is known from the intent.
```
daledupreez: This commentary is unnecessary, and wasn't needed before this PR.
```suggestion
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Done — now only resolves `generated_sepa_debit` when the retrievable type is SEPA.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Done — adopted your wording.

includes/payment-methods/class-wc-stripe-upe-payment-method.php
wjrosa: Agreed — pre-existing, leaving it as-is for now.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Done.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Done — switched to `is_object()`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Done.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Done — removed.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1480,6 +1481,33 @@ private function process_payment_with_checkout_session( int $order_id, string $c
 		];
 	}
 
+	/**
+	 * Sets setup_future_usage on the session so Stripe mints a reusable SEPA mandate for redirect-based
+	 * APMs (Bancontact, iDEAL, Sofort). Best-effort — failures are logged and swallowed so they cannot
+	 * block checkout.
```
daledupreez: Nit: This language is overly specific and not super-clear.
```suggestion
	 * Sets `setup_future_usage` to `off_session` for the checkout session, which will
	 * ensure the payment method is saved. This is necessary for payment methods that "convert"
	 * to SEPA debits when saved, including Bancontact and iDEAL/Wero.
	 * Note that failures are logged and ignored so we don't prevent any other processing.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3666,6 +3694,45 @@ public function handle_saving_payment_method( WC_Order $order, stdClass $payment
...
+	 *
+	 * @param stdClass           $payment_method_object The PaymentMethod object used for the payment.
+	 * @param object|string|null $charge                The latest charge from the intent, if available.
+	 *
+	 * @return stdClass|null The payment method to save, or null. The payment method may need to be
+	 *                       resolved for payment methods that have been converted to SEPA or similar.
+	 */
+	public function get_reusable_payment_method_for_saving( stdClass $payment_method_object, $charge ) {
+		$type     = $payment_method_object->type ?? '';
+		$instance = '' !== $type ? $this->get_payment_method_instance( $type ) : null;
```
daledupreez: Non-blocking nit: We could return early when `'' === $type`.
```suggestion
		if ( '' === $type ) {
			return $payment_method_object;
		}

		$instance = $this->get_payment_method_instance( $type );
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3666,6 +3694,45 @@ public function handle_saving_payment_method( WC_Order $order, stdClass $payment
...
+		if ( WC_Stripe_UPE_Payment_Method_Sepa::STRIPE_ID === $instance->get_retrievable_type() ) {
+			$generated_payment_method_id = is_object( $charge ) ? ( $charge->payment_method_details->{$type}->generated_sepa_debit ?? null ) : null;
+		}
+
+		if ( empty( $generated_payment_method_id ) ) {
+			return null;
+		}
+
+		$generated_payment_method = WC_Stripe_API::retrieve( 'payment_methods/' . $generated_payment_method_id );
+		if ( empty( $generated_payment_method ) || is_wp_error( $generated_payment_method ) || ! $generated_payment_method instanceof stdClass || ! empty( $generated_payment_method->error ) ) {
```
daledupreez: Is it worth using `! is_object( $generated_payment_method )` instead of `! $generated_payment_method instanceof stdClass`? It feels like the specificity of the `instanceof stdClass` check could be brittle in the long run.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -1938,6 +1938,132 @@ public function test_process_payment_with_checkout_session_returns_disambiguatio
...
+	 *
+	 * Part of STRIPE-1205.
+	 */
+	public function test_process_payment_with_checkout_session_requests_setup_future_usage_for_redirect_apm() {
+		$session_id = 'cs_test_sfu_bancontact';
+
+		$_POST['wc_stripe_checkout_session_id'] = $session_id;
+		$_POST['payment_method']                = WC_Stripe_UPE_Payment_Gateway::ID;
+		$_POST['wc-stripe-payment-method']      = WC_Stripe_UPE_Payment_Gateway::ID;
+		// Optimized Checkout reports 'card' as the gateway type; the real method is sent separately.
```
daledupreez: Nit: It's not being sent "separately", it's being sent in `wc_stripe_selected_upe_payment_type`.
```suggestion
		// Optimized Checkout reports 'card' as the gateway type; the real method is sent in `wc_stripe_selected_upe_payment_type`.
```

daledupreez (changes requested):
The changes LGTM, and mostly test well across Bancontact, iDEAL/Wero, and SEPA, with Bancontact configured to allow saved payment methods and iDEAL/Wero configured not to.

However, I think the logic to call `setup_future_usage` is incorrect, as the [checkout session update API](https://docs.stripe.com/api/checkout/sessions/update) doesn't support the `payment_method_options` parameter. So I think we need to remove that code, as I don't believe it is working -- I think we are updating the client-side checkout session in response to the saved payment method checkbox.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Personally, I don't see much gain with this change, but I applied it in my last commits

wjrosa:
Thanks for the review, Dale! I confirmed that against the [update API docs](https://docs.stripe.com/api/checkout/sessions/update). Removed the param in 2f2b614a3.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -2156,14 +2090,26 @@ public function test_process_payment_with_checkout_session_skips_setup_future_us
...
 	}
 
+	/**
+	 * Provides payment method types for the checkout session save-payment-method test.
+	 *
+	 * @return array[]
+	 */
+	public function provide_checkout_session_save_payment_method_types(): array {
+		return [
+			'redirect APM tokenized as SEPA (bancontact)' => [ WC_Stripe_Payment_Methods::BANCONTACT ],
```
daledupreez, line 2108: Is it worth adding iDEAL for completeness? (Not blocking.)

daledupreez (approved):
The updated changes look good to me based on inspection -- I can test further in the morning.

daledupreez (changes requested):
Actually, now that I am testing again -- the payment method is not being being saved via block checkout. :-/ So something is not working as expected.

wjrosa:
You're right — and it turns out the save never worked at any commit here: the session-update call was silently rejected (`parameter_unknown: payment_method_options`), and I verified against the test API that Checkout Sessions has no other way to request `setup_future_usage` for these methods (creation only accepts `none`, `payment_intent_data` errors, and client-side `savePaymentMethod` is cards-only). So I've scoped this PR down to the fatal fix and opened #5553 to hide the save checkbox for these methods in this flow.

daledupreez (approved):
I've re-tested with the reduced scope and everything is working as it should.

PR #5551, opened 2026-06-10
Agentic Commerce: populate shipping package contents so content-dependent methods price real carts

wjrosa:
Fixes STRIPE-986

The agentic shipping calculator and the order mapper's `map_shipping()` both handed `WC_Shipping::calculate_shipping()` a package with empty `contents` / `contents_cost: 0` (flagged with a `See STRIPE-986` comment in the source). Fixed-cost methods don't care, but content-dependent methods — table rate, weight-based, per-item, or flat-rate `[qty]`/`[cost]` expressions — were pricing an empty cart.

This PR adds `WC_Stripe_Agentic_Shipping_Package_Builder`, which builds WC cart-item-format `contents` entries (`data`, `quantity`, `line_total`, `line_subtotal`, …) from the two sources the agentic flow has:

- `build_contents_from_event()` — customize_checkout line items: resolves `sku_id` through the existing product resolver and prices lines from `unit_amount × quantity`, falling back to the catalog price when `unit_amount` is absent (new `get_unit_amount()` getter on the line-item wrapper).
- `build_contents_from_order()` — the mapped order's product items, used by `map_shipping()` where products and totals are already resolved.

Both filter to `$product->needs_shipping()`, variations get the core `product_id`/`variation_id` split, and `build_package()` derives `contents_cost` from the entries' line totals.

**Behavior changes**

- The shipping calculator now throws when a line item's `sku_id` can't be resolved, instead of silently quoting rates from an incomplete package. In the customize_checkout hook this is unreachable in practice: the tax calculator already throws for the same event before shipping runs.
- Stores with WC's legacy (non-zone) shipping methods enabled will now see those rates offered, because legacy methods only emit rates for packages with contents. This matches what regular checkout offers. It also surfaced in tests — `WC_Helper_Order` leaves the legacy flat rate enabled in the test environment — so both shipping test suites now remove it in `setUp()` for determinism.

**Design decisions**

- `cart_subtotal` is set to `contents_cost` (shippable items only). WC core sets it to the displayed cart subtotal including non-shippable items, but no core shipping method reads it from the package; keeping it derived from contents avoids re-pricing virtual items in a context with no cart.
- Free Shipping's min-amount check reads `WC()->cart` directly, which doesn't exist in webhook context — populating the package doesn't (and can't) change that method. Pre-existing limitation, out of scope.
- `line_total` from the event uses `unit_amount` verbatim, tax-exclusive by contract since this same hook computes the tax.
- `applied_coupons` stays empty; the customize_checkout event carries no coupon data.

### How can this code break?

- A third-party method reading exotic cart-item keys we don't populate (e.g. `variation` attribute arrays) could still misprice; the entries carry the keys WC core's methods read.
- An unresolvable SKU now fails the shipping quote (by design, see above) — a catalog drift between Stripe and WC turns from wrong-rates into no-rates, which the agent surfaces instead of quoting a bogus price.

## Testing instructions

> [!NOTE]
> The agentic hooks are HMAC-verified against the **Agentic** webhook secret, so they can't be hand-`curl`ed as-is. For local testing you can take `validate_request()`'s built-in bypass (`E2E_TESTING` + Stripe **test mode** → signature/secret checks skipped — see `WC_Stripe_Webhook_Handler::validate_request()`). Step 2 (the shipping quote) is a synchronous hook you can replay directly. Step 3 (order creation) is deferred to Action Scheduler and re-fetches the session from Stripe, so its cURL recipe also mocks that fetch — see the collapsible block under step 3.

**Setup**

1. Make sure the store is in Stripe **test mode**, Agentic Commerce is enabled, and there's a product with SKU `SKU0029` priced `24.99`.
2. Create a US shipping zone with a single flat rate titled `Expedited` whose cost is a content-dependent expression, `2 * [qty]`.

**Step 2 — quote shipping from a `customize_checkout` payload**

Skip the webhook signature check by dropping this temporary mu-plugin in `wp-content/mu-plugins/skip-stripe-webhook-sig.php` (**delete it when you're done**):

```php
<?php
/**
 * TEMPORARY — delete after testing.
 *
 * Replays agentic webhook payloads without a valid Stripe signature by taking
 * validate_request()'s built-in E2E bypass: it returns success when E2E_TESTING
 * is truthy and the store is in Stripe test mode. No plugin/core edit required.
 *
 * @see WC_Stripe_Webhook_Handler::validate_request()
 */
putenv( 'E2E_TESTING=true' );
```

Save the payload below to `customize-checkout.json`, then POST it (swap the host for your store — the local default is `localhost:8072`; worktrees use `8170`–`8189`). With the bypass active no `Stripe-Signature` header is needed:

```bash
curl -i -X POST 'http://localhost:8072/?wc-api=wc_stripe' \
  -H 'Content-Type: application/json' \
  --data-binary @customize-checkout.json
```

The response is JSON; check `shipping_options[].shipping_rate_data.fixed_amount.amount` — it should be `200` (= 2.00) for the qty-1 item with the `2 * [qty]` rate, where before this change the empty package made it `0`.

**Step 3 — complete an agentic checkout (end-to-end)**

The genuine end-to-end is a real test-mode agentic checkout choosing that rate: the created order's shipping line should be the matched WC rate (method `flat_rate`), not a free-form `stripe_agentic` fallback line. This is also covered by `WC_Stripe_Agentic_Commerce_Order_Mapper_Test`.

<details>
<summary>Optional: drive step 3 from cURL too (advanced — mocks the Stripe session fetch)</summary>

Order creation is deferred to Action Scheduler and the handler re-fetches the session from Stripe (`WC_Stripe_API::retrieve`) before mapping, so a forged `checkout.session.completed` needs the fetch mocked. Replace the step-2 mu-plugin with this one (still **temporary — delete after testing**); it keeps the signature bypass and returns a canned agentic session for the checkout-session retrieve only:

```php
<?php
/**
 * TEMPORARY — delete after testing.
 *
 * 1. Skips Stripe webhook signature verification (validate_request() E2E bypass).
 * 2. Mocks the checkout-session retrieve the agentic order-creation path performs,
 *    so a forged checkout.session.completed can build an order without a real
 *    Stripe session. Only intercepts GET checkout/sessions/<id>.
 *
 * @see WC_Stripe_Webhook_Handler::validate_request()
 * @see WC_Stripe_Webhook_Handler::handle_agentic_checkout_session()
 */
putenv( 'E2E_TESTING=true' );

add_filter(
	'pre_http_request',
	function ( $preempt, $args, $url ) {
		if ( false === strpos( $url, 'api.stripe.com/v1/checkout/sessions/' ) ) {
			return $preempt;
		}

		// is_agentic() is true only when payment_intent.agent_details.network_business_profile
		// is non-empty. amount_total (2699) = item 2499 + the WC '2 * [qty]' rate (200) so
		// verify_order_total() passes. shipping_rate.display_name must match the flat-rate
		// method title ('Expedited') so the mapper matches the WC rate, not the fallback.
		$session = [
			'id'               => 'cs_test_a1ExampleCheckoutSessionIdForDocsOnly000000000000000000',
			'currency'         => 'usd',
			'amount_subtotal'  => 2499,
			'amount_total'     => 2699,
			'payment_intent'   => [
				'id'            => 'pi_test_ExamplePaymentIntent001',
				'agent_details' => [ 'network_business_profile' => 'acct_example_agent' ],
			],
			'customer'         => 'cus_test_Example001',
			'customer_email'   => 'agent-shopper@example.com',
			'customer_details' => [
				'email'   => 'agent-shopper@example.com',
				'name'    => 'Agent Shopper',
				'phone'   => '+15555550123',
				'address' => [
					'line1'       => '123 Example Street',
					'line2'       => '',
					'city'        => 'San Francisco',
					'state'       => 'CA',
					'postal_code' => '94016',
					'country'     => 'US',
				],
			],
			'shipping_details' => [
				'name'    => 'Agent Shopper',
				'address' => [
					'line1'       => '123 Example Street',
					'line2'       => '',
					'city'        => 'San Francisco',
					'state'       => 'CA',
					'postal_code' => '94016',
					'country'     => 'US',
				],
			],
			'total_details'    => [
				'amount_shipping' => 200,
				'amount_tax'      => 0,
				'amount_discount' => 0,
			],
			'shipping_cost'    => [
				'shipping_rate' => [
					'display_name' => 'Expedited',
					'metadata'     => [],
				],
			],
			'line_items'       => [
				'data' => [
					[
						'id'              => 'li_ExampleLineItem001',
						'description'     => 'Sample Product',
						'quantity'        => 1,
						'amount_total'    => 2499,
						'amount_subtotal' => 2499,
						'amount_tax'      => 0,
						'price'           => [
							'unit_amount'        => 2499,
							'external_reference' => 'SKU0029',
							'currency'           => 'usd',
						],
					],
				],
			],
		];

		return [
			'response' => [ 'code' => 200, 'message' => 'OK' ],
			'body'     => wp_json_encode( $session ),
		];
	},
	10,
	3
);
```

1. **Run step 2's cURL first** with the same `checkout_session` id — the `customize_checkout` hook claims the session for this site; without the claim, order creation is skipped (the claim guards against sibling sites on a shared account).
2. Save the event below to `checkout-session-completed.json` and POST it (the response is `200` and the order is *not* created yet — the event is deferred):

   ```json
   {
     "id": "evt_test_checkout_session_completed",
     "type": "checkout.session.completed",
     "livemode": false,
     "data": {
       "object": {
         "id": "cs_test_a1ExampleCheckoutSessionIdForDocsOnly000000000000000000",
         "object": "checkout.session",
         "status": "complete",
         "payment_status": "paid"
       }
     }
   }
   ```

   ```bash
   curl -i -X POST 'http://localhost:8072/?wc-api=wc_stripe' \
     -H 'Content-Type: application/json' \
     --data-binary @checkout-session-completed.json
   ```
3. Run the deferred job so the order is created:

   ```bash
   npm run wp -- action-scheduler run
   ```
4. A new order should exist (created via `stripe-agentic`) with total `26.99` and a shipping line whose method is `flat_rate` (title `Expedited`, `2.00`), **not** a free-form `stripe_agentic` line.

This mirrors `WC_Stripe_Agentic_Commerce_Order_Mapper_Test` against a live store; the canned session is hand-maintained, so prefer a real agentic checkout for true end-to-end coverage.
</details>

<details>
<summary>Example <code>customize_checkout</code> event (sanitized) — <code>customize-checkout.json</code></summary>

```json
{
  "id": "evt_test_customize_checkout",
  "type": "v1.delegated_checkout.customize_checkout",
  "livemode": false,
  "data": {
    "amount_subtotal": 2499,
    "amount_total": 3498,
    "automatic_tax": { "enabled": false },
    "checkout_session": "cs_test_a1ExampleCheckoutSessionIdForDocsOnly000000000000000000",
    "currency": "usd",
    "discounts": [],
    "line_item_details": [
      {
        "id": "li_ExampleLineItem001",
        "amount_discount": 0,
        "amount_subtotal": 2499,
        "amount_tax": 0,
        "amount_total": 2499,
        "name": "Sample Product",
        "quantity": 1,
        "sku_id": "SKU0029",
        "tax_rates": [],
        "unit_amount": 2499
      }
    ],
    "metadata": {},
    "shipping_details": {
      "address": {
        "city": "San Francisco",
        "country": "US",
        "line1": "123 Example Street",
        "line2": "",
        "postal_code": "94016",
        "state": "CA"
      },
      "discount_details": [],
      "shipping_rate": {
        "id": "shr_ExampleExpedited001",
        "delivery_estimate": { "maximum": { "unit": "day", "value": 3 }, "minimum": { "unit": "day", "value": 2 } },
        "display_name": "Expedited",
        "fixed_amount": { "amount": 999, "currency": "usd", "currency_options": { "usd": { "amount": 999, "tax_behavior": "exclusive" } } },
        "metadata": {},
        "tax_behavior": "exclusive",
        "tax_code": "txcd_92010001"
      },
      "shipping_rates": [
        {
          "id": "shr_ExampleExpedited001",
          "delivery_estimate": { "maximum": { "unit": "day", "value": 3 }, "minimum": { "unit": "day", "value": 2 } },
          "display_name": "Expedited",
          "fixed_amount": { "amount": 999, "currency": "usd", "currency_options": { "usd": { "amount": 999, "tax_behavior": "exclusive" } } },
          "metadata": {},
          "tax_behavior": "exclusive",
          "tax_code": "txcd_92010001"
        },
        {
          "id": "shr_ExampleExpress001",
          "delivery_estimate": { "maximum": { "unit": "day", "value": 2 }, "minimum": { "unit": "day", "value": 1 } },
          "display_name": "Express",
          "fixed_amount": { "amount": 1499, "currency": "usd", "currency_options": { "usd": { "amount": 1499, "tax_behavior": "exclusive" } } },
          "metadata": {},
          "tax_behavior": "exclusive",
          "tax_code": "txcd_92010001"
        }
      ]
    },
    "total_details": { "amount_discount": 0, "amount_shipping": 999, "amount_tax": 0 }
  }
}
```

This exact body is exercised end-to-end by `WC_Stripe_Agentic_Shipping_Calculator_Test::test_realistic_customize_checkout_payload_prices_package_contents`. Identifiers and the address are placeholders.
</details>

Covered by unit tests: package-builder shape, shippable-only filtering and variation ID split, calculator `[qty]`/`[cost]`/price-fallback/virtual-exclusion/unresolvable-SKU cases, the realistic full-payload run above, and an order-mapper end-to-end run with a `2 * [qty]` rate.

---

### Changelog entry

Changelog entry added to `changelog.txt` and `readme.txt` under 10.9.0.

**Post merge**


includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,179 @@
...
+						$line_item->get_sku_id()
+					)
+				);
+			}
+
+			$product = WC_Stripe_Agentic_Commerce_Product_Resolver::resolve_product( $product_id );
+
+			if ( ! $product->needs_shipping() ) {
+				continue;
+			}
```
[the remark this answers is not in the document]
wjrosa: `resolve_product()` can't return `null`/`false` — it's typed `: WC_Product` and throws when the product is missing or deleted, and that exception is the intended outcome here (same contract as the sku-miss throw just above, covered by the method's `@throws`). The `instanceof` guard would be dead code, and PHPStan level 8 would flag the always-true `instanceof` as a new error.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,179 @@
...
+
+			$quantity    = $line_item->get_quantity();
+			$unit_amount = $line_item->get_unit_amount();
+			$line_total  = null !== $unit_amount
```
daledupreez: Question: If Stripe hasn't specified a unit amount, should we not throw an exception rather than try to infer the amount based on the current product price? It feels like we could generate mismatched prices in various situations if we rely on local data that may have drifted from the data Stripe has on file.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,179 @@
...
+				? WC_Stripe_Helper::convert_from_stripe_amount( $unit_amount * $quantity, $currency )
+				: (float) $product->get_price() * $quantity;
+
+			$contents[ $line_item->get_id() ] = self::build_contents_entry( $product, $quantity, $line_total, $line_total );
```
daledupreez, line 141: Question: is each line item guaranteed to have a unique ID within an order?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -440,22 +440,11 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
-				'country'  => $address->get_country() ?? '',
-				'state'    => $address->get_state() ?? '',
-				'postcode' => $address->get_postal_code() ?? '',
-				'city'     => $address->get_city() ?? '',
-				'address'  => '',
-			],
-			'cart_subtotal'   => 0,
-		];
+		$package = WC_Stripe_Agentic_Shipping_Package_Builder::build_package(
+			WC_Stripe_Agentic_Shipping_Package_Builder::build_contents_from_order( $order ),
```
daledupreez, line 559: Should we not be using the incoming checkout session data as the source of truth? This approach means that this logic depends on the upstream mapping for the order line items to include the details we need for building shipping data, and we'd need to make sure that we're making that expectation explicit in some way.

malithsen (commented):
```
2. Send a customize_checkout webhook with a line item of quantity 3 — the quoted shipping option should be 6.00 (previously the [qty] expression evaluated against an empty package).
3. Complete an agentic checkout choosing that rate — the created order's shipping line should be the matched WC rate (method flat_rate), not a free-form stripe_agentic fallback line.
```

For step 2 and 3 in the test instructions,

Can you please share the exact webhook payload and the URL. 
Maybe I should already know this, but for step 3, how exactly do I complete an agentic checkout?

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
wjrosa: This `line_total` only feeds the shipping package's `contents_cost`/`cart_subtotal`, never the order or payment totals (those come straight from Stripe), so a stale catalog price can't produce a price mismatch — the only effect would be a subtotal-threshold method (e.g. free shipping over X) seeing a slightly off subtotal. `unit_amount` is normally present; the fallback is defensive. Throwing would abort the whole shipping quote whenever Stripe omits it, which felt worse than a best-effort estimate for shipping-only math. Happy to make it a hard failure instead if you'd rather never infer.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
wjrosa: Yes — Stripe assigns a unique `id` per line item in the session, and the order-path mirror keys on WC's `item_id`, which WC guarantees unique. The key is only structural anyway: downstream (`build_package`, `calculate_shipping`) iterates the values and never indexes by it, so even a hypothetical collision would just merge an entry rather than corrupt the package.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
wjrosa: At this point the order is the post-resolution source of truth: `map_line_items()` runs first and has already resolved each session line item to a WC product and reconciled totals, so `build_contents_from_order()` reuses that rather than re-deriving from raw session data and risking divergence. I made the ordering dependency explicit in f284102 (a comment at the `map_shipping()` call site, plus the existing `build_contents_from_order()` docblock).

wjrosa:
@malithsen testing agentic commerce end-to-end can be involved, so I've made it replayable locally without a real agent. The PR testing instructions now include a signature-verification bypass (a small temporary mu-plugin) plus copy-paste `curl` commands for **both** step 2 (the `customize_checkout` shipping quote) and step 3 (order creation, which additionally mocks the session re-fetch + an Action Scheduler run). That's the quickest way to verify this change, and I shared the sanitized `customize_checkout` payload there too.

For a genuine "real" test you'd still set up your test store for agentic (doing the onboarding). Make sure you are gated into the feature (you may need to ask Stripe to gate your account), sync your catalog, and trigger a test checkout on your product feed: https://dashboard.stripe.com/test/agentic-commerce/catalog/viewer

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,179 @@
...
+
+			if ( ! $product->needs_shipping() ) {
+				continue;
+			}
+
+			$quantity    = $line_item->get_quantity();
+			$unit_amount = $line_item->get_unit_amount();
+			$line_total  = null !== $unit_amount
+				? WC_Stripe_Helper::convert_from_stripe_amount( $unit_amount * $quantity, $currency )
+				: (float) $product->get_price() * $quantity;
```
daledupreez: Concern: in general, `$product->get_price()` will include tax when prices include tax. While that's uncommon in the US, I think we need to be careful about incorrectly using the tax-inclusive value here. Might it be better to call `wc_get_price_excluding_tax()` to protect against that?

_Assisted by AI reviewer_
```suggestion
				: (float) wc_get_price_excluding_tax( $product, ['qty' => $quantity ] );
```

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,179 @@
...
+			}
+
+			$product = WC_Stripe_Agentic_Commerce_Product_Resolver::resolve_product( $product_id );
+
+			if ( ! $product->needs_shipping() ) {
+				continue;
+			}
+
+			$quantity    = $line_item->get_quantity();
+			$unit_amount = $line_item->get_unit_amount();
```
daledupreez, line 98: Note: my AI review tool flagged that we're not applying any bounds checks on these values, we're only casting to `int`s, so negative values would be permitted -- maybe we need a follow-up PR to address that.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
```diff
@@ -118,6 +124,181 @@ public function test_matching_zone_returns_valid_shipping_options( float $cost,
 		$this->assertEquals( 'usd', $fixed_amount['currency'] );
 	}
 
+	/**
+	 * Test that a quantity-based flat rate cost expression ('2 * [qty]') sees the
+	 * package contents quantities, proving contents are populated from the event.
+	 */
+	public function test_qty_cost_expression_uses_package_contents() {
```
daledupreez: Nit: It may be worth consolidating many of these tests using a data provider, as the main logic is pretty much identical across the test methods.

daledupreez: We should also add explicit tests to ensure that we handle multiple valid products correctly. The only multi-product test is adding a product we expect to be excluded.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
```diff
@@ -137,4 +318,139 @@ public function test_returns_empty_for_non_matching_zone() {
...
+	}
+
+	/**
+	 * Returns the sanitized customize_checkout event used by the realistic payload
+	 * test. This is the same JSON shared in the PR description so a reviewer can
+	 * replay it against a store; identifiers and the address are placeholders.
+	 *
+	 * @return \stdClass The decoded webhook event.
+	 */
+	private function realistic_customize_checkout_event(): \stdClass {
```
daledupreez: Nit: maybe call this `get_mock_customize_checkout_event()`? It wouldn't be worth testing if it wasn't realistic! 🙃

daledupreez (commented):
Overall, this is looking pretty good, but I have some feedback.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
wjrosa: Applied, thanks!

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
wjrosa: Agreed — I will handle bounds validation in a follow-up PR.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
wjrosa: Consolidated into a data provider and added multi-product `[qty]`/`[cost]` cases in 4f5a95ef9.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
wjrosa: Renamed in 4f5a95ef9.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,179 @@
...
+			$line_total  = null !== $unit_amount
+				? WC_Stripe_Helper::convert_from_stripe_amount( $unit_amount * $quantity, $currency )
+				: (float) wc_get_price_excluding_tax( $product, ['qty' => $quantity ] );
```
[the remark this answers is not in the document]
wjrosa: Guarded in e98628fa7 — a missing catalog price now throws (covering both the empty-string return on older WC and the 0.0 cast on newer WC), while numeric zero (free products) passes through. Test added.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
wjrosa: Follow-up: #5719.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
```diff
@@ -539,22 +555,11 @@ private function map_shipping( WC_Order $order, WC_Stripe_Agentic_Checkout_Sessi
...
+			$address,
+			$order->get_customer_id()
+		);
```
diegocurbelo, line 565: The initial shipping quote and this recalculation can use different user contexts. The Stripe webhook quotes with `get_current_user_id()` which is guest ID 0, and the previous mapper hardcoded it to 0 also. Could we preserve that here, or otherwise preserve the exact user ID used for the original quote?

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,217 @@
...
+	 * @param WC_Stripe_Agentic_Customize_Checkout_Event $event    The customization hook event.
+	 * @param string                                     $currency The three-letter currency code.
+	 * @return array Cart-item-format entries keyed by line item ID.
+	 * @throws Exception When a line item's sku_id cannot be resolved to a product, its quantity or unit_amount is out of bounds, or the product has no catalog price to fall back on.
+	 */
+	public static function build_contents_from_event( WC_Stripe_Agentic_Customize_Checkout_Event $event, string $currency ): array {
+		$contents = [];
+
+		foreach ( $event->get_line_items() as $line_item ) {
+			$product_id = WC_Stripe_Agentic_Commerce_Product_Resolver::resolve_product_id_by_external_reference( $line_item->get_sku_id() );
```
diegocurbelo: This repeats the SKU resolution already performed by `extract_line_items_from_customization_hook()` immediately before shipping runs and `wc_get_product_id_by_sku()` executes a database query, so this adds one query per line item to every request.

Could we resolve the line-item-to-product-ID map once and pass it to both tax and shipping?

That would also ensure both calculations use the same result if `woocommerce_get_product_id_by_sku` is filtered.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,217 @@
+<?php
+/**
+ * Class WC_Stripe_Agentic_Shipping_Package_Builder
+ *
+ * Builds WC shipping packages with resolved product contents for agentic checkout.
+ *
+ * @package WooCommerce_Stripe/Agentic_Commerce
+ * @since   10.9.0
```
diegocurbelo:
```suggestion
 * @since   11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * Builds WooCommerce shipping packages for agentic commerce flows.
+ *
+ * The agentic checkout flow has no WC_Cart, so packages handed to
+ * WC_Shipping::calculate_shipping() must be assembled by hand. Content-dependent
+ * shipping methods (table rate, weight-based, per-item) read the package's
+ * `contents` entries — product data, quantities, and line totals — so the
+ * entries mirror WC_Cart's cart item format.
+ *
+ * @since 10.9.0
```
diegocurbelo:
```suggestion
 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,217 @@
...
+ */
+class WC_Stripe_Agentic_Shipping_Package_Builder {
+
+	/**
+	 * Builds a WC shipping package array from contents entries and a destination.
+	 *
+	 * `contents_cost` and `cart_subtotal` are derived from the entries' line
+	 * totals, matching how WC_Cart::get_shipping_packages() sums shippable items.
+	 *
+	 * @since 10.9.0
```
diegocurbelo:
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	/**
+	 * Builds package contents from a customize_checkout event's line items.
+	 *
+	 * Resolves each line item's sku_id to a WooCommerce product and keeps only
+	 * shippable products. Line totals come from the event's unit_amount when
+	 * present, falling back to the catalog price otherwise.
+	 *
+	 * @since 10.9.0
```
diegocurbelo:
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-customize-checkout-line-item.php
```diff
@@ -83,4 +83,14 @@ public function get_quantity(): int {
 	public function get_name(): string {
 		return (string) ( $this->item->name ?? '' );
 	}
+
+	/**
+	 * Returns the unit amount in the smallest currency unit, or null when absent.
+	 *
+	 * @since 10.9.0
```
diegocurbelo:
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	/**
+	 * Builds package contents from a mapped order's product line items.
+	 *
+	 * Used by the order mapper, where line items (and their totals) have
+	 * already been resolved and added to the order. Keeps only shippable
+	 * products.
+	 *
+	 * @since 10.9.0
```
diegocurbelo:
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	/**
+	 * Builds a single contents entry in WC_Cart's cart item format.
+	 *
+	 * Shipping methods read `data` (for weight/shipping class), `quantity`,
+	 * `line_total`, and `line_subtotal`; tax fields are zeroed because tax is
+	 * calculated separately in both agentic flows.
+	 *
+	 * @since 10.9.0
```
diegocurbelo:
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php
wjrosa: Good point. 1acd6a12 passes user ID 0 here, matching the guest context the webhook quote always runs in, with a comment and a test that captures the package.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
wjrosa: Done in 1acd6a12: the webhook resolves the map once and passes it to both calculators. `calculate()` and `build_contents_from_event()` take it as an optional argument, so other callers are unchanged. Test added.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-calculator.php
```diff
@@ -46,22 +47,11 @@ public function calculate( WC_Stripe_Agentic_Customize_Checkout_Event $event, st
...
-			'destination'     => [
-				'country'  => $address->get_country() ?? '',
-				'state'    => $address->get_state() ?? '',
-				'postcode' => $address->get_postal_code() ?? '',
-				'city'     => $address->get_city() ?? '',
-				'address'  => '',
-			],
-			'cart_subtotal'   => 0,
-		];
+		$package = WC_Stripe_Agentic_Shipping_Package_Builder::build_package(
```
daledupreez, line 50: Not blocking: Should we offer a filter for splitting up the package into multiple sub-items? Or is it sufficient that the lower-level shipping calculations invoke the `woocommerce_shipping_packages` filter? Many merchants will need to be able to send multiple independent packages for various use cases.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
```diff
@@ -118,6 +124,304 @@ public function test_matching_zone_returns_valid_shipping_options( float $cost,
...
+			$raw_items[] = $raw_item;
+		}
+
+		$result = $this->calculator->calculate( $this->build_event_from_raw_items( $raw_items ), 'usd' );
+
+		$this->assertSame( $expected_amount, $result['shipping_options'][0]['shipping_rate_data']['fixed_amount']['amount'] );
+
+		foreach ( $products as $product ) {
+			$product->delete( true );
+		}
```
daledupreez: Nit: We should either perform cleanup before we assert or use `try/finally`.
```suggestion
		foreach ( $products as $product ) {
			$product->delete( true );
		}

		$this->assertSame( $expected_amount, $result['shipping_options'][0]['shipping_rate_data']['fixed_amount']['amount'] );
```

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
```diff
@@ -118,6 +124,304 @@ public function test_matching_zone_returns_valid_shipping_options( float $cost,
...
+					'id'       => 'li_pre_resolved',
+					'sku_id'   => 'NON-EXISTENT-SKU-' . uniqid(),
+					'quantity' => 1,
+				],
+			]
+		);
+
+		$result = $this->calculator->calculate( $event, 'usd', [ 'li_pre_resolved' => $product->get_id() ] );
+
+		$this->assertNotEmpty( $result['shipping_options'] ?? [] );
```
daledupreez, line 357: Nit: We should delete the product we create in this test.
```suggestion
		$product->delete( true );

		$this->assertNotEmpty( $result['shipping_options'] ?? [] );
```

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
```diff
@@ -137,4 +441,137 @@ public function test_returns_empty_for_non_matching_zone() {
...
+
+		$this->shipping_zone = $this->create_shipping_zone_with_flat_rate( 'US', '2 * [qty]' );
+
+		$raw    = $this->get_mock_customize_checkout_event();
+		$event  = new WC_Stripe_Agentic_Customize_Checkout_Event( $raw );
+		$result = $this->calculator->calculate( $event, $raw->data->currency );
+
+		$this->assertSame( 200, $result['shipping_options'][0]['shipping_rate_data']['fixed_amount']['amount'] );
+
+		$product->delete( true );
```
daledupreez: We should either delete the product first or use `try/finally`.
```suggestion
		$product->delete( true );

		$this->assertSame( 200, $result['shipping_options'][0]['shipping_rate_data']['fixed_amount']['amount'] );
```

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
```diff
@@ -137,4 +441,137 @@ public function test_returns_empty_for_non_matching_zone() {
 			$this->assertEmpty( $result );
 		}
 	}
+
+	/**
+	 * Test that a full customize_checkout webhook payload parses end-to-end: its
```
daledupreez: This test description is somewhat misleading, as it is only "end-to-end" within the shipping context.
```suggestion
	 * Test that a full customize_checkout webhook payload computes the correct shipping price: its
```

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder-test.php
```diff
@@ -0,0 +1,135 @@
...
+		$this->assertSame( $shippable->get_id(), $entry['product_id'] );
+		$this->assertSame( 0, $entry['variation_id'] );
+		$this->assertSame( 3, $entry['quantity'] );
+		$this->assertSame( $shippable->get_id(), $entry['data']->get_id() );
+		$this->assertSame( 30.0, $entry['line_total'] );
+		$this->assertSame( 30.0, $entry['line_subtotal'] );
+
+		$order->delete( true );
+		$shippable->delete( true );
+		$virtual->delete( true );
```
daledupreez: We should use `try/finally` here given that we use `$shippable` for assertions.
```suggestion
		try {
			$this->assertCount( 1, $contents );

			$entry = reset( $contents );
			$this->assertSame( $shippable->get_id(), $entry['product_id'] );
			$this->assertSame( 0, $entry['variation_id'] );
			$this->assertSame( 3, $entry['quantity'] );
			$this->assertSame( $shippable->get_id(), $entry['data']->get_id() );
			$this->assertSame( 30.0, $entry['line_total'] );
			$this->assertSame( 30.0, $entry['line_subtotal'] );
		} finally {
			$order->delete( true );
			$shippable->delete( true );
			$virtual->delete( true );
		}
```

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder-test.php
```diff
@@ -0,0 +1,135 @@
...
+
+		$this->assertCount( 1, $contents );
+
+		$entry = reset( $contents );
+		$this->assertSame( $variable->get_id(), $entry['product_id'] );
+		$this->assertSame( $variation->get_id(), $entry['variation_id'] );
+		$this->assertSame( 2, $entry['quantity'] );
+
+		$order->delete( true );
+		$variable->delete( true );
```
daledupreez: We should use `try/finally` for the cleanup given that we use `$variable` and `$variation` in the assertions.
```suggestion
		try {
			$this->assertCount( 1, $contents );

			$entry = reset( $contents );
			$this->assertSame( $variable->get_id(), $entry['product_id'] );
			$this->assertSame( $variation->get_id(), $entry['variation_id'] );
			$this->assertSame( 2, $entry['quantity'] );
		} finally {
			$order->delete( true );
			$variable->delete( true );
		}
```

daledupreez: Updated in 193aa52f23744c9e9479982663a4d88499396b1e.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder-test.php
daledupreez: Updated in 193aa52f23744c9e9479982663a4d88499396b1e.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
daledupreez: Updated in 193aa52f23744c9e9479982663a4d88499396b1e.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
daledupreez: Updated in 193aa52f23744c9e9479982663a4d88499396b1e.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-shipping-calculator-test.php
daledupreez: Updated in 193aa52f23744c9e9479982663a4d88499396b1e.

daledupreez (approved):
I've tested through step 2 (the mock local test), and I think we can get this merged.

includes/agentic-commerce/class-wc-stripe-agentic-shipping-calculator.php
wjrosa: `woocommerce_shipping_packages` runs only after each package is priced, and the real split filter (`woocommerce_cart_shipping_packages`) never fires here because there is no cart. So I added a `wc_stripe_agentic_shipping_packages` filter. It can split the package before calculation, and rates available for every package are offered, with costs summed

PR #5552, opened 2026-06-10
Fix uncaught "Stripe initialization data is not available" error on product pages with express checkout

Mayisha:
Fixes STRIPE-1206

This PR:
  
- Builds the express checkout API client from `wc_stripe_express_checkout_params` (which already carries the publishable key, locale, and AJAX URL) instead of `wc_stripe_upe_params`, removing the cross-bundle dependency.
- Makes `getStripeServerData()` return `null` instead of throwing when the data is unavailable, so the remaining optional-chaining call sites degrade gracefully rather than producing an uncaught error. Classic checkout and Blocks always localize their data, so those paths are unchanged.

## Testing instructions
1. Enable Express Checkout (Apple Pay / Google Pay) for the **product** page in the Stripe settings.
2. Simulate a scenario where `wc_stripe_upe_params` is not output on the product page — e.g., comment out the code where the param is localized.
3. Open a product page with the browser console open.
4. **Before:** `Uncaught Error: Stripe initialization data is not available`; the express button does
     not render. **After:** no console error, and the express checkout button initializes and is clickable.
5. Regression check: on classic checkout and Blocks checkout, confirm that card and express checkout still work normally.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


malithsen (approved):
Looks good

PR #5553, opened 2026-06-10
Hide the save payment method checkbox for SEPA-converting methods in the Adaptive Pricing checkout

wjrosa:
Follow-up to #5550. Part of STRIPE-1205.

With Optimized Checkout + Adaptive Pricing enabled, the checkout renders Stripe's Checkout Sessions form. In that flow, Stripe currently provides **no way to save Bancontact, iDEAL or Sofort as SEPA Direct Debit tokens** (verified against the Stripe test API, pinned version `2025-09-30.clover`):

- Session **update** with `payment_method_options` → `parameter_unknown`.
- Session **creation** with `payment_method_options[bancontact][setup_future_usage]=off_session` → `Invalid …: must be none`.
- Session **creation** with `payment_intent_data[setup_future_usage]=off_session` → ``unsupported for payment method `bancontact` ``.
- Client-side `checkout.confirm( { savePaymentMethod } )` only applies to cards per [Stripe's docs](https://docs.stripe.com/payments/checkout/save-during-payment?payment-ui=checkout-form).

So the "Save payment information" checkbox was a promise we could not keep: the shopper ticked it, the payment completed, and no token was ever created (after #5550, the webhook safely skips the save; before it, it fataled).

This PR hides and clears the save checkbox when one of these methods is selected while Adaptive Pricing is active: when building the OC `showSaveOptionByMethod` map, `get_enabled_payment_method_config()` now sets a method's value to `false` when Adaptive Pricing is supported and the method is saved as a different type (`get_id() !== get_retrievable_type()`). The frontend is unchanged — it already hides and clears the checkbox from that map on both Blocks and classic checkout.

**How can this code break / what prevents it:** without Adaptive Pricing, the map is built as before — covered by a test asserting Bancontact/iDEAL stay savable in the deferred-intent flow. Methods saved under their own type (card, SEPA) are unaffected (also covered). One accepted trade-off: the map is baked at page render, so on the rare runtime fallback from a failed Checkout Session to the deferred-intent flow the checkbox stays hidden even though saving would work there — an under-offer, not a broken promise. If Stripe later adds save support for these methods, removing the Adaptive Pricing condition from the map builder restores the checkbox with no client changes.

## Testing instructions

Prerequisites:
1. WooCommerce > Settings > Payments > Stripe > Advanced: **Optimized Checkout Suite** ON and **Adaptive Pricing** ON.
2. Enable **Bancontact** and "Save Bancontact payments as SEPA Direct Debit tokens".
3. Store currency **EUR**; log in as a customer.

Steps:
1. On the **Block** checkout, select **Bancontact** inside the payment element: the "Save payment information to my account for future purchases." checkbox must disappear. Select **card**: it must reappear.
2. Repeat on the **classic (shortcode)** checkout: same behavior.
3. Turn **Adaptive Pricing** OFF (deferred-intent flow): the checkbox must show for Bancontact again, and saving must still produce a SEPA token after payment.
4. Regression: with Adaptive Pricing ON, a card payment with the box ticked still saves a card token.

---

### Changelog entry

* Fix - Hide the save payment method checkbox for Bancontact, iDEAL and Sofort in the Adaptive Pricing checkout, where Stripe cannot save them as SEPA Direct Debit tokens

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -1490,29 +1490,28 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
 		$order_helper->update_stripe_checkout_session_id( $order, $checkout_session_id );
 
-		// Persist the 'save-payment-method' flag so the webhook handler can create the token once payment is completed.
-		$upe_payment_method = $this->payment_methods[ $selected_payment_type ] ?? null;
-		if ( $save_payment_method && $upe_payment_method && $upe_payment_method->get_id() === $upe_payment_method->get_retrievable_type() ) {
-			$order_helper->update_should_save_stripe_payment_method( $order, true );
-		}
-
-		// Eagerly set the payment method title from the customer's selection so the order
-		// reflects the right method as soon as it lands on the order-received page, even if
-		// the checkout.session.completed webhook is delayed. Prefer the type the customer
-		// actually picked inside the Stripe Element (sent via the hidden
-		// `wc_stripe_selected_upe_payment_type` input by the Optimized Checkout JS) over
-		// the gateway-derived `$selected_payment_type`, which for Optimized Checkout is
-		// always 'card' and would otherwise resolve to the OC pseudo-method's "Stripe"
-		// title. The webhook handler will reaffirm/refine this once the actual payment
-		// method type is known from the intent.
+		// Resolve the method the customer actually picked: for Optimized Checkout the gateway type is
+		// always 'card', so prefer the hidden `wc_stripe_selected_upe_payment_type` input when present.
 		// phpcs:ignore WordPress.Security.NonceVerification.Missing
 		$selected_upe_payment_type = isset( $_POST['wc_stripe_selected_upe_payment_type'] )
 			? sanitize_text_field( wp_unslash( $_POST['wc_stripe_selected_upe_payment_type'] ) )
 			: '';
 
-		$title_payment_type = ( ! empty( $selected_upe_payment_type ) && isset( $this->payment_methods[ $selected_upe_payment_type ] ) )
+		$resolved_payment_type   = ( ! empty( $selected_upe_payment_type ) && isset( $this->payment_methods[ $selected_upe_payment_type ] ) )
 			? $selected_upe_payment_type
 			: $selected_payment_type;
+		$resolved_payment_method = $this->payment_methods[ $resolved_payment_type ] ?? null;
+
+		// Persist the 'save-payment-method' flag so the webhook handler can create the token once payment is completed.
+		// Saving itself is requested client-side via `checkout.confirm( { savePaymentMethod } )`, which Stripe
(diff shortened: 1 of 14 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -955,7 +963,8 @@ private function get_enabled_payment_method_config() {
 			];
 
 			if ( ! empty( $show_save_option_by_method ) && $payment_method instanceof WC_Stripe_UPE_Payment_Method_OC ) {
-				$settings[ $payment_method_id ]['showSaveOptionByMethod'] = $show_save_option_by_method;
+				$settings[ $payment_method_id ]['showSaveOptionByMethod']      = $show_save_option_by_method;
+				$settings[ $payment_method_id ]['savedAsDifferentTypeMethods'] = $saved_as_different_type_methods;
```
daledupreez: Doesn't `showSaveOptionByMethod()` already encapsulate what we are trying to do? We can already determine that Adaptive Pricing is enabled, and the payment method will resolve to SEPA. So it feels like expanding that existing logic will be simpler than adding a secondary flag that needs to be checked in multiple places.

wjrosa: Makes sense! Done in cfe531979

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -930,12 +930,17 @@ private function get_enabled_payment_method_config() {
...
+		// Checkout Sessions flow cannot request `setup_future_usage` for them.
 		$show_save_option_by_method = [];
 		if ( $this->oc_enabled && $this->is_valid_optimized_checkout_page() ) {
+			$is_adaptive_pricing_active = $this->is_adaptive_pricing_supported();
 			foreach ( $original_method_ids as $method_id ) {
 				if ( isset( $this->payment_methods[ $method_id ] ) ) {
-					$show_save_option_by_method[ $method_id ] = $this->should_upe_payment_method_show_save_option( $this->payment_methods[ $method_id ] );
+					$payment_method                           = $this->payment_methods[ $method_id ];
+					$show_save_option_by_method[ $method_id ] = $this->should_upe_payment_method_show_save_option( $payment_method )
+						&& ! ( $is_adaptive_pricing_active && $payment_method->get_id() !== $payment_method->get_retrievable_type() );
```
malithsen: Non-blocking-nit: Maybe we can extract these conditions to separate variables to improve readability

```
$is_saved_as_different_type       = $payment_method->get_id() !== $payment_method->get_retrievable_type();
$is_savable_with_adaptive_pricing = ! ( $is_adaptive_pricing_active && $is_saved_as_different_type );

$show_save_option_by_method[ $method_id ] = $is_savable_with_adaptive_pricing
	&& $this->should_upe_payment_method_show_save_option( $payment_method );
```

malithsen (approved):
Looks good, works as described. One detail that tripped me for a bit while testing is that we need to have Link disabled to make "Save payment information ..." checkbox to show up for cards.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Good idea. Applied in 78dc7383e164d78769943dcc2135fd3134d0cdf8

PR #5556, opened 2026-06-11
Fix classic checkout blocking card/Link when the order has no billing country

Mayisha:

Classic (shortcode) checkout was failing for orders with no billing country, with the error *"The payment method type 'card' is not available in ."* (note the empty country). It hit card and Link — methods that have no country restriction at all — and aborted the payment **before** the PaymentIntent was created, so the customer couldn't complete checkout.

Root cause is in `WC_Stripe_UPE_Payment_Method::is_available_for_billing_country()` (introduced in 10.8.0). It returned `false` for **every** method when the billing country was empty:

```php
  if ( '' === $country_code ) {
      return false; // blocked even unrestricted methods like card/Link
  }
```

validate_selected_payment_method_type() then threw and stopped the payment. Because card/Link carry an empty supported_billing_countries (available everywhere), an unknown billing country should never have blocked them.

This PR reorders the check so a method with no country restriction is available regardless of the billing country (including when it's empty/unknown). Restricted methods (e.g. Klarna, BLIK) keep the previous behaviour: an unknown country still returns false, since a country-gated method can't be confirmed without one.

## Testing instructions
- Code review
- Regression check: a country-restricted method (e.g. Klarna/BLIK) with a valid supported billing country still works, and with an unsupported/empty country is still correctly blocked.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-upe-payment-method.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-method.php
@@ -469,16 +469,20 @@ abstract class WC_Stripe_UPE_Payment_Method extends WC_Payment_Gateway {
 	 * @return bool True when no restriction is set or the country is in the supported list.
 	 */
 	public function is_available_for_billing_country( $country_code ): bool {
+		// Methods with no country restriction (e.g. card, Link) are available everywhere,
+		// including when the billing country is unknown — an empty country must not block
+		// them, or checkout fails for orders that carry no billing country.
+		if ( empty( $this->supported_billing_countries ) ) {
+			return true;
+		}
+
+		// A restricted method can't be confirmed available without knowing the country.
 		$country_code = is_string( $country_code ) ? strtoupper( $country_code ) : '';
 		if ( '' === $country_code ) {
 			return false;
 		}
 
-		if ( ! empty( $this->supported_billing_countries ) ) {
-			return in_array( $country_code, $this->supported_billing_countries, true );
-		}
-
-		return true;
+		return in_array( $country_code, $this->supported_billing_countries, true );
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-method.php
```diff
@@ -469,16 +469,20 @@ public function is_allowed_on_country( $country ) {
 	 * @return bool True when no restriction is set or the country is in the supported list.
 	 */
 	public function is_available_for_billing_country( $country_code ): bool {
+		// Methods with no country restriction (e.g. card, Link) are available everywhere,
+		// including when the billing country is unknown — an empty country must not block
+		// them, or checkout fails for orders that carry no billing country.
+		if ( empty( $this->supported_billing_countries ) ) {
```
daledupreez: Tiny non-blocking nit: I have a slight preference for checking if we have `[]` rather than the broader `empty()` check. But definitely not blocking.
```suggestion
		if ( [] === $this->supported_billing_countries ) {
```

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-method-test.php
```diff
@@ -1209,27 +1209,37 @@ public function test_is_available_for_billing_country( array $supported_billing_
...
+				[],
+				'',
+				true,
+			],
+			'populated list rejects unknown country' => [
+				[ 'US', 'CA' ],
+				'',
+				false,
+			],
+			'populated list matches'                 => [
```
daledupreez, line 1242: Not blocking, but it may also be worth adding test cases for non-string values alongside these two new checks.
```suggestion
			'empty list permits null country'        => [
				[],
				null,
				true,
			],
			'populated list rejects null country'        => [
				[ 'US', 'CA', 'GB' ],
				null,
				false,
			],
			'populated list matches'                 => [
```

daledupreez (approved):
Thanks for catching this, @Mayisha! The fix looks good to me.

PR #5558, opened 2026-06-11
Add check for Agentic Commerce feature flag before running fallback logic

daledupreez:
## Changes proposed in this Pull Request:

This PR adds a check for the Agentic Commerce feature flag in our webhook handling to prevent unnecessary logging in the downstream code. The specific code path is when we receive a `checkout.session.completed` webhook, but can't locate the relevant order. Prior to this change, we would call `WC_Stripe_Webhook_Handler->handle_agentic_checkout_session()` which would then log an error message when the Agentic Commerce feature flag is disabled. This PR checks the feature flag is enabled before calling `handle_agentic_checkout_session()`.

## Testing instructions

Code inspection should be sufficient.

---

-   [N/A] Covered with tests (or have a good reason not to test in description ☝️)
-   [N/A] Tested on mobile (or does not apply)

### Changelog entry

-   [x] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 
This is a pretty tiny cleanup change -- I don't think it's worth adding a changelog for.
</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5559, opened 2026-06-11
Fix Adaptive Pricing checkout sessions being misclassified as agentic in webhooks

Mayisha:

In `handle_checkout_session_success()`, when a `checkout.session.completed` event couldn't be matched to a WooCommerce order, the handler assumed the session was **agentic** and routed it to `handle_agentic_checkout_session()`. That handler returns early when agentic commerce is disabled (`Agentic commerce is disabled, skipping…`) — *before* ever checking whether the session is actually agentic. So a paid Adaptive Pricing session whose order lookup was missed was discarded. (Worse, with agentic commerce *enabled*, it would have been run through the agentic order mapper and produced an incorrect order.)

 - Tag Adaptive Pricing checkout sessions with `metadata.checkout_type = adaptive_pricing_checkout` at creation, and promote the value to a shared `public const` so the producer (AJAX handler) and consumer (webhook handler) can't drift.
- In the webhook handler, before routing an order-less completed session to the agentic path, read `metadata->checkout_type`. An Adaptive Pricing session is never agentic, so log a warning and return instead of misrouting it.

## Testing instructions
1. Enable Adaptive Pricing / Optimized Checkout on a Stripe account, with **agentic commerce disabled**.
2. Simulate the edge case where a completed Adaptive Pricing `checkout.session.completed` webhook arrives with no matching WooCommerce order (e.g. the order wasn't created/linked) (comment out the order update with cs ID in `process_payment_with_checkout_session` method)
3. **Before this PR:** the log shows `Agentic commerce is disabled, skipping agentic checkout session: cs_…` and the session is dropped.
4. **With this PR:** the log instead shows `Completed Adaptive Pricing checkout session has no matching order: cs_…` and the agentic path is not invoked.
5. Regression check — a normal Adaptive Pricing checkout (order created during payment) still completes and is marked paid; a genuine agentic session (agentic commerce enabled) still processes as before.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
+++ b/includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
@@ -11,8 +11,6 @@ class WC_Stripe_Checkout_Sessions_Ajax_Handler {
 	/**
 	 * Checkout session metadata value identifying an Adaptive Pricing checkout.
 	 *
-	 * Set on the session at creation and read back in the webhook handler so a
-	 * completed Adaptive Pricing session is never mistaken for an agentic one.
 	 *
 	 * @var string
 	 */
--- a/includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
+++ b/includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
@@ -11,7 +11,6 @@ class WC_Stripe_Checkout_Sessions_Ajax_Handler {
 	/**
 	 * Checkout session metadata value identifying an Adaptive Pricing checkout.
 	 *
-	 *
 	 * @var string
 	 */
 	public const ADAPTIVE_PRICING_CHECKOUT_TYPE = 'adaptive_pricing_checkout';
--- a/includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
+++ b/includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
@@ -9,6 +9,16 @@ if ( ! defined( 'ABSPATH' ) ) {
  */
 class WC_Stripe_Checkout_Sessions_Ajax_Handler {
 	/**
+	 * Checkout session metadata value identifying an Adaptive Pricing checkout.
+	 *
+	 * Set on the session at creation and read back in the webhook handler so a
+	 * completed Adaptive Pricing session is never mistaken for an agentic one.
+	 *
+	 * @var string
+	 */
+	public const ADAPTIVE_PRICING_CHECKOUT_TYPE = 'adaptive_pricing_checkout';
+
+	/**
 	 * Initialize hooks.
 	 *
 	 * @return void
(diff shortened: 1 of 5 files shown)


daledupreez:
As a quick FYI I made a small related change that avoids error messages in #5558, but this PR is looking way more promising overall.

readme.txt
```diff
@@ -165,6 +165,8 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Fix - Send the billing address to Stripe on the Pay for Order page so payments aren't incorrectly blocked by Stripe Radar rules
 * Update - Replace shipping AJAX endpoints with Store API calls for Express Checkout Element
 * Tweak - Consolidate the default payment intent metadata fields into a shared method so they stay consistent across payment flows
+* Fix - Prevent a fatal error when a Bancontact, iDEAL or Sofort payment is placed with SEPA token saving enabled through the Adaptive Pricing checkout, which left the order unpaid
```
wjrosa: I think we should leave a single changelog entry

wjrosa (commented):
Nice, targeted fix — tagging AP sessions and reading the tag back is the right call, and the shared `public const` keeps producer/consumer from drifting. Two inline notes (one actionable on the changelog).

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1826,6 +1826,18 @@ protected function handle_checkout_session_success( object $notification ): void
 		// Look for an order. If one does not exists, this is probably an agentic hook.
 		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $checkout_session->id );
 		if ( ! $order instanceof \WC_Order ) {
+			// An Adaptive Pricing session is tagged at creation. It is never agentic,
+			// so don't route it into the agentic handler — that would silently drop a
+			// paid session whenever agentic commerce is disabled.
+			$checkout_type = $checkout_session->metadata->checkout_type ?? '';
```
wjrosa, line 1832: Verified this holds on the deferred path too: `defer_webhook_processing()` serializes the full notification and `normalize_deferred_webhook_notification_to_object()` JSON-decodes `metadata` back to a stdClass, so `->checkout_type` is intact in the Action Scheduler run; `??` also guards the whole chain when metadata is absent. The manual `delete( $lock_key )` before the early return correctly mirrors the agentic branch's `finally`. 👍

readme.txt
Mayisha: It mistakenly came here during the merge from 10.8.1. Removing this entry.

Mayisha: I also changed the actual changelog statement in [166aeed](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5559/commits/166aeed12c29de584b04e58c9db19440b321f6f5)

includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php
```diff
@@ -8,6 +8,16 @@
  * WC_Stripe_Checkout_Sessions_Ajax_Handler class.
  */
 class WC_Stripe_Checkout_Sessions_Ajax_Handler {
+	/**
+	 * Checkout session metadata value identifying an Adaptive Pricing checkout.
+	 *
+	 * Set on the session at creation and read back in the webhook handler so a
+	 * completed Adaptive Pricing session is never mistaken for an agentic one.
```
malithsen:
```suggestion

```

nit: The way we consume this metadata can easily change. I'd remove this line so that we don't have to keep this up to date with consumer changes.

malithsen (approved):
Looks good!

wjrosa (approved):
Looks good to me! 👍

PR #5562, opened 2026-06-11
Guard against null customer

malithsen:
Fixes STRIPE-1207

During a subscription purchase, the gateway saves the Stripe payment method details onto the subscription, including the Stripe customer ID. A payment method doesn't always have a customer attached to it yet, so that ID can be missing.

This PR adds a guard so we only save the customer ID when one is actually present. We already have similar checks everywhere else we save a customer. 

Also the skipping is safe. Renewal payments don't depend on the customer. When it's missing they fall back to the
customer ID [on the user account](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/8c8d95631427d588800f024dda1d8a2b95608d1c/includes/abstracts/abstract-wc-stripe-payment-gateway.php#L433). 

## Testing instructions

I've not been able to replicate the issue locally. Code inspection should be sufficient.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


diegocurbelo (approved):
The change looks good to me, the new guard makes sense 👍🏼

I wasn't able to reproduce the problem either, I tried using a guest customer (incognito + checkout sessions), but did not get any errors in `develop` either, it correctly used the `guest` grouping user in Stripe

Image

malithsen:
I will merge this, even though we cannot replicate the error. We have similar guards in other places where we update the customer.

PR #5568, opened 2026-06-12
Harden Adaptive Pricing checkout against third-party Stripe.js version conflicts

Mayisha:
Fixes STRIPE-STRIPE-1209

The plugin pins Stripe.js to the `clover` build, but a third-party script on the
store (e.g. a "Stripe Radar" snippet loading `https://js.stripe.com/dahlia/stripe.js`)
can replace the shared `window.Stripe` global with a different version — newer or
older. Because Stripe.js exposes a single global with no per-version namespace,
the plugin then runs against that other build and hits API differences it doesn't
expect — crashing checkout. This PR makes the checkout flows degrade gracefully
instead.

Changes:

- **`layout.radios` as a string enum.** Newer Stripe.js rejects the boolean `layout.radios` and requires `'always' | 'never' | 'auto' | 'if_multiple'`. Pass `'never'`/`'always'` instead of `false`/`true` in both classic and Blocks Optimized Checkout. Same rendered result.
- **Skip Adaptive Pricing when `initCheckout` is unusable (classic + Blocks).** AP uses `stripe.initCheckout()`, which newer Stripe.js keeps only as a stub that throws "has been removed" (replaced by `initCheckoutElementsSdk`). Method presence can't distinguish the versions (the stub is still a function), so we detect the replacement method and fall back to the standard elements flow before `<CheckoutProvider>` mounts. This has to be decided synchronously — the failure surfaces as an unhandled promise rejection inside the provider, so no error boundary or `onLoadError` can catch it after the fact.
- **Null-guard the optimized-checkout iframe.** `applyStyles()` assumed `.wcstripe-payment-element iframe` always exists and threw when it didn't.

**Scope**: the `initCheckout` guard handles both directions (older builds without the method and newer builds where it throws both fall back). The `layout.radios` change targets newer builds — an older build that predates the string enum accepts only the boolean, so a single hardcoded value can't satisfy both directions. Fully covering the older direction needs the more robust fix: capture the pinned `clover` factory at load time and route all Stripe access through it instead of the shared global, so the plugin always runs `clover` regardless of what else loads. Out of scope here; tracked as a follow-up.

## Testing instructions
- On a store with WooCommerce Blocks checkout and Optimized Checkout enabled, add a code snippet that loads a newer Stripe.js on the checkout page: `<script async src="https://js.stripe.com/dahlia/stripe.js"></script>`

```
add_action( 'wp_enqueue_scripts', function () {
        // Radar collects fraud signals from Stripe.js; load it on the checkout
        // page for both classic and block checkout (block checkout still uses
        // the WC checkout page, so is_checkout() matches both).
        if ( ! function_exists( 'is_checkout' ) || ! is_checkout() ) {
                return;
        }

        wp_enqueue_script(
                'stripe-radar',
                'https://js.stripe.com/dahlia/stripe.js',
                array(),
                null, 
                array(
                        'strategy'  => 'async',
                        'in_footer' => false,
                )
        );
  } );
```
- In `develop`
    - With OC enabled and AP disabled, the Payment Element fails to render and the console shows `Uncaught IntegrationError: Invalid value for elements.create('payment'): layout.radios should be one of the following strings: always, never, auto, if_multiple. You specified: false.` and `Uncaught TypeError: Cannot read properties of null (reading 'style')` errors.
    - With OC and AP both enabled, the Payment Element fails to render and the console shows `IntegrationError: stripe.initCheckout() has been removed`.
- In this branch, the payment element should load correctly. 

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/classic/upe/payment-processing.js
```diff
@@ -299,12 +299,16 @@ async function createStripePaymentElement( api, paymentMethodType ) {
...
 	let elements;
 	let shouldLoadStripeElements = true;
 	// If Adaptive Pricing is enabled, use the Checkout Session API to load the elements.
+	// initCheckout() is gone in newer Stripe.js (dahlia+); when a third-party script
+	// loads such a build, skip AP and fall back to the standard elements flow below.
 	if (
 		stripeServerData?.isAdaptivePricingEnabled &&
-		supportsDeferredIntent
+		supportsDeferredIntent &&
+		typeof stripe?.initCheckout === 'function'
```
wjrosa: Unlike the Blocks guard, this only checks `initCheckout` presence. But per the PR description dahlia+ keeps `initCheckout` as a throwing stub, so this stays `true` there — classic enters the AP branch, fires `checkoutSessionsCreateSession()` (creating an orphan Checkout Session server-side), then only falls back once `initCheckout()` throws + `console.error`s, on every dahlia+ classic load. Mirroring the Blocks check (`&& typeof stripe?.initCheckoutElementsSdk !== 'function'`) would skip the wasted session-create and the error log. The try/catch keeps it working today, but it's doing real work before failing.

client/classic/upe/__tests__/payment-processing.test.js
```diff
@@ -432,6 +432,23 @@ describe( 'payment-processing', () => {
 				expect( api._stripe.elements ).not.toHaveBeenCalled();
 			} );
 
+			it( 'skips Adaptive Pricing and loads standard elements when initCheckout is unavailable (dahlia+)', async () => {
+				const checkoutElements = createMockElements();
+				const api = createMockApi( checkoutElements );
+				// Newer Stripe.js removed initCheckout(); the gateway must degrade
+				// to the standard elements flow instead of crashing checkout.
+				delete api._stripe.initCheckout;
```
wjrosa, line 461: `delete`-ing `initCheckout` simulates the "method absent" build, but dahlia+ (per the description) keeps it as a throwing stub — that's exactly why the Blocks side needs the `initCheckoutElementsSdk` check. So this covers a branch real dahlia+ never takes; the actual build flows through the try/catch above. Worth renaming away from "(dahlia+)" or switching to a throwing stub so the test matches the scenario it claims (also lines up with CodeRabbit's ask for the throwing-path case).

wjrosa (commented):
One real asymmetry between the classic and Blocks guards, plus a follow-on in the classic test. (Skipping the test-coverage asks CodeRabbit already made.)

client/classic/upe/payment-processing.js
Mayisha: Good catch. Classic only checks `initCheckout` presence, so on dahlia+ (throwing stub) it still enters the AP branch, creates an orphan Checkout Session, and only bails after initCheckout() throws + logs. I'll mirror the Blocks guard with `&& typeof stripe?.initCheckoutElementsSdk !== 'function'` so it skips the session-create entirely, and switch the classic test from delete-ing initCheckout to a throwing stub + `initCheckoutElementsSdk` present so it actually exercises the dahlia+ path (also covers CodeRabbit's throwing-path ask).

client/classic/upe/__tests__/payment-processing.test.js
Mayisha: Update the classic test to exercise the real
dahlia+ throwing-stub path and add a separate older-build (initCheckout absent) case in [e69cd9c](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5568/commits/e69cd9c4b11534350620f6560fc654c8c99dc524)

client/classic/upe/payment-processing.js
Mayisha: Done [e69cd9c](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5568/commits/e69cd9c4b11534350620f6560fc654c8c99dc524)

client/optimized-checkout/apply-styles.js
```diff
@@ -21,6 +21,9 @@ export const applyStyles = () => {
 	const stripeIframe = document.querySelector(
 		'.wcstripe-payment-element iframe'
 	);
-	stripeIframe.style.margin = 0;
-	stripeIframe.style.width = '100%';
+
+	if ( stripeIframe ) {
+		stripeIframe.style.margin = 0;
+		stripeIframe.style.width = '100%';
+	}
```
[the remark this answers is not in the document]
Mayisha: Added in [bd08b01](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5568/commits/bd08b016fbe432d647a258b3f37e94586e945cfd)

wjrosa (approved):
Looks good to me! Thanks for addressing the comments

PR #5570, opened 2026-06-12
Update stable tag to 10.7.0

wjrosa:
This PR updates the stable tag to 10.7.0


malithsen (approved):

wjrosa:
Since we released 10.8.2 right after (updating the stable tag again), this PR is no longer needed.

PR #5571, opened 2026-06-12
 Prevent Adaptive Pricing from being enabled when Stripe webhooks aren't working

malithsen:

Adaptive Pricing depends on webhooks to mark orders as paid. If a store's webhooks aren't working, Stripe still charges the customer but the order will not be updated on the site and will stay in "Pending Payment" status. This PR ties Adaptive Pricing availability to webhook status so that can't happen.

CleanShot 2026-06-12 at 17 45 17@2x

With this change:

  - A merchant can no longer turn Adaptive Pricing on while their webhooks are disabled. 
  - For stores that already had Adaptive Pricing on, it stops being applied at checkout.
  - Adaptive Pricing default-on migration only applies when webhooks are enabled.
  - First-time onboarding still enables Adaptive Pricing as before

## Testing instructions

1. Manually bump the version to 10.8.2. 
2. Generate a build of this branch.

### (Regression) Test new merchants
1. Onboard as a new merchant using this branch.
2. Both OCS + AP should be enabled

### (Regression) Test backbook default-on migration
1. Setup a test store with Stripe 10.7.0 
2. Make sure Adaptive Pricing is disabled.
3. Upgrade to this build.
4. AP should be enabled

### Test backbook default-on migration with webhook disabled
1. Set up a test store with Stripe 10.7.0 
2. Make sure Adaptive Pricing is disabled.
3. Disable Webhooks for the site via the Stripe dashboard.
4. Upgrade to this build.
5. AP should NOT be enabled

### Test merchants who already have AP on but have webhooks disabled
1. Setup a site with a build from this branch
2. Enable AP.
3. Disable Webhooks via the Stripe dashboard.
4. AP should appear disabled on the merchant site and should not be available to shoppers.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -1190,6 +1190,10 @@ class WC_Stripe_Helper {
 			return 'account-country';
 		}
 
+		if ( ! $stripe_account->is_webhook_enabled() ) {
+			return 'webhooks-disabled';
+		}
+
 		// If we are in test mode, payout details are often missing and currency-based rules
 		// are not enforced.
 		if ( WC_Stripe_Mode::is_test() ) {
--- a/includes/connect/class-wc-stripe-connect.php
+++ b/includes/connect/class-wc-stripe-connect.php
@@ -303,14 +303,8 @@ if ( ! class_exists( 'WC_Stripe_Connect' ) ) {
 
 			$this->clear_caches_after_key_update();
 
-			// Runs after the keys are saved: the Adaptive Pricing decision needs the account data.
 			if ( 'connect' === $type && $should_default_optimized_checkout_on ) {
 				$options['optimized_checkout_element'] = 'yes';
-				if ( WC_Stripe_Helper::is_adaptive_pricing_available_for_account() ) {
-					$options['adaptive_pricing'] = 'yes';
-				} else {
-					WC_Stripe_Logger::info( 'OAuth: Not defaulting Adaptive Pricing on; it is not available for the connected account.' );
-				}
 				WC_Stripe_Helper::update_main_stripe_settings( $options );
 			}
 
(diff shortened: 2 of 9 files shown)


includes/class-wc-stripe-helper.php
```diff
@@ -1190,6 +1190,10 @@ public static function get_adaptive_pricing_account_unavailable_reason(): ?strin
 			return 'account-country';
 		}
 
+		if ( ! $stripe_account->is_webhook_enabled() ) {
```
wjrosa, line 1193: This now runs on every checkout render via `is_adaptive_pricing_supported()`. `is_webhook_enabled()` is transient-cached (2h) but does a synchronous Stripe GET on a cold cache, and returns `false` on any API error — so a transient blip silently drops AP for that render. Fail-closed is reasonable; just flagging the added checkout-path call.

Also note this check sits before the `is_test()` early-return, so AP now requires webhooks in test mode too (intentional per the new data-provider case).

includes/connect/class-wc-stripe-connect.php
```diff
@@ -355,6 +349,19 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 				WC_Stripe_API::set_secret_key( '' );
 			}
 
+			// Default Adaptive Pricing on for first-time connections, now that webhooks have been
+			// configured above. AP requires a working webhook endpoint, so this decision must run after
+			// configure_webhooks()
+			if ( 'connect' === $type && $should_default_optimized_checkout_on ) {
```
wjrosa, line 355: Good call moving this after `configure_webhooks()`. One edge: if `configure_webhooks()` throws (e.g. unreachable local/test sites), the `catch` returns a `WP_Error` above and this block is skipped, so AP won't default on there even if otherwise eligible. Likely acceptable given the intent.

wjrosa (commented):
A few inline notes — overall the approach (gating AP on webhook status and deferring the default-on decision until after `configure_webhooks()`) looks sound.

includes/connect/class-wc-stripe-connect.php
malithsen: Yeah, this is intentional. If the store doesn't have a working webhook setup, it won't be able to process AP payments, so we should not default it on.

includes/class-wc-stripe-helper.php
malithsen: I think this is okay. A transient blip would still allow the store to display accurate information, unlike the issue this PR addresses.

wjrosa (approved):
Looks good to me 👍

PR #5573, opened 2026-06-15
Fix Adaptive Pricing orders stuck in pending after successful payment

Mayisha:
Fixes STRIPE-1216

With Adaptive Pricing (Checkout Sessions on OCS), a successfully paid order could stay stuck in `pending`.

The order-received redirect handler (`process_checkout_session_redirect`) is UX-only — it just routes the browser and never settles the order. But it was acquiring the order payment lock and holding it across a ~1s Stripe API call. When the `checkout.session.completed` webhook (the path that actually settles the order) arrived in that window, it couldn't get the lock and silently dropped settlement. The redirect handler then released the lock without settling either — so the paid order was never marked `processing`.
  
This is a edge case but realistic race that I encountered during testing another PR.
  
**Fix:** the redirect handler no longer takes the payment lock. It doesn't need it — it never settles the order, and duplicate returns are already idempotent via the terminal-status short-circuit and the `_stripe_upe_redirect_processed` flag. The change is a net removal of the locking machinery from that handler; the webhook now settles unobstructed.
  
  ## Testing instructions
  
  1. Enable Adaptive Pricing and use the optimized checkout.
  2. Place an order with local currency (AP)
  3. Complete payment and let it redirect to the order-received page.
  4. Confirm the order moves to **Processing** (not stuck in Pending), the payment intent/charge are recorded, and the "Local currency purchase via Adaptive Pricing" order note is present.
  5. Also verify the non-success paths still behave: cancelling on a redirect-based method bounces back to checkout with a notice, and a hard decline marks the order Failed.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -2299,107 +2299,99 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 			exit;
 		}
 
-		$locked = $order_helper->lock_order_payment( $order );
-		if ( $locked ) {
-			WC_Stripe_Logger::info( "Skip processing checkout session redirect for order $order_id, order payment is already being processed (locked)" );
-			return;
-		}
-
-		try {
-			$checkout_session = $this->get_checkout_session_from_order( $order, true, true );
-			if ( null === $checkout_session ) {
-				// Either the order has no session id (defensive) or the Stripe API call failed.
-				// Logged downstream by get_checkout_session_from_order(). Leave the order
-				// untouched and let the webhook take over, but still strip the disambiguation
-				// query args from the address bar so a refresh doesn't loop on the API.
-				$order_helper->unlock_order_payment( $order );
-				wp_safe_redirect( wp_sanitize_redirect( $this->get_return_url( $order ) ) );
-				exit;
-			}
-
-			$status         = isset( $checkout_session->status ) ? $checkout_session->status : '';
-			$payment_intent = isset( $checkout_session->payment_intent ) && is_object( $checkout_session->payment_intent )
-				? $checkout_session->payment_intent
-				: null;
-			$payment_error  = $payment_intent && isset( $payment_intent->last_payment_error ) ? $payment_intent->last_payment_error : null;
-
-			// Success: session is complete (paid OR async-pending unpaid like Boleto).
-			// Mark the handler as having run so a refresh of the order-received page
-			// doesn't fall through into the cancel path while the webhook is still in flight,
-			// then strip the disambiguation query args from the address bar.
-			if ( 'complete' === $status ) {
-				$order_helper->update_stripe_upe_redirect_processed( $order, true );
-				$order->save();
-				$order_helper->unlock_order_payment( $order );
-				wp_safe_redirect( wp_sanitize_redirect( $this->get_return_url( $order ) ) );
-				exit;
-			}
-
-			// Anything else (open / expired) on a non-terminal order means the customer
-			// either cancelled on the redirect provider, the session expired, or there
-			// was a hard failure. Disambiguate via last_payment_error if present.
-			//
-			// NOTE: we intentionally do NOT set `_stripe_upe_redirect_processed` on these
-			// branches. The flag is order-scoped and never cleared, so setting it here
-			// would block a subsequent retry on the same order (the next return would
-			// short-circuit above and skip the bounce/notice logic). Replay protection
-			// for cancel/failure is unnecessary because these paths leave the order in
-			// a non-terminal state that is safe to re-enter.
-			$cancellation_codes = [
-				'payment_method_customer_decline',
-				'payment_intent_payment_attempt_expired',
-			];
-			$error_code         = $payment_error && isset( $payment_error->code ) ? $payment_error->code : '';
-			$error_message      = $payment_error && isset( $payment_error->message ) ? $payment_error->message : '';
-
-			// Hard failure: a real decline/error reported by the upstream payment provider.
-			// An `expired` session (Stripe's own timeout) is never a hard failure — the
-			// customer needs a fresh session to retry, regardless of any stale
-			// last_payment_error left on the intent from an earlier attempt.
-			$is_hard_failure = 'expired' !== $status
-				&& ! empty( $error_code )
-				&& ! in_array( $error_code, $cancellation_codes, true );
-
-			if ( $is_hard_failure ) {
-				WC_Stripe_Logger::error(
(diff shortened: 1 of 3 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2299,107 +2299,99 @@ public function process_checkout_session_redirect( int $order_id, bool $is_pay_f
...
-						'error_code'     => $error_code,
-						'error_message'  => $error_message,
-					]
-				);
+		// Success: session is complete (paid OR async-pending unpaid like Boleto).
+		// Mark the handler as having run so a refresh of the order-received page
+		// doesn't fall through into the cancel path while the webhook is still in flight,
+		// then strip the disambiguation query args from the address bar.
+		if ( 'complete' === $status ) {
+			$order_helper->update_stripe_upe_redirect_processed( $order, true );
```
daledupreez, line 2328: We're making updates to the order in both code paths - it feels like we should keep the lock if we're going to be doing that.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2299,107 +2299,99 @@ public function process_checkout_session_redirect( int $order_id, bool $is_pay_f
 			exit;
 		}
 
-		$locked = $order_helper->lock_order_payment( $order );
-		if ( $locked ) {
-			WC_Stripe_Logger::info( "Skip processing checkout session redirect for order $order_id, order payment is already being processed (locked)" );
-			return;
+		// No payment lock needed: this path only routes the browser and never settles the order,
```
wjrosa, line 2302: I think it is worth it asking AI to reduce all the comments it added to this file

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2299,107 +2299,99 @@ public function process_checkout_session_redirect( int $order_id, bool $is_pay_f
...
+		// and duplicate returns are already idempotent via the terminal-status short-circuit and
+		// the `_stripe_upe_redirect_processed` flag above. Taking it also risks harm — if the
+		// checkout.session.* webhook arrives while we hold it across the Stripe call below, the
+		// webhook can't settle the order and a paid order is left stuck pending.
+		$checkout_session = $this->get_checkout_session_from_order( $order, true, true );
```
wjrosa, line 2307: Dropping the lock also drops the dedup it incidentally gave us: a concurrent first-time return (double-submit, link prefetch, refresh before the webhook lands) used to short-circuit on `$locked`, now each one makes this ~1s Stripe call. The terminal-status/flag guards above only help *after* the first request finishes. Fine trade for the fix — just worth a conscious nod.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2299,107 +2299,99 @@ public function process_checkout_session_redirect( int $order_id, bool $is_pay_f
...
-						'error_message'  => $error_message,
-					]
-				);
+		// Success: session is complete (paid OR async-pending unpaid like Boleto).
+		// Mark the handler as having run so a refresh of the order-received page
+		// doesn't fall through into the cancel path while the webhook is still in flight,
+		// then strip the disambiguation query args from the address bar.
+		if ( 'complete' === $status ) {
+			$order_helper->update_stripe_upe_redirect_processed( $order, true );
+			$order->save();
```
wjrosa, line 2329: Re the "updates in both paths" concern: the two mutations aren't symmetric. This success `save()` only persists the `_stripe_upe_redirect_processed` meta (`update_..._processed` → `update_order_meta`); status is never touched here, so it's safe to write lock-free even if the webhook settles concurrently. The only settlement-class mutation is the hard-failure `update_status(FAILED)` at L2370 — so that's the single spot that actually races the webhook.

wjrosa (commented):
Two notes that I don't think the other threads cover (staying off the lock-vs-mutation and comment-volume points already raised).

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: They write different, mutually-exclusive data, so it's worth separating them:

  - **`complete` branch** writes only the `_stripe_upe_redirect_processed` meta flag. The webhook writes status + intent/charge meta. Different fields, and WC's CRUD `save()` only persists *changed* props on
  this object — it won't rewrite status or clobber the webhook's meta. This write is safe lock-free.
  - **hard-failure branch** writes order *status* (`FAILED`) — the same field the webhook writes (`processing`). These are mutually exclusive per session, but **not across a retry on the same order**:
  attempt 1 declines, the customer retries, attempt 2's webhook settles to `processing`, and attempt 1's still-in-flight return then stamps `FAILED` over a paid order. Narrow window, but bad outcome.

  So there is one genuinely racy write — the `FAILED` status — but keeping the lock doesn't actually fix it cleanly: under that same retry, the lock just starves the webhook again, so you'd get "stuck
  pending" instead of "wrongly failed." The lock's protection here is incidental, and its cost is the very bug this PR fixes.

**Options**
1. **Stay lock-free, guard only the hard-failure branch** — reload the order and bail if it's already terminal or no longer points at the returned session before calling `update_status(FAILED)`. Protects the one genuinely shared field, and keeps the webhook as the single settler.

2. **Keep the lock, and always settle the webhook asynchronously** — instead of settling inline on receipt, `checkout.session.completed` always schedules an Action Scheduler job and returns immediately.
  The redirect handler's lock is request-scoped and released when that request ends, so by the time the deferred job runs there's no contention — it takes the lock and settles. This keeps the lock on both
  paths (so the mutation-safety concern stays addressed) and is the smallest webhook change, but it adds settlement latency to *every* order (the deferred-job delay, currently ~2 min) and makes settlement
  depend on Action Scheduler firing on time, which can lag on low-traffic stores.

3. **Keep the lock here, but stop the webhook from dropping the event when it's locked** — today `handle_checkout_session_success()` just `return`s when it can't get the order lock, silently discarding the
settlement. Instead, we'd re-queue the webhook (via Action Scheduler) to retry shortly; the redirect handler releases the lock within ~1s, so the retry acquires it and settles. Keeps the lock but adds retry plumbing on the webhook side.

Mayisha:
Considering different scenarios as well as based on the raised concerns by @daledupreez and @wjrosa, removing the lock completely seems risky. I have created another PR with a different approach https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5576 to handle this issue.

Mayisha:
Closing in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5576

PR #5574, opened 2026-06-15
Run woocommerce_gateway_description filter on UPE payment methods so descriptions can be filtered

### LINKED ISSUE #5572: Remove WC_Stripe_UPE_Payment_Method get_description method to allow filtering by 3rd parties
The [`get_description`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/10.8.1/includes/payment-methods/class-wc-stripe-upe-payment-method.php#L346) method on `WC_Stripe_UPE_Payment_Method` explicitly replaces the WooCommerce `WC_Payment_Gateway` [`get_description`](https://github.com/woocommerce/woocommerce/blob/10.8.1/plugins/woocommerce/includes/abstracts/abstract-wc-payment-gateway.php#L393) method and returns "empty".

This breaks the expected WooCommerce behavior and makes it impossible for 3rd parties to use the `woocommerce_gateway_description` filter to add or replace the payment method description for Stripe payment methods.

Use case example: for Direct Debit, I would like to add a small description to inform the user that the "payment approval" can take some days, and is not immediate, like in Credit Card.

This is just an example. The primary issue here is that the expected WooCommerce behavior is broken deliberately, which is, well, not cool 🤓

Mayisha:
Fixes STRIPE-1213
Fixes #5572 

`WC_Stripe_UPE_Payment_Method::get_description()` hard-returned `''`, which short-circuited WooCommerce's `WC_Payment_Gateway::get_description()` and never fired the `woocommerce_gateway_description` filter. As a result, third parties could not add or replace the checkout description for individual Stripe payment methods (e.g. a "SEPA approval can take a few days" note on Direct Debit).

This change keeps the empty default but routes it through the standard filter, so extensions can hook in.

## Testing instructions
Code review should be sufficient.

1. Without any filter: no method description renders at checkout (unchanged).
2. Add a filter, e.g.:
```php
     add_filter( 'woocommerce_gateway_description', function ( $desc, $id ) {
         return 'stripe_sepa_debit' === $id ? 'Payment approval can take a few days.' : $desc;
     }, 10, 2 );
```
Confirm the text appears under SEPA Direct Debit on classic checkout, Blocks, and OC, and that other methods are unaffected.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -16,6 +16,7 @@ xxxx-xx-xx - version 10.9.0
 * Fix - Append the 3-letter currency code to the Stripe Fee and Stripe Payout amounts in WP Admin order totals when the Stripe account currency differs from the order currency
 * Fix - Only create an agentic commerce order on the site that produced the checkout, preventing duplicate or wrong-site orders when multiple stores share one Stripe account
 * Dev - Add a Code Comment Conventions section to AGENTS.md to standardize agent and contributor comment guidance
+* Fix - Run the woocommerce_gateway_description filter for Stripe payment methods so third parties can add or replace their checkout descriptions
 
 2026-06-11 - version 10.8.1
 * Fix - Prevent a fatal error when a Bancontact, iDEAL or Sofort payment is placed with SEPA token saving enabled through the Adaptive Pricing checkout, which left the order unpaid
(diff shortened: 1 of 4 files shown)


wjrosa (approved):
Change makes sense and works as expected 👍

PR #5576, opened 2026-06-15
Fix Adaptive Pricing orders stuck in pending by re-queueing the settlement webhook when the order is locked

Mayisha:
Fixes STRIPE-1216

With Adaptive Pricing (Checkout Sessions on OCS), a successfully paid order could stay stuck in pending.

The order-received redirect handler (process_checkout_session_redirect) is UX-only — it just routes the browser and never settles the order. But it was acquiring the order payment lock and holding it across a ~1s Stripe API call. When the checkout.session.completed webhook (the path that actually settles the order) arrived in that window, it couldn't get the lock and silently dropped settlement. The redirect handler then released the lock without settling either — so the paid order was never marked processing.

This is a edge case but realistic race that I encountered during testing another PR.

 **Fix:** when the order is locked, the webhook now re-queues itself (via Action Scheduler) instead of dropping the event. The lock holder releases within ~1s, so the retry acquires the lock and settles the order — nothing is lost. The lock's existing 5-minute TTL guarantees a wedged holder clears, so the retry always terminates (no retry counter needed). This keeps the redirect handler's lock in place; the only change is making the webhook resilient to contention rather than giving up.

## Testing instructions

- Enable Adaptive Pricing and use the optimized checkout.
- Place an order with local currency (AP)
- Complete payment and let it redirect to the order-received page.
- Confirm the order moves to **Processing** (not stuck in Pending), the payment intent/charge are recorded, and the "Local currency purchase via Adaptive Pricing" order note is present.
- Also verify the non-success paths still behave: cancelling on a redirect-based method bounces back to checkout with a notice, and a hard decline marks the order Failed.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -61,6 +61,17 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	protected $deferred_webhook_delay = 2 * MINUTE_IN_SECONDS;
 
 	/**
+	 * How long to wait before retrying a webhook that lost the order-payment lock race.
+	 *
+	 * The order-received redirect handler holds the lock only across a single Stripe API
+	 * call (~1s), so a short backoff settles the order quickly instead of leaving it pending
+	 * for the full deferred delay. Kept well above the typical hold to avoid a busy re-queue loop.
+	 *
+	 * @var int
+	 */
+	protected $locked_order_retry_delay = 10;
+
+	/**
 	 * The Action Scheduler hook to use when retrying a webhook.
 	 *
 	 * @var string
@@ -1465,10 +1476,12 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	 *
 	 * @param stdClass $webhook_notification The webhook payload received from Stripe.
 	 * @param array    $additional_data      Additional data to pass to the scheduled job.
+	 * @param int|null $delay                Seconds to wait before retrying. Defaults to $deferred_webhook_delay.
 	 */
-	protected function defer_webhook_processing( $webhook_notification, $additional_data ) {
+	protected function defer_webhook_processing( $webhook_notification, $additional_data, $delay = null ) {
+		$delay = null === $delay ? $this->deferred_webhook_delay : $delay;
 		$this->action_scheduler_service->schedule_job(
-			time() + $this->deferred_webhook_delay,
+			time() + $delay,
 			$this->deferred_webhook_action,
 			[
 				'type'         => $webhook_notification->type,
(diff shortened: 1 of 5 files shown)


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1861,8 +1861,12 @@ protected function handle_checkout_session_success( object $notification ): void
 
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
 
-		// Lock the order
+		// Lock the order. The order-received redirect handler briefly holds this same lock across a
+		// Stripe API call without settling; dropping the event here would leave a paid order stuck
+		// pending. Re-queue instead so settlement runs once the lock is released — the lock's 5-minute
+		// TTL guarantees a wedged holder clears, so the retry terminates.
 		if ( $order_helper->lock_order_payment( $order ) ) {
+			$this->defer_webhook_processing( $notification, [ 'session_id' => $session_id ] );
```
wjrosa: `defer_webhook_processing` schedules at the shared `deferred_webhook_delay` = **2 min**. So in exactly the race this fixes, the paid order sits in `pending` on the order-received page for ~2 minutes before the retry settles it — even though the redirect handler releases the lock in ~1s. Worth weighing against settling inline (no delay); if you keep the re-queue, a much shorter dedicated backoff (a few seconds) would close the window the lock TTL doesn't need.

wjrosa (commented):
Approach works, but two behavioral notes plus a baseline nit.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1861,8 +1861,12 @@ protected function handle_checkout_session_success( object $notification ): void
 
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
 
-		// Lock the order
+		// Lock the order. The order-received redirect handler briefly holds this same lock across a
+		// Stripe API call without settling; dropping the event here would leave a paid order stuck
+		// pending. Re-queue instead so settlement runs once the lock is released — the lock's 5-minute
+		// TTL guarantees a wedged holder clears, so the retry terminates.
 		if ( $order_helper->lock_order_payment( $order ) ) {
```
wjrosa, line 1909: When this defers, `process_checkout_session_success` still `return false`s (order exists), so `process_webhook` doesn't see it as deferred and fires `wc_stripe_webhook_received` on this live pass — before settlement actually happens — then the retry fires it again. The order-not-found defer path returns `true` precisely to skip that. Consider propagating a "deferred" signal up here too so the action fires once, after settlement.

malithsen (approved):
I'm not able to replicate the original issue, but the code makes sense, and I didn't notice any regressions during testing.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Fixed in 1917fa663.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Fixed in 1917fa663.

wjrosa (approved):

PR #5577, opened 2026-06-15
Update status of Adaptive Payment failed orders promptly

malithsen:
Forum thread: https://wordpress.org/support/topic/order-status-not-changing-to-failed-after-payment-decline/

When Adaptive Pricing is enabled, checkout runs through Stripe Checkout Sessions, where the payment is confirmed on Stripe's side and the order status is updated by webhooks. If a customer's card was declined, the order couldn't be matched back from the failure webhook, so it was left stuck in "Pending Payment" until the session expired - up to 24 hours later.

This change lets the failed-payment webhook find the order through its checkout session and mark it failed right away. 

## Testing instructions

1. Enable Optimized Checkout and Adaptive Pricing, and make sure webhooks are working.
2. Add a product to the cart and go to checkout.
3. Pay with a card that is declined (Stripe test card 4000 0000 0000 0002).
4. Confirm the payment is declined at checkout.
5. Check the order in the WooCommerce admin: 
- In develop, order status will be Pending Payments
- In this branch, it should be "Failed" 
6. Regression check: complete a normal successful Adaptive Pricing payment and confirm the order is marked paid as usual.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1730,6 +1730,18 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 		// Look for an order. If one does not exists, this is probably an agentic hook.
 		$order = WC_Stripe_Helper::get_order_by_checkout_session_id( $checkout_session->id );
 		if ( ! $order instanceof \WC_Order ) {
+			// An Adaptive Pricing session is tagged at creation. It is never agentic,
+			// so don't route it into the agentic handler — that would silently drop a
+			// paid session whenever agentic commerce is disabled.
+			$checkout_type = $checkout_session->metadata->checkout_type ?? '';
+			if ( WC_Stripe_Checkout_Sessions_Ajax_Handler::ADAPTIVE_PRICING_CHECKOUT_TYPE === $checkout_type ) {
+				WC_Stripe_Logger::warning(
+					'Completed Adaptive Pricing checkout session has no matching order: ' . $checkout_session->id
+				);
+				WC_Stripe_Database_Cache::delete( $lock_key );
+				return;
+			}
+
 			try {
 				$this->handle_agentic_checkout_session( $notification );
 			} finally {
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -2314,6 +2314,7 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 			$session    = $sessions->data[0] ?? null;
 			$session_id = isset( $session->id ) ? (string) $session->id : '';
 			if ( '' === $session_id ) {
+				WC_Stripe_Logger::warning( 'No checkout session found for intent ' . $intent_id . '; order left unresolved.' );
 				return null;
 			}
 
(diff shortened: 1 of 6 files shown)


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1255,6 +1255,15 @@ public function process_payment_intent( $notification ) {
 		$intent = $notification->data->object;
 		$order  = $this->get_order_from_intent( $intent );
 
+		$checkout_type = $intent->metadata->checkout_type ?? '';
+
+		// For AP, attempt to find the order via the checkout session.
+		if ( ! $order
+			&& 'payment_intent.payment_failed' === $notification->type
+			&& WC_Stripe_Checkout_Sessions_Ajax_Handler::ADAPTIVE_PRICING_CHECKOUT_TYPE === $checkout_type ) {
+			$order = $this->get_order_by_intent_checkout_session( isset( $intent->id ) ? (string) $intent->id : '' );
```
wjrosa, line 1264: This fallback hinges on the PI carrying `checkout_type`, which `build_payment_intent_data()` sets via `apply_filters( 'wc_stripe_payment_metadata', … )`. A third-party filter that *replaces* the metadata array (rather than merging) silently disables this fix, and `payment_intent.payment_failed` carries no other AP signal. Worth at least a note about the dependency.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2289,6 +2298,33 @@ private function get_order_from_intent( $intent ) {
...
+	 * @return WC_Order|null
+	 */
+	private function get_order_by_intent_checkout_session( string $intent_id ): ?WC_Order {
+		if ( '' === $intent_id ) {
+			return null;
+		}
+
+		try {
+			$sessions   = WC_Stripe_API::request( [], 'checkout/sessions?payment_intent=' . $intent_id . '&limit=1', 'GET' );
+			$session    = $sessions->data[0] ?? null;
```
wjrosa, line 2314: `WC_Stripe_API::request()` only throws on outage/empty-body; a Stripe API *error* (e.g. bad request) is returned as `{ error: … }` with no `data`, so it won't hit the `catch` — `$sessions->data[0] ?? null` quietly yields null and the order stays Pending with no trace. Consider checking `$sessions->error` (or logging when `data` is absent) so a failed lookup is diagnosable.

wjrosa (commented):
Marker is correctly wired onto both the session and the PI, no order exists at session-creation time so the lookup round-trip is justified, and the status guard in `process_payment_intent` keeps a late failed event from flipping an already-paid order. Two small robustness notes.

includes/class-wc-stripe-webhook-handler.php
malithsen: I don't think a guard is needed. If someone overrides the intent metadata, would break order resolution in general and not just this path.

includes/class-wc-stripe-webhook-handler.php
malithsen: Good point. Added a log in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5577/commits/56a344dac1a04d4a285c51207b229cec925b7d62

wjrosa (approved):
Looks good! Thanks for addressing the comments

PR #5587, opened 2026-06-17
Deprecate WC_Stripe_Feature_Flags::is_amazon_pay_available()

### LINKED ISSUE #4989: [Amazon Pay] Deprecate `WC_Stripe_Feature_Flags::is_amazon_pay_available()`
Once Amazon Pay has been launched, we need to deprecate `WC_Stripe_Feature_Flags::is_amazon_pay_available()`. We couldn't do that as part of the main release, as we still have internal code calling the function and many of our PHP unit tests were failing as a result. To get this to completion, we need to do the following:
 * Remove internal calls to `WC_Stripe_Feature_Flags::is_amazon_pay_available()`
 * Add `wc_deprecated_function()` to the implementation

This stems from the work to enable Amazon Pay in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/4859, where I added the `wc_deprecated_function()` call, but realised that it would cause unit test failures so backed out of that change.

wjrosa:
Fixes STRIPE-943
Fixes #4989

Amazon Pay has been permanently enabled since 10.4.0, so `WC_Stripe_Feature_Flags::is_amazon_pay_available()` always returns `true`. This PR completes the deprecation that was prepared in #4859 but backed out because the live `wc_deprecated_function()` call tripped unit-test deprecation assertions while internal callers still existed.

This removes all 5 internal callers first, then adds the deprecation:

- **`is_amazon_pay_available()`** — added `wc_deprecated_function( __METHOD__, '10.9.0' )`.
- **`class-wc-stripe.php`** — dropped the flag check from the `amazon_pay` admin-area branch and removed the dead `maybe_deactivate_amazon_pay()` method and its call site (it only ever returned `[]` once Amazon Pay shipped; its own TODOs said to delete it post-release).
- **`is_amazon_pay_enabled()`** (Amazon Pay method) — removed the always-false feature-flag guard.
- **Localized params** — removed `is_amazon_pay_available` (settings controller) and `isAmazonPayAvailable` (UPE gateway), plus the now-constant conditionals they gated in `client/settings/payment-request-section/index.js` and `client/blocks/upe/index.js`. Amazon Pay availability now derives from the enabled/available payment method IDs.

Because the flag was always `true`, this is behaviorally a no-op — Amazon Pay stays permanently enabled. The only externally observable change is that any lingering third-party caller of `is_amazon_pay_available()` now gets a deprecation notice.

## Testing instructions

This is an internal refactor with no intended user-facing change; the goal of testing is to confirm Amazon Pay behaves exactly as before.

1. Connect a Stripe account in a supported Amazon Pay country (e.g. US) and enable Amazon Pay.
2. **Settings:** go to the Stripe settings → Payment Methods / Express Checkout section and confirm the Amazon Pay row renders and its checkbox toggles as before.
3. **Blocks checkout:** add a product to the cart and confirm the Amazon Pay express button still renders on the Blocks cart/checkout.
4. **Classic / shortcode checkout:** confirm the Amazon Pay express button still renders.
5. Disable Amazon Pay and confirm it disappears from all surfaces.
6. Confirm no PHP `is_amazon_pay_available` deprecation notices appear during normal checkout (no internal code calls it anymore).

---

### Changelog entry

Added under 10.9.0 in both `changelog.txt` and `readme.txt`:

> Dev - Deprecate the internal WC_Stripe_Feature_Flags::is_amazon_pay_available() helper now that Amazon Pay is permanently enabled

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -21,6 +21,7 @@ xxxx-xx-xx - version 10.9.0
 * Dev - Add a Code Comment Conventions section to AGENTS.md to standardize agent and contributor comment guidance
 * Fix - Run the woocommerce_gateway_description filter for Stripe payment methods so third parties can add or replace their checkout descriptions
 * Fix - Show an error notice when a Stripe OAuth connection attempt fails its security check
+* Dev - Deprecate the internal WC_Stripe_Feature_Flags::is_amazon_pay_available() helper now that Amazon Pay is permanently enabled
 
 2026-06-15 - version 10.8.2
 * Fix - Disable Adaptive Pricing when webhooks are disabled
--- a/client/blocks/upe/index.js
+++ b/client/blocks/upe/index.js
@@ -55,10 +55,7 @@ Object.entries( paymentMethodsConfig )
 	} );
 
 // Register Express Checkout Elements.
-if (
-	getBlocksConfiguration()?.isAmazonPayAvailable && // Hide behind feature flag so the editor does not show the button.
-	getBlocksConfiguration()?.isAmazonPayEnabled
-) {
+if ( getBlocksConfiguration()?.isAmazonPayEnabled ) {
 	registerExpressPaymentMethod( expressCheckoutElementAmazonPay( api ) );
 }
 if ( getBlocksConfiguration()?.isExpressCheckoutEnabled ) {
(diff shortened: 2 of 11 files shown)


malithsen (approved):
Looks good and works as described.

PR #5588, opened 2026-06-17
Add "What's new" inbox note for plugin upgraders

wjrosa:
Fixes STRIPE-1188

The two post-update "what's new" surfaces added in #5386 (the plugin-row "Release notes" link and the "See what's new" link on the *Updated!* notice) are both gated on `'plugins.php' === $hook_suffix`. Stores on auto-update never return to `plugins.php`, so neither surface ever fires for them.

This PR adds a WC Inbox note that reaches upgraders on their next `wp-admin` visit anywhere, regardless of how the plugin was updated.

- **`WC_Stripe_Whats_New_Note`** (`includes/notes/class-wc-stripe-whats-new-note.php`) — follows the `WC_Stripe_OC_Promotion_Note` pattern (`NoteTraits` + `get_note()` + `init()`), registered in `WC_Stripe_Inbox_Notes::create_upe_notes()` and Composer-classmap autoloaded. The title carries the full version (e.g. "WooCommerce Stripe X.Y.Z: see what's new"). Note type is `info`: the Inbox only renders `info`/`marketing`/`survey`/`warning`, and `info` (unlike `marketing`) isn't suppressed by the marketplace-suggestions opt-out. The **See what's new** action points to the public WordPress.org changelog (`…/woocommerce-gateway-stripe/#developers`) and **opens in a new tab** — the Inbox only `window.open()`s action URLs that fall outside the admin URL, so an external target is required. No image is set (`set_image()` is deprecated as of WC Admin 10.8.0).
- **Version gate** — a new `wc_stripe_whats_new_note_version` option records the version a merchant was last shown the note for. `WC_Stripe::install()` calls `record_install_version()`: a fresh install banks the *current* version (so the note never fires); an upgrader banks the version it came *from*, but only when no baseline exists yet, so an existing baseline is never rolled back. `init()` fires only when the baseline is older than `WC_STRIPE_VERSION`, then advances it — so the note shows **once per upgrade** and stays dismissible.

`install()` and `create_upe_notes()` both run on `admin_init` in an undefined order; the gate is written to be order-independent (worst case is a one-page-load delay, never a note shown to the wrong audience).

Preview:
Screenshot 2026-06-17 at 11 05 54

### Out of scope: Remote Inbox Notification (RIN)

The RIN feed-filter approach raised in the Linear discussion is intentionally **not** included. Registering a `data_source_poller_data_sources` feed URL from the plugin would point WC's daily poller at a `notes.json` that this repo doesn't host (a recurring 404), and a Stripe release-notes RIN can already be delivered through woocommerce.com's existing core feed using `plugin_version`/`plugins_activated` rules with zero plugin code. RIN remains an optional, complementary campaign owned by the Remote Inbox team.

#### How can this code break?
- If the baseline option were written in a way that rolled back below the running version, the note could re-show every page load. The `record_install_version()` no-rollback guard and the dedicated init-only baseline advance prevent this.
- A fresh install could wrongly see the note if the baseline were absent. `init()` stays silent when the baseline is unset.

#### What keeps it from breaking?
Targeted PHPUnit coverage for the version-gate logic (below).

## Testing instructions

1. **Upgrader path:** On a store already running an older version, set the baseline to an older value and visit any `wp-admin` page:
   - `wp option update wc_stripe_whats_new_note_version 10.8.0`
   - Load `wp-admin` → confirm a note titled "WooCommerce Stripe X.Y.Z: see what's new" appears under **WooCommerce → Home → Inbox**.
   - Click **See what's new** → confirm it opens the WordPress.org changelog in a **new tab**.
2. **Once per upgrade:** Reload several `wp-admin` pages → confirm only one note exists and `wc_stripe_whats_new_note_version` now equals the current plugin version. Dismiss the note → confirm it does not reappear.
3. **Fresh install:** `wp option update wc_stripe_whats_new_note_version <current plugin version>` (simulates a fresh install's baseline) → load `wp-admin` → confirm **no** note appears.
4. Automated: `npm run test:php -- --filter WC_Stripe_Whats_New_Note_Test` (13 tests) and `--filter WC_Stripe_Inbox_Notes_Test`.

### Alternative ways to test (faster, no real upgrade)

`WC_Stripe_Whats_New_Note` is autoloaded, so these work from `wp shell`/`wp eval` without any `require`.

**A) Force-render the actual note immediately (bypasses the version gate):**

```bash
wp eval 'WC_Stripe_Whats_New_Note::get_note()->save();'
```

Then open **WooCommerce → Home → Inbox**. Uses the real `get_note()`, so it's the quickest way to eyeball the title/copy/CTA. Re-running duplicates it — reset first (see C).

**B) Exercise the real gate end-to-end (`install()` → `init()`):**

```bash
wp option update wc_stripe_whats_new_note_version 0.0.1
wp eval 'WC_Stripe_Inbox_Notes::create_upe_notes();'   # what admin_init runs
wp option get wc_stripe_whats_new_note_version          # should now equal WC_STRIPE_VERSION
```

The option advancing back to the current version proves the note fired exactly once.

**C) Reset between runs:**

```bash
wp eval 'Automattic\WooCommerce\Admin\Notes\Notes::delete_notes_with_name( WC_Stripe_Whats_New_Note::NOTE_NAME );'
```

**D) Confirm the note reaches the browser (DevTools console on any wc-admin page):**

```js
wp.apiFetch( { path: '/wc-analytics/admin/notes?status=unactioned&per_page=5&type=info,marketing,survey,warning' } )
  .then( n => console.log( n.map( x => x.name ) ) )
// expect 'wc-stripe-whats-new-note' in the list
```

> Note: the Inbox only renders notes of type `info`/`marketing`/`survey`/`warning` and only opens action URLs **outside** the admin URL in a new tab — both are exercised by the steps above.

---

### Changelog entry

> Add - Show a "What's new" inbox note after the plugin updates so merchants who auto-update are still pointed to the release notes

**Post merge**

cc @pierorocca

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,6 +22,7 @@ xxxx-xx-xx - version 10.9.0
 * Fix - Run the woocommerce_gateway_description filter for Stripe payment methods so third parties can add or replace their checkout descriptions
 * Fix - Show an error notice when a Stripe OAuth connection attempt fails its security check
 * Dev - Deprecate the internal WC_Stripe_Feature_Flags::is_amazon_pay_available() helper now that Amazon Pay is permanently enabled
+* Add - Show a "What's new" inbox note after the plugin updates so merchants who auto-update are still pointed to the release notes
 
 2026-06-15 - version 10.8.2
 * Fix - Disable Adaptive Pricing when webhooks are disabled
--- a/changelog.txt
+++ b/changelog.txt
@@ -22,7 +22,7 @@ xxxx-xx-xx - version 10.9.0
 * Fix - Run the woocommerce_gateway_description filter for Stripe payment methods so third parties can add or replace their checkout descriptions
 * Fix - Show an error notice when a Stripe OAuth connection attempt fails its security check
 * Dev - Deprecate the internal WC_Stripe_Feature_Flags::is_amazon_pay_available() helper now that Amazon Pay is permanently enabled
-* Add - Show a "What's new" inbox note after the plugin updates so merchants who auto-update are still pointed to the release notes
+* Add - Show a "What's new" inbox note after the plugin updates, including via auto-update
 
 2026-06-15 - version 10.8.2
 * Fix - Disable Adaptive Pricing when webhooks are disabled
--- a/composer.json
+++ b/composer.json
@@ -55,6 +55,7 @@
         "includes/migrations/class-wc-stripe-express-checkout-add-change-payment-method-location-update.php",
         "includes/migrations/class-wc-stripe-migrate-link-button-locations.php",
         "includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php",
+        "includes/notes/class-wc-stripe-whats-new-note.php",
         "includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php",
         "includes/payment-tokens/class-wc-stripe-ach-payment-token.php",
         "includes/payment-tokens/class-wc-stripe-acss-payment-token.php",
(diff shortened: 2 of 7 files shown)


Mayisha (approved):
Nice work 💯 

Image

PR #5590, opened 2026-06-17
Fix STRIPE-882: validate customer before saved cards

kamlesh08:
Fixes STRIPE-882

This PR prevents stale or missing Stripe customer IDs from leaving unusable saved cards visible at checkout.

- Hide the saved payment methods list when the WooCommerce customer does not have a stored Stripe customer ID.
- Validate the stored Stripe customer before showing saved cards, and hide the list when Stripe reports the customer no longer exists.
- Keep saved cards visible when Stripe returns a generic API error or throws before returning a response, so transient Stripe failures do not remove existing local payment methods from checkout.
- Add regression coverage for the saved-card list filtering behavior.

## Testing instructions

1. Run `php -l includes/payment-methods/class-wc-stripe-upe-payment-gateway.php`.
2. Run `php -l tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php`.
3. Run `npm run test:php -- --filter=test_filter_saved_payment_methods_list`.
4. As a logged-in customer with saved cards, clear the local `_stripe_customer_id` user option and confirm the saved payment methods list is hidden at checkout.
5. Set `_stripe_customer_id` to a deleted/missing Stripe customer and confirm the saved payment methods list is hidden at checkout.
6. Use a valid Stripe customer with saved cards and confirm saved payment methods still display at checkout.
7. Simulate a generic Stripe API failure while validating the customer and confirm existing saved payment methods remain visible.

---

-   [x] Covered with tests (or have a good reason not to test in description)
-   [x] Tested on mobile (or does not apply): Does not apply; this is server-side saved payment method list filtering.

### Changelog entry

-   [x] Added changelog entries to `changelog.txt` and `readme.txt`.

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,6 +7,7 @@ xxxx-xx-xx - version 10.9.0
 * Dev - Use a shared hook manager to prevent duplicate subscription hook registrations
 * Dev - Initial infrastructure for more complex Agentic feed filtering
 * Fix - Prevent unnecessary Stripe payment method creation when shortcode checkout has empty required fields
+* Fix - Hide saved cards at checkout when the stored Stripe customer ID is missing or no longer exists
 * Fix - Decommission the previously configured webhook before connecting via OAuth so reconnecting to a different Stripe account no longer leaves an orphaned webhook on the old account
 * Fix - Send the billing address to Stripe on the Pay for Order page so payments aren't incorrectly blocked by Stripe Radar rules
 * Update - Replace shipping AJAX endpoints with Store API calls for Express Checkout Element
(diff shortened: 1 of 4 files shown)


kamlesh08:
Closing this after a closer look: the scenario it targets is already handled on `develop` by the existing saved-token reconciliation.

On every logged-in checkout/account view, `WC_Stripe_Payment_Tokens::sync_and_retrieve_customer_payment_tokens()` calls `WC_Stripe_Customer::get_all_payment_methods()` — which already detects the `resource_missing` customer error (the same check this PR added) and caches it — and then `cleanup_invalid_tokens()` deletes local tokens that no longer exist in Stripe. So a saved card backed by a missing or deleted Stripe customer is already removed for customers with fewer than `posts_per_page` (10) saved tokens.

The behaviour only persists at 10+ saved tokens, where the reconciliation is skipped (the token data store returns only the first `posts_per_page`). That's a narrow edge case, and it would be better addressed in the token-retrieval/reconciliation path than by adding a second, uncached `payment_methods` call on the checkout render. Closing in favour of that finding.

PR #5591, opened 2026-06-17
Fix STRIPE-1187: honor renewal payment locks

kamlesh08:
Fixes STRIPE-1187

The renewal code calls `lock_order_payment()` but does not use its result. Two scheduled renewals for one order can both send a charge to Stripe.

- `process_subscription_payment()` takes the lock once and releases it in `finally`. The lock covers the Stripe request, the response handling, and all retries.
- If the lock is already held, the method returns without a charge. It writes an error to the log and adds a note to the order. The order keeps its status, so Woo Subscriptions does not schedule a retry; the worker that holds the lock is expected to finish the payment. A lock left behind by a worker that died blocks renewals for that order for up to five minutes.
- Retries stop when the lock expires. The method only removes the lock if the stored value is the value it wrote. Two workers that start in the same second write the same value; #5897 adds owner tokens to close that gap.
- The SEPA path now takes the same lock. Before, it took no lock. Open question for the reviewer: that path runs only when the gateway id is `stripe_sepa`, which is the removed legacy checkout id. If no gateway or subclass uses it, the SEPA branch and its tests can go.
- Two new private methods hold the attempt and retry code. Public method signatures do not change. Retries no longer re-enter `process_subscription_payment()`, because that would take the lock again. A subclass that overrides that method now runs its override once per renewal, not once per attempt.
- This PR removes the `phpstan-baseline.neon` entries that the change makes unnecessary.

Follow-up PRs build on this one. Merge order: #5897 (atomic lock), #5901 (renewal path uses it), #5900 (owner row cleanup), then #5898 (expanded charge) and #5899 (retry keys). #5898, #5899, and #5901 edit the same method, so each needs a rebase after the one before it merges.

## Testing instructions

**Environment:** Docker (`npm run up`), PHP dependencies installed.

**Action:** Run `npm run test:php -- --filter=WC_Stripe_Subscription_Renewal_Test`, `npm run phpstan`, and `npm run lint:php`.

**Validation:** All 18 tests pass. The tests cover these cases:

- A held lock stops the charge, for cards and for SEPA.
- The SEPA request runs while the lock is held.
- A retry takes the lock only once.
- The lock stays through `process_response()` and the code releases it after.
- An expired lock stops retries.
- The method does not remove a lock that another process wrote.

---

### Changelog entry

<details>

#### Comment

`Fix - Honor subscription renewal payment locks while preserving retry attempts`

</details>


includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -742,8 +728,61 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+			$response = $this->create_and_confirm_intent_for_off_session( $renewal_order, $prepared_source, $amount );
+			/** @var stdClass $response */
+
+			return [
+				'response'                   => $response,
+				'is_authentication_required' => $this->is_authentication_required_for_payment( $response ),
+			];
+		} finally {
 			$order_helper->unlock_order_payment( $renewal_order );
 		}
```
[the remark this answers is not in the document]
malithsen: From inspection, I agree with this comment. The lock window is now narrower than the previous one. Is there a reason for that? 

In general, it feels safer to have it locked for the duration of processing, like before.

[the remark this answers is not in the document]
kamlesh08: Hi @malithsen, I am still working on this, and I will update you here once this PR is ready for review.
I'd to take it out of draft so that Copilot can review.
Thanks for the feedback. I really appreciate it.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -741,11 +755,57 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+	 * The caller is responsible for the order payment lock; this method only performs a single
+	 * charge so the lock can be held across the response processing that follows.
+	 *
+	 * @param float    $amount         The amount to charge.
+	 * @param WC_Order $renewal_order  The renewal order.
+	 * @param object   $prepared_source The prepared renewal order source.
+	 * @return array{response: stdClass, is_authentication_required: bool}
+	 * @throws WC_Stripe_Exception When the Stripe API request fails.
+	 */
+	private function create_subscription_renewal_payment_response( $amount, WC_Order $renewal_order, $prepared_source ): array {
```
Mayisha:
```suggestion
	private function attempt_subscription_renewal_payment( $amount, WC_Order $renewal_order, $prepared_source ): array {
```
`attempt_subscription_renewal_payment` would be better suited as the name.

Mayisha (changes requested):
Thanks for the fix. This is mostly looking good to me. I have a suggestion about the method renaming and the phpstan issues are valid around $order_locked variable.

includes/compat/trait-wc-stripe-subscriptions.php
kamlesh08: Renamed to `attempt_subscription_renewal_payment()` in 426db718d. The same push also resolves the PHPStan issues: the `$order_locked` flag is gone entirely — `process_subscription_payment()` now acquires the lock once and releases it in a `finally`, with the attempt/retry logic moved to a private `process_subscription_payment_attempt()`. PHPStan and PHPCS both pass on the branch now.

includes/compat/trait-wc-stripe-subscriptions.php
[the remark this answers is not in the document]
kamlesh08: As of 426db718d the lock is held for the full duration again, and wider than before this PR: `process_subscription_payment()` acquires it before the first attempt and releases it in a `finally` after response processing — including in-process retries, which previously released and re-acquired it between attempts. `test_renewal_retry_acquires_the_lock_only_once` asserts the lock is taken exactly once across a retry chain.

kamlesh08:
@Mayisha @malithsen, I have updated the PR with all the suggestions.
Please do let me know if any other changes are needed.
Thank you.

daledupreez:
Thanks for your patience, @kamlesh08! I'll take a look at this PR tomorrow.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -501,7 +501,40 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+		$order_helper = WC_Stripe_Order_Helper::get_instance();
+		$order_id     = $renewal_order->get_id();
+
+		// Acquire the order payment lock once and hold it for the entire renewal attempt —
+		// including any retries — so a concurrent scheduled renewal cannot create a second charge
+		// for this order while this attempt is in flight. Returning early when the lock is already
+		// held mirrors the regular checkout handler. The finally guarantees release on every exit
+		// path (success, failure, retry exhaustion, or an unexpected error).
+		if ( $order_helper->lock_order_payment( $renewal_order ) ) {
+			WC_Stripe_Logger::warning( "Stripe: skipping duplicate renewal attempt for order {$order_id} because the payment lock is already held." );
```
daledupreez: @kamlesh08, can we bump this to an error so we are guaranteed to log something when this happens? Warnings are only written when debug logging is enabled, and we definitely want to guarantee this situation is logged.
```suggestion
			WC_Stripe_Logger::error( "Stripe: skipping duplicate renewal attempt for order {$order_id} because the payment lock is already held." );
```

daledupreez: And on a related note, can we (re)schedule a delayed retry? (I haven't looked up the stack, but it feels like that would be a better approach than existing immediately just because there was a lock on the order.)

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -501,7 +501,40 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+			return;
+		}
+
+		try {
+			$this->process_subscription_payment_attempt( $amount, $renewal_order, $retry, $previous_error );
+		} finally {
+			$order_helper->unlock_order_payment( $renewal_order );
+		}
```
daledupreez, line 529: There is a subtle failure mode that could be problematic with the new code:
When we lock the order payment, that lock only lasts 5 minutes, but the retries below could technically take longer than that given that we retry up to 6 times and the Stripe API call timeout is 70 seconds by default. That means we could still be retrying after the lock has timed out and/or our finally cleanup might release a lock taken out by another process.

I think we can work around this by calling `$order_helper->get_order_existing_payment_lock()` to get the timestamp at which the lock will expire, and then doing two things:
 * Adding checks to the retry logic to either exit early when our lock times out _or_ consider allowing a lock renewal/extension operation.
    - I would lean towards exiting, as if it's taken us 5 minutes, we are unlikely to start succeeding, and the lock extension code would be very tricky to use safely.
 * Adding a check to the finally clause to make sure the lock timeout is the same before we call unlock().

includes/compat/trait-wc-stripe-subscriptions.php
daledupreez: And in a further thought, is it worth adding an order note for this case, especially if we're not going to retry.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -726,8 +738,10 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
 				);
 
 				// Use the last charge within the intent or the full response body in case of SEPA.
-				$latest_charge = $this->get_latest_charge_from_intent( $response );
-				$this->process_response( ( ! empty( $latest_charge ) ) ? $latest_charge : $response, $renewal_order );
+				$latest_charge   = $this->get_latest_charge_from_intent( $response );
+				$charge_response = ( ! empty( $latest_charge ) ) ? $latest_charge : $response;
+				/** @var object $charge_response */
+				$this->process_response( $charge_response, $renewal_order );
```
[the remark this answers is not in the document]
daledupreez: I think we should add an explicit guard against `$latest_charge` being a string.

daledupreez (commented):
Thanks for your patience, @kamlesh08! I have taken another look at the changes and there are some subtle issues we still need to work through.

includes/compat/trait-wc-stripe-subscriptions.php
kamlesh08: > **can we bump this to an error so we are guaranteed to log something when this happens?**

Done.

> **And on a related note, can we (re)schedule a delayed retry?** 

I looked up the stack before wiring anything: if the lock holder fails, it marks the renewal order failed and the WooCommerce Subscriptions retry system schedules the retry, so that case is already covered. 

If it succeeds, no retry should run. 

The only gap is a stale lock from a crashed process (max 5 minutes). Re-firing the gateway hook after a delay isn't safe as-is because `trigger_gateway_renewal_payment_hook()` doesn't re-check `needs_payment()`, so it could double-charge a paid order. 

If you think the crashed-holder edge case is worth covering, I can add an Action Scheduler single action after lock expiry with an `explicit needs_payment()` recheck - happy to do that as a follow-up if you agree with the approach. 

For now the skip adds an order note.

> **is it worth adding an order note for this case**

Done. The skip path now adds a merchant-visible order note: "_Stripe: skipped this renewal payment attempt because another payment attempt for this order was already in progress_".

@daledupreez

includes/compat/trait-wc-stripe-subscriptions.php
kamlesh08: The caller now captures the lock value it wrote via `get_order_existing_payment_lock()` and passes the expiry into the retry chain; once `time()` passes it, retries stop with an error log, and the order goes through the normal failure path so WooCommerce Subscriptions retry handling takes over. 

The finally now re-reads the lock and only calls `unlock_order_payment()` when the stored value still matches ours, so a newer lock taken by another process after ours lapsed is left alone. 

Two new tests cover each: an expired lock stops after exactly one attempt, and unlock is never called when the stored lock differs from ours.

kamlesh08:
@daledupreez, I've addressed everything as per your suggestions.
Please do let me know if any other changes are needed.
Thank you.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -501,7 +501,52 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+			WC_Stripe_Logger::error( "Stripe: skipping duplicate renewal attempt for order {$order_id} because the payment lock is already held." );
+			$renewal_order->add_order_note( __( 'Stripe: skipped this renewal payment attempt because another payment attempt for this order was already in progress.', 'woocommerce-gateway-stripe' ) );
+			return;
+		}
+
+		// Remember the exact lock value this process wrote. The lock expires after 5 minutes while
+		// the retries below can run longer, so the expiry bounds the retry loop, and the finally
+		// must only release the lock while it is still ours — deleting the meta unconditionally
+		// could release a newer lock acquired by a concurrent process after ours lapsed.
+		$our_lock = (string) $order_helper->get_order_existing_payment_lock( $renewal_order );
```
daledupreez, line 514: Not a common edge case, but we should probably validate that `$our_lock` is a valid timestamp if we're going to be using it as an int in logic below.
```suggestion
		$our_lock = $order_helper->get_order_existing_payment_lock( $renewal_order );
		$is_valid_lock = false;
		if ( is_int( $our_lock ) ) {
			$is_valid_lock = true;
		} elseif( is_string( $our_lock ) && ctype_digit( $our_lock ) ) {
			// We could also check for non-zero values in the condition above
			$is_valid_lock = true;
		}
		if ( ! $is_valid_lock ) {
			// Log error and add order note
			return;
		}
		$lock_expiry = (int) $our_lock; // We can pass this value deeper down
		$our_lock    = (string) $our_lock; // We can use this for the equality comparison
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -501,7 +501,52 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+
+		try {
+			$this->process_subscription_payment_attempt( $amount, $renewal_order, $retry, $previous_error, (int) $our_lock );
+		} finally {
+			if ( (string) $order_helper->get_order_existing_payment_lock( $renewal_order ) === $our_lock ) {
```
daledupreez, line 526: Note that we also need the checks above to make sure that `$our_lock !== ''`, as we don't need to call unlock when no lock is currently taken out.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -501,7 +501,52 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+	 * re-acquiring it (which would let a concurrent renewal charge in the gap).
+	 *
+	 * @param float        $amount         The amount to charge.
+	 * @param WC_Order     $renewal_order  The renewal order.
+	 * @param bool         $retry          Should we retry the process?
+	 * @param object|false $previous_error Previous error object.
+	 * @param int          $lock_expiry    Unix timestamp at which the caller's payment lock expires. 0 when unknown.
+	 * @return void
+	 */
+	private function process_subscription_payment_attempt( $amount, $renewal_order, $retry = true, $previous_error = false, $lock_expiry = 0 ) {
```
daledupreez: Non-blocking nit: can we add strict types for the `$lock_expiry` argument, as we fully control it. The others are passed in and could be other types.
```suggestion
	private function process_subscription_payment_attempt( $amount, $renewal_order, $retry = true, $previous_error = false, int $lock_expiry = 0 ) {
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -818,10 +884,59 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+
+			// WC_Stripe_API::request() returns the decoded response body, which is not
+			// guaranteed to be an object; the renewal flow and process_response() read the
+			// charge via object property access. Round-tripping through json_decode() turns
+			// associative arrays (including nested ones) into objects while leaving JSON lists
+			// as arrays, which is what the downstream readers expect.
+			if ( is_array( $response ) ) {
+				$encoded_response = wp_json_encode( $response );
+				$decoded_response = is_string( $encoded_response ) ? json_decode( $encoded_response ) : null;
+				$response         = $decoded_response instanceof stdClass ? $decoded_response : (object) $response;
```
daledupreez: It feels like using `if/else` here will be a bit easier than the current code, which is packing in a lot of work on two lines. Also, we're not explicitly indicating that we want objects and not arrays, so it's worth passing in false to `json_decode()`. Also, I think that trying to cast any other value to an object is not the right approach, as it means we have a sequential array, which is not what any caller would expect.
```suggestion
				if ( is_string( $encoded_response ) ) {
					$decoded_response = json_decode( $encoded_response, false );
					$response         = $decoded_response instanceof stdClass ? $decoded_response : null;
				} else {
					$response = null;
				}
```

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -257,6 +262,760 @@ public function test_renewal_successful() {
...
+			}
+		);
+		$this->assertNotEmpty( $matching_errors, 'Expected a duplicate-attempt error mentioning the order id.' );
+
+		// The skip must also leave a merchant-visible trail on the order itself.
+		$notes          = wc_get_order_notes( [ 'order_id' => $renewal_order->get_id() ] );
+		$matching_notes = array_filter(
+			wp_list_pluck( $notes, 'content' ),
+			function ( $content ) {
+				return false !== strpos( $content, 'already in progress' );
```
daledupreez: Not blocking, but we can use `str_contains()` for readability.
```suggestion
				return str_contains( $message, $order_id );
			}
		);
		$this->assertNotEmpty( $matching_errors, 'Expected a duplicate-attempt error mentioning the order id.' );

		// The skip must also leave a merchant-visible trail on the order itself.
		$notes          = wc_get_order_notes( [ 'order_id' => $renewal_order->get_id() ] );
		$matching_notes = array_filter(
			wp_list_pluck( $notes, 'content' ),
			function ( $content ) {
				return str_contains( $content, 'already in progress' );
```

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -257,6 +262,760 @@ public function test_renewal_successful() {
...
+		$renewal_order = wc_get_order( $renewal_order->get_id() );
+
+		$this->assertSame( [], $api_requests );
+		$this->assertGreaterThan( time(), $order_helper->get_order_existing_payment_lock( $renewal_order ) );
+
+		$order_id        = (string) $renewal_order->get_id();
+		$matching_errors = array_filter(
+			$logged_errors,
+			function ( $message ) use ( $order_id ) {
+				return false !== strpos( $message, $order_id );
```
daledupreez: It may be worth making this a more specific check -- just mentioning the order ID could easily be true for other order notes.

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -257,6 +262,760 @@ public function test_renewal_successful() {
...
+			$this->wc_gateway_stripe->process_subscription_payment( 20, $renewal_order, false, false );
+		} finally {
+			remove_filter( 'pre_http_request', $pre_http_request_response_callback, 10 );
+			remove_action( 'wc_gateway_stripe_process_payment_charge', $capture_lock );
+		}
+
+		$renewal_order = wc_get_order( $renewal_order->get_id() );
+
+		// The lock must still be held while the successful charge is processed...
+		$this->assertGreaterThan( time(), $lock_during_processing );
```
daledupreez, line 571: Nit: It may be better to assert that the lock expired after a time that was before processing, and possibly event check that it expires less than 5 minutes in the future, i.e. `time() + 5 * MINUTE_IN_SECONDS`.

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -257,6 +262,760 @@ public function test_renewal_successful() {
...
+					'source'         => $source,
+					'source_object'  => (object) [
+						'type' => WC_Stripe_Payment_Methods::SEPA,
+					],
+					'payment_method' => null,
+				]
+			);
+
+		$pre_http_request_response_callback = function ( $preempt, $request_args, $url ) use ( $charges_api, &$requested, $renewal_order ) {
+			if ( 0 !== strpos( $url, $charges_api ) ) {
```
daledupreez: Non-blocking it: we can use `str_starts_with()`, as it is shimmed by WordPress.
```suggestion
			if ( ! str_starts_with( $url, $charges_api ) ) {
```

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -257,6 +262,760 @@ public function test_renewal_successful() {
...
+				return $preempt;
+			}
+
+			$requested[] = $url;
```
daledupreez: Not blocking, but a key change in this logic is to ensure that SEPA renewals now lock the order where they didn't before. It is worth adding some logic in this request mocking to confirm that the lock was taken out during the API call? It could also be applicable for a number of other tests, but this is something we should explicitly confirm because it is new.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -800,7 +841,32 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
 
 				// Use the last charge within the intent or the full response body in case of SEPA.
 				$latest_charge = $this->get_latest_charge_from_intent( $response );
-				$this->process_response( ( ! empty( $latest_charge ) ) ? $latest_charge : $response, $renewal_order );
+
+				// get_latest_charge_from_intent() can return a bare charge id, and
+				// process_response() reads the charge via object property access, so a string
+				// must be resolved to the full charge object before it is handed over.
+				if ( is_string( $latest_charge ) && '' !== $latest_charge ) {
+					$latest_charge = $this->get_charge_object( $latest_charge ); // @phpstan-ignore-line (get_charge_object is defined in the gateway classes using this trait)
```
daledupreez: Nit: Rather than using direct access that relies on the trait usage, can we explicitly call the method on the gateway instance?
```suggestion
					$latest_charge = WC_Stripe::get_instance()->get_main_stripe_gateway()->get_charge_object( $latest_charge );
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -800,7 +841,32 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+
+				$charge_response = ( ! empty( $latest_charge ) ) ? $latest_charge : $response;
+
+				// process_response() reads the charge via object property access; mirror the SEPA
+				// normalization so an associative-array response becomes an object and cannot fatal
+				// there. JSON lists stay arrays, as the readers expect.
+				if ( is_array( $charge_response ) ) {
+					$encoded_charge  = wp_json_encode( $charge_response );
+					$decoded_charge  = is_string( $encoded_charge ) ? json_decode( $encoded_charge ) : null;
+					$charge_response = $decoded_charge instanceof stdClass ? $decoded_charge : (object) $charge_response;
```
daledupreez: We have two code blocks doing the array->JSON->object round trip. Given that I had some comments on the other code location, it feels like it would be worth pushing this into a helper method. That may also make it easier to implement early returns and simplify the logic.

daledupreez (commented):
I have some more comments from inspection, but a lot of them are pretty minor.

client/classic/upe/__tests__/payment-processing.test.js
```diff
@@ -1769,10 +1769,6 @@ describe( 'payment-processing', () => {
 			);
 		} );
 
-		// WooCommerce posts a multiselect as implode( ', ', $value ), so both of the
```
daledupreez, line 1772: @kamlesh08, it looks like the comment-only changes from commit edc524b778385ed23c283339e952a21a7c30389c include cleanup of unrelated comments in other files, and those updates should not be handled as part of this PR. The changes in `client/classic/upe/__tests__/payment-processing.test.js` and `client/classic/upe/payment-processing.js` definitely fall into this set, and the commit also removes comments from a number of PHP unit test methods. Can you review the set of changes in the commit to confirm that they are all intentional and match your intent?

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -499,21 +499,360 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+
+	/**
+	 * Returns the timeout budget for one renewal attempt.
+	 *
+	 * Three Stripe calls may run before renewal; a fourth timeout is the safety margin.
+	 *
+	 * @return int Timeout budget in seconds.
+	 */
+	private function get_subscription_payment_request_timeout_budget(): int {
+		return 4 * WC_Stripe_API::REQUEST_TIMEOUT;
```
daledupreez: Minor note: this is only the initial value for the HTTP requests, as the `timeout` value may be modified by lower-level WordPress hooks. Those modifications may be to increase or decrease the value.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -499,21 +499,360 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+			return false;
+		}
+
+		$error_data = get_object_vars( $previous_error );
+
+		return ! empty( $source_object->status )
+			&& 'chargeable' === $source_object->status
+			&& 'idempotency_error' === ( $error_data['type'] ?? '' )
+			&& is_string( $error_data['message'] ?? null )
+			&& 1 === preg_match( '/Keys for idempotent requests can only be used with the same parameters they were first used with./i', $error_data['message'] );
```
daledupreez: Not blocking, but this feels like it could be relatively brittle due to the check for specific error text.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -548,74 +887,123 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+				$renewal_order->add_order_note( $localized_message );
+				throw new WC_Stripe_Exception(
+					"Stripe: subscription renewal order {$order_id} was not attempted because its payment lock could not cover the complete Stripe request chain.",
+					$localized_message
+				);
 			}
 
+			$idempotency_key_filter = function ( $idempotency_key, $request ) use ( $retry_interval ) {
+				return $this->change_subscription_renewal_idempotency_key( $idempotency_key, $request, $retry_interval );
+			};
```
daledupreez: Nit: It feels like we should only define this function when `$should_change_idempotency_key` is true.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -817,10 +1248,120 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
+	 *
+	 * @param mixed $charge Value resolved from a Stripe response.
+	 * @return bool
+	 * @phpstan-assert-if-true \stdClass $charge
+	 */
+	private function is_subscription_renewal_charge_object( $charge ): bool {
+		// get_charge_object() returns whatever the API layer decoded, and WP_Error carries no
+		// charge data, so neither can be trusted to be a Charge without checking.
+		return is_object( $charge ) && ! $charge instanceof WP_Error && ! empty( $charge->id );
+	}
```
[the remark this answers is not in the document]
daledupreez: Agreed - it feels like we should check the object type as well. ([docs](https://docs.stripe.com/api/charges/object#charge_object-object))
```suggestion
	private function is_subscription_renewal_charge_object( $charge ): bool {
		// get_charge_object() returns whatever the API layer decoded, and WP_Error carries no
		// charge data, so neither can be trusted to be a Charge without checking.
		return is_object( $charge ) && ! $charge instanceof WP_Error && ! empty( $charge->id ) && 'charge' === ( $charge->object ?? '' );
	}
```

daledupreez (changes requested):
@kamlesh08, your recent rounds of changes have introduced multiple changes in scope that aren't described well in the PR:
1. You have added a mechanism to cache charges during subscription renewals.
2. You have modified the order locking values and added in-process ownership semantics to the lock processing, including more complex SQL logic to manage the data.
3. You've implemented a lot of logic around error handling and reporting for various possible issues.
4. You have changed the details for how we check and update idempotency keys for subscription renewals.

While I agree with us needing most of these changes, pushing them all into one PR makes it _much_ harder to review, understand, and test the collective changes, especially given that the PR description doesn't capture many of the details. It feels like we should try to split these changes up into multiple PRs. We _really_ want to get the baseline locking improvements shipped, so that feels like the core of what we should aim to keep in this PR, while the caching and order locking changes feel like they should each be split into dedicated PRs.

Please let us know if you think you will be able to get that moving pretty soon.

kamlesh08:
@daledupreez, the split you asked for is done, and this PR is ready for another look.

This PR now holds only the baseline locking fix: the renewal takes the payment lock once, honours it, and releases it after all retries. The other changes moved to dedicated PRs:

- #5897: atomic lock owner row and owned-lock helpers
- #5901: renewal path uses the owned lock
- #5900: daily cleanup of expired owner rows
- #5898: expand the charge on the renewal confirm request
- #5899: idempotency keys on renewal retries

I think the merge order should be #5591, #5897, #5901, #5900, then #5898 and #5899. 

Each description lists its base, the rebase it needs, and impact. 

**On your inline notes**: the unrelated comment cleanup is done, the charge check now tests object === 'charge' (in #5898), the idempotency error match uses the error type on the renewal path (in #5899), and the key filter attaches only when an attempt can use it. 

**One open question**: the SEPA branch runs only for the removed stripe_sepa gateway id, so it may be dead code.

daledupreez:
@coderabbitai resume

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -499,7 +499,41 @@ protected function is_charge_blocked_by_radar( $response ) {
 	 * @return void
 	 */
 	public function process_subscription_payment( $amount, $renewal_order, $retry = true, $previous_error = false ) {
-		$order_locked = false;
+		$order_helper = WC_Stripe_Order_Helper::get_instance();
+		$order_id     = $renewal_order->get_id();
+
+		// One lock covers the whole attempt, retries included, so a concurrent renewal cannot charge in between.
+		if ( $order_helper->lock_order_payment( $renewal_order ) ) {
+			// Error, not warning: warnings are dropped unless debug logging is on.
```
daledupreez: I think we can rephrase this slightly for clarity.
```suggestion
			// Log an error to ensure we log the error even when debug logging is disabled.
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -499,7 +499,41 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+			WC_Stripe_Logger::error( "Stripe: skipping duplicate renewal attempt for order {$order_id} because the payment lock is already held." );
+			$renewal_order->add_order_note( __( 'Stripe: skipped this renewal payment attempt because another payment attempt for this order was already in progress.', 'woocommerce-gateway-stripe' ) );
+			return;
+		}
+
+		// The 5-minute lock can lapse mid-retry; only release it while it is still ours.
+		$our_lock    = (string) $order_helper->get_order_existing_payment_lock( $renewal_order );
+		$lock_expiry = (int) $our_lock;
```
daledupreez: I think we should add another check to make sure we have a valid lock before we proceed, as `$our_lock` being `''` would be highly problematic when removing the lock and because `(int) ''` is `0` which could lead to unexpected handling of `$lock_expiry` too.
```suggestion
		$our_lock = (string) $order_helper->get_order_existing_payment_lock( $renewal_order );
		if ( '' === $our_lock ) {
			WC_Stripe_Logger::error( "Stripe: skipping renewal attempt for order {$order_id} because we could not confirm the payment lock." );
			$renewal_order->add_order_note( __( 'Stripe: skipped this renewal payment attempt because we could not confirm the payment lock.', 'woocommerce-gateway-stripe' ) );
			return;
		}

		$lock_expiry = (int) $our_lock;
```

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -499,7 +499,41 @@ protected function is_charge_blocked_by_radar( $response ) {
 	 * @return void
 	 */
 	public function process_subscription_payment( $amount, $renewal_order, $retry = true, $previous_error = false ) {
-		$order_locked = false;
+		$order_helper = WC_Stripe_Order_Helper::get_instance();
+		$order_id     = $renewal_order->get_id();
+
+		// One lock covers the whole attempt, retries included, so a concurrent renewal cannot charge in between.
+		if ( $order_helper->lock_order_payment( $renewal_order ) ) {
```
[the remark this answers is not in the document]
daledupreez: Noting that we're shifting the logic in this PR and not fully guarding against all possible concurrency issues. The addition of more granular, atomic locking is being tackled in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5897.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -499,7 +499,57 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+		}
+
+		$lock_expiry = (int) $our_lock;
+		$our_lock    = (string) $our_lock;
+
+		try {
+			$this->process_subscription_payment_attempt( $amount, $renewal_order, $retry, $previous_error, $lock_expiry );
+		} finally {
+			if ( (string) $order_helper->get_order_existing_payment_lock( $renewal_order ) === $our_lock ) {
+				$order_helper->unlock_order_payment( $renewal_order );
```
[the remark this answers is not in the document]
daledupreez: Noting that true atomicity is being addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5897. The early return I added above for `$our_lock` being `''` also ensures that we're removing our lock.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -575,29 +611,39 @@ public function process_subscription_payment( $amount, $renewal_order, $retry =
...
-					if ( $retry ) {
-						// Don't do anymore retries after this.
+					// Stop retrying once the lock lapses; a concurrent renewal may hold its own lock by then.
+					$lock_expired = $lock_expiry > 0 && time() > $lock_expiry;
+
+					if ( $retry && ! $lock_expired ) {
+						// Last retry.
 						if ( 5 <= $this->retry_interval ) { // @phpstan-ignore-line (retry_interval is defined in classes using this class)
-							$this->process_subscription_payment( $amount, $renewal_order, false, $response->error );
+							$this->process_subscription_payment_attempt( $amount, $renewal_order, false, $response->error, $lock_expiry );
```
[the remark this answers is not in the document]
daledupreez: There are no overrides of this method, so the issue is somewhat moot.

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -66,14 +69,16 @@ public function set_up() {
 	 * Tears down the stuff we set up.
 	 */
 	public function tear_down() {
-		WC_Stripe_Helper::delete_main_stripe_settings();
-
 		// The tests in this file do not mock ALL the calls to the Stripe API, and as we use mocked API keys they trigger the 401 rate-limiter,
 		// this is not a problem for these tests as they don't depend on the reponses.
 		//
+		// Delete before the settings go: the cache key depends on the current mode.
```
daledupreez: I think we can clarify the wording.
```suggestion
		// Delete the cached data before we delete the settings: the cache key depends on the current mode.
```

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -264,6 +269,672 @@ public function test_renewal_successful() {
...
+					'source'         => $source,
+					'source_object'  => (object) [
+						'type' => WC_Stripe_Payment_Methods::CARD,
+					],
+					'payment_method' => null,
+				]
+			);
+
+		$order_helper = WC_Stripe_Order_Helper::get_instance();
+		$order_helper->lock_order_payment( $renewal_order );
```
daledupreez: I think we should confirm that we got the lock before we run the main function in the `try` block below. The test will fail in weird ways if we don't get a lock.
```suggestion
		$lock_result = $order_helper->lock_order_payment( $renewal_order );
```

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -264,6 +269,672 @@ public function test_renewal_successful() {
...
+				],
+				'cookies'  => [],
+				'filename' => null,
+			];
+		};
+
+		add_filter( 'pre_http_request', $pre_http_request_response_callback, 10, 3 );
+
+		try {
+			$this->wc_gateway_stripe->process_subscription_payment( 20, $renewal_order, false, false );
```
daledupreez, line 378: The other half of the initial lock checks from [this comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5591/changes#r3924230392).
```suggestion
			$this->assertFalse( $lock_result );
			$this->wc_gateway_stripe->process_subscription_payment( 20, $renewal_order, false, false );
```

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -264,6 +269,672 @@ public function test_renewal_successful() {
...
+			->getMock();
+		$order_helper_spy->expects( $this->once() )
+			->method( 'lock_order_payment' )
+			->willReturnCallback( [ $real_order_helper, 'lock_order_payment' ] );
+		$order_helper_spy->method( 'unlock_order_payment' )
+			->willReturnCallback( [ $real_order_helper, 'unlock_order_payment' ] );
+
+		$instance_property = new ReflectionProperty( WC_Stripe_Order_Helper::class, 'instance' );
+		$instance_property->setAccessible( true );
+		$original_instance = $real_order_helper;
```
daledupreez: Not blocking, but it feels redundant to have `$real_order_helper` and `$original_instance` pointing at the same initial object.

daledupreez (commented):
I have some comments from inspection; will loop back for testing shortly.

phpstan-baseline.neon
```diff
@@ -2328,6 +2292,12 @@ parameters:
 			count: 1
 			path: includes/class-wc-stripe-order-handler.php
 
+		-
+			message: '#^Parameter \#1 \$response of method WC_Stripe_Payment_Gateway\:\:process_response\(\) expects object, object\|string given\.$#'
+			identifier: argument.type
+			count: 1
+			path: includes/abstracts/abstract-wc-stripe-payment-gateway.php
```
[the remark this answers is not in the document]
daledupreez: We should definitely address this issue before merging as well.

includes/compat/trait-wc-stripe-subscriptions.php
daledupreez: Added in c13e672e775f653827369943017fadfcc5f312a3.

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
daledupreez: Addressed in 35125c450a7fe2fa2e5ed636cd5851aa2a14fd74.

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
daledupreez: Addressed in 9fc8e2d0974caaeb524a43c0f98d12bb891ba645.

includes/compat/trait-wc-stripe-subscriptions.php
daledupreez: Addressed in c13e672e775f653827369943017fadfcc5f312a3.

phpstan-baseline.neon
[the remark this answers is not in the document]
daledupreez: Addressed in bd21718fdd5ba9cff1a57e3a1ee9c367ebed13fd.

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
daledupreez: Added in 35125c450a7fe2fa2e5ed636cd5851aa2a14fd74.

daledupreez (approved):
Thanks for your persistence in getting this PR ready, @kamlesh08! Things are looking good and I validated the locking behaviour via some manual collision tests.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -499,7 +499,47 @@ protected function is_charge_blocked_by_radar( $response ) {
...
+			$renewal_order->add_order_note( __( 'Stripe: skipped this renewal payment attempt because another payment attempt for this order was already in progress.', 'woocommerce-gateway-stripe' ) );
+			return;
+		}
+
+		// The 5-minute lock can lapse mid-retry; only release it while it is still ours.
+		$our_lock = (string) $order_helper->get_order_existing_payment_lock( $renewal_order );
+		if ( '' === $our_lock ) {
+			WC_Stripe_Logger::error( "Stripe: skipping renewal attempt for order {$order_id} because we could not confirm the payment lock." );
+			$renewal_order->add_order_note( __( 'Stripe: skipped this renewal payment attempt because we could not confirm the payment lock.', 'woocommerce-gateway-stripe' ) );
+			return;
```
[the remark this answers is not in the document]
daledupreez: If the lock read returns `''`, then our lock has been removed somehow, so we should return early rather than proceeding without a confirmed lock.

kamlesh08:
Thank you so much, @daledupreez. 🙇‍♂️

tests/phpunit/compat/class-wc-stripe-subscription-renewal-test.php
```diff
@@ -264,6 +269,674 @@ public function test_renewal_successful() {
...
+				],
+				'cookies'  => [],
+				'filename' => null,
+			];
+		};
+
+		add_filter( 'pre_http_request', $pre_http_request_response_callback, 10, 3 );
+
+		try {
+			$this->assertFalse( $lock_result, 'Expect a new lock to be acquired before we run the main test.' );
```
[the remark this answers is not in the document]
daledupreez: Added in ea7a3b3ef236faec8e2c6d740b2ac054629d89b1.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -570,6 +578,14 @@ private function process_subscription_payment_attempt( $amount, $renewal_order,
 				]
 			);
 
+			if ( $this->is_lock_expired( $lock_expiry ) ) {
+				$this->log_renewal_lock_expired( $renewal_order, true );
+				throw new WC_Stripe_Exception(
+					"Failed to process renewal for order $order_id. The payment lock has expired.",
+					__( 'Payment lock expired', 'woocommerce-gateway-stripe' )
+				);
+			}
```
jaclync: non-blocking nit: these are exactly the same as L554-560, maybe worth considering a private helper like `throw_if_renewal_lock_expired`.

daledupreez: Good idea. I added a new `log_and_throw_if_lock_expired()` helper in 7e1537dfe374674788debea07ee3a769adddca12.

PR #5592, opened 2026-06-17
Fix express checkout tax preview for taxable product add-ons

kamlesh08:
Fixes STRIPE-XXXX

The express checkout (Apple Pay / Google Pay) product-data preview shows the selected add-on cost in the line-item total (`$line_subtotal + $addon_value`) but calculated tax from the product subtotal only. WooCommerce folds the add-on cost into the taxable cart-item price, so the wallet sheet understated tax — and therefore the total — whenever a taxable Product Add-On was selected. The shopper then authorizes less than the order is actually charged.

This taxes the full displayed line amount (`$total` = product subtotal + add-ons) via `get_taxes_like_cart()`, matching how the real cart taxes the add-on cost.

This is the same preview-tax-base family as #5585 (which scaled the base by quantity). It is **stacked on #5585** because it edits the same statement — this PR targets the #5585 branch so the diff is just the add-on change. Please merge #5585 first; this can then be retargeted to `develop`. Together they make the preview tax base both quantity- and add-on-correct.

### How this was found / verified

Reproduced on WordPress 7.0 / WooCommerce 10.8.1 with Product Add-Ons active and a single 10% tax rate, using a $10 taxable product and a $5 taxable add-on:

- Real cart (what is charged): item priced $15, tax **$1.50**.
- Preview before this change: tax **$1.00** (computed on $10) → wallet total $16.00 vs charged $16.50.
- A/B control: without an add-on the preview already matched the cart ($1.00 = $1.00); the $0.50 gap appeared only with the add-on, isolating the cause to the preview tax base (Product Add-Ons taxes the add-on correctly in the real cart).

## Testing instructions

1. Enable taxes with a standard rate (e.g. 10%); set price display to **excluding tax**.
2. Install Product Add-Ons; add a taxable add-on (e.g. a $5 flat-fee checkbox) to a $10 taxable product.
3. Enable Apple Pay / Google Pay express checkout and open the product page.
4. Select the add-on and open the wallet sheet.
5. **Before:** tax is calculated on $10 ($1.00); the wallet total is $16.00 — less than the $16.50 the order is actually charged.
6. **After:** tax reflects $15 ($1.50); the wallet total is $16.50 and matches the cart/order.

---

-   [x] Covered with tests (PHPUnit regression test for a taxable add-on in the preview tax base)

### Changelog entry

-   Fix - Include selected add-on costs in the Apple Pay / Google Pay tax preview so the wallet total is not understated when a taxable product add-on is selected

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@ xxxx-xx-xx - version 10.9.0
 * Fix - Show an error notice when a Stripe OAuth connection attempt fails its security check
 * Dev - Deprecate the internal WC_Stripe_Feature_Flags::is_amazon_pay_available() helper now that Amazon Pay is permanently enabled
 * Fix - Calculate the Apple Pay / Google Pay product tax preview on the selected quantity instead of a single unit
+* Fix - Include selected add-on costs in the Apple Pay / Google Pay tax preview so the wallet total is not understated when a taxable product add-on is selected
 
 2026-06-15 - version 10.8.2
 * Fix - Disable Adaptive Pricing when webhooks are disabled
--- a/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
@@ -414,8 +414,11 @@ class WC_Stripe_Express_Checkout_Ajax_Handler {
 				'amount' => WC_Stripe_Helper::get_stripe_amount( $total ),
 			];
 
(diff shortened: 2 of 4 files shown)


kamlesh08:
Folding this into #5585. It's the same one-line change to the express-checkout tax base, and #5585 now taxes the full taxable line total (quantity **and** add-ons) with a regression test for the add-on case. Closing here to avoid two stacked PRs patching the same statement.

PR #5593, opened 2026-06-17
Add client-side Stripe.js origin assertion as defense-in-depth against checkout skimmers

mgascam:
Partially fixes WOOPMNT-6210

Minimal, client-only defense-in-depth against checkout skimmers that swap Stripe.js. Ports the equivalent client-side origin assertion from WooPayments (woocommerce/woocommerce-payments#11853).

The gateway loads Stripe.js via a rewritable WordPress script handle and instantiates it from the ambient `window.Stripe` with no provenance check — on Blocks too, since the `@stripe/stripe-js` loader reuses the already-loaded global. On a compromised store, repointing that handle to a look-alike origin makes the card iframe render from an attacker while charges still succeed.

This adds an origin assertion at the two Stripe-instantiation choke points (`WCStripeAPI.createStripe()` and `client/blocks/load-stripe.js`): a new helper (`client/stripe-utils/verify-stripe-js-origin.js`) checks the loaded Stripe.js tag's origin is `https://js.stripe.com`; on a mismatch it warns (console only) and throws a generic error, blocking the payment. The `#stripe-js` handle is looked up explicitly so a legitimate `js.stripe.com` tag elsewhere in the DOM cannot mask a repointed handle.

**Scope / limitations:** defense-in-depth, not a gateway vulnerability fix. The gateway isn't the entry vector, and a fully-compromised box can bypass this. It defends the Stripe.js *substitution* vector specifically (the variant seen in the wild), not an attacker-injected custom form.

## Testing instructions

Simulate the attack on a test site, then open checkout (console warns, payment blocked on classic, Blocks, and express checkout):

    add_filter( 'script_loader_src', fn( $src, $h ) => 'stripe' === $h ? 'https://js.evil.example/v3/' : $src, PHP_INT_MAX, 2 );

---

-   [x] Covered with tests (or have a good reason not to test in description)

### DIFF
--- a/client/api/__tests__/index.test.js
+++ b/client/api/__tests__/index.test.js
@@ -24,6 +24,7 @@ describe( 'WCStripeAPI', () => {
 		} );
 
 		afterEach( () => {
+			warnSpy.mockRestore();
 			delete global.Stripe;
 			document.getElementById( 'stripe-js' )?.remove();
 		} );
--- a/client/api/__tests__/index.test.js
+++ b/client/api/__tests__/index.test.js
@@ -1,0 +1,65 @@
+import WCStripeAPI from '..';
+
+jest.mock( 'wcstripe/stripe-utils', () => ( {
+	getStripeServerData: jest.fn(),
+	getStripeDevWidgetOptions: jest.fn( () => ( {} ) ),
+} ) );
+
+describe( 'WCStripeAPI', () => {
+	describe( 'getStripe', () => {
+		let warnSpy;
+
+		const addStripeScriptTag = ( src ) => {
+			const script = document.createElement( 'script' );
+			script.id = 'stripe-js';
+			script.setAttribute( 'src', src );
+			document.body.appendChild( script );
+		};
+
+		beforeEach( () => {
+			global.Stripe = jest.fn( () => ( {} ) );
+			warnSpy = jest
+				.spyOn( console, 'warn' )
+				.mockImplementation( () => {} );
+		} );
+
+		afterEach( () => {
+			delete global.Stripe;
+			document.getElementById( 'stripe-js' )?.remove();
+		} );
+
+		it( 'instantiates Stripe when Stripe.js was loaded from the official origin', () => {
+			addStripeScriptTag( 'https://js.stripe.com/clover/stripe.js' );
+			const api = new WCStripeAPI( { key: 'pk_test_123', locale: 'en' } );
+
+			expect( api.getStripe() ).toBeTruthy();
+			expect( global.Stripe ).toHaveBeenCalledWith( 'pk_test_123', {
+				locale: 'en',
+			} );
+			expect( warnSpy ).not.toHaveBeenCalled();
+		} );
+
+		it( 'warns and blocks when Stripe.js was loaded from an unexpected origin', () => {
+			addStripeScriptTag(
+				'https://js.stripe.com.evil.example/clover/stripe.js'
+			);
+			const api = new WCStripeAPI( { key: 'pk_test_123', locale: 'en' } );
+
+			expect( () => api.getStripe() ).toThrow(
+				/provenance check failed/
+			);
+			expect( global.Stripe ).not.toHaveBeenCalled();
+			expect( warnSpy ).toHaveBeenCalled();
+		} );
+
+		it( 'warns and blocks when no Stripe.js tag is present', () => {
+			const api = new WCStripeAPI( { key: 'pk_test_123', locale: 'en' } );
+
+			expect( () => api.getStripe() ).toThrow(
+				/provenance check failed/
+			);
+			expect( global.Stripe ).not.toHaveBeenCalled();
+			expect( warnSpy ).toHaveBeenCalled();
+		} );
+	} );
+} );
--- a/client/api/index.js
+++ b/client/api/index.js
@@ -11,6 +11,7 @@ import {
 	getStripeServerData,
 	getStripeDevWidgetOptions,
 } from 'wcstripe/stripe-utils';
+import { assertStripeJsOrigin } from 'wcstripe/stripe-utils/verify-stripe-js-origin';
 import {
 	PAYMENT_INTENT_STATUS_REQUIRES_ACTION,
 	PAYMENT_METHOD_CASHAPP,
@@ -95,6 +96,8 @@ export default class WCStripeAPI {
 	 * @return {Object} The Stripe instance.
 	 */
 	createStripe( key, locale, betas = [] ) {
+		assertStripeJsOrigin();
+
 		const options = {
 			locale,
 			...getStripeDevWidgetOptions(),
--- a/client/stripe-utils/verify-stripe-js-origin.js
+++ b/client/stripe-utils/verify-stripe-js-origin.js
@@ -1,0 +1,79 @@
+/**
+ * The only origin Stripe.js is allowed to be served from.
+ *
+ * @type {string}
+ */
+export const STRIPE_JS_ORIGIN = 'https://js.stripe.com';
+
+/**
+ * Inspects the Stripe.js script tag on the page and verifies that it was
+ * loaded from the official Stripe origin.
+ *
+ * Only the origin is compared (not the full URL), so Stripe.js path changes
+ * (`/v3/`, `/clover/stripe.js`, named release trains) do not cause mismatches.
+ *
+ * The `#stripe-js` id is the fingerprint WordPress renders for the `stripe`
+ * script handle, so a repointed handle is read by id regardless of its
+ * current src. It is looked up explicitly (not via a selector list) so it
+ * always takes precedence: `querySelector` on a comma-separated selector
+ * returns the first match in document order, so a legitimate js.stripe.com
+ * tag inserted earlier in the DOM could otherwise mask a repointed handle.
+ * Only when no handle tag exists do we fall back to any script served from
+ * the official origin.
+ *
+ * @param {Document} [doc] The document to inspect. Defaults to the global document.
+ * @return {Object} Result object with `ok` (boolean), `detectedSrc` (string|null), and `detectedOrigin` (string|null).
+ */
+export const verifyStripeJsOrigin = ( doc = document ) => {
+	const tag =
+		doc.querySelector( '#stripe-js' ) ??
+		doc.querySelector( 'script[src^="https://js.stripe.com/"]' );
+
+	if ( ! tag || ! tag.src ) {
+		return { ok: false, detectedSrc: null, detectedOrigin: null };
+	}
+
+	try {
+		const origin = new URL( tag.src ).origin;
+		return {
+			ok: origin === STRIPE_JS_ORIGIN,
+			detectedSrc: tag.src,
+			detectedOrigin: origin,
+		};
+	} catch ( error ) {
+		return { ok: false, detectedSrc: tag.src, detectedOrigin: null };
+	}
+};
+
+/**
+ * Asserts that Stripe.js was loaded from the official Stripe origin,
+ * blocking payment processing otherwise (fail closed).
+ *
+ * This is a defense-in-depth measure against checkout skimmers that swap the
+ * Stripe.js script for a look-alike hosted on an attacker-controlled origin.
+ *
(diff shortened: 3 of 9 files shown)


client/stripe-utils/verify-stripe-js-origin.js
```diff
@@ -0,0 +1,79 @@
...
+ * @return {Object} Result object with `ok` (boolean), `detectedSrc` (string|null), and `detectedOrigin` (string|null).
+ */
+export const verifyStripeJsOrigin = ( doc = document ) => {
+	const tag =
+		doc.querySelector( '#stripe-js' ) ??
+		doc.querySelector( 'script[src^="https://js.stripe.com/"]' );
+
+	if ( ! tag || ! tag.src ) {
+		return { ok: false, detectedSrc: null, detectedOrigin: null };
+	}
```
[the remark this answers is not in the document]
mgascam: Good call. `#stripe-js` is the WordPress handle's id, but `querySelector('#stripe-js')` matches any element — so a non-script or src-less collision failed closed instead of falling back to a legitimately-loaded tag, blocking checkout for no reason. Fixed in ff8b1cfd9: the `#stripe-js` match only takes precedence when it's a `SCRIPT` with a non-empty `src`, otherwise we fall back to the `script[src^="https://js.stripe.com/"]` query. Masking protection is intact — a repointed handle is still a script-with-src, so it's always caught. Added a test for the fallback branch.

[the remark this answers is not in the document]
mgascam: With the `#stripe-js` guard above (fall back when it isn't a usable script), a src-less `#stripe-js` no longer reaches this branch — it falls through to the origin query, and we report "no script tag" only when there's genuinely no usable tag, which is accurate. I'd rather not add an errorCode/reason field just for console wording; the diagnostic is developer-facing and the two-branch message already distinguishes the meaningful cases (wrong origin vs. nothing usable).

client/api/__tests__/index.test.js
```diff
@@ -0,0 +1,65 @@
...
+			global.Stripe = jest.fn( () => ( {} ) );
+			warnSpy = jest
+				.spyOn( console, 'warn' )
+				.mockImplementation( () => {} );
+		} );
+
+		afterEach( () => {
+			delete global.Stripe;
+			document.getElementById( 'stripe-js' )?.remove();
+		} );
```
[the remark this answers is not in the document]
mgascam: This suite runs under `restoreMocks: true` (`tests/js/jest.config.js`), so Jest restores all spies before every test — the `console.warn` spy can't leak. Added the explicit `mockRestore()` anyway in eeaacf2b1 to stay consistent with `load-stripe.test.js`.

client/stripe-utils/verify-stripe-js-origin.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * The only origin Stripe.js is allowed to be served from.
+ *
+ * @type {string}
+ */
+export const STRIPE_JS_ORIGIN = 'https://js.stripe.com';
+
+/**
```
wjrosa, line 10: It would be good to reduce the length of this block

client/stripe-utils/verify-stripe-js-origin.js
```diff
@@ -0,0 +1,84 @@
...
+			detectedSrc: tag.src,
+			detectedOrigin: origin,
+		};
+	} catch ( error ) {
+		return { ok: false, detectedSrc: tag.src, detectedOrigin: null };
+	}
+};
+
+/**
+ * Asserts that Stripe.js was loaded from the official Stripe origin,
```
wjrosa: This one too

client/api/index.js
```diff
@@ -95,6 +96,8 @@ export default class WCStripeAPI {
 	 * @return {Object} The Stripe instance.
 	 */
 	createStripe( key, locale, betas = [] ) {
```
wjrosa, line 98: `getStripe()` is called unwrapped in several spots (e.g. `upe-element.js:128` `!! api.getStripe()`, `classic/upe/index.js:148`), so the new throw propagates there as an uncaught exception rather than the generic `{ error }` the two `loadStripe` wrappers resolve with. Fail-closed is right; worth confirming those surfaces degrade cleanly (PM availability / form validation) and don't just blow up.

client/stripe-utils/verify-stripe-js-origin.js
```diff
@@ -0,0 +1,84 @@
...
+	const reason =
+		result.detectedSrc === null
+			? 'no Stripe.js script tag was found'
+			: `Stripe.js was loaded from an unexpected origin (${ result.detectedSrc })`;
+
+	// eslint-disable-next-line no-console
+	console.warn(
+		`WooCommerce Stripe Gateway: blocking checkout — ${ reason }. Expected Stripe.js from ${ STRIPE_JS_ORIGIN }.`
+	);
+	throw new Error( 'Stripe.js provenance check failed.' );
```
wjrosa: This message can reach shoppers via checkout error rendering — wrap in `__( 'Stripe.js provenance check failed.', 'woocommerce-gateway-stripe' )` for i18n, consistent with other user-facing strings.

wjrosa (commented):
Solid defense-in-depth, good test coverage and the id-vs-fallback masking handling is careful. Two small notes inline.

client/stripe-utils/verify-stripe-js-origin.js
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @param {Document} [doc] The document to inspect. Defaults to the global document.
+ * @return {Object} Result object with `ok` (boolean), `detectedSrc` (string|null), and `detectedOrigin` (string|null).
+ */
+export const verifyStripeJsOrigin = ( doc = document ) => {
+	const handleTag = doc.querySelector( '#stripe-js' );
+	const tag =
+		handleTag?.tagName === 'SCRIPT' && handleTag.src
+			? handleTag
+			: doc.querySelector( 'script[src^="https://js.stripe.com/"]' );
```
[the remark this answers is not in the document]
mgascam: Agreed, and good refinement over the earlier pass. Fixed in 43f8a96c0: `#stripe-js` is now authoritative whenever it's a `<script>` (dropped the `&& src` condition), so a srcless or unparsable handle fails closed instead of falling back — closing the inject-srcless-handle + benign-fallback-script bypass. The fallback now applies only when `#stripe-js` is absent or not a script (the benign id-collision case). Added a test that a srcless `<script id="stripe-js">` does not fall back even when a legit js.stripe.com script is present.

client/stripe-utils/verify-stripe-js-origin.js
```diff
@@ -0,0 +1,84 @@
+/**
+ * The only origin Stripe.js is allowed to be served from.
+ *
+ * @type {string}
+ */
+export const STRIPE_JS_ORIGIN = 'https://js.stripe.com';
```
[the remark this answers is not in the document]
mgascam: Done in 43f8a96c0 — the fallback selector is now built from `STRIPE_JS_ORIGIN` (`script[src^="${ STRIPE_JS_ORIGIN }/"]`) so the constant stays the single source of truth.

client/stripe-utils/verify-stripe-js-origin.js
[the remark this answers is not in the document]
mgascam: Done in 43f8a96c0 — the fallback selector is now built from `STRIPE_JS_ORIGIN` (`script[src^="${ STRIPE_JS_ORIGIN }/"]`) so the constant stays the single source of truth.

client/stripe-utils/verify-stripe-js-origin.js
```diff
@@ -0,0 +1,84 @@
...
+	const result = verifyStripeJsOrigin( doc );
+
+	if ( result.ok ) {
+		return;
+	}
+
+	const reason =
+		result.detectedSrc === null
+			? 'no Stripe.js script tag was found'
+			: `Stripe.js was loaded from an unexpected origin (${ result.detectedSrc })`;
```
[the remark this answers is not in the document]
mgascam: Kept the full `detectedSrc` in the warning — it's console-only (developer-facing) and more useful for pinpointing the exact script, and `detectedOrigin` is `null` for the unparsable-URL case, which would make the message less precise there. Reworded "unexpected origin" → "unexpected source" in 43f8a96c0 so the full URL isn't mislabeled as an origin.

client/stripe-utils/verify-stripe-js-origin.js
mgascam: Trimmed both docstrings in 43f8a96c0 to the essential rationale (why `#stripe-js` is authoritative; why the thrown message stays generic), dropping the `querySelector`/DOM-order narration.

client/stripe-utils/verify-stripe-js-origin.js
mgascam: Trimmed both docstrings in 43f8a96c0 to the essential rationale (why `#stripe-js` is authoritative; why the thrown message stays generic), dropping the `querySelector`/DOM-order narration.

client/api/index.js
mgascam: Confirmed fail-closed on every surface, and the throw always lands at init, not mid-payment: `getStripe()` memoizes, so the first call wins — Blocks `canMakePayment` drops the card method (Blocks evaluates it defensively), classic mount is already `.catch()`-wrapped so UPE fields don't render, and express mount fails so ECE buttons don't render. The `confirmPayment`/`confirmSetup` callers are downstream of a successful mount, so they're unreachable when the origin is tampered.

The one rough edge was the express mount throwing uncaught — wrapped it in 3c917ae15 (try/catch + early return) so it degrades to "button doesn't render" like the other surfaces.

client/stripe-utils/verify-stripe-js-origin.js
mgascam: Good catch — wrapped the thrown message in `__( 'Stripe.js provenance check failed.', 'woocommerce-gateway-stripe' )` (with the `@wordpress/i18n` import) in 43f8a96c0, since it can surface to shoppers via checkout error rendering.

mgascam:
Thanks for the review @wjrosa! I addressed your comments, ready for a new pass when you get a moment 🙏

client/stripe-utils/verify-stripe-js-origin.js
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @param {Document} [doc] The document to inspect. Defaults to the global document.
+ * @return {Object} Result with `ok` (boolean), `detectedSrc` (string|null), and `detectedOrigin` (string|null).
+ */
+export const verifyStripeJsOrigin = ( doc = document ) => {
+	const handleTag = doc.querySelector( '#stripe-js' );
+	const tag =
+		handleTag?.tagName === 'SCRIPT'
+			? handleTag
+			: doc.querySelector( `script[src^="${ STRIPE_JS_ORIGIN }/"]` );
```
wjrosa: One residual masking vector remains: a non-`<script>` `#stripe-js` (e.g. an injected `<div id="stripe-js">`) placed before the real handle makes `querySelector('#stripe-js')` return the non-script, so we fall through to the origin query — which can match an earlier benign `js.stripe.com` script and report `ok` while a look-alike populated `window.Stripe`. That's the same masking the authoritative handle is meant to prevent, just reached via a non-script id collision instead of a srcless script. Using `doc.querySelector('script#stripe-js')` keeps the real handle authoritative (and lets you drop the `tagName` check); the fallback then fires only when no script handle exists at all.

mgascam: Sharp catch — right, the `tagName` guard still let a `<div id="stripe-js">` decoy divert to the fallback, which could match a benign official-origin script while a look-alike owned `window.Stripe`. Switched to `doc.querySelector('script#stripe-js')` in 08ed69dd2 so the real handle stays authoritative regardless of any non-script id collision (and dropped the `tagName` check); the fallback now fires only when there's no script handle at all. Added a test for the div-decoy-before-repointed-handle case.

wjrosa (approved):
Looks good to me and tests well. Thanks for addressing the suggestions!

Image

PR #5594, opened 2026-06-17
Use a live Store API nonce for Express Checkout order requests

Mayisha:
Related to STRIPE-1214

The final `/wc/store/v1/checkout` request sends the `wc_store_api` nonce that's rendered into the checkout page HTML. When that page is served from full-page cache, the embedded nonce can be stale or missing for guests — and logged-in users bypass full-page cache (Excluding `/wp-json/wc/store/` from cache doesn't help: the nonce lives in the checkout *page*, not the REST endpoint.)

Every Store API cart response rotates the `Nonce` header, and the cart calls the wallet sheet already makes (`updateCustomer` / `selectShippingRate`) store that rotated value. This change reuses it for the create-order / pay-for-order requests instead of the page-rendered nonce, falling back to the page nonce when no cart call has run. It is a synchronous lookup of a value we already hold.

## Testing instructions
**No-regression smoke check**
1. Enable Express Checkout (Apple Pay / Google Pay) under WooCommerce → Settings → Payments → Stripe → Express Checkout.
2. Use a classic shortcode `[woocommerce_checkout]` page.
3. As a **guest**, add a product to the cart, go to checkout, and pay with Apple Pay / Google Pay.
4. Expected: order completes successfully.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/api/index.js
+++ b/client/api/index.js
@@ -606,9 +606,10 @@ export default class WCStripeAPI {
 	 * Creates order based on Express Checkout ECE payment method.
 	 *
 	 * @param {Object} orderData Order data.
+	 * @param {Object} headers   Optional header overrides (e.g. a refreshed `Nonce`).
 	 * @return {Promise} Promise for the request to the server.
 	 */
-	expressCheckoutECECreateOrder( orderData ) {
+	expressCheckoutECECreateOrder( orderData, headers = {} ) {
 		return this.postToBlocksAPI(
 			'/wc/store/v1/checkout',
 			{
@@ -620,6 +621,7 @@ export default class WCStripeAPI {
 				'X-WCSTRIPE-EXPRESS-CHECKOUT-NONCE':
 					getExpressCheckoutData( 'nonce' )
 						?.wc_store_api_express_checkout,
+				...headers,
 			}
 		);
 	}
@@ -630,15 +632,21 @@ export default class WCStripeAPI {
 	 * @param {number} order        The order ID.
 	 * @param {Object} orderDetails Order details, including order key and billing email.
 	 * @param {Object} paymentData  Order data.
+	 * @param {Object} headers      Optional header overrides (e.g. a refreshed `Nonce`).
 	 * @return {Promise} Promise for the request to the server.
 	 */
-	expressCheckoutECEPayForOrder( order, orderDetails, paymentData ) {
+	expressCheckoutECEPayForOrder(
+		order,
+		orderDetails,
+		paymentData,
+		headers = {}
+	) {
 		paymentData.shipping_address = orderDetails.shippingAddress;
 
 		const billingEmail = orderDetails.billingEmail ?? '';
 		const key = orderDetails.orderKey ?? '';
 		const url = `/wc/store/v1/checkout/${ order }?key=${ key }&billing_email=${ billingEmail }`;
-		return this.postToBlocksAPI( url, paymentData );
+		return this.postToBlocksAPI( url, paymentData, headers );
 	}
 
 	/**
--- a/client/express-checkout/cart-api.js
+++ b/client/express-checkout/cart-api.js
@@ -43,6 +43,27 @@ export default class ExpressCheckoutCartApi {
 	}
 
 	/**
+	 * Returns the freshest Store API nonce we already hold, without a request.
+	 *
+	 * The nonce localized into the page can be stale or absent for guests when
+	 * the checkout page is served from full-page cache (logged-in users bypass
+	 * the cache, which is why only guests hit `woocommerce_rest_missing_nonce`).
+	 * Every Store API cart response rotates the `Nonce` header, which the cart
+	 * calls made while the wallet sheet is open (`updateCustomer` /
+	 * `selectShippingRate`) store here — prefer that over the page nonce. This
+	 * deliberately makes no request of its own so it adds no latency to the
+	 * express checkout submit path; it only reuses a value already fetched.
+	 *
+	 * @return {string|undefined} The freshest Store API nonce available.
+	 */
+	getStoreApiNonce() {
+		return (
+			this.cartRequestHeaders.Nonce ||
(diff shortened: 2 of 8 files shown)


Mayisha:
@coderabbitai review

readme.txt
```diff
@@ -165,6 +165,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Fix - Decommission the previously configured webhook before connecting via OAuth so reconnecting to a different Stripe account no longer leaves an orphaned webhook on the old account
 * Fix - Send the billing address to Stripe on the Pay for Order page so payments aren't incorrectly blocked by Stripe Radar rules
 * Update - Replace shipping AJAX endpoints with Store API calls for Express Checkout Element
+* Fix - Reuse the Store API nonce rotated during the Express Checkout cart requests when completing the order, so guest Apple Pay/Google Pay payments no longer fail with a nonce error on cached classic checkout pages
```
wjrosa: I would make this a bit shorter, something like:
```suggestion
* Fix - Reuse the Store API nonce from Express Checkout cart requests when completing orders, preventing guest Apple Pay/Google Pay failures on cached classic checkout pages.
```

client/express-checkout/event-handler.js
```diff
@@ -123,16 +123,23 @@ export const onConfirmHandler = async ( params ) => {
 		return handleChangePaymentMethodFlow( params );
 	}
 
+	// Prefer the nonce the cart calls already rotated while the wallet sheet was
```
wjrosa: This comment could also be shorter

client/express-checkout/cart-api.js
```diff
@@ -42,6 +42,27 @@ export default class ExpressCheckoutCartApi {
 		return response.json();
 	}
 
+	/**
+	 * Returns the freshest Store API nonce we already hold, without a request.
+	 *
+	 * The nonce localized into the page can be stale or absent for guests when
```
wjrosa: Same as previous comments

client/express-checkout/cart-api.js
```diff
@@ -42,6 +42,27 @@ export default class ExpressCheckoutCartApi {
...
+	 * `selectShippingRate`) store here — prefer that over the page nonce. This
+	 * deliberately makes no request of its own so it adds no latency to the
+	 * express checkout submit path; it only reuses a value already fetched.
+	 *
+	 * @return {string|undefined} The freshest Store API nonce available.
+	 */
+	getStoreApiNonce() {
+		return (
+			this.cartRequestHeaders.Nonce ||
+			getExpressCheckoutData( 'nonce' )?.wc_store_api
```
wjrosa, line 63: Virtual/downloadable carts never fire `shippingAddressChangeHandler`/`shippingRateChangeHandler`, so `cartRequestHeaders.Nonce` stays empty and this falls back to the same stale page nonce — guest checkout for no-shipping products on a cached page would still hit the nonce error. Is a cart call guaranteed to run before confirm?

client/api/index.js
```diff
@@ -620,6 +621,7 @@ export default class WCStripeAPI {
 				'X-WCSTRIPE-EXPRESS-CHECKOUT-NONCE':
 					getExpressCheckoutData( 'nonce' )
 						?.wc_store_api_express_checkout,
```
wjrosa, line 628: `X-WCSTRIPE-EXPRESS-CHECKOUT-NONCE` (checked by `is_express_checkout_context()`) is also page-localized and never rotated here — is it not subject to the same cache staleness, or does its `wp_verify_nonce` survive for guests where the Store API nonce doesn't?

client/express-checkout/payment-flow.js
```diff
@@ -63,9 +64,14 @@ const processOrder = async ( {
 	confirmationTokenId,
 	order = 0,
 	orderDetails = {},
+	storeApiNonce,
 } ) => {
 	let orderResponse;
 
+	// The page-localized nonce can be stale/absent on cached guest checkout
+	// pages, so override it with the live one resolved from the cart.
+	const headers = storeApiNonce ? { Nonce: storeApiNonce } : {};
```
wjrosa, line 72: Same rationale now lives in three spots (here, `getStoreApiNonce` docblock, `onConfirmHandler`); consider trimming this to a one-liner and keeping the full explanation only on `getStoreApiNonce`.

client/api/index.js
Mayisha: Good catch — same staleness, not an asymmetry: `wc_store_api_express_checkout` sits next to `wc_store_api` (`class-wc-stripe-express-checkout-element.php:237-238`) and uses the same guest `wp_verify_nonce`
  formula and ~2-tick lifetime, so they expire together.

  But it's a different failure from this one. `woocommerce_rest_missing_nonce` comes only from WC core's `check_nonce()` on the `wc_store_api` header. A stale EC nonce just makes
  `is_express_checkout_context()` return false — no error, it only skips address normalization / postcode relaxation. So it can't reproduce the nonce error, and the cart routes already depend on it today, so
  this PR doesn't add exposure. Leaving it as-is.

client/express-checkout/cart-api.js
Mayisha: You're right — no cart call is guaranteed before confirm. `getCart()` was never called in the shortcode flow, so for virtual/downloadable carts (no shipping step) `cartRequestHeaders.Nonce` stayed empty
  and we fell back to the stale page nonce.

  Fixed in 2cfddf92f: `getStoreApiNonce()` is now async and, when nothing has rotated a nonce yet, does a `GET /cart` to harvest a fresh one — a GET needs no nonce (`requires_nonce()` is false for reads) but
  still returns a rotated `Nonce` header. Falls back to the page nonce if that request fails. The common path (a cart call already rotated one) is unchanged, so the extra request only fires for the
  no-shipping case. Added tests for the harvest and both fallbacks.

Mayisha:

PR #5598, opened 2026-06-17
Add woocommerce_agentic_commerce_should_sync_product filter and deprecate the Stripe-prefixed one

wjrosa:
Towards STRIPE-1168

Merchants/integrations can exclude products from the Agentic Commerce sync (full feed, inventory updates, archive events) via a `should_sync_product` filter. Until now that hook used the Stripe-specific `wc_stripe_` prefix, which prevents other (non-Stripe) plugins from registering against a shared, generic hook name.

This PR:

- Adds a new canonical filter **`woocommerce_agentic_commerce_should_sync_product`** using the WooCommerce-core `woocommerce_` prefix, so the hook can be integrated by other plugins.
- Deprecates **`wc_stripe_agentic_commerce_should_sync_product`** via `apply_filters_deprecated()` (since 10.9.0). The deprecated filter's result seeds the default for the new filter, so:
  - Adapters still hooking the old name keep working (with a standard deprecation notice).
  - A hook on the new name takes precedence over the deprecated one.
- The existing `wp_validate_boolean()` normalization and the full lifecycle/convergence contract in the docblock are preserved.

**How can this break?** An adapter relying on the old filter would only be affected if the old hook stopped firing — it still does, and is covered by a dedicated backward-compat test. The new filter overriding the old one is also covered by a test.

## Testing instructions

This is a developer-facing hook change with no merchant UI. To verify:

1. Add a snippet hooking the **new** filter to exclude a product:
   ```php
   add_filter( 'woocommerce_agentic_commerce_should_sync_product', function ( $sync, $product ) {
       return 'My Excluded Product' !== $product->get_name();
   }, 10, 2 );
   ```
   Trigger an Agentic Commerce feed sync and confirm the product is omitted from the feed, inventory, and archive surfaces.
2. Repeat with the **deprecated** filter `wc_stripe_agentic_commerce_should_sync_product` and confirm it still excludes the product (and emits a `_deprecated_hook` notice with `WP_DEBUG` on, pointing to the new name).
3. Run the suites: `npm run test:php -- --filter WC_Stripe_Agentic_Commerce_Product_Mapper_Test` (and the inventory-tracker / feed-validator / integration agentic-commerce suites).

---

### Changelog entry

Added to the 10.9.0 block in both `changelog.txt` and `readme.txt`:

> Update - Add the woocommerce_agentic_commerce_should_sync_product filter for excluding products from Agentic Commerce sync and deprecate the Stripe-prefixed wc_stripe_agentic_commerce_should_sync_product so other plugins can share the hook

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,7 +23,7 @@ xxxx-xx-xx - version 10.9.0
 * Fix - Show an error notice when a Stripe OAuth connection attempt fails its security check
 * Dev - Deprecate the internal WC_Stripe_Feature_Flags::is_amazon_pay_available() helper now that Amazon Pay is permanently enabled
 * Tweak - Don't show the Stripe API outage notice on local and development sites, where Stripe is often unreachable for benign reasons
-* Update - Add the woocommerce_agentic_commerce_should_sync_product filter for excluding products from Agentic Commerce sync and deprecate the Stripe-prefixed wc_stripe_agentic_commerce_should_sync_product so other plugins can share the hook
+* Update - Deprecate the wc_stripe_agentic_commerce_should_sync_product filter in favor of the shareable woocommerce_agentic_commerce_should_sync_product
 
 2026-06-15 - version 10.8.2
 * Fix - Disable Adaptive Pricing when webhooks are disabled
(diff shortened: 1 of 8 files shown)


Mayisha (approved):
Looks good to me 👍

malithsen (approved):
Looks good

PR #5604, opened 2026-06-18
Add booking products to cart via Store API in Express Checkout

wjrosa:
Towards STRIPE-694

> **Stacked on #5602** (variable products). Base is that branch, so review #5602 first; this diff is bookings-only and GitHub will retarget to `develop` once #5602 merges.

Booking products on the product page added to the cart through the legacy `wc_stripe_add_to_cart` endpoint. WooCommerce Bookings **3.0.0+** ships a Store API integration (`WC_Bookings_Cart_Store_API`) that accepts a `booking_configuration` param on `cart/add-item`, so this routes supported bookings through the Store API like the other product types.

- New `buildBookingConfiguration` helper reads the Bookings product-page form and produces `{ date, resource_id?, local_timezone }`:
  - **date/day** bookings — from the `wc_bookings_field_start_date_{year,month,day}` inputs → `Y-m-d 00:00:00`
  - **datetime** bookings — converts the chosen slot's ISO-8601 value (`get_time_as_iso8601()`) → `Y-m-d H:i:s`
  - **month** bookings — from the `_yearmonth` (`Y-n`) field → first of the month
  - optional resource (`wc_bookings_field_resource`) and `local_timezone`
- It returns **null** — falling back to the legacy endpoint — for bookings the Store API can't represent (**persons**, **customer-defined duration**) or when no date/time is selected yet.
- Gated by a `has_bookings_store_api` capability flag localized from PHP (`class_exists( 'WC_Bookings_Cart_Store_API' )`), so stores on older Bookings keep the legacy path.

**Server-contract scope:** the Bookings Store API route only accepts `date`/`resource_id`/`local_timezone` (it computes duration from the product). Persons and customer-defined-duration bookings are therefore **not** representable and intentionally stay on legacy.

**How can this break?** A wrong `date` format or unsupported type reaching the Store API would 400/409 the add. **Mitigations:** the builder is unit-tested (10 cases) and returns null for unsupported types; the server contract was verified live for both date and datetime bookings.

## Testing instructions

Requires **WooCommerce Bookings 3.0.0+** and Express Checkout enabled on product pages.

1. Create a bookable product (fixed duration; try a day-unit and an hour-unit one).
2. On the product page, choose a date (and a time slot for the hour product).
3. DevTools → Network, filter `cart`. Click the Express Checkout button (Apple Pay / Google Pay / Link).
4. **Expected:** add-to-cart is `POST /wp-json/wc/store/v1/cart/add-item` with a `booking_configuration` payload — **not** `wc-ajax=wc_stripe_add_to_cart` — and the correct booking (date/time, resource) lands in the cart. Complete the purchase and confirm the order's booking.
5. **Fallback checks:** a **persons-based** or **customer-defined-duration** booking still uses the legacy `wc_stripe_add_to_cart` endpoint; and with WooCommerce Bookings deactivated/older than 3.0.0, all bookings use legacy.

Note: the wallet button can't be exercised over plain local HTTP; use HTTPS/a real device for step 4. The Store API booking contract and the form→payload extraction were verified live (date add-item → 201 with correct date; datetime → 201 with correct date+time).

---

### Changelog entry

> Update - Add booking products to the cart via the Store API in Express Checkout when WooCommerce Bookings supports it

**Post merge**

### DIFF
--- a/client/api/index.js
+++ b/client/api/index.js
@@ -551,8 +551,9 @@ export default class WCStripeAPI {
 	/**
 	 * Add product to cart from product page (legacy version, non-StoreAPI).
 	 *
-	 * Used for booking products, whose add-to-cart is not routed through the
-	 * Store API (see `expressCheckoutAddToCart` for the Store API path).
+	 * Fallback for booking products that can't be expressed as a Store API
+	 * `booking_configuration` (persons / customer-defined duration); the
+	 * representable ones go through `expressCheckoutAddToCart`.
 	 *
 	 * @param {Object} productData Product data.
 	 * @return {Promise} Promise for the request to the server.
--- a/client/api/index.js
+++ b/client/api/index.js
@@ -533,16 +533,16 @@ export default class WCStripeAPI {
 		// Rename qty to quantity to match StoreAPI expected parameter.
 		const { qty, ...rest } = productData;
 		const quantity = qty ?? 1;
-		const blocksApiProductData = {
+		const storeApiProductData = {
 			...rest,
 			quantity,
 		};
 
 		const data = applyFilters(
 			'wcstripe.express-checkout.cart-add-item',
-			blocksApiProductData
+			storeApiProductData
 		);
-		return this.postToBlocksAPI( '/wc/store/v1/cart/add-item', data );
+		return this.postToStoreApi( '/wc/store/v1/cart/add-item', data );
 	}
 
 	/**
@@ -577,7 +577,7 @@ export default class WCStripeAPI {
 				},
 			} );
 			const removeItemsPromises = cartData.items.map( ( item ) => {
-				return this.postToBlocksAPI( '/wc/store/v1/cart/remove-item', {
+				return this.postToStoreApi( '/wc/store/v1/cart/remove-item', {
 					key: item.key,
 					booking_id: bookingId,
 				} );
@@ -610,7 +610,7 @@ export default class WCStripeAPI {
 	 * @return {Promise} Promise for the request to the server.
 	 */
 	expressCheckoutECECreateOrder( orderData ) {
-		return this.postToBlocksAPI(
+		return this.postToStoreApi(
 			'/wc/store/v1/checkout',
 			{
 				...orderData,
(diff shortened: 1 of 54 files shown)


client/express-checkout/utils/bookings.js
```diff
@@ -0,0 +1,90 @@
...
+		return null;
+	}
+
+	const fieldValue = ( name ) =>
+		form.querySelector( `[name="${ name }"]` )?.value?.trim() ?? '';
+
+	// Persons / customer-defined-duration bookings can't be represented; legacy.
+	if (
+		form.querySelector( '[name^="wc_bookings_field_persons"]' ) ||
+		form.querySelector( '[name$="_to_year"]' )
```
Mayisha: The `[name$="_to_year"]` check for falling back to legacy is too narrow — it only catches a subset of customer-defined-duration bookings, so the rest get the wrong duration.

  `_to_year` (`wc_bookings_field_start_date_to_year`) is only rendered when the range picker is active, and `is_range_picker_enabled()` requires *all* of: range picker enabled, duration unit `day`/`month`,
  customer-defined duration, **and** base duration `1`. Customer-defined-duration bookings outside that slice — nights (the classic hotel "N nights" case), hours/minutes (datetime picker hard-disables the
  range picker), weeks, or day/month with range picker off — render `wc_bookings_field_duration` instead and have no `_to_year`. Those slip past the fallback and go to the Store API, which forces
  `wc_bookings_field_duration => $product->get_duration()` and drops the customer's chosen duration. Wrong duration, wrong price, no error.

  Suggest keying the fallback on the duration field instead:

  ```
  if (
      form.querySelector( '[name^="wc_bookings_field_persons"]' ) ||
      form.querySelector( '[name="wc_bookings_field_duration"]' )
  ) {
      return null;
  }
  ```

  `duration_field()` is rendered unconditionally for `duration_type === 'customer'` (every unit), which is exactly the set the endpoint can't represent, so there are no false positives and it subsumes the
  range-picker case too. The existing `_to_year` test is fine for the range subset — could we add a `wc_bookings_field_duration` case (e.g. a nights booking) to cover the gap?

client/api/index.js
```diff
@@ -551,7 +548,8 @@ export default class WCStripeAPI {
 	/**
 	 * Add product to cart from product page (legacy version, non-StoreAPI).
 	 *
-	 * @todo Remove this once WC 9.7.0 is the min. required version.
+	 * Used for booking products, whose add-to-cart is not routed through the
```
Mayisha: Nit: as this branch adds booking add-to-cart to the store API, we can update the comment to mention that it is a fallback for the booking products.

Mayisha (approved):
Functional test went well. I tested with 2 booking products from the template and paid with Google Pay-
✅ Product Tutoring - Store API flow, Google Pay
✅ Product Boat Cruise - Legacy flow, Google Pay

I have an AI-flagged feedback. Please check if that is worth fixing before the merge.

client/express-checkout/utils/bookings.js
wjrosa: Good catch — switched the fallback to `[name="wc_bookings_field_duration"]` in 92f3b5528. Since `duration_field()` renders for every `duration_type === 'customer'` unit, this now covers nights/hours/weeks too and subsumes the range-picker case, so I dropped the `_to_year` check. Added a nights test and updated the existing range test to carry the duration field (range pickers render both).

client/api/index.js
wjrosa: Updated in 92f3b5528 — the docblock now describes it as the fallback for bookings that can't be expressed as a `booking_configuration`, pointing at `expressCheckoutAddToCart` for the Store API path.

PR #5605, opened 2026-06-18
Scope admin gateway filter to block cart and checkout editors only

Mayisha:
Fixes STRIPE-1230

`WC_Stripe::add_gateways()` filtered out the non-card UPE payment methods on **every** admin request except order edit / refund. That meant any extension reading the gateway list elsewhere in wp-admin — AutomateWoo workflow conditions, WooCommerce Subscriptions payment-method selectors, etc. — could no longer see SEPA, iDEAL, Affirm and the rest whenever Optimized Checkout (OCS) was enabled.

That filter only exists to avoid block-support warnings when the Cart/Checkout/Express Checkout/OCS blocks are previewed in the editor. So instead of running on all of `is_admin()`, it now runs only on actual block-editor screens via a new `is_block_editor_request()` helper (`get_current_screen()->is_block_editor()`).

## Testing instructions

1. Stripe enabled with **Optimized Checkout on**, and a non-card method (e.g. SEPA or Affirm) enabled.
3. Edit the Checkout page in the block editor and confirm there are **no** block-support console warnings for the Stripe payment blocks.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


wjrosa:
Nit: the PR description have a duplicate part at the top

includes/class-wc-stripe.php
```diff
@@ -593,7 +593,7 @@ public function add_gateways( $methods ) {
 		// 3. Filter out UPE payment methods that are not enabled at checkout, as they are not available in the checkout block
 		// and including them in the list results in warnings about block support
 		// when viewing the payment methods block in the editor for the cart and checkout pages.
-		if ( is_admin() && ! $this->is_order_management_context() ) {
+		if ( is_admin() && $this->is_block_editor_request() ) {
```
wjrosa: `is_block_editor()` is true for *every* block-editor screen (any post/page edit, the site editor, widgets) — not only screens that can host the Stripe cart/checkout/ECE/OCS blocks. So the global gateway list is still filtered while editing unrelated content, and an extension reading gateways on those screens hits the same symptom this PR fixes, just narrowed. Intended, or worth limiting to screens that can actually contain these blocks?

includes/class-wc-stripe.php
```diff
@@ -620,48 +620,23 @@ function ( $method ) use ( $is_oc_enabled ) {
...
+	 * previewed in the editor. Scoping it to block-editor screens keeps the full gateway list
+	 * available everywhere else in wp-admin.
 	 *
 	 * @return bool
 	 */
-	private function is_order_management_context(): bool {
-		// Refund AJAX action — fired when the merchant clicks "Refund via Gateway".
-		if ( 'woocommerce_refund_line_items' === $this->get_request_var( 'action', INPUT_POST ) ) {
-			return true;
+	private function is_block_editor_request(): bool {
```
wjrosa: The dropped data-provider cases were the only coverage that order-edit / refund-AJAX screens keep all gateways registered (the refund-processing guarantee the old docblock called out). That's now only implicit in the generic `is_block_editor => false` cases — consider an explicit order-edit/refund case so a future change here can't silently regress refunds.

includes/class-wc-stripe.php
```diff
@@ -620,48 +620,23 @@ function ( $method ) use ( $is_oc_enabled ) {
...
+		if ( ! function_exists( 'get_current_screen' ) ) {
+			return false;
 		}
 
-		$page   = $this->get_request_var( 'page' );
-		$action = $this->get_request_var( 'action' );
+		$screen = get_current_screen();
```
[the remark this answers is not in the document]
Mayisha: Tested and confirmed that the changes works on block cart and checkout page editors.

includes/class-wc-stripe.php
Mayisha: Re-added the regression tests in [bf8e0a6](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5605/commits/bf8e0a62fea461ccf0fd856b02a3afa8dd6d1dd1)

includes/class-wc-stripe.php
Mayisha: Good call. I addressed this in [bf8e0a6](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5605/commits/bf8e0a62fea461ccf0fd856b02a3afa8dd6d1dd1)

wjrosa (approved):
Looks good to me 👍 . Thanks for addressing the comments

PR #5609, opened 2026-06-18
Clarify that manual capture doesn't apply to Agentic Commerce purchases

wjrosa:
Fixes STRIPE-1232

A merchant who enables **Manual capture** in the Stripe settings reasonably expects every order to be authorize-only. But it only applies to normal checkout — **Agentic Commerce purchases ignore it**, which is surprising.

Why: the plugin sets `capture_method` on PaymentIntents it creates (normal checkout, via `is_automatic_capture_enabled()`), but the **agentic** PaymentIntent is created by Stripe's hosted agentic checkout — not the plugin. The plugin only receives the finished session via the `checkout.session.completed` webhook and calls `payment_complete()` (`includes/agentic-commerce/class-wc-stripe-agentic-commerce-order-mapper.php`). It never creates the intent, so it **cannot** set its capture method; capture for agentic is controlled entirely by the **Stripe agentic commerce dashboard**.

Since the plugin can't make its own toggle drive agentic capture, this PR addresses the surprise directly: it adds a one-line clarification to the manual capture setting so merchants know the toggle doesn't govern agentic purchases.

Surfaces updated (same wording on both):
- The React **Manual capture** control help text (`client/settings/payments-and-transactions-section/manual-capture-control.js`) — the live settings UI.
- The legacy `capture` setting description (`includes/admin/stripe-settings.php`) — kept in parity.

> Agentic Commerce purchases follow the capture setting in your Stripe agentic commerce dashboard, not this option.

This is intentionally scoped to a notice. Making agentic orders *reflect* Stripe's actual capture state (e.g. set on-hold + leave uncaptured when Stripe only authorized) is a separate payment-flow change and is noted on the issue as a possible follow-up.

## Testing instructions

1. Go to **WooCommerce → Settings → Payments → Stripe → Settings → Payments & transactions**.
2. Under **Manual capture** (label: "Issue an authorization on checkout, and capture later"), read the help text below the checkbox.
3. Confirm it ends with: *"Agentic Commerce purchases follow the capture setting in your Stripe agentic commerce dashboard, not this option."*
4. (Legacy form parity) The same sentence appears on the `capture` field description in the underlying settings definition.

No behavior change — copy only.

### Changelog entry

* Tweak - Clarify on the manual capture setting that Agentic Commerce purchases follow the capture setting in the Stripe agentic commerce dashboard

---


client/settings/payments-and-transactions-section/__tests__/manual-capture-control.test.js
```diff
@@ -13,6 +13,18 @@ describe( 'ManualCaptureControl', () => {
 		useManualCapture.mockReturnValue( [ false, () => null ] );
 	} );
 
+	it( 'notes that agentic commerce purchases follow the Stripe dashboard capture setting', () => {
```
Mayisha: Nit and personal opinion. 
I have observed agents keep adding tests for every change, and sometimes the test does not add much value, just increases the test count. For example, I had one PR that adds a simple opacity in a checkbox and Claude added a test asserting that 🤦‍♀️ 
I feel this is one of the unnecessary kinds of tests. On second thought, we could may be assert that the whole note is rendered instead of this partial piece.

Mayisha (approved):
Looks good to me, just a nit comment about the test. 👍

aheckler:
IMHO it might not make sense to update the settings text for the main auth/capture setting. Why? Because everyone will see the "Agentic Commerce purchases follow the..." text there even if they do not use agentic, have no idea what agentic is, etc.

Some alternatives to think about:

- Only show the warning about agentic if people actually *enable* manual capture.
- Show the warning in the agentic settings section, not as part of the auth/capture setting.
- Make toggling the manual capture setting on automatically enable to same setting for agentic.

client/settings/payments-and-transactions-section/__tests__/manual-capture-control.test.js
wjrosa: I agree! But I will keep this one since the copy is the whole PR, but I tightened it: it now asserts the full agentic sentence (including "not this option.") instead of the fragment.

wjrosa:
Good call @aheckler ! It shouldn't sit in the always-visible help text that every merchant reads. I've moved it: the note now lives in the confirmation modal that already appears when you enable manual capture (alongside the existing caveats about supported methods and the 7-day window), and it now renders only when agentic commerce is enabled.

On the auto-sync idea: that one isn't possible. Agentic capture is configured entirely in the Stripe dashboard. Agentic payment intents are created server-side by Stripe's delegated checkout, so the plugin's `capture_method` never touches them, and there's no API to set the dashboard value remotely.

client/settings/payments-and-transactions-section/manual-capture-control.js
```diff
@@ -41,6 +41,11 @@ const ManualCaptureControl = () => {
 	const [ isConfirmationModalOpen, setIsConfirmationModalOpen ] =
 		useState( false );
 
+	// Agentic Commerce capture is configured in the Stripe dashboard, not by this
+	// plugin setting, so only flag the divergence to merchants who actually use it.
+	const isAgenticCommerceEnabled =
+		!! window.wc_stripe_settings_params?.is_agentic_commerce_enabled;
```
malithsen: This determines whether the agentic commerce feature flag is enabled, not whether the merchant has it enabled via the "Enable agentic commerce" checkbox in settings. 

I think that might be okay, since the Agentic Commerce feature is visible to merchants in settings, and the note might be useful if they later enable it. I just wanted to highlight the distinction.

malithsen (approved):
Looks good

Image

client/settings/payments-and-transactions-section/manual-capture-control.js
wjrosa: Nice catch, Malith! I've switched the gate to the merchant toggle, which makes more sense in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5609/commits/2674a1e2e4f96bff923ae23878fe630ece9cf428

Mayisha (approved):
:shipit:

PR #5610, opened 2026-06-18
PoC: Automatically configure test-mode when onboarding to a live account

malithsen:
Towards STRIPE-1174

Today, connecting a Stripe account only sets up one environment at a time. To use both live and test mode, a merchant has to run the "Connect to Stripe" flow twice, and each run switches the store's active mode. 

With this change, completing the live "Connect to Stripe" onboarding now also retrieves and stores the test mode credentials in the same step. So that "Enable test mode" toggle is immediately available, without having to go through test-mode onboarding flow separately.

> [!NOTE]
> This PR does not update the current UI. It's only a PoC to prove that configuring test-mode duing live onboarding flow is possible. If we proceed with this approach, we'd need additional UI changes.

## Testing instructions

1. As the connect-server component is not yet deployed, testing this requires setting it up locally. Follow the instructions in https://github.com/Automattic/woocommerce-connect-server/pull/3334
2. Configure the plugin to use your local connect server  with `define( 'WOOCOMMERCE_CONNECT_SERVER_URL', 'http://host.docker.internal:5500/' );`
3. Go through the live onboarding flow 
CleanShot 2026-06-19 at 08 55 52@2x
4. After onboarding, confirm test mode is now available - you can switch to test mode without doing a separate test connection.
CleanShot 2026-06-18 at 16 21 01@2x

6. Confirm the store is still in live mode (connecting did not flip it to test).
7. Enable test-mode and complete a test payment.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/connect/class-wc-stripe-connect.php
```diff
@@ -341,6 +341,11 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
 				WC_Stripe_Payment_Method_Configurations::maybe_migrate_payment_methods_from_db_to_pmc( true );
 			}
 
+			// A live Standard-OAuth onboarding can also carry the account's test keys
+			if ( 'connect' === $type && 'live' === $mode ) {
+				$this->save_oauth_test_keys_alongside_live( $result );
```
Mayisha, line 346: QQ: `save_oauth_test_keys_alongside_live` is called unconditionally without checking if test keys are already present or not. Is it intentional?

includes/connect/class-wc-stripe-connect.php
```diff
@@ -368,6 +373,47 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
...
+
+			update_option( 'wc_stripe_test_oauth_updated_at', time() );
+			update_option( 'wc_stripe_test_oauth_failed_attempts', 0 );
+			update_option( 'wc_stripe_test_oauth_last_failed_at', '' );
+
+			WC_Stripe_API::set_secret_key( $test_secret_key );
+			try {
+				WC_Stripe::get_instance()->account->configure_webhooks( 'test' );
+			} catch ( Exception $e ) {
+				WC_Stripe_Logger::error( 'OAuth: Failed to configure test webhooks during dual-fetch: ' . $e->getMessage() );
```
Mayisha, line 416: Note: I see webhooks are not configured in any mode and see this error in the logs.

```
OAuth: Failed to configure test webhooks during dual-fetch: There was a problem setting up your webhooks. You are not permitted to configure webhook endpoints on a connected account. Did you mean to create a Connect webhook on your account instead?
```

But I think this could be related to the test platform account.

malithsen: Yeah, I get that too. I think this has to do with my test platform account. We can retest this PR against production once we merge the Connect Server PR.

includes/connect/class-wc-stripe-connect.php
malithsen: It is intentional. I haven't tested, but Stripe [docs](https://docs.stripe.com/connect/oauth-reference#:~:text=Any%20existing%20access%20token%20with%20the%20same%20scope%20and%20mode%E2%80%94live%20or%20test%E2%80%94is%20revoked.) say that any existing keys created with the scope and mode would be revoked. I am not sure whether that applies only to keys created with `refresh_token` or to current ones, but replacing any existing test keys here seems safer. It also avoids any ambiguity we might encounter after we release this change.

malithsen:
@Mayisha Connect server change was merged. Can you retest this PR with the prod Connect server? I don't get the webhook configuration issues that we ran into [earlier](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5610#discussion_r3475858327) with my test connect account.

Mayisha:
This tests well now @malithsen 🎉 

Connecting a live account configured both test and live modes correctly. Webhooks are also configured for both of the modes.

**Live**

Screenshot 2026-07-17 at 12 55 28 AM

**Test**
Screenshot 2026-07-17 at 12 55 34 AM

includes/connect/class-wc-stripe-connect.php
```diff
@@ -368,6 +373,47 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
...
+		 *
+		 * @param object $result OAuth result that may carry testPublishableKey/testSecretKey.
+		 */
+		private function save_oauth_test_keys_alongside_live( $result ): void {
+			if ( ! $result instanceof stdClass || ! isset( $result->testPublishableKey, $result->testSecretKey ) ) { // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+				return;
+			}
+
+			$settings = WC_Stripe_Helper::get_stripe_settings();
+			$settings = is_array( $settings ) ? $settings : [];
```
Mayisha: Nit. This is redundant as `get_stripe_settings` has the same fallback.

Mayisha (approved):
Looks good from my end. 👍

includes/connect/class-wc-stripe-connect.php
malithsen: Thanks. Removed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5610/commits/feda9616f2428a5168a8e496cde91d6105f0bf34

malithsen:
@pierorocca @daledupreez Are we okay with shipping this PR in it's current state as an incremental improvement to the onboarding flow, or do you think we should wait for design changes being proposed in 0fiyiUe18lGU6kGyQNmXxE-fi-3915_41920

To be clear, this PR is entirely additive. For Live mode merchants, this allows switching to test mode without having to go through test mode onboarding again. No other onboarding UI changes are included here.

Before
CleanShot 2026-07-17 at 12 52 42@2x

Now
CleanShot 2026-06-18 at 16 21 01@2x

includes/connect/class-wc-stripe-connect.php
malithsen: @daledupreez I think I misrepresented how we handled this case in our 1:1. 
We, in fact, replace any existing test keys.

includes/connect/class-wc-stripe-connect.php
```diff
@@ -368,6 +373,47 @@ private function save_stripe_keys( $result, $type = 'connect', $mode = 'live' )
...
+			WC_Stripe_Helper::update_main_stripe_settings( $settings );
+
+			update_option( 'wc_stripe_test_oauth_updated_at', time() );
+			update_option( 'wc_stripe_test_oauth_failed_attempts', 0 );
+			update_option( 'wc_stripe_test_oauth_last_failed_at', '' );
+
+			WC_Stripe_API::set_secret_key( $test_secret_key );
+			try {
+				WC_Stripe::get_instance()->account->configure_webhooks( 'test' );
+			} catch ( Exception $e ) {
```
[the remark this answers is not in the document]
malithsen: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5610/commits/0e6a5f0b36172550846a95c92d6c2ba2e2a3028e

tests/phpunit/class-wc-stripe-connect-test.php
```diff
@@ -206,6 +206,126 @@ public function test_save_stripe_keys_decommissions_previous_webhook_before_savi
...
+
+		// The live-webhook failure is still surfaced to the caller.
+		$this->assertInstanceOf( WP_Error::class, $return );
+
+		// ...but the dual-fetched test keys were persisted regardless of it.
+		$settings = WC_Stripe_Helper::get_stripe_settings();
+		$this->assertSame( 'sk_test_123', $settings['test_secret_key'] );
+		$this->assertSame( 'pk_test_123', $settings['test_publishable_key'] );
+		$this->assertSame( 'connect', $settings['test_connection_type'] );
+	}
```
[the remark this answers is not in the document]
malithsen: Added a test in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5610/commits/cc81b6110eadeebdda0c7758d7d5c0a1c6c2780b

tests/phpunit/class-wc-stripe-connect-test.php
```diff
@@ -281,6 +359,97 @@ public function provide_refresh_connection_terminal_handling(): array {
...
+
+		// Live environment is fully saved despite the test-webhook failure.
+		$this->assertSame( 'sk_live_123', $settings['secret_key'] );
+		$this->assertSame( 'no', $settings['testmode'] );
+
+		// The dual-fetched test keys are persisted; the webhook failure happens after the
+		// settings write, so it does not roll them back.
+		$this->assertSame( 'sk_test_123', $settings['test_secret_key'] );
+		$this->assertSame( 'pk_test_123', $settings['test_publishable_key'] );
+		$this->assertSame( 'connect', $settings['test_connection_type'] );
```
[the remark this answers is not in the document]
malithsen: Added in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5610/commits/b85c192205995688ba0370c9be004b8d49701085

pierorocca:
Per our DM:

\[Piero\]

With the live and test account status tabs in a modal, it’s hard to see that both accounts were connected right? That info is hidden until a user clicks in. Now that both get enabled in one shot, there’s no way to tell both were enabled in one shot without clicking back into each tab?

\[Malith\]

The only visual indicator is that the "Enable test mode" checkbox will be enabled. Previously, it was disabled until merchants completed the test connect flow. I think we'd eventually want to get rid of the tabs altogether. Nevena's [designs](<https://www.figma.com/design/0fiyiUe18lGU6kGyQNmXxE/Stripe-Payments?node-id=3656-17767&p=f&t=Iz7Xsq89MNOJzGta-0>) are moving in that direction.

---

Adding that the merchant won't know which test account was added either based on a checkbox affordance. If we could move in the direction @nevena.ilic1013 conceptually like below so that important metadata and account status info is visible at a glance.

diegocurbelo (approved):
The code changes look good, and it tests as described 👍🏼 

I agree on shipping this as a first step towards a single onboarding flow for both live and test environments, I just found two details (that we can also tackle in a follow up PR):

**1. Test-mode caches are never invalidated after the dual-fetch write**
We are currently calling `clear_caches_after_key_update()` for live mode, so it clears only the live-prefixed entries; we should either call it for test mode, or make changes to enable/disable test mode to clear them; otherwise reconnecting a live account _might_ change the test account (depending on what we decide with 2. below) and keep the caches (e.g: ~2h for the account data)... not a blocker as this is somewhat of an edge case, but the fix is simple enough and very low risk


**2. A live connect now silently replaces pre-existing test credentials**
I think that for this case, we could save tests keys only when:
- there are no previous keys, or
- the previous keys were not obtained via OAuth

PR #5611, opened 2026-06-19
Guard Optimized Checkout against undercharging tax applied after the order total is summed

### LINKED ISSUE #5582: Optimized Checkout doesn't capture tax amount when tax is calculated by TaxJar
### Description

Disclaimer: this issue was identified due to Stripe amounts not reconciling against WooCommerce numbers. Claude AI took a look on our codebase and identified a potential fix for the issue, which appears to work. A human is writing this report, but I am including the AI's reasoning around the issue, as I have not thoroughly validated its claims. But, it all appears to make sense.

When the Optimized Checkout option is enabled, and TaxJar is being used to handle tax calculation, the Stripe PaymentIntent is created with an amount equal to the order total **without tax**. The order data in WC has the correct tax amount and is also sent along with the API call as `metadata.tax_amount`. But, the captured amount omits it. As a result, sales tax was not charged on the order.

**Toggling Optimized Checkout to `off`, the issue no longer persists.**

AI believes that due to updating, the following codepath triggered, making Optimized Checkout enabled instead of keeping our existing option (`no`). AI's reasoning:

<details><summary>AI's reading of the cause: </summary>
- Updating the gateway 10.7.0 → 10.8.1 ran a one-time migration (logged under the tag `[OCS+AP 10.8]`) that auto-enabled Optimized Checkout + Adaptive Pricing for our account, writing `optimized_checkout_element=yes` into the `woocommerce_stripe_settings` option. Our prior setting was `no`.
- We later rolled the plugin back to 10.7.0, but because the migration had written the value to the **database option**, the file rollback did not revert it — Optimized Checkout stayed enabled and the undercharging continued. This is why a rollback alone did not resolve it.

There appear to be **two** separable problems:
1. The 10.8.x migration force-enabled Optimized Checkout (overriding our explicit `no`), and the change survived a version rollback because it lives in a DB option.
2. With Optimized Checkout on + an external tax plugin (TaxJar), the captured amount excludes tax.

in the Optimized Checkout / checkout-session path, the PaymentIntent `amount` is derived from `$order->get_total()` (e.g. `WC_Stripe_Intent_Controller::update_intent()` sets `$amount = $order->get_total()`) **before** the order total has been recalculated to include TaxJar's tax. TaxJar only folds tax into the order grand total inside `$order->calculate_totals()` (hook `woocommerce_order_after_calculate_totals` → its `Order_Tax_Applicator`, which sets the total *including* tax). So at the moment the amount is locked, `get_total()` is still pre-tax while `get_total_tax()` is already correct — which is why `metadata.tax_amount` is right but the charged `amount` is not. The pre-10.8 (non-OC) create-and-confirm path reads `get_total()` after totals are finalized, so it was unaffected.

</details>

### Steps to reproduce

Prerequisites: 

* WooCommerce Stripe Gateway
  * Version 10.7.0
  * Optimized Checkout is turned off
* TaxJar (or another tax plugin that calculates order tax on `woocommerce_order_after_calculate_totals`)

1. Confirm a taxable order is charged the full tax-inclusive total in Stripe.
2. Update the Stripe plugin to 10.8.1, where the [`[OCS+AP 10.8]`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/a9aa74708541338f0cfabd5887a2b02cf512b958/includes/migrations/class-wc-stripe-ocs-ap-default-on-update.php#L45) migration runs. Optimized Checkout becomes enabled automatically.
3. Place an order shipping to a taxable jurisdiction so WooCommerce/TaxJar applies sales tax.
4. Complete payment via Stripe (now Optimized Checkout)
5. In the Stripe Dashboard / API, inspect the PaymentIntent: the captured `amount` / `amount_received` equals the order total **minus** tax, while `metadata.tax_amount` shows the correct (uncaptured) tax.

### Expected behavior

The amount captured by Stripe should equal the WooCommerce order total including tax, same as the non-Optimized checkout.

I also believe the database migration should not override our `optimized_checkout_element=no` option.

### Screenshots

After updating to 10.8.1, before disabling Optimized Checkout:

Image

Image

After disabling Optimized Checkout:

Image

Image

### System report

Important context for the following report:

* WooCommerce Stripe Gateway version reads 10.7.0, but this is after performing a rollback from 10.8.1
* A lot of information is redacted that I believe is not necessary for this report

<details><summary>System Status Report</summary>

```
### WordPress Environment ###

WC Version: 10.8.1
Legacy REST API Package Version: The Legacy REST API plugin is not installed on this site.
Log Directory Writable: ✔
WP Version: ❌ 6.9.4 - There is a newer version of WordPress available (7.0)
WP Multisite: –
WP Memory Limit: 512 MB
WP Cron: ✔
WP Environment Type: production
Language: en_US

Default Timezone is UTC: ✔
fsockopen/cURL: ✔
SoapClient: ✔
DOMDocument: ✔
GZip: ✔
Multibyte String: ✔
Remote Post: ✔
Remote Get: ✔

### Security ###

Secure connection (HTTPS): ✔
Hide errors from visitors: ✔

### Active Plugins () ###

TaxJar - Sales Tax Automation for WooCommerce: by TaxJar – 4.2.7
WooCommerce.com Update Manager: by Automattic – 1.0.3
WooCommerce Composite Products: by WooCommerce – 11.0.6
WooCommerce Stripe Gateway: by Stripe – 10.7.0
WooCommerce: by Automattic – 10.8.1

### Dropin Plugins () ###

db-error.php: db-error.php
db.php: db.php
maintenance.php: maintenance.php
php-error.php: php-error.php

### Must Use Plugins (1) ###

Protectli Must-Use Configs: by Ed Borgquist IV <ed.borgquist@protectli.com> – 1.1.0

### Settings ###

Legacy API Enabled: –
Force SSL: ✔
Currency: USD ($)
Currency Position: left
Thousand Separator: ,
Decimal Separator: .
Number of Decimals: 2

Taxonomies: Product Visibility: exclude-from-catalog (exclude-from-catalog)
exclude-from-search (exclude-from-search)
featured (featured)
outofstock (outofstock)
rated-1 (rated-1)
rated-2 (rated-2)
rated-3 (rated-3)
rated-4 (rated-4)
rated-5 (rated-5)

Connected to WooCommerce.com: ✔
Enforce Approved Product Download Directories: –
HPOS feature enabled: –
Order datastore: WC_Order_Data_Store_CPT
HPOS data sync enabled: –
Enabled Features: analytics
rate_limit_checkout
marketplace


### Logging ###

Enabled: ✔
Handler: Automattic\WooCommerce\Internal\Admin\Logging\LogHandlerFileV2

### WC Pages ###

### Theme ###

Name: Protectli
Version: 2.0.9
Author URL: https://protectli.com/
Child Theme: ✔
Parent Theme Name: Hello Elementor
Parent Theme Version: 3.4.9
Parent Theme Author URL: https://elementor.com/?utm_source=wp-themes&utm_campaign=author-uri&utm_medium=wp-dash
Theme type: Classic theme
WooCommerce Support: ✔

Version: 10.7.0
Account ID: acct_1K4D58E5s29Qbg5n
Test Mode Enabled: ✔
					No
OAuth Connected: ✔
					Yes
Sync Enabled: ✔
					Yes
Optimized Checkout Enabled: No
Enabled Payment Methods: card,link
Express Checkout: ✔
					Enabled (cart)
Auth and Capture: ✔
					Yes
Logging: ✔
					Yes

### Admin ###

Enabled Features: activity-panels
analytics
analytics-scheduled-import
product-block-editor
experimental-iapi-mini-cart
coupons
core-profiler
customize-store
customer-effort-score-tracks
import-products-task
experimental-fashion-sample-products
shipping-smart-defaults
shipping-setting-tour
homescreen
marketing
mobile-app-banner
onboarding
onboarding-tasks
pattern-toolkit-full-composability
product-custom-fields
remote-inbox-notifications
remote-free-extensions
payment-gateway-suggestions
printful
shipping-label-banner
subscriptions
store-alerts
transient-notices
woo-mobile-welcome
wc-pay-promotion
wc-pay-welcome-page
launch-your-store

Disabled Features: product-data-views
experimental-blocks
experimental-iapi-runtime
coming-soon-newsletter-template
minified-js
product-pre-publish-modal
products-catalog-api
settings
async-product-editor-category-field
product-editor-template-system
use-wp-horizon
rest-api-v4

Daily Cron: ✔ Next scheduled: 2026-06-16 19:56:00 -07:00
Options: ✔
Notes: 6
Onboarding: completed

### Status report information ###

Generated at: 2026-06-15 14:36:34 -07:00
```
</details>

### Additional context

The problem started happening immediately and consistently after updating from 10.7.0 to 10.8.1

**Example order:** order total $745.79 (incl. $43.29 tax). 
- PaymentIntent request/response:
  - `amount`: 70250 and `amount_received`: 70250. This is the total **without** tax
  - `metadata.tax_amount`: 4329 is correct, but not captured in `amount`
  - `capture_method`: automatic_async (before the update, this was "automatic")
  - 70250 + 4329 = 74579 = the correct total

Mayisha:
Fixes woocommerce/woocommerce-gateway-stripe#5582  <br>Fixes woocommerce/woocommerce-gateway-stripe#5582

On Optimized Checkout, taxable orders could be charged the pre-tax total: the PaymentIntent `amount` excluded sales tax while `tax_amount` metadata reported it, so the captured amount was short by exactly the tax.

**Note on reproduction:** I was **not able to reproduce the undercharge locally** (TaxJar 4.2.7, OCS on, Blocks + local pickup, CA tax). In my environment the order total was always reconciled correctly before the charge was read, so the charge matched the order. The reporter's production data, however, shows the exact persisted state this guards against — correct tax line items (`get_total_tax()`) alongside a stale, pre-tax `get_total()`. **This PR is therefore a guard against that known-possible code path rather than a fix verified against a live local repro.** The mechanism and the guard are covered by unit tests that fail against the unpatched code.

### Root cause

Timing, not the wrong field. Tax extensions like TaxJar apply tax on `woocommerce_order_after_calculate_totals` and store it on the order; if the grand total was last summed before that, `get_total()` stays pre-tax while the tax line items are correct. The amount-reading paths then charge the stale total. Three distinct paths read it, split by whether the payment method supports deferred intents:

* **Deferred methods** (card → OCS card, express): charged via `process_payment_with_deferred_intent`. The intent-controller paths are gated off for them.
* **Non-deferred methods** (Blik, ACSS): amount locked up-front in `WC_Stripe_Intent_Controller::create_payment_intent()` / `update_intent()`.

### The fix

A shared `WC_Stripe_Helper::recalculate_order_total()` calls `calculate_totals( false )` immediately before the amount is read, at each path:

* `process_payment_with_deferred_intent()` for the deferred (OCS card / express / saved-card) charges. Scoped to this dispatcher so the subscription change-payment flow — which reaches `prepare_payment_information_from_request()` with a `WC_Subscription` — is untouched. Guarded by `is_payment_needed()` so a deliberately zero total (free orders, free trials, pre-orders not charged upfront) is never resurrected by re-summing line items.
* `create_payment_intent()` **/** `update_intent()` for the non-deferred methods, whose amount never flows through the dispatcher.

`false` re-sums the tax already stored on the order rather than re-deriving it from store tax rates (which would clobber tax a third-party extension applied via its API). `calculate_totals()` saves the order, so the persisted total and the charge stay consistent. Idempotent for already-correct orders, so WooCommerce-core-tax stores are a no-op and never double-charged.

### How can this code break? What prevents it?

* Risk: recalculation changes a total deliberately set to something not derivable from line items (free orders, gift cards, manual adjustments). Mitigation: `calculate_totals( false )` doesn't re-derive taxes from store rates; the `is_payment_needed()` guard skips zero-total orders; these sites run on fresh checkout orders, not admin-edited ones. Known limitation: a partial reduction stored outside line items (e.g. some gift-card plugins) would be re-summed — accepted, same trade-off as recomputing the total.
* Prevention: a regression test reproduces TaxJar's persisted state (correct tax line items + stale pre-tax total, no active hook) and asserts the deferred-path charge includes the tax; it fails against the unpatched code on both the deferred and intent-controller paths.

## Testing instructions

* Add a tax plugin (like TaxJar) and configure taxes.
* Enable Optimized Checkout; place an OCS card order for a shippable product.
* In the Stripe dashboard, confirm `amount` equals the tax-inclusive order total and matches `metadata.tax_amount`.
* Regression test: with WooCommerce-core tax (no snippet), OCS on and off, confirm the charged amount is unchanged and tax isn't double-counted. Place a free (100%-off) order and confirm it still completes without a charge. **The main test for this PR is to ensure it doesn't cause any regressions.**

### Changelog entry

#### Comment

</details>

**Post merge**


includes/class-wc-stripe-helper.php
```diff
@@ -102,6 +102,21 @@ public static function get_stripe_amount( $total, $currency = '' ) {
...
+	 *
+	 * Folds in tax an extension applied on `woocommerce_order_after_calculate_totals` after the
+	 * total was last summed (otherwise `get_total()` stays pre-tax and the charge undercharges).
+	 * `false` re-sums stored tax rather than re-deriving from store rates, which would clobber an
+	 * extension's API-applied tax; safe to repeat.
+	 *
+	 * @param WC_Order $order The order being paid.
+	 * @return float The recalculated grand total, including taxes.
+	 */
+	public static function recalculate_order_total( $order ) {
```
wjrosa: We should add the return type if it is safe.  And specific unit tests to cover this new method

wjrosa (approved):
Looks good to me 👍 . Left a minor suggestion

includes/class-wc-stripe-intent-controller.php
```diff
@@ -514,8 +513,9 @@ public function update_intent( $intent_id = '', $order_id = null, $save_payment_
 
 			$is_setup_intent = substr( $intent_id, 0, 4 ) === 'seti';
 			if ( ! $is_setup_intent ) {
-				// These parameters are only supported for payment intents.
-				$request['amount']   = WC_Stripe_Helper::get_stripe_amount( $amount, strtolower( $currency ) );
+				// These parameters are only supported for payment intents. Setup
+				// intents carry no amount, so the recalculation is scoped here.
+				$request['amount']   = WC_Stripe_Helper::get_stripe_amount( WC_Stripe_Helper::recalculate_order_total( $order ), strtolower( $currency ) );
```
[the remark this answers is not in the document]
Mayisha: Good catch on the retry window. I went a different route than relaxing the update validation or making the save idempotent.

The key realization: `update_intent()` always overwrites the intent amount with the order total a few lines after this validation, so the amount comparison here isn't a charge-correctness guard — it's a consistency tripwire that runs *before* the very step that reconciles the amount. In the late-tax flow, an order total that's moved ahead of the intent is the expected pre-reconciliation state, not an anomaly. Enforcing equality there is what wedges the retry (order tax-inclusive, intent still pre-tax → throw before the amount can be corrected).

Fix: `validate_intent_for_order()` now takes an optional `$skip_amount_check`, and only `update_intent()` passes `true`. Currency, payment-type, and intent-ownership checks stay enforced everywhere, and the charge/confirm path (`process_payment_with_deferred_intent`) keeps the strict amount check — so nothing is weakened where the payment is actually captured. This also leaves the order's persisted total correct (tax-inclusive) instead of rolling it back to a stale value.

Added a regression test (`test_update_intent_retry_succeeds_when_order_total_ahead_of_intent`): first Stripe update fails, the retry revalidates cleanly and re-pushes the tax-inclusive amount. It fails against the unpatched code.

Mayisha:
Thanks @wjrosa for your previous review. I have made some changes in this commit [dec3e67](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5611/commits/dec3e67fceadab75abbe24a6c48b4673caa99f1c) to address [this coderabbitai feedback](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5611#discussion_r3460004767)

May I get another 👀 from you?

wjrosa:
> May I get another 👀 from you?

Sure! On it 👀 . Meanwhile, PHPUnit tests are all failing. Can you double-check if that's because of this PR?

includes/class-wc-stripe-order-helper.php
```diff
@@ -1152,10 +1158,12 @@ public function validate_intent_for_order( WC_Order $order, $intent, ?string $se
 		$order_amount          = WC_Stripe_Helper::get_stripe_amount( $order->get_total(), $order->get_currency() );
 		$order_intent_id       = self::get_intent_id_from_order( $order );
 
+		$amount_matches = $skip_amount_check || $order_amount === $intent->amount;
+
 		if ( 'payment_intent' === $intent->object ) {
 			$is_valid = $order_currency === $intent->currency
 				&& $is_valid_payment_type
-				&& $order_amount === $intent->amount
+				&& $amount_matches
```
wjrosa: On setup intents this dereferences `$intent->amount`, which doesn't exist on a setup-intent object, so it raises an "Undefined property" warning on every setup-intent confirmation. In tests `convertWarningsToExceptions="true"` escalates that to an exception, which `process_upe_redirect_payment()` catches and turns into `wp_safe_redirect` → "Cannot modify header information". That's the 3 red tests (`test_checkout_without_payment_uses_setup_intents`, the SEPA save, `test_pre_order_without_payment_uses_setup_intents`) failing across every PHPUnit matrix on CI right now.

The value is unused on the setup-intent branch, so just scope the read inside the `payment_intent` branch:

```suggestion
		if ( 'payment_intent' === $intent->object ) {
			// Only payment intents carry an amount; reading $intent->amount on a setup intent warns.
			$amount_matches = $skip_amount_check || $order_amount === $intent->amount;

			$is_valid = $order_currency === $intent->currency
				&& $is_valid_payment_type
				&& $amount_matches
```

Mayisha: Good catch. This was also the reason of the test failures. Fixed in [6c95a14](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5611/commits/6c95a142df000936a0acf26b2505b4542382269b)

wjrosa:
Something is off. Here's what I see in wp-admin:
Screenshot 2026-07-02 at 11 15 37

Here's what I see in the Stripe dashboard:
Screenshot 2026-07-02 at 11 16 11

Mayisha:
@wjrosa could you please share your store settings and shopper details? 
- Store base address
- Tax settings (tax percent, tax based on which address etc.)
- Shopper's billing/shipping country at the time of testing.

Mayisha:
I tested with different variation of the settings but could not reproduce this issue solely for OCS. This PR added some safeguards and forced recalculations in the OCS path. However, upon further investigation, I came to the conclusion that this is not an OCS issue. The issue lies in the adaptive pricing path and not in the OCS path. 

Closing this PR.

PR #5614, opened 2026-06-19
Use initCheckoutElementsSdk for Adaptive Pricing

Mayisha:
Fixes STRIPE-1209

PR #5568 guarded Adaptive Pricing on `initCheckout` being present **and** `initCheckoutElementsSdk` being **absent**. That check doesn't hold on current Stripe.js: `initCheckoutElementsSdk` is actually present, so the condition evaluated false and Adaptive Pricing was wrongly hidden.

This PR moves AP onto the method that's actually available:

- **Classic checkout:** gate the AP path on `initCheckoutElementsSdk` being present and call `stripe.initCheckoutElementsSdk()` instead of `stripe.initCheckout()`.
- **Blocks checkout:** drop the explicit guard entirely. The element is rendered by the `@stripe/react-stripe-js` npm package, which uses whatever method that version provides, so no manual `initCheckout` check is needed — AP support is
driven solely by whether it's enabled server-side.
- Update the classic and Blocks tests to match.

## Testing instructions
- Enable Adaptive Pricing
- Adaptive pricing should work in classic checkout
- Adaptive pricing should work in block checkout

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where Adaptive Pricing options were incorrectly hidden during checkout. Payment processing has been enhanced to properly detect and load available pricing options, ensuring customers see all pricing choices when completing their purchase and can take advantage of adaptive pricing features.


client/classic/upe/payment-processing.js
```diff
@@ -316,8 +316,7 @@ async function createStripePaymentElement( api, paymentMethodType ) {
 	if (
 		stripeServerData?.isAdaptivePricingEnabled &&
 		supportsDeferredIntent &&
-		typeof stripe?.initCheckout === 'function' &&
-		typeof stripe?.initCheckoutElementsSdk !== 'function'
+		typeof stripe?.initCheckoutElementsSdk === 'function'
```
malithsen, line 319: Do we need this `initCheckoutElementsSdk` at all? It is truthy, and AP is showing up as expected, but it doesn't look like we need it. The negation of this check was only introduced in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5568/. Also, we don't have it in the blocks checkout path.

client/classic/upe/payment-processing.js
```diff
@@ -336,7 +335,7 @@ async function createStripePaymentElement( api, paymentMethodType ) {
 			gatewayUPEComponents[ paymentMethodType ].checkoutSessionId =
 				sessionId;
 
-			elements = await stripe.initCheckout( {
+			elements = await stripe.initCheckoutElementsSdk( {
```
malithsen, line 338: What's the reason for switching to `initCheckoutElementsSdk` instead of `initCheckout`?

malithsen (commented):
This is testing as expected and AP is working. 

> Classic checkout: gate the AP path on initCheckoutElementsSdk being present and call stripe.initCheckoutElementsSdk() instead of stripe.initCheckout().

Can you comment on why switching to `initCheckoutElementsSdk` is needed for classic checkout?

client/classic/upe/payment-processing.js
Mayisha: In block checkout the element is rendered via @stripe/react-stripe-js npm package. But in classic we are directly calling `initCheckoutElementsSdk`. I am thinking of a scenario that if any site loads an older version of stripe.js build manually or via other plugins, this method could be absent. This is just a guard to prevent that.

client/classic/upe/payment-processing.js
Mayisha: Two reasons-
1. Stripe documentation do not have the `initCheckout` method anymore. The official docs only have [initCheckoutElementsSdk](https://docs.stripe.com/js/custom_checkout/init).
2. Stripe.js latest versions (like `dahlia`) throw an error for `initCheckout`. Using `initCheckoutElementsSdk` loads AP even if the site loads both dahlia and clover.

malithsen: hm, `initCheckoutElementsSdk` works on both `clover` and `dahlia` but are we sure that `dahlia` won't break some other API calls we make?

We officially support `clover` only for now. So I think it would be safer to use the documented `clover` API. `initCheckoutElementsSdk` is [officially](https://docs.stripe.com/changelog/dahlia/2026-03-25/rename-init-checkout-to-init-checkout-elements) a `dahlia` feature, though it's also available in `clover`. 

I think a long-term fix (a separate PR) would be to ensure we always use the version we are supposed to (`clover` in this case). 

Thoughts?

wjrosa: Maybe we could use the current JS version here somehow? We can use `stripe.VERSION` (returns an integer). But I am not sure how to map the integer version to the named version.

wjrosa (approved):
Looks good to me and works as expected! Malith's comment is relevant

Mayisha:
Closing in favor of https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5664.

PR #5621, opened 2026-06-22
Skip payment methods Stripe reports as unavailable at checkout

wjrosa:
Related to STRIPE-1233 (defensive hardening found while investigating; not a confirmed fix for that ticket — see note).

The plugin decides a payment method is enabled from the PMC's `display_preference.value` only, and never reads Stripe's per-method `available` flag. Stripe can report a method `display_preference.value: 'on'` **and** `available: false`; in that state the plugin still offers the method and adds it to the intent, but Stripe renders an empty Payment Element and checkout fails with *"We could not retrieve data from the specified Element…"*.

**Change:** add `WC_Stripe_Payment_Method_Configurations::is_payment_method_available()` and gate `is_enabled_at_checkout()` on it. A method counts as available unless the PMC explicitly reports `available: false`, so accounts without the flag are unaffected. Admin toggles are untouched; only checkout offering is gated. This is a defensive guard against offering a method Stripe won't render.

> **Status:** Investigation of the STRIPE-1233 merchant's account (NL, EUR, iDEAL capability Active, iDEAL ON in both configs) did **not** confirm `available: false` for that account, so this is **not** verified as the fix for that ticket — it's a general guard. Keeping as draft until either the raw API confirms an `available: false` case or we decide to land it as standalone hardening.

## Testing instructions

Needs an account whose PMC has a method with `display_preference.value: 'on'` but `available: false`.

1. Before: the method is offered at checkout → empty Payment Element, order placement errors.
2. After: the method is not offered (classic radio / OC / intent `payment_method_types`); other available methods unaffected.
3. A method with `available: true` (or no flag) still shows and completes.

Automated: `WC_Stripe_Payment_Method_Configurations_Test` (+5 cases) and `WC_Stripe_UPE_Payment_Method_Test` — 66 + 117 green; PHPCS clean.

---

### Changelog entry

Fix - Skip payment methods that Stripe reports as unavailable in the payment method configuration, fixing empty iDEAL checkout

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -33,6 +33,7 @@ xxxx-xx-xx - version 10.9.0
 * Dev - Rename internal Blocks API references to Store API in the Express Checkout client for naming consistency
 * Add - Verify that Stripe.js was served from the official Stripe origin before processing checkout payments
 * Fix - Exclude WooCommerce Subscriptions products from the Agentic Commerce product feed so they no longer make every sync report a partial success
+* Fix - Don't offer a payment method at checkout when Stripe reports it as unavailable in the payment method configuration, which caused an empty Payment Element and failed orders (e.g. iDEAL)
 
 2026-06-15 - version 10.8.2
 * Fix - Disable Adaptive Pricing when webhooks are disabled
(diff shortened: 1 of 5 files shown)


wjrosa:
Closing. The STRIPE-1233 investigation confirmed the affected account's payment method configuration has iDEAL **active and available** on the configuration checkout actually uses — so the `available: false` guard in this PR doesn't address that issue. The root cause is a front-end Payment Element submission problem, not a PMC/capability one.

The change itself (skip methods Stripe reports as unavailable at checkout) is a reasonable defensive guard, but I couldn't construct a real case where a method is `on` with `available: false` — Stripe ties availability to the enabled state — so I'm closing rather than merging speculative hardening. Easy to revive if we hit a confirmed `available: false` case.

PR #5626, opened 2026-06-23
Send async-method order emails by serializing deferred webhook settlement with the order lock

### LINKED ISSUE #5620: New Order (admin) & Customer Processing emails not sent for async methods (iDEAL/Klarna/Bancontact) completed via webhook
### Summary

For orders paid with **asynchronous redirect payment methods** (iDEAL, Klarna, Bancontact, EPS), WooCommerce's order-notification emails are intermittently **not sent at all**:

- the **admin "New order"** email is not sent, and
- the **customer "Processing order" / "Order received"** email is not sent.

Orders paid with **cards (synchronous)** are never affected — those always send both emails. Later status emails (e.g. "Completed" triggered when staff manually completes the order) are also sent normally. Only the notifications tied to the initial `pending → processing` transition that happens during asynchronous completion go missing.

The result is a paid order for which neither the shop nor the customer receives any confirmation, until staff notice it manually.

### Environment

- WooCommerce Stripe Gateway: 10.8.x
- WooCommerce: 10.8.x
- WordPress: 6.x
- PHP: 8.x
- HPOS (custom order tables): enabled
- Checkout: Cart/Checkout **Blocks** (Store API), UPE / optimized checkout enabled
- Affected methods: iDEAL, Klarna, Bancontact (redirect/webhook-confirmed). Card (sync) unaffected.

### Observed log pattern (sanitized)

For affected orders, the gateway log shows the payment being confirmed via the **deferred/asynchronous webhook path** while the customer's UPE redirect return runs concurrently, e.g.:

```
DEBUG Webhook received (payment_intent.requires_action)
DEBUG Processing payment_intent.succeeded (pi_XXX) asynchronously for order NNNN.
DEBUG Stripe PaymentIntent pi_XXX succeeded for order NNNN
DEBUG Webhook received (charge.succeeded)
DEBUG Webhook received (payment_intent.succeeded)
INFO  Begin processing UPE redirect payment for order NNNN for the amount of ...
DEBUG Processing charge response
INFO  Skip processing UPE redirect payment for order NNNN ..., order payment is already being processed (locked)
```

or, when the customer does not return to the site:

```
DEBUG Skipped processing deferred webhook for Stripe PaymentIntent pi_XXX for order NNNN - payment already complete.
```

The order ends in `processing` (or later `completed`) correctly — the **payment is captured and the order is marked paid** — but the `New order` / `Customer processing order` emails for that transition never fire.

### Analysis / suspected cause

It appears to be a race between the **deferred webhook processing** (scheduled via Action Scheduler in `defer_webhook_processing()` for orders awaiting redirect) and the **UPE redirect-return** handler, combined with `lock_order_payment()`. Depending on which path wins and acquires the lock, the order is moved to `processing` in a code path/context where WooCommerce's transactional email actions for that transition are not dispatched, so both the admin and customer notifications for the initial paid transition are skipped. Synchronous card payments don't hit this path and are unaffected.

### Steps to reproduce

1. Enable an asynchronous redirect method (iDEAL / Klarna / Bancontact) via UPE, HPOS + Blocks checkout.
2. Place an order with that method and authorize the payment at the bank/redirect page.
3. Observe the order is completed via the webhook / deferred-webhook path (especially if the customer does not return to the site).
4. Check the order: it is `processing`/paid, but no "New order" (admin) and no "Processing order" (customer) email was sent.

### Expected

The admin "New order" and customer "Processing order" emails should be sent exactly once when an asynchronous order is marked paid, the same as for card payments.

### Actual

Neither email is sent for asynchronous (iDEAL/Klarna/Bancontact) orders completed via the webhook/deferred path. Card orders are unaffected.

wjrosa:
Fixes woocommerce/woocommerce-gateway-stripe#5620

For asynchronous redirect payment methods (iDEAL, Klarna, Bancontact, EPS), the admin **New Order** and customer **Processing order** emails were intermittently not sent at all, even though the order was correctly marked paid. Synchronous card payments were never affected.

**Root cause:** the `payment_intent.succeeded` event for these methods is deferred to Action Scheduler (`process_webhook_payment_intent_success()` → `defer_webhook_processing()`), and runs concurrently with the customer's order-received redirect (`process_upe_redirect_payment()`). The redirect path and the synchronous webhook path both acquire the order payment lock before settling — but the **deferred** path called `payment_complete()` with **no lock**. So both paths could reach `payment_complete()` at once; whichever ran second found the order already paid and no-opped, so the initial `pending`/`on-hold → processing` transition — and the transactional emails tied to it — was never dispatched.

**Fix:** acquire the order lock in the deferred `payment_intent` path and re-queue while it's held, then release it once settlement completes. This mirrors the existing `checkout.session.*` deferred handling, which already locks-and-requeues for this exact redirect-handler race. The lock's 5-minute TTL bounds the retry, so a wedged holder can't loop forever.

* *How can this break?* If the lock were never released, the event would re-queue until the 5-minute TTL clears it. Settlement is wrapped in `try/finally` so the lock is always released, even on error.
* *What prevents that?* New unit coverage asserts the re-queue-while-locked path (no settlement, no unlock) and the happy path (settle exactly once, then unlock, no re-queue). Full webhook-handler suite stays green.

## Testing instructions

Automated:

* `npm run test:php -- --filter WC_Stripe_Webhook_Handler_Test` — green, including the two new cases (`test_deferred_payment_intent_requeues_while_order_locked`, `test_deferred_payment_intent_unlocks_after_processing`).
* `npm run lint:php`, `npm run phpstan` — clean.

Manual (HPOS + Blocks/Store API checkout, UPE enabled):

1. Enable iDEAL (or Klarna/Bancontact) via UPE.
2. Place an order with that method and authorize at the redirect page.
3. To force the webhook-only path, **do not return to the site** after authorizing (or close the tab) so the deferred webhook settles the order.
4. Confirm the order reaches `processing`/paid, **and** that both the admin New Order email and the customer Processing order email are received — exactly once.
5. Repeat returning to the site normally; confirm still exactly one of each email (no duplicates).
6. Place a card order and confirm no regression (both emails, as before).

The exact "no emails" symptom depends on concurrent webhook/redirect timing and is not faithfully reproducible in unit tests; the serialization gap that causes it is what the new tests cover.

---

- [X] Covered with tests (or have a good reason not to test in description ☝️)

### Changelog entry

Fix - Send the admin New Order and customer Processing emails when an asynchronous payment method (iDEAL, Klarna, Bancontact) is confirmed via the deferred webhook path

**Post merge**

- [ ] Added testing instructions to the Release Testing Instructions wiki page (or does not apply)
- [ ] Added the needs docs label (or does not apply)


malithsen (approved):
I am not able to replicate the issue but the change makes sense.

PR #5630, opened 2026-06-24
Surface agentic commerce product feed validation & preview to merchants

wjrosa:
Fixes STRIPE-1239

Today the agentic commerce product feed is generated, uploaded to the Stripe Files API, and deleted. The per-product validation errors collected during a sync are only written to the logs, so when a product doesn't show up for AI agents the merchant has no in-store way to find out why — they have to open the Stripe dashboard, locate the ImportSet, and read its error report.

This PR gives merchants feed visibility directly under **Settings → Payments → Stripe → Agentic Commerce**, on demand and without waiting for an upload:

- **`WC_Stripe_Agentic_Commerce_Feed_Preview`** walks the catalog using the exact same query args, mapper, and validator a real sync uses (`ProductWalker::from_integration()` parity, including the `woocommerce_product_feed_args` filter), streaming in bounded 100-product batches — no CSV is written and nothing is uploaded. It returns `included_count`, `excluded_count` (filter-excluded — a merchant choice), `invalid_count`, and a capped `validation_errors` list of `{ product_id, product_name, edit_link, errors[] }`, plus a `truncated` overflow count. Variations link to their parent product's edit screen. Products the mapper can't process (e.g. an orphaned variation) are surfaced as invalid instead of aborting the whole preview.
- **`GET /wc/v3/wc_stripe/agentic-commerce/preview`** exposes the preview. It's read-only and gated on the developer feature flag only (not the merchant toggle), so a merchant can inspect feed health before turning the feature on.
- A **Feed preview** panel in the settings UI renders the summary counts and a "products that need attention" table with edit links.
- **`wp stripe agentic-commerce preview`** rides along for support/debugging.

The validator's pre-formatted messages are surfaced per product (grouped with the product name + edit link) rather than split into a separate `field`/`message` shape — the field name is already embedded in each message, and grouping is what the UI needs to point the merchant at the right product.

### How can this code break? / What prevents it?

- Large catalogs: the walk is synchronous like the existing manual sync, and streams page-by-page (bounded memory) rather than buffering the catalog. The detail list is capped (default 50) with the remainder rolled into `truncated`.
- Counts drifting from real syncs: the preview deliberately reuses the same query-arg construction, mapper, and validator as the sync walker, with tests asserting the included/excluded/invalid split.

## Testing instructions

1. Enable the Agentic Commerce feature flag and open **Settings → Payments → Stripe → Agentic commerce**; toggle **Enable agentic commerce** on.
2. In the new **Feed preview** card, click **Preview feed**.
3. Confirm the summary shows Included / With errors / Excluded by filters counts.
4. Create a product with missing required data (e.g. no price) and re-run the preview — it should appear under "Products that need attention" with a plain-language reason and a link to edit it.
5. Optional CLI: `wp stripe agentic-commerce preview` prints the same summary and error list.

---

-   [x] Tested on mobile (or does not apply) — admin settings screen, N/A

### Changelog entry

Add - Agentic commerce: add a Feed preview in settings (plus a REST endpoint and WP-CLI command) that shows, without uploading, which products will be sent to AI agents and which are skipped for missing or invalid data.


includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
```diff
@@ -182,6 +192,39 @@ public function get_status() {
...
+	 * included vs. excluded, plus the products the feed validator would reject
+	 * and why.
+	 *
+	 * Read-only and gated only on the developer feature flag (not the merchant
+	 * toggle) so merchants can inspect feed health before enabling the feature.
+	 *
+	 * @since 10.9.0
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function get_preview() {
```
Mayisha, line 206: AI feedback:

get_preview() runs generate() with no total-product cap and no time guard, inside a single REST request triggered by the UI "Preview feed" button. The detail_limit only caps returned error detail, not the walk itself. The real sync is explicitly designed for 100k+ catalogs and runs batched in the background (ProductWalker + add_time_limit under a scheduled action). The preview re-walks that same catalog one-shot and synchronously — on a large store this will hit max_execution_time / a gateway 504, and the frontend button just spins until apiFetch rejects.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-cli.php
```diff
@@ -110,6 +110,64 @@ function ( WalkerProgress $progress ) {
...
+			$preview = ( new WC_Stripe_Agentic_Commerce_Feed_Preview() )->generate( $limit );
+		} catch ( Exception $e ) {
+			WP_CLI::error( 'Preview failed: ' . $e->getMessage() );
+			return; // Never reached — error() exits.
+		}
+
+		WP_CLI::success( 'Preview complete.' );
+		WP_CLI::log( sprintf( '  Products scanned: %d', $preview['total_count'] ) );
+		WP_CLI::log( sprintf( '  Included:         %d', $preview['included_count'] ) );
+		WP_CLI::log( sprintf( '  Excluded (filter):%d', $preview['excluded_count'] ) );
```
Mayisha:
```suggestion
		WP_CLI::log( sprintf( '  Excluded (filter): %d', $preview['excluded_count'] ) );
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-cli.php
```diff
@@ -110,6 +110,64 @@ function ( WalkerProgress $progress ) {
...
+		WP_CLI::log( 'Building feed preview...' );
+
+		try {
+			$preview = ( new WC_Stripe_Agentic_Commerce_Feed_Preview() )->generate( $limit );
+		} catch ( Exception $e ) {
+			WP_CLI::error( 'Preview failed: ' . $e->getMessage() );
+			return; // Never reached — error() exits.
+		}
+
+		WP_CLI::success( 'Preview complete.' );
```
Mayisha:
```suggestion
		WP_CLI::success( 'Preview generation complete.' );
```
Nit.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-cli.php
```diff
@@ -110,6 +110,64 @@ function ( WalkerProgress $progress ) {
...
+	 */
+	public function preview( $args, $assoc_args ): void {
+		$limit = (int) ( $assoc_args['limit'] ?? WC_Stripe_Agentic_Commerce_Feed_Preview::DEFAULT_DETAIL_LIMIT );
+
+		WP_CLI::log( 'Building feed preview...' );
+
+		try {
+			$preview = ( new WC_Stripe_Agentic_Commerce_Feed_Preview() )->generate( $limit );
+		} catch ( Exception $e ) {
+			WP_CLI::error( 'Preview failed: ' . $e->getMessage() );
```
Mayisha:
```suggestion
			WP_CLI::error( 'Preview generation failed: ' . $e->getMessage() );
```
Nit.

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -0,0 +1,293 @@
...
+									) }
+								</span>
+							</SummaryStat>
+							<SummaryStat>
+								<span className="wc-stripe-agentic-preview__stat-value">
+									{ excludedCount.toLocaleString() }
+								</span>
+								<span className="wc-stripe-agentic-preview__stat-label">
+									{ __(
+										'Excluded by filters',
```
Mayisha: When i ran this locally in my test store with subscription products and got 14 "excluded by filters", it was confusing for me. I never added a filter and the count was coming for my subscription products. 
I think this label creates a confusing UX for the user.

Mayisha: Image

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -0,0 +1,293 @@
...
+					{ __(
+						'Check which products will be made available to AI agents on the next sync, and which would be skipped because of missing or invalid data — without uploading anything.',
+						'woocommerce-gateway-stripe'
+					) }
+				</IntroDescription>
+
+				{ hasError && (
+					<Notice status="error" isDismissible={ false }>
+						{ __(
+							'Could not build the feed preview. Check the WooCommerce logs for details.',
```
Mayisha: Nit. We can link the logs page here.

readme.txt
```diff
@@ -156,6 +156,7 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 == Changelog ==
 
 = 10.9.0 - xxxx-xx-xx =
+* Add - Agentic commerce: add a Feed preview in settings (plus a REST endpoint and WP-CLI command) that shows, without uploading, which products will be sent to AI agents and which are skipped for missing or invalid data
```
Mayisha: Nit: This could be shorter. One example is-
```suggestion
* Add - Add feed preview feature in settings and via WP-CLI command for Agentic commerce.
```

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -0,0 +1,293 @@
...
+		<SectionCard>
+			<CardHeader>
+				<SectionHeading>
+					{ __( 'Feed preview', 'woocommerce-gateway-stripe' ) }
+				</SectionHeading>
+			</CardHeader>
+			<CardBody>
+				<IntroDescription>
+					{ __(
+						'Check which products will be made available to AI agents on the next sync, and which would be skipped because of missing or invalid data — without uploading anything.',
```
Mayisha: This looks very much AI-generated, probably because of the em-dash 😅

Mayisha (commented):
The feed preview is working well. I have a few minor comments and an AI flagged comment.

Image

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
wjrosa: Good catch. The preview now caps the walk at `DEFAULT_SCAN_LIMIT` (2,000 products) and `break 2`s out of the pagination loop once it's hit, so the synchronous request stays bounded no matter the catalog size — no more open-ended walk racing `max_execution_time`. When the cap trips, the response carries `scan_limited: true` and the UI shows a notice that the preview only covers the first N products while the real (background, batched) sync still does the whole catalog. The merchant gets a fast, honest sample instead of a spinner that ends in a 504. Done in 9538dbc6f.

client/settings/agentic-commerce/feed-preview.js
wjrosa: Linked. The error notice now points "WooCommerce logs" at `admin.php?page=wc-status&tab=logs` (same target as the View logs button on the sync-status card). Done in 9538dbc6f.

client/settings/agentic-commerce/feed-preview.js
wjrosa: That's fair 😅 . Dropped it in 9538dbc6f

client/settings/agentic-commerce/feed-preview.js
wjrosa: Agreed — "by filters" was misleading because almost none of that count is a filter you set. Those 14 are subscription products: their variations share the `product_variation` post type, so the feed query walks them, and `should_sync_product()` drops them by default since agents can't do recurring billing. The label made an automatic exclusion look like merchant config.

Two-part fix:
- Here on this PR I renamed the stat to "Excluded" and moved the explanation into a tooltip (subscriptions + visibility settings, and that these aren't validation errors).
- The breakdown you'd actually want — splitting the count into "12 subscription products" vs "2 hidden by visibility settings" — is in a stacked follow-up, #5635. It adds an `excluded_breakdown` to the preview response and renders a per-reason line under the stat, so 14 is never just an opaque number again.

readme.txt
```diff
@@ -198,5 +199,6 @@ If you get stuck, you can ask for help in the [Plugin Forum](https://wordpress.o
 * Add - Add an Agentic Commerce checkout mode that redirects shoppers to your store to check out instead of completing the purchase in the AI agent
 * Tweak - When enabling manual capture, clarify to agentic commerce merchants that Agentic Commerce purchases follow the capture setting in the Stripe agentic commerce dashboard
 * Dev - Centralize agentic commerce feed scheduling in integration class
+* Tweak - Cap the Agentic Commerce feed preview scan so it stays responsive on large catalogs instead of timing out
```
Mayisha, line 202: Is this new changelog entry intentional? The feed preview feature itself is added in the same PR, so this separate entry is not quite needed.

Mayisha (approved):
Thanks for the updates @wjrosa. Changes look good to me 👍 

Left a question about the double changelog entries.

readme.txt
wjrosa: Ooops, not really. Fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5630/commits/381d790683172018673a9aa762b2017ce05e0b2a

PR #5475, opened 2026-05-25
Map runtime CoreJS 3 polyfills to native globals

diegocurbelo:
Fixes STRIPE-1079

## Changes proposed in this Pull Request:

Externalize `core-js-pure` polyfill modules (`URL`, `URLSearchParams`, `Promise`, `Symbol`, `Map`, `Set`) to native browser globals via webpack's `externals` config, and alias
  `@babel/runtime-corejs3/helpers` → `@babel/runtime/helpers` (identical helpers, no bundled polyfills).

  All polyfilled features have native support in every browser within our
  [`@wordpress/browserslist-config`](https://github.com/WordPress/gutenberg/blob/trunk/packages/browserslist-config/index.js) target. `wp-polyfill` (already enqueued via `injectPolyfill:
  true`) backstops anything older that WP itself still supports.

  **Storefront bundle savings:**

  | Bundle | Before | After | Saved |
  |---|---|---|---|
  | `express-checkout.js` | 129 KB | 93 KB | −28% |
  | `upe-classic.js` | 160 KB | 113 KB | −29% |
  | `upe-blocks.js` | 307 KB | 270 KB | −12% |

## Testing instructions

1. Build assets: `npm run build:webpack`. Verify bundle sizes shrink as above.
2. On the latest two versions of Chrome / Firefox / Safari / Edge, complete a purchase via:
     - Classic checkout
     - Cart Block + Checkout Block
     - Apple Pay or Google Pay button on the product page
3. Confirm no `URL is not defined` / `Promise is not defined` / similar console errors.

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5631, opened 2026-06-24
Make classic checkout resilient to deferred render-blocking JS (SiteGround Speed Optimizer)

### LINKED ISSUE #5615: Classic checkout: Stripe credit card fields and payment options fail to render when SiteGround Speed Optimizer's "Defer Render-Blocking JS" is enabled (Regression in 10.8.0+)
### Description

Since updating to **WooCommerce Stripe Gateway 10.8.0**, the Stripe debit/credit card fields and payment options fail to render on the **Classic Checkout** (shortcode) page when **Apple Pay / Google Pay** and the **SiteGround Speed Optimizer** **"Defer Render-Blocking JS"** setting are both enabled.

The issue occurs regardless of whether **Optimized Checkout Suite (OCS)** is enabled or disabled.

Rolling back to **10.7.0** immediately resolves the issue under identical caching and optimisation settings.

| WooCommerce Stripe Gateway Version | OCS Enabled | OCS Disabled |
| --- | --- | --- |
| **10.7.0** | ✅ Credit card fields/Payment options render correctly | ✅ Credit card fields render correctly |
| **10.8.0** | ❌ Credit card fields/Payment options fail to render | ❌ Credit card fields fail to render |
| **10.8.1** | ❌ Credit card fields/Payment options fail to render | ❌ Credit card fields fail to render |
| **10.8.2** | ❌ Credit card fields/Payment options fail to render | ❌ Credit card fields fail to render |

**NB**  Stripe is configured to offer card payments and Apple Pay/Google Pay only.

### Actual Behaviour

Stripe credit card payment fields/payment options fail to render and initialise when **SiteGround Speed Optimizer**'s **Defer Render-Blocking JS** setting is enabled.

Merchants are unable to identify and exclude the relevant script through SiteGround's **Exclude from Deferral of Render-blocking JS** dropdown menu. As a result, card payment fields/payment options fail to render on Classic Checkout pages unless the **Defer Render-Blocking JS** feature is disabled.

From a merchant and customer perspective, this creates a critical checkout failure:

- Customers cannot enter payment details.
- Completing checkout with a credit card becomes impossible.
- Users experience confusion, resulting in rage clicks.
- Cart abandonment rates increase significantly.
- Merchant trust and conversion rates are negatively impacted.

### Expected behaviour

Stripe credit card payment fields/payment options should render and initialise correctly when common JavaScript optimisation settings, including the deferral of render-blocking JavaScript, are enabled.

If specific Stripe assets must be excluded from deferral, the plugin should expose identifiable script handles that can be easily targeted by hosting optimisation tools such as **SiteGround Speed Optimizer**.

#### Environment Setup

1. Install and activate **WooCommerce**, **WooCommerce Stripe Gateway (10.8.0+)**, and **SiteGround Speed Optimizer**.
2. Deactivate all other plugins to ensure a clean testing environment.
3. Configure the store to use the **Classic Checkout** page (using the shortcode `[woocommerce_checkout]`).
4. Set up **WooCommerce Stripe Gateway** to offer **card payments and Apple Pay/Google Pay only**.
5. Configure the **SiteGround Speed Optimizer** plugin with the following settings:

   - **Speed Optimizer → Frontend → JavaScript**: Set **Minify JavaScript Files** to `On`
   - **Speed Optimizer → Frontend → JavaScript**: Set **Combine JavaScript Files** to `Off`.
   - **Speed Optimizer → Frontend → JavaScript**: Set **Defer Render-Blocking JavaScript** to `On`.
   - **Speed Optimizer → Caching → Exclude URLs from Caching**: Exclude the `/checkout/` URL from caching.

#### Scenario A: 10.8.0+ (Optimized Checkout Suite Enabled)

6. Enable **Optimized Checkout Suite** in the **WooCommerce Stripe Gateway** advanced settings.
7. Open a clean incognito or private browser window as a logged-out guest.
8. Add a product to the cart and proceed to the checkout page.
9. Observe that the Stripe credit card fields and payment options fail to render.

#### Scenario B: 10.8.0+ (Optimized Checkout Suite Disabled)

10. Disable **Optimized Checkout Suite**.
11. Refresh the incognito checkout session.
12. Observe that the credit card fields and payment options still fail to render.

#### Scenario C: Verifying the Regression (Downgrade to 10.7.0)

13. Downgrade **WooCommerce Stripe Gateway** to **10.7.0** (leave all **SiteGround Speed Optimizer** settings unchanged).
14. Ensure **Optimized Checkout Suite** is disabled.
15. Repeat the guest checkout process in an incognito window.
16. Observe that the credit card fields render correctly and the purchase can be completed.
17. Enable **Optimized Checkout Suite**.
18. Repeat the guest checkout process in an incognito window.
19. Observe that the credit card fields still render correctly, confirming the regression started in **10.8.0**.

#### Additional Findings

- In **SiteGround Speed Optimizer → Frontend → JavaScript → Exclude from Deferral of Render-blocking JS**, the WooCommerce Stripe Gateway script handles are not available in the exclusion dropdown.
- This occurs with **Optimized Checkout Suite** both enabled and disabled.
- Searching the automatically detected handles does not return expected Stripe-related entries.
- As a result, the affected **WooCommerce Stripe Gateway** scripts cannot be excluded from JavaScript deferral through the **SiteGround Speed Optimizer** UI.

### Environment

- **WordPress Version:** `7.0`
- **WooCommerce Version:** `10.8.1`
- **Theme:** Storefront
- **Checkout Type:** Classic Checkout (shortcode)
- **WooCommerce Stripe Gateway Versions:** 10.7.0, 10.8.0, 10.8.1, 10.8.2
- **Stripe Payment Methods:** Credit card / debit card and Apple Pay / Google Pay only
- **Active Plugins (for isolation):**

  - **WooCommerce**
  - **WooCommerce Stripe Gateway**
  - **SiteGround Speed Optimizer**

wjrosa:
Fixes woocommerce/woocommerce-gateway-stripe#5615

When a host optimizer that defers **all** render-blocking JavaScript is enabled (reported with **SiteGround Speed Optimizer → "Defer Render-Blocking JS"**), the classic-checkout Stripe card fields fail to render. The merchant's console showed the tell:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'wcSettings')  // window.wc undefined
Uncaught TypeError: Cannot read properties of undefined (reading 'use')          // wp.data undefined
```

**Root cause.** Our classic bundle's module graph statically imports `@woocommerce/settings`, `@wordpress/data`, and `@wordpress/i18n`, which webpack externalizes to `window.wc.wcSettings` / `window.wp.data` / `window.wp.i18n`. WordPress normally guarantees the `wc-settings` / `wp-data` / `wp-i18n` scripts run first via our declared dependencies. The optimizer relocates and `defer`s our bundle **without preserving that order**, so our graph evaluates before those globals exist — the first externalized read throws at load and the whole script aborts before any card field initializes.

This regressed at **10.8.0**: the bundle-slimming change (woocommerce/woocommerce-gateway-stripe#5475, map CoreJS3 polyfills to native globals) shifted load timing enough to flip the optimizer's deferred execution order. It surfaces only with **Apple Pay/Google Pay enabled** (the express-checkout bundle joins the deferred set and tips the order), and **Pixel Manager for WooCommerce** had been coincidentally masking it by loading `wp-data`/`wc-settings`-dependent scripts early. (This is fundamentally the optimizer breaking WP's documented dependency ordering; the fix makes us resilient to it.)

**Fix.** Split the classic UPE entry into a tiny, dependency-free **bootstrap** (`client/classic/upe/index.js`) that imports nothing resolving a WP/WC external — so it can't throw at load. It waits for the required globals — the list is derived from the build's declared dependencies (e.g. `window.wp.data` / `window.wp.i18n` / `window.wc.wcSettings`) — **and** for the inline `wc_stripe_upe_params` settings object, in case the optimizer also reorders the inline settings script. Once ready, it executes the real init (`client/classic/upe/init.js`, the former entry body), which is bundled **eagerly into the same file** (`webpackMode: "eager"`): no separate chunk request, no `publicPath` handling, and an import rejection can only be an initialization failure, so it is logged rather than retried. A 10s fallback loads init anyway so a genuine misconfiguration fails the same way it would today.

### Verification status

The triggering condition is a third-party host optimizer that can't run in CI. The bootstrap gating is covered by Jest (including the reordered-inline-settings case), and reviewer testing verified classic + Blocks purchases with and without a JS-reordering simulation.

## Testing instructions

Regression (no optimizer) — must be unchanged:

1. Classic (shortcode) checkout: card fields render and a card payment completes.
2. Blocks checkout: card fields render and complete.
3. With Apple Pay/Google Pay enabled and disabled.

The actual fix (needs a SiteGround Speed Optimizer env, or equivalent "defer all JS"):  <br>4. Enable "Defer Render-Blocking JS" + Apple Pay/Google Pay, classic checkout, guest/incognito → card fields render (previously blank). Confirm no `window.wc`/`wp.data` undefined errors in console.  <br>5. Repeat with Pixel Manager deactivated → still renders.

---

- [X] Covered with tests (Jest for the bootstrap gating; manual verification noted above for the optimizer path)

### Changelog entry

Fix - Render classic-checkout card fields when a host optimizer defers render-blocking JavaScript (e.g. SiteGround Speed Optimizer) by waiting for required WordPress/WooCommerce globals before initializing the Stripe bundle.

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)


wjrosa:
The fix here was tested and confirmed working by the issue reporter (see https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5615)

client/classic/upe/index.js
```diff
@@ -1,361 +1,77 @@
...
-
-	// Handle hash change - used when authenticating payment with SCA on checkout page.
-	$( window ).on( 'hashchange', () => {
-		if ( window.location.hash.startsWith( '#wc-stripe-confirm-' ) ) {
-			maybeShowAuthenticationModal();
-		} else if ( window.location.hash.startsWith( '#confirm-' ) ) {
-			legacyHashchangeHandler( api, showError );
+			clearInterval( timer );
+			loadInit();
 		}
```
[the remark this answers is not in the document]
wjrosa: Done in ef17ee7d7 — `loadInit()` now retries the import once and logs a hard failure instead of leaving an unhandled rejection. Kept it UI-free on purpose so the bootstrap stays free of WP/WC externals (a notice would defeat that).

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -528,12 +528,18 @@ public function javascript_params() {
...
+			'isUPEEnabled'   => true,
+			'key'            => $this->publishable_key,
+			'locale'         => WC_Stripe_Helper::convert_wc_locale_to_stripe_locale( get_locale() ),
+			'apiVersion'     => WC_Stripe_API::STRIPE_API_VERSION,
+			// Base URL for the classic UPE bundle's async init chunk. The JS
+			// bootstrap pins webpack's publicPath to this so the chunk loads from
+			// the plugin even when a host optimizer serves the entry script from a
+			// rewritten/relocated path. Localized inline (runs at parse time), so
+			// it is available before a deferred bundle executes.
+			'pluginBuildUrl' => trailingslashit( WC_STRIPE_PLUGIN_URL ) . 'build/',
```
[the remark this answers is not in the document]
wjrosa: Added in ef17ee7d7 — `test_javascript_params_exposes_plugin_build_url` asserts the key and its value. Passing locally.

client/classic/upe/init.js
```diff
@@ -0,0 +1,360 @@
...
+			return new Promise( ( resolve, reject ) => {
+				jQuery.post( url, args ).then( resolve ).fail( reject );
+			} );
+		}
+	);
+
+	const elements = null;
+	const upeElement = null;
+	const paymentIntentId = null;
+	const isUPEComplete = false;
```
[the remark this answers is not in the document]
wjrosa: Out of scope for this PR: `init.js` is the previous `index.js` body moved **verbatim** by the entry split — `diff develop:index.js HEAD:init.js` is identical except the SCSS import. This behavior predates the PR (the active classic flow runs through `deferred-intent.js`; `paymentIntentId` is null so this submit branch is already inert). Flagging for a separate ticket rather than changing behavior in a load-resilience PR.

client/classic/upe/init.js
```diff
@@ -0,0 +1,360 @@
...
+			return;
+		}
+
+		blockUI( $form );
+		// Create object where keys are form field names and values are form field values
+		const formFields = $form.serializeArray().reduce( ( obj, field ) => {
+			obj[ field.name ] = field.value;
+			return obj;
+		}, {} );
+		try {
```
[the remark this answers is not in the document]
wjrosa: Out of scope for this PR: `init.js` is the previous `index.js` body moved **verbatim** by the entry split — `diff develop:index.js HEAD:init.js` is identical except the SCSS import. This behavior predates the PR (the active classic flow runs through `deferred-intent.js`; `paymentIntentId` is null so this submit branch is already inert). Flagging for a separate ticket rather than changing behavior in a load-resilience PR.

client/classic/upe/init.js
```diff
@@ -0,0 +1,360 @@
...
+
+		// Boolean `true` means that there is nothing to confirm.
+		if ( confirmation === true ) {
+			return;
+		}
+
+		const { request, isOrderPage } = confirmation;
+
+		if ( isOrderPage ) {
+			blockUI( $( '#order_review' ) );
```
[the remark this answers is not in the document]
wjrosa: Out of scope for this PR: `init.js` is the previous `index.js` body moved **verbatim** by the entry split — `diff develop:index.js HEAD:init.js` is identical except the SCSS import. This behavior predates the PR (the active classic flow runs through `deferred-intent.js`; `paymentIntentId` is null so this submit branch is already inert). Flagging for a separate ticket rather than changing behavior in a load-resilience PR.

client/classic/upe/init.js
```diff
@@ -0,0 +1,360 @@
...
+				}
+			}
+		} )
+		.on( 'change', '.wc_payment_methods', () => {
+			// Check to see whether we should display the Cash App limit notice.
+			if ( $( 'input#payment_method_stripe_cashapp' ).is( ':checked' ) ) {
+				maybeShowCashAppLimitNotice(
+					'.woocommerce-checkout-payment',
+					Number( getStripeServerData()?.cartTotal )
+				);
```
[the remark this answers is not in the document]
wjrosa: Out of scope for this PR: `init.js` is the previous `index.js` body moved **verbatim** by the entry split — `diff develop:index.js HEAD:init.js` is identical except the SCSS import. This behavior predates the PR (the active classic flow runs through `deferred-intent.js`; `paymentIntentId` is null so this submit branch is already inert). Flagging for a separate ticket rather than changing behavior in a load-resilience PR.

client/classic/upe/init.js
```diff
@@ -0,0 +1,360 @@
...
+	maybeShowAuthenticationModal();
+
+	// Handle hash change - used when authenticating payment with SCA on checkout page.
+	$( window ).on( 'hashchange', () => {
+		if ( window.location.hash.startsWith( '#wc-stripe-confirm-' ) ) {
+			maybeShowAuthenticationModal();
+		} else if ( window.location.hash.startsWith( '#confirm-' ) ) {
+			legacyHashchangeHandler( api, showError );
+		}
+	} );
```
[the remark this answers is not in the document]
wjrosa: Out of scope for this PR: `init.js` is the previous `index.js` body moved **verbatim** by the entry split — `diff develop:index.js HEAD:init.js` is identical except the SCSS import. This behavior predates the PR (the active classic flow runs through `deferred-intent.js`; `paymentIntentId` is null so this submit branch is already inert). Flagging for a separate ticket rather than changing behavior in a load-resilience PR.

mahangu:
@claude review

edit: was debugging AI reviews. please ignore!

client/classic/upe/index.js
```diff
@@ -1,361 +1,78 @@
...
+// Give up after this long and load anyway, so a genuinely missing dependency
+// fails as it would without this gate rather than silently never initializing.
+const READY_TIMEOUT_MS = 10000;
+const READY_POLL_MS = 50;
+
+const dependenciesReady = () =>
+	typeof window !== 'undefined' &&
+	!! ( window.wp && window.wp.data ) &&
+	!! ( window.wp && window.wp.i18n ) &&
+	!! ( window.wc && window.wc.wcSettings );
```
malithsen: Could this monitored dependency list drift from the actual implementation?

I wonder if we should flip the logic. Instead of "wait for globals A/B/C, then import," just `import('./init')` if it throws, wait and retry until it succeeds or the timeout fires. That way we don't need a list of dependencies.

wjrosa: Good call on the drift risk! Instead of your suggestion, I made the list derive from the build so it can't drift in 0b6bdd90e

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -500,15 +500,21 @@ public function payment_scripts() {
 			'woocommerce-gateway-stripe'
 		);
 
+		/**
+		 * Filters the UPE classic checkout JavaScript parameters.
+		 *
+		 * @param array $params UPE JavaScript parameters.
+		 */
+		$upe_params = apply_filters( 'wc_stripe_upe_params', $this->javascript_params() );
```
malithsen: If the filter returns a non-array, line 512 will result in a fatal error. Should we add a guard?

client/classic/upe/index.js
```diff
@@ -1,361 +1,128 @@
...
-			$form.removeClass( 'processing' ).unblock();
-			showError( error.message );
+	if ( HANDLE_GLOBAL_PATHS[ handle ] ) {
+		return HANDLE_GLOBAL_PATHS[ handle ];
+	}
+	if ( handle.indexOf( 'wp-' ) === 0 ) {
+		return [ 'wp', camelCaseDash( handle.slice( 'wp-'.length ) ) ];
+	}
+	// Unknown handle: skip rather than risk hanging on a wrong path.
+	return null;
```
malithsen, line 67: Should we add a `console.warn` here? I'm thinking of a future dependency with an unmapped handle.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Guarded in 28216d8a1 — falls back to the unfiltered params, with a test.

client/classic/upe/index.js
wjrosa: Added in 28216d8a1, with a test.

client/classic/upe/index.js
```diff
@@ -1,361 +1,132 @@
...
-				}
+	return true;
+};
+
+const dependenciesReady = () => requiredGlobalPaths.every( globalPathReady );
+
+// Retry once on a transient chunk-load failure; log a hard failure rather than
+// leave an unhandled rejection.
+const loadInit = ( retriesLeft = 1 ) =>
+	import( /* webpackChunkName: "upe-classic-init" */ './init' ).catch(
```
jaclync: The async chunk results in a second serial request on the checkout critical path for most stores, which is prone to network conditions. How about keeping the gate but not splitting with [`webpackMode: "eager"`](https://webpack.js.org/api/module-methods/#magic-comments) which keeps `./init` in the same file while still deferring *executing* it until the gate opens (the fix)? Then the publicPath block, `pluginBuildUrl` param, and retry won't be needed.

I didn't find pre-existing webpack eager mode usage in the codebase, just wanted to post it as an idea.

```suggestion
	import( /* webpackMode: "eager" */ './init' ).catch(
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -500,15 +500,26 @@ public function payment_scripts() {
...
+
+		// A filter callback returning a non-array must not fatal the checkout.
+		if ( ! is_array( $upe_params ) ) {
+			$upe_params = $this->javascript_params();
```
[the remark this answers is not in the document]
wjrosa: Fixed in d6cb409b6 — the pre-filter params are stored and reused as the fallback, and the test now requires `javascript_params()` to be called exactly once. Valid catch: it creates a payment token from the setup intent on the success-redirect path, so the second call risked duplicating that.

client/classic/upe/index.js
```diff
@@ -1,361 +1,130 @@
...
-import { initializeAlwaysExpandedOptimizedCheckout } from './always-expanded-optimized-checkout';
-import { legacyHashchangeHandler } from './legacy-support';
+/* global wc_stripe_upe_params */
+/*
+ * Dependency-free bootstrap for the classic-checkout Stripe (UPE) bundle.
+ *
+ * "Defer render-blocking JS" optimizers can run our bundle before the WP/WC
+ * globals ./init imports exist, throwing at load. Wait for those globals (the
+ * list is derived from the build's declared dependencies so it can't drift),
+ * then load ./init as an async chunk. This file imports no WP/WC external.
```
jaclync: nit: the async chunk might be outdated.

client/classic/upe/index.js
```diff
@@ -1,361 +1,130 @@
...
-	PAYMENT_METHOD_OXXO,
-} from 'wcstripe/stripe-utils/constants';
 
-jQuery( function ( $ ) {
-	const key = getStripeServerData()?.key;
-	if ( ! key ) {
-		// If no configuration is present, probably this is not the checkout page.
-		return;
+// Pin the async-chunk base URL to the plugin build dir; webpack's default "auto"
+// publicPath would resolve ./init against the optimizer's rewritten entry path.
```
jaclync: nit: the comment about async chunk might be outdated.

client/classic/upe/index.js
```diff
@@ -1,361 +1,130 @@
...
+// eslint-disable-next-line no-undef, camelcase
+const hasWebpackPublicPath = typeof __webpack_public_path__ !== 'undefined';
+if (
+	hasWebpackPublicPath &&
+	bootstrapParams &&
+	bootstrapParams.pluginBuildUrl
+) {
+	// eslint-disable-next-line no-undef, camelcase
+	__webpack_public_path__ = bootstrapParams.pluginBuildUrl;
+}
```
jaclync: nit: with webpack eager mode without a separate chunk request, is the public path still used?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -535,12 +551,15 @@ public function javascript_params() {
...
-			'apiVersion'   => WC_Stripe_API::STRIPE_API_VERSION,
+			'gatewayId'      => self::ID,
+			'title'          => $this->title,
+			'isUPEEnabled'   => true,
+			'key'            => $this->publishable_key,
+			'locale'         => WC_Stripe_Helper::convert_wc_locale_to_stripe_locale( get_locale() ),
+			'apiVersion'     => WC_Stripe_API::STRIPE_API_VERSION,
+			// Base URL the bootstrap pins webpack's publicPath to, so the async init
+			// chunk loads from the plugin even when an optimizer rewrites the path.
+			'pluginBuildUrl' => trailingslashit( WC_STRIPE_PLUGIN_URL ) . 'build/',
```
jaclync: nit: similar to the unused public path code, this is likely not needed anymore.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -874,6 +874,91 @@ public function test_payment_scripts_registers_script_with_correct_version(): vo
...
+	/**
+	 * The bootstrap pins webpack's publicPath to `pluginBuildUrl`; guard that
+	 * PHP→JS contract key here.
+	 */
+	public function test_javascript_params_exposes_plugin_build_url(): void {
+		$params = $this->mock_gateway->javascript_params();
+
+		$this->assertArrayHasKey( 'pluginBuildUrl', $params );
+		$this->assertSame( trailingslashit( WC_STRIPE_PLUGIN_URL ) . 'build/', $params['pluginBuildUrl'] );
+	}
```
jaclync: nit: similar to the unused public path code, this is likely not needed anymore.

client/classic/upe/index.js
```diff
@@ -1,361 +1,130 @@
...
-			// Check to see whether we should display the Cash App limit notice.
-			if ( $( 'input#payment_method_stripe_cashapp' ).is( ':checked' ) ) {
-				maybeShowCashAppLimitNotice(
-					'.woocommerce-checkout-payment',
-					Number( getStripeServerData()?.cartTotal )
-				);
-			} else {
-				removeCashAppLimitNotice();
-			}
+const dependenciesReady = () => requiredGlobalPaths.every( globalPathReady );
```
jaclync: ❓ This dependencies gate waits for the wp/wc globals but not for `wc_stripe_upe_params`, which is depended on `init.js`'s initialization but only read once at eval. If some optimizers reorder the inline script so that `wc_stripe_upe_params` becomes available later, `init.js` might fail silently. Does it make sense to also wait for the params?

```suggestion
const dependenciesReady = () => typeof wc_stripe_upe_params !== 'undefined' && requiredGlobalPaths.every( globalPathReady );
```

client/classic/upe/index.js
```diff
@@ -1,361 +1,130 @@
...
+	import( /* webpackMode: "eager" */ './init' ).catch( ( error ) => {
+		if ( retriesLeft > 0 ) {
+			return loadInit( retriesLeft - 1 );
 		}
+		// eslint-disable-next-line no-console
+		console.error(
+			'WooCommerce Stripe: failed to load the classic checkout init chunk.',
+			error
+		);
 	} );
```
jaclync: ❓ With eager mode, do we still need the retry (probably for network failures originally)? The retry could hide actual failures from init.

```suggestion
// Retry once on a transient chunk-load failure; log a hard failure rather than
// leave an unhandled rejection.
const loadInit = () =>
	import( /* webpackMode: "eager" */ './init' ).catch( ( error ) => {
		// eslint-disable-next-line no-console
		console.error(
			'WooCommerce Stripe: failed to load the classic checkout init chunk.',
			error
		);
} );
```

jaclync (commented):
Initial testing worked well (with the mu-plugin simulation below, pre-fix showed 2 wcSettings read errors in console + payment methods not loaded; post-fix showed 1 wcSettings read error from express checkout bundle + payment methods fixed). Had some questions, will do more testing tomorrow.

mu-plugin to simulate the deferred dependent JS's, feel free to suggest any other ways to repro:
```php
add_filter( 'script_loader_tag', function ( $tag, $handle ) {
	$is_library = 0 === strpos( $handle, 'wp-' )
		|| 0 === strpos( $handle, 'wc-' )
		|| in_array( $handle, [ 'react', 'react-dom' ], true );

	// Keep the plugin's own bundles blocking: deferring them too would restore
	// ordering (defer preserves execution order) and mask the bug.
	if ( $is_library && 0 !== strpos( $handle, 'wc-stripe' ) && false === strpos( $tag, ' defer' ) ) {
		return str_replace( ' src=', ' defer src=', $tag );
	}
	return $tag;
}, 10, 2 );
```

client/classic/upe/index.js
wjrosa: Adopted in 751e44139 — and the now-unneeded publicPath block, pluginBuildUrl param, and retry are removed in 137663f2e. Thanks, much simpler.

client/classic/upe/index.js
wjrosa: Updated in 137663f2e.

client/classic/upe/index.js
wjrosa: Right — the whole publicPath block only served the separate chunk. Removed in 137663f2e.

client/classic/upe/index.js
wjrosa: Not anymore — removed in 137663f2e.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Removed in 137663f2e.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Removed with the param in 137663f2e.

client/classic/upe/index.js
wjrosa: Agreed — your applied suggestion is kept (reformatted for lint in 137663f2e) with a Jest test for the reordered-inline-script case.

client/classic/upe/index.js
wjrosa: Agreed — with eager mode a rejection can only be an init failure, so retrying just re-runs it. Your suggestion is kept, reformatted in 137663f2e.

jaclync:
❓ While testing the PR (regression tests), I noticed an issue with ECE: with "automatically generate account password" disabled (`woocommerce_registration_generate_password = no`), every express checkout payment (Apple/Google Pay, Link, Amazon Pay) failed with `wc_stripe_express_checkout_missing_required_fields` ("Create account password is a required field") for any product and any user, logged-in or not, while the same cart completes classic checkout fine.

Does this happen on your end as well? Could it be related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5640?

jaclync (approved):
LGTM :shipit: shared an ECE issue before https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5631#issuecomment-5249775145 but I think it's pre-existing.

✅ Verified locally: classic + Blocks purchases complete with and without the simulation, classic checkout with the simulation.

Some non-blocking notes:

1. PR description could be updated, the fix section still refers to using an async chunk.
2. Observations from testing with the JS reordering simulation in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5631#pullrequestreview-4894224243, which I think are out of scope:
  - Express checkout error logs are still in the console in classic checkout.
  - The Blocks checkout renders as a permanent loading skeleton.

wjrosa:
Confirmed the ECE failure is pre-existing from #5640: WC core marks `account_password` required unconditionally when password generation is off, and the custom-field enforcement doesn't exclude the `account` fieldset — so every express payment 400s on a field the sheet can't collect. Fix in #5800 (STRIPE-1345), with regression tests.

Also updated the PR description (eager bundle instead of async chunk) and filed the simulation observations as STRIPE-1346 (ECE console errors) and STRIPE-1347 (Blocks skeleton).

PR #5632, opened 2026-06-24
Register the Optimized Checkout element in the Cart/Checkout block editor

wjrosa:
Fixes STRIPE-1240
Related to #5605

With Optimized Checkout (OC) enabled, the Cart/Checkout **block editor** marks the Stripe payment block as unsupported ("Stripe does not yet support this block" + empty Checkout block) whenever the **card method is disabled** and only non-card methods (e.g. Cash App Pay) are enabled. Enabling the card method clears it — the on/off behavior the report describes.

Root cause: the Blocks payment config is built from `is_valid_optimized_checkout_page()`, which calls `is_checkout()` and is therefore `false` on the wp-admin editor screen. So the editor registers the individual UPE methods (`stripe_cashapp`, …) instead of the single OC element, which is the only thing registered under the main `stripe` name. The admin gateway filter (added for the previous incompatibility fix) leaves only the `stripe` gateway enabled in the editor, so when nothing registers as `stripe`, WooCommerce reports the gateway as block-incompatible. With the card method enabled, the card method registers as `stripe` and masks the problem.

This complements #5605 (STRIPE-1230), which scoped `add_gateways()` to block-editor screens and added the `is_editing_cart_or_checkout_block()` helper this PR reuses. #5605 fixed the **gateway-list** side of the editor (leaving only the `stripe` gateway); it did not touch the **Blocks-config** side, which still gated the OC element on `is_valid_optimized_checkout_page()`. So #5605 alone does not resolve STRIPE-1240 — the two are complementary.

### What this changes

- Adds `WC_Stripe_UPE_Payment_Gateway::should_render_optimized_checkout_for_blocks()` — `true` on a real OC checkout page **and** while editing the Cart/Checkout block in admin. The editor branch reuses the same `WC_Stripe::is_editing_cart_or_checkout_block()` detection that `add_gateways()` already uses, so the gateway list and the registered Blocks methods stay in lockstep.
- Routes the three Blocks-config OC gates (`shouldShowOptimizedCheckout`, the OC method registration, and the per-method save-option map) and `WC_Stripe_Blocks_Support::is_active()` through it.

Net effect: the OC element represents Stripe in the editor under the `stripe` name regardless of which individual methods are enabled. Outside admin the helper returns exactly the previous `oc_enabled && is_valid_optimized_checkout_page()`, so frontend and classic checkout behavior is unchanged.

## Testing instructions

1. Connect a US Stripe account (USD), enable Stripe, and enable **Optimized Checkout**.
2. Under Stripe payment methods, **disable Credit/Debit card** and enable **only Cash App Pay**.
3. Edit the Checkout page in the block editor and select the Stripe payment block.
   - Before: the block shows "Your store does not have any payment methods that support the Checkout block" and the sidebar shows "Stripe does not yet support this block".
   - After: the block renders normally with no incompatibility warning.
4. Re-enable Credit/Debit card and confirm the editor still renders correctly.
5. On the frontend Checkout page, confirm Cash App Pay is selectable and a purchase completes (behavior unchanged).
6. Toggle Optimized Checkout **off** and confirm the editor and checkout still behave as before.

---

### Changelog entry

* Fix - Show the Stripe payment block as supported in the Checkout block editor when Optimized Checkout is enabled and only non-card methods (e.g. Cash App Pay) are active


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -635,8 +635,8 @@ static function ( $domain ) {
 			}
 		}
 
-		// Optimized Checkout feature flag + setting + whether we are on any of the pages that should not show OC.
-		$should_show_optimized_checkout                 = $this->oc_enabled && $this->is_valid_optimized_checkout_page();
+		// True on OC checkout pages and in the Cart/Checkout block editor (where the OC element registers as 'stripe').
+		$should_show_optimized_checkout                 = $this->should_render_optimized_checkout_for_blocks();
```
Mayisha: This method name `should_render_optimized_checkout_for_blocks` a bit misleading here. `$should_show_optimized_checkout` flag is used on both classic and block. We should rename it or just keep the name until `should_render_optimized_checkout`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -854,6 +854,25 @@ public function is_valid_optimized_checkout_page(): bool {
...
+	 * Whether the single OC element should represent Stripe when building the Blocks config.
+	 *
+	 * Covers real OC checkout pages plus the Cart/Checkout block editor, where is_checkout()
+	 * (and thus is_valid_optimized_checkout_page()) is false. Without the editor case the editor
+	 * registers the individual UPE methods instead of the OC element — the only one named 'stripe' —
+	 * so with card disabled nothing maps to the 'stripe' gateway and the block reports it unsupported.
+	 *
+	 * @return bool
+	 */
+	public function should_render_optimized_checkout_for_blocks(): bool {
```
Mayisha: With renaming the method, the docblock also needs update as this covers classic checkout too.

Mayisha (commented):
I am seeing this incompatibility notice in the block editor when I disable card and keep Cash App (only and/or other non-card methods) enbaled.

Image

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Good call — renamed to `should_render_optimized_checkout` in cdad06a2e since it gates the OC element on classic OC pages too, not just Blocks.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Done — docblock now says it covers real OC checkout pages (classic and Blocks) plus the Cart/Checkout block editor.

wjrosa:
Thanks for the review and testing, Mayisha! I have fixed the issue in 0208872ad

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -854,6 +854,26 @@ public function is_valid_optimized_checkout_page(): bool {
...
 	}
 
+	/**
+	 * Whether the single OC element should represent Stripe for the current request.
+	 *
+	 * Covers real OC checkout pages (classic and Blocks) plus the Cart/Checkout block editor,
+	 * where is_checkout() (and thus is_valid_optimized_checkout_page()) is false. Without the
+	 * editor case the editor registers the individual UPE methods instead of the OC element —
+	 * the only one named 'stripe' — so with card disabled nothing maps to the 'stripe' gateway
+	 * and the block reports it unsupported.
```
malithsen: nit: I'd consider removing this paragraph. This context is unnecessary

malithsen (approved):
Works as described, blocks edit page no longer has the error.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Done in 88c85f75d — trimmed the rationale paragraph, kept the one-line note that the editor case is covered even though `is_checkout()` is false there.

PR #5634, opened 2026-06-25
Fix Adaptive Pricing PaymentIntent fallback

rakshitwesley:
## Summary
- Recover Adaptive Pricing orders when a successful or capturable PaymentIntent webhook can only be linked through its Checkout Session.
- Store the recovered PaymentIntent ID on the order before deferring the existing settlement workflow.
- Keep the fallback scoped to Adaptive Pricing Checkout Session intents so non-AP lookalike events remain untouched.

## Changes
- Generalizes the Adaptive Pricing Checkout Session order fallback from failed intents to failed, succeeded, and capturable PaymentIntent events.
- Uses `payment_details.order_reference` as the direct Checkout Session link when Stripe includes it, falling back to the existing PaymentIntent session lookup.
- Requeues the recovered PaymentIntent webhook if the target order is currently locked, so the intent ID is not lost before deferred settlement can run.
- Adds webhook-handler PHPUnit coverage for AP success, AP capturable, AP failure, standard checkout, unmarked intents, stale session references, and locked-order retry handling.
- Enables legacy Woo Subscriptions product types during E2E setup so the existing subscription fixtures remain compatible with Woo Subscriptions 9.0.0.
- Adds a changelog/readme entry for the recovery fix.

Screenshots:
image
image

## Definition of Done
- Adaptive Pricing checkout orders can recover when `payment_intent.succeeded` or `payment_intent.amount_capturable_updated` arrives but the matching Checkout Session success webhook is missing or delayed.
- The recovered WooCommerce order records the Stripe PaymentIntent ID and continues through the existing deferred settlement workflow.
- Failed Adaptive Pricing intents continue to use the Checkout Session fallback behavior.
- Non-Adaptive-Pricing PaymentIntent events, including Checkout Session-looking references, do not recover or alter unrelated pending orders.
- Locked orders retry instead of dropping the PaymentIntent recovery path.
- Stripe E2E subscription tests can create their subscription fixtures with the latest Woo Subscriptions setup.

## Backwards Compatibility
No public API changes. No hooks, filters, classes, or files were renamed or removed. The changed helper is private and keeps a default value for the new optional parameter.

## Validation
- `php -l includes/class-wc-stripe-webhook-handler.php`
- `php -l tests/phpunit/class-wc-stripe-webhook-handler-test.php`
- `/Users/rakshitwesley/Desktop/claude/may12/woocommerce-gateway-stripe/vendor/bin/phpcs --standard=phpcs.xml.dist -n includes/class-wc-stripe-webhook-handler.php tests/phpunit/class-wc-stripe-webhook-handler-test.php`
- `npm run phpstan` / `composer phpstan` equivalent through the repo container — OK
- `npm run test:php -- --filter='WC_Stripe_Webhook_Handler_Test::test_process_adaptive_pricing_payment_intent_order_reference_fallback|WC_Stripe_Webhook_Handler_Test::test_process_payment_intent_checkout_session_fallback|WC_Stripe_Webhook_Handler_Test::test_process_payment_intent_checkout_session_fallback_retries_when_order_is_locked|WC_Stripe_Webhook_Handler_Test::test_process_deferred_webhook_retries_payment_intent_processing'` — OK, 11 tests / 46 assertions
- `docker compose exec -u www-data -e XDEBUG_MODE=coverage wordpress /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/vendor/bin/phpunit --configuration /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/phpunit.xml.dist --filter=WC_Stripe_Webhook_Handler_Test` — OK, 91 tests / 260 assertions
- `bash -n tests/e2e/bin/setup.sh`
- `node --check tests/e2e/config/global-setup.js`
- `node --check tests/e2e/utils/playwright-setup.js`
- `git diff --check`
- Targeted JS ESLint was not available locally because this checkout does not have a local `node_modules/.bin/eslint`; CI will run the project linter.
- Veloria: skipped — no hooks or public API modified.
- QIT: not configured locally; CI/QIT can run on the PR.

## Test Plan
- Enable Optimized Checkout Suite and Adaptive Pricing.
- Complete an Adaptive Pricing checkout where `payment_intent.succeeded` arrives but the matching Checkout Session success webhook is missing or delayed.
- Confirm the WooCommerce order records the PaymentIntent and proceeds through the existing deferred settlement workflow.
- Confirm non-Adaptive-Pricing PaymentIntent events with a Checkout Session-looking reference do not recover or alter unrelated pending orders.

STRIPE-1242

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1278,7 +1278,9 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 			&& $is_adaptive_pricing_checkout_intent
 			&& in_array( $notification->type, $checkout_session_intent_event_types, true );
 		$checkout_session_order_reference    = $intent->payment_details->order_reference ?? '';
-		$checkout_session_order_reference    = is_string( $checkout_session_order_reference ) ? $checkout_session_order_reference : '';
+		$checkout_session_order_reference    = is_string( $checkout_session_order_reference ) && 0 === strpos( $checkout_session_order_reference, 'cs_' )
+			? $checkout_session_order_reference
+			: '';
 
 		// Checkout Session PaymentIntents can arrive without order metadata; the session remains the durable order link.
 		if ( $should_resolve_via_checkout_session ) {
@@ -1528,22 +1530,22 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	 * object-shaped payload.
 	 *
 	 * @param array|object $notification Raw notification from the scheduled job.
-	 * @return object      The normalized notification object.
+	 * @return stdClass    The normalized notification object.
 	 * @throws Exception When the payload cannot be normalized.
 	 */
 	private function normalize_deferred_webhook_notification_to_object( $notification ) {
-		if ( is_object( $notification ) ) {
+		if ( $notification instanceof stdClass ) {
 			return $notification;
 		}
 
-		if ( is_array( $notification ) ) {
+		if ( is_array( $notification ) || is_object( $notification ) ) {
 			$json = wp_json_encode( $notification );
 			if ( false === $json ) {
 				throw new Exception( 'Failed to encode deferred webhook notification for object restoration.' );
 			}
 
 			$object = json_decode( $json );
-			if ( ! is_object( $object ) ) {
+			if ( ! $object instanceof stdClass ) {
 				throw new Exception( 'Failed to restore deferred webhook notification to an object.' );
 			}
 
@@ -1574,10 +1576,7 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 				case 'payment_intent.succeeded':
 				case 'payment_intent.amount_capturable_updated':
 					if ( ! empty( $additional_data['retry_process_payment_intent'] ) ) {
-						if ( $notification instanceof stdClass ) {
-							$this->process_payment_intent( $notification );
-						}
-
+						$this->process_payment_intent( $notification );
 						return;
 					}
 
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -1266,13 +1266,24 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 		$intent = $notification->data->object;
 		$order  = $this->get_order_from_intent( $intent );
 
-		$checkout_type = $intent->metadata->checkout_type ?? '';
+		$checkout_type                       = $intent->metadata->checkout_type ?? '';
+		$resolved_order_via_checkout_session = false;
+		$checkout_session_intent_event_types = [
+			'payment_intent.payment_failed',
+			'payment_intent.succeeded',
+			'payment_intent.amount_capturable_updated',
+		];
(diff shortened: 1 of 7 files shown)


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1296,9 +1307,33 @@ public function process_payment_intent( $notification ) {
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
 
 		if ( $order_helper->lock_order_payment( $order ) ) {
+			if (
+				$resolved_order_via_checkout_session
+				&& isset( $intent->id )
+				&& in_array( $notification->type, [ 'payment_intent.succeeded', 'payment_intent.amount_capturable_updated' ], true )
+			) {
```
wjrosa: This 3-part guard (`$resolved_order_via_checkout_session && isset( $intent->id ) && in_array( …succeeded/capturable… )`) is repeated verbatim in the unlocked branch right below. Worth hoisting into one named bool (e.g. `$is_recoverable_success_event`) so the two can't drift.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1543,6 +1578,14 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
 			switch ( $webhook_type ) {
 				case 'payment_intent.succeeded':
 				case 'payment_intent.amount_capturable_updated':
+					if ( ! empty( $additional_data['retry_process_payment_intent'] ) ) {
+						if ( $notification instanceof stdClass ) {
```
wjrosa: By this point `$notification` has already passed through `normalize_deferred_webhook_notification_to_object()`, which returns a deep `stdClass` or throws — so this `instanceof stdClass` is effectively always true, and the implicit else silently `return`s, dropping the recovery with no log (unlike the sibling cases that throw on bad data). Drop the guard, or log/throw in the else.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1266,13 +1266,24 @@ public function process_payment_intent( $notification ) {
...
+		$checkout_session_intent_event_types = [
+			'payment_intent.payment_failed',
+			'payment_intent.succeeded',
+			'payment_intent.amount_capturable_updated',
+		];
+		$is_adaptive_pricing_checkout_intent = WC_Stripe_Checkout_Sessions_Ajax_Handler::ADAPTIVE_PRICING_CHECKOUT_TYPE === $checkout_type;
+		$should_resolve_via_checkout_session = ! $order
+			&& $is_adaptive_pricing_checkout_intent
+			&& in_array( $notification->type, $checkout_session_intent_event_types, true );
+		$checkout_session_order_reference    = $intent->payment_details->order_reference ?? '';
```
wjrosa, line 1280: This treats `payment_details.order_reference` as a Checkout Session id (`cs_…`). Worth confirming Stripe always populates it with the session id on these events — if it's ever something else the direct lookup just no-ops and falls back to the intent-session lookup, so it's safe, just silently slower.

malithsen (commented):
From a higher level, I don't think this approach is correct for the underlying issue. Left a more detailed explanation in the Linear ticket.

mahangu:
@claude review

mahangu:
@claude review

mahangu:
@claude review

mahangu:
Sorry for the repeated AI Review runs on this PR. We were debugging why the automated review stopped posting and used it as a test case. Should be fixed now.

PR #5637, opened 2026-06-25
Add express checkout placement simulator to customize tabs

wjrosa:
Towards STRIPE-1323

Adds an eligibility/placement simulator (an **Eligibility** section, not an interactive preview) directly under the live **Preview** on each of the three "Customize express checkouts" tabs — Apple Pay / Google Pay, Amazon Pay, and Link. It answers the common support question *"why is the express checkout button hidden?"* for the merchant's **current configuration** (including unsaved changes), right where they configure these buttons.

The live Preview only proves the button can render in the admin's browser; it says nothing about *where* on the storefront it will appear given the current settings. The simulator fills that gap.

Each tab shows:

- **Eligibility checklist** (✓ / ✗ / ⓘ): Stripe account connected, method enabled, HTTPS (informational in test mode, required in live), mode — plus tab-specific gates. Amazon Pay adds account-country, store-currency, and tax-setup checks; Apple/Google Pay and Link add the "card method required" check.
- **Where the button appears**: one row per location (Product / Cart / Checkout, plus Change payment method for subscriptions on Express/Link). The first failing blocking check hides every location with that reason; otherwise a location shows only when its toggle is enabled.
- A footnote noting the storefront-only gates that can't be evaluated in wp-admin (supported device/browser, cart contents and product types, address-based taxes).

The simulator is fully reactive — it updates live as you toggle locations or enable/disable the method.

Preview:
Screenshot 2026-07-30 at 16 14 32

**How it's built**

- New shared presentational component `client/settings/express-checkout-simulator/`. It's intentionally dumb: each settings section computes the gate checks and per-location toggle state and passes them in.
- Reuses the existing `PAYMENT_METHOD_UNAVAILABLE_REASONS` wording (from the unavailable pills) so explanations stay consistent with the main settings page.
- The account-level gates (connected / HTTPS / mode) come from one shared `WC_Stripe_Helper::get_express_checkout_simulator_gate_params()` used by all three controllers, rather than duplicated per controller. The HTTPS gate reads `wp_is_using_https()` (the configured site URLs) since the admin request's own scheme isn't what the storefront gate sees.
- Amazon's **account-country** and **currency** checks are computed **server-side** (`WC_Stripe_UPE_Payment_Method_Amazon_Pay`) and localized to the page, so the client keeps no copy of those eligibility rules.
- No changes to storefront rendering or to the gateway's gating logic — the simulator only reflects existing rules.

**How can this break / what guards against it?** The component is presentational and reflects only already-localized params and reactive settings state; it can't affect checkout. Covered by Jest (verdict precedence, check builders) and PHPUnit (the shared gate params and the Amazon Pay eligibility params are localized).

> A follow-up refactor that de-duplicates the three tabs' shared UI (override notice, location checkboxes, size control) is stacked on top of this in #5639, kept separate to keep this PR focused on the feature.

## Testing instructions

1. Go to **WooCommerce → Settings → Payments → Stripe**, then open **Customize** for Apple Pay / Google Pay (repeat for Amazon Pay and Link).
2. Confirm an **Eligibility** section appears under the **Preview**, with an eligibility checklist and a "Where the button appears" table. (It's a static readout of the current settings, not an interactive preview.)
3. Toggle the location checkboxes (Product / Cart / Checkout) and confirm the table flips live — enabled locations show ✓ "Would display here", disabled ones show ✗ "Not enabled for this location".
4. Disable the method (uncheck Apple Pay / Google Pay on the main Stripe settings, or disable Amazon Pay / Link) and confirm every location flips to ✗ citing the method being disabled.
5. On the **Amazon Pay** tab, set the store currency (**WooCommerce → Settings → General**) to one Amazon Pay doesn't support and confirm every location flips to ✗ with the unsupported-currency reason; restore it and confirm it clears.

---

-   [x] Tested on mobile (or does not apply) — admin settings UI; no mobile-specific behavior.

### Changelog entry

* Add - Show a placement simulator on each Customize express checkouts tab that previews where the express checkout button would and wouldn't appear, with the reason

Added to both `changelog.txt` and `readme.txt` under 11.0.0.

**Post merge**


client/settings/express-checkout-simulator/index.js
```diff
@@ -0,0 +1,117 @@
...
+						<Icon icon={ STATUS_ICON[ c.status ] } size={ 20 } />
+						<span className="express-checkout-simulator__check-label">
+							{ c.label }
+						</span>
+						{ c.detail && (
+							<span className="express-checkout-simulator__check-detail">
+								{ c.detail }
+							</span>
+						) }
+					</li>
```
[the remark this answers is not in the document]
wjrosa: Addressed in 2a9a742a2 — each check row now renders a `VisuallyHidden` status ("Passed:" / "Failed:" / "Information:") before the label, with a test asserting the accessible output for all three statuses. The ast-grep key warnings are false positives: the `<li>` carries the key; its children don't need one.

client/settings/express-checkout-simulator/build-checks.js
```diff
@@ -0,0 +1,197 @@
...
+export const buildCardMethodCheck = ( { isCardEnabled, methodLabel } ) => ( {
+	key: 'card-method',
+	label: __( 'Card payment method enabled', 'woocommerce-gateway-stripe' ),
+	status: isCardEnabled ? STATUS.PASS : STATUS.FAIL,
+	detail: '',
+	blockingText: getReasonText(
+		PAYMENT_METHOD_UNAVAILABLE_REASONS.REQUIRES_CARD_METHOD,
+		methodLabel
+	),
+} );
```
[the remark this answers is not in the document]
wjrosa: Added in 01fe8883c — a dedicated `build-checks.test.js` covers all four builders directly: account/method failures with their blocking text, HTTPS as info in test mode vs. fail in live mode, currency pass/fail/null-when-unrestricted, the card-method gate, and location enablement mapping. 16 tests green across the two suites.

includes/class-wc-stripe-helper.php
```diff
@@ -2160,6 +2160,21 @@ public static function is_connected( $mode = null ) {
...
+	 * Returns the account-level gate values the Customize express checkouts settings pages localize
+	 * for their placement simulator. These gates apply to every express method (Apple Pay/Google Pay,
+	 * Amazon Pay, Link), so they live here rather than being duplicated across the three controllers.
+	 *
+	 * @return array{is_account_connected: bool, is_https: bool, is_test_mode: bool}
+	 */
+	public static function get_express_checkout_simulator_gate_params(): array {
+		return [
+			'is_account_connected' => self::is_connected(),
+			'is_https'             => is_ssl(),
```
jaclync: ❓ Same as what was raised in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5637#pullrequestreview-4952825599 by coderabbitai, `is_ssl` here is in the admin context and I think the storefront context is intended here? [`wp_is_using_https`](https://developer.wordpress.org/reference/functions/wp_is_using_https/) might be worth considering.

client/settings/express-checkout-simulator/build-checks.js
```diff
@@ -0,0 +1,197 @@
...
+
+/**
+ * Maps the location keys a tab supports to `{ key, label, enabled }` rows for the simulator,
+ * where `enabled` reflects the tab's current (reactive) location toggles.
+ *
+ * @param {string[]} keys             Location keys this tab supports, in display order.
+ * @param {string[]} enabledLocations The currently enabled location keys.
+ * @return {Array<Object>} Location rows for `<ExpressCheckoutSimulator locations />`.
+ */
+export const buildLocations = ( keys, enabledLocations ) => {
```
jaclync, line 29: non-blocking suggestion: does it make sense to make the location key an enum that is kept in sync with `WC_Stripe_Express_Checkout_Helper::get_button_locations` (`getLocationLabels` then maps the enum) so that it's clear what all the cases are with potential validation?

client/entrypoints/amazon-pay-settings/amazon-pay-settings-section.js
```diff
@@ -71,6 +83,54 @@ const AmazonPaySettingsSection = () => {
...
+				PAYMENT_METHOD_UNAVAILABLE_REASONS.TAX_BASED_ON_BILLING_ADDRESS,
+				methodLabel
+			),
+		},
+	].filter( Boolean );
+
+	const simulatorLocations = buildLocations(
+		[ 'product', 'cart', 'checkout' ],
+		amazonPayLocations
+	);
```
jaclync, line 93: ❓ There's also a list of location checkboxes in all the EC settings, does it make sense for the supported locations list to be shared? (Same for Express/Link settings.)

client/utils/use-payment-method-currencies.js
```diff
@@ -254,6 +254,41 @@ const getAmazonPayCurrencies = () => {
...
+];
+
+/**
+ * Whether the connected Stripe account's country can use Amazon Pay. Derived client-side from the
+ * localized account country, so callers don't need a separate server-side flag.
+ *
+ * @return {boolean} True when the account country supports Amazon Pay.
+ */
+export const isAmazonPayAccountCountrySupported = () =>
+	SUPPORTED_AMAZON_PAY_ACCOUNT_COUNTRIES.includes( accountCountry );
```
jaclync: Potential issue: `accountCountry` comes from `window.wc_stripe_settings_params`, but it doesn't look like `window.wc_stripe_settings_params` is set for Amazon Pay settings (I checked the developer tools and only `wc_stripe_amazon_pay_settings_params` was available). Therefore, it falls back to `US`, and thus always supported (I tested it by editing the fallback country to an unsupported country, and the country check fails in the simulation).

How about passing the country supported check on the server side to `wc_stripe_settings_params`? This way, `SUPPORTED_AMAZON_PAY_ACCOUNT_COUNTRIES` won't be needed on the client side as a separate copy of the server-side source of truth.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.9.0
+* Add - Show a placement simulator on each Customize express checkouts tab that previews where the express checkout button would and wouldn't appear, with the reason
```
jaclync, line 4: nit: the changelog/readme entry would be moved to 11.0.0.

client/settings/express-checkout-simulator/build-checks.js
```diff
@@ -0,0 +1,197 @@
...
+			label: sprintf(
+				/* translators: %s: payment method name. */
+				__( '%s enabled', 'woocommerce-gateway-stripe' ),
+				methodLabel
+			),
+			status: methodEnabled ? STATUS.PASS : STATUS.FAIL,
+			detail: '',
+			blockingText: sprintf(
+				/* translators: %s: payment method name. */
+				__( "%s isn't enabled.", 'woocommerce-gateway-stripe' ),
```
jaclync: non-blocking nit: might we want to use `%1$s` consistent with `getReasonText()`? Same for the string in `buildCurrencyCheck()`.

client/settings/express-checkout-simulator/index.js
```diff
@@ -0,0 +1,128 @@
...
+								{ verdict.reason }
+							</span>
+						</li>
+					);
+				} ) }
+			</ul>
+
+			<p className="express-checkout-simulator__caveat">
+				{ __(
+					'This simulation reflects your saved configuration only. On the storefront the button can still be hidden by checks that can only run there — a supported device and browser, the contents and product types in the cart, and taxes that need a customer address.',
```
jaclync: non-blocking nit: `This simulation reflects your saved configuration only` might read like the checks here have been saved, while they can include unsaved changes. Might be worth rephrasing a bit.

client/settings/express-checkout-simulator/index.js
```diff
@@ -0,0 +1,128 @@
...
+ *
+ * @param {Object}        props
+ * @param {Array<Object>} props.checks    Ordered eligibility checks; see `build-checks.js`.
+ * @param {Array<Object>} props.locations `{ key, label, enabled }` for the tab's locations.
+ */
+const ExpressCheckoutSimulator = ( { checks, locations } ) => {
+	// The first failing check with blocking text gates every location; checks are ordered by
+	// precedence, so the earliest failure is the reason a merchant sees.
+	const blocker = checks.find(
+		( c ) => c.status === STATUS.FAIL && c.blockingText
```
jaclync: For learning purposes: is it recommended to use single letter variables in callbacks like this & L66? Or can we use `check` instead of `c`? In this file, `locations.map( ( location )` uses `location` instead of `l`, which reads more fluently.

jaclync (commented):
> 5. On the Amazon Pay tab, set the store currency (WooCommerce → Settings → General) to one Amazon Pay doesn't support and confirm every location flips to ✗ with the unsupported-currency reason; restore it and confirm it clears.

This step wasn't working on my end, also mentioned in the "Potential issue" comment about the localized params.

> 2. Confirm a Simulator appears under the Preview, with an eligibility checklist and a "Where the button appears" table.

non-blocking: I think it's "Eligibility" section instead of "Simulator." When I first saw "simulator," I thought it's an interactive simulator like the preview feature. No changes needed except for maybe updating the PR description to mention it's more like a "eligibility/placement simulator."

includes/class-wc-stripe-helper.php
wjrosa: Good catch, switched to `wp_is_using_https()` in d73159708. The gate now reflects the configured site URLs rather than the admin request's scheme.

client/settings/express-checkout-simulator/build-checks.js
wjrosa: Done in d73159708: added a shared `BUTTON_LOCATIONS` enum in `build-checks.js` (documented as mirroring `get_button_locations()`), used by all three tabs; `buildLocations()` now drops unknown keys instead of rendering unlabeled rows.

client/entrypoints/amazon-pay-settings/amazon-pay-settings-section.js
wjrosa: Agreed, but the checkbox markup predates this PR and its de-duplication is scoped to the follow-up #5639. Here I only unified the keys through the new `BUTTON_LOCATIONS` enum (d73159708).

client/utils/use-payment-method-currencies.js
wjrosa: You were right, thanks. In d73159708 both the account-country and the currency checks are now computed server-side (`WC_Stripe_UPE_Payment_Method_Amazon_Pay`) and localized to the page; the client-side country list (and its tests) are gone. This also fixes the currency step from your review, since the currency list previously fell back to the US list on this page.

changelog.txt
wjrosa: Done while resolving the merge conflict with develop; the entry now sits under 11.0.0 in both files.

client/settings/express-checkout-simulator/build-checks.js
wjrosa: Done in d73159708, all three strings now use `%1$s`.

client/settings/express-checkout-simulator/index.js
wjrosa: Reworded in d73159708 to "the settings currently shown on this page, including unsaved changes".

client/settings/express-checkout-simulator/index.js
wjrosa: Just brevity, no convention behind it; `check` is clearer. Renamed in d73159708 (the `check` icon import is aliased to `checkIcon` to avoid shadowing).

wjrosa:
Thanks for the review, @jaclync! All comments addressed in d73159708 (plus a merge with develop to resolve the changelog conflicts).

Re the review notes: the currency step failing was the localized-params gap you flagged; the Amazon Pay tab now gets its currency list (and the account-country verdict) from the server, so that step should work. I also updated the PR description to frame this as an eligibility/placement simulator and renamed the testing step to match the "Eligibility" heading.

wjrosa:
Thanks for the review, @jaclync! All comments addressed in d73159708 (plus a merge with develop to resolve the changelog conflicts).

Re the review notes: the currency step failing was the localized-params gap you flagged; the Amazon Pay tab now gets its currency list (and the account-country verdict) from the server, so that step should work. I also updated the PR description to frame this as an eligibility/placement simulator and renamed the testing step to match the "Eligibility" heading.

jaclync (approved):
Thanks a lot for the quick updates, LGTM! :shipit: 

A separate question, I believe Apple/Google Pay aren't available in non-https sites in test mode (at least in my localhost sites and also in the preview), while the eligibility UI said so (screenshot below). Are these two a special case?

Image

wjrosa:
Thanks for the review! That's a good observation. I've updated the HTTPS row's detail to say so when the site isn't on HTTPS: d1a703d60

PR #5641, opened 2026-06-25
Guard removed Payment Request class callbacks with a no-op compat shim

wjrosa:
Fixes STRIPE-1243

`WC_Stripe_Payment_Request` was removed in 10.4.0 (#4904), but the public `WC_Stripe::$payment_request_configuration` property was kept and left `null`. Third-party integrations that still read that property and register its methods as hook callbacks — for example Avada's "WooCommerce One Page Checkout" module registering `display_payment_request_button_html()` / `display_payment_request_button_separator_html()` on `woocommerce_review_order_before_submit` — end up with a `[ null, 'method' ]` callback. When the hook fires, WordPress throws an uncaught `TypeError` and the entire checkout page renders "There has been a critical error on this website":

```
call_user_func_array(): Argument #1 ($callback) must be a valid callback,
first array member is not a valid class name or object
```

This also explains why toggling Stripe's Express Checkout setting doesn't clear the fatal: the integration registers off the null property regardless of that setting.

The permanent fix belongs to the third-party integration, but the blast radius is a full checkout outage on a live store, so this PR adds a defensive guard on our side:

- Add `WC_Stripe_Payment_Request_Compat`, a small no-op shim with the two legacy button-rendering methods plus a `__call` catch-all for any other removed method.
- Assign `WC_Stripe::$payment_request_configuration` to the shim instead of leaving it `null`, so legacy callbacks resolve to a valid object. No Payment Request behaviour is reintroduced.

**How can this code break?** The shim is inert — it only ever returns `null`/void. The one consideration is that a caller expecting the old `null` value back from the property would now get an object; in practice nothing in the plugin reads it, and a non-null deprecated config is the safer default for external callers.

## Testing instructions

Without a copy of Avada you can reproduce the root cause directly against the registration pattern it uses:

1. `git checkout develop` and load any store with the classic/shortcode checkout.
2. In a snippet, read `$config = woocommerce_gateway_stripe()->payment_request_configuration;` (it is `null`).
3. `add_action( 'woocommerce_review_order_before_submit', [ $config, 'display_payment_request_button_html' ] );`
4. Load the checkout page → "There has been a critical error on this website"; the log shows the `TypeError` above.
5. Switch to this branch and repeat steps 2–4 → checkout renders normally and the callback runs as a no-op.

Automated: new regression test in `WC_Stripe_Test` registers the legacy callbacks and fires the hook; full `WC_Stripe_Test` suite (48 tests) passes, PHPCS is clean on the new file, and PHPStan reports no errors.

---

### Changelog entry

Added: `Fix - Prevent a fatal checkout error when a third-party integration registers callbacks against the Payment Request class removed in 10.4.0`

**Post merge**


includes/payment-methods/class-wc-stripe-payment-request-compat.php
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * No-op stand-in for the WC_Stripe_Payment_Request class removed in 10.4.0.
+ *
+ * Third parties still read woocommerce_gateway_stripe()->payment_request_configuration and
+ * register its methods as hook callbacks; a null property made those callbacks fatal at
+ * hook-fire time. This keeps them resolving to a valid object without restoring any behaviour.
+ *
+ * @deprecated 10.4.0 Use WC_Stripe_Express_Checkout_Element instead.
+ */
+class WC_Stripe_Payment_Request_Compat {
```
Mayisha, line 19: We removed the deprecated, unused class but now we are introducing a no-op shim for that class. What is the plan for this no-op class? Is it going to be in the codebase infinitely or do you plan to remove this at some point?

includes/payment-methods/class-wc-stripe-payment-request-compat.php
```diff
@@ -0,0 +1,30 @@
...
+	public function display_payment_request_button_separator_html(): void {}
+
+	/**
+	 * Swallow any other removed method so legacy callbacks never fatal.
+	 *
+	 * @param string  $name
+	 * @param mixed[] $arguments
+	 * @return null
+	 */
+	public function __call( string $name, array $arguments ) {
```
Mayisha, line 35: Should we emot deprecation notices in these no-op methods? Just to make it loud enough that the 3rd party integration is still on a deprecated path.

Mayisha (approved):
Overall looks good to me. Have a questiona and a suggestion about deprecation notices.

includes/payment-methods/class-wc-stripe-payment-request-compat.php
wjrosa: That's a good question. I added some `wc_deprecated_function` calls in faa583236 to at least notify 3rd-party plugins, so they can address it in their code. We can then remove this once we know consumers like Avada have dropped the reference

includes/payment-methods/class-wc-stripe-payment-request-compat.php
wjrosa: Agreed! Done in faa583236

PR #5642, opened 2026-06-25
Unify express checkout button locations & sizing across methods

wjrosa:
Fixes STRIPE-1093

Each express checkout method (Apple Pay / Google Pay = `payment_request`, `link`, `amazon_pay`) has its own `*_button_locations` and `*_button_size` setting, which was redundant and could render express buttons at mismatched sizes on the same page.

This PR unifies the settings experience without changing what is stored:

- **One shared button size.** The settings UI exposes a single size (`express_checkout_button_size`). The Link and Amazon Pay height getters delegate to it. On save, `link_button_size` and `amazon_pay_button_size` are written with the same value so they stay in step.
- **Per-method locations, read directly.** `WC_Stripe_Express_Checkout_Helper::get_button_locations()` reads each method's option, and the settings UI composes them into one location → methods view. Nothing new is persisted: `express_checkout_button_locations`, `link_button_locations`, and `amazon_pay_button_locations` remain the stored options, in the flat-list shape they always had.
- **No migration.** Since the stored options do not change, there is nothing to convert or delete on upgrade.

`WC_Stripe_Express_Checkout_Helper` keeps its public seam (`get_button_locations()`, `is_enabled_for_location()`), so PHP consumers and renderers are unchanged. The existing Express Checkout / Link / Amazon Pay settings pages are kept; the REST controller writes each method's option and returns the per-method lists, and the deprecated per-method size params still update the shared size. Sizing is a single global value, the three method units stay as-is, and ordering is fixed canonical. A single location-first settings screen is a possible follow-up.

The meaning of each location option is unchanged: an absent option keeps the default product and cart placement, and a non-array value (WooCommerce stores an emptied multiselect as `""`) disables that method everywhere.

### Backward compatibility

- **Rollback is safe.** A site that installs this version and then rolls back to 11.0.0 finds the same per-method options it had before, including any change made through the unified UI, because saves write through to every option older code reads.
- **Public PHP surface.** No hook, class, or method is removed. The `DEFAULT_BUTTON_LOCATIONS` and `BUTTON_LOCATION_OPTION_KEYS` constants are new additions on the helper.
- **REST.** `link_button_size` and `amazon_pay_button_size` are still accepted and returned, aliased to the shared size. A future release can remove the per-method size options once the deprecation window has passed.

### How can this code break? What prevents it?

- **A save through the unified UI dropping another method's locations** → each request param updates only its own option; covered by `test_express_checkout_button_locations_are_stored_per_method` and `test_express_checkout_button_locations_partial_update_keeps_other_methods`.
- **A method that was never configured being treated as disabled** → the helper falls back to the default placement for an absent option; covered by `test_get_button_locations_with_partial_settings`.
- **Sizes drifting between the shared and per-method options after a rollback** → the controller writes all three on every size save; covered by `test_legacy_button_size_params_update_shared_size` and `test_shared_button_size_param_wins_over_legacy_alias`.
- **Wrong rendering across surfaces** → the helper's public predicates are unchanged; classic / Blocks / OC / ECE all go through them.

## Testing instructions

Automated: targeted PHPUnit suites (express checkout helper, REST settings controller, admin notices, update manager, helper class) plus the full suite via `npm run test:php:parallel`, `npm run test:js`, `npm run phpstan`, and lints, all green.

Manual:

1. On develop, set different button **sizes** and **locations** per method (Express Checkout, Link, Amazon Pay pages).
2. Switch to this branch. Confirm the locations are preserved per method, only the Express Checkout page has a size control (the Link and Amazon Pay previews reflect the shared size), and all enabled express buttons render at the **same** size on Product / Cart / Checkout.
3. Change locations on each page and confirm they persist and render in the right places.
4. Uncheck every location for one method and save. Confirm only that method disappears and the others keep rendering.
5. Switch back to develop. Confirm the locations and sizes you saved on this branch are still what the per-method settings pages show.

---

### Changelog entry

> Update - Unify express checkout button locations and sizing across Apple Pay / Google Pay, Link, and Amazon Pay

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1875,39 +1869,145 @@ public function get_login_confirmation_settings() {
...
+				continue;
+			}
+
+			foreach ( $locations as $location ) {
+				$map[ $location ][] = $method;
+			}
+		}
+
+		return $map;
+	}
```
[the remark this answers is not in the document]
wjrosa: Fixed in 439b4d9bd: the converter now skips non-string/empty locations and de-duplicates methods, and `maybe_migrate()` treats an already-unified map as authoritative — it never feeds it back through the converter, and per-method leftovers are dropped without rebuilding the map. Tests cover the second pass and the mixed leftover state. One deliberate deviation: the flag still writes at the end — with the migration now idempotent, concurrent re-entry is harmless, while writing the flag before mutating would leave settings unmigrated forever if the update failed mid-pass.

includes/admin/class-wc-rest-stripe-settings-controller.php
```diff
@@ -621,10 +611,9 @@ private function update_link_settings( WP_REST_Request $request ) {
 	 */
 	private function update_express_checkout_settings( WP_REST_Request $request ) {
 		$attributes = [
-			'express_checkout_button_type'      => 'express_checkout_button_type',
-			'express_checkout_button_size'      => 'express_checkout_button_size',
-			'express_checkout_button_theme'     => 'express_checkout_button_theme',
-			'express_checkout_button_locations' => 'express_checkout_button_locations',
+			'express_checkout_button_type'  => 'express_checkout_button_type',
+			'express_checkout_button_size'  => 'express_checkout_button_size',
+			'express_checkout_button_theme' => 'express_checkout_button_theme',
```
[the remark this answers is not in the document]
wjrosa: Restored in 439b4d9bd — `link_button_size` and `amazon_pay_button_size` are registered again as deprecated aliases (same enum as the shared size), they update `express_checkout_button_size` when the shared param is absent (shared wins otherwise), and GET exposes both aliased to the shared value. Covered by two new controller tests.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1875,39 +1869,145 @@ public function get_login_confirmation_settings() {
...
+
+		$locations = [];
+		foreach ( $map as $location => $methods ) {
+			if ( in_array( $method, (array) $methods, true ) ) {
+				$locations[] = $location;
+			}
 		}
 
-		return $this->stripe_settings[ $key ];
+		return $locations;
```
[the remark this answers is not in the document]
wjrosa: Preserved instead of documented (439b4d9bd): no-arg `get_button_locations()` returns the Apple Pay / Google Pay locations again, matching the pre-unification default — nothing in-repo used the all-methods union, so external callers keep the old semantics. Pinned by a test.

tests/phpunit/admin/migrations/class-wc-stripe-migrate-express-checkout-button-settings-test.php
```diff
@@ -0,0 +1,84 @@
...
+	 */
+	public function test_skips_rewrite_when_no_legacy_locations() {
+		WC_Stripe_Helper::update_main_stripe_settings( [ 'express_checkout_button_size' => 'large' ] );
+
+		( new WC_Stripe_Migrate_Express_Checkout_Button_Settings() )->maybe_migrate();
+
+		$settings = WC_Stripe_Helper::get_stripe_settings();
+		$this->assertArrayNotHasKey( 'express_checkout_button_locations', $settings );
+		$this->assertSame( 'yes', get_option( WC_Stripe_Migrate_Express_Checkout_Button_Settings::MIGRATED_OPTION ) );
+	}
```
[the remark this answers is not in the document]
wjrosa: Added in 439b4d9bd — `test_second_pass_leaves_unified_map_unchanged()` (cleared flag, unified map survives) plus `test_drops_leftovers_without_rebuilding_unified_map()` for the mixed state.

includes/migrations/class-wc-stripe-migrate-express-checkout-button-settings.php
```diff
@@ -0,0 +1,63 @@
...
+
+		// Only rewrite settings when there is legacy per-method data to collapse.
+		// Fresh installs have none and pick up the default map at render time.
+		$has_legacy_locations = isset( $settings['express_checkout_button_locations'] )
+			|| isset( $settings['link_button_locations'] )
+			|| isset( $settings['amazon_pay_button_locations'] );
+
+		if ( $has_legacy_locations ) {
+			// Build the map before overwriting the option it partly derives from.
+			$settings['express_checkout_button_locations'] = WC_Stripe_Express_Checkout_Helper::build_locations_map_from_legacy( $settings );
```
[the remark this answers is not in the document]
wjrosa: Confirmed and fixed in 0b51c2ce4: both readers now go through the map-aware helper. `should_load_scripts_for_ece_location()` uses `is_enabled_for_location( 'payment_request', … )` and `check_express_checkout_location()` reads `get_button_locations( 'payment_request' )` (still bailing when the option is absent, so the notice keeps targeting explicitly-configured stores). Behavior is unchanged for every legacy shape — absent option → default placement, non-array → false, flat list → honored — and the map shape now works on both paths. Covered by a location-shape data provider on the script gate and three unified-map scenarios on the notice, including checkout occupied by another method only.

jaclync:
Maybe the PR can be updated based on the latest `develop`? There might be some merge conflicts from the recent merges of ECE PRs.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1894,38 +1888,148 @@ public function get_login_confirmation_settings() {
...
+	public static function build_locations_map_from_legacy( array $settings ): array {
+		$legacy_keys = [
+			'payment_request' => 'express_checkout_button_locations',
+			'link'            => 'link_button_locations',
+			'amazon_pay'      => 'amazon_pay_button_locations',
+		];
+
+		$map = [];
+		foreach ( self::EXPRESS_CHECKOUT_METHODS as $method ) {
+			$locations = $settings[ $legacy_keys[ $method ] ] ?? [];
```
diegocurbelo: Missing legacy keys previously meant the default `[product, cart]` locations for that method. Using `?? []` changes missing settings to `disabled`.
For example, a store with only `link_button_locations` saved loses the default Apple Pay/Google Pay and Amazon Pay locations before and after migration. This also needs a partial legacy-settings test.

includes/migrations/class-wc-stripe-migrate-express-checkout-button-settings.php
```diff
@@ -0,0 +1,75 @@
...
+
+		$settings = WC_Stripe_Helper::get_stripe_settings();
+
+		$stored_locations = $settings['express_checkout_button_locations'] ?? null;
+
+		// A unified map means a previous (or concurrent) pass already migrated —
+		// possible when the one-shot flag is lost. Feeding the map back through the
+		// legacy converter would corrupt it, so it is kept authoritative as-is.
+		$already_unified = is_array( $stored_locations )
+			&& WC_Stripe_Express_Checkout_Helper::is_locations_map( $stored_locations );
```
diegocurbelo: An empty array is a valid legacy value meaning `no Apple Pay/Google Pay locations`, so if Link/AmazonPay still has locations, this treats the empty legacy list as already migrated and then deletes those per-method locations. We should treat an empty array as "already_unified" only when no legacy per-method location keys remain; and we should add a test to cover this upgrade case.

diegocurbelo (commented):
I’m concerned about a merchant needing a rollback; this PR is replacing the original `express_checkout_button_locations` list with a map (and removing other keys). If a merchant needs to rollback then Apple Pay/Google Pay would be disabled for them (and I think Amazon and Link would go back to the defaults).

I think we should consider using a new set of keys (and deprecate the current ones), so we keep the plugin backwards compatibility in case a merchants needs/wants to rollback; we can then remove the old settings keys in a future version.

wjrosa:
Thanks for the review and suggestions! That was a good call. I changed the approach in 21258c1d. The per-method options (locations and sizes) stay stored exactly as 11.0.0 reads them. The UI still shows one map and one size, but it now builds them from those options at read time, and each save writes the options back. 

No migration, nothing deleted, so a rollback keeps every setting, even ones changed through the new UI. We can remove the per-method options later, after a deprecation window. PR description and README note are updated too.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: You're right, that changed the meaning of a missing option. Fixed in 21258c1d: `build_locations_map_from_settings()` now treats an absent option as the default `product` and `cart` placement and a non-array value (WooCommerce stores an emptied multiselect as `""`) as disabled, which is what `get_button_locations()` always did. A store with only `link_button_locations` saved keeps Apple Pay / Google Pay and Amazon Pay on their defaults. Covered by `test_build_locations_map_from_settings_with_partial_settings`, which also has the empty-array and empty-string cases.

includes/migrations/class-wc-stripe-migrate-express-checkout-button-settings.php
wjrosa: Agreed, an empty list is a valid "Apple Pay / Google Pay off everywhere" value and the migration would have dropped the Link and Amazon Pay locations in that case. Since the stored options no longer change, I removed the migration entirely in 21258c1d, so this path is gone. The REST side has a test for the same scenario now: `test_express_checkout_button_locations_partial_update_keeps_other_methods` saves an empty list for one method and checks the others are untouched.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1894,38 +1888,106 @@ public function get_login_confirmation_settings() {
...
+		'link'            => 'link_button_locations',
+		'amazon_pay'      => 'amazon_pay_button_locations',
+	];
+
+	/**
+	 * Returns the express checkout locations map (location => enabled methods).
+	 *
+	 * @return array<string, string[]>
+	 */
+	public function get_express_checkout_locations_map(): array {
```
diegocurbelo: This function does not seem to be used, and in turn this is the only called of `self::build_locations_map_from_settings`.

`get_method_locations_from_settings` is called from `get_button_locations`; but the other two are not.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1894,38 +1888,106 @@ public function get_login_confirmation_settings() {
...
-				break;
-			case 'link':
-				$key = 'link_button_locations';
-				break;
-			case 'payment_request':
-			default:
-				$key = 'express_checkout_button_locations';
-				break;
-		}
+	public const DEFAULT_BUTTON_LOCATIONS = [ 'product', 'cart' ];
```
diegocurbelo, line 1895: Should we include `checkout` in the default locations?

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Good catch. Both removed in ab372f58; `get_button_locations()` reads the option directly and the tests now go through it.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: I kept `product` and `cart` because that is what the helper already returned on develop when the option was never saved, so fresh stores keep the same buttons. The form-field default does include `checkout`, so the two disagree today; that predates this PR, and I would rather align them in a follow-up than change where buttons appear here. Happy to include it now if you prefer.

PR #5643, opened 2026-06-25
Make Blocks checkout resilient to deferred render-blocking JS (SiteGround Speed Optimizer)

wjrosa:
Refs STRIPE-1236
Refs #5615

**Stacked on #5631** (the classic-checkout fix). Review/merge that first; this PR's diff shows only the Blocks-checkout changes.

#5631 made the **classic** checkout resilient to host "defer all render-blocking JS" optimizers (SiteGround Speed Optimizer). The reporting merchant verified that fix works for classic, but found the **Checkout Block** still fails to render under the same conditions (Defer Render-Blocking JS ON + Apple Pay/Google Pay ON).

Same root cause: the Blocks UPE bundle reads the localized config and calls `registerPaymentMethod` / `registerExpressPaymentMethod` at module-evaluation time, statically importing `@woocommerce/blocks-registry`, `@woocommerce/settings`, and `@wordpress/data`/`i18n` (externalized to `window.wc.wcBlocksRegistry` / `window.wc.wcSettings` / `window.wp.data` / `window.wp.i18n`). When the optimizer reorders our bundle ahead of those globals, the first externalized read throws — and because the throw aborts the integration script, the **entire Checkout Block** fails to render, not just our payment method.

**Fix** (mirrors #5631): split the entry into a dependency-free **bootstrap** (`client/blocks/upe/index.js`) that waits for `window.wp.data` / `window.wp.i18n` / `window.wc.wcSettings` / `window.wc.wcBlocksRegistry`, then loads the former entry (now `client/blocks/upe/init.js`) as an async chunk so the externals resolve and the methods register only once those globals exist. The bootstrap imports nothing that resolves a WP/WC external, so it can't throw at load. `publicPath` is pinned to the plugin build dir via a tiny dedicated localized global (`wc_stripe_upe_blocks_params.pluginBuildUrl`) — not `wcSettings`, which is itself deferred — so the async chunk loads from the plugin even when the optimizer serves the entry from a rewritten path.

**How this can break / guards:** the bootstrap's external module *definitions* are lazy (only executed when the async chunk requires them, after the readiness check), confirmed against the build output — identical to the proven classic bootstrap. A 10s fallback loads init anyway so a genuine misconfiguration fails as it would today. Covered by Jest for the gating; the entry/async split is verified in the build (`upe-blocks.js` ~16 KB bootstrap, heavy graph in `upe-blocks-init.js`, `upe-blocks.css` preserved).

## Why draft

The triggering condition is a third-party host optimizer that can't be reproduced in CI, and Blocks payment-method **registration timing** under defer-all can only be confirmed in a real environment. A combined test build (classic + Blocks) is being shared with the merchant for verification. One known limitation: the merchant's console also showed the optimizer breaking WordPress's own `wp-data` script — that part isn't fixable from our plugin, so we can only shield our bundle.

## Testing instructions

Regression (no optimizer) — must be unchanged:
1. Blocks checkout: card fields render and a card payment completes, with Apple Pay/Google Pay enabled and disabled.
2. Classic checkout unaffected (covered by #5631).

The actual fix (needs a SiteGround Speed Optimizer env, or equivalent "defer all JS"):
3. Enable "Defer Render-Blocking JS" + Apple Pay/Google Pay, Blocks checkout, guest/incognito → the Checkout Block renders and card fields appear (previously the whole block was blank). Confirm no `window.wc`/`wcBlocksRegistry`/`wp.data` undefined errors abort our bundle.
4. Repeat with Pixel Manager for WooCommerce deactivated → still renders.

---

-   [x] Covered with tests (Jest for the bootstrap gating; manual verification noted above for the optimizer path)

### Changelog entry

Fix - Render the Blocks checkout when a host optimizer defers render-blocking JavaScript (e.g. SiteGround Speed Optimizer)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 xxxx-xx-xx - version 10.9.0
 * Fix - Render classic-checkout card fields when a host optimizer defers render-blocking JavaScript (e.g. SiteGround Speed Optimizer) by waiting for required WordPress/WooCommerce globals before initializing the Stripe bundle
+* Fix - Render the Blocks checkout when a host optimizer defers render-blocking JavaScript (e.g. SiteGround Speed Optimizer) by waiting for required WordPress/WooCommerce globals before initializing the Stripe Blocks bundle
 * Fix - Use the Amazon Pay custom button size setting on the product, cart, and checkout pages instead of falling back to the Apple Pay/Google Pay size
 * Tweak - Render the Express Checkout button on the cart and checkout from page-bootstrapped data, removing a cart-details request from the critical path to first button render
 * Fix - Scope admin gateway filter to block cart and checkout editors only
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,7 +2,7 @@
 
 xxxx-xx-xx - version 10.9.0
 * Fix - Render classic-checkout card fields when a host optimizer defers render-blocking JavaScript (e.g. SiteGround Speed Optimizer) by waiting for required WordPress/WooCommerce globals before initializing the Stripe bundle
-* Fix - Render the Blocks checkout when a host optimizer defers render-blocking JavaScript (e.g. SiteGround Speed Optimizer) by waiting for required WordPress/WooCommerce globals before initializing the Stripe Blocks bundle
+* Fix - Render the Blocks checkout when a host optimizer defers render-blocking JavaScript (e.g. SiteGround Speed Optimizer)
 * Fix - Use the Amazon Pay custom button size setting on the product, cart, and checkout pages instead of falling back to the Apple Pay/Google Pay size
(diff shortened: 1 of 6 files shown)


wjrosa:
Closing this. Merchant testing showed the Blocks fix doesn't resolve the issue: under SiteGround Speed Optimizer's "Defer Render-Blocking JS", the entire WordPress/WooCommerce Blocks script stack (React/`@wordpress/element`, `@wordpress/data`, `wc-blocks-*`) is reordered and broken (`createContext`/`forwardRef`/`combineReducers`/`createReduxStore is not a function`), so the Checkout Block can't mount regardless of our bundle. The same errors appear on the stock 10.8.3 release; our split only moved our own (downstream) throw.

The root cause is the optimizer breaking WordPress core script ordering, which we can't fix from the plugin. The classic-checkout fix (#5631) still stands and is verified by the merchant. Tracking the Blocks case as a host-optimizer issue instead.

PR #5645, opened 2026-06-26
Filter country-restricted payment methods out of Optimized Checkout

### LINKED ISSUE #5644: Optimized Checkout offers country-restricted methods (e.g. iDEAL) that fail at payment for ineligible billing countries
## Description

With Optimized Checkout (OCS) enabled, the dynamic Stripe Payment Element offers payment methods based on the PaymentIntent **currency**, independent of the shopper's billing country. Country-restricted methods can therefore be shown to shoppers whose billing country can't use them, and the payment then fails at confirmation with:

> There was an error processing the payment: This payment method type is not available in the selected country.

This is a poor UX: the method is offered, the shopper selects it, fills the form, and only fails on submit.

## Steps to reproduce

1. Set the store currency to **EUR** and enable **iDEAL** (a Netherlands-only method) for checkout.
2. Enable Optimized Checkout.
3. Go to checkout with a **United States** billing address.
4. The dynamic Payment Element lists **iDEAL**. Select it and submit.

**Expected:** iDEAL is not offered for a US billing address (it isn't usable there).
**Actual:** iDEAL is offered, and the payment fails at confirmation with "not available in the selected country."

Reproduces on both classic and Blocks checkout. Confirmed pre-existing (not specific to any in-flight refactor).

## Root cause

The OCS element's `excludedPaymentMethodTypes` only ever excluded plugin-unsupported methods and Amazon Pay — never methods that are unavailable for the current billing country. The per-method country filtering that the non-OCS layouts apply (`togglePaymentMethodForCountry`) does not apply to the single collapsed OCS element.

## Affected surfaces

Classic checkout, Blocks checkout, and Optimized Checkout. The Adaptive Pricing (Checkout Sessions) flow uses `initCheckout()` and would need separate handling for in-page country changes.

---

wjrosa:
Fixes #5644
Fixes STRIPE-1245

With Optimized Checkout (OCS) enabled, the dynamic Stripe Payment Element offers methods based on the PaymentIntent **currency**, independent of billing country. A country-restricted method (e.g. iDEAL, which is NL-only) could therefore be offered to a shopper with an ineligible billing address (e.g. US) and then fail at confirmation with *"This payment method type is not available in the selected country."* The per-method country filter the non-OCS layouts use (`togglePaymentMethodForCountry`) does not apply to the single collapsed OCS element.

This PR excludes methods the current billing country can't use from the OCS element's `excludedPaymentMethodTypes`:

- **Server (load time):** `get_excluded_payment_method_types()` now also excludes enabled methods whose `get_available_billing_countries()` list doesn't include the shopper's billing country — the same source `togglePaymentMethodForCountry` uses on non-OCS layouts, so it also covers domestic-only methods (Affirm, Afterpay) and Klarna's per-account/currency matrix. Covers classic, Blocks, and OCS at page load.
- **Client (reactive):** a `countriesByMethod` map is exposed on the OCS config entry (the per-method `countries` is otherwise hidden once OCS collapses methods into one entry). On an in-page billing-country change, the classic (`updated_checkout`) and Blocks (`useEffect`) flows recompute the exclusion and push it with `elements.update({ excludedPaymentMethodTypes })`. `excludedPaymentMethodTypes` is an updatable Elements option.

**How can this break?** The Adaptive Pricing (Checkout Sessions) flow uses `initCheckout()`, which has no `elements.update()`, so it relies on the server load-time exclusion only and won't react to an in-page country change. Server seed and client recompute both read `get_available_billing_countries()`, so they can't disagree (an earlier asymmetry flagged in review was fixed by aligning the server on that source).

**What keeps it from breaking?** New PHPUnit coverage for the server country logic and Jest coverage for the client helper; full existing PHPUnit (UPE gateway class) and Jest suites pass; PHPStan and lint clean.

## Testing instructions

1. Set the store currency to **EUR** and enable **iDEAL** for checkout.
2. Enable Optimized Checkout.
3. Go to checkout with a **United States** billing address.
4. Confirm **iDEAL is not listed** in the Payment Element (classic and Blocks).
5. Change the billing country to **Netherlands** — iDEAL appears. Change back to **US** — it disappears. No page reload required.
6. Complete a card payment to confirm nothing else regressed.

Also verify a normal EUR/NL iDEAL checkout still succeeds, and that disabling OCS leaves the existing per-method behavior unchanged.

---

### Changelog entry

Fix - Hide country-restricted payment methods such as iDEAL from Optimized Checkout when the billing country can't use them, preventing a "not available in the selected country" failure at payment


Mayisha (commented):
Tested for iDeal and it is working as described.

Got a warning in AI review which is worth checking.

AI review:
Client and server use different country sources — they diverge for domestic-restriction methods

The server's exclusion path uses `is_available_for_billing_country()`, which reads supported_billing_countries:

// class-wc-stripe-upe-payment-gateway.php — 
```
get_country_restricted_excluded_payment_method_types()
if ( ... && ! $payment_method->is_available_for_billing_country( $billing_country ) ) { ... }
```

But the client `countriesByMethod` map is built from `get_available_billing_countries()`:

```
$method_countries = $payment_method->get_available_billing_countries();
```

Those two functions return different things for methods with `accept_only_domestic_payment = true`. 
`get_available_billing_countries()` returns `[ $account_country ]` for domestic methods, while `is_available_for_billing_country()` reads `supported_billing_countries`.

Concrete case — Affirm (supported_billing_countries = ['US','CA'], accept_only_domestic_payment = true), US Stripe account, shopper with CA billing:
- Server (is_available_for_billing_country('CA')) → true → Affirm not excluded.
- Client (countriesByMethod['affirm'] = ['US'], !['US'].includes('CA')) → Affirm excluded.

So on OC, Affirm/Afterpay get hidden for the CA shopper, while the server's own base computation and classic non-OC checkout show them — and the client's elements.update() overrides the server-seeded value, so this is user-visible. The JS docblock claims it "mirrors the server get_excluded_payment_method_types()," which isn't true for these methods.

wjrosa:
Thanks for reviewing and for the findings, Mayisha! I fixed the issue in 540fc43. It aligns the server on `get_available_billing_countries()` rather than loosening the client, which would have reintroduced the confirm-time failure for domestic-only methods.

Mayisha (changes requested):
Thanks @wjrosa.

Found an issue when testing with the iDeal. Looks like the client side recompute is append-only, so a hidden method never re-surfaces.

To reproduce-
- Set the store currency to Euro.
- Enabled iDeal and Card.
- Enablle OCS.
- As a shopper, go to the checkout page. Make sure when you land on the checkout page the billing address is US or any other country but NL.  
- The Card method should be present only. 
- Change the billing address to NL. Change it using the Stripe developer tools as well to simulate a real geo location.
- Notice that iDeal does not appear on the page.
- Refresh the page or as a shopper go to the checkout page where NL is selected as the billing country. You'll see iDeal as an option.

I got some other minor feedback from AI.

- **Missing coverage for the reactive paths:** No Jest coverage for the new runtime behavior — the Blocks `useEffect` in `payment-processor.js` and classic `maybeUpdateOptimizedCheckoutExclusions()`. Please cover the guard conditions (OC off, elements.update missing) and that update() is called with the country-derived exclusions on a country change. Only the pure helper + PHP path are tested right now.
- **Pay-for-order seeds from the wrong country:** `get_billing_country_for_checkout()` reads only `WC()->customer`, which for a guest pay-for-order can differ from the order's billing country → wrong server seed. Fixing the blocker lets the client self-correct on classic/Blocks, but prefer the order's billing country here regardless.
- **Efficiency (nits)**
    - `maybeUpdateOptimizedCheckoutExclusions()` fires on every `updated_checkout` (coupon, shipping, qty — not just country) and calls `elements.update()` unconditionally. Memoize the last-applied (sorted/serialized) list and skip when unchanged — avoids redundant Stripe calls and flicker.
    - The Blocks `useEffect` depends on the whole `stripeServerData` object; narrow it to `stripeServerData?.shouldShowOptimizedCheckout` so it doesn't re-fire on unrelated re-renders.

wjrosa:
Thanks for flagging it, @Mayisha! I fixed the issue in f9e264d: methods present in `countriesByMethod` are now dropped from the seeded base and re-derived for the current country, so they re-surface when the country allows them.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -837,6 +846,58 @@ function ( $method ) use ( $non_excludable_methods ) {
 		return array_values( array_unique( $excluded_methods ) );
 	}
 
+	/**
+	 * Returns enabled-at-checkout methods unavailable for the given billing country.
+	 *
+	 * @param string $billing_country Two-letter ISO billing country, or empty when unknown.
+	 * @return string[] Payment method types unavailable in the given country.
+	 */
+	protected function get_country_restricted_excluded_payment_method_types( string $billing_country ): array {
```
jaclync: non-blocking nit: could this function be `private` instead of `protected` if we don't intend to make it extensible? The only caller is in the same class, and the unit tests already work with `ReflectionMethod`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -837,6 +846,58 @@ function ( $method ) use ( $non_excludable_methods ) {
...
+			$order = wc_get_order( absint( get_query_var( 'order-pay' ) ) );
+
+			if ( $order instanceof WC_Order ) {
+				return (string) $order->get_billing_country();
+			}
+		}
+
+		$customer = WC()->customer;
+
+		return $customer instanceof WC_Customer ? (string) $customer->get_billing_country() : '';
```
jaclync, line 938: With OC enabled, a potential regression in the Cart/Checkout block editor preview: `javascript_params()` runs in wp-admin where `WC()->customer` is uninitialized, so the billing country resolves to `''` and every country-restricted method gets included in the seeded `excludedPaymentMethodTypes`. The editor has no recompute path, thus the preview shows only unrestricted methods (card, Link) now. On `develop`, it shows all enabled methods.

Repro steps: with a EUR store, iDEAL/Bancontact/Klarna enabled, OC on
- In wp-admin, Pages > Checkout
- Edit the Checkout page in the block editor and preview --> the Payment Element preview lists only cards, different from `develop`

Maybe the country-restriction filtering can be skipped when editing the cart/checkout block (or whenever the customer object is unavailable), in case there's no recompute with the updated billing address later.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -837,6 +846,58 @@ function ( $method ) use ( $non_excludable_methods ) {
...
+		return $excluded;
+	}
+
+	/**
+	 * Returns the shopper's billing country for checkout-time method gating.
+	 * Protected so tests can supply a deterministic country.
+	 *
+	 * @return string Two-letter ISO billing country, or empty when unknown.
+	 */
+	protected function get_billing_country_for_checkout(): string {
```
jaclync, line 924: non-blocking: the logic here differs from the billing country set in `javascript_params()` for `customerData`, which could result in `get_excluded_payment_method_types` not using the billing address the same way `customerData` is set. Could the billing address logic be consolidated here for both call sites?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -829,6 +829,15 @@ function ( $method ) use ( $non_excludable_methods ) {
...
 		);
 
+		// Hide methods the billing country can't use; recomputed on the frontend as it changes.
+		$country_excluded_methods = array_filter(
+			$this->get_country_restricted_excluded_payment_method_types( $this->get_billing_country_for_checkout() ),
+			function ( $method ) use ( $non_excludable_methods ) {
+				return ! in_array( $method, $non_excludable_methods, true );
+			}
+		);
+		$excluded_methods         = array_merge( $excluded_methods, $country_excluded_methods );
```
jaclync: 💭 On the Adaptive Pricing flow this country filter becomes permanent, and can now hide a valid method as `excluded_payment_method_types` [can't be updated for the Checkout Session](https://docs.stripe.com/api/checkout/sessions/update).

Example scenario: if the session guessed US and the shopper changes the billing address to NL, iDEAL stays hidden until the next page reload.

Do you think this could be an issue?

client/classic/upe/payment-processing.js
```diff
@@ -1024,6 +1049,38 @@ export function getMountedUPEComponent( paymentMethodType ) {
...
+	// quantity); skip the Stripe round-trip when the exclusions are unchanged.
+	const exclusionsKey = getExclusionsKey( excludedPaymentMethodTypes );
+	if (
+		appliedOptimizedCheckoutExclusions.get( elements ) === exclusionsKey
+	) {
+		return;
+	}
+	appliedOptimizedCheckoutExclusions.set( elements, exclusionsKey );
+
+	elements.update( { excludedPaymentMethodTypes } );
```
jaclync: non-blocking nit: the memo is recorded before `elements.update()` and if the update ever throws, the memo is still written and read later. Perhaps the order of these two lines could be swapped.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
```diff
@@ -337,6 +340,31 @@ const PaymentProcessor = ( {
...
+					billing?.billingAddress?.country || ''
+				),
+		} );
+		// Depend on the primitive flag actually read, not the whole config
+		// object, so a changed config identity can't re-fire the Stripe update.
+	}, [
+		elements,
+		billing?.billingAddress?.country,
+		stripeServerData?.shouldShowOptimizedCheckout,
+	] );
```
jaclync, line 370: non-blocking question: there are differences between blocks (here) and classic checkout, do you think it's worth consolidating the recompute like in the classic path's `maybeUpdateOptimizedCheckoutExclusions`? Or are these differences intentional?

Two differences vs. the classic path:

1. Creation: classic now creates the element `getExcludedPaymentMethodTypesForBillingCountry`, while blocks creation still uses `getExcludedPaymentMethodTypes()`.
2. Updates: classic uses a memo to prevent duplicate updates, while the blocks path here doesn't.

client/stripe-utils/utils.js
```diff
@@ -787,6 +787,46 @@ export const getExcludedPaymentMethodTypes = () => {
...
+		( method ) =>
+			method === PAYMENT_METHOD_AMAZON_PAY ||
+			! isCountryRestricted( countriesByMethod[ method ] )
+	);
+
+	Object.entries( countriesByMethod ).forEach( ( [ method, countries ] ) => {
+		// Empty list = no restriction; an unknown country can't confirm a restricted method.
+		if (
+			isCountryRestricted( countries ) &&
+			! countries.includes( billingCountry )
```
jaclync: non-blocking question: could the casing differ between these two values?

client/stripe-utils/utils.js
```diff
@@ -787,6 +787,46 @@ export const getExcludedPaymentMethodTypes = () => {
...
+		) {
+			excluded.push( method );
+		}
+	} );
```
jaclync, line 827: non-blocking suggestion on an edge scenario: this recompute can silently override two public hooks for the exclusion list:

1. `wc_stripe_upe_params` (filters the whole JS config including `excludedPaymentMethodTypes`). If a third party appends a method here for a non-country reason (e.g. hide Klarna), this function drops that entry and the re-add path skips it if the billing country is supported. A shopper therefore sees Klarna again despite the filter.
2. `wc_stripe_ocs_non_excludable_payment_methods` which lets a merchant mark methods that must never be excluded. The server honors it when building the seed, but this list doesn't reach the client and `countriesByMethod` is built without it. The recompute can push the protected method from the filter back into `excludedPaymentMethodTypes`.

These edge cases exist because the client can't tell why a seed entry exists. Maybe the country-agnostic exclusion list can be sent to the client alongside `countriesByMethod`, so that the client can apply the country-based check on the base list.

jaclync (commented):
Shared some initial comments, mostly non-blocking.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Done in d0d22edb0.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Good catch — fixed in d0d22edb0: the server skips country exclusions in wp-admin and the blocks recompute bails in the editor, restoring the develop preview. Added a regression test.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Consolidated the add-payment-method site onto `get_billing_country_for_checkout()` in d0d22edb0 (also fixes an unguarded `WC()->customer` there). Left the order-pay branch as is: it also serves subscriptions change-payment, which the helper deliberately doesn't handle.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: It's real, but I think it's the safer failure mode: a hidden-but-valid method costs a reload, while an offered-but-invalid one fails the payment at confirm. Since the session can't update exclusions, there's no better option at creation time — happy to revisit if it shows up in practice.

client/classic/upe/payment-processing.js
wjrosa: Swapped in d0d22edb0.

client/blocks/upe/upe-deferred-intent-creation/payment-processor.js
wjrosa: Both converge in practice: the blocks creation seed already bakes in the page-load country server-side, and the effect's deps (billing country) act as the memo — a same-country re-render doesn't refire the update. So I kept them as is.

client/stripe-utils/utils.js
wjrosa: It could in principle — normalized to uppercase in d0d22edb0.

client/stripe-utils/utils.js
wjrosa: Implemented your suggestion in d0d22edb0: the server now exposes the country-derived portion separately (`countryExcludedPaymentMethodTypes`) and the recompute subtracts exactly that from the seed, so `wc_stripe_upe_params` additions survive; `countriesByMethod` now omits non-excludable methods so the client can never re-add them. Both scenarios covered by tests.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -819,22 +851,87 @@ public function get_excluded_payment_method_types(): array {
...
+
+	/**
+	 * Returns the country-derived portion of the excluded payment method types:
+	 * methods the current billing country can't use. Recomputed on the frontend
+	 * as the country changes; also exposed to the client on its own so the
+	 * recompute can subtract exactly this portion from the seed.
+	 *
+	 * @return string[] Payment method types excluded for the current billing country.
+	 */
+	public function get_country_excluded_payment_method_types(): array {
```
jaclync: nit: this could be private if it's not meant to be extensible?

jaclync (approved):
Thanks for the updates, LGTM! :shipit: Tested well, the AP path had the expected acceptable behavior where the excluded payment methods aren't updated after the billing country changes.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Good idea! Done in ebfeb9ee3

diegocurbelo (approved):
The code changes look good, and it tests well! 👍🏼 

> On the Adaptive Pricing flow this country filter becomes permanent..

I agree with the reasoning for this (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5645#discussion_r3657008497), but now we are "hiding a method the shopper might be eligible for" instead of "offering it and then show an error when placing the order"; we should add an explicit comment block in `class-wc-stripe-upe-payment-gateway.php` -> `$excluded_methods` explaining this restriction/decision.

wjrosa:
Thanks for the review, @diegocurbelo !

> I agree with the reasoning for this (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5645#discussion_r3657008497), but now we are "hiding a method the shopper might be eligible for" instead of "offering it and then show an error when placing the order"; we should add an explicit comment block in class-wc-stripe-upe-payment-gateway.php -> $excluded_methods explaining this restriction/decision.

That's fair! Added the comment block at the `$excluded_methods` merge in 60eeb59f8

PR #5649, opened 2026-06-29
Fix express checkout modal breakdown showing the previously selected variation

### LINKED ISSUE #2718: Apple/Google Pay not retriving attribute data in product page (eg: size)
**Describe the bug**

Apple/Google Pay is not retrieving the variation attribute data on product pages which have variation lists in the Related products ex: [https://d.pr/i/KbMfnw](<https://d.pr/i/KbMfnw>) .

This happens because of this selector [https://github.com/woocommerce/woocommerce-gateway-stripe/blob/c32219dc2969c377f1a92d2601f57b5c8ba4f0d8/assets/js/stripe-payment-request.js#L42](<https://github.com/woocommerce/woocommerce-gateway-stripe/blob/c32219dc2969c377f1a92d2601f57b5c8ba4f0d8/assets/js/stripe-payment-request.js#L42>) which tries pulling all `.variation_form` select lists.

For example in this case this selector `$( '.variations_form' ).find( '.variations select' )` will return 5 results (the main product Size select list + the 4 related products Size select list) ex: [https://d.pr/i/rvAibQ](<https://d.pr/i/rvAibQ>) .

Therefore the data within the `get_selected_product_data` Ajax endpoint will grab the last queried select list's value, which is empty (because it's unselected).

The fix would be to ensure the queried variation list is the main product page's one.

A solution I foresee now would be something like:

```
var pdpID = '';
var pdpIDAttr = $('div.product').attr('id'); // get product page wrapper ID attribute. Usually it's something like `product-9999` .

if( pdpIDAttr ){
	var getPdpID = pdpIDAttr.split('-'); // split it to retrieve the ID.
	if( getPdpID.length > 1 ){
		pdpID = '[data-product_id="'+ getPdpID[1] +'"]'; // Get the unique selector 
	}
}

// Query the main product's select list
// ex: .variations_form[data-product_id="9999"]
var select = $( '.variations_form' + pdpID ).find( '.variations select' ); 
```

**To Reproduce**<br>Ideally would be to test with some Variation Swatches plugin which is adding support for variations in the product listings.

PS: I didn't include other information because I don't believe they're relevant to this case, however I will provide if needed.

Mayisha:
Fixes STRIPE-174
Fixes #2718 

On a variable product page, selecting a different variation updated the Apple Pay/Google Pay sheet **total** correctly but left the **line-item breakdown** showing the previously selected (or default) variation. For a product priced red $10 / blue $20 / green $30, picking blue showed a $20 total with a $10 "red" line item.

Root cause is client-side. On `woocommerce_variation_has_changed` (and quantity change), the fast-path calls `elements.update({ amount })`, which refreshes only the Stripe element's amount. The wallet breakdown is built at `event.resolve()` time from the line items captured in the click closure when the button was first rendered, and that fast-path never refreshes them — only a full re-init did. The server (`ajax_get_selected_product_data`) already returns the correct line items for the selected variation; they were simply being discarded.

This PR:
- `resolveClickEvent` now reads the latest line items from the product store, scoped to the legacy (variable/booking) display-item format where the staleness occurs and the format matches `normalizeLineItems`. Other flows keep using the captured items.
- Both variation change and quantity change now write the fresh `total`/`displayItems`  back to the product store, so the live read returns the current selection.

## Testing instructions
- Create a variable product with 3 variations (example: red: $10, blue: $20, green: $30)
- Add tax and shipping rates for your store.
- Enable Google/Apple Pay.
- As a shopper, go to the variable product page. Wait for the ECE buttons to load.
- Now select the 2nd variation blue and wait for the ECE buttons to refresh.
- Click the Google Pay button and check the line items breakdown.
- In develop branch, notice that the total amount is correct for the selected variation but the breakdown (variation, tax, shipping etc.) is wrong. The total is for the variation blue but the breakdown shows data for red.

Screenshot 2026-06-30 at 1 09 38 AM

- In this branch, both breakdown and total should be correct.

Screenshot 2026-06-30 at 12 35 18 AM

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


mahangu:
@claude /review

mahangu:
Sorry about the `@claude /review` comments. Was debugging Woo AI reviews. Should be fixed now.

client/entrypoints/express-checkout/index.js
```diff
@@ -909,6 +918,12 @@ jQuery( function ( $ ) {
 										response.requestShipping;
 
 								if ( ! isDeposits && needsShipping ) {
+									// Refresh stored items so the click breakdown matches this variation.
```
wjrosa, line 930: I would consider creating something for reuse, either for both lines (like `refreshTotals`) or just  encapsulate `getExpressCheckoutData( 'product' )` (like `expressCheckoutProduct`)

wjrosa (changes requested):
Code looks good to me. But I am not able to properly test this. When I switch variations with different prices, I get the following error in the console:
```
Uncaught (in promise) IntegrationError: The amount 1500 is less than the total amount of the line items provided.
```
After it shows up, clicking the ECE buttons does nothing

Mayisha:
Thanks for catching this @wjrosa. I could reproduced it by enabling Link along with Google and Apple Pay.

**Cause**: a product page mounts a separate Elements group per express button (Apple Pay, Google Pay, Link…), but `attachProductPageEventListeners` runs once per button and rebinds the shared `woocommerce_variation_has_changed` / `.qty` handlers with `$(document.body).off(...).on(...)`. The `.off()` wipes the earlier buttons' handlers, so on a variation change only the **last-created Elements group** actually got `elements.update({ amount })` — the others kept the amount from the initially-selected variation.

That was harmless before this PR because the click still resolved the stale (initial) line items, so a non-updated button had a stale amount and stale line items — internally consistent. This PR fixes the breakdown by reading line items live from the shared product store at click time, which exposed the latent bug: the non-updated button now pairs the initial variation's amount (your 1500) with the newly selected variation's line items, and Stripe rejects the click with amount … is less than the total amount of the line items. It only shows up when 2+ express buttons are mounted, which is why single-button setups don't hit it.

**Fix**: track every mounted Elements group in a small registry and push the new amount to all of them on variation/qty change, instead of only the last one. Added a regression test that mounts multiple buttons and asserts each group's amount is updated (verified it fails when only the last is updated). Fixed in [0ecff9d](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5649/commits/0ecff9d6cb632eb507b08dd02bbc3ae03e0abace)

wjrosa (approved):
Looks good to me and works as expected 🎉

PR #5657, opened 2026-07-01
Add shareable woocommerce_agentic_commerce_disable_checkout filter (Agentic Commerce) [1/3]

wjrosa:
Towards woocommerce/woocommerce-gateway-stripe#5656

Part 1 of 3 splitting woocommerce/woocommerce-gateway-stripe#5655 (superseded) into reviewable, stacked PRs.

Stack:

1. **This PR** — shareable `disable_checkout` filter (base `develop`)
2. woocommerce/woocommerce-gateway-stripe#5658 — shipping-zone diagnostics
3. [#5659](<https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5659>) — add-on/configurator detection + toggles

## Changes proposed in this Pull Request:

Introduces the canonical `woocommerce_agentic_commerce_disable_checkout` filter and deprecates the Stripe-prefixed `wc_stripe_agentic_commerce_disable_checkout` twin via `apply_filters_deprecated`, mirroring the `should_sync_product` migration in woocommerce/woocommerce-gateway-stripe#5598. Existing hooks on the old name still run — they seed the new filter's default — but emit a deprecation notice.

## Testing instructions

1. Add a `woocommerce_agentic_commerce_disable_checkout` filter returning `true`; confirm the product's `disable_checkout` is `true` in the feed / preview.
2. Add a filter on the old `wc_stripe_agentic_commerce_disable_checkout` name; confirm it still applies and a deprecation notice is logged.
3. `npm run test:php -- --filter WC_Stripe_Agentic_Commerce_Product_Mapper_Test`.

---

- [X] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

### Changelog entry

Added under the unreleased 10.9.0 block.

**Post merge**

- [ ] Added testing instructions to the [Release Testing Instructions wiki page](<https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions>) (or does not apply)
- [ ] Added the [needs docs label](<https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs>) (or does not apply)
- [ ] Included this PR in the Release Thread scope (or does not apply)

🤖 Generated with [Claude Code](<https://claude.com/claude-code>)

Issue #5656, opened 2026-07-01
Agentic Commerce: merchant-config affordances for add-on/configurator catalogs and live-rate shipping


wjrosa:
**Is your feature request related to a problem? Please describe.**

Merchants who run product configurator plugins (WooCommerce Product Add-Ons, TM Extra Product Options, Composite Products, individually-priced Bundles) alongside live-rate shipping (e.g. ShipStation Live Rates) can't join the Agentic Commerce pilot without hand-written `mu-plugin` code:

- **Configurator SKUs** carry runtime-variable pricing the static feed can't represent, and the order-creation path rejects any line whose live price drifts from what Stripe charged. Today the only way to keep them out of the feed (or redirect shoppers to the store) is a custom `disable_checkout` / `should_sync_product` filter.
- **Live-rate shipping** methods leave a zone with no feed-representable shipping, and there's no signal to the merchant that this happened — the shipping column silently comes out empty.

**Describe the solution you'd like**

Make these merchant-configurable and self-diagnosable, without custom code:

- Built-in add-on/configurator detection with two opt-in toggles (both default off): auto-exclude detected products from the feed, or keep them discoverable but auto-redirect (`disable_checkout`). Both are defaults a custom filter can still override.
- A shareable `woocommerce_agentic_commerce_disable_checkout` filter (deprecating the `wc_stripe_`-prefixed twin), plus a `woocommerce_agentic_commerce_addon_detection_meta_keys` filter for plugins not covered out of the box.
- Shipping-zone diagnostics in the logs (methods skipped for no flat rate; zones with no flat-rate fallback) and feed-preview granularity (`shipping_warnings`, per-product `advisories`) so merchants can self-diagnose in WooCommerce.

**Describe alternatives you've considered**

Documenting the required custom filters for merchants to paste into an `mu-plugin`. This works but is error-prone and gates the pilot on developer availability.

**Additional context**

Implemented in #5655.

PR #5655, opened 2026-06-30
Add merchant-config affordances for add-on/configurator catalogs (Agentic Commerce)

wjrosa:
Closes #5656.

## Changes proposed in this Pull Request:

Lets merchants running configurator plugins (Product Add-Ons, TM Extra Product Options, Composite Products, individually-priced Bundles) and live-rate shipping (e.g. ShipStation Live Rates) join the Agentic Commerce pilot **without custom mu-plugin code**:

1. **Shareable redirect filter.** Introduces the canonical `woocommerce_agentic_commerce_disable_checkout` and deprecates the `wc_stripe_`-prefixed twin via `apply_filters_deprecated` (same migration as `should_sync_product`, #5598).
2. **Add-on/configurator detection + two opt-in toggles (default off).** `product_has_addons()` inspects per-product postmeta (`_product_addons`, `tm_meta_cpf_options`, `composite_data`, and `_wc_pb_priced_individually=yes` only), resolved on the parent for variations and extensible via `woocommerce_agentic_commerce_addon_detection_meta_keys`. The toggles auto-exclude detected products from the feed, or auto-redirect them. Both are defaults a custom filter still overrides. Detection does not gate on `class_exists()` (an active plugin doesn't mean a given product is configured).
3. **Shipping diagnostics.** `get_shipping()` logs methods skipped for no flat cost and warns on zones with no flat-rate fallback (the live-rate-only case).
4. **Feed-preview granularity.** Adds `shipping_warnings` and per-product `advisories` (`no_sku`, `excluded` with reason, `disable_checkout` with source).
5. **Docs.** README "Merchant configuration cookbook" and an `AGENTS.md` invariants note.

New options/filters/helpers are additive and gated behind toggles that default off, so an upgrade with no new options set behaves identically. The filter rename keeps the old hook working with a deprecation notice.

## Testing instructions

Prerequisites: enable the developer feature flag (`_wcstripe_feature_agentic_commerce`) and the merchant toggle (**Stripe settings → Agentic commerce → Enable agentic commerce**).

1. **Add-on auto-exclude.** Add Product Add-Ons (or `_product_addons` meta) to a product, enable **"Exclude products with add-ons or configurators from the feed"**. In **Preview feed** the product appears under advisories as `excluded` (reason `addons`) and drops from the included count; disabling returns it.
[...]

PR #5658, opened 2026-07-01
Add shipping-zone diagnostics for the Agentic Commerce feed [2/3]

wjrosa:
Towards STRIPE-1249
Towards #5656

Part 2 of 3 splitting #5655 (superseded) into reviewable, stacked PRs.

Stack:
1. #5657 — shareable `disable_checkout` filter (merged)
2. **This PR** — shipping-zone diagnostics (base: `develop`, now that #5657 has merged)
3. #5659 — add-on/configurator detection + toggles

Only shipping methods with a static numeric cost (flat rate, free shipping) can be written to the feed's `shipping` column; live-rate and calculated methods (e.g. ShipStation Live Rates) price at checkout and are dropped. When that leaves a zone with no feed-representable shipping, agents see no shipping for that destination — previously with no signal to the merchant.

- `get_shipping()` now logs each skipped method (info) and warns when a zone has no flat-rate fallback at all (warning), on the `wc-stripe` channel.
- New `get_shipping_diagnostics()` exposes the zones-without-flat-rate list without emitting logs, so the feed preview surfaces them as a `shipping_warnings` array.
- The built-in catch-all zone 0 ("Locations not covered by your other zones") is included in both the feed entries and the diagnostics alongside named zones — so a store whose uncovered-locations zone has no flat-rate fallback is surfaced too, not only when the store has no named zones at all.

**BC impact:** `get_shipping_diagnostics()` is a new public method on the concrete `WC_Stripe_Agentic_Commerce_Product_Mapper`, not on `ProductMapperInterface`, so it is additive: existing mapper implementations and subclasses are unaffected. No hooks, option keys, or existing signatures change. The `shipping_warnings` key is a new addition to the preview REST response.

## Testing instructions

1. Create a shipping zone served only by a live-rate method (or any method with no flat cost) with no flat-rate fallback.
2. Run a feed sync → **WooCommerce → Status → Logs** (`wc-stripe`): confirm the "method omitted" info line and the "zone has no flat-rate method" warning.
3. **Stripe settings → Agentic commerce → Preview feed** → a warning notice lists the zone (the REST response carries it as `shipping_warnings`).
4. With named zones present, leave WooCommerce's "Locations not covered by your other zones" zone without a flat-rate method → it also appears in the logs/warnings (previously it was skipped whenever named zones existed).
5. `npm run test:php -- --filter WC_Stripe_Agentic_Commerce_Product_Mapper_Test` and `--filter WC_Stripe_Agentic_Commerce_Feed_Preview_Test`.

---

### Changelog entry

Added under the unreleased 10.9.0 block.

**Post merge**


includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -821,6 +844,59 @@ protected function get_shipping(): ?string {
...
+				}
+			}
+
+			if ( ! $has_flat_rate ) {
+				$zones_without_flat_rate[] = $zone_name;
+			}
+		}
+
+		return [ 'zones_without_flat_rate' => array_values( array_unique( $zones_without_flat_rate ) ) ];
+	}
```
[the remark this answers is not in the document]
wjrosa: Fixed in 2bfb9908c. `get_cached_shipping_zones()` now always appends the built-in zone 0 to the named-zone results instead of only building it when there are none, so its shipping (and missing-flat-rate warnings) flow through both the feed entries and diagnostics. Both consumers iterate that one method, so the single change covers all three sites. Added mapper + feed-preview tests asserting zone 0 is reported alongside a shipping-capable named zone (red without the fix).

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -249,6 +263,7 @@ public function generate( int $detail_limit = self::DEFAULT_DETAIL_LIMIT, int $s
 			'validation_errors'  => $validation_errors,
 			'truncated'          => $truncated,
 			'scan_limited'       => $scan_limited,
+			'shipping_warnings'  => $shipping_warnings,
```
diegocurbelo, line 266: `shipping_warnings` is added to the REST response, but `client/settings/agentic-commerce/feed-preview.js` does not read or render it. As a result, the Preview feed UI named in the description and manual step 3 cannot show these warnings.

includes/agentic-commerce/README.md
```diff
@@ -384,6 +384,26 @@ All of these are defaults rather than hard blocks: `woocommerce_agentic_commerce
...
+## Shipping diagnostics
+
+Only methods with a static numeric cost (flat rate, free shipping) are written
+to the feed's `shipping` column. Live-rate, calculated, and most third-party
+methods price at checkout and are omitted. When that leaves a zone with no
+shipping in the feed, agents see no shipping for that destination.
+
+This is discoverable two ways:
+
+- **Logs** (WooCommerce → Status → Logs, `wc-stripe` source):
```
diegocurbelo: The log source should be `woocommerce-gateway-stripe` instead of `wc-stripe`. 

Also, we should mention that the info and warning entries are disabled unless Stripe logging (or verbose mode) is enabled.

includes/agentic-commerce/README.md
```diff
@@ -384,6 +384,26 @@ All of these are defaults rather than hard blocks: `woocommerce_agentic_commerce
...
+  - *info* — "shipping method has no flat rate and was omitted from the feed"
+    (with the zone, method id, and method title), once per sync.
+  - *warning* — "shipping zone has no flat-rate method; it contributes no
+    shipping options to the feed" (with the zone name).
+- **Feed preview** (**Stripe settings → Agentic commerce → Preview feed**): the
+  `shipping_warnings` array names every zone with no flat-rate fallback.
+
+**Recommended fix:** add a low-cost or representative flat-rate method to each
```
diegocurbelo, line 406: Should we add a note clarifying that the flat-rate method recommended is not feed-only (it would be offered during real checkout)?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -821,6 +844,59 @@ protected function get_shipping(): ?string {
 		return $this->cached_shipping_data;
 	}
 
+	/**
+	 * The flat numeric cost a shipping method contributes to the feed's
+	 * `shipping` column, or null when it has no static cost (live-rate /
+	 * calculated methods price at checkout and can't be a fixed feed value).
+	 *
+	 * @since 10.9.0
```
diegocurbelo:
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -821,6 +844,59 @@ protected function get_shipping(): ?string {
...
+
+		return null;
+	}
+
+	/**
+	 * Shipping diagnostics for the feed preview: zones that contribute no
+	 * flat-rate shipping. Mirrors the skip logic in {@see self::get_shipping()}
+	 * without emitting log lines.
+	 *
+	 * @since 10.9.0
```
diegocurbelo:
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -984,32 +1060,27 @@ private function get_cached_shipping_zones(): array {
...
+		if ( $generic_zone instanceof \WC_Shipping_Zone ) {
+			$zones[] = [
+				'zone_name'        => __( 'Locations not covered by your other zones', 'woocommerce-gateway-stripe' ),
 				'zone_locations'   => [
 					(object) [
 						'type' => 'not_covered',
 						'code' => '',
 					],
 				],
 				'shipping_methods' => $generic_zone->get_shipping_methods(),
```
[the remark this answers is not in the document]
wjrosa: @claude addressed in 75fceb19: zone 0 now uses `get_shipping_methods( true )`, plus a test with a disabled flat rate on zone 0.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -984,32 +1060,27 @@ private function get_cached_shipping_zones(): array {
 			return $this->cached_shipping_zones;
 		}
 
-		// Get the main zones.
-		$this->cached_shipping_zones = \WC_Shipping_Zones::get_zones();
-
-		if ( ! empty( $this->cached_shipping_zones ) ) {
-			return $this->cached_shipping_zones;
-		}
+		$zones = \WC_Shipping_Zones::get_zones();
```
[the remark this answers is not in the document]
wjrosa: Fixed in 75fceb19: the appended zone 0 lists enabled methods only, like `get_zones()` already does for named zones, with a test for a disabled flat rate.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
wjrosa: Good catch. 75fceb19 renders them as a warning notice in the preview, with Jest coverage.

includes/agentic-commerce/README.md
wjrosa: Fixed in 75fceb19: source is `woocommerce-gateway-stripe`, and the README now says the entries need Stripe logging or verbose mode.

includes/agentic-commerce/README.md
wjrosa: Added in 75fceb19: the note says the flat rate is offered at checkout too, so the cost should be one the merchant will honour.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -152,6 +153,19 @@ public function generate( int $detail_limit = self::DEFAULT_DETAIL_LIMIT, int $s
...
 		$validation_errors      = [];
 
+		// Zones that contribute no flat-rate shipping to the feed. Computed once,
+		// independent of the product walk.
+		$shipping_warnings = [];
+		if ( $mapper instanceof WC_Stripe_Agentic_Commerce_Product_Mapper ) {
+			foreach ( $mapper->get_shipping_diagnostics()['zones_without_flat_rate'] as $zone_name ) {
+				$shipping_warnings[] = sprintf(
+					/* translators: %s: shipping zone name */
+					__( 'Shipping zone "%s" has no flat-rate method, so the feed carries no shipping for it (live-rate / calculated methods price at checkout).', 'woocommerce-gateway-stripe' ),
```
daledupreez: Can we deep-link to the problematic shipping zone configuration in wp-admin? That ensures that merchants can take action on the warning.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -759,23 +759,36 @@ protected function get_shipping(): ?string {
...
+					// a fixed feed value and is dropped; log the skip so missing
+					// shipping is diagnosable. Runs once per sync (memoised result).
+					WC_Stripe_Logger::info(
+						'Agentic Commerce: shipping method has no flat rate and was omitted from the feed.',
+						[
+							'zone'      => $zone_name,
+							'method_id' => $method->id,
+							'method'    => $method->get_title(),
+						]
+					);
```
daledupreez: Rather than logging an entry for every single shipping zone that doesn't include a flat rate, can we log one data point at the end of the loop? This could be _really_ noisy on some stores, and completely swamp the log file.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -821,6 +844,59 @@ protected function get_shipping(): ?string {
...
+	/**
+	 * The flat numeric cost a shipping method contributes to the feed's
+	 * `shipping` column, or null when it has no static cost (live-rate /
+	 * calculated methods price at checkout and can't be a fixed feed value).
+	 *
+	 * @since 11.1.0
+	 * @param \WC_Shipping_Method $method Shipping method instance.
+	 * @return string|null Formatted price (e.g. "5.00"), or null when not flat-rate.
+	 */
+	private function method_flat_cost( $method ): ?string {
```
daledupreez: Nit: can we give this a verb that describes what we are doing? e.g. `compute_shipping_method_flat_cost()`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -984,32 +1060,29 @@ private function get_cached_shipping_zones(): array {
...
 					(object) [
 						'type' => 'not_covered',
 						'code' => '',
 					],
 				],
-				'shipping_methods' => $generic_zone->get_shipping_methods(),
-			],
-		];
+				// Enabled-only, matching what get_zones() returns for named zones; a
+				// disabled flat rate must neither price the feed nor hide the warning.
```
daledupreez: Nit: The second part of this comment is not very helpful.
```suggestion
				// Get only enabled shipping methods, matching what get_zones() returns for named zones.
```

includes/agentic-commerce/README.md
```diff
@@ -384,6 +384,29 @@ All of these are defaults rather than hard blocks: `woocommerce_agentic_commerce
...
+live-rate-only zone so the catalog advertises a shipping price. That method is
+not feed-only: WooCommerce offers it at checkout next to the live rate, so pick
+a cost the merchant is willing to honour.
 ## Coupons and discounts
```
daledupreez, line 411: Nit: Add a newline before the next header.
```suggestion

## Coupons and discounts
```

daledupreez (commented):
I have some initial feedback based on inspection.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
wjrosa: Done — each preview warning now deep-links to that zone's shipping settings.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Done — one summary warning after the loop now lists all zones with no flat rate, and dropped the per-method info log.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Done — renamed to `compute_shipping_method_flat_cost()`.

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -214,6 +215,32 @@ const AgenticCommerceFeedPreview = () => {
...
+										<li key={ i }>
+											{ warning.edit_link ? (
+												<ExternalLink
+													href={ warning.edit_link }
+												>
+													{ warning.message }
+												</ExternalLink>
+											) : (
+												warning.message
+											) }
```
daledupreez: I have two concerns with this approach:
 * Wrapping the entire message in a link is going to result in it being harder to understand what the fix/action should
 * Using `ExternalLink` is a bit unexpected when the link is just to another part of wp-admin -- `ExternalLink` is generally used for links in other applications.
```suggestion
											{ warning }
											{ warning.edit_link && (
												<a href={ warning.edit_link }>
													{ __( 'Edit shipping zone' , 'woocommerce-gateway-stripe' ) }
												</a>
											) }
```

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -214,6 +215,32 @@ const AgenticCommerceFeedPreview = () => {
 							</Notice>
 						) }
 
+						{ shippingWarnings.length > 0 && (
+							<Notice status="warning" isDismissible={ false }>
```
daledupreez: I am also wondering whether we should make it easier to minimise this message, either by using something like a `<details>` element, or by making the notice dismissible. (Though any dismiss action should be temporary, not permanent or affecting other users.) If a merchant knows that some zones don't have flat rate options, but they are trying to debug something, these notices could get really annoying, especially if there are a lot of them.

client/settings/agentic-commerce/feed-preview.js
wjrosa: Agreed on both points. I could not take the suggestion as-is because `{ warning }` would render the object itself (each entry is `{ message, edit_link }`), so React would throw. Applied the same idea in aed0f7fab

client/settings/agentic-commerce/feed-preview.js
wjrosa: Good idea! I went with the details element in 1eebd6696

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -214,6 +215,53 @@ const AgenticCommerceFeedPreview = () => {
...
+											/* translators: %d: number of shipping zones the warning covers. */
+											_n(
+												'%d shipping zone will have no shipping in the feed:',
+												'%d shipping zones will have no shipping in the feed:',
+												shippingWarnings.length,
+												'woocommerce-gateway-stripe'
+											),
+											shippingWarnings.length
+										) }
+									</summary>
```
daledupreez, line 249: I think it would be worth making the summary bold so it stands out.
```suggestion
									<summary>
										<strong>
											{ sprintf(
												/* translators: %d: number of shipping zones the warning covers. */
												_n(
													'%d shipping zone will have no shipping in the feed:',
													'%d shipping zones will have no shipping in the feed:',
													shippingWarnings.length,
													'woocommerce-gateway-stripe'
												),
												shippingWarnings.length
											) }
										</strong>
									</summary>
```

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -214,6 +215,53 @@ const AgenticCommerceFeedPreview = () => {
...
+									<ul>
```
daledupreez: The list doesn't have any markers for each list item, which makes it harder to parse. Might it be worth reinstating the disc markers for this element?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -152,6 +153,30 @@ public function generate( int $detail_limit = self::DEFAULT_DETAIL_LIMIT, int $s
...
 
+		// Zones that contribute no flat-rate shipping to the feed. Computed once,
+		// independent of the product walk.
+		$shipping_warnings = [];
+		if ( $mapper instanceof WC_Stripe_Agentic_Commerce_Product_Mapper ) {
+			foreach ( $mapper->get_shipping_diagnostics()['zones_without_flat_rate'] as $zone ) {
+				$shipping_warnings[] = [
+					'message'   => sprintf(
+						/* translators: %s: shipping zone name */
+						__( 'Shipping zone "%s" has no flat-rate method, so the feed carries no shipping for it (live-rate / calculated methods price at checkout).', 'woocommerce-gateway-stripe' ),
```
daledupreez: This message is overly verbose, especially when a store in this situation may have many distinct shipping zones. It ends up making the warnings overwhelming and repetitive. In the long run, we'll likely need to integrate with other shipping integrations, so keeping the message neutral may accommodate that better.
```suggestion
						__( 'We cannot precompute a shipping price for shipping zone "%s", so we cannot send any shipping costs to Stripe.', 'woocommerce-gateway-stripe' ),
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -152,6 +153,30 @@ public function generate( int $detail_limit = self::DEFAULT_DETAIL_LIMIT, int $s
 		$scan_limited           = false;
 		$validation_errors      = [];
 
+		// Zones that contribute no flat-rate shipping to the feed. Computed once,
+		// independent of the product walk.
+		$shipping_warnings = [];
+		if ( $mapper instanceof WC_Stripe_Agentic_Commerce_Product_Mapper ) {
```
daledupreez, line 160: High-level question: how much of an issue is this for merchants who are redirecting shoppers to WooCommerce checkout? In those redirect cases, the shipping costs will be computed in checkout, so it's not clear how much of an issue not having prices would be.

daledupreez (commented):
This is working as described, but I think we still need some visual tweaks, and I think we should consider the messaging for merchants who redirect agentic traffic to the local checkout.

client/settings/agentic-commerce/feed-preview.js
wjrosa: Done in 800f5b5d2: wp-admin strips the markers, so the list gets an explicit disc style back.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
wjrosa: For redirect merchants it is only a display gap: agents show the feed price without shipping, and the real cost is computed at WooCommerce checkout as usual. The warning mainly protects in-agent checkout stores, where a missing shipping price means the shopper pays no shipping and the merchant absorbs it. If it reads as noise for store-wide redirect setups, I can gate the warning on the checkout mode in a follow-up.

wjrosa: Implemented in 2cdb64804: with store-wide redirect on, the preview shows the zones as an info notice with a line explaining shipping is computed at the WooCommerce checkout; in-agent stores keep the warning.

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -216,25 +224,38 @@ const AgenticCommerceFeedPreview = () => {
...
+												),
+												shippingWarnings.length
+											) }
+										</strong>
 									</summary>
-									<ul>
+									{ shippingWarningsSeverity === 'info' && (
+										<p>
+											{ __(
+												'Agentic shoppers are redirected to your WooCommerce checkout, where shipping is computed as usual; the feed just won’t advertise a shipping price for these zones.',
```
daledupreez: Nit: The wording here is a bit unexpected for me, as "the feed" doesn't actually "advertise" anything. Maybe we should focus on the shopper-facing outcome?
```suggestion
												'Agentic shoppers are redirected to your WooCommerce checkout, where shipping is computed as usual. The agent won’t be able to show shipping prices for these zones.',
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -261,20 +262,24 @@ public function generate( int $detail_limit = self::DEFAULT_DETAIL_LIMIT, int $s
...
-			'shipping_warnings'  => $shipping_warnings,
+			'invalid_count'              => $invalid_count,
+			'validation_errors'          => $validation_errors,
+			'truncated'                  => $truncated,
+			'scan_limited'               => $scan_limited,
+			'shipping_warnings'          => $shipping_warnings,
+			// A missing shipping price only undercharges shoppers in in-agent
+			// checkout. With store-wide redirect the cost is computed at the
+			// merchant's own checkout, so the zones are informational there.
+			'shipping_warnings_severity' => WC_Stripe_Agentic_Commerce_Integration::is_checkout_disabled() ? 'info' : 'warning',
```
daledupreez: Nit: Given the need for a comment and the decision, I think it makes more sense to compute the value before we build the return array.
```suggestion
		// Shipping issues should only be reported as warnings when Stripe hosts checkout.
		// When shoppers access WooCommerce checkout, shipping will be computed at that stage.
		$shipping_warnings_severity = WC_Stripe_Agentic_Commerce_Integration::is_checkout_disabled() ? 'info' : 'warning';

		return [
			'total_count'                => $total_count,
			'included_count'             => $included_count,
			'excluded_count'             => $excluded_count,
			'excluded_breakdown'         => [
				'subscriptions'      => $excluded_subscriptions,
				'password_protected' => $excluded_password,
				'hidden'             => $excluded_hidden,
				'filtered'           => $excluded_filtered,
			],
			'invalid_count'              => $invalid_count,
			'validation_errors'          => $validation_errors,
			'truncated'                  => $truncated,
			'scan_limited'               => $scan_limited,
			'shipping_warnings'          => $shipping_warnings,
			'shipping_warnings_severity' => $shipping_warnings_severity,
```

client/settings/agentic-commerce/feed-preview.js
```diff
@@ -216,25 +224,38 @@ const AgenticCommerceFeedPreview = () => {
...
-												'woocommerce-gateway-stripe'
-											),
-											shippingWarnings.length
-										) }
+										<strong>
+											{ sprintf(
+												/* translators: %d: number of shipping zones the warning covers. */
+												_n(
+													'%d shipping zone will have no shipping in the feed:',
+													'%d shipping zones will have no shipping in the feed:',
```
daledupreez: Nit: We should specify that they will have no shipping _prices_ in the feed. I also think we can drop the `:` now that we are using bold text, though keeping a `.` feels like it will be best. (I considered removing it, but think keeping a full sentence will be cleaner.)
```suggestion
													'%d shipping zone will have no shipping prices in the feed.',
													'%d shipping zones will have no shipping prices in the feed.',
```

daledupreez (approved):
Overall, this is looking really good. I do have some feedback on wording and commentary, but no functional items remain, so I am going to approve -- I think we can get this merged after addressing my wording issues.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
wjrosa: Applied in the latest commit with you as co-author. Note for next time: the suggestion was anchored to a single line but contained the whole return array, so accepting it via the UI would have duplicated the array.

PR #5659, opened 2026-07-01
Add merchant-config toggles for add-on/configurator catalogs (Agentic Commerce) [3/3]

wjrosa:
Towards [STRIPE-1249](https://linear.app/a8c/issue/STRIPE-1249)
Towards #5656

Part 3 of 3 splitting #5655 (superseded) into reviewable, stacked PRs.

Stack:
1. #5657 — shareable `disable_checkout` filter
2. #5658 — shipping-zone diagnostics
3. **This PR** — add-on/configurator detection + toggles (base: #5658)

> **Review note:** based on #5658 (which is based on #5657), so the diff includes those branches until they merge. Review this branch's commit in isolation, or wait for the lower PRs to merge and the base to retarget.

Lets merchants running configurator plugins join the Agentic Commerce pilot **without custom mu-plugin code**. Configurator products carry runtime-variable pricing the static feed can't represent (and the order-creation path rejects any line whose live price drifts from what Stripe charged).

- **Detection** — `WC_Stripe_Agentic_Commerce_Product_Mapper::product_has_addons()` inspects per-product postmeta (`_product_addons`, `tm_meta_cpf_options`, `composite_data`, and `_wc_pb_priced_individually=yes` only), resolved on the parent for variations, extensible via `woocommerce_agentic_commerce_addon_detection_meta_keys`. It deliberately does not gate on `class_exists()` (an active plugin doesn't mean a given product is configured).
- **Two opt-in toggles (default off)** — auto-exclude detected products from the feed, or auto-redirect them (`disable_checkout`). Both are *defaults*; a custom filter still wins. Exclude wins over redirect.
- **Feed-preview advisories** — per-product `advisories` (`no_sku`, `excluded` with the branch that excluded it, `disable_checkout` with its source).

Additive and gated behind toggles that default off, so an upgrade with no new options set behaves identically.

## Testing instructions

Prerequisites: developer feature flag (`_wcstripe_feature_agentic_commerce`) + merchant toggle (**Stripe settings → Agentic commerce → Enable agentic commerce**).

1. **Auto-exclude.** Add Product Add-Ons (or `_product_addons` meta) to a product, enable **"Exclude products with add-ons or configurators from the feed"**. Preview feed → product shows an `excluded` advisory (reason `addons`) and drops from the included count; disable → it returns.
2. **Auto-redirect.** With exclude off, enable **"Redirect shoppers to my store…"**. Preview → `disable_checkout` advisory (source `addons`); confirm `disable_checkout=true` in the feed.
3. **Filter still wins.** A `woocommerce_agentic_commerce_should_sync_product` / `woocommerce_agentic_commerce_disable_checkout` filter overrides the toggle.
4. `npm run test:php -- --filter WC_Stripe_Agentic_Commerce_Product_Mapper_Test`, `--filter WC_Stripe_Agentic_Commerce_Feed_Preview_Test`, `--filter WC_REST_Stripe_Agentic_Commerce_Controller_Test`; `npm run test:js -- client/settings/agentic-commerce`.

---

### Changelog entry

Added under the unreleased 10.9.0 block.

**Post merge**


includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -101,6 +87,27 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
...
+	/**
+	 * Option key ('yes'/'no', default off) for defaulting `disable_checkout=true`
+	 * on detected configurator products so agents redirect shoppers to the store
+	 * while the products still syndicate for discovery. Exclude wins over this
+	 * (an excluded product is never in the feed).
+	 *
+	 * @var string
+	 * @since 10.9.0
+	 */
+	public const AUTO_DISABLE_CHECKOUT_ADDONS_OPTION = 'wc_stripe_agentic_commerce_auto_disable_checkout_addons';
```
daledupreez: I find this naming very confusing! While "disable" applies to the specific Stripe setting we're working with, that clashes with the more generic meaning of the word: the effect of the setting is to _redirect_ for this case by marking Stripe's hosted checkout flow as disabled. I think it would be a _lot_ clearer if we called this option and the various getter methods something that reflects the outcome. I would propose something like `AUTO_REDIRECT_CHECKOUT_ADDONS_OPTION` and `wc_stripe_agentic_commerce_auto_redirect_checkout_addons`.
```suggestion
	/**
	 * Option key ('yes'/'no', default off) for defaulting `disable_checkout=true`
	 * on detected configurator products so agents redirect shoppers to the store
	 * while the products are still uploaded to Stripe for discovery. Exclude wins over this
	 * (an excluded product is never in the feed).
	 *
	 * @var string
	 * @since 10.9.0
	 */
	public const AUTO_REDIRECT_CHECKOUT_ADDONS_OPTION = 'wc_stripe_agentic_commerce_auto_redirect_checkout_addons';
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1198,15 +1228,105 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+		/**
+		 * Meta keys whose non-empty presence marks a product as carrying add-on /
+		 * configurator options. Extend to support plugins not covered here.
+		 *
+		 * @since 10.9.0
+		 * @param string[]    $meta_keys Meta keys treated as add-on signals.
+		 * @param \WC_Product $target    Product whose meta is inspected (parent for variations).
+		 */
+		$meta_keys = apply_filters(
+			'woocommerce_agentic_commerce_addon_detection_meta_keys',
```
daledupreez, line 1332: Is this filter one that exists upstream in WooCommerce or WooCommerce AI? Is there a reason we're using the `woocommerce_` prefix?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1198,15 +1228,105 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+				if ( ! is_string( $meta_key ) || '' === $meta_key ) {
+					continue;
+				}
+				if ( ! empty( $target->get_meta( $meta_key ) ) ) {
+					return true;
+				}
+			}
+		}
+
+		// Bundles only vary at runtime when priced individually; the meta is `no`
```
daledupreez, line 1354: Not blocking: is it worth adding an explicit filter that allows plugins to indicate whether a specific product is an addon? e.g. something like `wc_stripe_agentic_commerce_product_has_addon`?

daledupreez (commented):
I have some comments up from inspection.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: The `woocommerce_` prefix is deliberate, not an upstream filter: our Agentic Commerce hooks use the shareable `woocommerce_agentic_commerce_*` prefix so non-Stripe integrations can hook the same decision. It's new, so there's no `wc_stripe_` twin to deprecate.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Added `woocommerce_agentic_commerce_product_has_addon` on the final verdict (with a test + cookbook entry). Used the shareable `woocommerce_` prefix to match the sibling hooks; happy to switch to `wc_stripe_` if you'd rather.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
daledupreez: If this is not an upstream filter in WooCommerce, we should not be using the `woocommerce_` prefix, at least not without some sort of communication/understanding with WooCommerce about this particular usage or some way of publicising the use case outside of the plugin's documentation. I strongly object to us using the prefix in this way.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -311,4 +322,80 @@ private function collect_error( array &$validation_errors, int &$truncated, int
...
 	}
+
+	/**
+	 * Collect the non-fatal advisories for a single product: why it is excluded,
+	 * whether it redirects (and from which layer), and missing-SKU warnings.
+	 *
+	 * Excluded products are never in the feed, so their SKU / checkout state is
+	 * moot — only the exclusion reason is recorded for them.
+	 *
+	 * @since 10.9.0
```
daledupreez: Nit: since tags here and elsewhere should be bumped to 11.1.0.
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -101,6 +101,27 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
 	 */
 	public const DISABLE_CHECKOUT_OPTION = 'wc_stripe_agentic_commerce_disable_checkout';
 
+	/**
+	 * Option key ('yes'/'no', default off) for auto-excluding detector-flagged
+	 * add-on / configurator products from the feed. Drives a default branch in
+	 * {@see WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product()}.
```
daledupreez: I think we should simplify the comment.
```suggestion
	 * Option key ('yes'/'no', default off) for auto-excluding detector-flagged
	 * add-on / configurator products from the feed.
	 *
	 * @see WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product()
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -101,6 +101,27 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
...
+	 * @var string
+	 * @since 10.9.0
+	 */
+	public const AUTO_EXCLUDE_ADDONS_OPTION = 'wc_stripe_agentic_commerce_auto_exclude_addons';
+
+	/**
+	 * Option key ('yes'/'no', default off) for defaulting `disable_checkout=true`
+	 * on detected configurator products so agents redirect shoppers to the store
+	 * while the products are still uploaded to Stripe for discovery. Exclude wins over this
+	 * (an excluded product is never in the feed).
```
daledupreez:
```suggestion
	/**
	 * Option key ('yes'/'no', default off) for defaulting `disable_checkout=true`
	 * on detected add-on/configurator products, so shoppers will be redirected
	 * to the store for these products.
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -656,6 +677,32 @@ public static function is_checkout_disabled(): bool {
 		return 'yes' === get_option( self::DISABLE_CHECKOUT_OPTION, 'no' );
 	}
 
+	/**
+	 * Whether the merchant has opted to auto-exclude add-on / configurator
+	 * products from the feed entirely. Drives a default branch in
+	 * {@see WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product()};
+	 * a custom filter still overrides it.
+	 *
+	 * @since 10.9.0
```
daledupreez: As above RE the comment and the since tag.
```suggestion
	 * Whether the merchant has opted to auto-exclude add-on / configurator
	 * products from the feed entirely.
	 *
	 * @see WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product()
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -656,6 +677,32 @@ public static function is_checkout_disabled(): bool {
...
+	}
+
+	/**
+	 * Whether the merchant has opted to auto-set `disable_checkout` (feed-only /
+	 * redirect) on add-on / configurator products. Drives a default branch in
+	 * {@see WC_Stripe_Agentic_Commerce_Product_Mapper::get_disable_checkout()};
+	 * a custom filter still overrides it.
+	 *
+	 * @since 10.9.0
+	 * @return bool
```
daledupreez: As above RE simplification and the since tag.
```suggestion
	 * Whether the merchant has opted to set `disable_checkout` for add-on /
	 * configurator products so they redirect to the local checkout.
	 *
	 * @see WC_Stripe_Agentic_Commerce_Product_Mapper::get_disable_checkout()
	 * @since 11.1.0
	 * @return bool
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -26,6 +26,26 @@
...
+	 * Postmeta keys whose non-empty presence marks a product as carrying add-on /
+	 * configurator options. The first three seed the
+	 * `woocommerce_agentic_commerce_addon_detection_meta_keys` filter default;
+	 * the Bundles key is checked separately (only `yes` counts). See
+	 * {@see self::product_has_addons()}.
+	 */
+	public const ADDON_META_PRODUCT_ADDONS             = '_product_addons';
+	public const ADDON_META_EXTRA_PRODUCT_OPTIONS      = 'tm_meta_cpf_options';
+	public const ADDON_META_COMPOSITE                  = 'composite_data';
+	public const ADDON_META_BUNDLE_PRICED_INDIVIDUALLY = '_wc_pb_priced_individually';
```
daledupreez: Do these constants need to be public? And does it make sense to use an array constant for the group? It feels like a private constant would work for the array of general keys.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -256,14 +300,31 @@ protected function get_disable_checkout( \WC_Product $product, ?\WC_Product $par
...
-		 * @param bool             $disabled       Store-wide default.
+		 * @param bool             $disabled       Resolved default (store-wide option, plus the add-on auto-default).
 		 * @param \WC_Product      $product        Product object.
 		 * @param \WC_Product|null $parent_product Parent product for variations.
 		 */
 		// wp_validate_boolean() rather than a plain (bool) cast: a callback that
 		// returns the string 'false' would be truthy under a cast and wrongly
 		// enable redirect mode. This still normalises null / 0 / '' to false.
-		return wp_validate_boolean( apply_filters( 'woocommerce_agentic_commerce_disable_checkout', $disabled, $product, $parent_product ) );
+		$disabled = wp_validate_boolean( apply_filters( 'woocommerce_agentic_commerce_disable_checkout', $disabled, $product, $parent_product ) );
```
daledupreez: While we're updating this code, we should also fix the filter docs. I also have two other concerns with the filter name:
1. We're using a `woocommerce_` prefix for a Stripe-specific filter that resolves which checkout to use -- that's not a WooCommerce Core question.
2. We're using `disable_checkout` in the filter name, which doesn't match the effect of forcing a redirect to WooCommerce checkout. But maybe this aspect should be handled in a separate follow-up.
```suggestion
		 */
		$should_redirect = apply_filters( 'wc_stripe_agentic_commerce_should_redirect_to_woocommerce', $disabled, $product, $parent_product );

		// Use wp_validate_boolean() to handle edge cases for boolean values.
		$should_redirect = wp_validate_boolean( $should_redirect );
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1221,4 +1292,112 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+			foreach ( $meta_keys as $meta_key ) {
+				if ( ! is_string( $meta_key ) || '' === $meta_key ) {
+					continue;
+				}
+				if ( ! empty( $target->get_meta( $meta_key ) ) ) {
+					$has_addons = true;
+					break;
+				}
+			}
+		}
```
daledupreez: Not blocking, but this code could make sense in a private helper method, as it would allow us to handle a few situations more easily:
 * If the returned meta keys are not an array or empty, we can return early.
 * When we first find a matching value, we can also return early.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1221,4 +1292,112 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+		 *
+		 * Uses the shareable `woocommerce_` prefix so non-Stripe Agentic Commerce
+		 * integrations can hook the same decision.
+		 *
+		 * @since 11.1.0
+		 * @param bool        $has_addons Whether meta-key detection flagged the product.
+		 * @param \WC_Product $product    The product (or variation) inspected.
+		 * @param \WC_Product $target     The product whose meta was read (parent for variations).
+		 */
+		return (bool) apply_filters( 'woocommerce_agentic_commerce_product_has_addon', $has_addons, $product, $target );
```
daledupreez: As for the other cases, I don't think we should be using a `woocommerce_` prefix for this filter. It's not clear that the distinction is or will be relevant for other integrations, and we don't have a way to communicate these elements easily.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Done in 2062558ba: the loop lives in a private matches_addon_meta_keys() with early returns for a non-array/empty key set and on the first match.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Fair. Renamed in 2062558ba to `wc_stripe_agentic_commerce_product_has_addon` (and its sibling meta-keys filter to match); both are unreleased, so renamed in place with docs updated. The should_sync/disable_checkout prefix question from your other threads is left open there.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Done in 2062558ba: the three general keys are now one private array constant seeding the filter default, and the Bundles key is private too. Nothing outside the mapper read them.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Agreed, fixed in 135d24827: all four filters are canonical under `wc_stripe_`. The `woocommerce_` prefix came from WOOAI-636/637, a planned cross-plugin contract owned by woocommerce-ai; I double-checked and that side was never built (both issues still in Triage) and woocommerce-ai uses its own `woocommerce_ai_*` prefix anyway. The shipped `should_sync_product` / `disable_checkout` names keep working via `apply_filters_deprecated` with the roles swapped; the two add-on filters are unreleased, so renamed in place.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Done in 135d24827, together with the other prefix threads: `wc_stripe_` is canonical for all four filters now. Context in my reply on the product_has_addon thread.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Prefix and docs fixed in 135d24827 (`wc_stripe_agentic_commerce_disable_checkout` is canonical; the old name is deprecated). On the second point: `disable_checkout` matches the Stripe feed column the filter feeds, and the name shipped in 10.9.0, so I agree with your instinct to leave the rename to a follow-up if we still want it.

wjrosa: Filed the rename follow-up as STRIPE-1457.

client/settings/agentic-commerce/index.js
```diff
@@ -118,11 +125,17 @@ const AgenticCommerceSection = forwardRef( ( props, ref ) => {
 				data: {
 					is_enabled: isFeatureEnabled,
 					disable_checkout: disableCheckout,
+					auto_exclude_addons: autoExcludeAddons,
+					auto_redirect_checkout_addons: autoRedirectCheckoutAddons,
```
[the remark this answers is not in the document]
daledupreez: This feels like it would be worth flagging/preventing. Is it worth tackling this in a follow-up PR?

client/settings/agentic-commerce/index.js
```diff
@@ -226,6 +245,38 @@ const AgenticCommerceSection = forwardRef( ( props, ref ) => {
...
+										help={ __(
+											"Products built with add-on or configurator plugins (Product Add-Ons, Extra Product Options, Composite Products, individually-priced Bundles) have prices that depend on shopper choices, which the feed can't represent. When enabled, those products are kept out of the catalog entirely.",
+											'woocommerce-gateway-stripe'
+										) }
+										checked={ autoExcludeAddons }
+										onChange={ setAutoExcludeAddons }
+									/>
+								) }
+
+								{ isFeatureEnabled && ! autoExcludeAddons && (
```
daledupreez: UX question: is it worth showing this field as disabled when `! autoExcludeAddons` so merchants can always tell the option exists?

includes/agentic-commerce/AGENTS.md
```diff
@@ -30,6 +30,15 @@ Read `README.md` in this directory for domain-specific feed requirements.
 - Entry values must remain scalar or `null`; complex values must be serialized before `add_entry()`.
 - Keep logging and cleanup behavior on error paths intact.
 
+## Merchant Configuration Surface
+
+See README.md "Merchant configuration cookbook" for the merchant-facing recipes. When changing this surface, keep all of it in sync:
+
+- **Filters** use the plugin's own `wc_stripe_agentic_commerce_*` prefix. Canonical names: `wc_stripe_agentic_commerce_should_sync_product`, `wc_stripe_agentic_commerce_disable_checkout`, `wc_stripe_agentic_commerce_addon_detection_meta_keys`, `wc_stripe_agentic_commerce_product_has_addon`. The `woocommerce_agentic_commerce_*` names for `should_sync_product` and `disable_checkout` shipped in 10.9.0 anticipating a shared cross-plugin contract that never materialized (WOOAI-636); they are deprecated via `apply_filters_deprecated` (seed the canonical default) — do not drop them.
```
daledupreez: I think we can simplify this comment and should remove the `WOOAI-636` reference.
```suggestion
- **Filters** use the `wc_stripe_agentic_commerce_*` prefix. Canonical names: `wc_stripe_agentic_commerce_should_sync_product`, `wc_stripe_agentic_commerce_disable_checkout`, `wc_stripe_agentic_commerce_addon_detection_meta_keys`, `wc_stripe_agentic_commerce_product_has_addon`. The `woocommerce_agentic_commerce_*` names for `should_sync_product` and `disable_checkout` shipped in 10.9.0 anticipating a shared cross-plugin contract that never materialized, so they are deprecated via `apply_filters_deprecated()` — do not drop them.
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -153,6 +155,32 @@ public function generate( int $detail_limit = self::DEFAULT_DETAIL_LIMIT, int $s
...
+						[
+							'page'    => 'wc-settings',
+							'tab'     => 'shipping',
+							'zone_id' => $zone['id'],
+						],
+						admin_url( 'admin.php' )
+					),
+				];
+			}
+		}
```
daledupreez: This code block mostly duplicates the code block below, but will produce double errors for problematic shipping zones. (I suspect this was a bad merge.)
```suggestion
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -317,4 +352,80 @@ private function collect_error( array &$validation_errors, int &$truncated, int
...
+			$source = $mapper->resolve_disable_checkout( $product )['source'];
+			$this->add_advisory( $advisories, $truncated, $detail_limit, $product, 'disable_checkout', (string) $source );
+		}
+	}
+
+	/**
+	 * Append a single advisory to the list, or bump the overflow counter once the
+	 * detail cap is reached.
+	 *
+	 * @since 10.9.0
```
daledupreez: Should this not be 11.1.0?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -237,33 +264,76 @@ protected function get_link( \WC_Product $product ): string {
...
-		// Legacy filter's result seeds the default for the canonical filter below,
-		// so existing hooks keep working while a hook on the new name wins.
+	/**
+	 * Resolve the `disable_checkout` (feed-only / redirect) decision plus the
+	 * source it came from, for the feed preview. Precedence, lowest first:
+	 * store-wide option, add-on auto-default, then filters (a hook always wins,
+	 * keeping the first two opt-in defaults). Applies the `disable_checkout`
+	 * filters, so any merchant hook runs here.
+	 *
+	 * @since 10.9.0
```
daledupreez: Nit: Version.
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -237,33 +264,76 @@ protected function get_link( \WC_Product $product ): string {
...
+			&& WC_Stripe_Agentic_Commerce_Integration::is_auto_redirect_checkout_addons_enabled()
+			&& self::product_has_addons( $product );
+
+		$default = $store_wide || $addon_default;
+
+		// The woocommerce_-prefixed name shipped in 10.9.0 anticipating a shared
+		// cross-plugin contract that woocommerce-ai never implemented (WOOAI-636),
+		// so the plugin's own prefix is canonical again. The deprecated name's
+		// result seeds the canonical filter's default, so existing hooks keep
+		// working while a hook on the canonical name wins.
```
daledupreez: I think we should reduce the context/commentary here, and we should remove the `WOOAI-636` reference.
```suggestion
		// The woocommerce_-prefixed filter shipped in 10.9.0 anticipating a shared
		// cross-plugin contract that was never implemented, so it has been deprecated.
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -237,33 +264,76 @@ protected function get_link( \WC_Product $product ): string {
...
-		 * @param bool             $disabled       Store-wide default.
+		 * @param bool             $disabled       Resolved default (store-wide option, plus the add-on auto-default).
 		 * @param \WC_Product      $product        Product object.
 		 * @param \WC_Product|null $parent_product Parent product for variations.
 		 */
 		// wp_validate_boolean() rather than a plain (bool) cast: a callback that
 		// returns the string 'false' would be truthy under a cast and wrongly
 		// enable redirect mode. This still normalises null / 0 / '' to false.
-		return wp_validate_boolean( apply_filters( 'woocommerce_agentic_commerce_disable_checkout', $disabled, $product, $parent_product ) );
+		$disabled = wp_validate_boolean( apply_filters( 'wc_stripe_agentic_commerce_disable_checkout', $disabled, $product, $parent_product ) );
```
daledupreez: Nit: Please run the filter and then the boolean wrapper so the filter docs are picked up.
```suggestion
		 * @param \WC_Product      $product        Product object.
		 * @param \WC_Product|null $parent_product Parent product for variations.
		 */
		$disabled = apply_filters( 'wc_stripe_agentic_commerce_disable_checkout', $disabled, $product, $parent_product );

		// wp_validate_boolean() rather than a plain (bool) cast: a callback that
		// returns the string 'false' would be truthy under a cast and wrongly
		// enable redirect mode. This still normalises null / 0 / '' to false.
		$disabled = wp_validate_boolean( $disabled );
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1172,15 +1242,27 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+			&& self::product_has_addons( $product )
+		) {
+			$default_should_sync = false;
+		}
+
+		// The woocommerce_-prefixed name shipped in 10.9.0 anticipating a shared
+		// cross-plugin contract that woocommerce-ai never implemented (WOOAI-636),
+		// so the plugin's own prefix is canonical again. The deprecated name's
+		// result seeds the canonical filter's default, so existing adapters keep
+		// working while a hook on the canonical name wins.
```
daledupreez: As above RE the comments.
```suggestion
		// The woocommerce_-prefixed name shipped in 10.9.0 anticipating a shared
		// cross-plugin contract that was not implemented.
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1213,12 +1291,120 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
 		 *
-		 * @since 10.9.0
+		 * @since 10.8.0
 		 * @param bool        $should_sync Whether to include the product. Default true, except for
 		 *                                 subscriptions, password-protected products, and products
 		 *                                 hidden from catalog and search.
 		 * @param \WC_Product $product     Product being evaluated.
 		 */
-		return wp_validate_boolean( apply_filters( 'woocommerce_agentic_commerce_should_sync_product', $should_sync, $product ) );
+		return wp_validate_boolean( apply_filters( 'wc_stripe_agentic_commerce_should_sync_product', $should_sync, $product ) );
```
daledupreez: Nit: As above, we should split the filter docs and the variable cleanup.
```suggestion
		$should_sync = apply_filters( 'wc_stripe_agentic_commerce_should_sync_product', $should_sync, $product );
		return wp_validate_boolean( $should_sync );
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1213,12 +1291,120 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+	 * price depend on runtime shopper choices the static feed can't honour.
+	 *
+	 * Detection is per-product postmeta (resolved on the parent for variations)
+	 * and deliberately does NOT key off `class_exists()` of the configurator
+	 * plugins — an active plugin says nothing about whether *this* product is
+	 * configured, so a bare class check would flag the whole catalog. Covered:
+	 * Product Add-Ons (`_product_addons`), TM Extra Product Options
+	 * (`tm_meta_cpf_options`), Composite Products (`composite_data`), and Product
+	 * Bundles only when priced individually (`_wc_pb_priced_individually` = `yes`;
+	 * a fixed-price bundle stays eligible). The meta-key set is filterable.
```
daledupreez: I think we should refine this comment to make the implementation details clearer.
```suggestion
	 * Detection is via product meta fields, which are resolved on the parent for variations.
	 * The product meta checks include the following:
	 *  - Product Add-Ons - `_product_addons`
	 *  - TM Extra Product Options - `tm_meta_cpf_options`
	 *  - Composite Products - `composite_data` 
	 *  - Product Bundles when they are priced individually - `_wc_pb_priced_individually` is `'yes'`
	 *
	 * The set of meta keys to check can be filtered via the wc_stripe_agentic_commerce_addon_detection_meta_keys filter.

File: includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php [hunk 24]
```diff
@@ -1213,12 +1291,120 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+	 *
+	 * @since 10.9.0
```
daledupreez: Nit: Version.
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1213,12 +1291,120 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+		 * whose options are not stored in a detectable meta key mark a product as
+		 * configurable (or clear a false positive) directly, instead of only
+		 * extending the meta-key set.
+		 *
+		 * @since 11.1.0
+		 * @param bool        $has_addons Whether meta-key detection flagged the product.
+		 * @param \WC_Product $product    The product (or variation) inspected.
+		 * @param \WC_Product $target     The product whose meta was read (parent for variations).
+		 */
+		return (bool) apply_filters( 'wc_stripe_agentic_commerce_product_has_addon', $has_addons, $product, $target );
```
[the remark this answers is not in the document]
daledupreez: Should this be modified to run the return value through `wp_validate_boolean()`?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -1213,12 +1291,120 @@ public static function should_sync_product( \WC_Product $product ): bool {
...
+	}
+
+	/**
+	 * Diagnostic companion to {@see self::should_sync_product()}: report the most
+	 * likely reason a product is excluded from the feed, or null when it syncs.
+	 * Branch order matches how the default is built (subscription, then add-on
+	 * auto-exclude), falling through to the filter reason when a custom hook
+	 * forced it.
+	 *
+	 * @since 10.9.0
```
daledupreez: Nit: Version.
```suggestion
	 * @since 11.1.0
```

includes/agentic-commerce/README.md
```diff
@@ -363,7 +363,7 @@ add_filter(
 );
 ```
 
-> The Stripe-prefixed `wc_stripe_agentic_commerce_disable_checkout` filter is **deprecated since 10.9.0** in favour of the shareable `woocommerce_agentic_commerce_disable_checkout` above (mirroring the `woocommerce_agentic_commerce_should_sync_product` migration). Existing hooks on the old name still run — they seed the new filter's default — but emit a deprecation notice.
+> The `woocommerce_agentic_commerce_disable_checkout` name is **deprecated since 11.1.0** in favour of the Stripe-prefixed `wc_stripe_agentic_commerce_disable_checkout` above (mirroring the `should_sync_product` naming; the planned shared cross-plugin contract never materialized). Existing hooks on the deprecated name still run — they seed the canonical filter's default — but emit a deprecation notice.
```
daledupreez: Nit: Minor wording improvements.
```suggestion
> The `woocommerce_agentic_commerce_disable_checkout` filter is **deprecated since 11.1.0** in favour of the Stripe-prefixed `wc_stripe_agentic_commerce_disable_checkout` filter.
```

includes/agentic-commerce/README.md
```diff
@@ -419,3 +419,94 @@ WooCommerce coupons do not participate in delegated (in-agent) checkout. Prices
...
+
+Configurator products carry runtime-variable pricing the static feed can't
+represent, and the order-creation path rejects any line whose live price drifts
+from what Stripe charged. Two opt-in toggles (both default **off**) handle them
+without code, under **Stripe settings → Agentic commerce**:
+
+| Toggle | Option key | Effect |
+| --- | --- | --- |
+| **Exclude products with add-ons or configurators from the feed** | `wc_stripe_agentic_commerce_auto_exclude_addons` | Detected products never enter the catalog. Use when configurator SKUs should not appear in agents at all. |
+| **Redirect shoppers to my store for products with add-ons or configurators** | `wc_stripe_agentic_commerce_auto_redirect_checkout_addons` | Detected products stay discoverable but agents send shoppers to the store to configure and buy (`disable_checkout=true`). Use to keep discoverability while moving the actual purchase on-site. |
```
daledupreez, line 454: Should we not default the one of the two options to yes?

includes/agentic-commerce/README.md
```diff
@@ -419,3 +419,94 @@ WooCommerce coupons do not participate in delegated (in-agent) checkout. Prices
 This limitation only affects delegated (in-agent) purchases. In feed-only / redirect mode (see [Checkout Mode](#checkout-mode-embedded-vs-redirect) above) the shopper completes checkout on the store, where WooCommerce coupons and their usage limits apply as usual.
 
 Whether to support promotions in delegated checkout (mapping Stripe discounts to WooCommerce coupons and enforcing usage limits at `finalize_checkout`) is an open product decision, tracked in STRIPE-1257.
+
+## Merchant configuration cookbook
+
+Recipes for onboarding catalogs that mix standard SKUs with configurator
+products (WooCommerce Product Add-Ons, TM Extra Product Options, Composite
+Products, individually-priced Bundles) — without writing custom mu-plugin code.
```
daledupreez: I think we should remove the "mu-plugins reference" - this may still require filtering for some use cases.
```suggestion
Products, individually-priced Bundles).
```

includes/agentic-commerce/README.md
```diff
@@ -419,3 +419,94 @@ WooCommerce coupons do not participate in delegated (in-agent) checkout. Prices
...
+
+Exclude wins over redirect (an excluded product is never in the feed, so its
+checkout mode is moot). Both are **defaults** — a custom filter still wins.
+
+Detection is per-product postmeta (resolved on the parent for variations).
+Out of the box it covers `_product_addons`, `tm_meta_cpf_options`,
+`composite_data`, and `_wc_pb_priced_individually=yes`. It deliberately does not
+key off `class_exists()` of the plugins, since an active plugin says nothing
+about whether a given product is configured. Extend the detected set without
+forking:
```
daledupreez: This content is not easy to read. I think we should be more explicit.
```suggestion
The default detection includes the following checks:
 - WooCommerce Product Add-Ons - `_product_addons`
 - TM Extra Product Options - `tm_meta_cpf_options`
 - Composite Products - `composite_data`
 - WooCommerce Bundles that are individually priced - `_wc_pb_priced_individually=yes`

Additional meta fields can be added using the `wc_stripe_agentic_commerce_addon_detection_meta_keys` filter:
```

includes/agentic-commerce/README.md
```diff
@@ -419,3 +419,94 @@ WooCommerce coupons do not participate in delegated (in-agent) checkout. Prices
...
+
+```php
+add_filter(
+    'wc_stripe_agentic_commerce_addon_detection_meta_keys',
+    function ( array $meta_keys, WC_Product $product ): array {
+        $meta_keys[] = '_my_configurator_options';
```
daledupreez: Just to be extra clear, let's use "meta" in the example key.
```suggestion
        $meta_keys[] = '_my_configurator_meta';
```

includes/agentic-commerce/README.md
```diff
@@ -419,3 +419,94 @@ WooCommerce coupons do not participate in delegated (in-agent) checkout. Prices
...
+        return $meta_keys;
+    },
+    10,
+    2
+);
+```
+
+When a configurator stores nothing in a stable meta key, decide per product with
+`wc_stripe_agentic_commerce_product_has_addon` instead — it filters the final
+verdict, so it can also clear a false positive:
```
daledupreez: I think we should rephrase this content as well.
```suggestion
In cases where more nuanced checks are required, or when the meta keys to check are not consistent,
use the `wc_stripe_agentic_commerce_product_has_addon` filter:
```

includes/agentic-commerce/README.md
```diff
@@ -419,3 +419,94 @@ WooCommerce coupons do not participate in delegated (in-agent) checkout. Prices
...
+    },
+    10,
+    2
+);
+```
+
+### Per-product redirect without the toggles
+
+To redirect specific products regardless of the toggles, hook the
+`wc_stripe_agentic_commerce_disable_checkout` filter shown above. To exclude
```
daledupreez, line 501: Not blocking for this PR, but it may be worth renaming this filter in a follow-up PR.

includes/agentic-commerce/README.md
```diff
@@ -419,3 +419,94 @@ WooCommerce coupons do not participate in delegated (in-agent) checkout. Prices
...
+specific products from the feed, hook `wc_stripe_agentic_commerce_should_sync_product`:
+
+```php
+add_filter(
+    'wc_stripe_agentic_commerce_should_sync_product',
+    function ( bool $should_sync, WC_Product $product ): bool {
+        return get_post_meta( $product->get_id(), '_hide_from_agents', true ) ? false : $should_sync;
```
daledupreez: Question: Should we ever recommend `get_post_meta()`?

tests/phpunit/admin/class-wc-rest-stripe-agentic-commerce-controller-test.php
```diff
@@ -1365,6 +1385,29 @@ public function test_update_settings_persists_disable_checkout(): void {
...
+	 */
+	public function test_update_settings_persists_addon_toggles(): void {
+		$request = new WP_REST_Request( 'POST', self::REST_BASE . '/settings' );
+		$request->set_body(
+			wp_json_encode(
+				[
+					'auto_exclude_addons'           => true,
+					'auto_redirect_checkout_addons' => true,
+				]
+			)
```
daledupreez, line 1399: Nit: The tests should also cover the cases where the incoming values are false.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview-test.php
```diff
@@ -572,6 +676,23 @@ public function test_preview_flags_default_zone_alongside_named_zones(): void {
...
 	}
+
+	/**
+	 * Collect the product IDs of advisories of a given type.
+	 *
+	 * @param array  $advisories Advisory list from the preview.
+	 * @param string $type       Advisory type to filter on.
+	 * @return int[]
+	 */
+	private function advisory_ids_of_type( array $advisories, string $type ): array {
```
daledupreez: Nit: The method should include a verb describing the action. Something like `get_advisory_ids_of_type()` or `extract_advisory_ids_of_type()` would be much clearer.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview-test.php
```diff
@@ -493,6 +493,110 @@ public function test_variation_error_links_to_parent_product(): void {
...
 
+	/**
+	 * generate() exposes the advisory and shipping-warning surfaces added for
+	 * the merchant-configuration cookbook.
+	 *
+	 * @return void
+	 */
+	public function test_preview_exposes_advisory_and_shipping_keys(): void {
+		$product = $this->create_valid_product();
+		$this->scope_to( [ $product->get_id() ] );
```
daledupreez: Not blocking, but `scope_to()` is not easy to understand from the name. That may be worth expanding to something like `limit_product_query_to_product_ids()` so it's easier to know what the code does at first glance.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
```diff
@@ -1484,19 +1484,234 @@ public function test_disable_checkout_variation_inherits_parent_via_filter() {
...
 			$parent->delete( true );
 		}
 	}
 
+	/**
+	 * product_has_addons() detects the WooCommerce Product Add-Ons meta key.
+	 *
+	 * @return void
+	 */
+	public function test_product_has_addons_detects_product_addons_meta() {
```
daledupreez: The tests only cover the `_product_addons` meta key -- it feels like using a data provider for the other values would be worth it for coverage. The data provider could likely be re-used across most of these tests to ensure that all of the behaviours tested below are handled correctly, not just for `_product_addons`.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
```diff
@@ -1484,19 +1484,234 @@ public function test_disable_checkout_variation_inherits_parent_via_filter() {
...
+		$product = WC_Helper_Product::create_simple_product();
+		$product->update_meta_data( '_product_addons', [ [ 'name' => 'Engraving' ] ] );
+		$product->save();
+
+		$this->assertTrue( WC_Stripe_Agentic_Commerce_Product_Mapper::product_has_addons( $product ) );
+
+		$product->delete( true );
```
daledupreez: A number of these methods are asserting before any cleanup happens. We should either use `try/finally` or use a local variable for the method result, run the cleanup, and then assert.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
```diff
@@ -1484,19 +1484,234 @@ public function test_disable_checkout_variation_inherits_parent_via_filter() {
...
+	 * The final add-on verdict is filterable so a plugin can flag a configurator
+	 * that stores nothing in a detectable meta key, or clear a false positive.
+	 *
+	 * @return void
+	 */
+	public function test_product_has_addons_verdict_is_filterable() {
+		$product = WC_Helper_Product::create_simple_product();
+		$product->save();
+
+		$this->assertFalse( WC_Stripe_Agentic_Commerce_Product_Mapper::product_has_addons( $product ) );
```
daledupreez: Nit: the try/finally won't clean up after this assertion.

daledupreez (changes requested):

client/settings/agentic-commerce/index.js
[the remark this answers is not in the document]
wjrosa: Good catch. Tracked as #5980 so we can address it separately from this PR.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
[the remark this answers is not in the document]
wjrosa: Fixed in 99ab26b0b. Switched to `wp_validate_boolean()` so a callback returning the string `'false'` clears the verdict instead of casting to true, matching the sibling `should_sync` and `disable_checkout` filters. Added a test with a `'false'`-returning callback.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
```diff
@@ -317,4 +352,80 @@ private function collect_error( array &$validation_errors, int &$truncated, int
...
+	 * @param int                                      $detail_limit Maximum advisories to retain.
+	 * @param \WC_Product                              $product    Product being inspected.
+	 * @param array                                    $row        Mapped feed row for the product.
+	 * @param WC_Stripe_Agentic_Commerce_Product_Mapper $mapper    Concrete mapper for diagnostics.
+	 * @param bool                                     $excluded   Whether the product was excluded from the feed.
+	 * @return void
+	 */
+	private function collect_advisories( array &$advisories, int &$truncated, int $detail_limit, \WC_Product $product, array $row, WC_Stripe_Agentic_Commerce_Product_Mapper $mapper, bool $excluded ): void {
+		if ( $excluded ) {
+			$reason = WC_Stripe_Agentic_Commerce_Product_Mapper::get_sync_exclusion_reason( $product );
```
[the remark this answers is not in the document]
wjrosa: Fixed in 99ab26b0b. The preview now reads the exclusion reason and the disable_checkout source from a single mapper evaluation instead of re-invoking the filters. `map_product()` delegates to a new `map_product_with_diagnostics()` that captures the row and both reasons in one pass, so an advisory can no longer disagree with the row when a callback is stateful. `map_product()`'s output is unchanged. Added tests with stateful callbacks that flip across calls for both the exclusion and disable_checkout advisories.

One scope note: `Feed_Validator::validate_entry()` still calls `should_sync_product()` again to tell an excluded empty row from an invalid one. That is a separate skip-vs-invalid check, not the advisory re-derivation this finding was about, so I left it as is.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
wjrosa: Done in 03c026586: the `product_has_addons` tests now clean up the product via try/finally.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
wjrosa: Done in 03c026586: wrapped the product lifecycle in an outer try/finally so the first assertion can no longer skip cleanup.

tests/phpunit/admin/class-wc-rest-stripe-agentic-commerce-controller-test.php
wjrosa: Done in 03c026586: the add-on toggles test now round-trips through false, matching the disable_checkout coverage.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Applied by hand in fe507062d (the suggestion button no longer applied after the surrounding lines changed). Behavior is unchanged.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
wjrosa: Applied by hand in e98da0412 (button was outdated). Same split applied to `product_has_addon` too, so all three filters read consistently. Behavior unchanged.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview.php
wjrosa: Yes, bumped to `@since 11.1.0` in 03c026586.

includes/agentic-commerce/README.md
wjrosa: Done in 629f4828c: the example now uses `$product->get_meta( '_hide_from_agents' )`, which is HPOS-safe, instead of `get_post_meta()`.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview-test.php
wjrosa: Done in 629f4828c: renamed to `get_advisory_ids_of_type()`.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-feed-preview-test.php
wjrosa: Done in 629f4828c: renamed to `limit_product_query_to_product_ids()`.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
wjrosa: Done in 629f4828c: converted to a `@dataProvider` covering all three built-in meta keys (Product Add-Ons, TM Extra Product Options, Composite Products). The bundle case stays in its own test.

includes/agentic-commerce/README.md
wjrosa: Agreed, tracked as a follow-up in #5983.

client/settings/agentic-commerce/index.js
wjrosa: Done in 131954d93: the redirect toggle is now always rendered when the feature is on, shown disabled (with explanatory help text) while auto-exclude is enabled, so merchants can still see the option exists.

includes/agentic-commerce/README.md
wjrosa: Both default to off on purpose: on upgrade an existing store's feed and checkout routing should not change silently. Auto-exclude drops products from the catalog and auto-redirect changes where shoppers check out, so each is an opt-in a merchant makes deliberately. Happy to change the default if you'd prefer, but I'd lean on keeping both opt-in.

daledupreez: We don't have any live merchants, so I think it's worth having a better default for all future merchants.

client/settings/agentic-commerce/index.js
```diff
@@ -226,6 +245,46 @@ const AgenticCommerceSection = forwardRef( ( props, ref ) => {
...
+								{ isFeatureEnabled && (
+									<CheckboxControl
+										label={ __(
+											'Redirect shoppers to my store for products with add-ons or configurators',
+											'woocommerce-gateway-stripe'
+										) }
+										help={
+											autoExcludeAddons
+												? __(
+														'Add-on / configurator products are excluded from the feed above, so this redirect option does not apply while that is on.',
```
daledupreez: Nit: I think we can improve the wording slightly.
```suggestion
														'Add-on / configurator products are excluded from the feed above, so this redirect option does not apply to the excluded products.',
```

client/settings/agentic-commerce/index.js
```diff
@@ -226,6 +245,46 @@ const AgenticCommerceSection = forwardRef( ( props, ref ) => {
...
+														'woocommerce-gateway-stripe'
+												  )
+												: __(
+														'Instead of excluding add-on / configurator products, keep them discoverable in agents but send shoppers to your store to configure and complete the purchase. Has no effect on products already excluded above.',
```
daledupreez: I think we can also improve the wording for this case.
```suggestion
														'For included add-on / configurator products, ensure that Stripe redirects shoppers to your store to configure the products and make purchases.',
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper.php
```diff
@@ -67,13 +94,40 @@ public function __construct() {
...
+	/**
+	 * Map a product and capture the diagnostic context in the same pass: whether
+	 * it was excluded (and why), and which layer disabled its checkout.
+	 *
+	 * The sync and checkout decisions are hook-driven, and a stateful or
+	 * context-sensitive callback can return a different verdict when re-run. The
+	 * feed preview needs both the row and the reasons behind it, so it must read
+	 * them from this single evaluation rather than re-invoking the filters, which
+	 * could make an advisory disagree with the row it describes.
+	 *
```
daledupreez: We can drop this paragraph.
```suggestion
```

client/settings/agentic-commerce/index.js
```diff
@@ -226,6 +245,46 @@ const AgenticCommerceSection = forwardRef( ( props, ref ) => {
...
 								) }
 
+								{ isFeatureEnabled && (
+									<CheckboxControl
+										label={ __(
+											'Exclude products with add-ons or configurators from the feed',
+											'woocommerce-gateway-stripe'
+										) }
+										help={ __(
+											"Products built with add-on or configurator plugins (Product Add-Ons, Extra Product Options, Composite Products, individually-priced Bundles) have prices that depend on shopper choices, which the feed can't represent. When enabled, those products are kept out of the catalog entirely.",
```
daledupreez: Nit: I think we can improve the description.
```suggestion
											"Products built with add-on or configurator plugins (Product Add-Ons, Extra Product Options, Composite Products, individually-priced Bundles) have prices that depend on shopper choices, which the feed can't represent. When this option is enabled, those products are excluded from the Stripe agentic product catalog.",
```

daledupreez (commented):
I think we should use a better default behaviour for these options given that we don't have any merchants yet.

We also have a linting issue that we need to fix.

Other than that this is looking pretty good, though I didn't fully test the produced feeds.

wjrosa:
Thanks for the review, Dale! Just fixed the lint issue. 

About the default behavior: add-on/configurator products now default to being excluded from the feed (`auto_exclude_addons` is on). Their price depends on shopper choices the feed can't represent, so they're left in; an agent could check out at the wrong price. Since there are no merchants yet, I think that excluding by default is the safe starting point; merchants can opt them back in, and the redirect toggle stays off by default. If you meant redirect-on instead, I can switch it 👍

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
```diff
@@ -1683,24 +1683,47 @@ public function test_product_has_addons_verdict_filter_string_false_is_falsey()
...
 		$product = WC_Helper_Product::create_simple_product();
 		$product->update_meta_data( '_product_addons', [ [ 'name' => 'Engraving' ] ] );
 		$product->save();
 
-		$this->assertTrue( WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product( $product ) );
-		$this->assertNull( WC_Stripe_Agentic_Commerce_Product_Mapper::get_sync_exclusion_reason( $product ) );
+		$this->assertFalse( WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product( $product ) );
+		$this->assertSame( 'addons', WC_Stripe_Agentic_Commerce_Product_Mapper::get_sync_exclusion_reason( $product ) );
 
 		$product->delete( true );
```
daledupreez: Nit: We should use `try/finally` to remove the product.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -101,6 +101,29 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
...
+	 */
+	public const AUTO_EXCLUDE_ADDONS_OPTION = 'wc_stripe_agentic_commerce_auto_exclude_addons';
+
+	/**
+	 * Option key ('yes'/'no', default off) for defaulting `disable_checkout=true`
+	 * on detected add-on/configurator products, so shoppers will be redirected
+	 * to the store for these products.
+	 *
+	 * @var string
+	 * @since 10.9.0
```
daledupreez: Nit: Versions.
```suggestion
	 * @since 11.1.0
	 */
	public const AUTO_EXCLUDE_ADDONS_OPTION = 'wc_stripe_agentic_commerce_auto_exclude_addons';

	/**
	 * Option key ('yes'/'no', default off) for defaulting `disable_checkout=true`
	 * on detected add-on/configurator products, so shoppers will be redirected
	 * to the store for these products.
	 *
	 * @var string
	 * @since 11.1.0
```

daledupreez (approved):
The updates LGTM. I'll address my minor nits and we can then merge this.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-product-mapper-test.php
daledupreez: Fixed in a02caed0ca71899d3ed163381f141a566249dd8c.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
daledupreez: Updated in bf8a023890cd13eb7c01d74971175baa33220a12.

PR #5663, opened 2026-07-02
Disable in-agent checkout on the synced Stripe catalog when Agentic Commerce is turned off

wjrosa:
Fixes STRIPE-1290

Turning off the Agentic Commerce toggle previously only stopped outgoing sync and made the plugin ignore agentic webhooks — it pushed nothing to Stripe. Already-synced products stayed purchasable in-agent, and since Stripe charges the buyer before `checkout.session.completed` arrives, a dropped webhook could take a payment with no matching order.

Stripe has no catalog-delete API, and the true hard stop (disabling the agent) is Dashboard-only. So this PR uses the product feed's per-product "disable checkout" flag as the off switch:

- **On disable (yes→no):** queue a one-off job that re-uploads the catalog with in-agent checkout disabled for every product, redirecting buyers to the store instead of charging cards. Force-uploaded so dedup can't skip a change that's only the checkout flag.
- **Recurring sync now respects the merchant toggle** (`sync_feed()` no-ops while off), so a scheduled run can't re-upload a checkout-enabled catalog and undo the teardown. The teardown push bypasses that gate via `run_feed_sync()`.
- **On re-enable (no→yes):** cancel any teardown queued by a just-prior disable.

**Not a full hard stop:** per Stripe's [seller docs](https://docs.stripe.com/agentic-commerce/for-sellers), the guaranteed kill switch is disabling the agent in the Dashboard, which has no API. This makes products non-purchasable in-agent (redirect); a follow-up could add help text pointing to the Dashboard step.

## Testing instructions

1. Enable the feature flag and the merchant toggle; run a sync so a catalog is uploaded.
2. Turn the toggle **off**. Confirm a `wc_stripe_agentic_commerce_final_feed` action is queued (WooCommerce → Status → Scheduled Actions, group `wc-stripe-agentic-final-feed`) and, once run, the re-uploaded feed marks products checkout-disabled.
3. Confirm the recurring sync no longer uploads while off (scheduled run logs "Sync skipped - merchant toggle disabled").
4. Turn the toggle back **on** before the teardown runs; confirm the pending action is cancelled.
5. `WC_Stripe_Agentic_Commerce_Integration_Test` and `WC_REST_Stripe_Agentic_Commerce_Controller_Test` pass (136 tests).

---

### Changelog entry

Added to `changelog.txt` and `readme.txt` under 10.9.0:

> Update - Take already-synced products out of in-agent checkout when Agentic Commerce is disabled by pushing a final catalog feed, and stop the recurring product sync while it's off

**Post merge**


includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -243,6 +265,68 @@ public function cancel_pending_full_resync(): void {
...
+	 */
+	public function push_final_checkout_disabled_feed(): void {
+		if ( ! $this->is_enabled() ) {
+			return;
+		}
+
+		$force_disable_checkout = static function () {
+			return true;
+		};
+		add_filter( 'woocommerce_agentic_commerce_disable_checkout', $force_disable_checkout );
```
daledupreez: We should add this override using a might higher priority to ensure our filter gets priority.
```suggestion
		add_filter( 'woocommerce_agentic_commerce_disable_checkout', $force_disable_checkout, 99999 );
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -537,15 +621,41 @@ public static function is_checkout_disabled(): bool {
...
+
+	/**
+	 * Generate and upload the catalog feed, skipping the enablement gates.
+	 *
+	 * Extracted from sync_feed() so the teardown push can run while the merchant
+	 * toggle is off. Callers own the enablement checks; this guards delivery setup only.
+	 *
+	 * @since 10.9.0
+	 * @param bool $force_upload When true, bypass the content-hash dedup and always upload.
+	 * @return bool True on successful delivery, false on early returns or failure.
```
daledupreez: Nit: I think we can simplify this comment to remove PR-specific context.
```suggestion
	 * Generate and upload the catalog feed. Callers are responsible for deciding whether
	 * an upload should occur.
	 * 
	 * May be called when the agentic commerce feature has been
	 * disabled to ensure that the uploaded products are marked as unavailable.
	 *
	 * @since 10.9.0
	 * @param bool $force_upload When true, bypass the content-hash deduplication and always upload.
	 * @return bool True on successful delivery, false on early returns or failure.
```

tests/phpunit/admin/class-wc-rest-stripe-agentic-commerce-controller-test.php
```diff
@@ -1165,6 +1165,80 @@ public function test_update_settings_disables_feature(): void {
...
+	/**
+	 * Disabling (yes -> no) must queue the final-feed teardown push.
+	 */
+	public function test_update_settings_disable_schedules_final_feed_push(): void {
+		if ( ! function_exists( 'as_has_scheduled_action' ) ) {
+			$this->markTestSkipped( 'Action Scheduler not available.' );
+		}
+
+		// setUp() leaves the merchant toggle on, so this is a yes -> no transition.
+		as_unschedule_all_actions( WC_Stripe_Agentic_Commerce_Integration::FINAL_FEED_ACTION, [], 'wc-stripe-agentic-final-feed' );
```
daledupreez, line 1390: Should we assert that no actions remain before we proceed with the test? It feels like any lingering actions could give us an invalid test pass.

tests/phpunit/admin/class-wc-rest-stripe-agentic-commerce-controller-test.php
```diff
@@ -1165,6 +1165,80 @@ public function test_update_settings_disables_feature(): void {
 		$this->assertSame( 'no', get_option( WC_Stripe_Agentic_Commerce_Integration::ENABLED_OPTION ) );
 	}
 
+	/**
+	 * Disabling (yes -> no) must queue the final-feed teardown push.
+	 */
+	public function test_update_settings_disable_schedules_final_feed_push(): void {
```
daledupreez, line 1384: Not at all blocking: much of the code across these three test methods is very similar. This might work using a data provider. But it doesn't feel worth refactoring unless you think it will be much simpler.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-integration-test.php
```diff
@@ -586,6 +592,219 @@ public function test_sync_feed_skips_when_disabled() {
...
+	 *
+	 * @return void
+	 */
+	public function test_sync_feed_skips_when_merchant_disabled_but_feature_enabled() {
+		update_option( WC_Stripe_Feature_Flags::AGENTIC_COMMERCE_FEATURE_FLAG_NAME, 'yes' );
+		delete_option( \WC_Stripe_Agentic_Commerce_Integration::ENABLED_OPTION );
+
+		// The merchant gate must short-circuit before any delivery.
+		$http_guard = function () {
+			$this->fail( 'sync_feed() must not reach Stripe delivery while the merchant toggle is off.' );
```
daledupreez: Nit: I think we can state this differently.
```suggestion
			$this->fail( 'sync_feed() must not make an API call to Stripe when the merchant toggle is off.' );
```

daledupreez (commented):
I have some minor feedback from inspection. Will test tomorrow.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Applied — but it surfaced a bug: `remove_filter()` in the `finally` was still on the default 10, so the filter was never removed and leaked into later feed generation. Both are 99999 now, and the test spy moved above it. Fixed in `d05ccc858`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Applied. 👍

tests/phpunit/admin/class-wc-rest-stripe-agentic-commerce-controller-test.php
wjrosa: Good catch — added in `d05ccc858`. Only the schedules-push test could pass falsely (the cancels-push one already asserts its precondition), but I added the guard to the no-op test too for symmetry.

tests/phpunit/admin/class-wc-rest-stripe-agentic-commerce-controller-test.php
wjrosa: Agreed — the three differ in setup and in which direction they assert, so a provider would need setup closures. Leaving as-is.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-integration-test.php
wjrosa: Applied. 👍

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -243,6 +265,68 @@ public function cancel_pending_full_resync(): void {
...
+			return true;
+		};
+		add_filter( 'woocommerce_agentic_commerce_disable_checkout', $force_disable_checkout );
+
+		try {
+			$this->run_feed_sync( true );
+		} finally {
+			remove_filter( 'woocommerce_agentic_commerce_disable_checkout', $force_disable_checkout );
+		}
+	}
```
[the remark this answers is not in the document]
wjrosa: Valid — fixed in `568dba158`. `as_unschedule_all_actions()` only removes a job that's still pending, so one Action Scheduler has already claimed runs regardless, and the push was gated on the feature flag alone. A quick disable→enable published a checkout-disabled catalog to a live store.

`push_final_checkout_disabled_feed()` now bails when `is_merchant_enabled()` is true. Added a test that fails without the guard.

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-commerce-integration-test.php
```diff
@@ -687,6 +695,286 @@ public function test_sync_feed_skips_when_disabled() {
...
+			$this->assertSame( 1, (int) $last_sync['products'], 'Push must run and upload the catalog despite the merchant toggle being off.' );
+			$this->assertTrue( $captured, 'Push must force in-agent checkout off for every product.' );
+
+			remove_filter( 'woocommerce_agentic_commerce_disable_checkout', $spy, 100000 );
+			$this->assertFalse(
+				apply_filters( 'woocommerce_agentic_commerce_disable_checkout', false, $product ),
+				'Push must remove its forcing filter so it does not leak into later feed generation.'
+			);
+		} finally {
+			remove_filter( 'woocommerce_agentic_commerce_disable_checkout', $spy, 20 );
```
daledupreez: Nit: This should match the priority for the addition above. (CodeRabbit flagged this, but not inline.)
```suggestion
			remove_filter( 'woocommerce_agentic_commerce_disable_checkout', $spy, 100000 );
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -54,6 +54,25 @@ class WC_Stripe_Agentic_Commerce_Integration implements IntegrationInterface {
 	 */
 	protected const IMMEDIATE_SYNC_ACTION = 'wc_stripe_agentic_commerce_sync_feed_now';
 
+	/**
+	 * Action Scheduler hook for the one-off checkout-disabled feed pushed on
+	 * disable. Stripe has no catalog-delete API, so this is how already-synced
+	 * products are taken out of in-agent purchase.
+	 *
+	 * @var string
+	 * @since 10.9.0
```
daledupreez: Here and throughout: we should update the version to 11.0.0 (or maybe even 11.1.0) if we won't get this merged this week.
```suggestion
	 * @since 11.0.0
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -563,15 +656,42 @@ public static function is_checkout_disabled(): bool {
...
 		$this->feed_validator = null;
 
 		if ( ! $this->is_enabled() ) {
 			WC_Stripe_Logger::info( 'Agentic Commerce: Sync skipped - feature not enabled' );
 			return false;
 		}
 
+		// Gate the recurring push on the merchant toggle so a scheduled sync can't
+		// re-upload a checkout-enabled catalog after disable and undo the teardown.
+		// push_final_checkout_disabled_feed() bypasses this via run_feed_sync().
```
daledupreez: Nit: We can shorten this comment.
```suggestion
		// Skip the sync if Agentic Commerce is disabled.
```

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
daledupreez: Nit: This comment is unnecessarily verbose.
```suggestion
// Get settings in one fetch to prevent timing issues.
```

daledupreez (approved):
I have some non-blocking comments, but this is looking good to me and testing well. I do think we need to fix the `@since` tags before merging, but this can otherwise be shipped.

daledupreez:
@wjrosa, can you check whether this PR made it into 11.0.0? And can you make sure the changelog entry is in the right section? It looks like this ended up included in 11.0.0, but I am not sure that we recorded the details correctly due to the merge result.

wjrosa:
Sorry about that, Dale. Yes, it shipped in 11.0.0 (merged 2026-09-01). The readme and the release notes list it correctly, but in changelog.txt the entry ended up in the 10.9.0 block after the develop merges. I'll open a small PR to move it to the 11.0.0 block.

PR #5618, opened 2026-06-22
Ensure Adaptive Pricing renders on the checkout page when enabled

Mayisha:
Fixes STRIPE-1235

## Changes proposed in this Pull Request:
PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5568 guarded Adaptive Pricing on initCheckout being present and initCheckoutElementsSdk being absent. That check doesn't hold on current Stripe.js: initCheckoutElementsSdk is actually present, so the condition evaluated false and Adaptive Pricing was wrongly hidden.

This PR removes the capability check from both checkout pages so Adaptive Pricing renders whenever it's enabled:
  
- `client/classic/upe/payment-processing.js` — the Adaptive Pricing branch now keys solely on `isAdaptivePricingEnabled && supportsDeferredIntent`.
- `client/blocks/upe/upe-deferred-intent-creation/payment-elements.js` — `isAdaptivePricingSupported` now keys solely on `isAdaptivePricingEnabled`.

## Testing instructions

- Enable Adaptive Pricing
- Adaptive pricing should work in classic checkout
- Adaptive pricing should work in block checkout

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5664, opened 2026-07-03
Update Stripe API version from clover to dahlia

Mayisha:
Fixes STRIPE-1250, STRIPE-1209

## Changes proposed in this Pull Request:
Moves the plugin from the `clover` Stripe API release train (`2025-09-30.clover`) to `dahlia` (`2026-03-25.dahlia`), including the coordinated breaking changes the dahlia base release introduced for Checkout Sessions.

**API version pins**
- `WC_Stripe_API::STRIPE_API_VERSION` → `2026-03-25.dahlia`
- Agentic Commerce Files API version → `2026-03-25.dahlia;udap_beta=v1`
- Stripe.js release train URL `js.stripe.com/clover/` → `js.stripe.com/dahlia/` across all checkout entry points (classic, Blocks, OC, ECE)

**Checkout Sessions breaking-change migration**

Dahlia renamed the Checkout Sessions `ui_mode` enum and the corresponding Stripe.js / React bindings:
- Server: `ui_mode: 'custom'` → `'elements'` when creating the checkout session
- Classic client: `stripe.initCheckout()` → `stripe.initCheckoutElementsSdk()`
- Blocks client: `CheckoutProvider` → `CheckoutElementsProvider`
- Bumped `@stripe/react-stripe-js` `^5.4.1` → `^6.7.0` and `@stripe/stripe-js` `^8.6.0` → `^9.9.0` (required — the dahlia binding names only ship in these majors; peer deps verified)

Updated the affected Jest suites to the new names.

## Testing instructions
- Adaptive pricing checkout should work correctly on block and classic checkouts
- Regression test: With adaptive pricing disabled, OCS and non-OCS checkout should work on block and classic checkouts.

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Comment 

</details>

**Post merge**

-   [ ] Added testing instructions to the [Release Testing Instructions wiki page](https://github.com/woocommerce/woocommerce-gateway-stripe/wiki/Release-Testing-Instructions) (or does not apply)
-   [ ] Added the [needs docs label](https://github.com/woocommerce/woocommerce-gateway-stripe/labels?q=docs) (or does not apply)
-   [ ] Included this PR in the Release Thread scope (or does not apply)

PR #5665, opened 2026-07-03
Fix checkout Payment Element not rendering on newer Stripe.js builds (Adaptive Pricing)

wjrosa:
## Problem

On the Blocks checkout, the Stripe **Payment Element can fail to render entirely** on newer "dahlia+" Stripe.js builds when Adaptive Pricing is enabled — the shopper is left with no payment form. Classic/shortcode checkout is unaffected.

## Root cause

#5568 added a Stripe.js capability guard that skipped the Adaptive Pricing / Checkout Session path when Stripe.js no longer supports a working `initCheckout()`. #5618 removed that guard. On "dahlia+" builds `initCheckout()` is a **throwing stub**:

- **Classic** wraps `initCheckout()` in `try/catch`, so it falls back to standard Elements and still renders.
- **Blocks** mounts react-stripe-js's `CheckoutProvider`, which calls `initCheckout()` **synchronously** in an effect. That synchronous throw never sets `checkoutState.type === 'error'` — the only state `checkout-form.js` keys its standard-Elements fallback on — so the fallback never fires and the Payment Element does not render.

The plugin pins the **Clover** channel (`js.stripe.com/clover/stripe.js`), where `initCheckout()` works and the guard is inert; the failure only affects the edge-case loads Stripe serves as dahlia. It presents as a **Blocks-only** regression in Payment-Element render rate beginning with the guard removal.

## Fix

Restore the capability guard on **both** surfaces — gate Adaptive Pricing on a working `initCheckout()`, detected by `initCheckoutElementsSdk` being absent (its presence is the reliable dahlia marker, checked without calling the throwing stub):

```js
const stripeSupportsInitCheckout =
    typeof stripe?.initCheckout === 'function' &&
    typeof stripe?.initCheckoutElementsSdk !== 'function';
```

On dahlia builds, Adaptive Pricing now falls back to standard Elements (which render); on the pinned Clover channel the guard is `true`, so Adaptive Pricing is unchanged. Migrating to the replacement `initCheckoutElementsSdk()` API is the longer-term path and is tracked in #5614.

## Telemetry probe

Adds a `wcstripe_stripe_js_capability` Tracks event at Payment Element mount that classifies the loaded build (`clover` / `dahlia` / `legacy`) without calling the throwing stub, plus a `predicted_render_risk` flag for the dahlia + Adaptive Pricing + Blocks combination. This makes the real Clover:Dahlia ratio and the affected population measurable in production. It is best-effort and non-throwing, and no-ops where site tracking is disabled.

> **Note:** the event lands only where WooCommerce site tracking is enabled on the storefront. If build data is needed from all stores, a server-side beacon would be required instead.

## Testing

`npm run test:js` (targeted): 72 passing across 4 suites.

- **Repro** — `adaptive-pricing-initcheckout-fallback.test.js` drives the real `CheckoutProvider` to show the synchronous throw never reaches `checkoutState.type === 'error'` (fallback can't fire), contrasted with the async-failure path that does.
- **Classic** — skips Adaptive Pricing and renders standard Elements on dahlia+ (no orphan Checkout Session); retains a defense-in-depth `try/catch` fallback test.
- **Blocks** — falls back to standard Elements on dahlia+ even with Adaptive Pricing enabled; Clover still uses the Adaptive Pricing container.
- **Probe** — build classification, render-risk flag, once-per-surface dedupe, and never-throws.

Validated across classic and Blocks at the unit level. Not exercised via live E2E in this environment.

## Changelog

Added matching entries to `changelog.txt` and `readme.txt` under 10.9.0.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 10.9.0
+* Fix - Keep the checkout Payment Element visible on newer Stripe.js builds where Adaptive Pricing can't initialize, falling back to the standard element instead of rendering no payment form
 * Fix - Send the admin New Order and customer Processing emails when an asynchronous payment method (iDEAL, Klarna, Bancontact) is confirmed via the deferred webhook path
 * Fix - Use the Amazon Pay custom button size setting on the product, cart, and checkout pages instead of falling back to the Apple Pay/Google Pay size
 * Tweak - Render the Express Checkout button on the cart and checkout from page-bootstrapped data, removing a cart-details request from the critical path to first button render
--- a/client/blocks/checkout-sessions/__tests__/adaptive-pricing-initcheckout-fallback.test.js
+++ b/client/blocks/checkout-sessions/__tests__/adaptive-pricing-initcheckout-fallback.test.js
@@ -1,0 +1,128 @@
+/**
+ * Regression coverage for the OCS "Payment Element rendered" drop after 10.8.3.
+ *
+ * 10.8.3 (#5618) removed the Stripe.js capability guard that used to route
+ * Adaptive Pricing loads to standard Elements when Stripe.js no longer supported
+ * initCheckout(). On "dahlia+" Stripe.js, initCheckout() is a throwing stub.
+ *
+ * These tests exercise the REAL @stripe/react-stripe-js CheckoutProvider (the
+ * component the Blocks Adaptive Pricing path mounts) to demonstrate the asymmetry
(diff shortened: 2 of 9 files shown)


wjrosa:
Superseded by #5664, which migrates the plugin from the Clover to the Dahlia Stripe.js release train (`initCheckoutElementsSdk` + `CheckoutElementsProvider`) and fixes this at the source rather than guarding around the throwing `initCheckout()` stub. Closing in favor of that; #5614 is likewise superseded by #5664.

PR #5674, opened 2026-07-08
Add a filter so extensions can control Adaptive Pricing availability for a cart

malithsen:
Fixes STRIPE-1264

Adaptive Pricing isn't compatible with flows that need to charge the customer again after checkout, for example one-click post-purchase upsells. We already handle this for some extensions by hardcoding exclusions for subscriptions, pre-orders charged upon release, and deposits, but third-party plugins have no way to do the same.

This PR adds a public filter, `wc_stripe_is_adaptive_pricing_supported`, so extensions can opt a cart out of Adaptive Pricing. The current cart is passed to the filter for context.

A couple of things I wanted to highlight

- Adaptive Pricing remains enabled by default. Nothing changes unless a plugin hooks the filter.
- The filter only controls the cart-level decision. The merchant setting, account eligibility, and checkout page checks run first and can't be overridden.

## Testing instructions

1. Enable Adaptive Pricing
2. Add this snippet (eg: via Code Snippets plugin)  `add_filter( 'wc_stripe_is_adaptive_pricing_supported', '__return_false' );`
3. Add a simple product to the cart and go to the checkout page
5. Adaptive Pricing should not be offered to the shopper. (You can verify this by the absence of `wc_stripe_create_checkout_session` AJAX request when loading the page)
6. Remove the snippet and confirm Adaptive Pricing is active again.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -1225,6 +1225,9 @@ class WC_Stripe_Helper {
 	 * - A pre-order product that will be charged upon release.
 	 * - A deposit product.
 	 *
+	 * There's a hook to additionally opt out (but not back in) via the
+	 * `wc_stripe_is_adaptive_pricing_supported` filter.
+	 *
 	 * @return bool True if adaptive pricing is supported for the current checkout, false otherwise.
 	 * @since 10.6.0
 	 */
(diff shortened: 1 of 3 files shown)


includes/class-wc-stripe-helper.php
```diff
@@ -1250,40 +1250,54 @@ public static function is_adaptive_pricing_supported(): bool {
...
+		 *
+		 * This filter runs after the merchant setting, account availability, and checkout page
+		 * checks have passed, so it cannot enable Adaptive Pricing where those gates reject it.
+		 *
+		 * @since 10.9.0
+		 *
+		 * @param bool         $is_supported Whether Adaptive Pricing is supported for the current cart.
+		 * @param WC_Cart|null $cart         The current cart, or null if unavailable.
+		 */
+		return (bool) apply_filters( 'wc_stripe_is_adaptive_pricing_supported', $is_supported, WC()->cart );
```
Mayisha: > so extensions can opt a cart out of Adaptive Pricing

@malithsen if the intention is to allow extensions to disable it only, we need to arrange the logic differently. The current return value with the filter allows enabling/disabling both. As a result, if I use the following filter adaptive pricing is offered for subscription products which is not intended.

```
add_filter( 'wc_stripe_is_adaptive_pricing_supported', '__return_true' );
```

Mayisha (changes requested):

includes/class-wc-stripe-helper.php
malithsen: Thanks. That's a good catch!
Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5674/commits/3823580dd04e62f6bad3a8c7045111f7b59f8dde

Mayisha (approved):
Thanks for the update. LGTM 👍

PR #5531, opened 2026-06-05
Add simple admin stripe payment intents REST endpoint

en-adobrescu:
## Changes proposed in this Pull Request:

Add a new Stripe REST controller and endpoint for querying payment intents endpoint of Stripe API. The controller acts as a proxy, forwarding the received parameters to the remote endpoint. The supported parameters are documented here: https://docs.stripe.com/api/payment_intents/list?api-version=2026-05-27.preview&rds=1

The response received from Stripe API is returned after being filtered against a whitelist.

## Testing instructions
1. Make sure you can send HTTP requests to the backend using a HTTP client. For example, you can do this by setting define( 'WP_ENVIRONMENT_TYPE', 'local' ); in wp-config.php and adding an application password from 'admin' user profile.

2. Using a HTTP client (for example Postman or a browser), 'admin' and the new created password as Basic Auth , send a GET request to:
[WordPress instance URL]/wp-json/wc/v3/wc_stripe/payment_intents
The response should look something like this:
```
{
    "object": "list",
    "has_more": false,
    "data": [
        {
            "id": "pi_3TbL9RJlUF0dQbSB00q0FJS2",
            "amount": "24.60",
            "amount_received": "0.00",
            "currency": "USD",
            "payment_details": {
                "order_reference": "in_1TbKCUJlUF0dQbSBCq67mKfR"
            },
            "status": "requires_payment_method"
        },
        {
            "id": "pi_3TayovJlUF0dQbSB1v86EDwJ",
            "amount": "45.00",
            "amount_received": "45.00",
            "currency": "USD",
            "payment_details": {
                "order_reference": "in_1Taxs4JlUF0dQbSB036F5gea"
            },
            "status": "succeeded"
        }
   ]
}
```
3. The endpoint supports the following parameters:

- limit
- starting_after
- ending_before
- customer
- customer_account
- created

They are documented here: 
https://docs.stripe.com/api/payment_intents/list?api-version=2026-05-27.preview&rds=1

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️)
-   [ ] Tested on mobile (or does not apply)

### Changelog entry

-   [ ] This Pull Request does not require a changelog entry. (Comment required below)

<details>

<summary>Changelog Entry Comment</summary>

#### Message
Add a Stripe REST proxy endpoint for querying payment intents endpoint of Stripe API.

</details>

**Post merge**

[...]

PR #5680, opened 2026-07-10
Add query to admin stripe payment intents rest api endpoint 3

en-adobrescu:

Improves the payment intents endpoint introduced in [Add simple admin stripe payment intents REST endpoint](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5531) by adding support for the *query* parameter, described here: [Look up objects in your Stripe data | Query fields for PaymentIntents
](https://docs.stripe.com/search#query-fields-for-paymentintents).

It also refactors the code by splitting the functionality across a controller (WC_Stripe_REST_Payment_Intents_Controller), an input validation class (WC_Stripe_REST_Validator), and a helper/utility class (WC_Stripe_REST_Helper).

## Testing instructions

1. This functionality can be tested using the same instructions provided in [Add simple admin stripe payment intents REST endpoint](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5531)

2. In addition to the parameters already supported ( *limit*, *starting_after*, *ending_before*, *customer*, *customer_account*, *created*),  the endpoint now also supports a new *query* parameter. This parameter is a list/array of objects, each containing the following keys: *field*, *value*, and *operator* (for example query[0][field]=currency&query[0][value]=usd).

The *operator* key is optional. If it is not provided, a default operator is used for the search.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


daledupreez:
@en-adobrescu, could you resolve the conflicts in this branch? I'd like to get a review, but I don't have recent context for the related changes.

includes/admin/class-wc-stripe-rest-helper.php
```diff
@@ -0,0 +1,96 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ *
+ * Helper class for REST controller.
+ *
+ * @since 10.9.0
+ */
+abstract class WC_Stripe_REST_Helper {
```
daledupreez, line 14: Not blocking, but I am wondering whether code this generic should be in `includes/` rather than `includes/admin`. This applies to both this class and the other abstract helper classes.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -61,20 +61,77 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
+		'latest_charge.balance_transaction.currency'                 => 'strtoupper',
+		'latest_charge.billing_details'                              => '',
+		'latest_charge.payment_method_details.type'                  => 'ucfirst',
+		'latest_charge.payment_method_details.card.last4'            => '',
+		'latest_charge.payment_method_details.card.exp_month'        => '',
+		'latest_charge.payment_method_details.card.exp_year'         => '',
+		'latest_charge.payment_method_details.card.checks.cvc_check' => 'ucfirst',
+		'latest_charge.payment_method_details.card.brand'            => 'ucfirst',
+		'latest_charge.payment_method_details.card.funding'          => '',
+		'latest_charge.payment_method_details.card.country'          => 'strtoupper',
```
daledupreez, line 124: I don't think I am a big fan of the `strtoupper` and `ucfirst` transforms. Given that the Stripe API uses lower case, it feels like keeping those values will be simplest for any downstream developer using the API.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,8 +141,8 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 */
 	public function register_routes() {
 		register_rest_route(
-			$this->namespace,
-			'/' . $this->rest_base,
+			$this->namespace . '/wc_stripe',
+			'/payment_intents(?:/(?P<id>.+))?',
```
daledupreez, line 145: We should continue to use `$this->rest_base` in some form for the API registration.
I also think we should create clearly separate APIs for the baseline `GET /wc_stripe/payment_intents` query API and the more detailed `GET /wc_stripe/payment_intents/(?P<id>.+)` fetch API. This introduces a bunch of downstream complexity that would be easy to avoid if we register independent, dedicated endpoints.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -61,20 +61,77 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
+	protected static $rest_query_args = [
+		'amount'   => 'numeric',
+		'created'  => 'numeric',
+		'currency' => 'token',
+		'customer' => 'token',
+		'metadata' => 'token',
+		'status'   => 'token',
 	];
 
 	protected array $stripe_response_allowed_fields = [
```
daledupreez, line 89: If these are static, is it worth using a `const` for the data?

includes/admin/class-wc-stripe-rest-response-filter.php
```diff
@@ -132,4 +135,60 @@ private static function deep_clone( $obj ) {
...
 	}
+
+	/**
+	 * Format a unix timestamp as a date
+	 *
+	 * @param int $value The timestamp.
+	 *
+	 * @return string
+	 */
+	public static function date_format( $value ) {
```
daledupreez, line 146: I think we should add a more self-descriptive name, as the timestamp requirement is only visible via the PHPDoc.
```suggestion
	public static function format_timestamp_as_date( $value ) {
```

includes/admin/class-wc-stripe-rest-response-filter.php
```diff
@@ -132,4 +135,60 @@ private static function deep_clone( $obj ) {
...
+		return gmdate( 'F j, Y', $value );
```
daledupreez, line 147: We need stronger typing or guards somewhere in this stack, as `gmdate()` only accepts `int` arguments, but nothing enforces that, so invalid arguments will throw type errors. In my mind, that would ideally be at the edge (or closer to the edge) and not in this code.

Also, we should use `wc_date_format()` for the date format.
```suggestion
		$date_format = wc_date_format();
		return date_i18n( $date_format, $value );
```

includes/admin/class-wc-stripe-rest-response-filter.php
```diff
@@ -132,4 +135,60 @@ private static function deep_clone( $obj ) {
...
+	}
+
+	/**
+	 * Format a payment method details as a string
+	 *
+	 * @param object $value The payment method details.
+	 *
+	 * @return string
+	 */
+	public static function payment_method_details_format( $value ) {
```
daledupreez, line 157: We have a lot of logic getting details for payment tokens in `include/payment-tokens/`, which may need some refactoring, but that's something we can address in follow-up PRs.

includes/admin/class-wc-stripe-rest-response-filter.php
```diff
@@ -132,4 +135,60 @@ private static function deep_clone( $obj ) {
...
+
+	/**
+	 * Format a status
+	 *
+	 * @param string $value The payment method details.
+	 *
+	 * @return string
+	 */
+	public static function status_format( $value ) {
+		return ucfirst( str_replace( '_', ' ', $value ) );
```
daledupreez, line 192: This should be something we do on the client, as our API won't match the Stripe API. I don't think we should transform this type of field for this first pass.

includes/admin/class-wc-stripe-rest-validator.php
```diff
@@ -0,0 +1,174 @@
...
+	 * @param string $param The parameter name.
+	 *
+	 * @return bool
+	 */
+	public static function validate_timestamp( $value, WP_REST_Request $request, string $param ) {
+		if ( empty( $value ) ) {
+			return true;
+		}
+
+		$unix_timestamp_pattern = '^\d+$';
```
daledupreez, line 42: Why not declare the value as a regex to start with instead of interpolating it below?
```suggestion
		$unix_timestamp_pattern = '/^\d+$/';
```

daledupreez: Separately, should we enforce that the first character is not 0?
```suggestion
		$unix_timestamp_pattern = '^[1-9]\d+$';
```

includes/admin/class-wc-stripe-rest-validator.php
```diff
@@ -0,0 +1,174 @@
...
+				return false;
+			}
+
+			if ( ! isset( $query_field['operator'] ) ) {
+				continue;
+			}
+
+			$query_field_name      = $query_field['field'];
+			$query_field_type      = $rest_query_args[ $query_field_name ];
+			$query_field_operators = self::QUERY_OPERATORS[ $query_field_type ];
```
daledupreez, line 136: We should guard against keys that don't exist in `self::QUERY_OPERATORS`.
```suggestion
			$query_field_operators = self::QUERY_OPERATORS[ $query_field_type ] ?? null;
```

includes/admin/class-wc-stripe-rest-validator.php
```diff
@@ -0,0 +1,174 @@
...
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * Validator class for REST controller.
+ *
+ * @since 10.9.0
+ */
+abstract class WC_Stripe_REST_Validator {
+	public const QUERY_OPERATORS = [
```
daledupreez, line 14: Can QUERY_OPERATORS be protected or private? I am not sure that it needs to be public.

includes/admin/class-wc-stripe-rest-validator.php
```diff
@@ -0,0 +1,174 @@
...
+	 */
+	public static function validate_timestamp( $value, WP_REST_Request $request, string $param ) {
+		if ( empty( $value ) ) {
+			return true;
+		}
+
+		$unix_timestamp_pattern = '^\d+$';
+
+		if ( is_string( $value ) ) {
+			return preg_match( '/' . $unix_timestamp_pattern . '/', $value ) === 1;
```
daledupreez, line 45: Nit: We should use a Yoda condition here.
```suggestion
			return 1 === preg_match( '/' . $unix_timestamp_pattern . '/', $value );
```

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -250,4 +250,106 @@ public function test_pass_rest_params( $rest_params ) {
...
+	 * @dataProvider provide_rest_query_params
+	*/
+	public function test_query_param( $query_param ) {
+		$admin_id = $this->factory()->user->create( [ 'role' => 'administrator' ] );
+		wp_set_current_user( $admin_id );
+
+		$this->mock_http_call();
+		$response = $this->send_request( [ 'query' => $query_param ] );
+
+		$this->assertSame( 200, $response->get_status() );
```
daledupreez, line 353: Should we not validate that the query params are correctly processed? Just confirming that we have a 200 response status leaves a lot of ways for the code to fail silently.

composer.json
```diff
@@ -71,7 +71,9 @@
         "includes/payment-tokens/interface-wc-stripe-payment-method-comparison.php",
         "includes/payment-tokens/trait-wc-stripe-fingerprint.php",
         "includes/admin/class-wc-stripe-rest-payment-intents-controller.php",
-		"includes/admin/class-wc-stripe-rest-response-filter.php"
+		"includes/admin/class-wc-stripe-rest-response-filter.php",
```
daledupreez, line 74: Nit: Can you please add these files to the array in alphabetical order?

includes/admin/class-wc-stripe-rest-response-filter.php
```diff
@@ -132,4 +135,60 @@ private static function deep_clone( $obj ) {
...
+	}
+
+	/**
+	 * Format a payment destination as a string
+	 *
+	 * @param object $value The payment method details.
+	 *
+	 * @return string
+	 */
+	public static function destination_bank_format( $value ) {
```
daledupreez, line 180: As a separate comment, I am not sure that these formatting helpers belong on a `Response_Filter` class.

daledupreez (commented):

en-adobrescu:
> I have some feedback based on inspection.

Looking at it now, thanks!

PR #5683, opened 2026-07-10
Enable Stripe Dynamic Payment Methods for on-session Optimized Checkout payments

Mayisha:
Fixes STRIPE-1254

On-session Optimized Checkout payments currently create the PaymentIntent with an explicit
`payment_method_types` list. This PR lets eligible OC payments use Stripe **Dynamic Payment Methods** instead: when
Optimized Checkout and PMC are both enabled, the intent is created with
`automatic_payment_methods` so Stripe surfaces every eligible method from the merchant's PMC,
and honors their configuration/ordering.

- **Eligibility gate** (`is_automatic_payment_methods_eligible()`): only on-session OC + PMC
  flows qualify. Saved tokens, express wallets, subscriptions, and delayed-confirmation
  methods (Boleto/OXXO/Multibanco/Cash App) are excluded — those already target a known
  method, or need the type known before confirmation.
- When eligible, the create-intent request drops `payment_method_types` and sends
  `automatic_payment_methods: { enabled: true, allow_redirects: always }`.
- The client Payment Element's exclusion list is mirrored onto the intent
  (`excluded_payment_method_types`) so PMC methods the plugin doesn't support are never
  accepted server-side.
- `return_url` is backfilled when needed, since `allow_redirects: always` with `confirm: true`
  requires one.
- Extracted a shared `DELAYED_CONFIRMATION_PAYMENT_METHODS` constant, now used by both the
  eligibility gate and `is_delayed_confirmation_required()`.
- On-session OC reuses a compatible existing PaymentIntent on retries. Because
  `automatic_payment_methods` is a create-time-only Stripe parameter, DPM-eligible orders now
  skip that reuse and always create a fresh intent — otherwise a retry would re-send
  `payment_method_types` to `/confirm` on a DPM intent and fail. The already-succeeded-intent
  guard (duplicate-charge protection) is preserved ahead of it.

**Scope:** this only applies to stores with Optimized Checkout + PMC and **Adaptive Pricing off**. When Adaptive Pricing is on, `is_checkout_sessions_available()` is satisfied by the same PMC + OC conditions and `process_payment()` routes through `process_payment_with_checkout_session()`, which never reaches the DPM path.

## Testing instructions
- Enable OCS.
- Place and order with card. The payment should be successful.
- Check the logs and confirm that in the payment intent request `automatic_payment_methods` is sent instead of `payment_method_types`.
- Test OCS with other payment methods like redirect based payment, wallet etc. and make sure they work without regression.
-  Trigger a recoverable failure on an OCS card payment (e.g. a decline), then retry with a good card — confirm it still uses `automatic_payment_methods` and succeeds (no `payment_method_types` error on the retry).
- Confirm excluded flows are unaffected and still send `payment_method_types`: saved card,
  express wallet (Apple/Google Pay), subscription, and a voucher method (Boleto/OXXO/Cash App).
- Disable OCS and confirm that non-OCS checkout works without any error/regression.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


Mayisha:
@coderabbitai review

includes/class-wc-stripe-intent-controller.php
```diff
@@ -982,6 +982,29 @@ public function create_and_confirm_payment_intent( $payment_information ) {
 			unset( $request['return_url'], $request['mandate_data'] );
 		}
 
+		// Under Optimized Checkout + PMC, let Stripe surface every eligible method from the merchant's
+		// Payment Method Configuration (Dynamic Payment Methods) instead of the explicit list. The
+		// eligibility decision is made in WC_Stripe_UPE_Payment_Gateway::prepare_payment_information_from_request(),
+		// here we only apply it to the request.
+		if ( ! empty( $payment_information['automatic_payment_methods'] ) ) {
```
wjrosa, line 989: The description says DPM-eligible orders skip intent reuse, "otherwise a retry would re-send `payment_method_types` to `/confirm` on a DPM intent and fail." I can't find that guard, and the failure looks live.

`get_existing_compatible_payment_intent()` is still called unconditionally, and its subset check passes for a DPM intent — Stripe returns the full PMC list, so `array_intersect(['card','klarna',…], ['card'])` matches and the intent is reused. It then hits `update_and_confirm_payment_intent()`, which has no `automatic_payment_methods` branch and still sends `payment_method_types`.

So testing step 5 (decline, then retry) should fail. There's no test covering it either.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3487,6 +3513,12 @@ protected function prepare_payment_information_from_request( WC_Order $order ) {
 			'capture_method'                => $capture_method,
 		];
 
+		// The exclusion list keeps plugin-unsupported PMC methods off the intent, mirroring the client Payment Element.
+		if ( $this->is_automatic_payment_methods_eligible( $selected_payment_type, $is_using_saved_payment_method, (bool) $payment_information['has_subscription'], $express_payment_type ) ) {
```
wjrosa, line 3538: Worth noting how narrow this is: `is_checkout_sessions_available()` needs PMC + OC + auto capture — nearly the same conditions — and with Adaptive Pricing on, `process_payment()` routes to `process_payment_with_checkout_session()` and never reaches here. So DPM only applies to OC + PMC stores with AP off. Worth stating in the description so the testing matrix covers the right stores.

includes/class-wc-stripe-intent-controller.php
wjrosa: Taking this PR over from @Mayisha, so I've fixed this one myself in `3c3da765f`.

DPM-eligible orders now skip reuse and always create a fresh intent, which is what the description already claimed. The already-succeeded guard still runs first, so duplicate-charge protection is unchanged. Added a test that fails without the fix.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Taking this PR over from @Mayisha. Leaving the code as-is here — the narrow reach is correct, not a bug — but I'll add the Adaptive Pricing interaction to the description so the testing matrix targets OC + PMC stores with AP off.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -4181,6 +4181,53 @@ public function test_process_payment_deferred_intent_with_existing_intent() {
 	 * @return void
 	 * @throws Exception If test fails.
 	 */
+	/**
```
diegocurbelo, line 3870: This new test was added between the function `test_process_payment_reuses_successful_payment_intent` and its docblock, we should move the docblock below this new test

diegocurbelo (approved):
The code changes look good, and it tests as described 👍🏼  

Left a small comment on a test docblock, we should fix that before merging.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Moved in 68701eae3.

PR #5684, opened 2026-07-10
Show a clear message when manual capture is blocking Adaptive Pricing

malithsen:
Fixes STRIPE-1267

Adaptive Pricing can't be used together with the "Issue an authorization on checkout, and capture later" setting. When that setting is on, Adaptive Pricing is correctly turned off behind the scenes, but the settings screen only showed a
generic "Adaptive Pricing is currently unavailable" message, giving merchants no idea why or how to fix it.

This updates that message to say specifically that manual capture is the reason, so merchants know which setting to change. This is a text/help change only, it does not alter when Adaptive Pricing actually runs.

Before
CleanShot 2026-07-10 at 14 00 00@2x

Now
CleanShot 2026-07-10 at 13 59 23@2x

## Testing instructions

1. Enable the Optimized Checkout Suite in the Stripe settings.
2. Turn on the "Issue an authorization on checkout, and capture later" setting.
3. Open the Stripe settings and find the Adaptive Pricing option.
4. Confirm the option is disabled and the help text now explains that Adaptive Pricing requires automatic capture and is unavailable while manual capture is enabled.
5. Turn manual capture back off and confirm the message disappears, and Adaptive Pricing becomes available again

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-settings-controller.php
+++ b/includes/admin/class-wc-stripe-settings-controller.php
@@ -232,8 +232,11 @@ class WC_Stripe_Settings_Controller {
 			&& ! $is_ap_enabled
 			&& 'yes' === get_option( 'wc_stripe_show_ocs_only_banner', 'no' );
 
-		$adaptive_pricing_unavailable_reason = WC_Stripe_Helper::get_adaptive_pricing_unavailable_reason();
-		$is_checkout_sessions_available      = null === $adaptive_pricing_unavailable_reason;
+		$adaptive_pricing_unavailable_reason = WC_Stripe_Helper::get_adaptive_pricing_account_unavailable_reason();
+		if ( null === $adaptive_pricing_unavailable_reason && ! WC_Stripe_Helper::is_checkout_sessions_available() ) {
+			$adaptive_pricing_unavailable_reason = 'disabled';
+		}
+		$is_checkout_sessions_available = null === $adaptive_pricing_unavailable_reason;
 
 		$params = [
 			'time'                                  => time(),
--- a/includes/admin/class-wc-stripe-settings-controller.php
+++ b/includes/admin/class-wc-stripe-settings-controller.php
@@ -232,14 +232,8 @@ class WC_Stripe_Settings_Controller {
 			&& ! $is_ap_enabled
 			&& 'yes' === get_option( 'wc_stripe_show_ocs_only_banner', 'no' );
 
-		$is_checkout_sessions_available      = false;
-		$adaptive_pricing_unavailable_reason = 'disabled';
-		if ( WC_Stripe_Helper::is_checkout_sessions_available() ) {
-			$adaptive_pricing_unavailable_reason = WC_Stripe_Helper::get_adaptive_pricing_account_unavailable_reason();
-			if ( null === $adaptive_pricing_unavailable_reason ) {
-				$is_checkout_sessions_available = true;
-			}
-		}
+		$adaptive_pricing_unavailable_reason = WC_Stripe_Helper::get_adaptive_pricing_unavailable_reason();
+		$is_checkout_sessions_available      = null === $adaptive_pricing_unavailable_reason;
 
 		$params = [
 			'time'                                  => time(),
--- a/includes/class-wc-stripe-helper.php
+++ b/includes/class-wc-stripe-helper.php
@@ -1217,6 +1217,28 @@ class WC_Stripe_Helper {
 	}
 
 	/**
+	 * Returns the reason why Adaptive Pricing is unavailable for the store, or null if it is available.
+	 *
+	 * Checks for:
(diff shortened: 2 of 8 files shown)


includes/class-wc-stripe-helper.php
```diff
@@ -1216,6 +1216,28 @@ public static function get_adaptive_pricing_account_unavailable_reason(): ?strin
...
+	 * - Account level restrictions
+	 * - Store level restrictions (manual capture disabled)
+	 *
+	 * @return string|null The reason why Adaptive Pricing is unavailable, or null if it is available.
+	 */
+	public static function get_adaptive_pricing_unavailable_reason(): ?string {
+		if ( ! self::is_checkout_sessions_available() ) {
+			$is_automatic_capture_enabled = ( self::get_stripe_settings()['capture'] ?? 'yes' ) === 'yes';
+			if ( ! $is_automatic_capture_enabled ) {
+				return 'manual-capture';
```
Mayisha: Why don't we add the `manual-capture` reason also in the `get_adaptive_pricing_account_unavailable_reason` method instead of creating this new method? That method also includes the `'amount-mismatch-detected'` case. 🤔

malithsen: I extracted it because it's not an account-level config, but rather a store-level config. Account-level restrictions are verified when onboarding as well.

`amount-mismatch-detected` wasn't there when creating this PR. So given that we already have non-account checks there, maybe we should move `manual-capture` there and rename the method so that it's more generic.

Mayisha: Out of scope for the original issue, but a suggestion. Should we take the opportunity to add messages for the PMC disabled scenario?

Mayisha (commented):
Missed to post this yesterday 🤦‍♀️ 

Changes looks good to me, just have a suggestion about the `get_adaptive_pricing_unavailable_reason` method. It also tests well.

Image

includes/class-wc-stripe-helper.php
malithsen: Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5684/commits/af34a8e4b03ed954fa6e38315a3a8b9c3628f5a7

I didn't rename it because it would require deprecating the existing function, which I think is beyond the scope of the PR.

> Should we take the opportunity to add messages for the PMC disabled scenario?

Yes, makes sense to add a message to that case too.

malithsen:
@Mayisha Thanks for the suggestion. Updated in https://github.com/woocommerce/woocommerce-gateway-stripe/commit/af34a8e4b03ed954fa6e38315a3a8b9c3628f5a7

Will appreciate another review

includes/class-wc-stripe-helper.php
```diff
@@ -1204,6 +1204,11 @@ public static function get_adaptive_pricing_account_unavailable_reason(): ?strin
 			return 'amount-mismatch-detected';
 		}
 
+		// Adaptive Pricing requires automatic capture.
+		if ( 'yes' !== ( self::get_stripe_settings()['capture'] ?? 'yes' ) ) {
+			return 'manual-capture';
+		}
```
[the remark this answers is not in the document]
malithsen: It's an one-line check, so I'll keep it as is

includes/admin/class-wc-stripe-settings-controller.php
```diff
@@ -232,14 +232,11 @@ public function admin_scripts( $hook_suffix ) {
...
-		if ( WC_Stripe_Helper::is_checkout_sessions_available() ) {
-			$adaptive_pricing_unavailable_reason = WC_Stripe_Helper::get_adaptive_pricing_account_unavailable_reason();
-			if ( null === $adaptive_pricing_unavailable_reason ) {
-				$is_checkout_sessions_available = true;
-			}
+		$adaptive_pricing_unavailable_reason = WC_Stripe_Helper::get_adaptive_pricing_account_unavailable_reason();
+		if ( null === $adaptive_pricing_unavailable_reason && ! WC_Stripe_Helper::is_checkout_sessions_available() ) {
+			$adaptive_pricing_unavailable_reason = 'disabled';
 		}
+		$is_checkout_sessions_available = null === $adaptive_pricing_unavailable_reason;
```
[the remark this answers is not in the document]
malithsen: CS and AP are tightly coupled, so at least for now, current behavior is correct. Also this issue is pre-existing and not a result of this PR

Mayisha (approved):
Thanks @malithsen. It looks good to me 👍 

Image

**Note**
I do not see pmc off scenario handled here. However, when pmc is off ocs is also off so this message is displayed which is also correct and makes sense.

Screenshot 2026-07-17 at 12 45 36 AM

PR #5688, opened 2026-07-14
Update stable tag to 10.8.4

diegocurbelo:
This PR updates the stable tag to 10.8.4


jaclync (approved):

jaclync:
For learning purposes, is this stable tag part of the release process?

diegocurbelo:
> For learning purposes, is this stable tag part of the release process?

Correct, during the release we don't update the stable tag, the new version is available for download on wordpress.org immediately but it's not the default and it does not trigger auto-updates until we bump the stable tag.

PR #5692, opened 2026-07-14
Fix Apple Pay address error for provinces entered with a curly apostrophe

malithsen:
Fixes STRIPE-1270

When paying with Apple Pay, checkout can fail because the province field has a curly ("smart") apostrophe instead of a straight one (eg: L’Aquila instead of L'Aquila). 

To validate the address, the plugin looks up the province name in the list of valid Italian provinces to find its code. The curly and straight apostrophes didn't match, the lookup failed, resulting in an error on the payment sheet.

CleanShot 2026-07-14 at 13 44 41@2x

This PR makes that lookup ignore which style of apostrophe is used based on the UTF-8 literals in https://www.unicode.org/Public/security/latest/confusables.txt

## Testing instructions
1. On checkout, select Apple Pay
2. Add a new Address with a curly apostrophe as the shipping address 
eg: 
```
Street: Corso Vittorio Emanuele II
City: L’Aquila
Province: L’Aquila
Postal code:  67100
Country: Italy
```
3. Select the above address as the shipping address
4. On **develop**: payment sheet will show an error. On this branch, you should be able to place the order with no errors
5. (Optional) On Stripe dashboard under payment intent data, you should see that state is mapped to `AQ`

CleanShot 2026-07-14 at 14 01 18@2x

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -61,6 +61,7 @@ xxxx-xx-xx - version 10.9.0
 * Add - Add a wc_stripe_subscription_renewal_blocked_by_radar action hook that fires when Stripe Radar blocks a subscription renewal payment
 * Fix - Include the statement descriptor when creating payment intents for ACSS Debit and BLIK payments
 * Fix - Ensure all WordPress hooks have filter documentation
+* Fix - Accept express checkout addresses whose province is entered with a curly apostrophe, so Apple Pay no longer reject them
 
 2026-07-14 - version 10.8.4
 * Fix - Improve Checkout Session amount integrity
--- a/changelog.txt
+++ b/changelog.txt
@@ -64,7 +64,7 @@ xxxx-xx-xx - version 10.9.0
 * Add - Add a wc_stripe_subscription_renewal_blocked_by_radar action hook that fires when Stripe Radar blocks a subscription renewal payment
 * Fix - Include the statement descriptor when creating payment intents for ACSS Debit and BLIK payments
 * Fix - Ensure all WordPress hooks have filter documentation
-* Fix - Accept express checkout addresses whose province is entered with a curly apostrophe, so Apple Pay no longer reject them
+* Fix - Accept express checkout addresses whose province is entered with a curly apostrophe, so Apple Pay no longer rejects them
 
 2026-07-14 - version 10.8.4
 * Fix - Improve Checkout Session amount integrity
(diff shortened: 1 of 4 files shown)


changelog.txt
```diff
@@ -57,13 +56,12 @@ xxxx-xx-xx - version 10.9.0
...
-* Fix - Fall back to the standard Stripe payment form when another plugin loads an older, incompatible version of Stripe.js, so Adaptive Pricing and Optimized Checkout keep working
 * Update - Move the checkout sessions availability check to WC_Stripe_Helper and remove the deprecated wc_stripe_is_checkout_sessions_available filter
 * Fix - Prevent Adaptive Pricing payments from failing when the checkout return URL is relative
 * Add - Add a wc_stripe_subscription_renewal_blocked_by_radar action hook that fires when Stripe Radar blocks a subscription renewal payment
 * Fix - Include the statement descriptor when creating payment intents for ACSS Debit and BLIK payments
 * Fix - Ensure all WordPress hooks have filter documentation
+* Fix - Accept express checkout addresses whose province is entered with a curly apostrophe, so Apple Pay no longer reject them
 
 2026-07-14 - version 10.8.4
 * Fix - Improve Checkout Session amount integrity
```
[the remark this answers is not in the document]
malithsen: Addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5692/commits/e49fd38b9caa3211d766e21ab090fb6b6baaee07

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
jaclync, line 1484: non-blocking, pre-existing: maybe this fallback could use the `sanitize_string` for comparison as well.

jaclync: nit, non-blocking, pre-existing: as a small optimization, `$sanitized_state_string` could be moved out of the for loop.

jaclync (approved):
LGTM, tested on a Pressable test site (Apple Pay requires a public domain) and verified that the issue was reproducible before and fixed after :shipit:

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
malithsen: Good point. Moved in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5692/commits/a3d0599ff9a2d4d6d69d44161259c0b444f22abe

malithsen: Makes sense to sanitize it as well but I will leave it out of this PR as it's a functional change that requires additional testing.

PR #5702, opened 2026-07-16
Prevent deferred payment methods from using the PaymentIntent creation endpoint

diegocurbelo:
Fixes STRIPE-1278
Fixes #<github_issue_id>

Restricts the checkout PaymentIntent creation endpoint to only payment methods that do not support deferred intent creation.

The existing enabled payment methods lookup is preserved, but methods that support deferred intents (such as card) are filtered out. Currently only ACSS Debit and BLIK can continue creating PaymentIntents.

```
curl -ksS 'https://stripe.local/?wc-ajax=wc_stripe_create_payment_intent' --data-urlencode '_ajax_nonce=<NONCE>' --data-urlencode 'payment_method_type=card'
```

## Testing instructions

- The linear issue has reproduction steps, but you can use [this script](https://mc.a8c.com/includes/img-uploader/files/1784229176-wcstripe-create-payment-intent-repro.mjs) to automate the test:
    ```
    node wcstripe-create-payment-intent-repro.mjs --base_url=https://scallop-of-macaronis.jurassic.ninja
    ```
    In develop, an intent is created/returned:
    ```
    {
      "intent_created": true,
      "fix_behavior_observed": false
    }
    ```
    With this branch no intent is created/returned:
    ```
    {
      "intent_created": false,
      "fix_behavior_observed": true
    }
    ```

- Test that Credit Card still work as expected (regression tests): 
1. Add a product to the cart
2. Navigate to the Blocks checkout
3. Place the order using Credit Card (use 4242424242424242 for example)
4. Verify the checkout completes without error

- Test that ACSS/BLIK still work as expected (regression tests): 
1. Make sure ACSS and BLIK are enabled in the plugin settings
2. Add a product to the cart
5. Navigate to the Blocks checkout
6. Place the order using ACSS/BLIK

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-intent-controller.php
+++ b/includes/class-wc-stripe-intent-controller.php
@@ -401,6 +401,19 @@ class WC_Stripe_Intent_Controller {
 
 		$gateway                 = $this->get_upe_gateway();
 		$enabled_payment_methods = $payment_method_type ? [ $payment_method_type ] : $gateway->get_upe_enabled_at_checkout_payment_method_ids( $order_id );
+		$enabled_payment_methods = array_values(
+			array_filter(
+				$enabled_payment_methods,
+				function ( $payment_method_id ) use ( $gateway ) {
+					$payment_method = $gateway->payment_methods[ $payment_method_id ] ?? null;
+					return $payment_method && ! $payment_method->supports_deferred_intent();
+				}
+			)
+		);
+
+		if ( empty( $enabled_payment_methods ) ) {
+			throw new Exception( __( 'Unable to process your request. Please reload the page and try again.', 'woocommerce-gateway-stripe' ) );
+		}
 
 		$capture = $gateway->is_automatic_capture_enabled();
 		$request = [
--- a/includes/class-wc-stripe-intent-controller.php
+++ b/includes/class-wc-stripe-intent-controller.php
@@ -400,7 +400,10 @@ class WC_Stripe_Intent_Controller {
 		}
 
 		$gateway                 = $this->get_upe_gateway();
-		$enabled_payment_methods = $payment_method_type ? [ $payment_method_type ] : $gateway->get_upe_enabled_at_checkout_payment_method_ids( $order_id );
+		$enabled_payment_methods = $gateway->get_upe_enabled_at_checkout_payment_method_ids( $order_id );
+		if ( $payment_method_type ) {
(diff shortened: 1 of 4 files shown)


includes/class-wc-stripe-intent-controller.php
```diff
@@ -401,6 +401,19 @@ public function create_payment_intent( $order_id = null, $payment_method_type =
 
 		$gateway                 = $this->get_upe_gateway();
 		$enabled_payment_methods = $payment_method_type ? [ $payment_method_type ] : $gateway->get_upe_enabled_at_checkout_payment_method_ids( $order_id );
```
jaclync: Potential edge case: when a `payment_method_type` is passed, `$enabled_payment_methods` is just `[ $payment_method_type ]` and not checked if it's enabled for the store here nor in the filter below. Maybe `payment_method_type` can be checked for its enabled state first if non-null, and throw an error if not enabled?

includes/class-wc-stripe-intent-controller.php
jaclync: ❓ Do you think this function callable from guest checkout could be subject to the same risk as `create_payment_intent` since they both return a client secret from Stripe PI request? If so, maybe they can both filter the enabled & non-deferred payment methods.

jaclync (approved):
Code: Had 2 questions, but I don't think they're blocking.
Testing: I was able to reproduce the carding test with the testing steps in the issue in the console of a checkout page, and verified that it wasn't reproducible anymore in the PR branch. Also tested the 2 regression cases, with BLIK for the non-deferred payment method.

includes/class-wc-stripe-intent-controller.php
diegocurbelo: Makes sense, added in [a752a98](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5702/commits/a752a9887a55017785475305b1cc6c4e30b9ea4b)

includes/class-wc-stripe-intent-controller.php
diegocurbelo: Nice catch, it should have the exact same guard as its used in the same way (just ACSS that uses this flow, as BLIK is not reusable)... added the guard in [cfb3aa6](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5702/commits/cfb3aa6fc28bb8e71db35084ff85e8c82c48396b)

malithsen (approved):
The approach makes sense and code looks good.

PR #5705, opened 2026-07-16
Ignore PHPStan `missingType.generics` errors for `WP_REST_Request` type hints

### LINKED ISSUE #4973: [PHPStan] Prevent `missingType.generics` errors for `WP_REST_Request` usage
The `WP_REST_Request` class implements the `ArrayAccess` interface, which means that PHPStan expects the types of the incoming values to be specified. I don't believe any of our code leverages that feature of the class, so it would be better if we had some way to either default the types or ignore this specific type of error as per [the PHPStan docs for ignoring errors](https://phpstan.org/user-guide/ignoring-errors). Looking at those docs, it would probably be best if we used a regex for `message` in addition to matching the `identifier` field, but that likely requires some experimentation.

jaclync:
Fixes STRIPE-935
Fixes #4973

The WordPress stubs declare `WP_REST_Request` as generic over its parameter types (`@phpstan-template T of array`), so every plain `WP_REST_Request` type hint raises `missingType.generics` at level 8 — the source of all 41 such entries in our baseline.

This PR replaces those entries with a single scoped `ignoreErrors` rule (identifier plus a message regex that only matches `WP_REST_Request`), as suggested in the issue:

- New un-parameterized `WP_REST_Request` hints no longer fail analysis — baseline entries only cover the exact file and method they were generated for, so they never prevented new occurrences; this rule does.
- `missingType.generics` for other generic classes is still reported.
- The generic stays opt-in: a handler can still declare `WP_REST_Request<array{...}>` and get full type checking.

Our handlers treat request params as `mixed` and validate at runtime, so requiring `T` on every handler adds annotation noise without safety — a `T` annotation is a hand-maintained copy of the route's `args` schema, and PHPStan can't warn when the two drift apart.

## Testing instructions

Developer-tooling change only, no runtime behavior; verify via PHPStan:

1. `composer install`, then `npm run phpstan` → `[OK] No errors`.
2. Check the baseline diff: 246 deletions (41 entries, all `missingType.generics` for `WP_REST_Request`), no additions.
3. Prevention: add a temp file in `includes/` with a method typed `WP_REST_Request $request` → still green (this fails on `develop`).
4. Scope: change that param type to `ArrayObject $things` → `missingType.generics` is reported. Delete the temp file.

---

-   [x] Covered with tests (or have a good reason not to test in description ☝️) — static-analysis config only; running PHPStan is the verification.
-   [x] Tested on mobile (or does not apply) — does not apply.

### Changelog entry

<details>

#### Comment

Developer-tooling change only (PHPStan configuration and baseline); no user-facing or runtime behavior change.

</details>

**Post merge**

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1105,70 +1105,10 @@ parameters:
 			path: includes/admin/class-wc-rest-stripe-account-keys-controller.php
 
 		-
-			message: '#^Method WC_REST_Stripe_Account_Keys_Controller\:\:configure_webhooks\(\) has parameter \$request with generic class WP_REST_Request but does not specify its types\: T$#'
-			identifier: missingType.generics
-			count: 1
-			path: includes/admin/class-wc-rest-stripe-account-keys-controller.php
-
-		-
 			message: '#^Method WC_REST_Stripe_Account_Keys_Controller\:\:mask_key_value\(\) has parameter \$value with no type specified\.$#'
 			identifier: missingType.parameter
 			count: 1
 			path: includes/admin/class-wc-rest-stripe-account-keys-controller.php
-
-		-
-			message: '#^Method WC_REST_Stripe_Account_Keys_Controller\:\:set_account_keys\(\) has parameter \$request with generic class WP_REST_Request but does not specify its types\: T$#'
-			identifier: missingType.generics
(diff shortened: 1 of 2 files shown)


Mayisha (approved):
LGTM 👍

PR #5706, opened 2026-07-17
Prevent duplicate orders and charges when an interrupted checkout is resubmitted

### LINKED ISSUE #4878: Duplicate order processing
### Description

When an order is placed on the WooCommerce checkout, and the checkout page is **reloaded before the response completes**, the cart items remain on the checkout page. During this time, the original order **is successfully created in the backend**.

If the user accidentally proceeds to place the order again, **a second duplicate order is created** with the same products.

On a clean WooCommerce installation **without Stripe**, this behavior appears to be handled correctly by the logic introduced in the following PR, which attempts to retrieve and resume the previously created order instead of creating a new one:

https://github.com/woocommerce/woocommerce/pull/40964

However, **when Stripe is enabled**, this recovery behavior does not occur. Instead, WooCommerce allows a new order to be created, resulting in duplicate orders.

### Steps to reproduce

1. Install a clean **WordPress** + **WooCommerce** setup.
2. Enable **Stripe** as the payment gateway.
3. Add products to the cart and proceed to checkout.
4. Click **Place order**.
5. Before the checkout request completes (slow response / throttled network), **reload the checkout page**.
6. Observe that:
    - The order is already created in the backend.
    - The checkout page still contains the same cart items.
7. Click **Place order** again.
8. A **second duplicate order** is created.

### Expected behavior

After reloading the checkout page:
- WooCommerce should detect that an order has already been created for the current checkout session.
- The existing order should be retrieved and resumed (as intended by PR [#40964](https://github.com/woocommerce/woocommerce/pull/40964)).
- A second order **should not** be created.

This is the behavior observed with internal WC gateways, but **not with Stripe**.

### Actual behavior

- The checkout page reloads with the cart still intact.
- WooCommerce does not restore or reuse the previously created order.
- Submitting checkout again creates a **duplicate order**.
- This happens consistently when Stripe is enabled.

### Additional context

- This appears to be a **race condition/idempotency issue** when the checkout POST request is interrupted.
- Stripe’s checkout/payment flow does not appear to integrate with WooCommerce’s order-recovery logic introduced in PR [#40964](https://github.com/woocommerce/woocommerce/pull/40964).
- The issue is more noticeable on **slow servers or delayed payment responses**.

### Notes/Questions for maintainers

- Is Stripe expected to implement its own idempotency handling to align with WooCommerce’s order recovery logic?
- Should WooCommerce core enforce a stronger server-side checkout intent lock to prevent duplicate order creation regardless of gateway?
- Is the recovery logic from PR [#40964](https://github.com/woocommerce/woocommerce/pull/40964) intentionally bypassed for Stripe, or is this a bug?

Mayisha:
Related to woocommerce/woocommerce-gateway-stripe#4878

When Stripe charges the cart but the checkout response never reaches the shopper (page reload, slow network, a hook that throws or corrupts the JSON response), the cart is left intact and the shopper resubmits. Core can't resume the first order because it's already paid, so it builds a second order — and Stripe charges the card again. Core has no way to prevent this on its own: only the gateway knows Stripe already charged this cart.

This adds a gateway-side guard:

* **Marker on payment.** When an order is actually paid (`get_date_paid()`), the gateway records `{cart_hash, order_id, billing_email, paid_at}` in the session and flushes it immediately, so it survives the lost response. Written on both the synchronous path (`process_payment()`) and the browser-side async return for 3DS/SCA and redirect APMs — iDEAL, Klarna, Bancontact (`process_upe_redirect_payment()`).
* **Guard at the top of** `process_payment()`**.** If the order being processed matches a recent, still-paid marker (same cart hash **and** billing email, within the window), the gateway skips payment, cancels the redundant order core just created, empties the cart, and redirects to the order that was already paid.
* **Marker cleared at order confirmation** (`woocommerce_thankyou`, fired by both classic and Blocks). Reaching the confirmation page proves the shopper saw the order succeed, so a later repurchase of the same items is treated as deliberate, not a duplicate.
* **Detection window** defaults to 2 minutes, filterable via `wc_stripe_duplicate_charge_detection_window` (return `0` to disable the guard).

Excluded flows: change-payment-method, pay-for-order, add-payment-method, and Adaptive Pricing checkout sessions (which carry their own guard).

**How can this break / what guards against it:**

* *Blocking a genuine repurchase of an identical cart.* Mitigated by clearing the marker at the confirmation page, scoping by billing email, and the short window; the filter is the escape hatch.
* *Matching a different shopper's order.* Scoped by billing email, not cart hash alone.
* Covered by PHPUnit across the sync path, async/redirect completion, marker clearing, stray-order cancellation, cart survival, and the customer/window/exclusion cases.

**Known limitations (out of scope):**

* Orders completed **only** via webhook (server-to-server, no session) don't get a marker — but that path has no resubmit risk, since the shopper isn't on checkout.
* Authorize-only (manual capture) orders have no `date_paid`, so they aren't covered.

## Testing instructions

1. **Lost response after charge (classic + Blocks).** Add this as a snippet

```
php
   add_action( 'wc_gateway_stripe_process_response', function () {
       throw new \Exception( 'simulated lost response' );
   } );
```

Check out with a test card. You'll see an error. Click Place order again. → No second charge; the redundant order is cancelled (see its order note); you land on the first order's received page. Check the orders list as an admin. Notice that there is a cancelled order with the following order note-

Screenshot 2026-07-17 at 11 16 48 PM

2. Genuine repurchase. Remove the snippet. Place an order and let it reach the confirmation page. Place another order with the same items. → It goes through and is charged normally.
3. 3DS. Use a 3DS-required test card (4000 0025 0000 3155), complete the challenge, and confirm the marker path works on the return leg.
4. Regression test: verify that checkout goes through correctly for different payment methods and the redirected thank-you page is always for the correct order.

### Changelog entry

#### Comment

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -2525,6 +2525,13 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 			WC_Stripe_Logger::info( "Begin processing UPE redirect payment for order $order_id for the amount of {$order->get_total()}" );
 
 			$this->process_order_for_confirmed_intent( $order, $intent_id, $save_payment_method );
+
+			// A 3DS or redirect payment method finishes here, on the browser return from the challenge, rather than inside
+			// process_payment(), so record the marker on this leg too or a lost response could still resubmit the cart.
+			// Pay-for-order settles an existing order, so its cart isn't what was charged and must not be recorded.
+			if ( ! $is_pay_for_order && $order instanceof WC_Order && $order->get_date_paid( 'edit' ) ) {
+				$this->record_paid_cart_marker( $order );
+			}
 		} catch ( WC_Stripe_Payment_Cancelled_Exception $e ) {
 			if ( $order instanceof WC_Order ) {
 				$order_helper->delete_stripe_upe_waiting_for_redirect( $order );
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -317,6 +317,9 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 		// Add a notice about currency conversion when the order currency is different from the store currency on the order details page.
 		add_action( 'woocommerce_order_details_after_order_table', [ $this, 'add_currency_conversion_notice' ], 10 );
 
+		// Retire the duplicate-charge marker once the shopper has seen an order confirmed; both classic and Blocks fire this.
+		add_action( 'woocommerce_thankyou', [ $this, 'clear_paid_cart_marker_for_confirmed_order' ] );
+
 		// Add a notice about currency conversion in the order confirmation emails when the order currency is different from the store currency.
 		add_action( 'woocommerce_email_after_order_table', [ $this, 'add_email_currency_conversion_notice' ], 10, 3 );
 
@@ -1699,6 +1702,33 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 			'billing_email' => (string) $marker['billing_email'],
 			'paid_at'       => (int) $marker['paid_at'],
 		];
+	}
+
+	/**
+	 * Retires the paid-cart marker once the shopper has reached an order's confirmation page.
+	 *
+	 * Reaching the thank-you page is proof the shopper saw the order succeed, so the cart they are now holding is a
+	 * deliberate new purchase rather than a resubmission of a response they never received. Without this a genuine
+	 * repurchase of the same items within the detection window would be mistaken for a duplicate and cancelled.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @param int $order_id ID of the order whose confirmation page is being shown.
+	 * @return void
+	 */
+	public function clear_paid_cart_marker_for_confirmed_order( $order_id ): void {
+		$marker = $this->get_paid_cart_marker();
+
+		if ( null === $marker || (int) $order_id !== $marker['order_id'] ) {
+			return;
+		}
+
+		if ( ! WC()->session || ! method_exists( WC()->session, 'save_data' ) ) {
+			return;
+		}
+
+		WC()->session->set( self::PAID_CART_MARKER_SESSION_KEY, null );
+		WC()->session->save_data();
 	}
 
 	/**
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -1579,7 +1579,9 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 		$paid_order = wc_get_order( $marker['order_id'] );
 
 		// The recorded order has to still be paid; a refunded or cancelled one means the shopper is legitimately buying again.
-		if ( ! $paid_order instanceof WC_Order || ! $paid_order->get_date_paid( 'edit' ) ) {
+		if ( ! $paid_order instanceof WC_Order
+			|| ! $paid_order->get_date_paid( 'edit' )
+			|| $paid_order->has_status( [ OrderStatus::CANCELLED, OrderStatus::REFUNDED, OrderStatus::FAILED ] ) ) {
 			return null;
 		}
 
--- a/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
@@ -1454,7 +1454,11 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 	 * @return array|null An array with result of payment and redirect URL, or nothing.
 	 */
 	public function process_payment( $order_id, $retry = true, $force_save_source = false, $previous_error = false, $use_order_source = false ) {
-		$duplicate_response = $this->maybe_prevent_duplicate_charge( $order_id );
+		// Decided here because the signals it reads decay once payment succeeds: needs_payment() flips, so a paid
+		// order-pay stops looking like one, and the marker is written from the finally block below.
+		$is_paying_for_session_cart = $this->is_paying_for_session_cart( $order_id );
+
+		$duplicate_response = $this->maybe_prevent_duplicate_charge( $order_id, $is_paying_for_session_cart );
 
 		if ( null !== $duplicate_response ) {
 			return $duplicate_response;
@@ -1482,8 +1486,47 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 
 			return $this->process_payment_with_deferred_intent( $order_id );
 		} finally {
-			$this->handle_paid_order_after_checkout( $order_id );
+			$this->handle_paid_order_after_checkout( $order_id, $is_paying_for_session_cart );
 		}
+	}
+
+	/**
+	 * Tells whether this request is paying for the cart the shopper is holding, rather than settling an existing order.
+	 *
+	 * Only meaningful before payment is taken; see the call site in {@see process_payment()}.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @param int $order_id ID of the order being processed.
+	 * @return bool
+	 */
+	private function is_paying_for_session_cart( int $order_id ): bool {
+		return ! $this->is_changing_payment_method_for_subscription()
+			&& ! $this->is_on_add_payment_method_page()
+			&& ! $this->is_settling_an_existing_order( $order_id );
+	}
+
+	/**
+	 * Tells whether this request is settling an order that already exists, as pay-for-order and express checkout do.
+	 *
+	 * Keyed on the order key rather than the request path: paying by key is how WooCommerce settles an existing order
+	 * on both the classic order-pay page and the Store API, so this catches express checkout too, which the classic
+	 * is_valid_pay_for_order_endpoint() detector misses. Such an order's cart hash is its original purchase's.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @param int $order_id ID of the order being processed.
+	 * @return bool
+	 */
+	private function is_settling_an_existing_order( int $order_id ): bool {
+		if ( empty( $_GET['key'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+			return false;
+		}
+
+		$order = wc_get_order( $order_id );
+
+		return $order instanceof WC_Order
+			&& hash_equals( $order->get_order_key(), wc_clean( wp_unslash( $_GET['key'] ) ) ); // phpcs:ignore WordPress.Security.NonceVerification.Recommended
 	}
 
 	/**
@@ -1500,15 +1543,14 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 	 *
 	 * @since 10.9.0
 	 *
-	 * @param int $order_id ID of the order that was processed.
+	 * @param int  $order_id ID of the order that was processed.
+	 * @param bool $is_paying_for_session_cart Whether this request was paying the shopper's cart, decided before payment.
 	 * @return void
 	 */
-	private function handle_paid_order_after_checkout( int $order_id ): void {
-		// These flows pay an existing order or no order at all, so the shopper's cart isn't what was charged and the
-		// order's hash is the original purchase's.
-		if ( $this->is_changing_payment_method_for_subscription()
-			|| $this->is_valid_pay_for_order_endpoint()
-			|| $this->is_on_add_payment_method_page() ) {
+	private function handle_paid_order_after_checkout( int $order_id, bool $is_paying_for_session_cart ): void {
+		// Settling an existing order means the shopper's cart isn't what was charged and the order's hash is the
+		// original purchase's.
+		if ( ! $is_paying_for_session_cart ) {
 			return;
 		}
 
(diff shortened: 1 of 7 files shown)


Mayisha:
@coderabbitai review

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1459,11 +1475,280 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
...
+		}
+
+		$order = wc_get_order( $order_id );
+
+		if ( ! $order instanceof WC_Order || ! $order->get_date_paid( 'edit' ) ) {
+			return;
+		}
+
+		$this->record_paid_cart_marker( $order );
+	}
```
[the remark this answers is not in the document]
Mayisha: Good catch — fixed for the browser-side async paths in 83899ff. `process_upe_redirect_payment()` now records the marker after the intent is confirmed, so 3DS/SCA and redirect APMs (iDEAL, Klarna, Bancontact) are covered on the return leg, symmetric with the synchronous path. Pay-for-order is excluded there since it settles an existing order whose cart isn't what was charged; add/change payment method already return early before that code runs.

I left the webhook `payment_complete()` path alone deliberately: it runs server-to-server with no `WC()->session`, so there's nowhere to write the marker — and it'd be the wrong session anyway. That path also carries no resubmit risk, since if only the webhook completed the order the shopper isn't sitting on checkout to resubmit.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -310,6 +317,9 @@ public function __construct() {
 		// Add a notice about currency conversion when the order currency is different from the store currency on the order details page.
 		add_action( 'woocommerce_order_details_after_order_table', [ $this, 'add_currency_conversion_notice' ], 10 );
 
+		// Retire the duplicate-charge marker once the shopper has seen an order confirmed; both classic and Blocks fire this.
+		add_action( 'woocommerce_thankyou', [ $this, 'clear_paid_cart_marker_for_confirmed_order' ] );
```
wjrosa: AI-assisted:

On Blocks, `woocommerce_thankyou` fires from the Additional Information inner block ([AdditionalInformation.php#L33](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Blocks/BlockTypes/OrderConfirmation/AdditionalInformation.php#L33)): a template customized without it never clears the marker, so a deliberate identical repurchase within the window gets silently swallowed. Maybe clear on `template_redirect` for the `order-received` endpoint instead, so it doesn't depend on template contents?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1459,11 +1475,280 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
...
+
+	/**
+	 * Cancels an order that will never be paid because its cart was charged under a different order.
+	 *
+	 * Core creates the order before handing over to the gateway, so by the time the duplicate is spotted the order
+	 * already exists and would otherwise sit pending: WooCommerce only sweeps unpaid orders up when the hold stock
+	 * setting is non-empty, so on a store that has cleared it the record would linger and read as an abandoned sale.
+	 *
+	 * Cancelling from pending or failed is inert beyond the status change. Stock was never reduced, so the restore
+	 * hook returns early, and the cancelled order email only fires out of processing or on-hold.
```
jaclync, line 1670: ❓ Core also registers the cancelled-order email for pending -> cancelled [here](https://github.com/woocommerce/woocommerce/blob/b1aba6b4622835f9a203c079005e31604e24d9a1/plugins/woocommerce/includes/emails/class-wc-email-cancelled-order.php#L43-L46), along with processing/on-hold -> cancelled triggers. Does it mean cancelling the duplicate order can send the cancellation email to both customer and merchant/admin? Maybe the canceled email can be disabled for such duplicate orders with the `woocommerce_email_enabled_cancelled_order` filter.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1459,11 +1475,280 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
...
+		$window_in_seconds = $this->get_duplicate_charge_detection_window( $order );
+
+		if ( $window_in_seconds <= 0 || time() - $marker['paid_at'] > $window_in_seconds ) {
+			return null;
+		}
+
+		$paid_order = wc_get_order( $marker['order_id'] );
+
+		// The recorded order has to still be paid; a refunded or cancelled one means the shopper is legitimately buying again.
+		if ( ! $paid_order instanceof WC_Order || ! $paid_order->get_date_paid( 'edit' ) ) {
```
jaclync: ❓ from my understanding, `date_paid` remains populated when an order transitions to refunded or cancelled, so this check might not hold for these two status changes. During the detection window, orders like this might continue beyond this check and redirect back to the old order. Maybe an additional check for the order status can be added for the refunded/cancelled edge cases with unit testing coverage?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1459,11 +1475,280 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
...
+	}
+
+	/**
+	 * Returns the recorded paid-cart marker, or null when there isn't a usable one.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @return array|null
+	 */
+	private function get_paid_cart_marker(): ?array {
```
jaclync, line 1758: non-blocking: the marker now has its own lifecycle with schema, session persistence, expiry, matching, and clearing, used from several completion paths. Would it be worth extracting it from this class that is already big? e.g. `record( $order )` / `find_match( $order )` / `clear_for_order( $order_id ))` with focused unit tests. All other conditions/decisions (cancellation, redirect) remain in the gateway. Totally fine skipping this.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1459,11 +1475,280 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
...
+			[
+				'order_id'      => $order_id,
+				'paid_order_id' => $paid_order->get_id(),
+			]
+		);
+
+		$this->cancel_order_superseded_by( $order, $paid_order );
+
+		if ( WC()->cart ) {
+			WC()->cart->empty_cart();
```
jaclync, line 1653: non-blocking, edge case: `empty_cart` can discard cart changes made *after* the duplicate order was created (e.g. the shopper edited the cart in another tab before resubmitting). Is it worth rechecking the live cart hash first?

Maybe this potential edge case can be verified with a unit test.
```suggestion
			$live_cart_hash = WC()->cart->get_cart_hash();
			if ( '' !== $live_cart_hash && hash_equals( $marker['cart_hash'], $live_cart_hash ) ) {
                 WC()->cart->empty_cart();
			}
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1459,11 +1475,280 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
...
+	 * @since 10.9.0
+	 *
+	 * @param int $order_id ID of the order being processed.
+	 * @return array|null A response for the already-paid order, or null to carry on with payment.
+	 */
+	private function maybe_prevent_duplicate_charge( int $order_id ): ?array {
+		// These flows pay an existing order or no order at all, so the shopper's cart isn't what's being paid for and a hash
+		// match would prove nothing about it. Blocking a pay-for-order would strand a customer paying a legitimate invoice.
+		if ( $this->is_changing_payment_method_for_subscription()
+			|| $this->is_valid_pay_for_order_endpoint()
```
jaclync: ❓ might be an edge case scenario, feel free to correct: `is_valid_pay_for_order_endpoint` might not be a reliable pay-for-order detector for the new marker logic, in two independent ways:

1. In the post-payment call site `handle_paid_order_after_checkout`, it returns false once the order no longer `needs_payment()` - which is the case since `handle_paid_order_after_checkout` is called in the `finally` case after a successful payment. Therefore, a successful order-pay payment records a paid-cart marker built from the order's original checkout cart hash. If the shopper still holds the same cart and checks out within the window after a successful order-pay, this guard cancels a genuinely new purchase and redirects them to the invoice they just paid.

2. Store API route for Express Checkout. ECE pays existing orders via `POST /wc/store/v1/checkout/{order_id}`, where `is_wc_endpoint_url( 'order-pay' )` is false. Similar to above, the marker can be mistakenly recorded after a successful EC payment. Additionally, this could be false even before payment (the `needs_payment` check).

Maybe a new helper can be created for all the conditions of "is this request paying the session cart?" that work for all payment paths? Then, it can shared in the cases that currently call the code below, or as a flag upfront that gets passed down.

```
$this->is_changing_payment_method_for_subscription()
			|| $this->is_valid_pay_for_order_endpoint()
			|| $this->is_on_add_payment_method_page()
```

jaclync (commented):
With AI assistance, I spent some time understanding the changes with some testing. Noticed an issue + a general question + some inline comments for consideration:
 
Issue from testing:
> 3. 3DS. Use a 3DS-required test card (4000 0025 0000 3155), complete the challenge, and confirm the marker path works on the return leg.

With the simulated response snippet still enabled, I was able to create any number of duplicate & charged orders if I kept clicking "Place Order" after "Something went wrong. Please contact us to get assistance." error banner shown after a completed 3DS payment. Maybe the marker is missing in this 3D completion path?

--

A general ❓ : what's the feasibility of preventing the duplicate order from being created in the first place?

On the classic checkout, it provides a `woocommerce_create_order` filter that we could potentially short-circuit order creation by returning an existing order ID. Hooking the filter when the paid-cart marker matches would mean no stray order at all: no consumed order number, no cancellation note, and no order-created/cancelled webhooks for ERP or analytics integrations to ingest.

Blocks/Store API doesn't seem to have an equivalent extension point (the draft-order reuse decision doesn't seem filterable), so we'd need this cancel-then-redirect fix. I wonder if it's worth raising this use case for support in core.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2240,6 +2525,13 @@ public function process_upe_redirect_payment( $order_id, $intent_id, $save_payme
 			WC_Stripe_Logger::info( "Begin processing UPE redirect payment for order $order_id for the amount of {$order->get_total()}" );
 
 			$this->process_order_for_confirmed_intent( $order, $intent_id, $save_payment_method );
+
+			// A 3DS or redirect payment method finishes here, on the browser return from the challenge, rather than inside
+			// process_payment(), so record the marker on this leg too or a lost response could still resubmit the cart.
+			// Pay-for-order settles an existing order, so its cart isn't what was charged and must not be recorded.
+			if ( ! $is_pay_for_order && $order instanceof WC_Order && $order->get_date_paid( 'edit' ) ) {
```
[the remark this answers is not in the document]
Mayisha: The caller gates on is_valid_order_received_endpoint(), so subscription payment-method changes and add-payment-method can't reach here and need no exclusion of their own.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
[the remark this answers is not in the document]
Mayisha: The async gap was wider than the redirect leg this thread concluded on. Deferred-intent card 3DS doesn't go through `process_upe_redirect_payment()` — it completes over the `wc_stripe_update_order_status` AJAX endpoint → `process_order_for_confirmed_intent()`, which was recording no marker. Fixed by moving marker recording into that shared completion method (covering both 3DS and redirect APMs), in a `finally` so a thrown response after the charge still records it. So the earlier "resolved" was incomplete — the learned rule that `process_upe_redirect_payment()` is solely responsible isn't accurate.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Good call — done in 2e74575. Switched the marker clearing from `woocommerce_thankyou` to `template_redirect`, keyed off the `order-received` endpoint's order id, so it no longer depends on the confirmation template rendering that inner block. Same path for classic and Blocks. Thanks.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Checked this against core and don't think it bites. `WC_Email_Cancelled_Order` only hooks `processing_to_cancelled` and `on-hold_to_cancelled` — there's no `pending`/`failed → cancelled` trigger in the supported WC range. `cancel_order_superseded_by()` bails unless the order is `pending` or `failed`, so the transition it produces never fires that email. And its recipient is `admin_email`, so it's admin-only regardless, never customer-facing. Leaving as-is; the method docblock already notes the processing/on-hold constraint.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Fixed in a8ce518. The guard now checks status alongside `date_paid` and bails when the recorded order is `cancelled`/`refunded`/`failed` (all keep `date_paid`), so a genuine repurchase goes through. Added a data-provider test over the three statuses, and corrected the inline comment, which had claimed the opposite.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Going to leave this as-is. Emptying here is intentional: once we're redirecting the shopper to the order they already paid for, they're leaving checkout, and we don't try to preserve a cart they built in another tab in the meantime — consistent with how we already treat the cart elsewhere (ECE's `ajax_clear_cart()` empties it unconditionally). The one case a live-hash guard would "save" — a cart edited after the duplicate order was created — is exactly the case where the hash no longer matches, so we'd skip the empty and strand stale items against a completed purchase. Cleaner to just clear.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
Mayisha: Both fixed in 07f4c25. Replaced `is_valid_pay_for_order_endpoint()` on the marker path with a check keyed on the order key — `$_GET['key']` matching the order — computed once at the top of `process_payment()` and threaded down. That's present and validated on both classic pay-for-order and the Store API `POST /checkout/{id}` route ECE uses, and because it's decided before payment it doesn't decay when `needs_payment()` flips. Covers scenario 1 (post-pay order-pay) and scenario 2 (ECE). Added tests for matching-key vs unrelated-key.

Mayisha:

PR #5709, opened 2026-07-20
Reuse the express checkout Elements group on cart/checkout updates instead of re-initialising

Mayisha:
Related to STRIPE-1224

On the cart and checkout pages, every `updated_cart_totals` / `updated_checkout` event ran the full express checkout `init()`, which re-created the Stripe `elements()` group from scratch. On a typical recalc (coupon, quantity, shipping-method change) the only thing that actually moves is the amount — yet re-creating the group re-hit `GET /v1/elements/sessions`, re-probed the Apple Pay / Google Pay capability, and, because the previous group was dropped without being unmounted, stacked duplicate buttons and leaked the abandoned wallet iframes and telemetry. This produced the request flood and button stacking reported in the issue.

This PR makes the cart/checkout path reuse the mounted group:

- **In-place update.** When a group is already mounted and the structural signature — `currency` and `requestShipping`, the only inputs `elements.update()` can't change — is unchanged, the new amount, line items, and derived shipping rates are pushed to every mounted button in place. No new group, no `/v1/elements/sessions`, no wallet re-probe.
- **Rebuild only on a structural change.** A currency or shipping-requirement change (or nothing currently mounted) still does a full rebuild, which now first tears down the old buttons (`destroy()` + container cleanup) so nothing stacks or leaks.
- **Zero total.** A drop to a `0` total hides the buttons rather than pushing an invalid amount, and a return above `0` refreshes the still-mounted group in place.
- **Debounce.** The two handlers share a leading-edge debounce, so a burst of `updated_checkout` (e.g. address fields on classic checkout) coalesces into one Store API cart fetch while first paint stays immediate.

**Why this shape:** the shipping rates are derived from the same display items as the line items, so both are refreshed together — refreshing the items but not the rates would leave the wallet sheet summing to a stale total and the click would be rejected. The in-place path lives inside the cart/checkout branch of `init()`, so the setup-mode change-payment page and pay-for-order (handled in their own branches) never reach `elements.update()`.

**How this can break, and what guards it:** the risk is reusing a group whose structural inputs actually changed. The signature check is exact — both varying inputs (`currency`, `needs_shipping`) come straight from the cart-details response already fetched each update, and anything that changes them forces a rebuild. Unit tests lock in the in-place-vs-rebuild decision, the teardown, and the zero-total path.

## Testing instructions
**Prerequisite**: Stripe enabled with Express Checkout (Apple Pay / Google Pay / Link) turned on for Cart and Checkout, in a browser where a wallet button renders (e.g. Safari for Apple Pay, or Chrome signed in to Google Pay).

1. **Checkout — no repeated sessions.** On classic checkout, open DevTools → Network, filter `elements/sessions`. Change the address / shipping method / quantity a few times. Expected: the buttons stay put (no flicker/stacking) and you do **not** see a new `elements/sessions` per update — the group is reused. (Before: one per update.)
2. **Cart — clean rebuild, correct button.** On the cart page, change quantity / apply a coupon. Expected: the express buttons stay visible and update, with exactly one set of buttons (no duplicates stacking). Note: a single `elements/sessions` per cart update is expected here — WooCommerce rebuilds the cart totals container each update.
3. **No leaks.** Repeat updates on both pages and confirm the number of Stripe wallet iframes / background requests stays flat instead of climbing.
4. **Zero total.** Empty the cart / reach a $0 total.
   - Expected: the express buttons hide. Add an item again → they reappear with the correct amount.
5. **Other pages unaffected.** Confirm express checkout still works on the change-payment-method and pay-for-order pages.

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/entrypoints/express-checkout/__tests__/index.test.js
+++ b/client/entrypoints/express-checkout/__tests__/index.test.js
@@ -352,3 +352,151 @@ describe( 'Express Checkout product page variation breakdown', () => {
 		} );
 	} );
 } );
+describe( 'Express Checkout cart in-place amount update', () => {
+	const cartParams = () => ( {
+		...baseParams(),
+		stripe: {
+			publishable_key: 'pk_test_123',
+			locale: 'en',
+			is_express_checkout_enabled: true,
+		},
+		cart: {
+			total: 1500,
+			currency: 'usd',
+			requestShipping: false,
+			requestPhone: false,
+			displayItems: [],
+		},
+	} );
+
+	// Stripe stub whose elements() factory is a spy, so a full re-init (which
+	// creates a new group) is distinguishable from an in-place elements.update().
+	const stubStripe = () => {
+		const elementsList = [];
+		const button = {
+			on: () => button,
+			mount: jest.fn(),
+			destroy: jest.fn(),
+		};
+		const elementsFactory = jest.fn( () => {
+			const elements = {
+				create: jest.fn( () => button ),
+				update: jest.fn(),
+			};
+			elementsList.push( elements );
+			return elements;
+		} );
+		mockGetStripe.mockReturnValue( { elements: elementsFactory } );
+		return { elementsList, elementsFactory, button };
+	};
+
+	// The cart-details total flows through the mocked transformPrice.
+	const setCartTotal = ( amount ) => {
+		// eslint-disable-next-line global-require
+		require( 'wcstripe/express-checkout/transformers/wc-to-stripe' ).transformPrice.mockReturnValue(
+			amount
+		);
+	};
+
+	// Fire a cart update and drain the debounced handler + the cart-details
+	// promise chain it kicks off.
+	const triggerCartUpdate = async () => {
+		// eslint-disable-next-line global-require
+		require( 'jquery' )( document.body ).trigger( 'updated_cart_totals' );
+		try {
+			await jest.advanceTimersByTimeAsync( 300 );
+		} finally {
+			jest.useRealTimers();
+		}
+	};
+
+	beforeEach( () => {
+		jest.resetModules();
+		jest.useFakeTimers();
+		mockGetStripe.mockReset();
+		mockGetCartDetails.mockReset();
+		mockGetCartDetails.mockResolvedValue( {
+			totals: { total_price: '2500', total_refund: '0' },
+			needs_shipping: false,
+		} );
+		document.body.innerHTML =
+			'<div id="wc-stripe-express-checkout-element"></div>';
+	} );
+
+	afterEach( () => {
+		jest.useRealTimers();
+		delete global.wc_stripe_express_checkout_params;
+	} );
+
+	it( 'pushes the new amount to the existing groups without re-creating them on a second cart event', async () => {
+		global.wc_stripe_express_checkout_params = cartParams();
+		const { elementsList, elementsFactory } = stubStripe();
+		setCartTotal( 2500 );
+
+		loadEntrypoint();
+
+		// First paint mounts one Elements group per express method from the snapshot.
+		const groupsAfterFirstPaint = elementsFactory.mock.calls.length;
+		expect( groupsAfterFirstPaint ).toBeGreaterThan( 0 );
+		elementsList.forEach( ( e ) =>
+			expect( e.update ).not.toHaveBeenCalled()
+		);
+
+		await triggerCartUpdate();
+
+		// No new Elements group was created (no fresh /v1/elements/sessions) ...
+		expect( elementsFactory.mock.calls.length ).toBe(
+			groupsAfterFirstPaint
+		);
+		// ... and every existing group had the new amount pushed in place.
+		elementsList.forEach( ( e ) =>
+			expect( e.update ).toHaveBeenCalledWith( { amount: 2500 } )
+		);
+	} );
+
+	it( 'rebuilds the group when the shipping requirement changes', async () => {
+		global.wc_stripe_express_checkout_params = cartParams();
+		const { elementsFactory, button } = stubStripe();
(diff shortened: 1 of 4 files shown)


Mayisha:
@coderabbitai review

client/entrypoints/express-checkout/index.js
```diff
@@ -84,6 +84,21 @@ jQuery( function ( $ ) {
...
 	let cartBootstrapConsumed = false;
 
+	// Monotonic id for cart-details fetches. The leading+trailing debounce on
+	// cart-update events can leave two fetches in flight, and responses may
+	// arrive out of order — drop any but the latest so a stale total can't
+	// win the last-writer race against the mounted Elements group.
+	let cartFetchGeneration = 0;
+
+	// Structural inputs elements.update() cannot change, so a change in either
+	// forces a rebuild rather than an in-place refresh.
```
daledupreez, line 94: Nit: The comment here should not use `either`, as it's not clear what the referenced items are.
```suggestion
	// Structural inputs elements.update() cannot change, so a change should
	// force a rebuild rather than an in-place refresh.
```

client/entrypoints/express-checkout/index.js
```diff
@@ -168,6 +193,26 @@ jQuery( function ( $ ) {
...
 		},
 
+		// Destroy the previous render's buttons and drop their containers, so a
+		// re-init neither stacks duplicate buttons nor leaks the wallet iframes
+		// the abandoned groups keep alive.
+		teardownExpressCheckout: () => {
+			( wcStripeECE.expressCheckoutElements ?? [] ).forEach(
+				( { button } ) => {
+					try {
+						button.destroy();
```
daledupreez, line 203: Not blocking: just to avoid unnecessary exceptions, it may be worth ensuring button is defined.
```suggestion
						button?.destroy();
```

client/entrypoints/express-checkout/index.js
```diff
@@ -662,19 +718,47 @@ jQuery( function ( $ ) {
...
+						wcStripeECE.refreshExpressCheckoutAmount( {
+							total,
+							displayItems,
+						} );
+						// Only reveal if a container survived: the 'ready' handler removes
+						// containers for methods the browser can't offer, and un-hiding
+						// an empty flex box leaves a blank gap. Separator mirrors hide().
+						if ( wcStripeECE.getElements().children().length ) {
+							wcStripeECE.show();
+							wcStripeECE.getButtonSeparator().show();
```
daledupreez, line 749: I think we may have a subtle race condition here: there is no guarantee that the `ready` callback has been triggered yet. For the express checkout elements that would likely result in a flash, but for the separator, we'd have a lingering element that shouldn't be shown. We need some better tracking for that state, and possibly some shared code for ensuring that elements get (re)shown.

client/entrypoints/express-checkout/index.js
```diff
@@ -168,6 +193,26 @@ jQuery( function ( $ ) {
 			wcStripeECE.getButtonSeparator().hide();
 		},
 
+		// Destroy the previous render's buttons and drop their containers, so a
+		// re-init neither stacks duplicate buttons nor leaks the wallet iframes
+		// the abandoned groups keep alive.
+		teardownExpressCheckout: () => {
+			( wcStripeECE.expressCheckoutElements ?? [] ).forEach(
```
daledupreez, line 200: Is there a way for us to detect when we have an express checkout in flight and we trigger any events that call this new code?

_I used an AI tool to identify this as a possible issue._

client/entrypoints/express-checkout/index.js
```diff
@@ -436,6 +475,17 @@ jQuery( function ( $ ) {
 
 			wcStripeECE.renderButton( eceButton, expressPaymentType );
 
+			// A page can mount several express buttons (Apple Pay, Google Pay, …),
+			// each with its own Elements group. Retain each button alongside its
+			// group and click-closure options so a cart update can refresh every
+			// one, and a structural change can tear them all down.
+			wcStripeECE.expressCheckoutElements.push( {
```
daledupreez, line 482: By moving this push() until after the `wcStripeECE.createButton()` and `wcStripeECE.renderButton()` calls, any errors or exceptions in those two functions will prevent us from adding the element to `wcStripeECE.expressCheckoutElements`. I think we need to consider how we can ensure we guarantee the entry will be there. That may involve additional guards, or it may involve pushing incomplete data to start with and then augmenting it.

_I used an AI tool to identify this as a possible issue._

daledupreez (changes requested):

client/entrypoints/express-checkout/index.js
diegocurbelo: Agree, and also both the buttons container and the separator are initially rendered with `display:none` inline (`class-wc-stripe-express-checkout-element.php`, and with no wallet buttons availableit results in the ”— OR —” separator to be shown on every `updated_checkout` after the first render. This was not a probnlem before because they were only ever shown by the ready handler.

client/entrypoints/express-checkout/__tests__/index.test.js
```diff
@@ -352,3 +352,349 @@ describe( 'Express Checkout product page variation breakdown', () => {
...
+
+		await triggerCartUpdate();
+
+		expect( button.destroy ).toHaveBeenCalled();
+		expect( elementsFactory.mock.calls.length ).toBeGreaterThan(
+			groupsAfterFirstPaint
+		);
+	} );
+
+	it( 'hides the buttons when the cart total drops to zero', async () => {
```
diegocurbelo, line 485: Shouldn't we also cover the "hidden at $0, then show() when the total changes" case?

Mayisha:

PR #5719, opened 2026-07-21
Agentic Commerce: reject out-of-bounds line item quantity and unit_amount in shipping packages

wjrosa:
Towards STRIPE-986

Follow-up to #5551, addressing [this review comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5551#discussion_r3623607250).

The customize_checkout event line item getters only cast payload values to `int`, so a malformed payload could feed a zero or negative quantity, or a negative `unit_amount`, into the shipping package's line totals — corrupting content-dependent rate calculations (`[qty]`/`[cost]` expressions, contents_cost).

This PR makes `WC_Stripe_Agentic_Shipping_Package_Builder::build_contents_from_event()` throw on out-of-bounds values (quantity < 1, unit_amount < 0), consistent with its existing behavior for unresolvable products and missing catalog prices. The webhook handler surfaces the exception as a failed customization request instead of returning corrupted quotes.

The order-mapping path gets a matching guard in `map_line_items()`: its line-total reconciliation (WC catalog price × quantity vs Stripe's pre-tax total) catches inconsistent quantities, but not a payload internally consistent with a non-positive quantity (e.g. quantity 0 with amount_total 0), so those now throw before the order is created.

**BC impact:** none — this only tightens validation of webhook payload values inside a private flow; no public signatures, hooks, or options change.

## Testing instructions

1. Run `npm run test:php -- --filter WC_Stripe_Agentic_Shipping_Calculator_Test` — the new `test_throws_on_out_of_bounds_line_item_values` cases cover zero quantity, negative quantity, and negative unit_amount.
2. Optionally, send a `v1.delegated_checkout.customize_checkout` webhook (see #5551's testing instructions) with `"quantity": -2` on a line item — the request should fail instead of returning shipping options.

---

### Changelog entry

Changelog entries added to `changelog.txt` and `readme.txt` under 10.9.0.

<details>

#### Comment

</details>

**Post merge**

### DIFF
--- a/includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
+++ b/includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
@@ -70,7 +70,7 @@ class WC_Stripe_Agentic_Shipping_Package_Builder {
 	 * @param WC_Stripe_Agentic_Customize_Checkout_Event $event    The customization hook event.
 	 * @param string                                     $currency The three-letter currency code.
 	 * @return array Cart-item-format entries keyed by line item ID.
-	 * @throws Exception When a line item's sku_id cannot be resolved to a product, or the product has no catalog price to fall back on.
+	 * @throws Exception When a line item's sku_id cannot be resolved to a product, its quantity or unit_amount is out of bounds, or the product has no catalog price to fall back on.
 	 */
 	public static function build_contents_from_event( WC_Stripe_Agentic_Customize_Checkout_Event $event, string $currency ): array {
 		$contents = [];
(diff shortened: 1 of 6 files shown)


includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
```diff
@@ -97,6 +97,20 @@ public static function build_contents_from_event( WC_Stripe_Agentic_Customize_Ch
 			$quantity    = $line_item->get_quantity();
 			$unit_amount = $line_item->get_unit_amount();
 
+			// The event getters only cast to int, so a malformed payload can
+			// carry a zero/negative quantity or a negative unit_amount; either
+			// would corrupt line totals, so fail the request instead.
+			if ( $quantity < 1 || ( null !== $unit_amount && $unit_amount < 0 ) ) {
```
jaclync, line 103: non-blocking question: from PR description

> The order-mapping path (build_contents_from_order()) is unchanged: its quantities and totals come from already-validated WC order items.

I'm guessing the quantity & unit amount validation here is specific to the shipping requirements, and doesn't apply to the existing validation on the WC order items mentioned here?

jaclync (approved):
LGTM, tested locally and worked as expected :shipit:

includes/agentic-commerce/class-wc-stripe-agentic-shipping-package-builder.php
wjrosa: Correct — this covers the event path, where the values come raw off the webhook with only an int cast. The order path relies on `map_line_items()` reconciling catalog price × quantity against Stripe's totals, which misses a payload internally consistent with qty ≤ 0 — added a matching bounds check there in 128c91d4d.

jaclync (approved):

PR #5726, opened 2026-07-22
Record the Stripe charge-captured flag for asynchronously confirmed orders so refunds work

### LINKED ISSUE #5699: `_stripe_charge_captured` is not set for asynchronously confirmed orders - wp-admin refunds fail and a Dashboard refund cancels the order
When a payment's charge is created after checkout (e.g. ACH `us_bank_account` with microdeposit verification, where the intent is `requires_action` at checkout with no charge and only later transitions to `succeeded`), `_stripe_charge_captured` is never written:

* `process_response()` (the meta's writer) runs at checkout, when no charge object exists yet.
* The async completion webhook paths (`charge.pending` → on-hold "Stripe charge awaiting payment", then `charge.succeeded` → "Stripe charge complete … via webhook") do not set the meta.

Two flows read the flag and both result in unexpected behavior with it missing:

1. **wp-admin refunds fail** (`process_refund()`): `is_stripe_charge_captured()` → `false` skips the refunds API call as above and results in a generic error.
2. **A Stripe-Dashboard refund cancels the entire order** (`process_webhook_refund()`): the `charge.refunded` webhook hits the uncaptured branch, which assumes a voided pre-authorization - it cancels the order, adds a "Pre-Authorization … voided from the Stripe Dashboard" note, and creates no refund record.

**Steps to reproduce** (verified locally on WC 10.9.4 `develop` branch, test mode):

1. Check out with a test US bank account via the microdeposit flow (order ends `pending`, intent `requires_action`, no charge yet).
2. Complete verification via the hosted microdeposit link (test descriptor code `SM11AA`) and wait for the webhooks: order moves to `processing` and the `py_...` charge ID is back-filled.
3. **wp-admin refund fails:** open the order and issue any partial refund via `Refund $# via ACH Direct Debit` → the generic error above, no refund created, nothing logged, while Stripe shows the charge as refundable.
4. **Dashboard refund cancels the order:** with webhooks delivering, refund any partial amount for the same payment from the Stripe Dashboard → the order status changes to `cancelled`, a "Pre-Authorization … voided from the Stripe Dashboard" note is added, and no WC refund record exists, while Stripe shows the charge `succeeded` with the partial amount refunded.
5. Optional - causation check (ensure to do this last, or on a second order as it removes the repro): set `_stripe_charge_captured` to `yes` directly, matching Stripe's actual state, and the identical wp-admin refund succeeds and the Dashboard refund is recorded normally.

**Impact:** merchants cannot refund async-confirmed bank-debit orders via Stripe from wp-admin ("Refund manually" still works but moves no money), and refunding from the Stripe Dashboard instead cancels the order outright with no refund record - order status, customer communication, and accounting all diverge from Stripe. The same generic wp-admin error string also appears pre-verification (where refund failing is correct, since no charge exists) so merchants cannot distinguish "refund impossible" from "refund should have worked".

**Fix directions to evaluate:** set the captured meta in the async completion webhook path(s); and/or have `process_refund()` consult the retrieved charge/intent state rather than only order meta (it already fetches the intent); plus an explicit return value and log entry on the currently-silent path.

jaclync:
Fixes STRIPE-1280
Fixes #5699

> **A note on diff size:** most of this PR is test coverage — ~515 of the added lines are tests (every writer/reader path plus regressions), a few are changelog. The production change is ~165 insertions across 4 files.

For asynchronously confirmed payments (e.g. ACH with microdeposit verification), the charge is created after checkout, so `_stripe_charge_captured` is never recorded. Both refund flows misread the missing flag:

1. **wp-admin "Refund via Stripe" fails**: `process_refund()` treats missing as "not captured", silently skips the refunds API call, and surfaces a generic error.
2. **A Stripe-Dashboard refund cancels the order**: the `charge.refunded` webhook takes the voided-pre-authorization branch and cancels the order instead of recording the refund.

### The fix

- **Record the flag when the charge first appears** (`WC_Stripe_Webhook_Handler::process_payment_intent()`, `::process_webhook_charge_succeeded()`): the `payment_intent.processing` and `charge.succeeded` webhooks now write it — for async orders these deliver the first charge that ever exists.
- **`charge.refunded` repairs the flag from its own payload** (`WC_Stripe_Webhook_Handler::process_webhook_refund()`; the payload is the charge itself, so authoritative) before the branch that could cancel a paid order. This also heals orders created before the fix.
- **`process_refund()` resolves a still-missing flag from Stripe** (`WC_Stripe_Payment_Gateway::process_refund()` → `::resolve_charge_captured_state()`): one read-only GET `/charges/{id}`, only when the flag was never written (`''`; an explicit `no` is trusted with no API call), persisted so it runs at most once per order. This is what makes the fix reach pre-existing orders, stores with broken webhook delivery, and refunds racing the completion webhook. Please feel free to share if you think it's better not to include this repair for pre-existing orders as it requires a network request to Stripe.

### Guard rails the fix requires

- **`cancel_payment()` voids only charges explicitly recorded as uncaptured** (strict `no`, matching `capture_payment()`). Its cancelled/refunded status hooks also fire on "Refund manually"; with refund-time resolution, a missing flag there could have turned a captured charge into an unrequested full refund. Trade-off: an uncaptured order with a missing flag is no longer voided on cancellation — the authorization expires on its own, moving no money.
- **`process_refund()` rejects negative amounts up front with an explicit `WP_Error`** (`WC_Stripe_Helper::get_stripe_amount()` strips the sign, so an unguarded negative amount would refund the absolute value on a captured order). The guard sits before the captured-state resolution, so a doomed negative refund can no longer trigger the charge lookup either.

### Refactoring (no behavior change)

- New single writer `WC_Stripe_Order_Helper::sync_stripe_charge_captured()`; every path that receives a charge object routes through it (checkout response, the webhooks above, refund-time resolution, charge-ID recovery). The one remaining direct write is `capture_payment()`'s post-capture `yes`.
- New `WC_Stripe_Order_Helper::is_stripe_charge_authorized_only()` names the strict-`no` check the capture/void flows share.
- Fixed a `$charge` variable shadowing in `process_webhook_capture()`; typed `process_webhook_refund()`'s param as `stdClass` (PHPStan baseline shrinks rather than grows).

### Behavior notes

- `process_response()` handed a chargeless intent used to record a blind `no`; it now records nothing and the refund path resolves the real state. Card intents always carry their charge at `process_response()` time, so no real flow changes. (Six REST terminal-capture test mocks gained the `captured` field real charges always carry.)
- The pre-existing silent path for a known-uncaptured charge with nothing to void (implicit `null` → WooCommerce's generic refund failure) is deliberately unchanged as out of scope.
- Pre-existing orders are healed on their next refund; no migration.

**How can this code break?** The riskiest spots are the `''` vs `no` distinction and the `charge.refunded` repair overriding stored state. Guards: unit tests for every writer/reader path (including that genuine pre-authorization voids still cancel the order), the resolver only fires when the flag was never written, and the repair trusts the charge payload — Stripe's actual state.

**Backward compatibility impact:** No hooks, options, or existing public signatures changed or removed; two additive public helper methods. `process_refund()` returns `WP_Error` on two new failure paths (captured-state resolution failure, negative refund amount), consistent with its documented `bool|WP_Error` contract. `cancel_payment()` no longer attempts a void when the flag is missing (previously a silent no-op for captured charges); explicit `no` behaves as before. No cart/session reads on the touched webhook paths; nothing multisite- or install-layout-sensitive changed. Not tested under multisite.

## Testing instructions

Full async repro (test mode, webhooks delivering — e.g. `npm run listen`):

1. Check out with a test US bank account via the microdeposit flow. Click "Enter bank details manually" and auto-fill, other test banks verify immediately and won't reproduce the issue --> in wp-admin, the order should have "pending payment" status and no charge yet.
2. Find the hosted verification link: in the Stripe Dashboard open the payment's **Events**, select the last `payment_intent.requires_action` event, and drill into `next_action` → `verify_with_microdeposits` → `hosted_verification_url` (Stripe also emails it to the billing address).
3. Open it, enter the test descriptor code `SM11AA`, and wait for the webhooks --> in wp-admin, the order first moves to "on hold" (charge created, awaiting ACH settlement) and then to "processing" once the test-mode payment settles a few seconds later, with the charge ID back-filled.
4. In wp-admin order, click **Refund $# via ACH Direct Debit** for a partial refund --> it used to fail with a generic error on `develop`, now it should succeed with the Stripe refund ID in the order note.
5. Refund another amount from the Stripe Dashboard --> it used to cancel the order on `develop`, now it should create a refund record as expected.

Quicker card-based regression test:

1. Place a test card order, then blank its `_stripe_charge_captured` meta (recreates the async state).
2. **Refund via Stripe** → succeeds, and the meta is repaired to `yes`.

## Example screencasts

Shopper flow:

https://github.com/user-attachments/assets/833e3024-58d6-4a98-9eb2-de894008ca8e

Merchant refund flow:

https://github.com/user-attachments/assets/d5aae217-52c5-4808-ae82-141172e9c0ad

---

### Changelog entry

**Post merge**


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -951,8 +953,18 @@ public function process_webhook_refund( $notification ) {
...
+
+			// Repair the stored captured flag from the payload (the charge itself) when absent
+			// or stale, so the uncaptured branch below can't cancel a paid order as a voided
+			// pre-authorization.
+			$was_captured = $order_helper->is_stripe_charge_captured( $order );
+			$captured     = $order_helper->sync_stripe_charge_captured( $order, $charge_payload ) ?? $was_captured;
+
+			if ( $captured !== $was_captured ) {
+				$order->save();
+			}
```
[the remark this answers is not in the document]
jaclync: Confirmed and fixed in f836d89e7. Before, the save check compared boolean projections, so the `''→'no'` repair only persisted when the order is saved later, which an already-cancelled order misses from an early return. Now it compares the stored string before/after syncing the same way, with a regression test for that path. The `→'yes'` transitions were unaffected, so this was a latent unsaved repair, not misbehavior.

jaclync:
@coderabbitai review

@claude please review this PR, try finding any bugs and security issues from the PR changes

changelog.txt
```diff
@@ -72,12 +71,12 @@ xxxx-xx-xx - version 10.9.0
...
 * Fix - Represent negative fees and taxes as discounts in Level 3 data so manually capturing an order edited to include a negative fee no longer fails
 * Fix - Ensure all WordPress hooks have filter documentation
-* Fix - Restrict payment intent creation to payment methods that do not support deferred intent
 * Fix - Clear the Adaptive Pricing order total error notice once a later update succeeds
 * Fix - Accept express checkout addresses whose province is entered with a curly apostrophe, so Apple Pay no longer rejects them
 * Update - Show a clear message when manual capture is blocking Adaptive Pricing from being enabled
 * Fix - Lock the order while confirming a 3DS card payment so a concurrent webhook can't complete it twice, duplicating stock reduction, order notes, and emails
 * Fix - Store the Stripe refund ID on each WooCommerce refund record so orders with multiple partial refunds retain every refund ID
+* Fix - Record the charge-captured state for asynchronously confirmed payments (e.g. ACH with microdeposit verification) so refunds from wp-admin succeed and refunds from the Stripe Dashboard no longer cancel the order
```
wjrosa: I think we should try to reduce the entries a bit, it is too long

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -624,10 +624,11 @@ public function process_response( $response, $order ) {
 
 		$order_id     = $order->get_id();
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
-		$captured     = isset( $response->captured ) && $response->captured;
 
-		// Store charge data.
-		$order_helper->set_stripe_charge_captured( $order, $captured );
+		// Record the captured flag while the charge is in hand; refund and capture flows read
```
wjrosa: We should try to reduce this and other comments, AI (in case you used it) tends to be very verbose

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -1489,6 +1510,50 @@ private function recover_charge_id_from_intent( WC_Order $order ): string {
 		return $charge_id;
 	}
 
+	/**
+	 * Resolves whether the order's charge is captured, asking Stripe when the stored flag is missing.
```
wjrosa, line 1514: Same as my previous comment, this doc block is too long

includes/class-wc-stripe-order-handler.php
```diff
@@ -420,9 +420,11 @@ public function cancel_payment( $order_id ) {
...
 
 		if ( WC_Stripe_Helper::payment_method_allows_manual_capture( $order->get_payment_method() ) ) {
-			$captured = WC_Stripe_Order_Helper::get_instance()->is_stripe_charge_captured( $order );
-
-			if ( ! $captured ) {
+			// Only a known authorize-only charge is voided: this hook also fires on transitions
+			// made without wanting a gateway refund (e.g. "Refund manually"). With the flag merely
+			// missing, process_refund() would resolve it from Stripe and, if the charge turns out
+			// to be captured, issue a full refund nobody asked for.
+			if ( WC_Stripe_Order_Helper::get_instance()->is_stripe_charge_authorized_only( $order ) ) {
```
wjrosa, line 435: Non-blocking question: with `process_response()` no longer writing `'no'` for intent-shaped responses, could a manual-capture order ever reach cancellation with the flag unrecorded? It would now silently skip the void and let the pre-auth expire.

tests/phpunit/class-wc-stripe-payment-gateway-test.php
```diff
@@ -667,16 +667,45 @@ public function test_process_refund_on_zero_amount() {
 	}
 
 	/**
-	 * Tests that process_refund returns false for negative amounts.
+	 * Tests that a failed captured-state lookup surfaces the lookup error instead of
+	 * being treated as an uncaptured charge (which would suggest a manual refund for
+	 * money Stripe may still hold).
 	 */
-	public function test_process_refund_fails_on_negative_amount() {
+	public function test_process_refund_fails_when_captured_state_cannot_be_resolved() {
```
wjrosa, line 674: This replaces `test_process_refund_fails_on_negative_amount` — worth keeping negative-amount coverage too. Related: with the flag missing, a negative amount now triggers the charge fetch before any amount validation.

wjrosa (commented):
Solid fix — the single-writer consolidation, the ''-vs-'no' distinction, and the thorough tests all read well. Two small non-blocking comments inline. Also verified: `$charge` in `process_webhook_capture()` is in scope, `process_response()`'s downstream `$captured` is unchanged, and WC core handles the new `WP_Error` return.

changelog.txt
jaclync: Shortened in 1a51ee25e (and the matching readme.txt entry).

includes/abstracts/abstract-wc-stripe-payment-gateway.php
jaclync: Yup, trimmed in 1a51ee25e.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
jaclync: Yes, condensed in 1a51ee25e.

tests/phpunit/class-wc-stripe-payment-gateway-test.php
jaclync: Updated in daa6e4412, now it throws an error on invalid refund amount up front. I'd actually added this amount guard + test earlier (fd25b2ff2) but reverted it (2d2b21944) to keep the PR scope smaller, since the actual negative amount issue in `get_stripe_amount` is pre-existing (`test_process_refund_fails_on_negative_amount` passing previously was not because of any guard on the amount). I was planning on creating a separate issue, but since you also mentioned this and an extra charge fetch request is introduced in this PR without this guard, I thought it's worth adding it here. Feel free to let me know if it's better handling it separately.

tests/phpunit/class-wc-stripe-payment-gateway-test.php
```diff
@@ -667,16 +667,79 @@ public function test_process_refund_on_zero_amount() {
...
+	 * Tests that process_refund rejects negative amounts without contacting Stripe.
+	 * WC_Stripe_Helper::get_stripe_amount() strips the sign, so an unguarded negative
+	 * amount would refund the absolute value instead of failing.
 	 */
 	public function test_process_refund_fails_on_negative_amount() {
 		$order = WC_Helper_Order::create_order();
+		$order->set_currency( 'USD' );
 		$order->set_transaction_id( 'ch_123' );
+		$this->updateOrderMeta( $order, '_stripe_charge_captured', 'yes' );
 		$order->save();
```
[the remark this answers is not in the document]
jaclync: Good catch - updated the test to a data provider in a13224812 covering the stored states 'yes', 'no', never-recorded, and an unexpected value.

includes/class-wc-stripe-order-handler.php
jaclync: In practice, it shouldn't - manual capture is card-family only, and cards confirm synchronously, so `process_response` sees the latest charge and records `'no'` at checkout. If the flag were somehow missing anyway, the skipped void is self-limiting - the pre-auth auto-releases after ~7 days and the `charge.expired` webhook updates the order. That seemed safer than resolving a missing flag here before, which could turn a captured charge into a full refund unexpectedly.

wjrosa: Makes sense, thanks!

tests/phpunit/class-wc-stripe-payment-gateway-test.php
wjrosa: Keeping it here makes sense. This PR is what introduces the charge fetch ahead of amount validation, so the guard belongs with it

wjrosa (approved):
Looks good to me and works as intended! Thanks for the improvement!

jaclync:
Updated changelog/readme to 11.0.0, and did another round of testing. Merging now.

PR #5729, opened 2026-07-22
Agentic Commerce: document the coupon limitation and reject discounted sessions explicitly

wjrosa:
Towards STRIPE-1257

The delegated (in-agent) checkout order path never applies WooCommerce coupons: line-item prices come from the synced feed and are computed by Stripe, and the order is created only after the shopper has paid inside the agent. (Feed-only / redirect mode is unaffected — redirected shoppers use the store's standard checkout, where coupons work as usual.) Two consequences follow: WooCommerce coupon usage limits are neither enforced nor consumed by agentic sales, and a Stripe-side discount on a session can never be represented on the order — today it fails `verify_order_total()` with an opaque "total mismatch" *after* the order was built and payment captured.

Whether promotions should be *supported* in agentic checkout (mapping Stripe discounts to WC coupons and enforcing usage limits at `finalize_checkout`) is a product decision this PR deliberately does not make. It implements the decision-neutral pieces:

- **Fail fast with a clear reason:** `validate_checkout_session()` now rejects a session whose `total_details.amount_discount` is non-zero with an explicit "discounts are not supported" error, before the order is built. This changes no outcomes — discounted sessions always failed — only where and how clearly they fail. New `get_amount_discount()` accessor on the session wrapper.
- **Document the limitation:** new "Coupons and discounts" section in `includes/agentic-commerce/README.md`, and copy in the Agentic Commerce settings panel telling merchants that WooCommerce coupons and their usage limits do not apply to purchases completed inside AI agents, while redirect-mode purchases keep normal coupon behavior. Both surfaces scope the limitation to delegated checkout.

If the eventual decision is to support promotions, the fail-fast check is the single seam to replace with discount mapping.

**BC impact:** no public signatures, hooks, or options change. The only behavior change is the earlier, clearer failure for discounted sessions (previously an unavoidable total-mismatch failure on the same sessions).

## Testing instructions

1. Load Stripe settings → Agentic commerce — the section description now states the coupon limitation.
2. Unit coverage: `npm run test:php -- --filter "WC_Stripe_Agentic_Commerce_Order_Mapper_Test|WC_Stripe_Agentic_Checkout_Session_Test"` — covers the discounted-session rejection and the new accessor; `npm run test:js -- client/settings/agentic-commerce` for the settings panel.

---

### Changelog entry

Changelog entries added to `changelog.txt` and `readme.txt` under 10.9.0.

<details>

#### Comment

</details>

**Post merge**

### DIFF
--- a/tests/phpunit/agentic-commerce/class-wc-stripe-agentic-checkout-session-test.php
+++ b/tests/phpunit/agentic-commerce/class-wc-stripe-agentic-checkout-session-test.php
@@ -56,15 +56,39 @@ class WC_Stripe_Agentic_Checkout_Session_Test extends WP_UnitTestCase {
 
 	/**
 	 * Test get_amount_discount reads total_details and defaults to 0 when absent.
+	 *
+	 * @dataProvider provide_amount_discount_cases
+	 *
+	 * @param object $raw      Raw session data.
+	 * @param int    $expected Expected discount amount.
 	 */
-	public function test_get_amount_discount() {
-		$with_discount = new WC_Stripe_Agentic_Checkout_Session(
-			(object) [ 'total_details' => (object) [ 'amount_discount' => 500 ] ]
-		);
-		$this->assertSame( 500, $with_discount->get_amount_discount() );
+	public function test_get_amount_discount( object $raw, int $expected ) {
+		$session = new WC_Stripe_Agentic_Checkout_Session( $raw );
+		$this->assertSame( $expected, $session->get_amount_discount() );
+	}
 
-		$without = new WC_Stripe_Agentic_Checkout_Session( (object) [] );
-		$this->assertSame( 0, $without->get_amount_discount() );
+	/**
+	 * @return array
+	 */
+	public function provide_amount_discount_cases(): array {
+		return [
+			'discount present'          => [
+				(object) [ 'total_details' => (object) [ 'amount_discount' => 500 ] ],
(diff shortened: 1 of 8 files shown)


tests/phpunit/agentic-commerce/class-wc-stripe-agentic-checkout-session-test.php
```diff
@@ -54,6 +54,19 @@ public function test_get_currency_lowercase() {
 		$this->assertSame( 'eur', $session->get_currency_lowercase() );
 	}
 
+	/**
+	 * Test get_amount_discount reads total_details and defaults to 0 when absent.
+	 */
+	public function test_get_amount_discount() {
```
diegocurbelo: [nit] We should use a @dataProvider like in `test_get_amount_total` and `test_get_customer_email` below

diegocurbelo (approved):
The code changes look good to me 👍🏼 

I left a suggestion, but is not a blocker to merge

tests/phpunit/agentic-commerce/class-wc-stripe-agentic-checkout-session-test.php
wjrosa: Done in 55a5044d4 — converted to a `@dataProvider` matching the siblings, with two extra cases (zero discount, `total_details` without the key).

PR #5739, opened 2026-07-23
Add e2e coverage for Adaptive Pricing currency conversion (simulated shopper location)

wjrosa:
Fixes STRIPE-1289

Follow-up to #5737.

#5737 could not cover the Adaptive Pricing happy path — an actual currency conversion — because CI shoppers browse from the store's own currency. This PR adds that coverage using Stripe's documented test hook for Adaptive Pricing: a `customer_email` of `test+location_XX@example.com` on the Checkout Session simulates a shopper browsing from country `XX` ([docs](https://docs.stripe.com/payments/currencies/localize-prices/adaptive-pricing?payment-ui=embedded-components#testing)).

- New e2e-only mu-plugin (`tests/e2e/env/mu-plugins/`) that injects the location email into guest session requests through the `wc_stripe_request_body` filter, gated by a `wc_stripe_e2e_location` cookie so only opted-in tests are affected. Copied into the container by `run-tests.sh` for the `adaptive-pricing` project.
- New test: a guest simulated as browsing from France sees EUR offered in the Adaptive Pricing currency selector element, completes the purchase, sees the currency-conversion notice on the order-received page, and the admin order edit page shows the “Paid by customer” row in EUR. The last two read presentment meta populated by a lazy checkout-session fetch at render time, so they don’t depend on the settlement webhook CI can’t receive.

No plugin runtime code changes; the injection lives entirely in the e2e environment.

## Testing instructions

1. `npm run test:e2e-setup`
2. `npm run test:e2e-adaptive-pricing`
3. All Adaptive Pricing tests pass, including `guest sees converted prices as a simulated French shopper`.

---

### Changelog entry

<details>

#### Comment

Follow-up to #5737 in the same unreleased version; its entries cover this work.

</details>

**Post merge**

### DIFF
--- a/client/blocks/checkout-sessions/checkout-form.js
+++ b/client/blocks/checkout-sessions/checkout-form.js
@@ -162,8 +162,8 @@ const CheckoutForm = ( {
 					} }
 				/>
 			) }
-			{ /* Wrapped so the selector is addressable from the DOM, matching the
-			     `wc-stripe-currency-selector` hook the classic checkout renders. */ }
+			{ /* Wrapped to give the selector a DOM hook for e2e tests. The class
+			     name mirrors the one classic renders; neither is styled. */ }
 			<div className="wc-stripe-currency-selector">
 				<CurrencySelectorElement />
 			</div>
--- a/client/blocks/checkout-sessions/checkout-form.js
+++ b/client/blocks/checkout-sessions/checkout-form.js
@@ -162,9 +162,9 @@ const CheckoutForm = ( {
 					} }
 				/>
 			) }
-			{ /* Wrapped to give the selector a DOM hook for e2e tests. The class
-			     name mirrors the one classic renders; neither is styled. */ }
-			<div className="wc-stripe-currency-selector">
+			{ /* Wrapped only to give e2e tests a DOM hook — classic exposes an
+			     equivalent element, but via a class rather than a test id. */ }
+			<div data-testid="wc-stripe-currency-selector">
 				<CurrencySelectorElement />
 			</div>
 			<PaymentElement
--- a/client/blocks/checkout-sessions/checkout-form.js
+++ b/client/blocks/checkout-sessions/checkout-form.js
@@ -162,7 +162,11 @@ const CheckoutForm = ( {
 					} }
 				/>
 			) }
-			<CurrencySelectorElement />
+			{ /* Wrapped so the selector is addressable from the DOM, matching the
+			     `wc-stripe-currency-selector` hook the classic checkout renders. */ }
+			<div className="wc-stripe-currency-selector">
+				<CurrencySelectorElement />
+			</div>
 			<PaymentElement
 				options={ elementOptions }
 				onChange={ onSelectedPaymentMethodChange }
--- a/tests/e2e/bin/run-tests.sh
+++ b/tests/e2e/bin/run-tests.sh
@@ -50,6 +50,7 @@ if [[ "adaptive-pricing" == "$project" ]]; then
 	# `set -e` would abort seeding on any re-run inside the TTL. Deleting a missing transient is a
 	# no-op, so this stays correct on a fresh site and refreshes the TTL on a repeat run.
 	cli wp transient delete wcstripe_webhook_status_test
+	cli wp transient delete wcstripe_webhook_status_test
 	cli wp transient set wcstripe_webhook_status_test enabled 7200
 	# Cookie-gated mu-plugin that simulates the shopper's country for
 	# conversion tests (Stripe's "+location_XX" customer_email test hook).
--- a/tests/e2e/bin/run-tests.sh
+++ b/tests/e2e/bin/run-tests.sh
@@ -39,4 +39,16 @@ fi
 TEST_ENV="$TEST_ENV DOCKER=true E2E_ROOT=${E2E_ROOT} BASE_URL='http://localhost:8088'"
 TEST_ENV="$TEST_ENV ADMIN_USER='admin' ADMIN_PASSWORD='admin'"
 
+# Adaptive Pricing is only offered while the plugin believes webhooks work and
+# the payment method configurations API is on. The docker site cannot receive
+# real webhooks (no tunnel), so seed exactly what the availability checks read:
+# webhook data plus the cached status transient — which is_webhook_enabled()
+# trusts without calling Stripe — and the PMC flag.
+if [[ "adaptive-pricing" == "$project" ]]; then
+	echo "Seeding Adaptive Pricing prerequisites"
+	cli wp option patch insert woocommerce_stripe_settings pmc_enabled 'yes'
+	cli wp option patch insert woocommerce_stripe_settings test_webhook_data --format=json '{"id":"we_e2e_placeholder","secret":"whsec_e2e_placeholder"}'
+	cli wp transient set wcstripe_webhook_status_test enabled 7200
+fi
+
 cross-env $TEST_ENV playwright test --config=tests/e2e/config/playwright.config.js $TEST_ARGS ${project:+--project=$project}
--- a/tests/e2e/bin/run-tests.sh
+++ b/tests/e2e/bin/run-tests.sh
@@ -39,11 +39,9 @@ fi
 TEST_ENV="$TEST_ENV DOCKER=true E2E_ROOT=${E2E_ROOT} BASE_URL='http://localhost:8088'"
 TEST_ENV="$TEST_ENV ADMIN_USER='admin' ADMIN_PASSWORD='admin'"
 
-# Adaptive Pricing is only offered while the plugin believes webhooks work and
-# the payment method configurations API is on. The docker site cannot receive
-# real webhooks (no tunnel), so seed exactly what the availability checks read:
-# webhook data plus the cached status transient — which is_webhook_enabled()
-# trusts without calling Stripe — and the PMC flag.
+# The docker site can't receive real webhooks (no tunnel), so seed what the
+# Adaptive Pricing availability checks read: webhook data, the cached status
+# transient is_webhook_enabled() trusts without calling Stripe, and PMC.
 if [[ "adaptive-pricing" == "$project" ]]; then
 	echo "Seeding Adaptive Pricing prerequisites"
 	cli wp option patch insert woocommerce_stripe_settings pmc_enabled 'yes'
--- a/tests/e2e/bin/run-tests.sh
+++ b/tests/e2e/bin/run-tests.sh
@@ -50,7 +50,6 @@ if [[ "adaptive-pricing" == "$project" ]]; then
 	# `set -e` would abort seeding on any re-run inside the TTL. Deleting a missing transient is a
 	# no-op, so this stays correct on a fresh site and refreshes the TTL on a repeat run.
 	cli wp transient delete wcstripe_webhook_status_test
-	cli wp transient delete wcstripe_webhook_status_test
 	cli wp transient set wcstripe_webhook_status_test enabled 7200
 	# Cookie-gated mu-plugin that simulates the shopper's country for
 	# conversion tests (Stripe's "+location_XX" customer_email test hook).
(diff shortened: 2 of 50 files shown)


tests/e2e/env/mu-plugins/wc-stripe-e2e-location-simulation.php
```diff
@@ -0,0 +1,31 @@
+<?php
+/**
+ * Plugin Name: WC Stripe E2E Location Simulation
+ * Description: Lets e2e tests simulate the shopper's country for Adaptive Pricing using Stripe's documented "+location_XX" customer_email test hook.
```
jaclync, line 4: non-blocking nit: is it worth linking the [Stripe documentation](https://docs.stripe.com/payments/currencies/localize-prices/adaptive-pricing?payment-ui=embedded-components#testing) here?

tests/e2e/tests/checkout/adaptive-pricing.spec.js
```diff
@@ -94,6 +95,66 @@ test.describe( 'Adaptive Pricing checkout', () => {
...
+			// HPOS redirects this legacy edit URL to the new order screen.
+			await adminPage.goto(
+				`/wp-admin/post.php?post=${ orderId }&action=edit`
+			);
+			await expect(
+				adminPage
+					.locator( '.wc-order-totals tr' )
+					.filter( { hasText: 'Paid by customer' } )
+					.locator( '.total' )
+			).toContainText( '€' );
```
jaclync: non-blocking nit: do you think the same code as in [`verifyOrderChargedAmount`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/933963bc39ed95381aeec973a6c4315cacd7fea8/tests/e2e/utils/admin.js#L214-L229) could be shared? Like the navigation to order page `await adminPage.goto(order screen)` (the HPOS note can go to the shared helper) and the order total element.

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/933963bc39ed95381aeec973a6c4315cacd7fea8/tests/e2e/utils/admin.js#L214-L229

tests/e2e/bin/run-tests.sh
```diff
@@ -47,6 +47,9 @@ if [[ "adaptive-pricing" == "$project" ]]; then
 	cli wp option patch insert woocommerce_stripe_settings pmc_enabled 'yes'
 	cli wp option patch insert woocommerce_stripe_settings test_webhook_data --format=json '{"id":"we_e2e_placeholder","secret":"whsec_e2e_placeholder"}'
 	cli wp transient set wcstripe_webhook_status_test enabled 7200
+	# Cookie-gated mu-plugin that simulates the shopper's country for
+	# conversion tests (Stripe's "+location_XX" customer_email test hook).
+	cli sh -c "mkdir -p /var/www/html/wp-content/mu-plugins && cp /var/www/html/wp-content/plugins/woocommerce-gateway-stripe/tests/e2e/env/mu-plugins/wc-stripe-e2e-location-simulation.php /var/www/html/wp-content/mu-plugins/"
```
jaclync, line 56: non-blocking: how about adding a mount like `./mu-plugins:/var/www/html/wp-content/mu-plugins` to the e2e docker-compose, so that it has the same life cycle as the container?

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/63c14228f264d9f723cb5a7e84ebfc799550be11/tests/e2e/env/docker-compose.yml#L27-L30

The caveat is to require a container creation in the testing steps.

jaclync (commented):
Nice focused changes, the code looks good with some minor non-blocking comments. I ran the e2e tests locally on this branch, but the new test failed intermittently (passed on another run). The currency conversion worked (both currencies shown), but the step to fill in card number timed out waiting for the card number field:

```
TimeoutError: locator.fill: Timeout 15000ms exceeded.
waiting for locator('#wc-stripe-upe-form .StripeElement iframe[name^="__privateStripeFrame"]').filter({ visible: true }).first().contentFrame().locator('[name="number"]')
```

Image

Have you seen this before?

---

Separate observation when running e2e tests: when I tried running the AP e2e tests multiple times, it failed at setting the transient:

```
> woocommerce-gateway-stripe@10.8.4 test:e2e-adaptive-pricing
> ./tests/e2e/bin/run-tests.sh --project=adaptive-pricing

Seeding Adaptive Pricing prerequisites
Success: Value passed for 'woocommerce_stripe_settings' option is unchanged.
Success: Value passed for 'woocommerce_stripe_settings' option is unchanged.
Error: Transient could not be set.
```

Deleting the transient fixed it, it's likely from setting the transient while it was set before within the TTL.

tests/e2e/env/mu-plugins/wc-stripe-e2e-location-simulation.php
```diff
@@ -0,0 +1,31 @@
...
+add_filter(
+	'wc_stripe_request_body',
+	function ( $request, $api ) {
+		// Stripe rejects customer_email alongside customer, so only guest
+		// sessions (no attached Stripe customer) can carry the simulation.
+		if ( 'checkout/sessions' !== $api || ! empty( $request['customer'] ) ) {
+			return $request;
+		}
+
+		$location = isset( $_COOKIE['wc_stripe_e2e_location'] ) ? strtoupper( sanitize_key( wp_unslash( $_COOKIE['wc_stripe_e2e_location'] ) ) ) : '';
```
daledupreez: Nit: Why not bail early if the cookie is not set?
```suggestion
		if ( ! isset( $_COOKIE['wc_stripe_e2e_location'] ) ) {
			return $request;
		}
		$location = strtoupper( sanitize_key( wp_unslash( $_COOKIE['wc_stripe_e2e_location'] ) ) );
```

tests/e2e/tests/checkout/adaptive-pricing.spec.js
```diff
@@ -94,6 +95,66 @@ test.describe( 'Adaptive Pricing checkout', () => {
 		await payWithAdaptivePricing( page, 'shortcode' );
 	} );
 
+	test( 'guest sees converted prices as a simulated French shopper', async ( {
```
daledupreez: Can we add an equivalent test for block checkout?

tests/e2e/bin/run-tests.sh
```diff
@@ -47,6 +47,9 @@ if [[ "adaptive-pricing" == "$project" ]]; then
 	cli wp option patch insert woocommerce_stripe_settings pmc_enabled 'yes'
 	cli wp option patch insert woocommerce_stripe_settings test_webhook_data --format=json '{"id":"we_e2e_placeholder","secret":"whsec_e2e_placeholder"}'
 	cli wp transient set wcstripe_webhook_status_test enabled 7200
```
daledupreez, line 53: As noted in @jaclync's [comment](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5739#pullrequestreview-4804448207), we may want to delete the transient and then set it.
```suggestion
	cli wp transient delete wcstripe_webhook_status_test
	cli wp transient set wcstripe_webhook_status_test enabled 7200
```

daledupreez (commented):
This is looking good overall, but the tests only cover classic checkout without stating why we're not covering block checkout as well.

tests/e2e/env/mu-plugins/wc-stripe-e2e-location-simulation.php
wjrosa: Added in `b32e7a9b7`. 👍

tests/e2e/tests/checkout/adaptive-pricing.spec.js
wjrosa: Good call — extracted `gotoOrderEditPage()` and `getOrderTotalForLabel()` in `utils/admin.js`, now used by both `verifyOrderChargedAmount()` and this spec. `b32e7a9b7`

tests/e2e/bin/run-tests.sh
wjrosa: Tempted, but I've left the copy for now. The mount would apply the plugin to every e2e project rather than just `adaptive-pricing`, and — as you note — force everyone to recreate their container. It's cookie-gated so it'd be inert elsewhere, but that's a fair bit of blast radius for hygiene.

Happy to revisit if we add more mu-plugins; at that point the mount clearly wins.

wjrosa:
Thanks @jaclync — both items:

**Transient error:** reproduced and fixed in `b32e7a9b7`. `wp transient set` returns exit 1 ("Transient could not be set") when the value is unchanged, because `set_transient()` → `update_option()` returns false. Under `set -e` that aborted seeding on any re-run inside the 2h TTL. It now deletes first — a no-op on a fresh site, and it refreshes the TTL on repeats.

**Card field flake:** yes, but not from this PR. I went through the local Allure results and that Stripe-iframe timeout shows up in `optimized-checkout` (8), `default` (4) and `acss` (2) — and zero times in `adaptive-pricing`. Same locator, same 15s timeout, on runs predating this branch. So it looks like a pre-existing race in the shared OC helper where the Payment Element iframe isn't interactive yet, rather than anything conversion-specific — which matches your run showing both currencies rendering fine before it failed.

I'd rather not paper over it here with a retry; worth its own issue against the shared helper so the fix covers every project. Happy to open one.

tests/e2e/tests/checkout/adaptive-pricing.spec.js
wjrosa: Added in `e8b5e20ca` — the conversion assertions are now a shared helper taking the checkout type, with a blocks variant alongside the shortcode one.

Blocks rendered `<CurrencySelectorElement />` bare, so there was no DOM hook to assert against. I wrapped it in `wc-stripe-currency-selector` to match what classic already renders, and both tests target that class.

Caveat: I can't verify this locally — a real conversion needs an onboarded Stripe account, and our dev one has its capabilities inactive. Worth watching its first CI run.

client/blocks/checkout-sessions/checkout-form.js
```diff
@@ -162,7 +162,11 @@ const CheckoutForm = ( {
 					} }
 				/>
 			) }
-			<CurrencySelectorElement />
+			{ /* Wrapped so the selector is addressable from the DOM, matching the
+			     `wc-stripe-currency-selector` hook the classic checkout renders. */ }
+			<div className="wc-stripe-currency-selector">
```
daledupreez: I just checked, and we don't have any CSS rules that target the `wc-stripe-currency-selector` class.

tests/e2e/tests/checkout/adaptive-pricing.spec.js
```diff
@@ -94,6 +161,36 @@ test.describe( 'Adaptive Pricing checkout', () => {
...
```
daledupreez: Nit: We should specify classic/shortcode checkout in the description.
```suggestion
	test( 'guest sees converted prices as a simulated French shopper in classic checkout', async ( {
```

daledupreez (approved):
The changes are LTGM, and the e2e tests are passing. I think we should fix the one test description I flagged before merging, but that's a minor tweak.

client/blocks/checkout-sessions/checkout-form.js
wjrosa: Confirmed — the class is inert in both checkouts. Classic's spacing comes from the inline `style="margin-top: 12px;"` on its div, not the class, so my "matching classic" comment implied a parity that doesn't exist. Reworded in `9a39a4db9` to say it's a DOM hook for e2e only.

Happy to switch it to a `data-testid` if you'd rather that intent were explicit. Separately and pre-existing: blocks never gets classic's 12px top margin, since it lives in that inline style.

wjrosa: Switched in `cc12fa3c7` — blocks now uses `data-testid="wc-stripe-currency-selector"`. Left classic's class alone, so the spec matches either:

```js
page.locator( '.wc-stripe-currency-selector, [data-testid="wc-stripe-currency-selector"]' )
```

PR #5749, opened 2026-07-26
Center card field content on the Blocks checkout

kamlesh08:
Fixes STRIPE-1276

On the Blocks checkout, the content inside the Stripe card fields sat about 3px above center — easiest to see on the card brand icons. We build Stripe's styling from the theme's own input, whose padding is top-heavy for the floating label, and we subtracted the label's room from the top only. Now the same total comes off both sides: content is centered, field height unchanged.

Review found a second problem in the same function, fixed here too. When a sampled value doesn't parse (a selector that matches nothing, a `line-height` keyword), the fallback emitted invalid CSS like `calc(24px - undefined - 4px - 1px)` and still adjusted the other side. It now removes both padding overrides together and lets Stripe's defaults apply. Checked against a live Payment Element: Stripe drops only the single invalid declaration, so this renders the same as the old invalid output did.

Applies to every Payment Element field on the Blocks checkout, OCS on or off. Classic checkout is untouched — the code only runs for Blocks. No hooks, options, or PHP changes.

### Screenshots

Same site, theme, and Stripe account; only the plugin build differs (release 10.9.x vs this PR).
All screenshots are added in the comment.

## Testing instructions

1. Connect Stripe in test mode and use the Checkout block for the checkout page (not the shortcode).
2. Add a product, go to checkout, and leave the card fields empty.
3. Check the Card number field: the brand icons and label text should sit on the field's center line, and the card fields should match the height of the address inputs above.
4. Type `4242 4242 4242 4242` with any expiry and CVC. The label floats up, the value stays centered. Place the order.
5. Go to WooCommerce → Settings → Payments → Stripe, enable **Optimized Checkout Suite**, reload the checkout, and repeat steps 3–4 on the "Payment methods" element.
6. Optional: run `npm run test:js -- client/styles/upe` — 21 tests, red on `develop`, green here.

---

-   [x] Covered with tests (10 new unit tests: balanced padding and the non-numeric fallback)

### Changelog entry

Added to `changelog.txt` and `readme.txt`.

**Post merge**

---

### DIFF
--- a/client/styles/upe/utils.js
+++ b/client/styles/upe/utils.js
@@ -186,26 +186,66 @@ export const handleAppearanceForFloatingLabel = (
 		delete appearance.rules[ '.Label--floating' ].transform;
 	}
 
-	// Subtract the label's lineHeight from padding-top to account for floating label height.
-	// Minus STRIPE_PADDING_TOP which is a constant value added by Stripe to the padding-top.
-	// Minus STRIPE_PADDING_OFFSET for each vertical padding to account for unpredictable input height.
-	if ( appearance.rules[ '.Input' ].paddingTop ) {
-		appearance.rules[
-			'.Input'
-		].paddingTop = `calc(${ appearance.rules[ '.Input' ].paddingTop } - ${ appearance.rules[ '.Label--floating' ].lineHeight } - ${ STRIPE_PADDING_TOP } - ${ STRIPE_PADDING_OFFSET })`;
-	}
-	if ( appearance.rules[ '.Input' ].paddingBottom ) {
-		const paddingOffset = parseFloat( STRIPE_PADDING_OFFSET );
-		const originalPaddingBottom = parseFloat(
-			appearance.rules[ '.Input' ].paddingBottom
-		);
-		appearance.rules[ '.Input' ].paddingBottom = `${
-			originalPaddingBottom - paddingOffset
-		}px`;
+	// Reserve room for the floating label without letting the field grow taller
+	// than the theme's own inputs. The space the label needs is its lineHeight,
+	// plus STRIPE_PADDING_TOP which Stripe adds to padding-top on its own, plus
+	// STRIPE_PADDING_OFFSET per side to absorb unpredictable input height.
+	const paddingTop = parseFloat( appearance.rules[ '.Input' ].paddingTop );
+	const paddingBottom = parseFloat(
+		appearance.rules[ '.Input' ].paddingBottom
+	);
+	const floatingLabelLineHeight = parseFloat(
+		appearance.rules[ '.Label--floating' ].lineHeight
+	);
 
-		appearance.rules[ '.Label' ].marginTop = `${ Math.floor(
-			( originalPaddingBottom - paddingOffset ) / 3
-		) }px`;
+	if (
+		Number.isFinite( paddingTop ) &&
+		Number.isFinite( paddingBottom ) &&
+		Number.isFinite( floatingLabelLineHeight )
+	) {
+		// Themes that float their labels park the extra vertical space at the
+		// top of the input, where the label sits. Stripe centers what it draws
+		// inside .Input — the value, and the card brand icons on the card
+		// number field — against the padding box, so carrying that top-heavy
+		// split across leaves them visibly high. Take the same total out of
+		// both sides instead: the field keeps its height and the contents stay
+		// centered. Clamped at zero because a negative padding is dropped, and
+		// the field would then outgrow the theme's inputs.
+		const reservedForLabel =
+			floatingLabelLineHeight +
+			parseFloat( STRIPE_PADDING_TOP ) +
+			parseFloat( STRIPE_PADDING_OFFSET ) * 2;
+		const balancedPadding = Math.max(
+			0,
+			( paddingTop + paddingBottom - reservedForLabel ) / 2
+		);
+
+		appearance.rules[ '.Input' ].paddingTop = `${ balancedPadding }px`;
+		appearance.rules[ '.Input' ].paddingBottom = `${ balancedPadding }px`;
+	} else {
+		// Fall back to adjusting each side on its own when a padding or the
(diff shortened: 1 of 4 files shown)


client/styles/upe/utils.js
```diff
@@ -186,26 +186,66 @@ export const handleAppearanceForFloatingLabel = (
...
+			);
+			appearance.rules[ '.Input' ].paddingBottom = `${
+				originalPaddingBottom - paddingOffset
+			}px`;
+
+			// Pulls the resting label back toward center against the top-heavy
+			// input this branch leaves behind.
+			appearance.rules[ '.Label' ].marginTop = `${ Math.floor(
+				( originalPaddingBottom - paddingOffset ) / 3
+			) }px`;
```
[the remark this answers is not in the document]
wjrosa: We should address this one

client/styles/upe/utils.js
```diff
@@ -186,26 +186,66 @@ export const handleAppearanceForFloatingLabel = (
...
-	if ( appearance.rules[ '.Input' ].paddingTop ) {
-		appearance.rules[
-			'.Input'
-		].paddingTop = `calc(${ appearance.rules[ '.Input' ].paddingTop } - ${ appearance.rules[ '.Label--floating' ].lineHeight } - ${ STRIPE_PADDING_TOP } - ${ STRIPE_PADDING_OFFSET })`;
-	}
-	if ( appearance.rules[ '.Input' ].paddingBottom ) {
-		const paddingOffset = parseFloat( STRIPE_PADDING_OFFSET );
-		const originalPaddingBottom = parseFloat(
-			appearance.rules[ '.Input' ].paddingBottom
+	// Reserve room for the floating label without letting the field grow taller
```
wjrosa: Can you shorten all comments in this file? AI (if you used it) tends to be too verbose

client/styles/upe/utils.js
```diff
@@ -186,26 +186,66 @@ export const handleAppearanceForFloatingLabel = (
...
+			( paddingTop + paddingBottom - reservedForLabel ) / 2
 		);
-		appearance.rules[ '.Input' ].paddingBottom = `${
-			originalPaddingBottom - paddingOffset
-		}px`;
 
-		appearance.rules[ '.Label' ].marginTop = `${ Math.floor(
-			( originalPaddingBottom - paddingOffset ) / 3
-		) }px`;
+		appearance.rules[ '.Input' ].paddingTop = `${ balancedPadding }px`;
```
wjrosa, line 225: This new padding logic also applies to Optimized Checkout: `handleAppearanceForFloatingLabel()` runs for every Blocks checkout, whether OCS is enabled or not, and nothing in the OCS-specific styling overrides the `.Input` padding set here. So OCS fields will look different after this change too. 

Could you add an OCS step (and screenshot) to the testing instructions? 🙏

client/styles/upe/utils.js
kamlesh08: Done.

client/styles/upe/utils.js
kamlesh08: Added an OCS step to the testing instructions and before/after screenshots for both OCS modes. On the per-method card element, the icons measure ~3px above center on the release build and are centered with this PR, so the fix shows up there too.

client/styles/upe/utils.js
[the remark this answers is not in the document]
kamlesh08: Done.

kamlesh08:
### OCS Disabled - Before
ocs-disabled-before

<hr>

### OCS Disabled - After
ocs-disabled-after

<hr>

### OCS Enabled - Before
ocs-enabled-before

<hr>

### OCS Enabled - After
ocs-enabled-after

@wjrosa

client/styles/upe/utils.js
```diff
@@ -186,26 +186,38 @@ export const handleAppearanceForFloatingLabel = (
...
+		appearance.rules[ '.Input' ].paddingBottom = `${ balancedPadding }px`;
+	} else {
```
wjrosa, line 227: The two early `return appearance` exits in the transform block above (non-finite scale, keyword `lineHeight` with a transform) still bypass this fallback — the raw top-heavy padding stays while the label floats, which is the exact case this branch now cleans up. Worth routing them through the same cleanup?

client/styles/upe/utils.js
```diff
@@ -186,26 +186,38 @@ export const handleAppearanceForFloatingLabel = (
...
-		appearance.rules[
-			'.Input'
-		].paddingTop = `calc(${ appearance.rules[ '.Input' ].paddingTop } - ${ appearance.rules[ '.Label--floating' ].lineHeight } - ${ STRIPE_PADDING_TOP } - ${ STRIPE_PADDING_OFFSET })`;
-	}
-	if ( appearance.rules[ '.Input' ].paddingBottom ) {
-		const paddingOffset = parseFloat( STRIPE_PADDING_OFFSET );
-		const originalPaddingBottom = parseFloat(
-			appearance.rules[ '.Input' ].paddingBottom
+	// Reserve room for the floating label without letting the field grow taller.
+	const paddingTop = parseFloat( appearance.rules[ '.Input' ].paddingTop );
```
wjrosa: nit: `parseFloat( '1.5rem' )` passes the finite guard below and gets treated as px. Computed styles should always resolve to px, but since the tests already defend equally-impossible `calc()`/`var()` inputs, a cheap `endsWith( 'px' )` check would close this gap too.

client/styles/upe/utils.js
kamlesh08: Done.

client/styles/upe/utils.js
kamlesh08: Done.

kamlesh08:
@wjrosa all done. 
Please do let me know if any other changes are needed here.
Thanks a lot for reviewing.

wjrosa (approved):

PR #5752, opened 2026-07-27
Scope Agentic Commerce catalog sync state to the active Stripe mode

wjrosa:
Fixes STRIPE-1306

Mode *identification* in the Agentic Commerce catalog sync is correct — every entry point resolves the secret key from `testmode` at call time — but the persisted state around the sync was mode-blind, so switching test↔live silently misbehaved:

- The unchanged-feed dedup record is a single global option, so after a mode switch the first sync to the new mode's environment was skipped (same content hash) and the newly active catalog stayed empty until products changed or the cache TTL expired.
- Sync history and the settings status card reported the other mode's sync as current.
- The pending-status poll called `get_import_set()` with the current mode's key for ImportSets created under the other mode, which 404s — leaving a pending badge that could never resolve.

Fix:

- The dedup record and each sync-history entry now record their `mode` (`test`/`live`). A mode mismatch (or a legacy record with no mode) invalidates dedup, and the REST status endpoint filters history/last-sync/polling to the active mode (legacy entries stay visible).
- Toggling `testmode` in the Stripe settings clears the dedup record and schedules an immediate resync (only when the merchant has the integration enabled), so the new mode's catalog populates without waiting for the next content change.

**BC impact:** additive — new keys in existing option records, one new public static helper (`get_current_mode()`), one new hook registration on the existing `update_option_woocommerce_stripe_settings` action. Legacy records without a mode trigger one extra (idempotent) upload and remain visible in history.

## Testing instructions

1. Enable Agentic Commerce in test mode and let a sync complete; note the status in the settings.
2. Switch the plugin to live mode (with live keys configured): an immediate resync is scheduled (visible in Action Scheduler), and the sync status/history no longer shows the test-mode entries.
3. Switch back to test mode: the test-mode history reappears.

---

### Changelog entry

Changelog entries added to `changelog.txt` and `readme.txt` under 10.9.0 (Fix).

<details>

#### Comment

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
+++ b/includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
@@ -407,7 +407,13 @@ class WC_REST_Stripe_Agentic_Commerce_Controller extends WC_Stripe_REST_Base_Con
 		$status_updates = [];
 		$mode_updates   = [];
 		$delivery       = null;
-		$current_mode   = WC_Stripe_Agentic_Commerce_Integration::get_current_mode();
+
+		// One settings snapshot supplies both the mode used to filter entries
+		// and the key used to poll them: separate reads could straddle a
+		// concurrent settings save and poll this mode's entries with the other
+		// mode's key, misreading the resulting 404s as cross-mode entries.
+		$context      = WC_Stripe_Agentic_Commerce_Integration::get_delivery_context();
+		$current_mode = $context['mode'];
 
 		foreach ( $history as $entry ) {
 			$current_status = $entry['status'] ?? '';
@@ -427,7 +433,7 @@ class WC_REST_Stripe_Agentic_Commerce_Controller extends WC_Stripe_REST_Base_Con
 			}
 
 			try {
-				$delivery   = $delivery ?? $this->create_delivery();
+				$delivery   = $delivery ?? new WC_Stripe_Agentic_Commerce_Files_Api_Delivery( $context['secret_key'] );
 				$import_set = $delivery->get_import_set( $import_set_id );
 				$new_status = $import_set['status'] ?? $current_status;
 
@@ -461,22 +467,6 @@ class WC_REST_Stripe_Agentic_Commerce_Controller extends WC_Stripe_REST_Base_Con
 		}
 
 		WC_Stripe_Agentic_Commerce_Integration::update_pending_statuses( $status_updates, $mode_updates );
-	}
-
-	/**
-	 * Create a Files API delivery instance using the current Stripe settings.
-	 *
-	 * @since 10.7.0
-	 * @return WC_Stripe_Agentic_Commerce_Files_Api_Delivery
-	 */
-	private function create_delivery(): WC_Stripe_Agentic_Commerce_Files_Api_Delivery {
-		$settings  = WC_Stripe_Helper::get_stripe_settings();
-		$test_mode = isset( $settings['testmode'] ) && 'yes' === $settings['testmode'];
-		$secret    = $test_mode
-			? ( $settings['test_secret_key'] ?? '' )
-			: ( $settings['secret_key'] ?? '' );
-
-		return new WC_Stripe_Agentic_Commerce_Files_Api_Delivery( $secret );
 	}
 
 	/**
--- a/includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
+++ b/includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
@@ -177,6 +177,13 @@ class WC_REST_Stripe_Agentic_Commerce_Controller extends WC_Stripe_REST_Base_Con
 		$last_sync   = WC_Stripe_Agentic_Commerce_Integration::get_last_sync();
 		$history_raw = WC_Stripe_Agentic_Commerce_Integration::get_sync_history();
 
+		// Syncs from the other mode describe a different Stripe environment's
+		// catalog; showing them here would report the wrong catalog as synced.
+		if ( ! $this->is_entry_for_current_mode( $last_sync ) ) {
+			$last_sync = [];
+		}
+		$history_raw = array_values( array_filter( $history_raw, [ $this, 'is_entry_for_current_mode' ] ) );
+
 		// Return the most recent history entries, newest first.
 		$history = array_map(
 			[ $this, 'format_entry' ],
@@ -393,6 +400,12 @@ class WC_REST_Stripe_Agentic_Commerce_Controller extends WC_Stripe_REST_Base_Con
 				continue;
 			}
 
+			// An ImportSet from the other mode can't be retrieved with the
+			// current mode's key (Stripe returns a 404).
+			if ( ! $this->is_entry_for_current_mode( $entry ) ) {
+				continue;
+			}
+
 			$import_set_id = $entry['import_set_id'] ?? '';
 			if ( '' === $import_set_id ) {
 				continue;
@@ -454,6 +467,22 @@ class WC_REST_Stripe_Agentic_Commerce_Controller extends WC_Stripe_REST_Base_Con
 			'file_id'       => $entry['file_id'] ?? null,
 			'error'         => $entry['error'] ?? null,
 		];
+	}
+
+	/**
+	 * Whether a persisted sync entry belongs to the currently active mode.
+	 *
+	 * Entries with no recorded mode predate mode tracking and are kept, since
+	 * their mode can no longer be determined.
+	 *
+	 * @since 10.9.0
+	 * @param array $entry Raw entry from the options table.
+	 * @return bool
+	 */
+	private function is_entry_for_current_mode( array $entry ): bool {
+		$mode = $entry['mode'] ?? '';
+
+		return '' === $mode || WC_Stripe_Agentic_Commerce_Integration::get_current_mode() === $mode;
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
```diff
@@ -393,6 +400,12 @@ private function refresh_pending_sync_statuses(): void {
 				continue;
 			}
 
+			// An ImportSet from the other mode can't be retrieved with the
+			// current mode's key (Stripe returns a 404).
+			if ( ! $this->is_entry_for_current_mode( $entry ) ) {
```
jaclync: ❓ Legacy history entries without a mode pass this guard, and could continue with the key in the opposite mode. How about keeping polling mode-less entries, but making a 404 error stop retrying indefinitely in the same opposite mode?

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
```diff
@@ -177,6 +177,13 @@ public function get_status() {
 		$last_sync   = WC_Stripe_Agentic_Commerce_Integration::get_last_sync();
 		$history_raw = WC_Stripe_Agentic_Commerce_Integration::get_sync_history();
 
+		// Syncs from the other mode describe a different Stripe environment's
+		// catalog; showing them here would report the wrong catalog as synced.
+		if ( ! $this->is_entry_for_current_mode( $last_sync ) ) {
+			$last_sync = [];
+		}
+		$history_raw = array_values( array_filter( $history_raw, [ $this, 'is_entry_for_current_mode' ] ) );
```
jaclync: ❓ Dropping the other-mode snapshot in L182-L184 leaves a gap after switching back: since `LAST_SYNC_OPTION` is only rewritten when a sync actually completes, in a case like test -> live -> test, it still holds the live entry until the queued resync runs. The sync cron job can take minutes or never (test keys missing, integration disabled at switch time). In that window, this returns `last_sync: null`, and the card shows "No syncs yet" while an entry exists in history. This could probably be verified in a test case.

Since `store_sync_result()` already defines the snapshot as `end( $history )`, how about deriving it similarly from the mode-filtered history:

```diff
-		$last_sync   = WC_Stripe_Agentic_Commerce_Integration::get_last_sync();
 		$history_raw = WC_Stripe_Agentic_Commerce_Integration::get_sync_history();

-		if ( ! $this->is_entry_for_current_mode( $last_sync ) ) {
-			$last_sync = [];
-		}
 		$history_raw = array_values( array_filter( $history_raw, [ $this, 'is_entry_for_current_mode' ] ) );
+		$last_sync   = end( $history_raw ) ?: [];
```

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
```diff
@@ -456,6 +469,22 @@ private function format_entry( array $entry ): array {
...
+	 * Whether a persisted sync entry belongs to the currently active mode.
+	 *
+	 * Entries with no recorded mode predate mode tracking and are kept, since
+	 * their mode can no longer be determined.
+	 *
+	 * @since 10.9.0
+	 * @param array $entry Raw entry from the options table.
+	 * @return bool
+	 */
+	private function is_entry_for_current_mode( array $entry ): bool {
```
jaclync: non-blocking suggestion: how about adding the mode as a parameter instead of calling `get_current_m ode()` per entry (it can be the default value of the parameter)? Especially when it's used to filter an array, a fixed mode value to compare to feels safer.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -1032,11 +1066,25 @@ protected function remember_feed_upload( string $hash, array $result ): void {
 				'uploaded_at'   => time(),
 				'file_id'       => is_string( $result['file_id'] ?? null ) ? $result['file_id'] : '',
 				'import_set_id' => is_string( $result['import_set_id'] ?? null ) ? $result['import_set_id'] : '',
+				'mode'          => self::get_current_mode(),
```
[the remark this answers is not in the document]
jaclync: +1, getting the mode at the beginning of `sync_feed` that generates the feed feels safer in case the mode changes.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -994,6 +1021,13 @@ protected function is_feed_unchanged( string $current_hash ): bool {
 			return false;
 		}
 
+		// A record from the other mode (or a legacy record with no mode) says
+		// nothing about the current mode's environment — its catalog may have
+		// never received this feed, so an identical hash must still upload.
+		if ( ( $last['mode'] ?? '' ) !== self::get_current_mode() ) {
```
jaclync, line 1056: non-blocking question: what about the account check in addition to the test/live mode? If the merchant happens to switch their Stripe account while being in the same mode, would the upload be skipped here?

jaclync (commented):
Some questions about corner cases, I will get to testing next Monday!

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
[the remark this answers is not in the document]
wjrosa: Done in a2eec2a2f — the mode is captured once at the start of `sync_feed()` (next to the delivery client) and threaded through `remember_feed_upload()` and both `store_sync_result()` call sites as an appended optional argument, so a mid-sync testmode flip can't mislabel the upload record or history entries. Regression tests cover the explicit-mode persistence.

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
wjrosa: Implemented in a2eec2a2f, slightly extending your idea: mode-less entries keep getting polled, and on a 404 the entry is stamped with the *other* mode — that both stops the indefinite retry under the wrong key and makes the entry visible/pollable again if the merchant activates that mode, without faking a terminal status. The delivery now carries the HTTP status as the exception code so the 404 check doesn't parse messages. A recorded mode is never overwritten by this reclassification.

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
wjrosa: Applied your diff in a2eec2a2f — `last_sync` is now derived from the mode-filtered history, so the stale other-mode snapshot can never blank the card during the test → live → test window. `test_get_status_filters_out_other_mode_entries` now asserts the current-mode entry surfaces as `last_sync` while the snapshot option still holds the other mode's entry, per your suggestion to verify it in a test.

includes/admin/class-wc-rest-stripe-agentic-commerce-controller.php
wjrosa: Done in a2eec2a2f — now `is_entry_for_mode( $entry, $mode )`, with the mode resolved once per filtering pass in `get_status()` and the refresh loop.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Good question — yes: a same-mode account switch keeps `mode` matching, so an unchanged catalog would skip upload for up to `FEED_CACHE_TTL` (a week), and the new account's catalog would sit empty. The `maybe_resync_after_mode_switch` hook only watches `testmode`, so it doesn't catch this either. I'd rather handle it as a small follow-up than widen this PR: extend that hook to also clear the dedup record and queue a resync when the secret keys change (which covers account switches and key rotation in one move). Happy to open it after this lands.

jaclync (approved):
Thanks for the updates, LGTM! :shipit: Saw the `wc_stripe_agentic_commerce_sync_feed_now` action scheduled after connecting to an account in live mode from test mode, and the feed was synced successfully. Didn't test the live mode agentic purchase flow, as the live mode account isn't set up with live payments.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -546,6 +572,12 @@ public function sync_feed( bool $force_upload = false ): bool {
...
 		}
 
+		// Capture the mode once, alongside the delivery client, and hold it for
+		// the whole run: a settings save can flip testmode while the sync is
+		// generating the feed, and every record this run persists must describe
+		// the environment it actually delivered to, not the mode at write time.
+		$mode = self::get_current_mode();
+
 		// Check delivery setup before generating the feed.
 		$delivery = $this->get_push_delivery_method();
```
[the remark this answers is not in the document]
wjrosa: Addressed. Both paths now capture one settings snapshot via a new `WC_Stripe_Agentic_Commerce_Integration::get_delivery_context()` — the mode label and the secret key derive from a single `get_stripe_settings()` read, so a concurrent save can no longer pair one mode's label with the other mode's key. `sync_feed()` builds the delivery client from the captured key, and the controller's refresh uses the same snapshot for both entry filtering and polling (the now-redundant `create_delivery()` is removed).

Coverage added: a unit test that alternates `testmode` on every option read and asserts the pair is never split; an end-to-end sync test flipping the settings mid-walk and asserting the upload authenticates with the captured key while the history entry records the matching mode; and a refresh test flipping the settings mid-poll and asserting the current-mode entry is refreshed with the matching key and never reclassified.

PR #5757, opened 2026-07-27
Retry matching a completed Adaptive Pricing checkout session to its order instead of giving up after the first attempt

### LINKED ISSUE #5691: Stripe payments succeeding but WooCommerce orders stuck in "Pending"/cancelled — webhook returns 204 but order never updates
### Prerequisites

- [x] I have carried out troubleshooting steps and I believe I have found a bug.
- [x] I have searched for similar bugs in both open and closed issues and cannot find a duplicate.

### Describe the bug

Plugin: WooCommerce Gateway Stripe, version 10.8.3
Site: gaiaretreat.com.au

Hi,

We're seeing a recurring issue where customers are successfully charged via Stripe, but the corresponding WooCommerce order stays in "Pending payment" and is later auto-cancelled — even though the Stripe Dashboard shows the webhook was delivered.

Affected orders: #17924, #17925, #17926, #17927 (25–27 June 2026), #17929 (30 June 2026). All had to be manually corrected to Processing/Completed after the fact because they never updated automatically.

Image
Example image.​​
​

What we've found so far:

Our currently registered webhook endpoint is https://www.gaia.tattarangvps1.cloudsites.net.au/?wc-api=wc_stripe — an old staging domain, not our live site (https://www.gaiaretreat.com.au). We understand this needs correcting, but wanted to flag it in case it's contributing.

For several of these orders, Stripe's dashboard shows the webhook events (checkout.session.completed, payment_intent.succeeded, charge.succeeded) as "Delivered" with an HTTP 204 response. Looking at your plugin's own webhook handler code (includes/class-wc-stripe-webhook-handler.php, check_for_webhook()), it appears 204 is returned when webhook signature/secret validation fails — not just when a payment is processed successfully with nothing to report. This makes it impossible to tell from the Stripe Dashboard alone whether a webhook was actually processed or silently rejected. Could you confirm whether that's the intended behavior, and help us determine why validation might be failing (e.g. secret mismatch between the registered endpoint and our stored webhook secret)?

Separately, for orders using Adaptive Pricing checkout (checkout_type: adaptive_pricing_checkout in the payment metadata), we noticed in handle_checkout_session_success() that if the order isn't found by checkout session ID within the 2-minute deferred retry window, the plugin logs a warning ("Completed Adaptive Pricing checkout session has no matching order") and does not retry further — leaving a successfully captured payment permanently disconnected from its order. This looks related to the fixes described in your 10.8.2 changelog ("Prevent Adaptive Pricing orders from being stuck in pending after a successful payment"), but we're still seeing the symptom on 10.8.3.

Could you help us with:

Confirming whether our webhook signing secret matches what's currently registered for our live domain (we suspect the staging URL above may be using a stale/mismatched secret)
Advising on the correct way to update our webhook endpoint from the staging URL to the live domain without causing a gap in coverage
Letting us know if there's a more complete fix for the Adaptive Pricing order-matching gap beyond what shipped in 10.8.2
Happy to share the raw Stripe event JSON, screenshots, and exact order/event IDs for each case — just let us know the best way to send them through.

Thanks,
Claud Chan

### Expected behavior

Automatically marked the order is completed.

### Actual behavior

Remain processing until it expired change to cancel.

### Steps to reproduce

Orders from now on processing

### WordPress Environment

Affected orders (for reference, not for the public issue if you'd rather keep these private — see note below)
#17924–17927 (25–27 Jun 2026), #17929 (30 Jun 2026)

Additional context / what we've found in the code

Plugin version: 10.8.3
In includes/class-wc-stripe-webhook-handler.php, check_for_webhook() returns HTTP 204 both on validation failure and is the only 2xx-family status shown when the deeper processing logic silently can't match an order — meaning the Stripe Dashboard's "204 Delivered" gives no signal about whether the order was actually updated.
In handle_checkout_session_success(), when a webhook can't find a matching order by checkout session ID for an Adaptive Pricing checkout (even after the 2-minute deferred retry), the code logs a warning and permanently gives up with no further retry — this looks related to the 10.8.2 changelog entry "Prevent Adaptive Pricing orders from being stuck in pending after a successful payment," but the symptom still occurs on 10.8.3.
Our webhook endpoint is currently registered against an old staging domain rather than our live domain, which we're correcting separately, but wanted to flag in case it's a contributing factor (e.g. signing secret mismatch).
Happy to share the raw Stripe event JSON and screenshots privately if that's more appropriate than posting them in a public GitHub issue.

### Isolating the problem

- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/products/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

wjrosa:
Fixes STRIPE-1272
Fixes #5691

When a `checkout.session.completed` (or `checkout.session.async_payment_succeeded`) webhook arrives for an Adaptive Pricing session and no order is linked to the session yet, the handler deferred the event once (2 minutes) and then gave up permanently with a warning. The order is created and linked by the shopper's browser via the Store API, so a slow, dropped, or retried checkout request could outlast that single window — permanently orphaning a captured payment and leaving the order stuck in `pending` until WooCommerce's hold-stock cron cancelled it.

This PR re-queues the event a bounded number of times (3 additional retries at the standard 2-minute deferred delay, so lookups at roughly 2/4/6/8 minutes after the event) before conceding with the existing warning, which now includes the retry count. The retry count travels in the Action Scheduler job args (`retry_count`), so jobs queued before this change simply start at 0 and gain the retries.

Related: #5634 recovers the same class of orphaned Adaptive Pricing payments from the `payment_intent.succeeded` side; this PR closes the gap on the `checkout.session.*` side. The two are complementary.

### Backward compatibility

- `handle_checkout_session_success()` (protected) keeps its single-parameter signature. The deferred retry count is carried on a private instance field rather than a new parameter, so third-party subclasses overriding this protected method are unaffected. Adding the parameter would fatal such a subclass on PHP 8+ that kept the one-argument override (warning on 7.4), so the signature is deliberately left unchanged.
- The `wc_stripe_deferred_webhook` Action Scheduler job args gain an optional `retry_count` key; jobs without it behave as before (plus retries).
- `defer_webhook_processing()`'s docblock param type is widened from `stdClass` to `object` to match all existing call sites (doc-only).
- No hooks, options, or public signatures change.

## Testing instructions

1. Enable Optimized Checkout / Adaptive Pricing and start a Blocks checkout so a Stripe Checkout Session with `checkout_type: adaptive_pricing_checkout` metadata is created.
2. Simulate the order-creation request being slow or lost (e.g. block the Store API checkout request in dev tools after `checkout.confirm()` succeeds, or add a long `sleep` in the Store API checkout path).
3. Let Stripe deliver `checkout.session.completed` (e.g. via `npm run listen`).
4. Observe in the logs that the plugin now logs `Completed Adaptive Pricing checkout session has no matching order yet; re-queueing.` and schedules `wc_stripe_deferred_webhook` jobs (visible under WooCommerce → Status → Scheduled Actions) instead of warning and dropping the event.
5. Let the order-creation request complete before the retries are exhausted: the next retry finds the order and settles it (order moves to processing/completed).
6. If the order never appears, after the 3rd retry the handler logs `Completed Adaptive Pricing checkout session has no matching order after 3 retries: cs_…` and stops re-queueing.

Automated coverage: `WC_Stripe_Webhook_Handler_Test::test_deferred_adaptive_pricing_session_without_order_retries_are_bounded` (data provider: first run re-queues, last allowed retry re-queues, max retries gives up; also asserts `wc_stripe_webhook_received` does not fire while re-queued).

---

### DIFF
--- a/includes/class-wc-stripe-webhook-handler.php
+++ b/includes/class-wc-stripe-webhook-handler.php
@@ -81,6 +81,14 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	protected $locked_order_retry_delay = 10;
 
 	/**
+	 * How many times to re-queue a completed Adaptive Pricing checkout session whose order
+	 * cannot be found yet. Bounded so a session that will never gain an order stops retrying.
+	 *
+	 * @var int
+	 */
+	protected $adaptive_pricing_order_lookup_max_retries = 3;
+
+	/**
 	 * The Action Scheduler hook to use when retrying a webhook.
 	 *
 	 * @var string
@@ -1548,7 +1556,7 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	 *
 	 * Each Webhook type which is deferred should be supported by @see process_deferred_webhook().
 	 *
-	 * @param stdClass $webhook_notification The webhook payload received from Stripe.
+	 * @param object   $webhook_notification The webhook payload received from Stripe.
 	 * @param array    $additional_data      Additional data to pass to the scheduled job.
 	 * @param int|null $delay                Seconds to wait before retrying. Defaults to $deferred_webhook_delay.
 	 */
@@ -1667,9 +1675,9 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 					break;
 				case 'checkout.session.completed':
 				case 'checkout.session.async_payment_succeeded':
-					// If the order is still locked, this re-queues itself again; don't fire the
-					// action now — the next retry fires it once settlement actually runs.
-					if ( $this->handle_checkout_session_success( $notification ) ) {
+					// If the order is still locked or not created yet, this re-queues itself again;
+					// don't fire the action now — the retry that settles the payment fires it.
+					if ( $this->handle_checkout_session_success( $notification, (int) ( $additional_data['retry_count'] ?? 0 ) ) ) {
 						return;
 					}
 					break;
@@ -2096,10 +2104,11 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 	/**
 	 * Handles a deferred checkout session success event.
 	 *
-	 * @param object        $notification The Stripe notification containing the checkout session data.
+	 * @param object $notification The Stripe notification containing the checkout session data.
+	 * @param int    $retry_count  How many times this event has already been re-queued while waiting for its order.
 	 * @return bool True if the event was re-queued for async processing, false if handled inline.
 	 */
-	protected function handle_checkout_session_success( object $notification ): bool {
+	protected function handle_checkout_session_success( object $notification, int $retry_count = 0 ): bool {
 		$checkout_session = $notification->data->object;
 
 		$session_id = $checkout_session->id;
@@ -2127,10 +2136,31 @@ class WC_Stripe_Webhook_Handler extends WC_Stripe_Payment_Gateway {
 			// paid session whenever agentic commerce is disabled.
 			$checkout_type = $checkout_session->metadata->checkout_type ?? '';
 			if ( WC_Stripe_Checkout_Sessions_Ajax_Handler::ADAPTIVE_PRICING_CHECKOUT_TYPE === $checkout_type ) {
-				WC_Stripe_Logger::warning(
-					'Completed Adaptive Pricing checkout session has no matching order: ' . $checkout_session->id
-				);
 				WC_Stripe_Database_Cache::delete( $lock_key );
+
+				// The order is created and linked by the shopper's browser via the Store API,
+				// which can outlast a single deferred window — re-queue before conceding.
+				if ( $retry_count < $this->adaptive_pricing_order_lookup_max_retries ) {
+					WC_Stripe_Logger::info(
+						'Completed Adaptive Pricing checkout session has no matching order yet; re-queueing.',
+						[
+							'session_id'  => $session_id,
+							'retry_count' => $retry_count + 1,
+						]
+					);
+					$this->defer_webhook_processing(
+						$notification,
+						[
+							'session_id'  => $session_id,
+							'retry_count' => $retry_count + 1,
+						]
+					);
+					return true;
+				}
+
+				WC_Stripe_Logger::warning(
+					'Completed Adaptive Pricing checkout session has no matching order after ' . $retry_count . ' retries: ' . $checkout_session->id
+				);
 				return false;
 			}
 
--- a/tests/phpunit/class-wc-stripe-webhook-handler-test.php
+++ b/tests/phpunit/class-wc-stripe-webhook-handler-test.php
@@ -1333,6 +1333,101 @@ class WC_Stripe_Webhook_Handler_Test extends WP_UnitTestCase {
 	}
 
 	/**
+	 * An Adaptive Pricing session with no linked order yet must be re-queued a bounded number
+	 * of times instead of being dropped on the first deferred miss.
+	 *
+	 * @param int      $retry_count      Retry count carried by the deferred job being executed.
+	 * @param int|null $expected_requeue Retry count of the re-queued job, or null when the handler gives up.
+	 * @dataProvider provide_adaptive_pricing_order_lookup_retry_counts
+	 */
+	public function test_deferred_adaptive_pricing_session_without_order_retries_are_bounded( int $retry_count, ?int $expected_requeue ): void {
+		$checkout_session_id = 'cs_test_ap_retry_' . $retry_count;
+
+		WC_Stripe_Database_Cache::delete( 'checkout_session_lock_' . $checkout_session_id );
+		WC_Stripe_Database_Cache::delete( 'checkout_session_' . $checkout_session_id );
+
+		$notification = (object) [
+			'type' => 'checkout.session.completed',
+			'data' => (object) [
+				'object' => (object) [
+					'id'             => $checkout_session_id,
+					'payment_intent' => 'pi_test_ap_retry',
+					'metadata'       => (object) [
+						'checkout_type' => WC_Stripe_Checkout_Sessions_Ajax_Handler::ADAPTIVE_PRICING_CHECKOUT_TYPE,
+					],
+				],
+			],
+		];
+
+		$start          = time();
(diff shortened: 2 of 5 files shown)


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2297,10 +2306,31 @@ protected function handle_checkout_session_success( object $notification ): bool
 			// paid session whenever agentic commerce is disabled.
 			$checkout_type = $checkout_session->metadata->checkout_type ?? '';
 			if ( WC_Stripe_Checkout_Sessions_Ajax_Handler::ADAPTIVE_PRICING_CHECKOUT_TYPE === $checkout_type ) {
+				WC_Stripe_Database_Cache::delete( $lock_key );
+
+				// The order is created and linked by the shopper's browser via the Store API,
+				// which can outlast a single deferred window — re-queue before conceding.
+				if ( $retry_count < $this->adaptive_pricing_order_lookup_max_retries ) {
+					WC_Stripe_Logger::info(
```
daledupreez, line 2336: Not blocking: Given that we're computing `$retry_count + 1` in two locations, is it worth using a local variable to compute that once?
```suggestion
					$next_retry_count = $retry_count + 1;
					WC_Stripe_Logger::info(
```

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -1676,6 +1676,101 @@ public function provide_deferred_checkout_session_success_event_types(): array {
...
+		}
+	}
+
+	/**
+	 * @return array<string, array{0: int, 1: int|null}>
+	 */
+	public function provide_adaptive_pricing_order_lookup_retry_counts(): array {
+		return [
+			'first deferred run re-queues'   => [ 0, 1 ],
+			'last allowed retry re-queues'   => [ 2, 3 ],
```
daledupreez, line 1788: Just to be exhaustive, it would probably be worth checking that the second iteration also re-queues.
```suggestion
			'second deferred run re-queues'  => [ 1, 2 ],
			'last allowed retry re-queues'   => [ 2, 3 ],
```

daledupreez (approved):
The changes are working as described for me, but I was not able to follow the testing instructions, as we need the Store API response to return before we trigger the `confirm()` call in the client. I worked around that by adding a server-side sleep and then using WP CLI to get the checkout session ID meta, remove the meta for the order, save the updated order meta, wait a few minutes, and then reinstate the checkout session ID meta field.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Done in fd010945b. Extracted `$next_retry_count` and used it in both the logger and the defer call.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2266,10 +2274,11 @@ public function process_checkout_session_success( object $notification ): bool {
 	/**
 	 * Handles a deferred checkout session success event.
 	 *
-	 * @param object        $notification The Stripe notification containing the checkout session data.
+	 * @param object $notification The Stripe notification containing the checkout session data.
+	 * @param int    $retry_count  How many times this event has already been re-queued while waiting for its order.
 	 * @return bool True if the event was re-queued for async processing, false if handled inline.
 	 */
-	protected function handle_checkout_session_success( object $notification ): bool {
+	protected function handle_checkout_session_success( object $notification, int $retry_count = 0 ): bool {
```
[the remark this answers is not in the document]
wjrosa: Confirmed and fixed in 281092d5e. Verified the break: adding the param is a warning on PHP 7.4 but a fatal on 8.x for a subclass that kept the one-arg override. Reverted `handle_checkout_session_success()` to its single-parameter signature and carry the deferred retry count on a private instance field instead. Added a test asserting the signature stays single-parameter, and noted the BC impact in the PR description.

wjrosa:
@daledupreez thanks for the review! I'd appreciate another round if possible since I addressed https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5757#discussion_r4047338247

tests/phpunit/class-wc-stripe-webhook-handler-test.php
```diff
@@ -1666,6 +1666,24 @@ function ( $passed ) use ( $notification ) {
...
+		$method = new ReflectionMethod( WC_Stripe_Webhook_Handler::class, 'handle_checkout_session_success' );
+
+		$this->assertTrue( $method->isProtected() );
+		$this->assertSame(
+			1,
+			$method->getNumberOfParameters(),
+			'Adding a parameter to this protected method breaks subclasses that override the one-argument signature on PHP 8+.'
+		);
+	}
+
```
daledupreez: This feels unnecessary. I think inspection is sufficient to verify that a method signature didn't change. 😁 
```suggestion
```

daledupreez:
@wjrosa, it looks like there is a legitimate linting issue to address. (Testing now.)

daledupreez (approved):
I just retested and the restructured code tested well. It does look like CodeRabbit has some legitimate comments that could be worth addressing too.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2297,10 +2327,32 @@ protected function handle_checkout_session_success( object $notification ): bool
...
+						[
+							'session_id'  => $session_id,
+							'retry_count' => $next_retry_count,
+						]
+					);
+					$this->defer_webhook_processing(
+						$notification,
+						[
+							'session_id'  => $session_id,
+							'retry_count' => $next_retry_count,
```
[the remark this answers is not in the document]
wjrosa: The dedup gap is real: `schedule_job()` replaces only on a full-args match, so deliveries at different `retry_count`s (or the `completed` vs `async_payment_succeeded` pair) keep separate retry chains for one session. But the severe outcome is already covered: the per-session `checkout_session_lock` serializes concurrent runs and the `wc_stripe_allowed_payment_processing_statuses` guard makes a chain that finds an already-paid order return early, so there is no double settlement or double charge. The remaining effect is a possible extra `wc_stripe_webhook_received` fire on a terminal duplicate, which is inherent to Stripe's at-least-once delivery and something listeners already handle. Reducing redundant chains via a session-scoped dedup key is worthwhile, but it changes the scheduling contract, so I would rather track it as a follow-up than expand this PR.

PR #5758, opened 2026-07-27
Close order-lock race gaps: re-queue dropped webhooks, pre-order settlement, double-checked locking

### LINKED ISSUE #5753: Address AI review issues on PR 5050
### Description

This issue reports some findings from an AI review of #5050, which still need to be fully validated:

#### 1. Skip paths forgo confirm-time side effects no webhook path replicates

`includes/class-wc-stripe-intent-controller.php:765`

When the already-settled or lock-loser branch skips `process_order_for_confirmed_intent()`:
- charged-upfront pre-orders never get `mark_order_as_pre_ordered()` — the strongest harm, since the pre-order lifecycle is silently broken;
- the `woocommerce_stripe_add_payment_method` and `wc_stripe_after_set_payment_method_title_for_confirmed_intent` hooks never fire — a hook-firing-condition change, which should be considered a backwards-incompatible contract change.

Verified: neither the immediate nor the deferred webhook succeeded handler replicates any of this (both call only `process_response()`).

**Reachability (corrected by the critic pass):** this endpoint is only called from `confirmIntent()` on the `#wc-stripe-confirm-pi|si` hash — 3DS/SCA cards and default-case methods (e.g. BLIK, ACH). Wallet (CashApp/WeChat) and voucher (Boleto/OXXO/Multibanco) flows use their own hash routes and redirect straight to the return URL, so they never reach this skip. The race that triggers it is a 3DS card confirmation vs the deferred Action Scheduler webhook job, or ACH `processing` → on-hold events — real but narrower than "systematic". An earlier claim that saved-payment-method choices could be dropped for retrievable-type-mismatch methods was withdrawn: those methods (Bancontact/iDEAL/Sofort) are redirect flows that never hit this endpoint.

**Recommendation:** extract these side effects into a shared idempotent post-settlement routine invoked regardless of which path (AJAX or webhook) settles the order — or at minimum replicate pre-order marking in the webhook settlement path.

#### 2. Concurrent payment_intent webhooks are silently dropped while the AJAX holds the lock

`includes/class-wc-stripe-intent-controller.php:770`

The immediate webhook handler acks-and-drops with a 200 on lock contention (`class-wc-stripe-webhook-handler.php:1350-1352`) — no Stripe retry, no re-queue — unlike the deferred handler, which re-queues (`:1653-1656`). Before this change the AJAX path never held this lock, so this is new exposure: an immediate-mode event (e.g. an ACH `processing` → on-hold event) landing while the AJAX holds the lock across its Stripe API round-trips is lost, with no automatic repair if the AJAX subsequently fails the order. (Card `payment_intent.succeeded` events go through the deferred handler, which re-queues — those are safe.) Note the team fixed this exact drop-on-contention shape on the webhook side twice recently (#5576, #5626) by replacing drop with re-queue.

**Recommendation:** make the immediate `process_payment_intent()` handler re-queue on lock contention, mirroring the deferred handler.

#### 3. Residual TOCTOU between the settled-check and lock acquisition

`includes/class-wc-stripe-intent-controller.php:765`

`$already_settled` reads the in-memory order snapshot loaded earlier in the request (no DB re-read), while `lock_order_payment()` immediately after forces a fresh meta read. A concurrent request that fully settles and unlocks inside that window is invisible to the settled-check, so this request can still double-process — the exact duplication the PR targets, with a much smaller but still-open window.

**Recommendation:** re-check settlement state after acquiring the lock (double-checked locking), or force a fresh order read before the settled-check.

</details>

wjrosa:
Fixes STRIPE-1308
Fixes #5753

Closes the three race gaps the AI review found in the #5050 order-lock change:

1. **Immediate `payment_intent` webhooks were dropped on lock contention.** The immediate handler acked 200 and returned while another request (now including the AJAX confirm, which holds this lock since #5050) held the payment lock — Stripe never retries an acked event, so `payment_intent.processing`-style events were lost. It now re-queues via the deferred machinery with the short locked-order retry delay (mirroring #5576/#5626), and the deferred processor re-enters the immediate handler for `processing`/`requires_action`/`payment_failed` retries, re-running its guards each time. Retries are bounded by the lock's 5-minute TTL and stop once the allowed-status guard sees the order settled.
2. **Webhook-settled charged-upfront pre-orders skipped the pre-order lifecycle.** When the AJAX confirm loses the settlement race (or the shopper never returns), the webhook was the only settle path and ran the standard completion. Both webhook settle paths (deferred and inline) now route pre-orders through `mark_order_as_pre_ordered()` instead of `process_response()`, matching the confirm path.
3. **Residual TOCTOU in `update_order_status_ajax`.** The settled-check read a snapshot loaded before the lock, so a concurrent request settling and unlocking in that window was invisible. Added double-checked locking: after acquiring the lock, settlement is re-checked against a fresh order read and processing is skipped (with the lock released) when a concurrent request won.

Not addressed here (noted in the issue as the larger refactor): full parity for the `woocommerce_stripe_add_payment_method` / `wc_stripe_after_set_payment_method_title_for_confirmed_intent` hooks on webhook-settled orders — that needs the shared post-settlement routine and payment-method saving in the webhook path.

**BC impact:** no public signatures changed. The immediate webhook handler's behavior on lock contention changes from silent drop to deferred retry (new Action Scheduler jobs); a new protected method on the webhook handler.

## Testing instructions

1. With a 3DS card, complete checkout and let the deferred webhook race the AJAX confirm — the order settles exactly once either way.
2. For the webhook drop: hold the lock (e.g. pause the AJAX in a debugger) while an ACH `payment_intent.processing` event arrives — an Action Scheduler retry job is created instead of the event disappearing.
3. With WooCommerce Pre-Orders and a charged-upfront pre-order paid via 3DS: kill the browser after authentication so only the webhook settles — the order ends up pre-ordered, not processing.

---

### Changelog entry

Changelog entry added to `changelog.txt` and `readme.txt` under 10.9.0 (Fix).

<details>

#### Comment

</details>

**Post merge**


includes/class-wc-stripe-intent-controller.php
```diff
@@ -771,12 +771,26 @@ public function update_order_status_ajax() {
...
-					} finally {
-						// This request owns the lock (we just acquired it above), so it must release it.
+					// The settled-check above ran before the lock; re-check against a
+					// fresh read in case a concurrent request settled in between.
+					$fresh_order = wc_get_order( $order_id );
+					if ( $fresh_order instanceof WC_Order ) {
+						$order = $fresh_order;
+					}
+					$settled_at_lock = $order->has_status( [ OrderStatus::PROCESSING, OrderStatus::COMPLETED, OrderStatus::ON_HOLD ] )
+						|| $order_helper->get_stripe_upe_redirect_processed( $order );
```
jaclync: nit: could this `settled_at_lock` logic be a private function shared in the two places, here and L765-L766? Something like `WC_Stripe_Order_Helper`'s new helper `is_order_payment_settled( $order )`?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1347,7 +1347,17 @@ public function process_payment_intent( $notification ) {
 
 		$order_helper = WC_Stripe_Order_Helper::get_instance();
 
+		// Re-queue instead of dropping: the endpoint already acked 200, so Stripe
+		// never retries. Bounded by the lock TTL and the status guard above.
```
jaclync: non-blocking ❓ about the comment: where is the 200 ack that already happened before this line?

includes/class-wc-stripe-intent-controller.php
```diff
@@ -771,12 +771,26 @@ public function update_order_status_ajax() {
...
 					WC_Stripe_Logger::info( "Skipping update_order_status_ajax for order $order_id; order payment is already locked." );
 				} else {
-					$processing_started = true; // Past validation; a failure from here on is a genuine payment failure.
-					try {
-						$gateway->process_order_for_confirmed_intent( $order, $intent_id_received, $save_payment_method );
-					} finally {
-						// This request owns the lock (we just acquired it above), so it must release it.
+					// The settled-check above ran before the lock; re-check against a
+					// fresh read in case a concurrent request settled in between.
+					$fresh_order = wc_get_order( $order_id );
```
jaclync: ❓ The re-check after L770 lock reads through WordPress's request-local object cache, so on legacy (CPT) order storage without a persistent object cache it can't observe a concurrent settlement: `wc_get_order()` returns the row cached by the `wc_get_order()` at the top of this handler, and the meta read is non-forced. It's not easy to simulate this in unit tests with another PHP process.

The redirect handler already handles this case by clearing the cache before re-reading:

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/b8eae223e4a16350a43ca923df1212754c0ef831/includes/payment-methods/class-wc-stripe-upe-payment-gateway.php#L4741-L4743

Do you think the cache could be an issue here?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1851,6 +1880,26 @@ protected function handle_deferred_payment_intent_succeeded( $order, $intent_id
...
+	 *
+	 * @param WC_Order $order The order being settled.
+	 * @return bool True when handled as a pre-order (skip standard completion).
+	 */
+	protected function maybe_mark_order_as_pre_ordered( $order ): bool {
+		$gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+
+		if ( ! $gateway->has_pre_order( $order->get_id() ) ) {
+			return false;
+		}
```
jaclync: ❓ `has_pre_order()` is true for the order's lifetime, so the release payment of a charge-upon-release pre-order also passes through this gate - the status gets re-marked pre-ordered and skips `process_response()` -> a captured release payment never completes the order (reachable via the failed/3DS release flow, since failed passes the allowed-status guard). Do you think this could be an issue?

non-blocking: could we add E2E coverage for the 3DS pre-orders + charged-upfront/charged-upon-release, since the pre-orders handling is a bit scattered and it's not a flow we test often?

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1399,6 +1409,11 @@ public function process_payment_intent( $notification ) {
 
 				// Process the webhook now if it's for a voucher, wallet, or BLIK payment, or if filtered to process immediately and order is not awaiting action.
 				if ( $is_voucher_payment || $is_wallet_payment || $is_blik_payment || ( ! $process_webhook_async && ! $is_awaiting_action ) ) {
+					if ( $this->maybe_mark_order_as_pre_ordered( $order ) ) {
+						$this->run_webhook_received_action( (string) $notification->type, $notification );
+						break;
+					}
```
jaclync: ❓ This gate (and its twin in `handle_deferred_payment_intent_succeeded()`) also catches `payment_intent.amount_capturable_updated`, which shares this switch branch with `payment_intent.succeeded`.

On a manual-capture store, a charged-upfront pre-order's authorization now skips `process_response()` - is this intentional? The capture flow in `capture_payment()` can silently skip because of the lack of `process_response()` and the 7-day authorization expires uncaptured - pre-order fulfilled, money never collected.

(The confirm path has the same gap: `process_order_for_confirmed_intent()` also skips `process_response()` for pre-orders, but pre-PR the webhook path was the one that still recorded the capture meta while this PR removes the last path that did.)

Potentially we could:
1. (Probably the smallest change) Exclude uncaptured intents from the gate so `amount_capturable_updated` falls through to `process_response()` as before.
2. Record the transaction ID / captured flag inside `maybe_mark_order_as_pre_ordered()`, which would fix the confirm path too.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1677,6 +1692,16 @@ public function process_deferred_webhook( $webhook_type, $additional_data, $noti
...
+				case 'payment_intent.processing':
+				case 'payment_intent.requires_action':
+				case 'payment_intent.payment_failed':
+					/**
+					 * Retry of an immediate-mode event that hit the payment lock.
+					 *
+					 * @var stdClass $notification
+					 */
+					$this->process_payment_intent( $notification );
+					return;
```
jaclync: ❓ The `return;` here changes when `wc_stripe_webhook_received` fires for these three types, as `run_webhook_received_action()` below the switch in L1710 is skipped. WDYT of making `process_payment_intent()` return whether it's processed or requeued, so that it can determine whether to call `run_webhook_received_action()`?

The checkout-session events solved a similar case with a "was-deferred" signal: [process_checkout_session_success()](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/b8eae223e4a16350a43ca923df1212754c0ef831/includes/class-wc-stripe-webhook-handler.php#L2112) returns whether the event was deferred.

jaclync (commented):
Some initial questions about the changes, I will test the steps in the next pass!

includes/class-wc-stripe-intent-controller.php
wjrosa: Done in fe0374aa3: `WC_Stripe_Order_Helper::is_order_payment_settled()`, used in both places.

includes/class-wc-stripe-webhook-handler.php
wjrosa: It is after, not before: the endpoint returns 200 whether or not the event was applied here, so Stripe will not retry it. Reworded the comment in fe0374aa3.

includes/class-wc-stripe-intent-controller.php
wjrosa: Yes. The re-read now clears the post cache, and the HPOS order cache when present, before loading the order again, like the redirect handler. fe0374aa3.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Good catch. The gate now requires an unreleased charged-upfront pre-order, so the release payment of a charge-upon-release pre-order goes through `process_response()` as usual. fe0374aa3.

E2E for the pre-order flows: agreed, I will track it as a follow-up.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Went with option 2. `maybe_mark_order_as_pre_ordered()` now stores the transaction ID and captured flag from the charge before marking, so `amount_capturable_updated` leaves the order capturable. Tests cover the captured and uncaptured cases. fe0374aa3.

includes/class-wc-stripe-webhook-handler.php
wjrosa: Done. `process_payment_intent()` returns true on a lock re-queue, and both the live pass and the deferred retry hold the action until the event is actually applied. fe0374aa3.

includes/class-wc-stripe-intent-controller.php
```diff
@@ -795,21 +795,38 @@ public function update_order_status_ajax() {
...
 						$order_helper->unlock_order_payment( $order );
+					} else {
+						$processing_started = true; // Past validation; a failure from here on is a genuine payment failure.
+						try {
+							$gateway->process_order_for_confirmed_intent( $order, $intent_id_received, $save_payment_method );
+						} finally {
+							// This request owns the lock (we just acquired it above), so it must release it.
+							$order_helper->unlock_order_payment( $order );
+						}
 					}
```
jaclync: non-blocking nit: it might be easier to read with the branches in the `try`, since the final logic is the same:

```suggestion
					try {
						if ( $order_helper->is_order_payment_settled( $order ) ) {
							WC_Stripe_Logger::info( "Skipping update_order_status_ajax for order $order_id; order was settled by a concurrent request." );
						} else {
							$processing_started = true; // Past validation; a failure from here on is a genuine payment failure.
							$gateway->process_order_for_confirmed_intent( $order, $intent_id_received, $save_payment_method );
						}
					} finally {
						// This request owns the lock (we just acquired it above), so it must release it.
						$order_helper->unlock_order_payment( $order );
					}
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1931,6 +1968,53 @@ protected function handle_deferred_payment_intent_succeeded( $order, $intent_id
...
+	 * Whether the order is a charged-upfront pre-order that has not been released yet.
+	 *
+	 * The pre-order flag stays on the order for its lifetime, so has_pre_order() alone
+	 * would also catch the release payment of a charge-upon-release pre-order, which
+	 * must complete the order through process_response() like any other payment.
+	 *
+	 * @param WC_Order $order The order.
+	 * @return bool
+	 */
+	protected function is_unreleased_charged_upfront_pre_order( $order ): bool {
```
jaclync, line 2015: non-blocking nit: could these two functions be private?

jaclync (approved):
Tested with scripts and the pre-orders case, worked as expected :shipit:

includes/class-wc-stripe-webhook-handler.php
wjrosa: I kept them protected on purpose: both are stubbed in the webhook handler tests (`setMethods` with `maybe_mark_order_as_pre_ordered` and `is_unreleased_charged_upfront_pre_order`), so the pre-order path can be exercised without building a real pre-order product and gateway state in the test. PHPUnit cannot stub private methods, so making them private means rewriting those tests. It also matches the sibling seams in this file, like `maybe_flag_unexpected_charge_on_order` and `handle_deferred_payment_intent_succeeded`, which are protected for the same reason.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -1851,6 +1880,26 @@ protected function handle_deferred_payment_intent_succeeded( $order, $intent_id
...
+	 */
+	protected function maybe_mark_order_as_pre_ordered( $order ): bool {
+		$gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+
+		if ( ! $gateway->has_pre_order( $order->get_id() ) ) {
+			return false;
+		}
+
+		WC_Stripe_Logger::info( "Marking order {$order->get_id()} as pre-ordered via webhook settlement." );
+		$gateway->mark_order_as_pre_ordered( $order );
```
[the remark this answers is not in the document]
wjrosa: Already covered by the current version of `maybe_mark_order_as_pre_ordered()`: it takes the charge, stores the transaction ID and the captured flag, and saves the order before marking it pre-ordered, so later captures and refunds can find the charge. `test_deferred_payment_intent_succeeded_records_charge_on_pre_order` covers both. The intent ID is already on the order, since that is how the webhook finds it. The mandate ID is not stored here on purpose: only the subscriptions renewal code reads it, and a charged-upfront pre-order never charges the shopper again from this order.

daledupreez:
@wjrosa, I am just noting for the record that this PR was never cherry-picked across to the `release/11.0.0` branch, so it will only be included in 11.1.0.

PR #5759, opened 2026-07-27
Verify the payment status with Stripe before auto-cancelling an unpaid order

wjrosa:
Fixes STRIPE-1272
Related to #5691

WooCommerce's hold-stock cron cancels pending orders that look unpaid locally — but a lost or misrouted webhook can leave a *paid* order in exactly that state: the shopper stays charged while the order is auto-cancelled.

The existing `woocommerce_cancel_unpaid_order` guard (`prevent_cancelling_orders_awaiting_action()`) already fetches the order's PaymentIntent from Stripe; it just never looked at its **status**. Now, when Stripe reports `succeeded` or `requires_capture`, the cancellation is blocked and the order is settled through `process_response()` — the same path a webhook would have taken — with an order note explaining what happened. No additional API calls.

If a concurrent webhook/redirect holds the order payment lock, only the cancellation is blocked; the lock holder finishes settling. Any other intent status (e.g. an abandoned checkout) falls through to the existing behavior unchanged.

Complements #5757 (fixes the primary webhook-side path; this is the safety net at cancellation time) and #5703 (refunds payments that land after an order was already cancelled).

### Backward compatibility

- No hooks, options, or public signatures change; one new private method.
- The `woocommerce_cancel_unpaid_order` filter can now return `false` in one more situation — strictly fewer wrong cancellations.
- Cron-context only, no session/cart dependencies; all reads/writes are order-scoped, so multisite-safe.

## Testing instructions

1. Enable hold stock (WooCommerce → Settings → Products → Inventory).
2. Pay for an order with Stripe but keep the confirmation from reaching the site (e.g. stop `npm run listen` and set the order back to `pending` so it keeps its intent meta).
3. Run the `woocommerce_cancel_unpaid_orders` scheduled action (WooCommerce → Status → Scheduled Actions).
4. The order must NOT be cancelled: it moves to processing/completed (captured) or on-hold (uncaptured), with an explanatory order note.
5. A genuinely abandoned checkout is still cancelled as before.

Automated coverage in `WC_Stripe_Order_Handler_Test`: paid-intent settlement (data provider: `succeeded`, `requires_capture`), locked-order case, unpaid-intent fall-through.

---

cc @daledupreez


includes/class-wc-stripe-order-handler.php
```diff
@@ -509,11 +509,22 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
 		// Bail if the order doesn't have an intent yet.
-		if ( ! $this->get_intent_from_order( $order ) ) {
+		if ( ! $intent ) {
 			return $cancel_order;
 		}
 
+		// A lost or misrouted webhook leaves a paid order looking unpaid locally, so ask
+		// Stripe before cancelling — cancelling would orphan the captured payment.
+		if ( 'payment_intent' === ( $intent->object ?? '' )
+			&& in_array( $intent->status ?? '', [ WC_Stripe_Intent_Status::SUCCEEDED, WC_Stripe_Intent_Status::REQUIRES_CAPTURE ], true )
```
diegocurbelo: Shouldn't this also check for `PROCESSING` status?

diegocurbelo (commented):
The the approach seems right, however I was unable to test it following the test instructions:

> 1. Enable hold stock (WooCommerce → Settings → Products → Inventory).

- `Hold stock (minutes)` set to 2 minutes

> 2. Pay for an order with Stripe but keep the confirmation from reaching the site (e.g. stop npm run listen and set the order back to pending so it keeps its intent meta).

- Webhooks configured/enabled, but not listening.
- Placed an order and paid with CashApp.
    - After redirection to the order received page the the order status was still `Pending payment` as the webhook was never processed (as expected)

> 3. Run the woocommerce_cancel_unpaid_orders scheduled action (WooCommerce → Status → Scheduled Actions).

- After 2 min, run the `woocommerce_cancel_unpaid_orders` and executed with no errors

> 4. The order must NOT be cancelled: it moves to processing/completed (captured) or on-hold (uncaptured), with an explanatory order note.
A genuinely abandoned checkout is still cancelled as before.

- The order was `Cancelled` with this note:
    Image

includes/class-wc-stripe-order-handler.php
wjrosa: Good catch — it does now: switched the check to `WC_Stripe_Intent_Status::SUCCESSFUL_STATUSES`, which includes `processing` (27c139135).

wjrosa:
Thanks for testing @diegocurbelo! Your Cash App order hit exactly the missing `processing` status — fixed in 27c139135, so it now goes on-hold instead of being cancelled. Conflicts with `develop` are also resolved.

includes/class-wc-stripe-order-handler.php
```diff
@@ -551,6 +564,47 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+	 * payment still processing), via process_response()
+	 * — the same path a webhook would take. If a concurrent process holds the payment lock,
+	 * nothing is settled here; blocking the cancellation is enough and the lock holder finishes.
+	 *
+	 * @since 10.9.0
+	 *
+	 * @param WC_Order $order  The pending order about to be cancelled as unpaid.
+	 * @param object   $intent The PaymentIntent fetched from Stripe.
+	 */
+	private function settle_paid_order_instead_of_cancelling( $order, $intent ): void {
```
daledupreez: Is it worth prefixing the method with `maybe_` given that the code may or may not perform the action based on whether the order is locked? And might it be more accurate to use `process` instead of `settle` to reflect that we're calling the `process_response()` method?

includes/class-wc-stripe-order-handler.php
```diff
@@ -529,12 +529,25 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+		// Bail if the order doesn't have an intent yet.
 		if ( ! $intent ) {
 			return $cancel_order;
 		}
 
+		// A lost or misrouted webhook leaves a paid order looking unpaid locally, so ask
+		// Stripe before cancelling — cancelling would orphan the captured payment. Async
+		// methods (e.g. Cash App Pay, ACH) sit at `processing` while the charge clears and
+		// count as paid here too; process_response() parks those orders on-hold until the
+		// charge settles.
```
daledupreez: I think we can simplify the comment here.
```suggestion
		// Before cancelling, check with Stripe if the payment intent has already been paid or
		// is still being processed (which is common for async methods like ACH and Cash App Pay).
		// For those cases, we want to mimic the processing from webhooks.
```

includes/class-wc-stripe-order-handler.php
```diff
@@ -551,6 +564,47 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
 		return $cancel_order;
 	}
 
+	/**
+	 * Settles a pending order whose PaymentIntent Stripe reports as paid (or as an async
+	 * payment still processing), via process_response()
+	 * — the same path a webhook would take. If a concurrent process holds the payment lock,
+	 * nothing is settled here; blocking the cancellation is enough and the lock holder finishes.
+	 *
+	 * @since 10.9.0
```
daledupreez: Nit: We should update the version to 11.1.0.
```suggestion
	 * @since 11.1.0
```

includes/class-wc-stripe-order-handler.php
```diff
@@ -551,6 +564,47 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+		}
+
+		$intent_id     = $intent->id ?? '';
+		$intent_status = $intent->status ?? '';
+
+		try {
+			$charge = $this->get_latest_charge_from_intent( $intent );
+
+			if ( ! is_object( $charge ) ) {
+				WC_Stripe_Logger::warning( "PaymentIntent {$intent_id} reports status {$intent_status} but has no charge; blocked cancellation of order {$order->get_id()} without settling it." );
```
daledupreez: I don't think we should use "settling" in this message, as it's not a term we generally use. I think we should tweak the wording slightly, as well as logging some additional details.
```suggestion
				WC_Stripe_Logger::warning(
					"PaymentIntent {$intent_id} reports status {$intent_status} but has no charge; order cancellation blocked for order {$order->get_id()}, but order status was not updated from {$order->get_status()}.",
					[
						'order_id'      => $order->get_id(),
						'order_status'  => $order->get_status( 'edit' ),
						'intent_id'     => $intent_id,
						'intent_status' => $intent_status,
						'charge'        => $charge,
					]
				);
```

includes/class-wc-stripe-order-handler.php
```diff
@@ -551,6 +564,47 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+				return;
+			}
+
+			WC_Stripe_Logger::info( "Settling order {$order->get_id()} from PaymentIntent {$intent_id} ({$intent_status}) instead of cancelling it as unpaid." );
```
daledupreez: As above, I think we should avoid "Settling" and we should log some additional details in the context.
```suggestion
			WC_Stripe_Logger::info(
				"Processing order {$order->get_id()} from PaymentIntent {$intent_id} ({$intent_status}) instead of cancelling it as unpaid.",
				[
					'order_id'      => $order->get_id(),
					'order_status'  => $order->get_status( 'edit' ),
					'intent_id'     => $intent_id,
					'intent_status' => $intent_status,
					'charge_id      => $charge->id ?? '(no id)',
				]
			);
```

includes/class-wc-stripe-order-handler.php
```diff
@@ -551,6 +564,47 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+
+			$order->add_order_note( __( 'Stripe reports this payment as successful or still processing, but the payment confirmation never reached this site (e.g. a missed webhook). The order was updated from Stripe instead of being auto-cancelled as unpaid.', 'woocommerce-gateway-stripe' ) );
+
+			$this->process_response( $charge, $order );
+		} catch ( Exception $e ) {
+			WC_Stripe_Logger::error( "Failed to settle order {$order->get_id()} from its PaymentIntent before cancellation: " . $e->getMessage() );
```
daledupreez: As above RE "settle".
```suggestion
			WC_Stripe_Logger::error( "Failed to update order {$order->get_id()} from its PaymentIntent before cancellation: " . $e->getMessage() );
```

changelog.txt
```diff
@@ -11,6 +11,7 @@ xxxx-xx-xx - version 11.1.0
 * Fix - Stop disabling Level 3 data account-wide when Stripe rejects it for a request that declares a non-card payment method
 * Fix - Open testing and payment settings documentation links in new tabs
 * Fix - Accept shipping addresses in the express checkout wallet sheet for free-trial subscription carts whose shipping is charged with the recurring payments
+* Fix - Verify the payment status with Stripe before auto-cancelling an unpaid order, completing paid orders whose confirmation never reached the site
```
daledupreez: This description feels a but unclear to me. Maybe something like the following could be clearer?
```suggestion
* Fix - When checking if unpaid orders should be cancelled, get the Stripe payment status, and update orders with Stripe payment details
```

daledupreez (commented):
I have some feedback from inspection that touches on wording and naming, so will pause before manual testing.

includes/class-wc-stripe-order-handler.php
wjrosa: Good call — renamed to `maybe_process_paid_order_instead_of_cancelling()`. `maybe_` reflects the lock no-op and `process` matches the `process_response()` call.

includes/class-wc-stripe-order-handler.php
```diff
@@ -551,6 +562,65 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
+				[
+					'order_id'      => $order->get_id(),
+					'order_status'  => $order->get_status( 'edit' ),
+					'intent_id'     => $intent_id,
+					'intent_status' => $intent_status,
+					'charge_id'     => $charge->id ?? '(no id)',
+				]
+			);
+
+			$order->add_order_note( __( 'Stripe reports this payment as successful or still processing, but the payment confirmation never reached this site (e.g. a missed webhook). The order was updated from Stripe instead of being auto-cancelled as unpaid.', 'woocommerce-gateway-stripe' ) );
```
daledupreez, line 616: cc @diegocurbelo - this would be a clear location for actions relating to webhook (re)configuration.

tests/phpunit/class-wc-stripe-order-handler-test.php
```diff
@@ -317,6 +317,133 @@ public function provide_test_cancel_payment_voids_explicit_preauth(): array {
...
 
+	/**
+	 * An order whose PaymentIntent Stripe reports as paid must be settled, not cancelled as unpaid.
+	 *
+	 * @param string $intent_status PaymentIntent status reported by Stripe.
+	 * @dataProvider provide_paid_intent_statuses
+	 */
+	public function test_settles_order_with_paid_intent_instead_of_cancelling( $intent_status ) {
+		$order = WC_Helper_Order::create_order();
+		$order->set_payment_method( WC_Stripe_UPE_Payment_Gateway::ID );
```
daledupreez: Non-blocking nit: Might it be worth picking a payment method that is expected to have a delayed status update, like ACH?

includes/class-wc-stripe-order-handler.php
```diff
@@ -529,12 +529,23 @@ public function prevent_cancelling_orders_awaiting_action( $cancel_order, $order
...
 		$intent = $this->get_intent_from_order( $order );
+
+		// Bail if the order doesn't have an intent yet.
 		if ( ! $intent ) {
 			return $cancel_order;
 		}
 
+		// Before cancelling, check with Stripe if the payment intent has already been paid or
+		// is still being processed (which is common for async methods like ACH and Cash App Pay).
+		// For those cases, we want to mimic the processing from webhooks.
```
daledupreez, line 543: Is it worth mentioning that we don't care about whether the payment has been captured yet? Even if it's just authorised, we want the order status to be updated to match.

daledupreez (approved):
The changes LGTM and test as described.

tests/phpunit/class-wc-stripe-order-handler-test.php
wjrosa: Good point. Done in 02f86dc26: the test order now uses the ACH gateway, since its delayed status update is the typical way this path is reached.

includes/class-wc-stripe-order-handler.php
wjrosa: Yes, worth stating. Added in 02f86dc26: the comment now says capture state does not matter, an authorized-but-uncaptured payment must also settle the order instead of being cancelled.

PR #5768, opened 2026-07-30
Gate the Agentic Commerce catalog sync on completed onboarding

wjrosa:
Towards STRIPE-1319
Stacked on #5663.

Follow-up to #5663. That PR adds a merchant-toggle gate to `sync_feed()` (so a scheduled sync can't re-upload after the merchant disables Agentic Commerce). This PR **tightens that same gate to require completed onboarding** — the merchant toggle on **and** the webhook secret saved — via a new `is_onboarding_complete()` helper.

Why: the scheduled and immediate feed syncs run through `sync_feed()`, and the recurring job is scheduled as soon as the developer feature flag is on. Before this, once the flag was on and an account connected, the catalog could be pushed to Stripe before the merchant enabled the feature or finished webhook setup. #5663 closes the toggle half; this closes the pre-onboarding half and adds the webhook-secret requirement.

The delta over #5663 is small: swap its `is_merchant_enabled()` gate for `is_onboarding_complete()`. The teardown push still bypasses the gate via `run_feed_sync()`, so disabling still emits the final checkout-disabled feed.

Per review feedback: `wp stripe agentic-commerce sync --push` delivered the feed directly without going through `sync_feed()`, so it skipped the gate entirely. The CLI now applies the same `is_onboarding_complete()` check before pushing, with a new `--force` flag as an explicit development escape hatch.

**BC impact:** No public signatures/hooks/option keys change (adds one public static method). Behavior: a store with the feature flag on but incomplete onboarding no longer syncs until the toggle is on and the webhook secret is saved.

## Testing instructions

1. Enable the feature flag and connect Stripe (test keys); do **not** enable Agentic Commerce / save a webhook secret. Trigger a scheduled/immediate sync (recurring action, or "Sync now" in the UI) → no upload; log shows `Sync skipped - onboarding incomplete`.
2. Enable in settings but leave the webhook secret empty → still skipped.
3. Save a webhook secret → the catalog syncs.
4. With onboarding incomplete, `wp stripe agentic-commerce sync --push` errors with `Onboarding incomplete (...). Cannot push feed. Use --force to override.` and uploads nothing; `--push --force` still pushes; with onboarding complete, `--push` works as before.
5. `npm run test:php -- --filter WC_Stripe_Agentic_Commerce_Integration_Test` (84) and `--filter WC_Rest_Stripe_Agentic_Commerce_Controller_Test` (57) → green.

New tests: `is_onboarding_complete()` data provider (toggle/secret combos) and `test_sync_feed_skips_when_webhook_secret_missing` (toggle on + feature on + key present, no secret → bail before delivery). #5663's toggle-off test still holds under the tightened gate.

## Follow-up (not in this PR)

The settings UI should disable the "Sync now" button until onboarding is complete and show a label explaining which step is missing, so a merchant isn't left clicking a button that silently no-ops. Tracking separately.

---

### Changelog entry

Added a `Fix` entry to `changelog.txt` and `readme.txt`.


includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -632,6 +597,19 @@ public static function is_merchant_enabled(): bool {
...
 
+	/**
+	 * Whether the merchant has completed Agentic Commerce onboarding: the feature
+	 * is enabled in settings and the webhook signing secret has been saved. Both
+	 * are set by the settings UI; a catalog sync should not push before they are.
+	 *
+	 * @since 10.9.0
+	 * @return bool True once onboarding is complete.
+	 */
+	public static function is_onboarding_complete(): bool {
```
jaclync, line 643: ❓ Does it make sense to apply the same guard to `WC_Stripe_Agentic_Commerce_Inventory_Tracker`'s `sync_inventory` and `sync_archives`? They seem to have different gates right now.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
```diff
@@ -632,6 +597,19 @@ public static function is_merchant_enabled(): bool {
...
+		return self::is_merchant_enabled()
+			&& '' !== (string) get_option( self::WEBHOOK_SECRET_OPTION, '' );
```
jaclync, line 645: non-blocking ❓ on the product side: if a store configures the agentic checkout to redirect to their storefront page, do they still need the webhook setup?

jaclync (commented):
(Updated after reviewing https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5769.)

While testing the PR, I tried emptying the previously filled webhook secret and had some questions about the UI/UX:

- Do we want to disable unsetting the webhook secret (non-empty -> empty)? ~~Or at least a warning about this action? Right now, unsetting it -> saving settings -> tapping "Sync now" results in a generic sync failure `Sync failed. Check the WooCommerce logs for details.`~~
- Maybe the onboarding state can be more clear?
- If we do allow unsetting the webhook secret, would this leave the agentic catalog live but stuck while the secret is unset?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Good catch. Aligned in 4218f305b: `sync_inventory()` and `sync_archives()` now use the same gate as `sync_feed()` (feature flag + `is_onboarding_complete()`). Pending entries are retained while the gate is closed, so the full sync covers them once onboarding completes.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-integration.php
wjrosa: Probably not strictly: in redirect mode there is no in-agent checkout, so no delegated checkout webhooks. I kept the single gate for simplicity, and so that flipping checkout back on later cannot outrun webhook setup. Happy to relax it if product prefers.

wjrosa:
Re the webhook-secret UX questions (https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5768#pullrequestreview-4924125413):

- Agreed we should not allow silently unsetting a saved secret; I will fold a guard (or at least a warning) into the settings-UI follow-up.
- That same follow-up covers clearer onboarding state: disable "Sync now" until onboarding is complete and label the missing step.
- And yes, unsetting the secret leaves the already-uploaded catalog live on Stripe while syncs are skipped; the teardown feed only runs on toggle-off. One more reason to block the unset.

jaclync (approved):
Worked well! The follow-up improvements sound good :shipit: 

> Testing instructions
> 1. Enable the feature flag and connect Stripe (test keys); do not enable Agentic Commerce / save a webhook secret. Trigger a sync (recurring action, or `wp stripe agentic-commerce sync --push`) → no upload; log shows Sync skipped - onboarding incomplete.

Just a non-blocking note that `wp stripe agentic-commerce sync --push` didn't have the same gate. The product feed was still uploaded to Stripe even if the Agentic Commerce toggle was disabled & webhook secret was empty in wp-admin.

wjrosa:
Thanks for the review! I fixed the CLI issue in e9f267e

includes/agentic-commerce/class-wc-stripe-agentic-commerce-inventory-tracker.php
```diff
@@ -554,6 +562,14 @@ public function sync_archives(): void {
 			return;
 		}
 
+		// Same gate as the full catalog sync: no delta may push before the
+		// catalog itself is allowed to. Post-onboarding disables are handled
+		// by the final checkout-disabled feed, not by archive deltas.
+		if ( ! WC_Stripe_Agentic_Commerce_Integration::is_onboarding_complete() ) {
+			WC_Stripe_Logger::info( 'Agentic Commerce: Archive sync skipped - onboarding incomplete' );
+			return;
```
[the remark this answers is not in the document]
wjrosa: Addressed in 50cb5d526. The settings handler now schedules an archive flush once onboarding is complete, so rows queued while the gate was closed no longer wait for the next delete event. Covered by tests on the tracker and the settings endpoint.

The MAX_PENDING_UPDATES behavior is unchanged by this PR: a full queue already defers to the catalog sync, which cannot archive deleted products. That is a pre-existing limit of the archive queue, so I'm leaving it out of this PR.

Issue #1439, opened 2021-01-05
Payment buttons massively impact page speed


Mingendahl:
**Describe the bug**
If you activate the payment buttons (Apple- / Google-Pay), additional 33 (!!) requests to Stripe get fired in order to just create the payment button.
This results in massive degradation of speed metrics (GTMetrix drop from A->B, Pingdom drop A->C, Pagespeed drop 77->39 ) which results in bad user experience and lower ranking in search results.
On our pages, activating the payment buttons results in a 133% increase in total requests per page load.

**To Reproduce**
1. Disable payment buttons
2. go to tools.pingdom.com and probe the page, check number of requests with stripe URLs
3. enable payment buttons
4. redo step 2 --> some 33 additional requests can be counted

**Expected behavior**
Payment button integration uses a reasonable amount of requests (less than 5, best 1) and thus has no negative impact on speed metrics.

**Screenshots**
Without Payment Buttons
Pingdom
![image](https://user-images.githubusercontent.com/27972200/103663171-15c40a00-4f71-11eb-8d86-4d664cbb63a5.png)

Gtmetrix
![image](https://user-images.githubusercontent.com/27972200/103663978-03969b80-4f72-11eb-8f4c-a20ba716c5ee.png)

Google Pagespeed Mobile
![image](https://user-images.githubusercontent.com/27972200/103664183-448eb000-4f72-11eb-8745-901402c48bbb.png)

With Payment Buttons (no other changes)
Pingdom
![image](https://user-images.githubusercontent.com/27972200/103663381-51f76a80-4f71-11eb-973e-d9043f1b01af.png)
![image](https://user-images.githubusercontent.com/27972200/103663447-62a7e080-4f71-11eb-8152-95625b47ea8f.png)

Gtmetrix
![image](https://user-images.githubusercontent.com/27972200/103663637-9c78e700-4f71-11eb-9721-f5566b3feed8.png)

Google Pagespeed Mobile
![image](https://user-images.githubusercontent.com/27972200/103663783-c92cfe80-4f71-11eb-957c-5135fd4ee028.png)

**Environment (please complete the following information):**
 - WordPress Version 5.6
 - WooCommerce Version 4.8.0
 - Stripe Plugin Version 4.7.0
 - Browser [e.g. chrome, safari] and Version all
 - Any other plugins installed (many, but it is a quite local issue and reproducible with and without)

**Additional context**

crazyserb:
[...]

PR #5771, opened 2026-07-30
Serve express checkout AJAX nonces on demand instead of at page render

wjrosa:
Towards STRIPE-956
Towards #1439

`javascript_params()` minted 13 nonces into every page where express checkout can render. Auditing consumers showed:

- **4 keys had no client consumer at all** (`payment`, `checkout`, `pay_for_order`, `log_errors`) — removed.
- **7 keys are only needed after the shopper interacts with a button** (`shipping`, `normalize_address`, `get_cart_details`, `update_shipping`, `add_to_cart`, `get_selected_product_data`, `clear_cart`) — now served on demand by a new `wc_stripe_get_express_checkout_nonces` wc-ajax endpoint. The client fetch is memoized (one request per page at most) and prefetched at the first pointer/touch/focus over the button container so wallet event handlers with tight resolve deadlines don't pay the round trip.
- **2 Store API nonces stay eager** (`wc_store_api`, `wc_store_api_express_checkout`) — deliberately untouched so this composes with #5594's live-nonce work.

Besides doing less work per render, this fixes a cache-staleness class of bugs: full-page caches can no longer pin an expired nonce into the HTML they serve — the bundle is always minted fresh for the session that will use it. Pages cached *before* this change still embed the old nonce keys, and the client falls back to them if the on-demand fetch fails.

**Backward compatibility:** `wc_stripe_express_checkout_params.nonce` loses 11 keys. These localized params are not a documented public surface, but any third-party code reading those keys directly would need to use the new endpoint. All `wc_ajax_*` endpoints and their nonce actions are unchanged. The new endpoint intentionally has no nonce check of its own: nonces are per-session CSRF tokens, not secrets, and cross-origin callers can't read the response; it sends `nocache_headers()`.

## Testing instructions

1. Enable express checkout (Apple Pay / Google Pay) on product, cart, and checkout pages.
2. Load a product page and view source: the localized `wc_stripe_express_checkout_params.nonce` object should contain only `wc_store_api` and `wc_store_api_express_checkout`.
3. Hover over (or tap) the express checkout button area with DevTools open: a single `wc_stripe_get_express_checkout_nonces` request fires.
4. Complete a Google Pay/Apple Pay purchase from the product page, changing the shipping address inside the wallet sheet: shipping options recalculate normally (these calls carry the on-demand nonces).
5. On a variable product, switch variations: the preview updates normally (`get_selected_product_data` path).

---

### Changelog entry

<details>

#### Comment

</details>

**Post merge**


client/api/index.js
```diff
@@ -448,18 +448,48 @@ export default class WCStripeAPI {
...
+	 * @return {Promise<Object>} Promise resolving to the nonce map (empty on failure).
+	 */
+	expressCheckoutFetchNonces() {
+		if ( ! this.expressCheckoutNoncesPromise ) {
+			this.expressCheckoutNoncesPromise = this.request(
+				getExpressCheckoutAjaxURL( 'get_express_checkout_nonces' ),
+				{}
+			)
+				.then( ( response ) => response?.data ?? {} )
+				.catch( () => ( {} ) );
```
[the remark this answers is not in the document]
jaclync: +1 to the retry concern above. How about just memoizing success not failure, so that it can be retried later?

Then the caller can be:
```js
async expressCheckoutGetNonce( key ) {
	const nonces = await this.expressCheckoutFetchNonces().catch( () => ( {} ) );
	return nonces[ key ] ?? getExpressCheckoutData( 'nonce' )?.[ key ];
}
```

```suggestion
				.catch( ( error ) => {
				    this.expressCheckoutNoncesPromise = null;
   				    throw error;
  			    } );
```

client/api/index.js
```diff
@@ -448,18 +448,48 @@ export default class WCStripeAPI {
...
 	}
 
+	/**
+	 * Fetches the express checkout wc-ajax nonce bundle, once per page.
+	 *
+	 * @return {Promise<Object>} Promise resolving to the nonce map (empty on failure).
+	 */
+	expressCheckoutFetchNonces() {
+		if ( ! this.expressCheckoutNoncesPromise ) {
+			this.expressCheckoutNoncesPromise = this.request(
```
jaclync, line 466: non-blocking nit: does it make sense to initialize this promise in the constructor, so that it's clear that this variable is tracked and available at the top level?

client/api/index.js
```diff
@@ -448,18 +448,48 @@ export default class WCStripeAPI {
...
+	 * Resolves a single express checkout wc-ajax nonce, falling back to the
+	 * localized params (pages cached before the bundle became lazy still
+	 * embed them) when the fetch fails.
+	 *
+	 * @param {string} key Nonce key.
+	 * @return {Promise<string|undefined>} Promise resolving to the nonce value.
+	 */
+	async expressCheckoutGetNonce( key ) {
+		const nonces = await this.expressCheckoutFetchNonces();
+		return nonces[ key ] ?? getExpressCheckoutData( 'nonce' )?.[ key ];
```
jaclync: non-blocking nit: as this PR removes the EC nonce entries from `wc_stripe_express_checkout_params`, `getExpressCheckoutData( 'nonce' )?.[ key ]` is not defined all the time. Maybe this default can be removed so that it doesn't read like it can still be there?

client/entrypoints/express-checkout/index.js
```diff
@@ -1063,6 +1063,21 @@ jQuery( function ( $ ) {
 		wcStripeECE.init();
 	}
 
+	// Warm the on-demand nonce bundle at the first sign of intent so wallet
+	// event handlers (tight resolve deadlines) don't pay the round trip.
+	const eceContainer = document.getElementById(
+		'wc-stripe-express-checkout-element'
+	);
+	if ( eceContainer ) {
+		[ 'pointerenter', 'touchstart', 'focusin' ].forEach( ( eventName ) =>
```
jaclync, line 1072: ❓ Would this warmup work for blocks checkout / mobile web? From testing blocks checkout (cart) and mobile web, the `wc-ajax=wc_stripe_get_express_checkout_nonces` request seemed to be made only after tapping the ECE button. The hovering worked in classic checkout.

jaclync (commented):
Express checkout payment testing is currently blocked by the account issue that https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5800 likely fixes. I'll do more testing once the fix is included in this branch.

client/api/index.js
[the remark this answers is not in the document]
wjrosa: Applied: your catch suggestion landed in 70872ce00, and ebcbf4af8 finishes the caller side. Only success is memoized now; a failure rejects and the next lookup retries. Jest covers the fail-then-retry path (endpoint requested twice).

client/api/index.js
wjrosa: Done in ebcbf4af8: initialized to null in the constructor.

client/api/index.js
wjrosa: Agreed, removed in ebcbf4af8. The localized params now only carry the Store API keys, so the fallback was misleading; a failed fetch rejects like any other wc-ajax request and retries on the next lookup. Updated the Jest case accordingly.

client/entrypoints/express-checkout/index.js
wjrosa: Expected on both counts. The warm-up lives only in the classic bundle, where the deadline-bound handlers (shipping address/option change) hit these wc-ajax endpoints; on Blocks the only consumer of these nonces runs at confirm time, so fetching on tap is already early enough. And on mobile there is no pre-tap intent signal: touchstart fires with the tap itself, which still starts the fetch before the wallet sheet opens and the deadline-bound events fire.

jaclync (approved):
Thanks for the updates, working as expected and LGTM 🚢

PR #5776, opened 2026-07-31
Implement a simple admin rest endpoint to retrieve a Stripe payment intent details

en-adobrescu:

Implements a simple  admin REST endpoint for retrieveing a Stripe payement intent details ("get one").

## Testing instructions

1. Make sure you can send HTTP requests to the backend using a HTTP client. For example, you can do this by setting define( 'WP_ENVIRONMENT_TYPE', 'local' ); in wp-config.php and adding an application password from 'admin' user profile.

2. Using a HTTP client (for example Postman or a browser), 'admin' and the new created password as Basic Auth , send a GET request to:
[WordPress instance URL]/wp-json/wc/v3/wc_stripe/payment_intents/{id}
```
{
    "object": "payment_intent",
    "id": "pi_3TtPY4JlUF0dQbSB0Jb3hqkX",
    "amount": 7900,
    "amount_received": 7900,
    "currency": "usd",
    "status": "succeeded",
    "description": "WooCommerce Stripe Dev - Order 37",
    "latest_charge": {
        "balance_transaction": "txn_3TtPY4JlUF0dQbSB0EdqQ43J",
        "billing_details": {
            "address": {
                "city": "Bucuresti Sector 3",
                "country": "RO",
                "line1": "Bd. Octavian Goga nr. 4",
                "line2": "",
                "postal_code": "030101",
                "state": "B"
            },
            "email": "admin@example.com",
            "name": "John Smith",
            "phone": "+40722112945",
            "tax_id": null
        },
        "payment_method_details": {
            "type": "card",
            "card": {
                "last4": "0259",
                "exp_month": 12,
                "exp_year": 2030,
                "checks": {
                    "cvc_check": "pass"
                },
                "brand": "visa",
                "funding": "credit",
                "country": "US"
            }
        }
    }
}
```
---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
+++ b/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
@@ -6,9 +6,7 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * REST controller exposing Stripe payment intents data to the admin UI.
- *
- * The controller acts as a proxy that forwards the received parameters to the remote Stripe API and returns the received response.
+ * REST controller exposing a Stripe payment intent detais to the admin UI.
  *
  * @since 10.9.0
  */
@@ -19,62 +17,33 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 *
 	 * @var string
 	 */
-	protected $rest_base = 'wc_stripe/payment_intents';
-
-	/**
-	 * Endpoint args.
-	 *
-	 * @var array
-	 */
-	protected $rest_args = [
-		'limit'            => [
-			'type'              => 'integer',
-			'default'           => 25,
-			'minimum'           => 1,
-			'maximum'           => 100,
-			'sanitize_callback' => 'absint',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'starting_after'   => [
-			'type'              => 'string',
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'ending_before'    => [
-			'type'              => 'string',
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'customer'         => [
-			'type'              => 'string',
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'customer_account' => [
-			'type'              => 'string',
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'created'          => [
-			'required'          => false,
-			'sanitize_callback' => [ self::class, 'sanitize_created_field' ],
-			'validate_callback' => [ self::class, 'validate_created_field' ],
-		],
-	];
+	protected $rest_base = 'payment_intents';
 
 	protected array $stripe_response_allowed_fields = [
-		'object'                               => '',
-		'has_more'                             => '',
-		'data.id'                              => '',
-		'data.amount'                          => [ WC_Stripe_REST_Response_Filter::class, 'money_format' ],
-		'data.amount_received'                 => [ WC_Stripe_REST_Response_Filter::class, 'money_format' ],
-		'data.currency'                        => 'strtoupper',
-		'data.payment_details.order_reference' => '',
-		'data.status'                          => '',
+		'object',
+		'id',
+		'amount',
+		'amount_received',
+		'currency',
+		'payment_details.order_reference',
+		'status',
+		'description',
+		'latest_charge.balance_transaction.fee',
+		'latest_charge.balance_transaction.net',
+		'latest_charge.balance_transaction.currency',
+		'latest_charge.billing_details',
+		'latest_charge.payment_method_details.type',
+		'latest_charge.payment_method_details.card.last4',
+		'latest_charge.payment_method_details.card.exp_month',
+		'latest_charge.payment_method_details.card.exp_year',
+		'latest_charge.payment_method_details.card.checks.cvc_check',
+		'latest_charge.payment_method_details.card.brand',
+		'latest_charge.payment_method_details.card.funding',
+		'latest_charge.payment_method_details.card.country',
+	];
+
+	protected array $stripe_expand_fields = [
+		'latest_charge',
 	];
 
 	/**
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 */
 	public function register_routes() {
 		register_rest_route(
-			$this->namespace,
-			'/' . $this->rest_base,
+			$this->namespace . '/wc_stripe',
+			'/' . $this->rest_base . '(?:/(?P<id>.+))',
 			[
 				'methods'             => WP_REST_Server::READABLE,
-				'callback'            => [ $this, 'get_payment_intents' ],
+				'callback'            => [ $this, 'get_payment_intent' ],
 				'permission_callback' => [ $this, 'check_permission' ],
-				'args'                => $this->get_payment_intents_route_args(),
+				'args'                => [],
 			]
 		);
 	}
 
 	/**
-	 * Return route args.
-	 *
-	 * @return array
-	 */
-	public function get_payment_intents_route_args(): array {
-		return $this->rest_args;
-	}
-
-	/**
-	 * Given an incoming REST request, build and return an array of query parameters to be appended to Stripe API request URL.
-	 *
-	 * @param WP_REST_Request<array<string, mixed>> $request An incoming REST request.
-	 *
-	 * @return array
-	 */
-	public function build_http_query_array_from_request( $request ): array {
-		/**
-		 * Route args.
-		 *
-		 * @var array<string, mixed> $rest_args
-		 */
-		$rest_args = $this->get_payment_intents_route_args();
-
-		$search_params = [];
-
-		foreach ( $rest_args as $search_param_name => $search_param_definition ) {
-			$search_param_value = $request->get_param( $search_param_name );
-
-			if ( '' === $search_param_value || is_null( $search_param_value ) ) {
-				continue;
-			}
-
-			$search_params[ $search_param_name ] = $search_param_value;
-		}
-
-		return $search_params;
-	}
-
-	/**
-	 * Given an incoming REST request, build and return a query parameters string to be appended to Stripe API request URL.
-	 *
-	 * @param WP_REST_Request<array<string, mixed>> $request An incoming REST request.
-	 *
-	 * @return string
-	 */
-	public function build_http_query_string_from_request( $request ): string {
-		return http_build_query( $this->build_http_query_array_from_request( $request ) );
-	}
-
-	/**
-	 * Retrieve a paginated list of Stripe payment intents.
+	 * Retrieve, filters and return one Stripe payment intent.
 	 *
 	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 *
 	 * @return WP_REST_Response|WP_Error
 	 */
-	public function get_payment_intents( $request ) {
-		$response = WC_Stripe_API::retrieve( 'payment_intents?' . $this->build_http_query_string_from_request( $request ) );
+	public function get_payment_intent( $request ) {
+		$response = $this->fetch_from_stripe( 'payment_intents/' . $request['id'], [ 'expand' => $this->stripe_expand_fields ] );
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$filtered_response = WC_Stripe_REST_Response_Filter::filter_response( $response, $this->stripe_response_allowed_fields );
+
+		return rest_ensure_response( $filtered_response );
+	}
+
+	/**
+	 * Fetch data from an Stripe API endpoint and returns its raw data or a WP_Error if an error occurs.
+	 *
+	 * @param string $endpoint The Stripe endpoint.
+	 * @param array $params Parameters to pass to the endpoint.
+	 *
+	 * @return StdClass|WP_Error
+	 */
+	protected function fetch_from_stripe( $endpoint, $params ) {
+		$query_string = http_build_query( $params );
+
+		$response = WC_Stripe_API::retrieve( $endpoint . ( $query_string ? '?' . $query_string : '' ) );
 
 		if ( null === $response ) {
 			return new WP_Error(
-				'wc_stripe_payment_intents_error',
-				__( 'Unable to retrieve payment intents from Stripe.', 'woocommerce-gateway-stripe' ),
+				'wc_stripe_error',
+				__( 'Unable to fetch data from Stripe.', 'woocommerce-gateway-stripe' ),
 				[ 'status' => 401 ]
 			);
 		}
@@ -173,72 +115,6 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 			return new WP_Error( $error_code, $error_message, [ 'status' => 400 ] );
 		}
 
-		$filtered_response = WC_Stripe_REST_Response_Filter::filter_response( $response, $this->stripe_response_allowed_fields );
-
-		return rest_ensure_response( $filtered_response );
-	}
-
-	/**
-	 * Sanitize a "created" parameter value.
-	 *
-	 * @param string $value The parameter value.
-	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
-	 * @param string $param The parameter name.
-	 *
-	 * @return mixed
-	 */
-	public static function sanitize_created_field( $value, WP_REST_Request $request, string $param ) {
-		if ( ! is_array( $value ) ) {
-			$value = sanitize_text_field( $value );
-		} else {
-			$sanitized_value = [];
-
-			foreach ( $value as $operator => $operand ) {
-				$sanitized_value[ sanitize_key( $operator ) ] = sanitize_text_field( $operand );
-			}
-
-			$value = $sanitized_value;
-		}
-
-		return $value;
-	}
-
-	/**
-	 * Validate a "created" parameter value.
-	 *
-	 * @param string $value The parameter value.
-	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
-	 * @param string $param The parameter name.
-	 *
-	 * @return bool
-	 */
-	public static function validate_created_field( $value, WP_REST_Request $request, string $param ): bool {
-		if ( empty( $value ) ) {
-			return true;
-		}
-
-		$unix_timestamp_pattern = '^\d+$';
-
-		if ( is_string( $value ) ) {
-			return preg_match( '/' . $unix_timestamp_pattern . '/', $value ) === 1;
-		}
-
-		if ( ! is_array( $value ) ) {
-			return false;
-		}
-
-		$allowed_operators = [ 'gt', 'gte', 'lt', 'lte' ];
-
-		foreach ( $value as $operator => $operand ) {
-			if ( ! in_array( $operator, $allowed_operators ) ) {
-				return false;
-			}
-
-			if ( ! is_scalar( $operand ) || preg_match( '/' . $unix_timestamp_pattern . '/', (string) $operand ) !== 1 ) {
-				return false;
-			}
-		}
-
-		return true;
+		return $response;
 	}
 }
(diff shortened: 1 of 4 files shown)


includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -6,9 +6,7 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * REST controller exposing Stripe payment intents data to the admin UI.
- *
- * The controller acts as a proxy that forwards the received parameters to the remote Stripe API and returns the received response.
+ * REST controller exposing a Stripe payment intent detais to the admin UI.
```
daledupreez: Nit: Typos.
```suggestion
 * REST controller exposing Stripe payment intent details to the admin UI.
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -19,62 +17,33 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
+		'latest_charge.balance_transaction.currency',
+		'latest_charge.billing_details',
+		'latest_charge.payment_method_details.type',
+		'latest_charge.payment_method_details.card.last4',
+		'latest_charge.payment_method_details.card.exp_month',
+		'latest_charge.payment_method_details.card.exp_year',
+		'latest_charge.payment_method_details.card.checks.cvc_check',
+		'latest_charge.payment_method_details.card.brand',
+		'latest_charge.payment_method_details.card.funding',
+		'latest_charge.payment_method_details.card.country',
```
daledupreez: These types are pretty narrow and will exclude all details for non-card transactions. Can we allow all of `latest_charg.payment_method_details` through without requiring explicit line items for every payment method?

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -19,62 +17,33 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'created'          => [
-			'required'          => false,
-			'sanitize_callback' => [ self::class, 'sanitize_created_field' ],
-			'validate_callback' => [ self::class, 'validate_created_field' ],
-		],
+	protected array $stripe_response_allowed_fields = [
```
daledupreez: Out of interest, any reason for making this a property and not a `protected const`?

Also, I think we can pre-emptively name the property (or const) in a way that indicates it is for the individual get API. But it's possible that the same fields will apply for the list API.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -19,62 +17,33 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
-	protected array $stripe_response_allowed_fields = [
-		'object'                               => '',
-		'has_more'                             => '',
-		'data.id'                              => '',
-		'data.amount'                          => [ WC_Stripe_REST_Response_Filter::class, 'money_format' ],
-		'data.amount_received'                 => [ WC_Stripe_REST_Response_Filter::class, 'money_format' ],
-		'data.currency'                        => 'strtoupper',
-		'data.payment_details.order_reference' => '',
-		'data.status'                          => '',
+	protected array $stripe_expand_fields = [
```
daledupreez: Nit: I think we should be more explicit about what this represents via the name. Is this the list of permitted expand fields? Or the list of fields we should always request should be expanded?

Also, this could be a `const` rather than a variable.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 */
 	public function register_routes() {
 		register_rest_route(
-			$this->namespace,
-			'/' . $this->rest_base,
+			$this->namespace . '/wc_stripe',
```
daledupreez: Rather than appending this here, might it be worth initialising `$this->rest_base` to include `/wc_stripe`? That way all paths are guaranteed to have the same base path. (The existing code is using that approach.)

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
+			'/' . $this->rest_base . '(?:/(?P<id>.+))',
```
daledupreez: I have three thoughts here:
1. We can drop the look-ahead piece of the regular expression.
2. We should probably harden the expression to require word and _ characters only (see my comments about query encoding later)
3. As we look forward, is it worth tightening this regular expression now to only allow `pi_`-prefixed values? That way we won't have issues if we add a `/search` route later.

```suggestion
			'/' . $this->rest_base . '/(?P<id>pi_.+(\w|_)+\w+)',
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
 				'permission_callback' => [ $this, 'check_permission' ],
-				'args'                => $this->get_payment_intents_route_args(),
+				'args'                => [],
 			]
 		);
 	}
 
 	/**
-	 * Return route args.
+	 * Retrieve, filters and return one Stripe payment intent.
```
daledupreez, line 61: Nit: typo.
```suggestion
	 * Retrieve, filter, and return one Stripe payment intent.
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
 	/**
-	 * Given an incoming REST request, build and return a query parameters string to be appended to Stripe API request URL.
+	 * Fetch data from an Stripe API endpoint and returns its raw data or a WP_Error if an error occurs.
 	 *
-	 * @param WP_REST_Request<array<string, mixed>> $request An incoming REST request.
+	 * @param string $endpoint The Stripe endpoint.
+	 * @param array $params Parameters to pass to the endpoint.
 	 *
-	 * @return string
+	 * @return StdClass|WP_Error
```
daledupreez, line 85: Nit: this should be `stdClass`
```suggestion
	 * @return stdClass|WP_Error
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
 
-		return $search_params;
+		$filtered_response = WC_Stripe_REST_Response_Filter::filter_response( $response, $this->stripe_response_allowed_fields );
+
+		return rest_ensure_response( $filtered_response );
 	}
 
 	/**
-	 * Given an incoming REST request, build and return a query parameters string to be appended to Stripe API request URL.
+	 * Fetch data from an Stripe API endpoint and returns its raw data or a WP_Error if an error occurs.
```
daledupreez, line 80: Nit: typo.
```suggestion
	 * Fetch data from an Stripe API endpoint and return its raw data or a WP_Error if an error occurs.
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
+	 * @param string $endpoint The Stripe endpoint.
+	 * @param array $params Parameters to pass to the endpoint.
 	 *
-	 * @return string
+	 * @return StdClass|WP_Error
 	 */
-	public function build_http_query_string_from_request( $request ): string {
-		return http_build_query( $this->build_http_query_array_from_request( $request ) );
-	}
+	protected function fetch_from_stripe( $endpoint, $params ) {
```
daledupreez, line 87: Nit: can we add explicit types for the arguments here?
```suggestion
	protected function fetch_from_stripe( string $endpoint, array $params = [] ) {
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
+		$query_string = http_build_query( $params );
```
daledupreez: Not blocking, but is it worth explicitly specifying `PHP_QUERY_RFC3986` encoding given that this is going to be in a URL? This is only material if we allow space values in the REST API.
```suggestion
		$query_string = http_build_query( $params, '', '&', PHP_QUERY_RFC3986 );
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
 
-	/**
-	 * Retrieve a paginated list of Stripe payment intents.
-	 *
-	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
-	 * @return WP_REST_Response|WP_Error
-	 */
-	public function get_payment_intents( $request ) {
-		$response = WC_Stripe_API::retrieve( 'payment_intents?' . $this->build_http_query_string_from_request( $request ) );
+		$response = WC_Stripe_API::retrieve( $endpoint . ( $query_string ? '?' . $query_string : '' ) );
```
daledupreez: Nit: To keep things simple, it may be worth using a local variable to build the partial URL in one step and then calling the downstream function after that has been built.
```suggestion
		$stripe_resource_url = $endpoint . ( '' === $query_string ? '' : '?' . $query_string );
		$response = WC_Stripe_API::retrieve( $stripe_resource_url );
```

includes/admin/class-wc-stripe-rest-response-filter.php
```diff
@@ -42,7 +42,7 @@ protected static function expand_allowed_property_paths( array $allowed_property
 				$ref = &$ref[ static::IDX_PATH ][ $property_name ];
 			}
 
-			$ref[ static::IDX_FORMAT_CALLBACK ] = $format_callback;
+			$ref[ static::IDX_FORMAT_CALLBACK ] = '';
```
daledupreez: Given that we've removed the logic for the formatting, I don't know that we need this. If we want to process the data, then I think we should add distinct fields.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -6,7 +6,7 @@
  * @preserveGlobalState disabled
  */
 class WC_Stripe_REST_Payment_Intents_Controller_Test extends WP_UnitTestCase {
-	const ENDPOINT_URL = '/wc/v3/wc_stripe/payment_intents';
+	const ENDPOINT_URL = '/wc/v3/wc_stripe/payment_intents/pi_3TtPY4JlUF0dQbSB0Jb3hqkX';
```
daledupreez: Two nits:
1. We should use `private` or `protected` visibility.
2. We should name the value to be more specific.
```suggestion
	private const SINGLE_INTENT_ENDPOINT_URL = '/wc/v3/wc_stripe/payment_intents/pi_test_9876543210';
```

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -6,7 +6,7 @@
  * @preserveGlobalState disabled
  */
 class WC_Stripe_REST_Payment_Intents_Controller_Test extends WP_UnitTestCase {
```
daledupreez, line 8: The tests in this class are not validating that the data returned from Stripe is correctly returned, nor are there any tests confirming that the fields not on the allowlist are being returned. That should be fairly easy to add by allowing the response body to be specified by `mock_http_call()`. But it may be easier to have the `pre_http_request` filtering and cleanup handled locally within each test for those cases.

daledupreez (commented):
I have feedback based on inspection.

en-adobrescu:
> I have feedback based on inspection.

Thanks  Dale, looking at it now

daledupreez:
@en-adobrescu, it also looks like the response filter unit tests are failing.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
wjrosa: The PR description says this adds a "get one" endpoint, but it replaces the list endpoint rather than adding alongside it. The route group here isn't optional either (no trailing `?`), so `GET /wc_stripe/payment_intents` now errors. That endpoint was just added in #5531. Is the removal intentional? If so, could we mention that in the description?

On the same topic, we usually never simply remove public methods to avoid breaking stores and 3rd-party code. We always opt for deprecating them

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
-		 * @var array<string, mixed> $rest_args
-		 */
-		$rest_args = $this->get_payment_intents_route_args();
-
-		$search_params = [];
-
-		foreach ( $rest_args as $search_param_name => $search_param_definition ) {
-			$search_param_value = $request->get_param( $search_param_name );
+	public function get_payment_intent( $request ) {
+		$response = $this->fetch_from_stripe( 'payment_intents/' . $request['id'], [ 'expand' => $this->stripe_expand_fields ] );
```
wjrosa: `id` reaches the Stripe URL unvalidated (`'args' => []`) and unencoded, and `.+` matches `/` — so `pi_x/../../customers` resolves to another Stripe resource before the request goes out. Note the regex suggested above doesn't close this: `pi_.+(\w|_)+\w+` still matches slashes. `WC_Stripe_Ability_Get_Payment_Intent` on this branch has the pattern we want: `^pi_[A-Za-z0-9_]+$` plus `rawurlencode()`.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -19,62 +17,33 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
-		'object'                               => '',
-		'has_more'                             => '',
-		'data.id'                              => '',
-		'data.amount'                          => [ WC_Stripe_REST_Response_Filter::class, 'money_format' ],
-		'data.amount_received'                 => [ WC_Stripe_REST_Response_Filter::class, 'money_format' ],
-		'data.currency'                        => 'strtoupper',
-		'data.payment_details.order_reference' => '',
-		'data.status'                          => '',
+	protected array $stripe_expand_fields = [
+		'latest_charge',
```
wjrosa, line 38: We only expand `latest_charge`, so `balance_transaction` comes back as a bare string ID and the allowlisted `.fee`/`.net`/`.currency` can never be present — the sample response in the description shows it. Needs `latest_charge.balance_transaction` expanded too.

Worth noting `filter_value()` returns the raw value when a nested rule meets a scalar, so the allowlist fails open rather than dropping the field.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
daledupreez: On this topic, note that the target is a feature branch, so we can iterate a bit more safely in the short term.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: fixed

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: yes, done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: changed the prop to  const STRIPE_RESPONSE_ALLOWED_FIELDS. I think that the field whitelist can be shared by all 3 payment_intents endpoints (get-one/list/search)

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: changed the property to const STRIPE_EXPAND_PARAM.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: done

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: done

en-adobrescu:
> @en-adobrescu, it also looks like the response filter unit tests are failing.

fixed

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Added a simple test, given a test body it tests that only the allowed part is returned. There are more filtering tests in WC_Stripe_REST_Response_Filter_Test

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Hi Wesley! 

Added 'latest_charge.balance_transaction' to STRIPE_EXPAND_PARAM (renamed from $stripe_expand_fields).

We can add/remove entries to/from STRIPE_EXPAND_PARAM and STRIPE_RESPONSE_ALLOWED_FIELDS base on UI needs. 

The $value received by filter_value is already whitelisted, but if it is an object then filter_value  checks its properties.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
wjrosa: @en-adobrescu I think you missed this one on your latest commits

en-adobrescu: Added a more restrictive regex for matching payment intent IDs

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -85,79 +47,54 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	public function register_routes() {
 		register_rest_route(
 			$this->namespace,
-			'/' . $this->rest_base,
+			'/' . $this->rest_base . '(?:/(?P<id>pi_[A-Za-z0-9_]+))?$',
```
wjrosa: The `?` makes the id optional, so `GET /wc/v3/wc_stripe/payment_intents` (no id) now matches with `$request['id']` as `NULL`. That builds `payment_intents/` and hits Stripe's **list** endpoint, returning 200:

```
https://api.stripe.com/v1/payment_intents/?expand[0]=latest_charge&expand[1]=latest_charge.balance_transaction
```

Nothing leaks today — `data` isn't on the allowlist, so the client just gets `{"object":"list"}` — but the store issues an unbounded list call with none of the pagination validation the old list endpoint had. Also `rawurlencode( null )` is a deprecation on PHP 8.1+, and `/payment_intents/` (trailing slash) still 404s while the bare path returns 200.

Verified against a local install. Dropping the `?` so the id is required should cover it.

en-adobrescu: removed the '?'

wjrosa: Awesome! Thanks

wjrosa (approved):

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -85,79 +47,54 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	public function register_routes() {
 		register_rest_route(
 			$this->namespace,
-			'/' . $this->rest_base,
+			'/' . $this->rest_base . '(?:/(?P<id>pi_[A-Za-z0-9_]+))$',
```
daledupreez, line 50: Nit: Can we remove the lookahead here?
```suggestion
			'/' . $this->rest_base . '/(?P<id>pi_[A-Za-z0-9_]+)$',
```

daledupreez:
There are still some items we can tackle in follow-up PRs, but getting this merged for now.

PR #5786, opened 2026-08-05
Add admin Stripe payment intents list rest api endpoint

en-adobrescu:

Extends the admin Stripe REST Controller implemented by https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5776 PR by adding a new route  for querying payment intents endpoint of Stripe API. The controller acts as a proxy, forwarding the received parameters to the remote endpoint. The supported parameters are documented here: https://docs.stripe.com/api/payment_intents/list?api-version=2026-05-27.preview&rds=1

The response received from Stripe API is returned after being filtered against a whitelist.

## Testing instructions
1. Make sure you can send HTTP requests to the backend using a HTTP client. For example, you can do this by setting define( 'WP_ENVIRONMENT_TYPE', 'local' ); in wp-config.php and adding an application password from 'admin' user profile.

2. Using a HTTP client (for example Postman or a browser), 'admin' and the new created password as Basic Auth , send a GET request to:
[WordPress instance URL]/wp-json/wc/v3/wc_stripe/payment_intents
```
{
    "object": "list",
    "has_more": true,
    "data": [
        {
            "id": "pi_3TtPY4JlUF0dQbSB0Jb3hqkX",
            "created": 1784109376,
            "amount": 7900,
            "currency": "usd",
            "status": "succeeded",
            "description": "WooCommerce Stripe Dev - Order 37",
            "latest_charge": {
                "billing_details": {
                    "name": "John Smith"
                },
                "payment_method_details": {
                    "card": {
                        "amount_authorized": 7900,
                        "authorization_code": "747254",
                        "brand": "visa",
                        "checks": {
                            "address_line1_check": "pass",
                            "address_postal_code_check": "pass",
                            "cvc_check": "pass"
                        },
                        "country": "US",
                        "exp_month": 12,
                        "exp_year": 2030,
                        "extended_authorization": {
                            "status": "disabled"
                        },
                        "fingerprint": "lt4nMrHlZlkzWWm1",
                        "funding": "credit",
                        "incremental_authorization": {
                            "status": "unavailable"
                        },
                        "installments": null,
                        "last4": "0259",
                        "mandate": null,
                        "multicapture": {
                            "status": "unavailable"
                        },
                        "network": "visa",
                        "network_token": {
                            "used": false
                        },
                        "network_transaction_id": "104738610711282",
                        "overcapture": {
                            "maximum_amount_capturable": 7900,
                            "status": "unavailable"
                        },
                        "regulated_status": "unregulated",
                        "three_d_secure": null,
                        "transaction_link_id": null,
                        "wallet": null
                    },
                    "type": "card"
                }
            }
        }
    ]
}
```
3. The endpoint supports the following parameters:

- limit
- starting_after
- ending_before
- customer
- customer_account
- created

They are documented here: 
https://docs.stripe.com/api/payment_intents/list?api-version=2026-05-27.preview&rds=1

---

### Changelog entry

<details>

#### Message
Add a Stripe REST proxy endpoint for querying payment intents endpoint of Stripe API.

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
+++ b/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
@@ -51,7 +51,6 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 		'object',
 		'has_more',
 		'data.id',
-		'data.payment_intent',
 		'data.created',
 		'data.amount',
 		'data.currency',
--- a/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
+++ b/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
@@ -47,7 +47,7 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	public function register_routes() {
 		register_rest_route(
 			$this->namespace,
-			'/' . $this->rest_base . '(?:/(?P<id>pi_[A-Za-z0-9_]+))$',
+			'/' . $this->rest_base . '/(?P<id>pi_[A-Za-z0-9_]+)$',
 			[
 				'methods'             => WP_REST_Server::READABLE,
 				'callback'            => [ $this, 'get_payment_intent' ],
--- a/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
+++ b/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
@@ -88,7 +88,7 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 					'limit'            => [
 						'type'              => 'integer',
 						'required'          => false,
-						'default'           => 25,
+						'default'           => 10,
 						'minimum'           => 1,
 						'maximum'           => 100,
 						'sanitize_callback' => 'absint',
--- a/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
+++ b/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
@@ -6,9 +6,7 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
- * REST controller exposing Stripe payment intents data to the admin UI.
- *
- * The controller acts as a proxy that forwards the received parameters to the remote Stripe API and returns the received response.
+ * REST controller exposing a Stripe payment intent detais to the admin UI.
  *
  * @since 10.9.0
  */
@@ -19,62 +17,33 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 *
 	 * @var string
 	 */
-	protected $rest_base = 'wc_stripe/payment_intents';
-
-	/**
-	 * Endpoint args.
-	 *
-	 * @var array
-	 */
-	protected $rest_args = [
-		'limit'            => [
-			'type'              => 'integer',
-			'default'           => 25,
-			'minimum'           => 1,
-			'maximum'           => 100,
-			'sanitize_callback' => 'absint',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'starting_after'   => [
-			'type'              => 'string',
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'ending_before'    => [
-			'type'              => 'string',
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'customer'         => [
-			'type'              => 'string',
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'customer_account' => [
-			'type'              => 'string',
-			'required'          => false,
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-		],
-		'created'          => [
-			'required'          => false,
-			'sanitize_callback' => [ self::class, 'sanitize_created_field' ],
-			'validate_callback' => [ self::class, 'validate_created_field' ],
-		],
-	];
+	protected $rest_base = 'payment_intents';
 
 	protected array $stripe_response_allowed_fields = [
-		'object'                               => '',
-		'has_more'                             => '',
-		'data.id'                              => '',
-		'data.amount'                          => [ WC_Stripe_REST_Response_Filter::class, 'money_format' ],
-		'data.amount_received'                 => [ WC_Stripe_REST_Response_Filter::class, 'money_format' ],
-		'data.currency'                        => 'strtoupper',
-		'data.payment_details.order_reference' => '',
-		'data.status'                          => '',
+		'object',
+		'id',
+		'amount',
+		'amount_received',
+		'currency',
+		'payment_details.order_reference',
+		'status',
+		'description',
+		'latest_charge.balance_transaction.fee',
+		'latest_charge.balance_transaction.net',
+		'latest_charge.balance_transaction.currency',
+		'latest_charge.billing_details',
+		'latest_charge.payment_method_details.type',
+		'latest_charge.payment_method_details.card.last4',
+		'latest_charge.payment_method_details.card.exp_month',
+		'latest_charge.payment_method_details.card.exp_year',
+		'latest_charge.payment_method_details.card.checks.cvc_check',
+		'latest_charge.payment_method_details.card.brand',
+		'latest_charge.payment_method_details.card.funding',
+		'latest_charge.payment_method_details.card.country',
+	];
+
+	protected array $stripe_expand_fields = [
+		'latest_charge',
 	];
 
 	/**
@@ -84,80 +53,53 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 */
 	public function register_routes() {
 		register_rest_route(
-			$this->namespace,
-			'/' . $this->rest_base,
+			$this->namespace . '/wc_stripe',
+			'/' . $this->rest_base . '(?:/(?P<id>.+))',
 			[
 				'methods'             => WP_REST_Server::READABLE,
-				'callback'            => [ $this, 'get_payment_intents' ],
+				'callback'            => [ $this, 'get_payment_intent' ],
 				'permission_callback' => [ $this, 'check_permission' ],
-				'args'                => $this->get_payment_intents_route_args(),
+				'args'                => [],
 			]
 		);
 	}
 
 	/**
-	 * Return route args.
-	 *
-	 * @return array
-	 */
-	public function get_payment_intents_route_args(): array {
-		return $this->rest_args;
-	}
-
-	/**
-	 * Given an incoming REST request, build and return an array of query parameters to be appended to Stripe API request URL.
-	 *
-	 * @param WP_REST_Request<array<string, mixed>> $request An incoming REST request.
-	 *
-	 * @return array
-	 */
-	public function build_http_query_array_from_request( $request ): array {
-		/**
-		 * Route args.
-		 *
-		 * @var array<string, mixed> $rest_args
-		 */
-		$rest_args = $this->get_payment_intents_route_args();
-
-		$search_params = [];
-
-		foreach ( $rest_args as $search_param_name => $search_param_definition ) {
-			$search_param_value = $request->get_param( $search_param_name );
-
-			if ( '' === $search_param_value || is_null( $search_param_value ) ) {
-				continue;
-			}
-
-			$search_params[ $search_param_name ] = $search_param_value;
-		}
-
-		return $search_params;
-	}
-
-	/**
-	 * Given an incoming REST request, build and return a query parameters string to be appended to Stripe API request URL.
-	 *
-	 * @param WP_REST_Request<array<string, mixed>> $request An incoming REST request.
-	 *
-	 * @return string
-	 */
-	public function build_http_query_string_from_request( $request ): string {
-		return http_build_query( $this->build_http_query_array_from_request( $request ) );
-	}
-
-	/**
-	 * Retrieve a paginated list of Stripe payment intents.
+	 * Retrieve, filters and return one Stripe payment intent.
 	 *
 	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 *
 	 * @return WP_REST_Response|WP_Error
 	 */
-	public function get_payment_intents( $request ) {
-		$response = WC_Stripe_API::retrieve( 'payment_intents?' . $this->build_http_query_string_from_request( $request ) );
+	public function get_payment_intent( $request ) {
+		$response = $this->fetch_from_stripe( 'payment_intents/' . $request['id'], [ 'expand' => $this->stripe_expand_fields ] );
+
+		if ( is_wp_error( $response ) ) {
+			return $response;
+		}
+
+		$filtered_response = WC_Stripe_REST_Response_Filter::filter_response( $response, $this->stripe_response_allowed_fields );
+
+		return rest_ensure_response( $filtered_response );
+	}
+
+	/**
+	 * Fetch data from an Stripe API endpoint and returns its raw data or a WP_Error if an error occurs.
+	 *
+	 * @param string $endpoint The Stripe endpoint.
+	 * @param array $params Parameters to pass to the endpoint.
+	 *
+	 * @return StdClass|WP_Error
+	 */
+	protected function fetch_from_stripe( $endpoint, $params ) {
+		$query_string = http_build_query( $params );
+
+		$response = WC_Stripe_API::retrieve( $endpoint . ( $query_string ? '?' . $query_string : '' ) );
 
 		if ( null === $response ) {
 			return new WP_Error(
-				'wc_stripe_payment_intents_error',
-				__( 'Unable to retrieve payment intents from Stripe.', 'woocommerce-gateway-stripe' ),
+				'wc_stripe_error',
+				__( 'Unable to fetch data from Stripe.', 'woocommerce-gateway-stripe' ),
 				[ 'status' => 401 ]
 			);
 		}
@@ -173,72 +115,6 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 			return new WP_Error( $error_code, $error_message, [ 'status' => 400 ] );
 		}
 
-		$filtered_response = WC_Stripe_REST_Response_Filter::filter_response( $response, $this->stripe_response_allowed_fields );
-
-		return rest_ensure_response( $filtered_response );
-	}
-
-	/**
-	 * Sanitize a "created" parameter value.
-	 *
-	 * @param string $value The parameter value.
-	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
-	 * @param string $param The parameter name.
-	 *
-	 * @return mixed
-	 */
-	public static function sanitize_created_field( $value, WP_REST_Request $request, string $param ) {
-		if ( ! is_array( $value ) ) {
-			$value = sanitize_text_field( $value );
-		} else {
-			$sanitized_value = [];
-
-			foreach ( $value as $operator => $operand ) {
-				$sanitized_value[ sanitize_key( $operator ) ] = sanitize_text_field( $operand );
-			}
-
-			$value = $sanitized_value;
-		}
-
-		return $value;
-	}
-
-	/**
-	 * Validate a "created" parameter value.
-	 *
-	 * @param string $value The parameter value.
-	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
-	 * @param string $param The parameter name.
-	 *
-	 * @return bool
-	 */
-	public static function validate_created_field( $value, WP_REST_Request $request, string $param ): bool {
-		if ( empty( $value ) ) {
-			return true;
-		}
-
-		$unix_timestamp_pattern = '^\d+$';
-
-		if ( is_string( $value ) ) {
-			return preg_match( '/' . $unix_timestamp_pattern . '/', $value ) === 1;
-		}
-
-		if ( ! is_array( $value ) ) {
-			return false;
-		}
-
-		$allowed_operators = [ 'gt', 'gte', 'lt', 'lte' ];
-
-		foreach ( $value as $operator => $operand ) {
-			if ( ! in_array( $operator, $allowed_operators ) ) {
-				return false;
-			}
-
-			if ( ! is_scalar( $operand ) || preg_match( '/' . $unix_timestamp_pattern . '/', (string) $operand ) !== 1 ) {
-				return false;
-			}
-		}
-
-		return true;
+		return $response;
 	}
 }
--- a/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
+++ b/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
@@ -22,7 +22,7 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 */
 	protected $rest_base = 'wc_stripe/payment_intents';
 
-	protected const STRIPE_RESPONSE_ALLOWED_FIELDS = [
+	protected const STRIPE_SINGLE_RESPONSE_ALLOWED_FIELDS = [
 		'object',
 		'id',
 		'amount',
@@ -37,7 +37,7 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 		'latest_charge.payment_method_details',
 	];
 
-	protected const STRIPE_EXPAND_PARAM = [
+	protected const STRIPE_SINGLE_EXPAND_PARAM = [
 		'latest_charge',
 		'latest_charge.balance_transaction',
 	];
@@ -138,13 +138,13 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 * @return WP_REST_Response|WP_Error
 	 */
 	public function get_payment_intent( $request ) {
-		$response = $this->fetch_from_stripe( 'payment_intents/' . rawurlencode( $request['id'] ), [ 'expand' => self::STRIPE_EXPAND_PARAM ] );
+		$response = $this->fetch_from_stripe( 'payment_intents/' . rawurlencode( $request['id'] ), [ 'expand' => self::STRIPE_SINGLE_EXPAND_PARAM ] );
 
 		if ( is_wp_error( $response ) ) {
 			return $response;
 		}
 
-		$filtered_response = WC_Stripe_REST_Response_Filter::filter_response( $response, self::STRIPE_RESPONSE_ALLOWED_FIELDS );
+		$filtered_response = WC_Stripe_REST_Response_Filter::filter_response( $response, self::STRIPE_SINGLE_RESPONSE_ALLOWED_FIELDS );
 
 		return rest_ensure_response( $filtered_response );
 	}
--- a/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
+++ b/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
@@ -265,12 +265,12 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 */
 	public static function sanitize_created( $param_value, $request, $param_name ) {
 		if ( ! is_array( $param_value ) ) {
-			$sanitized_value = absint( $param_value );
+			$sanitized_value = (int) $param_value;
 		} else {
 			$sanitized_value = [];
 
 			foreach ( $param_value as $operator => $operand ) {
-				$sanitized_value[ sanitize_key( $operator ) ] = absint( $operand );
+				$sanitized_value[ sanitize_key( $operator ) ] = (int) $operand;
 			}
 		}
 
--- a/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
+++ b/includes/admin/class-wc-stripe-rest-payment-intents-controller.php
@@ -240,7 +240,7 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 		if ( $request->has_param( 'ending_before' ) ) {
 			return new WP_Error(
 				'invalid_starting_after',
-				'Received both starting_after and ending_before parameters. Please pass in only one.'
+				__( 'Received both starting_after and ending_before parameters. Please pass in only one.', 'woocommerce-gateway-stripe' )
 			);
 		}
 
@@ -261,7 +261,7 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 		if ( $request->has_param( 'starting_after' ) ) {
 			return new WP_Error(
 				'invalid_ending_before',
-				'Received both starting_after and ending_before parameters. Please pass in only one.'
+				__( 'Received both starting_after and ending_before parameters. Please pass in only one.', 'woocommerce-gateway-stripe' )
 			);
 		}
 
@@ -305,7 +305,7 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
 	 */
 	public static function sanitize_created( $param_value, $request, $param_name ) {
 		if ( ! is_array( $param_value ) ) {
-			$sanitized_value = self::is_valid_timestamp( $param_value ) ? (int) $param_value : 0;
+			$sanitized_value = self::is_valid_timestamp( $param_value ) ? (int) $param_value : '';
 		} else {
 			$sanitized_value = [];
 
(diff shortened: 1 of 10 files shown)


tests/phpunit/admin/stripe-api-test-response-payloads/1-stripe-rest-api-response.json
```diff
@@ -0,0 +1,281 @@
...
+                "billing_details": {
+                    "address": {
+                        "city": "Austin",
+                        "country": "US",
+                        "line1": "9024 Northgate Blvd",
+                        "line2": null,
+                        "postal_code": "78758",
+                        "state": "TX"
+                    },
+                    "email": "adi.dobrescu@gmail.com",
```
jaclync: This fixture contains real data: a personal email address (also lines 108 and 224), a working receipt URL (line 179) that opens publicly, a Connect application ID, a card fingerprint, and network transaction IDs. Let's use placeholder data in such mock responses, as these values are just for validating the schema.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -39,6 +42,25 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
+	protected const STRIPE_LIST_EXPAND_PARAM = [
+		'data.latest_charge',
+		'data.latest_charge.balance_transaction',
+	];
+
+	protected const STRIPE_LIST_RESPONSE_ALLOWED_FIELDS = [
+		'object',
+		'has_more',
+		'data.id',
+		'data.payment_intent',
```
jaclync: ❓ The [PaymentIntent object](https://docs.stripe.com/api/payment_intents/object) doesn't seem to have `payment_intent` attribute, could you confirm this with the doc reference?

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +209,119 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	 *
+	 * @param string $param_value The parameter value.
+	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 * @param string $param_name The parameter name.
+	 *
+	 * @return bool
+	 */
+	public static function validate_customer_account( $param_value, $request, $param_name ) {
+		return preg_match( '/^' . self::CUSTOMER_ID_PATTERN . '$/', $param_value ) === 1;
+	}
```
daledupreez, line 295: I think we should collapse these methods into two helpers that validate a payment intent and a customer ID so we can clearly guard against drift between them and avoid code duplication. It also makes it clearer what the accepted values are if the validation method describes the expected value rather than the field name.

Also as a smaller nit, we should use Yoda conditions. 😁 
```suggestion
	/**
	 * Validate a parameter value that should be a payment intent ID.
	 *
	 * @param string $param_value The parameter value.
	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
	 * @param string $param_name The parameter name.
	 *
	 * @return bool
	 */
	public static function validate_payment_intent_id( $param_value, $request, $param_name ) {
		return 1 === preg_match( '/^' . self::PAYMENT_INTENT_ID_PATTERN . '$/', $param_value );
	}

	/**
	 * Validate a parameter value that should be a customer ID.
	 *
	 * @param string $param_value The parameter value.
	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
	 * @param string $param_name The parameter name.
	 *
	 * @return bool
	 */
	public static function validate_customer_id( $param_value, $request, $param_name ) {
		return 1 === preg_match( '/^' . self::CUSTOMER_ID_PATTERN . '$/', $param_value );
	}
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -39,6 +42,25 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
+		'has_more',
+		'data.id',
+		'data.payment_intent',
+		'data.created',
+		'data.amount',
+		'data.currency',
+		'data.status',
+		'data.description',
+		'data.latest_charge.billing_details.name',
+		'data.latest_charge.payment_method_details',
```
jaclync: ❓ What is the expected use case of `payment_method_details` for the list response? I'm a bit concerned because now anyone with the wp-admin permission can see the full payment details for a list, while currently there's still a gate on the Stripe dashboard from the Stripe link in the wp-admin order page. How about exposing minimum visible details like the type, card's brand/last4, etc.?

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -76,6 +148,31 @@ public function get_payment_intent( $request ) {
...
+	 *
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function get_payment_intents( $request ) {
+		$response = $this->fetch_from_stripe(
+			'payment_intents',
+			array_merge(
+				$request->get_params(),
+				[ 'expand' => self::STRIPE_LIST_EXPAND_PARAM ]
+			),
```
jaclync: Potential issue: `get_payment_intents` forwards the raw `$request->get_params`, thus any unregistered query params can reach the Stripe API. Stripe rejects unknown params with a 400, and these unsupported params aren't validated. How about ensuring that only the validated & supported params are sent?

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +209,119 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	 * @param string $param_value The parameter value.
+	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 * @param string $param_name The parameter name.
+	 *
+	 * @return bool
+	 */
+	public static function validate_created( $param_value, $request, $param_name ) {
+		if ( empty( $param_value ) ) {
+			return true;
+		}
```
daledupreez: I am not sure if this is correct. I believe that `''` should be invalid, while `0` should be allowed, and both will be treated as `empty` here.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +209,119 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	 * or an array of Unix timestamps keyed by comparison operators (gt, gte, lt, lte).
+	 *
+	 * @param string $param_value The parameter value.
+	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 * @param string $param_name The parameter name.
+	 *
+	 * @return bool
+	 */
+	public static function validate_created( $param_value, $request, $param_name ) {
+		if ( empty( $param_value ) ) {
```
jaclync: Potential issue: the `empty` guard lets an empty string through, thus `?created=` passes validation and is forwarded as `created=`, which Stripe rejects based on the [list API docs](https://docs.stripe.com/api/payment_intents/list#list_payment_intents-created) as `created` must be a Unix timestamp or a dictionary with gt/gte/lt/lte keys. Validation callbacks only run when the param is present, so I don't think the guard is needed for the absent case.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +209,119 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+		}
+
+		$allowed_operators = [ 'gt', 'gte', 'lt', 'lte' ];
+
+		foreach ( $param_value as $operator => $operand ) {
+			if ( ! in_array( $operator, $allowed_operators ) ) {
+				return false;
+			}
+
+			if ( ! is_scalar( $operand ) || preg_match( '/^\d+$/', (string) $operand ) !== 1 ) {
```
daledupreez: Nit: the values are all being passed through `sanitize_text_field()` during sanitization, so we should only be dealing with strings. I am not sure why we're checking for `is_scalar()` and  casting to string for the pre_match() in the loop.

Relatedly, is it worth adding a basic `is_valid_timestamp()` helper method rather than separate regex checks?

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +209,119 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+		}
+
+		if ( ! is_array( $param_value ) ) {
+			return false;
+		}
+
+		$allowed_operators = [ 'gt', 'gte', 'lt', 'lte' ];
+
+		foreach ( $param_value as $operator => $operand ) {
+			if ( ! in_array( $operator, $allowed_operators ) ) {
```
jaclync: I believe this requires the strict flag: on PHP 7.4 (our minimum supported version), `0 == 'gt'` is true, thus a numeric key like `created[0]=123` passes validation and is forwarded to Stripe.

This will probably be flagged this once CI runs on the PR.

```suggestion
			if ( ! in_array( $operator, $allowed_operators, true ) ) {
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
daledupreez: Nit: It may be worth updating these constant names to reflect that they are specifically for the single intent endpoint.

jaclync (commented):
Thanks for the work, I took an initial pass with some comments.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -258,8 +259,218 @@ public function test_malformed_payment_intent_id() {
...
+			],
+			[
+				[
+					'limit'            => 100,
+					'customer'         => 'cus_sad8s6dasd',
+					'customer_account' => 'cus_sad8s6dasdxsa123',
+					'created'          =>
+						[
+							'lt' => '1779802569',
+						],
```
daledupreez, line 351: Can we update this test case to include multiple additional values, just to make sure we're iterating over them correctly?

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -258,8 +259,218 @@ public function test_malformed_payment_intent_id() {
...
+		wp_set_current_user( $admin_id );
+
+		rest_get_server()->dispatch( $request );
+
+		remove_filter(
+			'pre_http_request',
+			$http_stub,
+			10,
+			3
+		);
```
daledupreez: Nit: It may be safer to use `try/finally`. Also, I think we can use a single line call here -- the arguments are not that long.
```suggestion
		try {
			rest_get_server()->dispatch( $request );
		}
		finally {
			remove_filter( 'pre_http_request', $http_stub, 10, 3 );
		}
```

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -258,8 +259,218 @@ public function test_malformed_payment_intent_id() {
...
+
+		$this->assertEquals( $rest_params, $passed_rest_params );
+		$this->assertEquals(
+			$rest_params,
+			array_intersect_key( $pre_http_request_params['search_params'], $rest_params )
+		);
```
daledupreez, line 446: Can we also verify that we're not adding any keys?
```suggestion
		);
		$this->assertEmpty( array_diff_key( $pre_http_request_params['search_params'], $rest_params ) );
```

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -258,8 +259,218 @@ public function test_malformed_payment_intent_id() {
...
+	}
+
+	public static function provide_intent_list_filtering_test_data(): Generator {
+		$test_data_directory = __DIR__ . '/stripe-api-test-response-payloads';
+
+		$test_data_files = glob( $test_data_directory . '/*' );
```
daledupreez: Nit: is it worth limiting the results to `.json` files?
```suggestion
		$test_data_files = glob( $test_data_directory . '/*.json' );
```

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -258,8 +259,218 @@ public function test_malformed_payment_intent_id() {
...
+
+			foreach ( $files as $file ) {
+				if ( str_contains( basename( $file ), 'expected' ) ) {
+					$expected_file = $file;
+				} else {
+					$test_file = $file;
+				}
+			}
+
+			yield "case {$id}" => [
```
daledupreez: Can we set this up so the `id` value has some contextual meaning rather than being just an integer?

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -57,7 +58,7 @@ protected function mock_http_call() {
 	 *
 	 * If non-empty, adds the entries from $params to the request object.
 	 */
-	private function send_request( $params = [], $url = self::SINGLE_INTENT_ENDPOINT_URL ) {
+	private function send_request( $url, $params = [] ) {
```
daledupreez: General nit: for the test code, can we please add explicit types for arguments, especially if we don't have PHPDoc? The only cases where we can skip this is for union types, as we're only maintaining compatibility with PHP 7.4.
```suggestion
	private function send_request( string $url, array $params = [] ) {
```

daledupreez (commented):
Comments up from code inspection.

tests/phpunit/admin/stripe-api-test-response-payloads/1-stripe-rest-api-response.json
en-adobrescu: Replaced real data with placeholders

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: You are right, removed 'data.payment_intent' from the list

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Will tweak the list based on the UI requirements.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Fixed, there is a new validation function that validates time stamps as positive int or strings containing digits only

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Fixed

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Implemented ::is_valid_timestamp and removed is_scalar and preg_match calls

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Done

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Done

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Replaced $id with a description based on file basenames.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Done.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Added 2 more tests for lt + gt and lte+gte combinations.

en-adobrescu:
> Thanks for the work, I took an initial pass with some comments.

Thanks, Jaclyn! I’ve updated this based on your suggestions.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Added the suggested assertion for the case when 'limit' is specified in the request.  When it is not specified, the controller adds the 'limit' param with a default value of 25. I think we should either require the UI to always send the limit (make the parameter required) or just remove the limit param and always send the default to Stripe. Option 1 is probably better.

daledupreez: I think we should keep the current value optional for the UI, if only to match the Stripe API, where it's not required.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
daledupreez: For this one, it feels like dropping the field now will result in a better default behaviour -- if we keep the fields in, we'll need to make sure we remember to remove them later.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +227,110 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	public static function sanitize_created( $param_value, $request, $param_name ) {
+		if ( ! is_array( $param_value ) ) {
+			$sanitized_value = (int) $param_value;
+		} else {
+			$sanitized_value = [];
+
+			foreach ( $param_value as $operator => $operand ) {
+				$sanitized_value[ sanitize_key( $operator ) ] = (int) $operand;
+			}
+		}
```
daledupreez, line 319: Minor note: it could be worth hardening these checks to make sure we're only getting the types we expect.
```suggestion
		if ( is_array( $param_value ) ) {
			$sanitized_value = (int) $param_value;
		} elseif ( is_array( $param_value ) ) {
			$sanitized_value = [];

			foreach ( $param_value as $operator => $operand ) {
				$sanitized_value[ sanitize_key( $operator ) ] = (int) $operand;
			}
		} else {
			$sanitized_value = ''; // Not sure if this is the right fallback value
		}
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +227,110 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	 * @return mixed
+	 */
+	public static function sanitize_created( $param_value, $request, $param_name ) {
+		if ( ! is_array( $param_value ) ) {
+			$sanitized_value = (int) $param_value;
+		} else {
+			$sanitized_value = [];
+
+			foreach ( $param_value as $operator => $operand ) {
+				$sanitized_value[ sanitize_key( $operator ) ] = (int) $operand;
```
daledupreez: I think we should make sure `$operand` is a scalar before we assume it can be cast to `int`.
```suggestion
				if ( is_string( $operand ) && ctype_digit( $operand ) ) ) {
					$sanitized_value[ sanitize_key( $operator ) ] = (int) $operand;
				} else {
					$sanitized_value[ sanitize_key( $operator ) ] = '';
				}
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +227,110 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	public static function is_valid_timestamp( $value ) {
+		if ( is_int( $value ) ) {
+			return $value >= 0;
+		}
+
+		if ( ! is_string( $value ) || '' === $value ) {
+			return false;
+		}
+
+		return 1 === preg_match( '/^\d+$/', $value );
```
daledupreez: Not blocking, but this can also be handled via `ctype_digit()`. Also, in this path, we're not validating that the eventual value is non-zero.
```suggestion
		return ctype_digit( $value ) && ( (int) $value > 0 );
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +227,110 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	 *
+	 * Validates that the value is either a int greater than 0 or a non-empty string containing digits only.
+	 *
+	 * @param mixed $value The value.
+	 *
+	 * @return bool
+	 */
+	public static function is_valid_timestamp( $value ) {
+		if ( is_int( $value ) ) {
+			return $value >= 0;
```
daledupreez: The comment and the code are not fully aligned. The docs claim `> 0`, but the code has `>= 0`.

tests/phpunit/admin/stripe-api-test-response-payloads/1-stripe-rest-api-response.json
```diff
@@ -0,0 +1,281 @@
+{
+    "object": "list",
+    "data": [
+        {
+            "id": "pi_NNNNNNNNNN",
```
daledupreez: Not blocking, but we can use `<id>_test` or similar for these values so they are at least human-readable.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -258,8 +259,262 @@ public function test_malformed_payment_intent_id() {
...
+							'gt' => '1779802569',
+						],
+				],
+			],
+			[
+				[
+					'created' =>
+						[
+							'lte' => '1779802821',
+							'gte' => '1779802569',
```
daledupreez, line 369: Can we also add test cases for a timestamp of `0` within the array? That feels like it would be a good edge case to cover with tests.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -258,8 +259,262 @@ public function test_malformed_payment_intent_id() {
...
+		$this->assertEquals(
+			$rest_params,
+			array_intersect_key( $pre_http_request_params['search_params'], $rest_params )
+		);
+		$test = array_diff_key( $pre_http_request_params['search_params'], $rest_params );
+
+		if ( array_key_exists( 'limit', $rest_params ) ) {
+			$this->assertEmpty( array_diff_key( $pre_http_request_params['search_params'], $rest_params ) );
+		} else {
+			$this->assertEquals(
```
daledupreez, line 453: This feels like it would benefit from an inline comment.
```suggestion
			// We default the `limit` argument when not supplied by the caller.
			$this->assertEquals(
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +227,110 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	/**
+	 * Validate a timestamp value.
+	 *
+	 * Validates that the value is either a int greater than 0 or a non-empty string containing digits only.
+	 *
+	 * @param mixed $value The value.
+	 *
+	 * @return bool
+	 */
+	public static function is_valid_timestamp( $value ) {
```
jaclync: How about marking this as private to avoid unnecessary public overrides / future backward compatibility risk?

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -65,13 +138,55 @@ public function register_routes() {
...
+	/**
+	 * Builds an array of parameters to forward to Stripe API.
+	 *
+	 * @param WP_REST_Request<array<string, mixed>> $request An incoming REST request.
+	 * @param array $params_to_forward Names of params to forward.
+	 * @param array $expand_param Array of value to populate the 'expand' Stripe AAPI param.
+	 *
+	 * @return array
+	 */
+	public static function build_params_to_forward( $request, $params_to_forward, $expand_param ) {
```
jaclync: Can we make this function private so that it doesn't get overridden with future backward compatibility concerns? It's only called internally in this file.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -34,11 +37,31 @@ class WC_Stripe_REST_Payment_Intents_Controller extends WC_Stripe_REST_Base_Cont
...
 		'latest_charge',
 		'latest_charge.balance_transaction',
 	];
 
+	protected const STRIPE_LIST_EXPAND_PARAM = [
+		'data.latest_charge',
+		'data.latest_charge.balance_transaction',
+	];
+
+	protected const STRIPE_LIST_RESPONSE_ALLOWED_FIELDS = [
```
jaclync: super nit: could the order follow the single request, like `STRIPE_LIST_RESPONSE_ALLOWED_FIELDS` followed by `STRIPE_LIST_EXPAND_PARAM`?

daledupreez (commented):
Overall, this is testing well, but I think we should remove the payment details from the response before we merge, as well as considering some of the hardening changes.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
```diff
@@ -258,8 +259,262 @@ public function test_malformed_payment_intent_id() {
...
+					'limit'            => 100,
+					'customer'         => 'cus_sad8s6dasd',
+					'customer_account' => 'cus_sad8s6dasdxsa123',
+					'created'          =>
+						[
+							'lt' => '1779802569',
+						],
+					'starting_after'   => 'pi_3TbL9RJlUF0dQbSB00q0FJS2',
+					'ending_before'    => 'pi_3TbL9RJlUF0dQbSB00q0FJS2',
+				],
```
jaclync, line 353: ❓ These cases in `provide_intent_list_params` send `starting_after` and `ending_before` together, but based on the [Stripe pagination doc](https://docs.stripe.com/api/pagination) these two parameters are mutually exclusive:

> You can use either the starting_after or ending_before parameter, but not both simultaneously.

I tested locally and it returned a 400 error from Stripe (`Received both starting_after and ending_before parameters. Please pass in only one.`). They're passing because the tests are using a mock response, but I think it's worth providing valid cases for developer/reader's reference.

jaclync (commented):
Thanks for the updates, I had a few more comments and will test more in the next pass.

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Created two separated validation methods (validate_starting_after & validate_ending_before) to validate tehe two parameters as timestamps and check that the other one is not specified

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Swapped the two definitions.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Done

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Done

tests/phpunit/admin/class-wc-stripe-rest-payment-intents-controller-test.php
en-adobrescu: Done

tests/phpunit/admin/stripe-api-test-response-payloads/1-stripe-rest-api-response.json
en-adobrescu: Done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Fixed the comment the condition is >= 0.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Replaced preg_match with ctype_digit call and the condition $value >= 0 as the timestamped could be 0.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +225,156 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 * @param string $param_name The parameter name.
+	 *
+	 * @return WP_Error|bool
+	 */
+	public static function validate_starting_after( $param_value, $request, $param_name ) {
+		if ( $request->has_param( 'ending_before' ) ) {
+			return new WP_Error(
+				'invalid_starting_after',
+				'Received both starting_after and ending_before parameters. Please pass in only one.'
```
daledupreez: We should ensure the error message is translated.
```suggestion
				__( 'Received both starting_after and ending_before parameters. Please pass in only one.', 'woocommerce-gateway-stripe' )
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +225,156 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 * @param string $param_name The parameter name.
+	 *
+	 * @return WP_Error|bool
+	 */
+	public static function validate_ending_before( $param_value, $request, $param_name ) {
+		if ( $request->has_param( 'starting_after' ) ) {
+			return new WP_Error(
+				'invalid_ending_before',
+				'Received both starting_after and ending_before parameters. Please pass in only one.'
```
daledupreez: As above, the error message should be translated.
```suggestion
				__( 'Received both starting_after and ending_before parameters. Please pass in only one.', 'woocommerce-gateway-stripe' )
```

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
```diff
@@ -112,4 +225,156 @@ protected function fetch_from_stripe( $endpoint, $params ) {
...
+		if ( ! is_array( $param_value ) ) {
+			$sanitized_value = self::is_valid_timestamp( $param_value ) ? (int) $param_value : 0;
+		} else {
+			$sanitized_value = [];
+
+			foreach ( $param_value as $operator => $operand ) {
+				if ( self::is_valid_timestamp( $operand ) ) {
+					$sanitized_value[ sanitize_key( $operator ) ] = (int) $operand;
+				} else {
+					$sanitized_value[ sanitize_key( $operator ) ] = 0;
```
daledupreez: I don't think we should sanitize this to `0` by default, especially because that means the fallback for invalid parameters will be to request all intents that were ever created by setting it to `0`.

daledupreez (commented):
Overall, the changes are looking good to me, but the `0` defaults in the sanitization don't seem safe. I think it would be better to sanitize to `''`, at which point the incoming value will be treated as invalid.

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Done

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
en-adobrescu: Changed 0 to ''

daledupreez (approved):
Thanks for the updates! These are looking good to me, though I think we wil need to update the fields that are included in follow-up PRs.

jaclync (approved):
Thanks for the updates and it worked well from my testing :shipit: Before merging, could you handle https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5786#discussion_r3747445706 so that it doesn't get missed? Thank you!

includes/admin/class-wc-stripe-rest-payment-intents-controller.php
daledupreez: Noting that this got removed in one of the later commits. Marking as resolved.

PR #5788, opened 2026-08-06
Fix currency warnings in wp-admin when Adaptive Pricing is active

daledupreez:

The primary purpose of this PR is to prevent Stripe currency mismatch notices from displaying in wp-admin when Adaptive Pricing is enabled, but the PR also refactors the code and takes on some related cleanup. The key changes are as follows:
 * Add a secondary CTA when the message does show so a merchant can take action -- it links to the payment method settings UI
 * Implement all exit conditions as early returns, including being in the Stripe settings UI, the dismiss option being set, and adding a check for Adaptive Pricing being active
 * Change the logic to treat only the `stripe` section as part of the settings UI -- previously any section starting with `stripe` would hide the notice
 * Drive the main loop based on the result of `WC_Stripe_UPE_Payment_Gateway->get_upe_enabled_payment_method_ids()` instead of looping over all possible payment methods
 * Access the payment method instances via the `WC_Stripe_UPE_Payment_Gateway->payment_methods[ $payment_method_id ]` instead of constructing a new instance for every payment method, which has some complex side-effects
 * Update the logic to fully rely on the return value from `$payment_method->get_supported_currencies()`, especially `null` returns for payment methods without currency restrictions, especially card and Link, which previously had special checks

#### Before

Shown in all admin pages that show notices and are not the Stripe settings pages, regardless of whether Adaptive Pricing is enabled.

Screenshot 2026-08-06 at 14 40 22

#### After

When Adaptive Pricing is disabled and one or more payment methods don't support the store currency, and the page is not the Stripe settings UI.

Screenshot 2026-08-06 at 14 39 51

_When Adaptive Pricing is active, no notice is shown._

## Testing instructions

 - Run a build from develop or some other branch
 - Ensure that you are connected to Stripe and have Adaptive Pricing enabled
 - Ensure that one or more payment methods do not support your current store currency
 - Run `wp option delete wc_stripe_show_upe_payment_methods_notice` to ensure the notice is not dismissed
 - Navigate to a wp-admin page that shows notices, such as WooCommerce -> Settings or event Appearance -> Themes
 - Confirm that you see the currency notice, and that there is no action you can take in response
 - Now switch to this build, and reload the relevant admin pages
 - Verify that Adaptive Pricing is still enabled and active
 - Confirm that you no longer see the currency notice
 - Now disable Optimized Checkout, and reload your original test page (this keeps the Adaptive Pricing option enabled, but the feature is off)
 - Verify that you now see the notice, and that the CTA is present
 - Click on the CTA, and confirm that you're taken to the Stripe payment methods settings UI
 - Re-enable Optimized Checkout and disable Adaptive Pricing
 - Reload your test page, and confirm that you still see the notice with the CTA
---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -461,28 +461,57 @@ class WC_Stripe_Admin_Notices {
 	 * @return void
 	 */
 	public function payment_methods_check_environment() {
-		// phpcs:ignore
+		// phpcs:ignore WordPress.Security.ValidatedSanitizedInput.MissingUnslash,WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 		$is_stripe_settings_page = isset( $_GET['page'], $_GET['section'] ) && 'wc-settings' === $_GET['page'] && 0 === strpos( $_GET['section'], 'stripe' );
-		$currency_messages       = '';
 
-		foreach ( WC_Stripe_UPE_Payment_Gateway::UPE_AVAILABLE_METHODS as $method_class ) {
-			if ( WC_Stripe_UPE_Payment_Method_CC::class === $method_class || WC_Stripe_UPE_Payment_Method_Link::class === $method_class ) {
-				continue;
-			}
-			$method     = $method_class::STRIPE_ID;
-			$upe_method = new $method_class();
-			if ( ! $upe_method->is_enabled() ) {
-				continue;
-			}
-
-			if ( ! $is_stripe_settings_page && ! in_array( get_woocommerce_currency(), $upe_method->get_supported_currencies(), true ) ) {
-				/* translators: %1$s Payment method, %2$s List of supported currencies */
-				$currency_messages .= sprintf( __( '%1$s is enabled - it requires store currency to be set to %2$s<br>', 'woocommerce-gateway-stripe' ), $upe_method->get_label(), implode( ', ', $upe_method->get_supported_currencies() ) );
-			}
+		if ( $is_stripe_settings_page ) {
+			return;
 		}
 
 		$show_notice = get_option( 'wc_stripe_show_upe_payment_methods_notice' );
-		if ( ! empty( $currency_messages ) && 'no' !== $show_notice ) {
+		if ( 'no' === $show_notice ) {
+			return;
+		}
+
+		// If Adaptive Pricing is enabled, we should not show currency compatibility notices.
+		$settings = WC_Stripe_Helper::get_stripe_settings();
+		if (
+			WC_Stripe_Helper::is_checkout_sessions_available() &&
+			'yes' === ( $settings['adaptive_pricing'] ?? 'no' ) &&
+			WC_Stripe_Helper::is_adaptive_pricing_available_for_account()
+		) {
+			return;
+		}
+
+		$currency_messages = '';
+
+		$gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+		if ( ! $gateway instanceof WC_Stripe_UPE_Payment_Gateway ) {
+			return;
+		}
+
+		foreach ( $gateway->get_upe_enabled_payment_method_ids() as $payment_method_id ) {
+			if ( ! isset( $gateway->payment_methods[ $payment_method_id ] ) ) {
+				continue;
+			}
+
+			$upe_method = $gateway->payment_methods[ $payment_method_id ];
+			if ( ! $upe_method instanceof WC_Stripe_UPE_Payment_Method ) {
+				continue;
+			}
+
+			$supported_currencies = $upe_method->get_supported_currencies();
+			if ( null === $supported_currencies ) {
+				continue;
+			}
+
+			if ( ! in_array( get_woocommerce_currency(), $supported_currencies, true ) ) {
+				/* translators: %1$s Payment method, %2$s List of supported currencies */
+				$currency_messages .= sprintf( __( '%1$s is enabled - it requires store currency to be set to %2$s<br>', 'woocommerce-gateway-stripe' ), $upe_method->get_label(), implode( ', ', $supported_currencies ) );
+			}
+		}
+
+		if ( ! empty( $currency_messages ) ) {
 			$this->add_admin_notice( 'upe_payment_methods', 'notice notice-error', $currency_messages, true );
 		}
 	}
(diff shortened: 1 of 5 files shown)


tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
```diff
@@ -595,24 +608,341 @@ public function test_currency_notice_is_shown_for_upe_methods(): void {
...
+				'adaptive pricing'                       => 'yes',
+				'adaptive pricing available for account' => 'IN',
+				'store currency'                         => 'USD',
+				'enabled payment method IDs'             => [
+					WC_Stripe_Payment_Methods::EPS,
+					WC_Stripe_Payment_Methods::BANCONTACT,
+					WC_Stripe_Payment_Methods::IDEAL,
+					WC_Stripe_Payment_Methods::KLARNA,
+				],
+				'expected payment method IDs in notice'  => [
```
diegocurbelo:
```suggestion
				'is stripe settings page'               => false,
				'dismiss option'                        => '',
				'checkout sessions available'           => true,
				'adaptive pricing'                      => 'yes',
				'account country'                       => 'IN',
				'store currency'                        => 'USD',
				'enabled payment method IDs'            => [
					WC_Stripe_Payment_Methods::EPS,
					WC_Stripe_Payment_Methods::BANCONTACT,
					WC_Stripe_Payment_Methods::IDEAL,
					WC_Stripe_Payment_Methods::KLARNA,
				],
				'expected payment method IDs in notice' => [
```

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -461,29 +461,71 @@ protected function get_environment_type(): string {
...
+		$gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+		if ( ! $gateway instanceof WC_Stripe_UPE_Payment_Gateway ) {
+			return;
+		}
+
+		$store_currency = get_woocommerce_currency();
+		if ( empty( $store_currency ) ) {
+			return;
+		}
+
```
diegocurbelo, line 485: The old implementation used `$upe_method->is_enabled()`, which checks `$main_settings['enabled']` so it required the gateway itself to also be enabled. As a result of the change, now the notices are shown even when the Stripe gateway is disabled.

I think we should keep the previous behavior and bail out if the gateway is disabled:

```suggestion

		if ( ! $gateway->is_enabled() ) {
			return;
		}

```

diegocurbelo (commented):
The code changes look good, and it tests well 👍🏼 

Just one small detail, we now show the notices even when the gateway is disabled; and left  a code suggestion to restore the previous behavior.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -461,29 +461,71 @@ protected function get_environment_type(): string {
...
+		if ( 'no' === $show_notice ) {
+			return;
+		}
+
+		// If Adaptive Pricing is enabled, we should not show currency compatibility notices.
+		$settings = WC_Stripe_Helper::get_stripe_settings();
+		if (
+			WC_Stripe_Helper::is_checkout_sessions_available() &&
+			'yes' === ( $settings['adaptive_pricing'] ?? 'no' ) &&
+			WC_Stripe_Helper::is_adaptive_pricing_available_for_account()
```
wjrosa: AI review:

`is_adaptive_pricing_available_for_account()` reaches `WC_Stripe_Account::is_webhook_enabled()`, which on a cold transient makes a synchronous `webhook_endpoints/{id}` GET during render — and calls `WC_Stripe_API::set_secret_key()` with the key snapshotted at webhook creation, without restoring it. If keys were rotated, every Stripe call later in the request (the PMC fetch below, `check_express_checkout_location()`) runs with the stale key.

This hook runs on every wp-admin page, and OC+AP stores never see/dismiss the notice, so it runs there indefinitely — previously this path was limited to Stripe settings screens and checkout.

Suggestion: evaluate this condition only when `$currency_messages` is non-empty (move it after the loop). Same output, but stores with no mismatch never touch account/webhook state. The missing key restore in `is_webhook_enabled()` is pre-existing — worth a separate follow-up.

daledupreez: Excellent catch! I've just pushed up https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5793 to address the bug with the key not being reset.

daledupreez: @wjrosa, I've just pushed updated the implementation to defer the availability checks until we know we have currencies that need to be checked. I think we need to make sure that the webhook status is stored in our database cache rather than a transient, but that's definitely a separate PR.

includes/admin/class-wc-stripe-admin-notices.php
daledupreez: Thanks for flagging this - I've updated the code to add this check as one of the early return conditions in 923f72f93edbddbe11f6a294f252eec5fc22f506.

tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
daledupreez: Good catch. Fixed in 923f72f93edbddbe11f6a294f252eec5fc22f506.

includes/admin/class-wc-stripe-admin-notices.php
daledupreez: Also, I've created https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5794 to switch the web hook status caching from a transient to our database cache, as sites with unreliable transient implementations may be making unnecessary API calls to Stripe.

wjrosa: Thanks! Both looks good 👍

wjrosa (approved):
Looks good to me! 👍

diegocurbelo (approved):
LGTM, the notices are no longer shown when the gateway is disabled 👍🏼

tests/phpunit/admin/class-wc-stripe-admin-notices-test.php
```diff
@@ -571,48 +554,410 @@ public function options_to_notices_map(): array {
...
+				'store currency'                        => '',
+				'enabled payment method IDs'            => [
+					WC_Stripe_Payment_Methods::CARD,
+					WC_Stripe_Payment_Methods::EPS,
+					WC_Stripe_Payment_Methods::BANCONTACT,
+					WC_Stripe_Payment_Methods::IDEAL,
+					WC_Stripe_Payment_Methods::KLARNA,
+				],
+				'expected payment method IDs in notice' => [],
+			],
```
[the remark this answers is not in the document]
daledupreez: Updated in 7ae0cfd52b6671975013fc803cc0bc996547878c.

PR #5789, opened 2026-08-06
Exclude password-protected and hidden products from the Agentic Commerce feed

wjrosa:
Fixes STRIPE-1340

Password-protected and catalog-hidden products were syncing to Stripe's Agentic Commerce catalog. The feed query filters on post status and type only, so both are still `publish` and produce a valid feed row. For a password-protected product the exported `link` renders a password prompt, so an agent can list it, quote a price, and send the shopper to a wall.

Both are now defaults in `should_sync_product()`, alongside the existing subscription exclusion.

Notable details:

- **A variation does not inherit its parent's `post_password`** (it *does* inherit `catalog_visibility`). Variations are what the feed exports, so the password predicate resolves to the parent — otherwise every variation of a protected variable product keeps syncing. A test asserts the non-inheritance so an upstream change surfaces as a failure.
- The partial visibility values (`catalog`, `search`) are deliberately **not** excluded; the product is still reachable by the other route.
- **New `WC_Stripe_Agentic_Commerce_Product_Visibility`.** The filter only governs what is *sent*, so an already-exported product would linger in Stripe until the next full replacement — and `track_product_archive()` bails on the same predicate, so a later trash would never reach Stripe either. The watcher fires the documented `wc_stripe_agentic_commerce_schedule_full_resync` when a product crosses the boundary, keyed on a stored marker so ordinary price edits don't trigger a full resync.
- **No-price products are not excluded here.** As a `should_sync_product()` default it breaks variable products, whose parent can read as priceless before `_price` sync runs and then never flips state. Deferred on the issue.

**Backward compatibility:** no signature, hook, or option changed — only the default the existing filter receives is narrower, and returning `true` still opts a product back in. The behaviour change is intentional and merchant-visible: password-protected and hidden products drop out of the catalog on the next sync.

## Testing instructions

1. Enable Agentic Commerce and complete onboarding.
2. Publish a simple product with a price and confirm it appears in the feed (**Sync now** → feed preview).
3. Set **Publish → Visibility → Password protected**. The product should drop out, and a resync should be enqueued (pending `wc_stripe_agentic_commerce_sync_feed_now` in Action Scheduler). Remove the password and confirm it returns.
4. Repeat with **Catalog visibility → Hidden**.
5. Set **Catalog visibility → Search results only** and confirm it **still** syncs.
6. Password-protect the **parent** of a variable product — all its variations should drop out.
7. Change a product's price and confirm no full resync is enqueued.

---

### Changelog entry

<details>

#### Comment

Changelog and readme entries are included in this PR.

</details>

**Post merge**


includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
```diff
@@ -0,0 +1,173 @@
...
+	 * Schedule a resync when eligibility differs from the stored marker.
+	 *
+	 * Runs `should_sync_product()` rather than the predicates directly so a
+	 * third-party filter's verdict converges too.
+	 *
+	 * @since 10.9.0
+	 * @param WC_Product|false|null $product Product to reconcile.
+	 * @return bool True when a resync was scheduled.
+	 */
+	protected function reconcile( $product ): bool {
```
jaclync: non-blocking nit: any reason this isn't private? I don't see callers outside the class and the tests go through the hooks.

jaclync: non-blocking questions:

1. The name `reconcile` sounds a bit vague to me, how about something like `schedule_resync_if_needed`/`maybe_schedule_resync`?
2. The return value doesn't seem used in this PR. Is it meant for usage in the near future?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
```diff
@@ -0,0 +1,173 @@
...
+		return true;
+	}
+
+	/**
+	 * The eligibility-marker meta key.
+	 *
+	 * @since 10.9.0
+	 * @return string
+	 */
+	public static function get_state_meta_key(): string {
```
jaclync: non-blocking nit: could this be private? Since the meta key is for internal tracking within this class, I think the tests could verify the public behavior instead of the state meta values.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
```diff
@@ -0,0 +1,173 @@
...
+ * when a product's `should_sync_product()` outcome flips.
+ *
+ * The filter only governs what is *sent*, so an already-exported product stays
+ * live until a full feed replacement. The tracker can't close that gap — its
+ * archive path bails on the same predicate, so a later trash never lands.
+ *
+ * @internal
+ * @since 10.9.0
+ */
+class WC_Stripe_Agentic_Commerce_Product_Visibility {
```
jaclync: non-blocking nit: does it make sense to mark this internal class as `final`? It's meant for internal use from docblock, and not mocked in unit tests.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
```diff
@@ -0,0 +1,173 @@
...
+
+	/**
+	 * Last-exported eligibility ('yes' = excluded).
+	 *
+	 * Keyed on state, not raw property changes, so ordinary price edits don't
+	 * enqueue a resync.
+	 */
+	protected const STATE_META_KEY = '_wc_stripe_agentic_commerce_sync_excluded';
```
jaclync: non-blocking nit: could this be private?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
```diff
@@ -0,0 +1,173 @@
...
+	 */
+	public function init(): void {
+		// `post_updated` fires for every product save, so skip registering at all
+		// rather than loading a product just to bail.
+		if ( ! WC_Stripe_Agentic_Commerce_Integration::is_merchant_enabled() ) {
+			return;
+		}
+
+		add_action( 'woocommerce_update_product', [ $this, 'handle_product_save' ], 10, 2 );
+		add_action( 'woocommerce_new_product', [ $this, 'handle_product_save' ], 10, 2 );
```
jaclync: non-blocking❓: do these two cover variation creation/update events? From checking core, variations might have their own actions like `woocommerce_new_product_variation`/`woocommerce_update_product_variation`. A test case can probably verify it. All tests in `WC_Stripe_Agentic_Commerce_Product_Visibility_Test` are currently on simple products.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
jaclync: non-blocking high-level thought: does it make sense to include the product/variation status check here?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
```diff
@@ -0,0 +1,173 @@
...
+		// Missing marker means 'no': products predating this class were exported
+		// under the older defaults, so one excluded now has to be pulled out.
+		$previous_state = get_post_meta( $product_id, self::STATE_META_KEY, true );
+		if ( '' === $previous_state ) {
+			$previous_state = 'no';
+		}
+
+		if ( $previous_state === $new_state ) {
+			return false;
+		}
```
jaclync, line 156: non-blocking observation: because it assumes a missing marker as "not excluded", a newly created product that should be excluded passes through and triggers a resync unnecessarily. I thought about suggesting passing a "new product" parameter, but it looks like product created from wp-admin could have the product created with "auto-draft" then updated from publishing the product. It might only make sense if we add in the product status check later.

jaclync (approved):
LGTM! :shipit: worked well, with some non-blocking comments.

UI/UX observations from testing:
- Is `Preview feed` CTA supposed to display the list of products/variations in the feed in the "Feed preview" section? The subtitle says "Check which products will be made available to AI agents on the next sync, ..." while it just includes the breakdown data. I verified the product in the feed in the Stripe dashboard.
- After clicking `Sync now`, the "Recent syncs" section was updated when the sync is updated. Does it make sense to also update the "Feed preview" section?
- The "Excluded" layout is a bit off from "Included"/"With errors":

Image

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
wjrosa: No reason — made it private in 6ebe254.

wjrosa: Renamed to `maybe_schedule_resync()` in 6ebe254. The return value had no planned consumer, so I dropped it too.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
wjrosa: Done in 6ebe254.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
wjrosa: Done in 6ebe254 (getter removed too — see the other thread).

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
wjrosa: Done in 6ebe254: constant private, getter removed, and the test now asserts the behavior (a follow-up ordinary edit stays quiet) instead of the meta value.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
wjrosa: They don't — good catch. 6ebe254 adds the `woocommerce_new/update_product_variation` hooks, accepts `product_variation` in the `post_updated` handler, and adds a variation test. The test also surfaced that the object passed to the variation-creation hook lacks the parent-data snapshot `WC_Product_Variation::get_catalog_visibility()` reads, so the save handler now re-reads by ID.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
wjrosa: I think so, as a follow-up: today a publish→draft flip isn't caught by anything (the feed query filters on status, so the product just lingers in Stripe until the next full sync). Adding status to `should_sync_product()` would close that, but it changes eligibility for all three sync surfaces, so I'd rather do it with its own tests than widen this PR.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
wjrosa: Agreed. Worst case is one redundant full-resync schedule when a product is created already-excluded, which seemed acceptable versus threading creation context through the hooks. Revisiting alongside the status check sounds right.

wjrosa:
Thanks for the thorough testing! On the UI/UX observations:

- **Excluded layout**: fixed in 102e3cc + 8e8daab — the Details toggle rendered inline beside the label, making that stat taller, and the row centered items vertically, floating the value above its siblings. The row is now top-aligned (via the Flex `align` prop — its emotion class beats styled() CSS) with the toggle on its own line.
- **Preview vs. subtitle copy**: the preview endpoint only returns counts, the excluded breakdown, and validation errors today — listing the included products needs pagination for large catalogs, so I'd treat it as a follow-up (or soften the subtitle copy).
- **Refreshing the preview after Sync now**: the two sections keep separate fetch-on-demand state, so this needs shared state/a store subscription — also follow-up material.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-visibility.php
```diff
@@ -0,0 +1,164 @@
...
+
+		if ( $previous_state === $new_state ) {
+			return;
+		}
+
+		update_post_meta( $product_id, self::STATE_META_KEY, $new_state );
+
+		// Fire the documented contract rather than calling the integration, so
+		// this path behaves identically to an adapter's.
+		do_action( 'wc_stripe_agentic_commerce_schedule_full_resync' );
```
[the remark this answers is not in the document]
wjrosa: Valid — fixed in 151f666: a parent boundary flip now brings each variation's marker current (the full resync it schedules already covers them), with the suggested regression test. Verified the test fails without the fix.

PR #5790, opened 2026-08-06
Agentic Commerce: bulk edit, quick edit, and a sync-status column for catalog exclusion

wjrosa:
Fixes STRIPE-1339

Merchants can currently exclude a product from the Agentic Commerce catalog sync only by opening each product and ticking the Inventory-tab checkbox. This PR adds a bulk path on the **Products list table**, reusing the existing `_wc_stripe_agentic_commerce_exclude` meta through `WC_Stripe_Agentic_Commerce_Product_Exclusion` — no new data model, option, or migration.

New `WC_Stripe_Agentic_Commerce_Product_List_Table` (registered alongside the existing meta box, only when the merchant has Agentic Commerce enabled):

- **Bulk actions** — "Exclude from Agentic Commerce sync" / "Include in Agentic Commerce sync". Per-product `current_user_can( 'edit_product' )` check, the same feature and product-type gates as the meta box (so a bulk action can't clobber state the single-product editor would refuse to touch), a single immediate resync scheduled per batch (only when a flag actually changed), and an admin notice with the processed count.
- **Quick edit** — an "Exclude from the Stripe Agentic Commerce catalog sync" checkbox, persisted via `woocommerce_product_quick_edit_save` (WooCommerce verifies its quick-edit nonce before firing it). A small admin script (`assets/js/stripe-agentic-product-quick-edit.js`) copies the row's current state into the cloned quick-edit form.
- **Column + filter** — an "Agentic Commerce" column (Synced / Excluded, dash for unsupported types) and a sync-status dropdown that narrows the main query via `meta_query`, treating unset meta as synced.

The supported-types list moved from the meta box into `Product_Exclusion` behind a `get_supported_types()` getter so all editor surfaces share one source of truth.

### Backward compatibility

Additive only. New class, new bulk action keys, new column/query-var names — all namespaced `wc_stripe_agentic_*`. No existing hook, option key, or public signature changes; the meta box refactor removes only a `private` constant. The exclusion meta and its should-sync filter semantics are untouched. Not exercised under multisite; the feature reads/writes site-scoped post meta and options only.

## Testing instructions

1. Enable Agentic Commerce in **Payments → Stripe → Agentic commerce** (requires the feature flag).
2. Go to **Products → All Products**. Confirm a new **Agentic Commerce** column shows "Synced" for simple/variable products and "—" for other types (e.g. grouped/external).
3. Select several products, apply **Bulk actions → Exclude from Agentic Commerce sync**. Confirm the success notice, the column flipping to "Excluded", and a `wc_stripe_agentic_commerce_immediate_sync` job queued in **WooCommerce → Status → Scheduled Actions**.
4. Re-apply the same bulk action — the notice repeats but no new sync job is queued (no-op batch).
5. Apply **Include in Agentic Commerce sync** to flip them back.
6. Open **Quick Edit** on an excluded product: the checkbox is pre-ticked; untick and update, and the column returns to "Synced". Confirm the product editor's Inventory-tab checkbox agrees.
7. Use the **Filter by Agentic Commerce** dropdown with Synced / Excluded and confirm the list narrows accordingly, including products that have never been toggled (they count as synced).
8. Confirm exclusion still takes effect in the feed preview (**Stripe settings → Agentic commerce → Preview feed**).

`npm run test:php -- --filter Agentic`, `npm run lint:php`, and `npm run phpstan` all pass.


assets/js/stripe-agentic-product-quick-edit.js
```diff
@@ -0,0 +1,39 @@
...
+
+		var excluded =
+			'yes' ===
+			$( '#post-' + postId )
+				.find( '.column-wc_stripe_agentic_sync [data-excluded]' )
+				.attr( 'data-excluded' );
+
+		$( '#edit-' + postId )
+			.find( 'input[name="_wc_stripe_agentic_commerce_exclude"]' )
+			.prop( 'checked', excluded );
```
jaclync: ❓ Instead of deriving the excluded status from the sync status column attribute, how about using `add_inline_data` to add a hidden div for each row (like the excluded value), and reading the excluded value to populate here + hiding the checkbox if there's no inline state (e.g. unsupported product types)? Right now, the excluded checkbox is still shown for unsupported product types.

Also, the column attribute might not be the most reliable because it's removable. Some mu-plugin can unset the column at a later priority that causes the quick-edit checkbox not being rendered.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
```diff
@@ -0,0 +1,425 @@
...
+
+		foreach ( (array) $post_ids as $post_id ) {
+			$post_id = (int) $post_id;
+
+			if ( ! current_user_can( 'edit_product', $post_id ) ) {
+				continue;
+			}
+
+			$product = wc_get_product( $post_id );
+			if ( ! $product instanceof WC_Product || ! in_array( $product->get_type(), WC_Stripe_Agentic_Commerce_Product_Exclusion::get_supported_types(), true ) ) {
```
jaclync: non-blocking suggestion: this `instanceof WC_Product && in_array( get_type(), get_supported_types(), true )` guard exists at five sites: here, `render_column` (L239), `save_quick_edit` (L381), and twice in the meta box file. Can this be a shared helper + unit testing coverage like in `WC_Stripe_Agentic_Commerce_Product_Exclusion::supports( $product ): bool` in case the eligibility criteria change in the future?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
```diff
@@ -0,0 +1,425 @@
...
+					'key'     => $meta_key,
+					'compare' => 'NOT EXISTS',
+				],
+				[
+					'key'     => $meta_key,
+					'value'   => 'yes',
+					'compare' => '!=',
+				],
+			];
+		}
```
jaclync, line 366: ❓ The UI/UX from testing the filter felt confusing. The filter options are "synced" and "excluded", the excluded filter results were fine. But the synced filter results include unsupported product types and draft products, which should not and aren't synced to the feed, just in the filtered list UI.

Image

Potential fixes: 
1. Constrain the filter to supported product types + publish status.
2. Rename the filter labels to the query here ("Excluded" / "Not excluded").

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
```diff
@@ -0,0 +1,425 @@
...
+		if ( ! in_array( $product->get_type(), WC_Stripe_Agentic_Commerce_Product_Exclusion::get_supported_types(), true ) ) {
+			return;
+		}
+
+		// phpcs:ignore WordPress.Security.NonceVerification.Missing -- WC verified its quick-edit nonce before firing this hook.
+		$meta_key     = WC_Stripe_Agentic_Commerce_Product_Exclusion::get_meta_key();
+		$posted_value = isset( $_POST[ $meta_key ] ) ? sanitize_text_field( wp_unslash( $_POST[ $meta_key ] ) ) : '';
+		$excluded     = 'yes' === $posted_value;
+
+		$changed = WC_Stripe_Agentic_Commerce_Product_Exclusion::set_excluded( $product->get_id(), $excluded );
```
jaclync, line 457: If we do hide the checkbox for unsupported product types, we might want to skip persisting the excluded field if it's not in the POST data.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
```diff
@@ -0,0 +1,425 @@
...
+
+		$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
+
+		wp_enqueue_script(
+			'wc-stripe-agentic-product-quick-edit',
+			plugins_url( 'assets/js/stripe-agentic-product-quick-edit' . $suffix . '.js', WC_STRIPE_MAIN_FILE ),
+			[ 'jquery', 'inline-edit-post' ],
+			WC_STRIPE_VERSION,
+			true
+		);
```
jaclync, line 490: ❓ The new column isn't legible on my desktop under certain font size threshold (fixed after shrinking the font size):

Image

This is likely due to no width given to this column. Can some inline style to give the column some (safe yet flexible) width fix it here?

jaclync (commented):
Noticed a few things in the initial review & testing, will test more and check on the unit tests in the next pass!

assets/js/stripe-agentic-product-quick-edit.js
wjrosa: Done in 79d568e — the state now lives in an `add_inline_data` hidden div (rendered only for supported types), and the script hides + disables the checkbox when the div is missing. Quick edit no longer depends on the removable column.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
wjrosa: Done in 79d568e: `WC_Stripe_Agentic_Commerce_Product_Exclusion::supports( $product )`, used at all five sites, with a dataProvider test.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
wjrosa: Went with option 1 in 79d568e: the synced filter now constrains to supported product types + publish status, with query-level test coverage for drafts and grouped products.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
wjrosa: Done in 79d568e. Since an unchecked checkbox is also absent from the POST, a hidden marker field travels with the checkbox (disabled together with it for unsupported rows); no marker → the save is skipped instead of clobbering a stored `yes`.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
wjrosa: Done in 79d568e — the column now gets `width: 10%` via `wp_add_inline_style` on the `list-tables` handle.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
```diff
@@ -0,0 +1,425 @@
...
+	 * @param string $post_type   Current post type.
+	 * @return void
+	 */
+	public function render_quick_edit_field( $column_name, $post_type ): void {
+		if ( self::COLUMN_KEY !== $column_name || 'product' !== $post_type ) {
+			return;
+		}
+
+		printf(
+			'<fieldset class="inline-edit-col-right"><div class="inline-edit-col"><label class="alignleft"><input type="checkbox" name="%s" value="yes"><span class="checkbox-title">%s</span></label></div></fieldset>',
```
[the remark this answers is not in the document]
wjrosa: Addressed in 79d568e — the quick-edit script now hides and disables the checkbox for rows without the `add_inline_data` state div (rendered only for supported types), and the save handler skips rows whose field never submitted.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
```diff
@@ -0,0 +1,493 @@
...
+	public function render_quick_edit_field( $column_name, $post_type ): void {
+		if ( self::COLUMN_KEY !== $column_name || 'product' !== $post_type ) {
+			return;
+		}
+
+		// The hidden marker travels with the checkbox so the save handler can
+		// tell "field never rendered" (skip) from "checkbox unchecked" (clear
+		// the flag) — an unchecked checkbox is simply absent from the POST.
+		printf(
+			'<fieldset class="inline-edit-col-right"><div class="inline-edit-col"><label class="alignleft"><input type="hidden" name="%1$s" value="1"><input type="checkbox" name="%2$s" value="yes"><span class="checkbox-title">%3$s</span></label></div></fieldset>',
```
jaclync: non-blocking nit: how about disabling this field by default and then enabled in the JS later? If the JS is replaced by a third-party plugin or not run, this field stays enabled and is included in the submission even for unsupported products.

jaclync (approved):
Thanks for the updates, tested well and LGTM! :shipit:

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
wjrosa: Good call, done in 2539a0f. The template now ships both the checkbox and the hidden marker with `disabled`, and the quick-edit script enables them only after populating the checkbox from the row's inline data. This also closes a worse gap than the unsupported-types one: without the script, the checkbox would always submit unchecked for supported products too, so quick edit could silently clear a stored exclusion. Now no script means no marker in the POST, and the save handler skips the row.

PR #5791, opened 2026-08-06
Agentic Commerce: eligibility-aware sync column and settings link

wjrosa:
Part of STRIPE-1339. Stacked on #5790 (targets its branch; merge after review there).

Follow-up to #5790, addressing two discoverability gaps:

- **Eligibility-aware column.** The "Agentic Commerce" column derived its "Synced" label from the exclude toggle alone, so a product kept out of the feed by any other rule — the built-in sync-eligibility defaults (which #5789 extends to password-protected and hidden products) or a third-party `woocommerce_agentic_commerce_should_sync_product` filter — still read as "Synced". The column now renders a third state, **"Not synced"** (with an explanatory tooltip), from the full `should_sync_product()` verdict. The toggle keeps priority for the label, and the `data-excluded` attribute the quick-edit script reads continues to mirror the toggle alone.
- **Filter label honesty.** Eligibility is computed per row at render time and cannot be expressed as SQL, so the filter dropdown keeps querying the toggle meta only; its "Synced products" option is renamed **"Not excluded"** so the filter never claims more than it checks.
- **Settings-tab link.** The Agentic Commerce settings tab now links to the pre-filtered excluded-products view (`edit.php?post_type=product&wc_stripe_agentic_sync_status=excluded`), so the settings hub can answer "what have I excluded" without duplicating a management UI. The link is relative on purpose — the settings page lives in wp-admin, so it resolves on any install layout.

### Backward compatibility

Additive/label-only changes to UI introduced in #5790 (unreleased). No hook, option, or signature changes. The per-row `should_sync_product()` call in the column applies the existing public filters in admin context, same as the feed already does.

## Testing instructions

1. With Agentic Commerce enabled, go to **Products → All Products**. Products excluded via the toggle show **Excluded**; ordinary products show **Synced**.
2. Add a `woocommerce_agentic_commerce_should_sync_product` filter returning false for some product (or, once #5789 lands, password-protect a product or set its catalog visibility to Hidden): its column shows **Not synced** with a tooltip, while quick edit still shows the exclude checkbox unticked.
3. Toggle-exclude that same product: the column shows **Excluded** (toggle wins).
4. The filter dropdown now offers **Not excluded** / **Excluded products** and filters by the toggle as before.
5. In **Payments → Stripe → Agentic commerce**, with the feature enabled, confirm the "View excluded products" link opens the Products list pre-filtered to excluded products. Disable the feature and confirm the link disappears.

`npm run test:php -- --filter WC_Stripe_Agentic_Commerce_Product_List_Table_Test` (22 tests), `npm run test:js -- client/settings/agentic-commerce` (50 tests), `npm run lint:php`, and `npm run phpstan` all pass.


jaclync (commented):
I think there are some conflicts with the latest base PR branch https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5790, I'll take a look once the base PR is merged & this PR is updated (code and description).

wjrosa:
@jaclync thanks for flagging. Fixed the conflicts and re-targeted the branch to `develop` after merging #5790

client/settings/agentic-commerce/index.js
```diff
@@ -190,6 +190,25 @@ const AgenticCommerceSection = forwardRef( ( props, ref ) => {
...
+											mixedString: __(
+												'To keep specific products out of the catalog, exclude them on the Products screen — individually, via Quick Edit, or with the bulk actions. {{excludedLink}}View excluded products{{/excludedLink}}',
+												'woocommerce-gateway-stripe'
+											),
+											components: {
+												excludedLink: (
+													// Relative on purpose: the settings page lives in
+													// wp-admin, so this resolves on any install layout.
+													// eslint-disable-next-line jsx-a11y/anchor-has-content
+													<a href="edit.php?post_type=product&wc_stripe_agentic_sync_status=excluded" />
```
jaclync: non-blocking: maybe this link can come from the backend, like `agentic_commerce_logs_url` in `wc_stripe_settings_params`? If the status key/value ever changes, the url here might get out of sync from the backend.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
```diff
@@ -246,9 +246,26 @@ public function render_column( $column, $post_id ): void {
...
+			esc_html_e( 'Excluded', 'woocommerce-gateway-stripe' );
+			return;
+		}
+
+		// The exclude toggle is off, but the full eligibility verdict can still
+		// say no — sync rules like catalog visibility, or a third-party filter.
+		// Show that as its own state so "Synced" is never a lie; the quick-edit
+		// checkbox is unaffected because it reads the toggle alone from the
+		// inline-data div.
+		if ( ! WC_Stripe_Agentic_Commerce_Product_Mapper::should_sync_product( $product ) ) {
```
jaclync, line 283: non-blocking: `should_sync_product()` includes applying a filter (+ deprecated one), and is now run per row during admin render. Third-party hooks can make it slow. Maybe we can add a note to the hook's docblock that it also runs during admin rendering for performance considerations?

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
```diff
@@ -246,9 +246,26 @@ public function render_column( $column, $post_id ): void {
...
+			printf(
+				'<span title="%s">%s</span>',
+				esc_attr__( 'Kept out of the catalog by sync eligibility rules, not by the exclude toggle.', 'woocommerce-gateway-stripe' ),
+				esc_html__( 'Not synced', 'woocommerce-gateway-stripe' )
+			);
+			return;
+		}
+
+		esc_html_e( 'Synced', 'woocommerce-gateway-stripe' );
```
jaclync, line 292: ❓ Do we want to update the label for non-published products, which render "Synced" right now? Because `supports()` nor `should_sync_product()` checks post status, draft/pending/private products are labeled "Synced" while they are never sent to Stripe.

Adding the publish status to `should_sync_product()` might have some complications, maybe just a label update here.

Image

jaclync (approved):
Worked as expected! :shipit:

client/settings/agentic-commerce/index.js
wjrosa: Done in 5b24988. The URL now comes from `wc_stripe_settings_params`, built by a getter on the list-table class next to the query var, with the relative URL as fallback.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
wjrosa: Done in 5b24988, added a note to the filter docblock.

includes/agentic-commerce/class-wc-stripe-agentic-commerce-product-list-table.php
wjrosa: Good catch, done in 5b24988. Non-published products now render "Not synced" with a "Only published products are synced." tooltip; went with the label-only fix, and the check runs before the filter so drafts skip it.

PR #5795, opened 2026-08-07
Fix STRIPE-1142: Strip platform fee fields injected via wc_stripe_generate_create_intent_request

### LINKED ISSUE #5417: Harden `wc_stripe_generate_create_intent_request` filter against unintended fee injection
## Summary

The `wc_stripe_generate_create_intent_request` filter (applied in `includes/class-wc-stripe-intent-controller.php` at lines ~923, ~975, and ~1024, plus references in `abstract-wc-stripe-payment-gateway.php`, `trait-wc-stripe-subscriptions.php`, and `class-wc-stripe-upe-payment-gateway.php`) passes the **entire** `$request` array to third-party code immediately before it's POSTed to `/v1/payment_intents`. That's the documented extension point for things like adding mandate options or rewriting currency for multi-currency add-ons.

However, because the filter passes the full request body, third-party plugins, themes, or mu-plugins can also inject **platform-level fields** that the merchant never intended to set — most consequentially `application_fee_amount`, but also `transfer_data`, `on_behalf_of`, and `application_fee`. When that happens on a Stripe-Connect-direct-charge install, the field is sent verbatim to Stripe and the platform account ends up collecting fees it never asked for.

## How we noticed

The Automattic platform account that publishes this plugin has been seeing small unexpected `application_fee_amount` charges. Stripe traced two recent ones to PI creation requests where `application_fee_amount` was set in the request itself; the plugin codebase does **not** set this field anywhere (only `null` values appear in test-fixture response JSON). Cross-referencing the affected merchant's installed plugins pointed at a multi-currency add-on hooking this filter — consistent with the structure of WPML's `class-wcml-payment-gateway-stripe.php` which is one of ~300 third-party files publicly hooking this filter.

## Suggested hardening (open for discussion)

1. **Defensive `unset()` after the filter.** Strip platform-only fields the plugin doesn't intend to send, regardless of who added them:
   ```php
   $request = apply_filters( 'wc_stripe_generate_create_intent_request', $request, $order, $prepared_source );
   unset(
       $request['application_fee_amount'],
       $request['application_fee'],
       $request['transfer_data'],
       $request['on_behalf_of']
   );
   ```
   Pro: closes the loop completely. Con: would break any merchant deliberately using this hook to set those fields (rare, but possible).

2. **Allowlist filter approach.** Document a stable set of keys third-party code may add, and `array_intersect_key` against it after the filter.

3. **Docs-only change.** Add an inline PHPDoc warning above each `apply_filters` call clarifying that platform-level fields will produce surprising behavior and shouldn't be set via this hook.

Happy to send a PR for whichever direction the team prefers.

## References

- Filter call sites: `includes/class-wc-stripe-intent-controller.php` lines ~923, ~975, ~1024
- Stripe API docs on [`application_fee_amount`](https://stripe.com/docs/api/payment_intents/create#create_payment_intent-application_fee_amount)

johndcoy:
Fixes STRIPE-1142 · Closes #5417

The `wc_stripe_generate_create_intent_request` filter hands the whole request array to third-party code immediately before it goes to Stripe. Some plugins use that to set `application_fee_amount`. On a Connect direct charge that fee transfers to the *platform* account, so on stores connected through this plugin's Stripe App OAuth flow it is collected by the plugin platform rather than by the store, and the merchant nets less on the sale. Refunds compound it, since Stripe does not refund an application fee automatically, so the store repays the customer in full while the fee stays with the platform.

## What this does

Adds `WC_Stripe_Helper::strip_platform_fee_fields()`, applied after every `apply_filters()` call for this hook. It removes `application_fee_amount` and `application_fee`, and logs what it removed along with the value, so the source plugin can be traced from the store's own logs without needing source access.

## What it deliberately does not do

Stores connected with their own API keys are left untouched. They may legitimately be their own Connect platform, for example a multi-vendor marketplace charging vendor commission ( [a public case](https://github.com/thelight413/stripecustom/blob/46b48dfaa6d7ddc59806a5e954f48afa9cb0569a/stripecustom.php#L60-L86) ), and stripping the field there would break a supported setup.

`transfer_data` and `on_behalf_of` are also left alone. Those have real Connect on-behalf-of patterns in the wild.

## Scope note

The issue lists three call sites. There are six `apply_filters` calls for this hook and all six are covered:

| File | Sites |
| --- | --- |
| `includes/class-wc-stripe-intent-controller.php` | 3 |
| `includes/abstracts/abstract-wc-stripe-payment-gateway.php` | 2 |
| `includes/payment-methods/class-wc-stripe-upe-payment-gateway.php` | 1 |

The filter's canonical docblock in the abstract gateway now carries a note explaining why platform-level fields should not be set through it.

## Testing

Seven unit tests in `tests/phpunit/class-wc-stripe-helper-test.php`:

- Strips the fee fields on `connect` and on `app` connection types
- Leaves `manual` and unrecorded connection types alone
- Leaves stores with no API keys alone
- Honours `test_connection_type` when the store is in test mode
- Leaves `metadata`, `payment_method_options`, `transfer_data` and `on_behalf_of` intact
- Returns the request unchanged when no fee fields are present

PHPCS is clean on every changed line. I was not able to run the PHPUnit suite locally as I had no database available, so CI is the first real execution of these tests.

## Reviewer note

@daledupreez worth a specific look at the OAuth gate. A store that connected via OAuth *and* runs its own marketplace commission through this filter will see a behaviour change. My read is that this is a fix rather than a regression, because that commission was already being collected by the plugin platform rather than by them, but I would like you to agree with that read before this ships.


includes/class-wc-stripe-helper.php
```diff
@@ -1495,6 +1460,64 @@ public static function add_payment_method_to_request_array( string $payment_meth
 		return $request;
 	}
 
+	/**
+	 * The platform-level fee fields that this gateway never sets itself.
+	 *
+	 * @var string[]
+	 */
+	public const PLATFORM_FEE_FIELDS = [ 'application_fee_amount', 'application_fee' ];
```
jaclync, line 1552: I'd suggest marking these constants as private as they're only used within the helper function, unless it's intended to be extensible. Any public interface can be modified externally with backward compatibility risk.

jaclync: nit: let's move the constants up to the top of the file along with other constants.

includes/class-wc-stripe-helper.php
```diff
@@ -1495,6 +1460,64 @@ public static function add_payment_method_to_request_array( string $payment_meth
...
+				continue;
+			}
+
+			WC_Stripe_Logger::warning(
+				sprintf(
+					'Removed %1$s (value: %2$s) from an intent request. It was added by third-party code hooking wc_stripe_generate_create_intent_request. This store is connected via Stripe App OAuth, so the fee would have been collected by the plugin platform account rather than by this store.',
+					$field,
+					wp_json_encode( $request[ $field ] )
+				)
+			);
```
jaclync, line 1597: ❓ This is more of a product level question: since warnings require debug logging enabled (default is off and probably common in production stores), such warnings can go unnoticed. Do you think something more visible like an order note makes sense here? It can help inform merchants if the fee amount is intentional for them to take actions.

jaclync (commented):
❓ An overall question about the stripping code placement, tagging @daledupreez for input as another reviewer:

After the `wc_stripe_generate_create_intent_request` filter, the request goes through `WC_Stripe_API::request` which [applies another filter `wc_stripe_request_body/woocommerce_stripe_request_body (deprecated)`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/bd1d857f878139c3f699bbf9527badd5b3d5c43c/includes/class-wc-stripe-api.php#L243) on the request body before the API request goes to Stripe. Third-party plugins can still add platform fee fields with the `*_request_body` filters. WDYT of placement in `WC_Stripe_API::request` after the `*_request_body` filters? This way, we just need one call to the helper instead of 6 individual call sites. It'd also help with legacy charges API requests.

```php
// WC_Stripe_API::request(), right after the wc_stripe_request_body filter:
$request = apply_filters( 'wc_stripe_request_body', $request, $api );

if ( 'POST' === $method ) {
	$request = WC_Stripe_Helper::strip_platform_fee_fields( $request );
}
```

and the helper can early return on empty fee fields:

```php
// WC_Stripe_Helper::strip_platform_fee_fields
if ( // no fee fields ) { return $request; }
// existing OAuth gate + unset + log…
```

The main downside I can think of is that it applies to all Stripe requests, with the overhead minimized by the POST request & fee fields checks.

---

If we're keeping the current placement in in individual call sites, I would suggest combining the `wc_stripe_generate_create_intent_request` filter and the stripping logic into one shared helper.

daledupreez:
I agree with @jaclync's suggestion to run this from the Stripe API class when we're making a POST call to a path that starts with `payment_intents` - as far as I can tell, the paths where this would take effect are either POST calls to `payment_intents` or POST calls to `payment_intents/<intent_id>` or a sub-path like `/confirm`.

jaclync:
@johndcoy If you might not have the time to update the PR, I'm happy to take over the development - just let me know 🙂 .

daledupreez:
@johndcoy and @jaclync, I've adapted the code here to run from within `WC_Stripe_API::request()` in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5947, so I am going to close this PR for the time being.

PR #5640, opened 2026-06-25
Enforce and save custom required checkout fields in express checkout

wjrosa:
Fixes STRIPE-672
Fixes #4597

## Changes proposed in this Pull Request:

Custom **required** checkout fields added by third-party plugins (e.g. conditional/checkout-field editors) did not stop the Express Checkout (Apple Pay / Google Pay) flow on classic checkout, and any values a shopper entered in them were not saved to the order. Merchants had to chase customers for the missing information.

The classic-checkout custom-field integration already existed (#4424 / #4455) but shipped behind a controlled-rollout filter (`wc_stripe_express_checkout_enable_classic_checkout_custom_fields`) that defaulted to `false`, so `WC_Stripe_Express_Checkout_Custom_Fields::init()` never ran and the Store API order never saw the data.

This PR:

- **Enables the integration by default.** The filter remains as an opt-out (`add_filter( 'wc_stripe_express_checkout_enable_classic_checkout_custom_fields', '__return_false' )`).
- **Persists the entered values** to the order keyed by field ID in `process_custom_checkout_data()`, mirroring how block-checkout additional fields are stored in `WC_Stripe_Express_Checkout_Element::add_order_meta()`. This runs before the `wc_stripe_express_checkout_update_order_meta` action so the data is saved even when no third-party plugin hooks that stand-in action, while still letting third-party handlers override.
- **Guards processing with `is_express_checkout_context()`.** The underlying `woocommerce_store_api_checkout_update_order_from_request` hook fires for *every* Store API checkout. Without the guard, a missing required classic custom field would wrongly block normal Block/OC checkout orders that never went through express checkout. The guard scopes validation + save to genuine ECE requests (identified by the `X-WCSTRIPE-EXPRESS-CHECKOUT` header + nonce).
- **Directs buyers to the checkout page when the fields can't be filled in.** When a required custom field blocks an ECE payment started on a page without the classic checkout form (product/cart), the error now also says "Please go to the checkout page, fill in the required fields, and complete your order from there." Detection relies on the custom-data payload: the client attaches it (even when empty) only when `form[name="checkout"]` exists on the page, so its absence identifies flows where the buyer never saw the fields — and One Page Checkout product pages, which do render the form, correctly keep the plain per-field errors.

### How can this code break? What prevents it?

[...]

PR #5800, opened 2026-08-11
Exclude core account fields from express checkout custom-field enforcement

wjrosa:
Fixes STRIPE-1345

With **"When creating an account, automatically generate an account password" disabled** (`woocommerce_registration_generate_password = no`), every express checkout payment (Apple Pay, Google Pay, Link, Amazon Pay) fails with `wc_stripe_express_checkout_missing_required_fields` — *"Create account password is a required field."* — for any product and any user, logged-in or guest, while the same cart completes classic checkout fine. Reported in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5631#issuecomment-5249775145; regression from the custom-fields enforcement introduced in #5640 (unreleased, 10.9.0).

**Root cause.** When that option is off, WooCommerce core adds `account_password` (and `account_username`, for the equivalent username option) to `WC()->checkout()->get_checkout_fields()` with `required => true` **unconditionally** — core itself only enforces those fields when an account is actually being created. `WC_Stripe_Express_Checkout_Custom_Fields::get_custom_checkout_fields( 'classic' )` treats every non-standard field as a merchant custom field, and its exclusion list only covers address fields, phone, email, and `order_comments` — so `account_password` becomes a "required custom field" the express sheet can never collect, and `process_custom_checkout_data()` rejects every express order.

**Fix.** Skip the `account` fieldset in the classic custom-field detection. Those are core's conditional account-creation fields, not merchant custom fields, and express flows can't collect them: express checkout relies on auto-generated credentials (`is_account_creation_possible()` already hides the buttons when guest checkout is disabled and generation is off).

## Testing instructions

1. In **WooCommerce → Settings → Accounts & Privacy**, uncheck **"When creating an account, automatically generate an account password"**. Keep guest checkout enabled.
2. Enable Apple Pay/Google Pay in the Stripe settings.
3. Add any product to the cart and pay with an express checkout button (product page, cart, or checkout).
4. Before this fix: the payment sheet errors with "Create account password is a required field." After: the payment completes.
5. Regression: with a required custom checkout field added by a plugin/snippet, express checkout from a product page still directs the buyer to the checkout page (unchanged #5640 behavior).

---

-   [x] Covered with tests

### Changelog entry

Fix - Allow express checkout payments when automatic account password generation is disabled

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
@@ -278,6 +278,16 @@ class WC_Stripe_Express_Checkout_Custom_Fields {
 			}
 
 			foreach ( $all_fields as $fieldset => $fields ) {
+				// Core's account fieldset (account_username/account_password when
+				// auto-generation is off) is marked required unconditionally, but core
+				// only enforces it when an account is actually being created. The
+				// express sheet can never collect these values — express flows rely on
+				// auto-generated credentials — so treating them as required custom
+				// fields would block every express payment.
+				if ( 'account' === $fieldset ) {
+					continue;
+				}
+
 				foreach ( $fields as $field_key => $field ) {
 					if ( in_array( $field_key, $standard_checkout_fields, true ) ) {
 						continue;
--- a/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
@@ -278,12 +278,8 @@ class WC_Stripe_Express_Checkout_Custom_Fields {
 			}
 
 			foreach ( $all_fields as $fieldset => $fields ) {
-				// Core's account fieldset (account_username/account_password when
-				// auto-generation is off) is marked required unconditionally, but core
-				// only enforces it when an account is actually being created. The
-				// express sheet can never collect these values — express flows rely on
-				// auto-generated credentials — so treating them as required custom
-				// fields would block every express payment.
+				// Core only enforces account fields when creating an account; the
+				// express sheet can't collect them, so never treat them as required.
 				if ( 'account' === $fieldset ) {
 					continue;
 				}
--- a/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
@@ -267,7 +267,8 @@ class WC_Stripe_Express_Checkout_Custom_Fields {
 			}
 		}
 
-		// Classic checkout page
+		// Classic checkout page. This branch also feeds the cart/product merge below,
+		// where the legacy field definitions apply even on block-based stores.
 		if ( is_checkout() || 'classic' === $context ) {
 			$classic_custom_checkout_fields = [];
 			$standard_checkout_fields       = $this->get_standard_checkout_fields();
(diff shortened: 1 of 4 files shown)


includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
```diff
@@ -278,6 +278,12 @@ public function get_custom_checkout_fields( $context = '' ) {
 			}
 
 			foreach ( $all_fields as $fieldset => $fields ) {
+				// Core only enforces account fields when creating an account; the
+				// express sheet can't collect them, so never treat them as required.
+				if ( 'account' === $fieldset ) {
```
jaclync, line 285: non-blocking ❓ : skipping the whole `account` fieldset (rather than excluding just the core `account_username`/`account_password` keys) means a field a (third-party) plugin injects into `$fields['account']` via `woocommerce_checkout_fields` is now dropped here. Just want to confirm that this is an intentional tradeoff?

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
jaclync: non-blocking observation: this code path reads like it's specific to classic checkout, but this PR fix applies to both blocks & classic checkout from testing before & after. It might help future readers if the code, or the docblock at least, matches the actual code paths.

jaclync (approved):
Thanks for fixing this quickly! :shipit: 

Verified: ECE payment with classic/blocks cart checkout + account required/optional + logged-in/out + Amazon Pay/Link, one pre-existing issue with Amazon Pay below ✅ 

Non-blocking notes:

1. When an account is required (`Enable guest checkout (recommended)` disabled in Settings > Account & Privacy), some pre-existing issues:
  - Blocks checkout (cart), logged out: the login alert was only shown after the EC form was complete, regardless of the EC payment method.
  - Classic checkout (cart), logged out: after clicking on Amazon Pay, the modal appeared shortly then disappeared. For other EC methods like Link/Apple Pay/Google Pay, the login alert was shown first.

https://github.com/user-attachments/assets/4c08658e-52b3-4c51-9724-a8ba9d770f84

2. E2E coverage: the current ECE E2E specs seem to only assert the Link button renders and the wallet modal opens, but there's no payment path verification to catch payment issues like this. The new PHPUnit tests cover this specific issue (I verified they fail with the fix reverted), but maybe E2E coverage on the EC payment paths + classic/blocks checkout + logged-in/out + account required/optional + payment methods is worth adding to prevent future regressions.

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
wjrosa: Intentional — the sheet can't collect any account-fieldset field (third-party included), so enforcing them would recreate the hard block this PR fixes; clarified the comment in bc98a167e.

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
wjrosa: Good catch — the branch also feeds the cart/product merge path, which is why block stores are affected too; noted in the comment in bc98a167e.

wjrosa:
Thanks for the thorough matrix testing! Both non-blocking notes are captured as follow-ups: STRIPE-1361 (login-required flow inconsistencies, with your recording) and STRIPE-1362 (E2E coverage for the express payment paths).

PR #5804, opened 2026-08-13
Restore Adaptive Pricing after amount mismatch update

diegocurbelo:
Fixes #<github_issue_id>

This PR adds a one-shot update migration that restores Adaptive Pricing for stores that got it auto-disabled because of incompatibility errors with 3rd party plugins in version 10.8.x.

At a high level, the migration:
1. Checks whether `wc_stripe_restore_adaptive_pricing_after_amount_mismatch_migration_ran` is already set.
2. Uses `wc_stripe_adaptive_pricing_session_amount_mismatch_detected` to distinguish an automatic safety disable from a merchant’s decision to disable Adaptive Pricing.
3. Sets `adaptive_pricing` to _yes_ in `woocommerce_stripe_settings`.
4. Confirms that the updated setting was persisted.
5. Deletes the `wc_stripe_adaptive_pricing_session_amount_mismatch_detected` flag.
6. Sets the `wc_stripe_restore_adaptive_pricing_after_amount_mismatch_migration_ran` migration flag.
7. Logs an informational message after the AP restoration completes.

## Testing instructions

1. Install a build from this branch (and make sure Logging is enabled)
2. Disable _Adaptive Pricing_ in the Stripe settings
3. Simulate an affected 10.8.5 installation:
    ```
     wp option delete wc_stripe_restore_adaptive_pricing_after_amount_mismatch_migration_ran
     wp option update wc_stripe_adaptive_pricing_session_amount_mismatch_detected yes
     wp option update wc_stripe_version 10.8.5
    ```
4. Load a WordPress admin page to trigger the update routine
5. Confirm that:
    - Adaptive Pricing is enabled
    - `wc_stripe_adaptive_pricing_session_amount_mismatch_detected` no longer exists
    - `wc_stripe_restore_adaptive_pricing_after_amount_mismatch_migration_ran` is yes
    - The Stripe logs contain the Adaptive Pricing restoration info message
6. Disable _Adaptive Pricing_ again
7. Simulated update without the mismatch marker:
    ```
     wp option delete wc_stripe_restore_adaptive_pricing_after_amount_mismatch_migration_ran
     wp option delete wc_stripe_adaptive_pricing_session_amount_mismatch_detected
     wp option update wc_stripe_version 10.8.5
    ```
8. Load a WordPress admin page
9. Confirm Adaptive Pricing remains disabled and the migration flag is set to yes

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php
```diff
@@ -0,0 +1,69 @@
...
+			return;
+		}
+
+		$stripe          = WC_Stripe::get_instance();
+		$stripe_settings = $stripe->get_settings();
+
+		$stripe_settings['adaptive_pricing'] = 'yes';
+		$stripe->update_settings( $stripe_settings );
+
+		// Keep the safety marker in place unless the canonical settings read confirms the write.
```
daledupreez: Nit: I think we could describe this better.
```suggestion
		// Get the updated settings data.
		// If adaptive pricing is still disabled, return _without_ setting the migration flag so we can run at a later time.
```

includes/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php
```diff
@@ -0,0 +1,69 @@
...
+		if ( 'yes' !== ( $stripe->get_settings()['adaptive_pricing'] ?? 'no' ) ) {
+			return;
+		}
+
+		// A failed marker deletion must remain retryable on a later plugin update.
+		if ( ! delete_option( self::AMOUNT_MISMATCH_OPTION ) ) {
```
daledupreez, line 58: Can we add a secondary check here to make sure the option still exists? I think `delete_option()` will return false when the option doesn't exist, which can happen for this code path because the option is not set to autoload and `delete_option()` queries the database.
```suggestion
		if ( ! delete_option( self::AMOUNT_MISMATCH_OPTION ) ) {
			// If the option still exists, then return early.
			if ( false !== get_option( self::AMOUNT_MISMATCH_OPTION, false ) ) ) {
				return;
			}
			// Otherwise the option no longer exists, mark the migration as complete.
```

includes/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php
```diff
@@ -0,0 +1,69 @@
...
+			return;
+		}
+
+		update_option( self::MIGRATION_FLAG_OPTION, 'yes' );
+
+		WC_Stripe_Logger::info(
```
daledupreez: I think this should be an `error` so we are guaranteed to log the update.
```suggestion
		WC_Stripe_Logger::error(
```

daledupreez: Hmm... However, if OCS is not enabled, the setting may not take effect.

tests/phpunit/admin/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update-test.php
```diff
@@ -0,0 +1,198 @@
...
+		return [
+			'mismatch disable is restored on upgrade'        => [
+				'10.8.5',
+				'yes',
+				null,
+				'no',
+				'yes',
+				false,
+				'yes',
+				true,
```
daledupreez: Can you please update these tests to use an associative array so the values have descriptive keys and it's _much_ easier to see what is being tested for each test case?
```suggestion
				'previous_version'          => '10.8.5',
				'mismatch_marker'           => 'yes',
				'migration_flag'            => null,
				'adaptive_pricing'          => 'no',
				'expected_adaptive_pricing' => 'yes',
				'expected_marker'           => false,
				'expected_migration_flag'   => 'yes',
				'expect_log'                => true,
```

includes/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php
```diff
@@ -0,0 +1,69 @@
...
+ * Restore Adaptive Pricing for stores affected by the amount-mismatch safety disable.
+ *
+ * @since 10.9.0
+ */
+class WC_Stripe_Restore_Adaptive_Pricing_After_Amount_Mismatch_Update {
+
+	/**
+	 * Option flag used to ensure the migration only runs once.
+	 */
+	private const MIGRATION_FLAG_OPTION = 'wc_stripe_restore_adaptive_pricing_after_amount_mismatch_migration_ran';
```
daledupreez: Not blocking, but I think we can use a shorter option name without losing too much meaning.
```suggestion
	private const MIGRATION_FLAG_OPTION = 'wc_stripe_restored_ap_after_amount_mismatch';
```

daledupreez (approved):
This is working as described, and tests well too. I have some minor feedback on some small readability improvements. I think the two things we should do before merging are making the test data provider easier to understand and adding a check for the option being removed.

includes/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php
daledupreez: I went ahead and renamed the option to `wc_stripe_restored_adaptive_pricing_after_amount_mismatch`.

includes/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php
```diff
@@ -0,0 +1,69 @@
...
+
+		if ( 'yes' !== get_option( self::AMOUNT_MISMATCH_OPTION, 'no' ) ) {
+			update_option( self::MIGRATION_FLAG_OPTION, 'yes' );
+			return;
+		}
+
+		$stripe          = WC_Stripe::get_instance();
+		$stripe_settings = $stripe->get_settings();
+
+		$stripe_settings['adaptive_pricing'] = 'yes';
```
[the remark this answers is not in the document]
daledupreez: This is a valid issue, especially if we start ignoring the value of the `wc_stripe_adaptive_pricing_session_amount_mismatch_detected` option in the settings UI as we are doing in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5806. I _think_ that is something we should handle in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5806 by marking the migration as complete if the merchant toggles the Adaptive Pricing setting before this migration has run.

includes/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php
```diff
@@ -0,0 +1,75 @@
...
+		$stripe->update_settings( $stripe_settings );
+
+		// Get the updated settings data.
+		// If adaptive pricing is still disabled, return _without_ setting the migration flag so we can run at a later time.
```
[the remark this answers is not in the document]
daledupreez: Simplified in 67b3bbbbdaa445a2696c97bba2fb5c3e4872459d.

includes/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php
```diff
@@ -0,0 +1,75 @@
...
+		if ( 'yes' !== ( $stripe->get_settings()['adaptive_pricing'] ?? 'no' ) ) {
+			return;
+		}
+
+		// A failed marker deletion must remain retryable on a later plugin update.
+		if ( ! delete_option( self::AMOUNT_MISMATCH_OPTION ) ) {
+			// If the option still exists, then return early.
+			if ( false !== get_option( self::AMOUNT_MISMATCH_OPTION, false ) ) ) {
```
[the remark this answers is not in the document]
daledupreez: Fixed in 09924d13a9f74258c21b8abdbe94e9e3930eaee8.

tests/phpunit/admin/migrations/class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update-test.php
daledupreez: Implemented in 197b73699e317f541d1da85b43f027b5091553fd.

PR #5806, opened 2026-08-13
Temporarily fallback to non Checkout Session flow after an amount mismatch

diegocurbelo:
Fixes STRIPE-1312

As a follow-up of #5799 and the CS/AP `WC_Stripe_Checkout_Session_Context::validate_for_order()` incompatibility issues with 3rd party plugins from `10.8.x` this PR adds a temporally fallback to the payment intent flow when a mismatch/incompatibility is detected to avoid a second retry error.

#5799 fixed the underlying incompatibility problem, but we discussed including this fallback as a preemptive workaround.

## Testing instructions

Code review should be enough

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-checkout-session-context.php
+++ b/includes/class-wc-stripe-checkout-session-context.php
@@ -392,6 +392,7 @@ class WC_Stripe_Checkout_Session_Context {
 			[
 				'checkout_session_id' => $session_id,
 				'order_id'            => $order->get_id(),
+				'customer_id'         => $order->get_customer_id(),
 				'session_amount'      => $context['amount'],
 				'order_amount'        => $order_amount,
 				'session_currency'    => $context['currency'],
--- a/includes/class-wc-stripe-checkout-session-context.php
+++ b/includes/class-wc-stripe-checkout-session-context.php
@@ -387,7 +387,7 @@ class WC_Stripe_Checkout_Session_Context {
 
 		self::delete_context( $session_id );
 
-		WC_Stripe_Logger::warning(
+		WC_Stripe_Logger::error(
 			'Checkout Session amount mismatch; Session invalidated.',
 			[
 				'checkout_session_id' => $session_id,
--- a/includes/class-wc-stripe-checkout-session-context.php
+++ b/includes/class-wc-stripe-checkout-session-context.php
@@ -23,11 +23,12 @@ class WC_Stripe_Checkout_Session_Context {
 	private const MUTATION_LOCK_OPTION_PREFIX = 'wc_stripe_checkout_session_lock_';
 
 	/**
-	 * Stable marker for support/admin code to distinguish an automatic safety disable from a manual setting change.
+	 * WC session key flagging that this customer hit a Checkout Session amount mismatch, so their
+	 * later checkouts use the standard intent flow instead of Adaptive Pricing.
 	 *
 	 * @var string
 	 */
-	private const ADAPTIVE_PRICING_AMOUNT_MISMATCH_OPTION = 'wc_stripe_adaptive_pricing_session_amount_mismatch_detected';
+	private const AMOUNT_MISMATCH_SESSION_KEY = 'wc_stripe_checkout_session_amount_mismatch';
 
 	/**
 	 * WooCommerce rewrites the session customer ID when checkout creates an account.
@@ -86,12 +87,16 @@ class WC_Stripe_Checkout_Session_Context {
 	}
 
 	/**
-	 * Check if an amount mismatch was detected.
+	 * Check if the current customer session previously hit a Checkout Session amount mismatch.
 	 *
 	 * @return bool
 	 */
 	public static function was_amount_mismatch_detected(): bool {
-		return 'yes' === get_option( self::ADAPTIVE_PRICING_AMOUNT_MISMATCH_OPTION, 'no' );
+		if ( ! WC()->session || ! is_callable( [ WC()->session, 'get' ] ) ) {
+			return false;
+		}
+
+		return 'yes' === WC()->session->get( self::AMOUNT_MISMATCH_SESSION_KEY );
 	}
 
 	/**
@@ -272,7 +277,7 @@ class WC_Stripe_Checkout_Session_Context {
 			(int) ( $context['amount'] ?? 0 ) !== $order_amount
 			|| strtolower( (string) ( $context['currency'] ?? '' ) ) !== $order_currency
 		) {
-			self::disable_adaptive_pricing_after_amount_mismatch( $session_id, $order, $context, $order_amount, $order_currency );
+			self::invalidate_session_after_amount_mismatch( $session_id, $order, $context, $order_amount, $order_currency );
 
 			throw new Exception( __( 'The payment amount no longer matches the order total. Please refresh the page and try again.', 'woocommerce-gateway-stripe' ) );
 		}
@@ -366,7 +371,7 @@ class WC_Stripe_Checkout_Session_Context {
 
 	/**
 	 * A stored Checkout Session total that disagrees with the order means the store may have an
-	 * incompatible totals integration, so future shoppers should use the standard intent flow.
+	 * incompatible totals integration; invalidate the session to use the standard intent flow.
 	 *
 	 * @param string   $session_id Stripe Checkout Session ID.
 	 * @param WC_Order $order WooCommerce order.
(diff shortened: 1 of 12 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1542,7 +1542,7 @@ public function process_payment( $order_id, $retry = true, $force_save_source =
 			}
 		}
 
-		if ( is_string( $checkout_session_id ) && ! empty( $checkout_session_id ) && ! WC_Stripe_Checkout_Session_Context::was_amount_mismatch_detected() ) {
+		if ( is_string( $checkout_session_id ) && ! empty( $checkout_session_id ) ) {
```
daledupreez, line 1545: Shouldn't we keep this condition? It will only ever trigger within a shopper session, so it feels like we should keep the guard.

includes/class-wc-stripe-checkout-session-context.php
```diff
@@ -375,19 +380,15 @@ private static function validate_owner( array $context ): void {
...
-			$stripe_settings['adaptive_pricing'] = 'no';
-			WC_Stripe_Helper::update_main_stripe_settings( $stripe_settings );
+	private static function invalidate_session_after_amount_mismatch( string $session_id, WC_Order $order, array $context, $order_amount, string $order_currency ): void {
+		if ( WC()->session && is_callable( [ WC()->session, 'set' ] ) ) {
+			WC()->session->set( self::AMOUNT_MISMATCH_SESSION_KEY, 'yes' );
 		}
 
 		self::delete_context( $session_id );
 
 		WC_Stripe_Logger::warning(
```
daledupreez: Is it worth elevating this to an `error` so we are guaranteed to have these events logged?

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -13,6 +13,13 @@ class WC_Stripe_UPE_Payment_Gateway_Test extends WC_Mock_Stripe_API_Unit_Test_Ca
 	 */
 	private const ADAPTIVE_PRICING_AMOUNT_MISMATCH_OPTION = 'wc_stripe_adaptive_pricing_session_amount_mismatch_detected';
 
+	/**
+	 * Asserts the exact persisted WC session flag without exposing a production setter only for tests.
+	 *
+	 * @var string
+	 */
+	private const AMOUNT_MISMATCH_SESSION_KEY = 'wc_stripe_checkout_session_amount_mismatch';
```
daledupreez: Not blocking, but we can also use `WC_Stripe_Test_Helper::get_class_const_value()` instead of replicating the value in a way that could drift.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -2163,20 +2174,25 @@ public function test_process_payment_with_checkout_session_rejects_amount_mismat
...
+			$adaptive_pricing_setting   = WC_Stripe_Helper::get_settings( null, 'adaptive_pricing' );
+			$customer_mismatch_detected = WC_Stripe_Checkout_Session_Context::was_amount_mismatch_detected();
+			$legacy_marker_option       = get_option( self::ADAPTIVE_PRICING_AMOUNT_MISMATCH_OPTION, 'no' );
+			$remaining_context          = WC_Stripe_Checkout_Session_Context::get_context( $session_id );
 		} finally {
 			WC_Stripe_Checkout_Session_Context::delete_context( $session_id );
 			unset( $_POST['wc_stripe_checkout_session_id'], $_POST['payment_method'], $_POST['wc-stripe-payment-method'] );
 			WC_Stripe_Helper::update_main_stripe_settings( $original_stripe_settings );
 			delete_option( self::ADAPTIVE_PRICING_AMOUNT_MISMATCH_OPTION );
+			WC()->session->set( self::AMOUNT_MISMATCH_SESSION_KEY, null );
```
daledupreez: Is WC()->session guaranteed to be valid?

tests/phpunit/class-wc-stripe-helper-test.php
```diff
@@ -10,6 +10,13 @@
  * WC_Stripe_Helper_Test class.
  */
 class WC_Stripe_Helper_Test extends WC_Mock_Stripe_API_Unit_Test_Case {
+	/**
+	 * Mirrors the WC session key WC_Stripe_Checkout_Session_Context uses to flag a customer amount mismatch.
+	 *
+	 * @var string
+	 */
+	private const AMOUNT_MISMATCH_SESSION_KEY = 'wc_stripe_checkout_session_amount_mismatch';
```
daledupreez: I think we should also use the helper for this code location.

daledupreez (commented):
I have some feedback from code inspection.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
diegocurbelo: There the guard makes things a bit worse for the customer. The payload carries no `wc-stripe-payment-method / wc-stripe-confirmation-token`, so `process_payment_with_payment_method()` fails with a generic "There was an error processing the payment: …" and marks the order failed. Without the guard, the deleted context would make `validate_for_order()` return the actionable "please refresh the page and try again" on an order that's still pending.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
daledupreez: Addressed in c87011f8339736ed9d4e5f96a703ba3d86ecd222.

tests/phpunit/class-wc-stripe-helper-test.php
daledupreez: Addressed in c87011f8339736ed9d4e5f96a703ba3d86ecd222.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
daledupreez: Thanks for clarifying. That makes sense to me.

includes/class-wc-stripe-checkout-session-context.php
```diff
@@ -375,19 +380,15 @@ private static function validate_owner( array $context ): void {
...
+		if ( WC()->session && is_callable( [ WC()->session, 'set' ] ) ) {
+			WC()->session->set( self::AMOUNT_MISMATCH_SESSION_KEY, 'yes' );
 		}
 
 		self::delete_context( $session_id );
 
-		WC_Stripe_Logger::warning(
-			'Adaptive Pricing disabled after Checkout Session amount mismatch.',
+		WC_Stripe_Logger::error(
+			'Checkout Session amount mismatch; Session invalidated.',
```
daledupreez, line 391: Is it worth logging the WooCommerce customer ID as well? It will most likely be 0 for guests, but I think it's worth considering.

daledupreez (approved):
The changes look OK to me based on code inspection. I have one open question, but we can make a decision on that without another round of review.

includes/class-wc-stripe-checkout-session-context.php
diegocurbelo: Good idea, added in [e385946](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5806/commits/e385946b4cc92ce81c91ea26d367e84fe2e4081b)

wjrosa (approved):
Looks good to me! https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5806/changes#r3785401178 looks relevant, but I understand if we need to keep it

includes/class-wc-stripe-checkout-session-context.php
```diff
@@ -375,22 +380,19 @@ private static function validate_owner( array $context ): void {
...
 			[
 				'checkout_session_id' => $session_id,
 				'order_id'            => $order->get_id(),
+				'customer_id'         => $order->get_customer_id(),
```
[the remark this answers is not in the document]
diegocurbelo: Keeping this as-is.

`$order->get_customer_id()` returns the WooCommerce user ID — an integer, `0` for guests. It isn't a Stripe customer ID.

The same log entry already includes `order_id`, which resolves to that customer in one click in wp-admin. So this field doesn't widen exposure, and a keyed pseudonymous replacement wouldn't narrow it — `order_id` would still identify the customer just as directly. It would only make support correlation harder, for no privacy gain.

diegocurbelo:
> Looks good to me! https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5806/changes#r3785401178 looks relevant, but I understand if we need to keep it

I added the customer_id it as a result of this thread: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5806#pullrequestreview-4938284819, I think it makes sense to keep it as it does not log any PII and the customer id is visible on the order details page anyway.

tests/phpunit/class-wc-stripe-test.php
```diff
@@ -929,6 +929,161 @@ public function provide_test_maybe_reconfigure_webhooks_after_adaptive_pricing_e
...
+			do_action( 'update_option_' . WC_Stripe::SETTINGS_OPTION_NAME, $old_value, $new_value, WC_Stripe::SETTINGS_OPTION_NAME );
+		} finally {
+			remove_filter( 'pre_update_option_' . $flag_option, $mark_migration_complete_spy );
+		}
+
+		$this->assertSame( $expect_marked ? 1 : 0, $mark_migration_complete_calls );
+		$this->assertSame(
+			$expect_marked ? 'yes' : ( $flag_before ?? false ),
+			get_option( $flag_option )
+		);
```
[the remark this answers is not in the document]
diegocurbelo: Added in [2ac622c](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5806/commits/2ac622c4bdb7dc566db31ed02717f94fa92b9eb6)

includes/class-wc-stripe-checkout-session-context.php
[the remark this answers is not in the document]
diegocurbelo: Keeping `customer_id`, final answer.

Three things worth weighing:

1. This field was added at a reviewer's explicit request in the thread directly above ([#discussion](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5806#discussion_r0) — the "Is it worth logging the WooCommerce customer ID as well?" thread).
2. It isn't a new class of data for this plugin `includes/payment-tokens/class-wc-stripe-payment-tokens.php:594` already logs the WordPress user ID at error level, as do lines `308` and `458` with the WooCommerce customer ID.
3. The proposed remedy is self-defeating. A keyed pseudonymous identifier exists precisely to provide stable correlation across log entries — the property the objection is based on. It would preserve the correlation while adding key management and making support lookups harder, for no privacy gain.

These logs are server-side, manage_woocommerce-gated, and the entry already carries order_id.

PR #5807, opened 2026-08-14
Improve the E2E env setup.sh script plugin dependency fetching

diegocurbelo:
Fixes #<github_issue_id>

The E2E setup.sh hard-required GITHUB_TOKEN to install Woo Subscriptions and Woo Pre-Orders, which live in private repositories, without a token there was no way to bring up the local Docker E2E environment; and the failure was misleading, it surfaced as a corrupt-plugin error rather than an auth error

This PR updates the local Docker setup script to support additional plugin packages sources (CI behavior is unchanged).
A new `fetch_plugin_zip <repo> <zip>` helper resolves each plugin from the first available source:
1. GITHUB_TOKEN - the existing curl flow against the GitHub Releases API
2. The GH CLI - `gh release download`, used when no token is set and `gh auth status` is clean
3. tests/e2e/deps/<zip> - a copy placed there by hand (or left behind by an earlier run)

When none of the three can supply a plugin, a preflight check aborts before starting the Docker containers

**_Trade-offs_**

Failure handling is deliberately asymmetric. A token that fails to authenticate aborts instead of falling through, so an expired token surfaces as an auth error rather than a silent downgrade to whatever older zip happens to be sitting in `deps/`

A gh failure is treated as "this source is unavailable" and falls through

`deps/` is a fallback, not a cache. Whenever a token or gh is available the latest release is still downloaded on every run, preserving the previous always-latest behaviour. The stored copies come into play when gh cannot reach GitHub, or when you have neither a token nor a gh login

**_Backward compatibility_**

CI flow is unaffected, `run-e2e-tests.yml` and `e2e-tests.yml` both set GITHUB_TOKEN so runners take source 1

## Testing instructions

Verify the E2E workflows of this PR run successfully

**GH Token (the CI path)** (if you have a valid token, otherwise this is the GH action flow, see below)
1. Set a valid `GITHUB_TOKEN` in `tests/e2e/config/local.env`
2. Run `rm -rf tests/e2e/deps && npm run test:e2e-setup`
3. Verify that both plugins download and install
    - the summary banner reports their versions

**GitHub CLI**
1. Clear `GITHUB_TOKEN` in `tests/e2e/config/local.env`
2. Check that `gh auth status` is valid (run `gh auth login` if needed)
3. Run `rm -rf tests/e2e/deps && npm run test:e2e-setup`
4. Verify the log shows `Fetching latest version with the gh CLI`
5. Verify that both plugins download and install
    - the summary banner reports their versions

**Local zips**
1. Clear `GITHUB_TOKEN`, log out of `gh cli`
2. Put both the `woocommerce-subscriptions.zip` and `woocommerce-pre-orders.zip` in the `tests/e2e/deps` folder
3. Run `rm -rf tests/e2e/deps && npm run test:e2e-setup`
4. Verify the log shows `Using tests/e2e/deps/<zip>`
5. Verify that both plugins download and install
    - the summary banner reports their versions

**No source available**
1. Clear `GITHUB_TOKEN`, log out of `gh cli`
2. Run `rm -rf tests/e2e/deps && npm run test:e2e-setup`
3. Verify that the setup aborts without starting the containers

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


tests/e2e/bin/setup.sh
```diff
@@ -21,19 +21,105 @@ DEBUG=false
...
+
 cd "$CWD"
 
 check_dep 'docker'
 check_dep 'curl'
 check_dep 'jq'
 check_dep 'php'
 
+gh_authenticated() {
+	command -v gh > /dev/null 2>&1 && gh auth status > /dev/null 2>&1
```
wjrosa: `gh auth status` checks every host configured in GH and exits non-zero if any of them fail. So for anyone who also has a GHE host configured (like many of us at a8c), an unreachable GHE host disables this fallback even though github.com is logged in and downloads would work. I hit this locally.

Suggest checking only the relevant host: `gh auth status --hostname github.com`.

tests/e2e/bin/setup.sh
```diff
@@ -21,19 +21,105 @@ DEBUG=false
...
+	if gh_authenticated; then
+		echo " - Fetching latest version with the gh CLI"
+
+		if redirect_output gh release download --repo "$repo" --pattern '*.zip' --output "$tmp" --clobber; then
+			mv "$tmp" "$dest"
+			return
+		fi
+
+		rm -f "$tmp"
+		echo " - gh could not download $zip, falling back to tests/e2e/deps/"
```
wjrosa: When the gh download fails, this silently falls back to whatever zip is already in `deps/`, which can be months old (gh's actual error only goes to `e2e-setup.log`). That contradicts the README's "The latest release is fetched on every run whenever a token or the gh CLI is available" and its no-silent-downgrade NOTE, which the code only implements for the token path. Suggest aborting here like the token path does, or making the downgrade loud.

tests/e2e/bin/setup.sh
```diff
@@ -21,19 +21,105 @@ DEBUG=false
...
+	local repo=$1
+	local zip=$2
+	local dest="$DEPS_DIR/$zip"
+	# Downloads land on a temp path so a failed fetch cannot truncate the cached
+	# zip we may still need to fall back to.
+	local tmp="$dest.part"
+
+	mkdir -p "$DEPS_DIR"
+
+	if [[ -n "$GITHUB_TOKEN" ]]; then
```
wjrosa, line 58: This also picks up a `GITHUB_TOKEN` exported in the developer's shell (common for npm/act/homebrew tooling), not just one set in `local.env`. If that inherited token cannot access the private repos, the script aborts at the curl step without ever trying gh or a valid zip already in `deps/`, and only after the full environment build, since the preflight also accepts any non-empty token. Consider falling through to the next sources when the token cannot resolve the release.

tests/e2e/bin/setup.sh
```diff
@@ -21,19 +21,105 @@ DEBUG=false
...
+}
+
 if ! docker info > /dev/null 2>&1; then
 	echo
 	error "Docker is not running, please start it and try again."
 	exit 1
 fi
 
+# Fail before the environment is built, rather than minutes later at the install step.
+if [[ -z "$GITHUB_TOKEN" ]] && ! gh_authenticated; then
```
wjrosa: This preflight only catches the empty-credentials case. An expired/invalid `GITHUB_TOKEN`, or a gh login without access to the private repos, passes it and still dies minutes later at the install step, which is exactly what the comment above says it prevents. An alternative that covers every source: run the two `fetch_plugin_zip` calls before starting the containers, then this block can go away.

tests/e2e/README.md
```diff
@@ -36,6 +36,19 @@ We use [Playwright](https://playwright.dev/) as our test runner.
...
+**Woo Subscriptions and Woo Pre-Orders**
+
+The local Docker setup (`npm run test:e2e-setup` with no `--base_url`) installs both plugins, which live in private repositories. It gets them from the first available of:
+
+1. `GITHUB_TOKEN` in `local.env` — a token with access to both repositories.
+2. The [GitHub CLI](https://cli.github.com/), if you are logged in (`gh auth login`).
+3. `woocommerce-subscriptions.zip` and `woocommerce-pre-orders.zip` placed in `/tests/e2e/deps/`.
+
+Downloads are kept in /tests/e2e/deps/ (gitignored).
+The latest release is fetched on every run whenever a token or the gh CLI is available, so these copies act as a fallback rather than a cache. This is also how you pin specific plugin versions for a run.
```
wjrosa: Pinning does not work as described when a token or gh login is present: `fetch_plugin_zip` downloads the latest release and `mv`s it over the pinned zip, so the run tests the latest version and the pinned copy is gone. Either have setup.sh skip the download when the zip already exists, or document that pinning requires unsetting the token and logging out of gh.

tests/e2e/bin/setup.sh
```diff
@@ -21,19 +21,105 @@ DEBUG=false
...
+			return
+		fi
+
+		rm -f "$tmp"
+		echo " - gh could not download $zip, falling back to tests/e2e/deps/"
+	fi
+
+	if [[ -f "$dest" ]]; then
+		echo " - Using tests/e2e/deps/$zip"
+		return
```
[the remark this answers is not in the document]
diegocurbelo: Both releases publish exactly one asset, so neither path can pick the wrong archive — and `.assets[0].id` is pre-existing code this PR only moved. On the `gh` path, `--output` refusing multiple assets means a second zip would fail and fall through rather than fetch the wrong file; pinning `--pattern "$zip"` would instead break if the asset name ever gains a version suffix. Not taking that one.

On validation: downloads only land in place after `curl -sf` succeeds, and both zips are now resolved before the containers are built, so a bad file fails up front. I've changed the fallback check from `-f` to `-s` rather than making `unzip` a new required dependency.

tests/e2e/bin/setup.sh
diegocurbelo: Rather than scoping it with `--hostname`, I dropped the probe entirely. Moving the fetches ahead of the container build (your other comment) left `gh_authenticated` with one caller, and there the download attempt is the authoritative test — `command -v gh`, then let `gh release download` succeed or fall through. No auth state to interpret, and one less round-trip per plugin.

tests/e2e/bin/setup.sh
diegocurbelo: Fair point on the inconsistency, but we don't want to abort here: a gh login isn't proof of access to these private repos, so aborting would stop anyone logged in with an account outside the org from using their own `deps/` zips without logging out of gh entirely — which is the case this PR exists to serve.

Fixed it the other way instead: all three sources now behave the same. One that can't deliver says why and moves on (the gh line points at `e2e-setup.log` for the actual error), and only "no source could supply it" is fatal.

tests/e2e/bin/setup.sh
diegocurbelo: Fixed in [fa7ef93](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5807/commits/fa7ef93e3448c54d50ba359beb6fe95d1b9ab755)

tests/e2e/bin/setup.sh
diegocurbelo: Fixed in [fa7ef93](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5807/commits/fa7ef93e3448c54d50ba359beb6fe95d1b9ab755)

tests/e2e/README.md
diegocurbelo: Went with the documentation fix rather than skip-if-exists, to keep the always-latest behavior that CI relies on.

`deps/` is now described as a fallback that's only used when neither other source can deliver

tests/e2e/bin/setup.sh
```diff
@@ -4,7 +4,10 @@ set -e
 . ./tests/e2e/bin/common.sh
 
 if [[ -f "$E2E_ROOT/config/local.env" ]]; then
-	. "$E2E_ROOT/config/local.env"
+	# Unreplaced <placeholder> values from local.env.example are not valid shell:
+	# sourcing one aborts the rest of the file, silently dropping every variable
+	# below it. Blank them out so only the values actually filled in take effect.
+	eval "$(sed -E 's/=<[^>]*>[[:space:]]*$/=/' "$E2E_ROOT/config/local.env")"
```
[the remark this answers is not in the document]
diegocurbelo: The suggested fix doesn't work: `source <(...)` sets nothing on bash 3.2

`local.env` was already executed as shell by the `.` this replaced, so there's no new capability or trust boundary — same file, same execution, with unfilled placeholders neutralised first.

Blanking rather than deleting is deliberate: a placeholder left in `local.env` should read as "unset" regardless of ambient shell state, which is what closes the inherited-`GITHUB_TOKEN` thread above.

wjrosa (approved):
Looks good to me 👍

PR #5809, opened 2026-08-14
Ensure Adaptive Pricing settings changes bypass amount mismatch migration

daledupreez:
Related to #5804
Related to #5806

This PR addresses the problem raised in [this review comment on #5804](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5804#discussion_r3776360043), where we need to ensure that merchant modifications to the Adaptive Pricing settings are not overridden after the migration does not succeed during an upgrade. With only #5804, we prevent merchants from toggling the value after a mismatch was detected, but #5806 removes the UI restriction, so merchants can update the Adaptive Pricing settings when a migration didn't succeed. This PR handles that edge case by marking the migration as complete when such a manual update occurs.

The PR does that by listening to the `option_update_woocommerce_stripe_settings` action and detecting when the `adaptive_pricing` key is modified, and when that happens, checking if we need to complete the migration. If we do need to complete the migration, we remove the amount mismatch option and mark the migration as complete.

## Testing instructions

Get a site connected to Stripe, and then mimic the mismatch migration failure mode by running the following commands:
```
wp option update wc_stripe_adaptive_pricing_session_amount_mismatch_detected yes
wp option delete wc_stripe_restored_adaptive_pricing_after_amount_mismatch
```

Navigate to the Stripe Settings UI and modify the Adaptive Pricing setting, and ensure the settings got saved.

Verify that the migration was run by verifying that the amount mismatch option was deleted and the migration option was run:
 * Verify that the migration option is set to `'yes'`:
```
wp option get wc_stripe_restored_adaptive_pricing_after_amount_mismatch
```
* Verify that the amount mismatch option was deleted:
```
wp option get wc_stripe_adaptive_pricing_session_amount_mismatch_detected
```
This should return something like the following:
```
Error: Could not get 'wc_stripe_adaptive_pricing_session_amount_mismatch_detected' option. Does it exist?
```

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -107,6 +107,7 @@ xxxx-xx-xx - version 10.9.0
 * Fix - Allow express checkout payments when automatic account password generation is disabled
 * Tweak - Add PHPDoc for some deprecated hooks
 * Dev - Remove the deprecated @woocommerce/settings npm package; settings are still read from the wc-settings script WooCommerce provides at runtime
+* Fix - Ensure that manual Adaptive Pricing changes after upgrades get precedence
 
 2026-08-05 - version 10.8.5
 * Update - Improve webhook handling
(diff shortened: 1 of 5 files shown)


diegocurbelo (approved):
LGTM 👍🏼 

We might want to use `WC_Stripe::get_instance()` methods instead if the deprecated `WC_Stripe_Helper`

daledupreez:
@diegocurbelo, because we're wanting this in 10.9.0 where those accessors don't exist, we need to get this version of the code across to 10.9.0. We can either do that as part of the port across to `release/10.9.0` or as a follow-up PR -- I don't have a strong opinion on that either way.

PR #5810, opened 2026-08-14
Improve handling of expired checkout sessions

daledupreez:
Fixes STRIPE-1330

A Stripe Checkout Session that expires means the shopper **abandoned** the purchase, not that a payment failed — no payment was ever attempted. Today `handle_checkout_session_failure()` treats `checkout.session.expired` exactly like `checkout.session.async_payment_failed`: it marks the order Failed and sends the failed-order emails, so the shopper gets "your payment failed" up to 24 hours after a purchase that never began.

This PR splits the two events apart, and adds an early `$order->is_paid()` check to make sure that the order has not been paid for via some other means. This additional check handles situations where another payment method was used for the order, and the case where custom order statuses are in use, as reported in #5812.

The two code paths are changed as follows:

* `checkout.session.async_payment_failed` event handling is unchanged. We still mark the order as failed and will try to send an email for this case.
* `checkout.session.expired` handling is now as follows:
  - The order is set to **Cancelled** instead of Failed, and no failed-order email is sent to either the shopper or the merchant.
  - The expiry event is now a **no-op on an already-cancelled order**. This is the common case in practice: core's `wc_cancel_unpaid_orders()` cancels the pending order at the hold-stock timeout (default 60 minutes), well inside Stripe's 24h session TTL, so the webhook was arriving and reverting a correctly-cancelled order back to Failed.
  - **Neither cancelled-order email is sent by default**, though this only takes effect when emails are sent immediately, and the merchant is not using WooCommerce's deferred email handling from 11.0 and newer. Two filters are used to control the behaviour, with both defaulting to _not_ sending the email:
    * `wc_stripe_checkout_session_expired_should_send_cancelled_order_email_to_customer`
    * `wc_stripe_checkout_session_expired_should_send_cancelled_order_email_to_merchant`
Both filters receive `( bool $should_send, WC_Order $order )` and default to `false`. Note the email type must also be enabled in **WooCommerce → Settings → Emails** for a filter to have any effect — `customer_cancelled_order` is disabled by default — because `trigger()` still honours that setting, the same way `send_failed_order_email()` does.

Because WooCommerce only fires its cancelled-order emails on *some* transitions (never on `pending → cancelled` for the customer email, and only from WC 11.0 for the merchant email), a filter that merely declined to block would never actually send anything. So the plugin suppresses WooCommerce's automatic send for this cancellation and dispatches explicitly instead. That makes both filters behave identically regardless of the WooCommerce version or which status the order was in.

`wc_gateway_stripe_process_webhook_payment_error` still fires for both events, in its original position between the status change and the failed-order email — an earlier revision of this branch had dropped it for expired sessions.

### Backward compatibility

* Two new filters (additive). No hook removed, renamed, or reordered. No new persisted state.
* `cancelled` orders return `false` from `WC_Order::needs_payment()`, so a shopper can no longer pay one of these orders from My Account. This matches what core's own unpaid-order auto-cancel already does to the same orders.
* `wc_release_stock_for_order` runs on `cancelled` but not on `failed`, so held stock reservations are now released on abandonment.
* `woocommerce_order_status_cancelled` now fires on this path, which reaches `WC_Stripe_Order_Handler::cancel_payment()` and, for manual-capture-capable methods, `woocommerce_stripe_process_manual_cancel` — a change in *when* an existing hook fires. On an abandoned session the order carries no payment or setup intent, so `process_refund()` returns early without calling Stripe.

### Known limitation

With WooCommerce's opt-in **Deferred emails** feature enabled, the automatic send is queued to Action Scheduler and its `is_enabled()` check runs in a later request, where the request-scoped suppression no longer applies. On transitions WooCommerce hooks itself (`on-hold → cancelled` on any version, `pending → cancelled` on WC 11.0+) the suppressed email can still go out. Closing this would require per-order persisted state and site-wide filters, which we will take on in a follow-up PR.

## Testing instructions

1. Enable Optimized Checkout and Adaptive Pricing, Stripe in test mode, classic shortcode checkout.
2. Add a product to the cart, go to checkout, enter card `4000 0027 6000 3184`, and click **Place order**.
3. When the 3DS challenge appears, **abandon it** (navigate away). Confirm the order is `Pending payment`, has `_stripe_checkout_session_id` set, and has no transaction ID.
4. Expire the session immediately:
   `POST https://api.stripe.com/v1/checkout/sessions/{cs_test_...}/expire`
5. **Expected:** the order becomes **Cancelled** with the note "The checkout session has expired.", and **no** emails appear in **WooCommerce → Status → Logs → `transactional-emails`**.
6. **Already-cancelled case:** repeat steps 1–3, set `woocommerce_hold_stock_minutes` to `1` and wait for core to cancel the order, then expire the session. Expect no status change, no new order note, and no email. (On WC 11.0+ you will see a merchant email from core's auto-cancel a minute earlier — that one is not from this webhook.)
7. **Filter opt-in:** add `add_filter( 'wc_stripe_checkout_session_expired_should_send_cancelled_order_email_to_merchant', '__return_true' );` and repeat steps 1–5 — the merchant Cancelled order email is sent, the customer stays silent. Then enable **Cancelled order** for customers in WooCommerce → Settings → Emails and use the `..._to_customer` filter — the customer email is sent.
8. **Regression:** trigger `checkout.session.async_payment_failed` (e.g. a Boleto or Konbini order whose payment expires). The order must still become **Failed** and still send the failed-order emails, exactly as before this PR.

---

### Changelog entry

<details>

#### Comment

Entry added to `changelog.txt` and `readme.txt`:

`Fix - Treat an expired Stripe Checkout Session as an abandoned order and cancel it, instead of marking it failed and emailing the shopper a payment failure for a purchase that was never attempted`

</details>

**Post merge**


includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2514,24 +2516,93 @@ protected function handle_checkout_session_failure( object $notification ): void
...
+		} finally {
+			remove_filter( 'woocommerce_email_enabled_customer_cancelled_order', '__return_false' );
+			remove_filter( 'woocommerce_email_enabled_cancelled_order', '__return_false' );
+		}
+
+		$emails         = WC()->mailer()->get_emails();
+		$customer_email = $emails['WC_Email_Customer_Cancelled_Order'] ?? null;
+		$merchant_email = $emails['WC_Email_Cancelled_Order'] ?? null;
+
+		if ( $send_to_customer && $customer_email instanceof WC_Email_Customer_Cancelled_Order ) {
```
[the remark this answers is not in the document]
diegocurbelo: The class `WC_Email_Customer_Cancelled_Order` is in the repo 🤷🏼:
https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/emails/class-wc-email-customer-cancelled-order.php

includes/class-wc-stripe-webhook-handler.php
jaclync: nit: we might want to update the comment here, as this doesn't apply to the `checkout.session.expired` case anymore.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2514,24 +2516,93 @@ protected function handle_checkout_session_failure( object $notification ): void
 				return;
 			}
 
-			$message = 'checkout.session.expired' === $notification->type ? __( 'The checkout session has expired.', 'woocommerce-gateway-stripe' ) : __( 'The async payment for this checkout session has failed.', 'woocommerce-gateway-stripe' );
```
jaclync, line 2517: non-blocking: it might be easier to follow if both cases (expired or not) read from start to end, as both cases only share one line `do_action( 'wc_gateway_stripe_process_webhook_payment_error', ...)` (if we want to preserve the comment it can be extracted into a private function):

```php
if ( $is_expired ) {
    if ( $order->has_status( OrderStatus::CANCELLED ) ) {
        return;
    }
    $this->cancel_expired_checkout_session_order( $order );
    do_action( 'wc_gateway_stripe_process_webhook_payment_error', $order, $notification, null );
} else {
    $from_status = $order->get_status();
    $order->update_status( OrderStatus::FAILED, __( 'The async payment for this checkout session has failed.', 'woocommerce-gateway-stripe' ) );
    do_action( 'wc_gateway_stripe_process_webhook_payment_error', $order, $notification, null );
    $this->send_failed_order_email( $order->get_id(), [ 'from' => $from_status, 'to' => OrderStatus::FAILED ] );
}

// finally
```

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2514,24 +2516,93 @@ protected function handle_checkout_session_failure( object $notification ): void
...
+		 * @param bool     $should_send Whether to send the email. Default false.
+		 * @param WC_Order $order       The cancelled order.
+		 *
+		 * @since 10.9.0
+		 */
+		$send_to_customer = (bool) apply_filters( 'wc_stripe_checkout_session_expired_should_send_cancelled_order_email_to_customer', false, $order );
+
+		/**
+		 * Whether to email the merchant when an order is cancelled because its Stripe Checkout
+		 * Session expired.
```
jaclync: nit: for the docblock for these 2 new filters, maybe the notes about other conditions (emails enabled state, email type state, deferred email limitation, etc.) are worth mentioning? Just in case the merchants might think that returning true for these filters guarantees the email being sent.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2514,24 +2516,93 @@ protected function handle_checkout_session_failure( object $notification ): void
...
+		 *
+		 * @param bool     $should_send Whether to send the email. Default false.
+		 * @param WC_Order $order       The cancelled order.
+		 *
+		 * @since 10.9.0
+		 */
+		$send_to_merchant = (bool) apply_filters( 'wc_stripe_checkout_session_expired_should_send_cancelled_order_email_to_merchant', false, $order );
```
jaclync: nit: how about early returning if both `$send_to_*` values are `false`, since they are the default? It can save loading the WC email classes later when no email should be sent.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2514,24 +2516,93 @@ protected function handle_checkout_session_failure( object $notification ): void
...
+
+		$emails         = WC()->mailer()->get_emails();
+		$customer_email = $emails['WC_Email_Customer_Cancelled_Order'] ?? null;
+		$merchant_email = $emails['WC_Email_Cancelled_Order'] ?? null;
+
+		if ( $send_to_customer && $customer_email instanceof WC_Email_Customer_Cancelled_Order ) {
+			$customer_email->trigger( $order->get_id(), $order );
+		}
+
+		if ( $send_to_merchant && $merchant_email instanceof WC_Email_Cancelled_Order ) {
```
jaclync, line 2616: nit: in sibling methods like `send_failed_order_email`, it checks if the email is set by `isset( $emails['WC_Email_*'])` probably to account for cases when the email class might be replaced with another class externally. Maybe `isset` can be used here for consistency.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2514,24 +2516,93 @@ protected function handle_checkout_session_failure( object $notification ): void
...
+		add_filter( 'woocommerce_email_enabled_cancelled_order', '__return_false' );
+
+		try {
+			$order->update_status( OrderStatus::CANCELLED, __( 'The checkout session has expired.', 'woocommerce-gateway-stripe' ) );
+		} finally {
+			remove_filter( 'woocommerce_email_enabled_customer_cancelled_order', '__return_false' );
+			remove_filter( 'woocommerce_email_enabled_cancelled_order', '__return_false' );
+		}
+
+		$emails         = WC()->mailer()->get_emails();
```
jaclync, line 2608: nit: how about early returning right before the `WC()->mailer()` if both `$send_to_*` values are `false`, since they are the default? It can save loading the WC email classes later when no email should be sent.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2514,24 +2516,93 @@ protected function handle_checkout_session_failure( object $notification ): void
...
+		 *
+		 * @since 10.9.0
+		 */
+		$send_to_merchant = (bool) apply_filters( 'wc_stripe_checkout_session_expired_should_send_cancelled_order_email_to_merchant', false, $order );
+
+		add_filter( 'woocommerce_email_enabled_customer_cancelled_order', '__return_false' );
+		add_filter( 'woocommerce_email_enabled_cancelled_order', '__return_false' );
+
+		try {
+			$order->update_status( OrderStatus::CANCELLED, __( 'The checkout session has expired.', 'woocommerce-gateway-stripe' ) );
```
jaclync, line 2598: ❓ in `WC_Stripe_Order_Handler.prevent_cancelling_orders_awaiting_action`, it performs a few other checks on whether an order should be canceled from the core `woocommerce_cancel_unpaid_order` action. Do you think it's worth going through the same pre-cancelation path with a shared order handler helper?

jaclync: Also, the PR description mentions that the order "carries no payment or setup intent, so process_refund() returns early without calling Stripe," a potential corner case where an intent could exist: attempt 1 abandons the session (session still open), attempt 2 like from pay-for-order fails as a standalone intent (fires `payment_intent.payment_failed` webhook, saves intent to the order), and then the session hits its TTL and fires `checkout.session.expired`.

In this case, there could be a Stripe GET API call: the PR cancels the order here -> `cancel_payment()` -> `process_refund()` -> `recover_charge_id_from_intent()` -> live Stripe GET. If this is possible, maybe the PR description can be updated.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2514,24 +2516,93 @@ protected function handle_checkout_session_failure( object $notification ): void
...
+		 * Session expired.
+		 *
+		 * @param bool     $should_send Whether to send the email. Default false.
+		 * @param WC_Order $order       The cancelled order.
+		 *
+		 * @since 10.9.0
+		 */
+		$send_to_merchant = (bool) apply_filters( 'wc_stripe_checkout_session_expired_should_send_cancelled_order_email_to_merchant', false, $order );
+
+		add_filter( 'woocommerce_email_enabled_customer_cancelled_order', '__return_false' );
```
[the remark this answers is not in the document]
jaclync: +1 to this point of suppressing third-party callbacks site wide. Maybe we can use a callback taking 2 args like:

```php
$suppress_emails = static function ( $enabled, $email_order = null ) use ( $order_id ) {
	return $email_order instanceof WC_Order && $email_order->get_id() === $order_id ? false : $enabled;
};
```

This still won't work for deferred emails, but it's noted in the known limitations.

includes/class-wc-stripe-webhook-handler.php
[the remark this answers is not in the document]
daledupreez: And noting that it was added in https://github.com/woocommerce/woocommerce/pull/57987, which was first in WooCommerce 10.0.0. So this email has been in WooCommerce Core for a long time.

jaclync (commented):
Tested well, I think the email filter `__return_false` https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5810/changes#r3788458440 is worth addressing.

- ✅ Expired session (steps 1–5)
- ✅ Already-cancelled no-op (step 6)
- ✅ Merchant email opt-in (step 7)
- ✅ Customer email opt-in (step 7)
- ✅ Async-failure regression (step 8) (since my local store was using a US account, I used ACH with a test bank account that failed async)

includes/class-wc-stripe-webhook-handler.php
daledupreez: Good catch: I've added distinct comments for expiration and payment failures in 4e37644e50d01fd10b52ddd12df8e9685d9760e9.

includes/class-wc-stripe-webhook-handler.php
daledupreez: I added more context in 4e37644e50d01fd10b52ddd12df8e9685d9760e9. I think the context is important.

includes/class-wc-stripe-webhook-handler.php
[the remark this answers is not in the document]
daledupreez: This is a super-subtle issue. I've switched to a custom function in 884b43d55ce09bcb1eb76d700d7fce40e78b7dbb so we won't override other filters.

includes/class-wc-stripe-webhook-handler.php
daledupreez: Good suggestion. I've added an early return in 08ee69290b5fb56517de467b188cf7f266c5ed54.

includes/class-wc-stripe-webhook-handler.php
```diff
@@ -2505,33 +2506,118 @@ protected function handle_checkout_session_failure( object $notification ): void
...
 
+		$is_expired = 'checkout.session.expired' === $notification->type;
+
 		try {
 			if ( $order_helper->is_stripe_status_final( $order ) ) {
 				return;
 			}
 
-			if ( $order->has_status( [ OrderStatus::PROCESSING, OrderStatus::COMPLETED, OrderStatus::FAILED ] ) ) {
+			if ( $order->is_paid() || $order->has_status( [ OrderStatus::PROCESSING, OrderStatus::COMPLETED, OrderStatus::FAILED ] ) ) {
```
daledupreez, line 2516: Noting for the record that I've included the suggestion from https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5812 in this PR as it is touching the same code -- we should not process expiration or cancellation for orders that have already been paid.

daledupreez:
Merging as the e2e test failures look unrelated.

diegocurbelo (approved):

mbaierl:
## The "improved handling" leads to cancelled orders. 

Imagine the user can't do 3DS to verify the payment. Calls the shop owner, wire-transfers the money... the shop owner sets the order to "payment received" or "order completed". A few days later Stripe resets the order status to "cancelled", which automatically invalidates the invoice and notifies the user....
Legally there is now a cancelled order which has been paid.

Solution(s):
1. Add a filter to be able to prevent automatically cancelling the order (and maybe just notifying the shop admin)
2. Prior updating the status of the order validate that it's still in "waiting for a payment" state, not anything else.

Thanks for fixing!

daledupreez:
Thanks for flagging this, @mbaierl. The logic in this change keeps the order in place if it has been paid for. It sounds like the case you're outlining looks more like the following:
1. Shopper attempts to pay, but can't complete 3DS processing
2. Shopper may pay via a different async payment method, or may leave the order unpaid -- _I am not sure which would be the case in your example_
3. Shopper contacts the shop owner to arrange for a bank transfer
4. Shop Owner marks the order as paid -- _Does this also include updating the payment method type? Some details on how you would expect this to work would be helpful.__
5. Webhook from Stripe notes that the checkout session has expired -- the following checks include a number of checks for the order being paid or the order being in a pending or complete state
https://github.com/woocommerce/woocommerce-gateway-stripe/blob/2f06db5baabe5966cbc2c693ce7ede48fcee70ae/includes/class-wc-stripe-webhook-handler.php#L2693-L2710

If you're also using stock management, then the changes in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5759 may be relevant: when WooCommerce's auto-cancellation logic for orders runs, we check the Stripe payment status to ensure we don't cancel paid-for orders.

mbaierl:
@daledupreez 
Regarding #2: both happens in this shop - leaving orders unpaid or choosing another payment method.
Regarding #4: No, the payment method is not updated in Woo. The order is simply marked as "paid". I am not aware that the shop owner can change the payment method for an order? 

You are right, the code from #5 should handle the case where the order has been marked as completed... I'll further investigate what the exact flow is... (might be a PEBKAC style problem).

PR #5816, opened 2026-08-18
Show Apple Pay and Google Pay express buttons only on their own enabled locations

wjrosa:
Fixes STRIPE-1363

With multiple express wallets available, unchecking every location under Apple Pay & Google Pay's "Show express checkouts on" setting did not remove their buttons: they kept rendering on any page covered by *another* wallet's locations (e.g. Amazon Pay), and only unchecking that other wallet's locations made them disappear.

The root cause was on the client boundary. PHP already computes a per-method, location-aware answer (`WC_Stripe_Express_Checkout_Helper::is_apple_google_pay_enabled()`), but neither frontend ever received it:

- the classic entrypoint created the Apple Pay and Google Pay elements whenever the localized `is_express_checkout_enabled` aggregate (any wallet enabled for the page) was true, and the ECE options pin `applePay`/`googlePay` to `always`, so Stripe rendered them unconditionally;
- the Blocks integration registered both express payment methods off the same aggregate (`isExpressCheckoutEnabled`), and `canMakePayment` only checks wallet availability, not locations.

Both frontends now receive method-specific flags and gate each wallet's creation/registration on its own flag: `is_apple_pay_enabled` / `is_google_pay_enabled` in the express checkout params (classic/OC), `isApplePayEnabled` / `isGooglePayEnabled` in the Blocks config. Amazon Pay and Link already had their own flags and are unchanged.

### Why per-wallet flags when the settings are shared?

Apple Pay and Google Pay share one setting today, so the two flags are always equal — both are fed from the same `is_apple_google_pay_enabled()` helper. They are still shipped as separate keys deliberately: these localized params are new surface introduced by this PR, and external code can read them once released, so this is the one cheap moment to pick names. Stripe already models the two wallets separately (distinct payment method configuration IDs, per-wallet Express Checkout Element options), and a follow-up (STRIPE-1373) proposes splitting them in wp-admin. With per-wallet keys, that split only changes how PHP computes each value — the frontend contract and rendering path need no changes, and no combined key ever ships that would need a deprecation window afterwards.

### Backward compatibility

- Purely additive: four new localized config keys; the existing aggregate flags remain localized with unchanged values, so external readers of the params objects are unaffected.
- No hooks, options, or PHP signatures change.
- Location settings are read per site, as before; no multisite impact.

## Testing instructions

1. Connect a Stripe account with Apple Pay/Google Pay and Amazon Pay available.
2. Under **Payment methods → Apple Pay / Google Pay → Customize express checkouts**, untick Checkout, Product page, and Cart. Save.
3. Under **Amazon Pay**, tick all three locations. Save.
4. Visit a product page, the cart, and checkout (both shortcode and Blocks): only the Amazon Pay button should render.
5. Re-tick a location for Apple Pay / Google Pay: their buttons return on that page only.

Automated coverage: `WC_Stripe_Express_Checkout_Element_Test` asserts both localized flags carry the method-specific value while the aggregate stays true; Jest tests on the classic entrypoint assert Apple/Google elements are not created when only another wallet covers the page, are created when their own locations do, and that each wallet follows its own flag independently.

---

### DIFF
--- a/includes/payment-methods/class-wc-stripe-express-checkout-element.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-element.php
@@ -222,12 +222,8 @@ class WC_Stripe_Express_Checkout_Element {
 				'is_link_enabled'             => $this->express_checkout_helper->is_link_enabled(),
 				'is_express_checkout_enabled' => $this->express_checkout_helper->is_express_checkout_enabled(),
 				'is_amazon_pay_enabled'       => $this->express_checkout_helper->is_amazon_pay_enabled(),
-				// Apple/Google Pay need their own flags: `is_express_checkout_enabled` is the
-				// any-method aggregate, so it stays true when only another wallet's locations
-				// cover this page and must not decide whether these two buttons render.
-				// Both flags are currently backed by the shared Apple/Google Pay setting, so
-				// they are always equal — per-wallet keys so a future settings split only has
-				// to change how each value is computed, not the frontend contract.
+				// `is_express_checkout_enabled` aggregates all methods, so Apple/Google Pay need their
+				// own flags; per-wallet keys keep the contract stable if the shared setting ever splits.
 				'is_apple_pay_enabled'        => $this->express_checkout_helper->is_apple_google_pay_enabled(),
 				'is_google_pay_enabled'       => $this->express_checkout_helper->is_apple_google_pay_enabled(),
 			],
--- a/includes/payment-methods/class-wc-stripe-express-checkout-element.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-element.php
@@ -222,10 +222,14 @@ class WC_Stripe_Express_Checkout_Element {
 				'is_link_enabled'             => $this->express_checkout_helper->is_link_enabled(),
 				'is_express_checkout_enabled' => $this->express_checkout_helper->is_express_checkout_enabled(),
 				'is_amazon_pay_enabled'       => $this->express_checkout_helper->is_amazon_pay_enabled(),
-				// Apple/Google Pay need their own flag: `is_express_checkout_enabled` is the
+				// Apple/Google Pay need their own flags: `is_express_checkout_enabled` is the
 				// any-method aggregate, so it stays true when only another wallet's locations
 				// cover this page and must not decide whether these two buttons render.
-				'is_apple_google_pay_enabled' => $this->express_checkout_helper->is_apple_google_pay_enabled(),
+				// Both flags are currently backed by the shared Apple/Google Pay setting, so
+				// they are always equal — per-wallet keys so a future settings split only has
+				// to change how each value is computed, not the frontend contract.
+				'is_apple_pay_enabled'        => $this->express_checkout_helper->is_apple_google_pay_enabled(),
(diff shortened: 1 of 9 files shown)


includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -222,6 +222,14 @@ public function javascript_params() {
 				'is_link_enabled'             => $this->express_checkout_helper->is_link_enabled(),
 				'is_express_checkout_enabled' => $this->express_checkout_helper->is_express_checkout_enabled(),
 				'is_amazon_pay_enabled'       => $this->express_checkout_helper->is_amazon_pay_enabled(),
+				// Apple/Google Pay need their own flags: `is_express_checkout_enabled` is the
+				// any-method aggregate, so it stays true when only another wallet's locations
+				// cover this page and must not decide whether these two buttons render.
+				// Both flags are currently backed by the shared Apple/Google Pay setting, so
+				// they are always equal — per-wallet keys so a future settings split only has
+				// to change how each value is computed, not the frontend contract.
```
jaclync: non-blocking nit: the comment here and `includes/payment-methods/class-wc-stripe-upe-payment-gateway.php` feels a bit verbose, could probably be trimmed.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -603,6 +603,14 @@ public function javascript_params() {
...
 		$stripe_params['isAmazonPayEnabled']                = $express_checkout_helper->is_amazon_pay_enabled();
 		$stripe_params['isLinkEnabled']                     = $express_checkout_helper->is_link_enabled();
+		// Apple/Google Pay need their own flags: `isExpressCheckoutEnabled` is the any-method
+		// aggregate, so it stays true when only another wallet's locations cover this page
+		// and must not decide whether the Apple/Google Pay buttons register. Both flags are
+		// currently backed by the shared Apple/Google Pay setting, so they are always equal —
+		// per-wallet keys so a future settings split only has to change how each value is
+		// computed, not the frontend contract.
+		$stripe_params['isApplePayEnabled']  = $express_checkout_helper->is_apple_google_pay_enabled();
+		$stripe_params['isGooglePayEnabled'] = $express_checkout_helper->is_apple_google_pay_enabled();
```
jaclync, line 609: non-blocking: do you think we can add unit testing coverage on this change? The classic checkout version in `includes/payment-methods/class-wc-stripe-express-checkout-element.php` has it.

jaclync (approved):
Nice fix, LGTM :shipit:

includes/payment-methods/class-wc-stripe-express-checkout-element.php
wjrosa: Trimmed in both files (fbc0631f7).

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Added in fbc0631f7, mirroring the classic-side test. javascript_params() built the helper inline, so I added a protected get_express_checkout_helper() seam to substitute it.

PR #5820, opened 2026-08-18
Stop webhook auto-configuration from clobbering manual webhook setups

wjrosa:
Fixes STRIPE-1356

Automatic webhook reconfiguration (triggered silently on plugin updates and when enabling Adaptive Pricing) destroyed manually configured webhook setups: it deleted **every** endpoint pointing at the site's webhook URL — including ones the merchant created by following the settings page's own instructions — and silently overwrote a manually pasted signing secret. A stored webhook ID whose endpoint no longer exists also kept being displayed with no warning.

The plugin now keeps full control of its **own** endpoint while leaving merchant configuration alone:

- New endpoints are stamped with `metadata[created_by]`; cleanup only deletes stamped or settings-recorded endpoints and skips the rest.
- `webhook_data` records the signing secret auto-configuration wrote. If the stored secret was set manually, silent reconfiguration is skipped with a dismissible admin notice; the explicit Configure webhooks flow still replaces it. Legacy data without the recorded secret keeps the old behavior.
- If the stored webhook ID 404s during the update check, an admin notice asks the merchant to re-configure instead of displaying a ghost ID.

### Backward compatibility

All additive: endpoint metadata, one new `webhook_data` key, two `wc_stripe_show_*` options, one new public method. Cleanup deletes strictly fewer endpoints; stores with plugin-written secrets reconfigure exactly as before.

## Testing instructions

1. Let the plugin configure webhooks; the endpoint gets `created_by` metadata in the Dashboard.
2. Manually create a second endpoint at the same URL and trigger a reconfiguration (e.g. toggle Adaptive Pricing): the manual endpoint survives.
3. Paste the manual endpoint's secret into settings and trigger the silent path (`do_action( 'woocommerce_stripe_updated' )`): the secret is untouched and a notice appears; Configure webhooks replaces it and clears the notice.
4. Delete the plugin's endpoint in the Dashboard and trigger the update check: a "webhook no longer exists" notice appears.

Covered in `WC_Stripe_Account_Test`: deletion scoping, manual-secret skip (data provider), plugin-written and legacy secrets proceeding, ghost-ID notice, and endpoint stamping/recording/notice-clearing.

---

cc @daledupreez

### DIFF
--- a/includes/admin/class-wc-stripe-admin-notices.php
+++ b/includes/admin/class-wc-stripe-admin-notices.php
@@ -264,6 +264,28 @@ class WC_Stripe_Admin_Notices {
 				$this->add_admin_notice( '3ds', 'notice notice-warning', $message, true );
 			}
 
+			if ( 'yes' === get_option( WC_Stripe_Account::WEBHOOK_MISSING_NOTICE_OPTION ) ) {
+				$message = sprintf(
+					/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag */
+					__( 'WooCommerce Stripe - The webhook endpoint saved in your settings no longer exists in your Stripe account, so order updates from Stripe are not being received. Please %1$sre-configure your webhooks%2$s.', 'woocommerce-gateway-stripe' ),
+					'<a href="' . $this->get_setting_link() . '">',
+					'</a>'
+				);
+
+				$this->add_admin_notice( 'webhook_missing', 'notice notice-error', $message, true );
+			}
+
+			if ( 'yes' === get_option( WC_Stripe_Account::WEBHOOK_MANUAL_SECRET_NOTICE_OPTION ) ) {
+				$message = sprintf(
+					/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag */
+					__( 'WooCommerce Stripe - Automatic webhook reconfiguration was skipped because the webhook signing secret in your settings was set manually. Please verify your webhook configuration in the Stripe Dashboard, or %1$sre-configure your webhooks%2$s to let the plugin manage them.', 'woocommerce-gateway-stripe' ),
+					'<a href="' . $this->get_setting_link() . '">',
+					'</a>'
+				);
+
+				$this->add_admin_notice( 'webhook_manual_secret', 'notice notice-warning', $message, true );
+			}
+
 			if ( empty( $show_style_notice ) ) {
 				$message = sprintf(
 				/* translators: 1) HTML anchor open tag 2) HTML anchor closing tag */
@@ -842,6 +864,12 @@ class WC_Stripe_Admin_Notices {
 					break;
 				case 'changed_keys':
 					update_option( 'wc_stripe_show_changed_keys_notice', 'no' );
+					break;
+				case 'webhook_missing':
+					delete_option( WC_Stripe_Account::WEBHOOK_MISSING_NOTICE_OPTION );
+					break;
+				case 'webhook_manual_secret':
+					delete_option( WC_Stripe_Account::WEBHOOK_MANUAL_SECRET_NOTICE_OPTION );
 					break;
 				case 'legacy_deprecation':
 					update_option( 'wc_stripe_show_legacy_deprecation_notice', 'no' );
--- a/includes/class-wc-stripe-account.php
+++ b/includes/class-wc-stripe-account.php
@@ -83,6 +83,25 @@ class WC_Stripe_Account {
 	protected const WEBHOOK_STATUS_CACHE_KEY = 'webhook_status';
 
 	/**
+	 * Metadata key/value stamped on webhook endpoints this plugin creates, so cleanup can
+	 * tell them apart from endpoints the merchant created at the same URL by hand.
+	 */
+	public const WEBHOOK_METADATA_CREATED_BY_KEY   = 'created_by';
+	public const WEBHOOK_METADATA_CREATED_BY_VALUE = 'woocommerce_gateway_stripe';
+
+	/**
+	 * Option set when automatic reconfiguration is skipped because the stored signing
+	 * secret was entered manually; read by WC_Stripe_Admin_Notices.
+	 */
+	public const WEBHOOK_MANUAL_SECRET_NOTICE_OPTION = 'wc_stripe_show_webhook_manual_secret_notice';
+
+	/**
+	 * Option set when the stored webhook endpoint no longer exists in the connected
+	 * Stripe account; read by WC_Stripe_Admin_Notices.
+	 */
+	public const WEBHOOK_MISSING_NOTICE_OPTION = 'wc_stripe_show_webhook_missing_notice';
+
+	/**
 	 * The Stripe connect instance.
 	 *
 	 * @var WC_Stripe_Connect
@@ -336,6 +355,9 @@ class WC_Stripe_Account {
 			'enabled_events' => self::WEBHOOK_EVENTS,
 			'url'            => WC_Stripe_Helper::get_webhook_url(),
 			'api_version'    => self::get_webhooks_api_version(),
+			// Stamp the endpoint so later cleanup only ever deletes endpoints this plugin
+			// created, never ones the merchant added at the same URL by hand.
+			'metadata'       => [ self::WEBHOOK_METADATA_CREATED_BY_KEY => self::WEBHOOK_METADATA_CREATED_BY_VALUE ],
 		];
 
 		$response = WC_Stripe_API::request( $request, 'webhook_endpoints', 'POST' );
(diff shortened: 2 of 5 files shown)


includes/class-wc-stripe-account.php
```diff
@@ -363,13 +375,19 @@ public function configure_webhooks( $mode = 'live' ) {
...
+			'url'            => wc_clean( $response->url ),
+			'secret'         => WC_Stripe_API::get_secret_key(),
+			// Lets reconfiguration tell a plugin-written secret from a manually pasted one.
+			'signing_secret' => wc_clean( $response->secret ),
 		];
 
 		WC_Stripe_Helper::update_main_stripe_settings( $settings );
 
+		// A successful (re)configuration resolves whatever the webhook notices were flagging.
+		delete_option( self::WEBHOOK_MANUAL_SECRET_NOTICE_OPTION );
```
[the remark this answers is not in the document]
daledupreez: This is a valid issue.

includes/class-wc-stripe-account.php
```diff
@@ -569,6 +607,27 @@ public function get_existing_webhook() {
...
 	}
 
+	/**
+	 * Whether a webhook endpoint still exists in the connected Stripe account.
+	 * Only resource_missing counts as gone, so a transient API failure can't raise a false alarm.
+	 *
+	 * @param string $webhook_id The webhook endpoint ID to check.
+	 * @return bool
+	 */
+	public function webhook_endpoint_exists( string $webhook_id ): bool {
```
daledupreez: Should the current mode be taken into consideration for this API call? Or should this method be private/protected? At present, it works because the local call to this method is setting the relevant secret key before calling this logic.

includes/class-wc-stripe-account.php
```diff
@@ -82,6 +82,16 @@ class WC_Stripe_Account {
...
 
+	/** Metadata stamped on endpoints this plugin creates, so cleanup can tell them from merchant-created ones. */
+	public const WEBHOOK_METADATA_CREATED_BY_KEY   = 'created_by';
+	public const WEBHOOK_METADATA_CREATED_BY_VALUE = 'woocommerce_gateway_stripe';
+
+	/** Option flagging that reconfiguration was skipped over a manually set signing secret. */
+	public const WEBHOOK_MANUAL_SECRET_NOTICE_OPTION = 'wc_stripe_show_webhook_manual_secret_notice';
+
+	/** Option flagging that the stored webhook endpoint no longer exists in Stripe. */
+	public const WEBHOOK_MISSING_NOTICE_OPTION = 'wc_stripe_show_webhook_missing_notice';
```
daledupreez: Do these constants need to be public, especially the metadata fields?

daledupreez (commented):
I have some comments from inspection.

includes/class-wc-stripe-account.php
wjrosa: Made it `protected` in 785135230. It only queries whichever key `WC_Stripe_API` holds, so it relies on the caller setting the mode key first; not safe as public API.

includes/class-wc-stripe-account.php
wjrosa: The metadata constants are now `protected` (785135230), only this class uses them. Kept the two `*_NOTICE_OPTION` constants public since `WC_Stripe_Admin_Notices` reads them.

includes/class-wc-stripe-account.php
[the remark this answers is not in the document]
wjrosa: Fixed in 41207cfc1. Both flags are now per mode via `WC_Stripe_Account::get_webhook_notice_option( $base, $mode )` (`_live`/`_test`); the account class writes/clears them per mode, and the admin notice shows if either is set and clears both on dismiss. Added a test that a test reconfigure leaves a live-mode notice intact.

includes/admin/class-wc-stripe-admin-notices.php
```diff
@@ -277,7 +281,11 @@ public function stripe_check_environment() {
 				$this->add_admin_notice( 'webhook_missing', 'notice notice-error', $message, true );
 			}
 
-			if ( 'yes' === get_option( WC_Stripe_Account::WEBHOOK_MANUAL_SECRET_NOTICE_OPTION ) ) {
+			$webhook_manual_secret_flagged =
+				'yes' === get_option( WC_Stripe_Account::get_webhook_notice_option( WC_Stripe_Account::WEBHOOK_MANUAL_SECRET_NOTICE_OPTION, 'live' ) )
+				|| 'yes' === get_option( WC_Stripe_Account::get_webhook_notice_option( WC_Stripe_Account::WEBHOOK_MANUAL_SECRET_NOTICE_OPTION, 'test' ) );
+
+			if ( $webhook_manual_secret_flagged ) {
```
daledupreez: For both of these notices, I think it would help to know whether they apply to live or test mode. Merchants should care a _lot_ more about live webhooks, so showing notices for test webhooks would be panic-inducing. Furthermore, without knowing which mode is affected, merchants could end up continually reconfiguring webhooks for the wrong mode.

tests/phpunit/class-wc-stripe-account-test.php
```diff
@@ -438,7 +440,7 @@ public function test_delete_previously_configured_webhooks_with_exclusion() {
 	 */
 	public function test_delete_previously_configured_webhooks_without_exclusion() {
 		$webhook_url = WC_Stripe_Helper::get_webhook_url();
-		$plugin_meta = (object) [ WC_Stripe_Account::WEBHOOK_METADATA_CREATED_BY_KEY => WC_Stripe_Account::WEBHOOK_METADATA_CREATED_BY_VALUE ];
+		$plugin_meta = (object) [ 'created_by' => 'woocommerce_gateway_stripe' ];
```
daledupreez: Rather than hard-coding the values, we can also use `WC_Stripe_Test_Helper::get_class_const_value()`.

tests/phpunit/class-wc-stripe-account-test.php
```diff
@@ -859,18 +861,62 @@ public function test_configure_webhooks_records_signing_secret_and_stamps_endpoi
 			parse_str( $captured_body, $captured_body );
 		}
 
+		// Literals pin the exact wire contract stamped on the endpoint (the
+		// constants are protected on WC_Stripe_Account).
 		$this->assertSame(
-			WC_Stripe_Account::WEBHOOK_METADATA_CREATED_BY_VALUE,
-			$captured_body['metadata'][ WC_Stripe_Account::WEBHOOK_METADATA_CREATED_BY_KEY ] ?? null
+			'woocommerce_gateway_stripe',
+			$captured_body['metadata']['created_by'] ?? null
```
daledupreez: As above, we can use `WC_Stripe_Test_Helper::get_class_const_value()` to get the actual values.

includes/admin/class-wc-stripe-admin-notices.php
wjrosa: Done in fec0d03. Both notices now name the affected mode and are raised and dismissed per mode, so dismissing a test notice can't clear a live one.

tests/phpunit/class-wc-stripe-account-test.php
wjrosa: Done in d7c03a0.

tests/phpunit/class-wc-stripe-account-test.php
wjrosa: Done in d7c03a0.

includes/class-wc-stripe-account.php
wjrosa: It's `protected` now, and its docblock states the caller sets the mode's secret key first (as the reconfigure loop does), so it stays keyless rather than duplicating the mode-to-key resolution.

includes/class-wc-stripe-account.php
wjrosa: The webhook metadata constants are `protected` now, not public.

includes/class-wc-stripe-account.php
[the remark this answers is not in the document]
wjrosa: The notice options are scoped per mode now, and fec0d03 adds a test that dismissing one mode does not clear the other.

includes/class-wc-stripe-account.php
```diff
@@ -597,9 +675,26 @@ public function maybe_reconfigure_webhooks_on_update( string $update_type = 'plu
...
+				// The stored endpoint can vanish (deleted in the Dashboard, or created under a
+				// previously connected account); notice instead of displaying the ghost ID forever.
+				if (
+					'' !== $stored_webhook_id
+					&& ( ! $existing_webhook || ( $existing_webhook->id ?? '' ) !== $stored_webhook_id )
+					&& ! $this->webhook_endpoint_exists( $stored_webhook_id )
+				) {
+					update_option( self::get_webhook_notice_option( self::WEBHOOK_MISSING_NOTICE_OPTION, $mode ), 'yes' );
+					WC_Stripe_Logger::info( "Stored webhook {$stored_webhook_id} for {$mode} mode no longer exists in the Stripe account." );
+				}
```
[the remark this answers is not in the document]
wjrosa: Fixed in 0da2ef6. Reconfiguration now resolves the comparison target to the plugin's stored endpoint by ID (`get_webhook_endpoint_by_id`), falling back to the URL match only when there's no stored ID, so a merchant endpoint sharing the site URL can't cause the plugin's own outdated endpoint to be skipped. Added a test for the two-endpoints-at-one-URL case.

wjrosa:
Changelog validation job failures being fixed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5991

includes/class-wc-stripe-account.php
```diff
@@ -597,9 +694,40 @@ public function maybe_reconfigure_webhooks_on_update( string $update_type = 'plu
...
+				// merchant's manual endpoint sharing that URL. Compare events/version against the
+				// plugin's own recorded endpoint when it still exists, so an outdated plugin endpoint
+				// isn't skipped just because a merchant endpoint at the same URL is up to date.
+				if (
+					'' !== $stored_webhook_id
+					&& ( ! $existing_webhook || ( $existing_webhook->id ?? '' ) !== $stored_webhook_id )
+				) {
+					$stored_endpoint = $this->get_webhook_endpoint_by_id( $stored_webhook_id );
+					if ( false !== $stored_endpoint ) {
+						$existing_webhook = $stored_endpoint;
```
diegocurbelo: Should we check that the endpoint matched the site URL? Otherwise, on a staging clone (with production’s keys/settings), it would now trigger the creation of a new webhook for staging.

We should also add a test for a stored endpoint that belongs to another site URL

PR #5824, opened 2026-08-19
Retry failed remote-config syncs with backoff and surface cache staleness

wjrosa:
Fixes STRIPE-1381

Hardens the remote-config cache lifecycle against endpoint unreachability:

- **Retry with backoff.** A failed fetch previously waited ~24h for the next daily run. It now schedules in-cycle retries at +1h and +4h, with the attempt number as the action arg and duplicates guarded. The deliberate disabled-channel error is not retried.
- **Staleness observability.** Failure logs include a consecutive-failure counter (`_wcstripe_remote_config_sync_failures`, reset on first success) and per-mode cache age (new `get_cache_age()`), escalating to warning once a connected mode's cache is older than 7 days.
- **No TTL expiry, recorded as a decision.** Expiring stale overrides back to local defaults would re-enable a remotely disabled feature on exactly the sites the disable can no longer reach. Rationale and the manual recovery path (delete the per-mode option, or the internal override) are now in the `WC_Stripe_Remote_Config` docblock.

BC impact: no public surface changes. The `wc_stripe_remote_config_sync` action gains an optional attempt arg (default 0), so existing schedules keep working.

## Testing instructions

1. `wp option update _wcstripe_remote_config_enabled yes`, then block reach to `public-api.wordpress.com` (e.g. /etc/hosts).
2. Run the sync action; confirm a pending `wc_stripe_remote_config_sync` action with args `[1]` ~1h out and `_wcstripe_remote_config_sync_failures` = 1.
3. Restore connectivity, run the pending action; confirm the failure option is deleted and `_wcstripe_remote_config_{mode}` has a fresh `fetched_at`.

Covered by PHPUnit: backoff per attempt, chain cap, no retry on the disabled error, counter reset on success, `get_cache_age()`.

---

### Changelog entry

**Post merge**

### DIFF
--- a/includes/remote-config/class-wc-stripe-remote-config-scheduler.php
+++ b/includes/remote-config/class-wc-stripe-remote-config-scheduler.php
@@ -9,6 +9,7 @@ if ( ! defined( 'ABSPATH' ) ) {
  * - Daily Action Scheduler job (`wc_stripe_remote_config_sync`).
  * - Immediate single-action enqueue on plugin upgrade
  * - Immediate single-action enqueue when API keys or the test/live mode change
+ * - In-cycle retries with backoff after a failed fetch (see RETRY_DELAYS)
  *
  * On each run (skipped entirely when no mode has keys), makes one combined
  * Client::fetch_all call and applies each mode's payload.
@@ -17,6 +18,26 @@ class WC_Stripe_Remote_Config_Scheduler {
 
 	public const SYNC_ACTION     = 'wc_stripe_remote_config_sync';
 	public const SCHEDULER_GROUP = 'woocommerce-gateway-stripe';
+
+	/**
+	 * Option tracking consecutive failed fetches. Reset on the first
+	 * successful fetch; never autoloaded.
+	 */
+	public const FAILURE_COUNT_OPTION = '_wcstripe_remote_config_sync_failures';
+
+	/**
+	 * Delays (seconds) before each in-cycle retry after a failed fetch: a run
+	 * at attempt N that fails schedules attempt N+1 after RETRY_DELAYS[N].
+	 * Past the end of the list, the sync waits for the next daily run, so a
+	 * fully failed cycle contacts the endpoint three times per day.
+	 */
+	protected const RETRY_DELAYS = [ HOUR_IN_SECONDS, 4 * HOUR_IN_SECONDS ];
+
+	/**
+	 * Cache age (seconds) past which a failed fetch logs at warning level:
+	 * remote flag changes have not reached this site for this long.
+	 */
+	protected const STALENESS_WARNING_SECONDS = 7 * DAY_IN_SECONDS;
 
 	/**
 	 * Outbound HTTP client used to fetch remote-config payloads.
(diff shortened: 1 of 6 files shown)


includes/remote-config/class-wc-stripe-remote-config-scheduler.php
```diff
@@ -18,6 +19,22 @@ class WC_Stripe_Remote_Config_Scheduler {
...
+	/**
+	 * Backoff delays: a failed run at attempt N schedules attempt N+1 after
+	 * RETRY_DELAYS[N]; past the end, the next contact is the daily run.
+	 */
+	protected const RETRY_DELAYS = [ HOUR_IN_SECONDS, 4 * HOUR_IN_SECONDS ];
+
+	/**
+	 * Cache age (seconds) past which a failed fetch logs at warning level.
+	 */
+	protected const STALENESS_WARNING_SECONDS = 7 * DAY_IN_SECONDS;
```
jaclync, line 36: non-blocking nit: could these three constants be private? They're only used in this file + tests.

includes/remote-config/class-wc-stripe-remote-config-scheduler.php
```diff
@@ -108,6 +120,89 @@ public function run(): void {
 				$this->remote_config->apply( $mode, $response['modes'][ $mode ] );
```
jaclync, line 120: ❓ If the response is malformed and thus `remote_config->apply` fails (returning `false`) for both modes, might we want to consider this as a failure in addition to the `is_wp_error` case above?

includes/remote-config/class-wc-stripe-remote-config-scheduler.php
```diff
@@ -108,6 +120,89 @@ public function run(): void {
...
 			}
 		}
+
+		if ( 0 !== (int) get_option( self::FAILURE_COUNT_OPTION, 0 ) ) {
```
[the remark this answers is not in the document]
jaclync: +1, cancelling scheduled sync actions from previous sync failures after a successful sync makes sense for efficiency.

includes/remote-config/class-wc-stripe-remote-config-scheduler.php
```diff
@@ -108,6 +120,89 @@ public function run(): void {
...
+	 * Records a failed fetch and schedules the next in-cycle retry. The cache
+	 * is left untouched: last-known-good wins (see WC_Stripe_Remote_Config).
+	 *
+	 * @param WP_Error $error   The fetch failure.
+	 * @param int      $attempt In-cycle attempt number of the failed run.
+	 * @return void
+	 */
+	private function handle_failed_fetch( WP_Error $error, int $attempt ): void {
+		$failures = (int) get_option( self::FAILURE_COUNT_OPTION, 0 ) + 1;
+		update_option( self::FAILURE_COUNT_OPTION, $failures, false );
```
jaclync, line 139: non-blocking: is the `FAILURE_COUNT_OPTION` option reliable, could `attempt` be sufficient for logging? Since the option is only reset out after a successful sync at the end, if remote config becomes disabled after some failures, the option stays around the next time it's enabled. Concurrent writes could also make the count not always update to date.

includes/remote-config/class-wc-stripe-remote-config-scheduler.php
```diff
@@ -108,6 +120,89 @@ public function run(): void {
...
+	/**
+	 * Schedules the next retry attempt with backoff. The deliberate
+	 * disabled-channel error is not retried.
+	 *
+	 * @param WP_Error $error   The fetch failure.
+	 * @param int      $attempt In-cycle attempt number of the failed run.
+	 * @return void
+	 */
+	private function maybe_schedule_retry( WP_Error $error, int $attempt ): void {
+		if ( 'wc_stripe_remote_config_disabled' === $error->get_error_code() ) {
```
jaclync, line 192: non-blocking: could this error code constant be shared with its trigger(s)?

jaclync (approved):
Working as expected! Left some comments but more for edge cases, approving first :shipit:

PR #5829, opened 2026-08-20
Stop offering Sofort at checkout while keeping refunds and subscription renewals working

### LINKED ISSUE #5828: Stop offering Sofort at checkout (discontinued by Stripe) while keeping refunds and subscription renewals working
### Describe the bug

Stripe discontinued Sofort as a standalone payment method on **March 31, 2025** and now instructs integrations to remove it to avoid payment failures ([Stripe docs](https://docs.stripe.com/payments/sofort), [support article](https://support.stripe.com/questions/sofort-is-being-deprecated-as-a-standalone-payment-method)).

The plugin already hides Sofort from merchants who don't have it enabled (`WC_Stripe_UPE_Payment_Gateway` constructor), but merchants who had it enabled before the deprecation still get Sofort offered at checkout — where every new payment attempt now fails at Stripe. The full Sofort wiring is still active for those stores: `$is_reusable = true`, tokenization support, subscriptions and pre-orders initialization, Blocks support, and icons.

### What should happen instead

Sofort should no longer be offered at checkout for anyone, following the same pattern used for giropay (instantiated only on the order details page / refund requests).

Two things must keep working for a while:

1. **Refunds** — existing Sofort orders must remain refundable from the order details page, so the `stripe_sofort` gateway must stay registered in those admin/refund contexts (same approach as giropay).
2. **Subscription renewals** — subscriptions signed up via Sofort renew through SEPA Direct Debit mandates, which Stripe keeps active after the discontinuation ("Existing SEPA direct debit mandates initiated by SOFORT payments remain active"). The renewal path (`woocommerce_scheduled_subscription_payment_stripe_sofort`) and saved (SEPA-typed) tokens must not break.

Full removal of the class, constant, icons, and Blocks wiring can happen in a follow-up once the refund/renewal window is over.

### Acceptance criteria

- [ ] Sofort is not offered on classic checkout, Blocks checkout, optimized checkout, or express checkout — including for merchants that still have it in their enabled payment methods.
- [ ] Refunding an existing Sofort order from the admin order details page still works.
- [ ] Existing Sofort-initiated subscriptions continue to renew via their SEPA mandate.
- [ ] Saved payment methods created via Sofort remain usable/visible where they were before.
- [ ] Targeted PHPUnit coverage for the availability change.

*Drafted with Claude Code; reviewed by me.*

wjrosa:
Fixes #5828

Stripe [discontinued Sofort on March 31, 2025](https://docs.stripe.com/payments/sofort), so new Sofort payments now fail. The plugin already hid Sofort from merchants who never enabled it, but those who had it enabled before the deprecation still got it offered at checkout.

- `WC_Stripe_UPE_Payment_Method_Sofort::is_enabled_at_checkout()` now always returns `false`, the single choke point for classic checkout, Blocks, Optimized Checkout, and the add-payment-method page, so Sofort disappears from every checkout surface.
- Refunds and renewals for existing Sofort orders keep working: the gateway is still instantiated on the order-details and refund paths (mirroring giropay), stores that still have Sofort enabled keep their renewal hooks, and tokenization and the SEPA-typed token mapping are untouched.
- The Sofort and giropay classes and their `WC_Stripe_Payment_Methods` constants are tagged `@deprecated 11.0.0` (PHPDoc only, since both still run on refund/renewal paths and the constant strings read historical order and webhook data).

Full removal is deferred to a follow-up once the refund/renewal window is over.

### Backward compatibility

- `WC_Stripe_UPE_Payment_Method_Sofort::is_enabled_at_checkout()` changes behavior (always `false`). Any third-party code calling it will now see Sofort reported as unavailable at checkout — which matches reality at Stripe, where these payments fail.
- No hooks, option keys, or public symbols are renamed or removed. The `stripe_sofort` gateway remains registered in the same contexts as before, plus the order-details/refund contexts for stores that disabled Sofort.
- Settings are read/written through the existing site-scoped helpers; no multisite behavior changes. Not explicitly tested under multisite.

### Testing instructions

Sofort can no longer be created at checkout, which is what this PR enforces, so a fresh store has no Sofort orders or subscriptions to test refunds and renewals against. Sofort settles as a SEPA Direct Debit on Stripe, so seed the test data from SEPA. Testers who already have historical Sofort orders or subscriptions can use those and skip the setup.

**Setup**

1. On a store connected to Stripe, set the store currency to EUR and enable both SEPA Direct Debit and Sofort. The UI no longer offers Sofort, so add it to `woocommerce_stripe_settings` > `upe_checkout_experience_accepted_payments` in the DB. Keep Sofort enabled for the whole test, so its gateway stays instantiated and the refund and renewal hooks stay registered.
2. Place a SEPA Direct Debit order. For the renewal test, also start a SEPA Direct Debit subscription with WooCommerce Subscriptions active.
3. Change the payment method of that order (and subscription) to `stripe_sofort` by setting the `_payment_method` post meta in the DB. They now look like Sofort-initiated records.

**Checks**

4. Visit classic checkout and Blocks checkout: Sofort must not be offered. Before this change it was, and paying with it failed at Stripe.
5. Open the seeded Sofort order in wp-admin and issue a partial refund: it must succeed. SEPA charges settle asynchronously, so the charge must be settled before it can be refunded.
6. Trigger a renewal for the seeded Sofort subscription (scheduled action `woocommerce_scheduled_subscription_payment_stripe_sofort`): the renewal must process against the saved SEPA-typed token.

`npm run test:php -- --filter WC_Stripe_UPE_Payment_Method_Test` → 118 tests OK; `--filter WC_Stripe_UPE_Payment_Gateway_Test` → 241 tests OK. PHPStan clean; PHPCS unchanged.

### Changelog entry

> Update - Stop offering Sofort at checkout (discontinued by Stripe on March 31, 2025); refunds and subscription renewals for existing Sofort orders keep working

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,6 +20,7 @@ xxxx-xx-xx - version 11.0.0
 * Dev - Make GITHUB_TOKEN optional for the local E2E Docker setup by falling back to the GitHub CLI and then to plugin zips placed in tests/e2e/deps
 * Fix - Reload the plugin's payment method settings when refreshing account details, so the settings screen no longer keeps showing stale values
 * Update - Stop offering Sofort at checkout (discontinued by Stripe on March 31, 2025); refunds and subscription renewals for existing Sofort orders keep working
+* Dev - Mark the Sofort and giropay payment method classes and constants as deprecated ahead of their removal
 
 2026-08-17 - version 10.9.0
 * Fix - Re-enable Adaptive Pricing if it was automatically disabled after a Checkout Session amount mismatch
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,6 +19,7 @@ xxxx-xx-xx - version 11.0.0
 * Fix - Name password protection and hidden visibility as their own reasons in the Agentic Commerce feed preview
 * Dev - Make GITHUB_TOKEN optional for the local E2E Docker setup by falling back to the GitHub CLI and then to plugin zips placed in tests/e2e/deps
 * Fix - Reload the plugin's payment method settings when refreshing account details, so the settings screen no longer keeps showing stale values
+* Update - Stop offering Sofort at checkout (discontinued by Stripe on March 31, 2025); refunds and subscription renewals for existing Sofort orders keep working
 
 2026-08-17 - version 10.9.0
 * Fix - Re-enable Adaptive Pricing if it was automatically disabled after a Checkout Session amount mismatch
--- a/includes/payment-methods/class-wc-stripe-upe-payment-method-giropay.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-method-giropay.php
@@ -5,6 +5,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 /**
  * The giropay Payment Method class extending UPE base class
+ *
+ * @deprecated 11.0.0 Discontinued by Stripe on 2024-06-30; kept only for refunds on existing orders.
  */
 class WC_Stripe_UPE_Payment_Method_Giropay extends WC_Stripe_UPE_Payment_Method {
 
--- a/includes/payment-methods/class-wc-stripe-upe-payment-method-sofort.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-method-sofort.php
@@ -46,4 +46,19 @@ class WC_Stripe_UPE_Payment_Method_Sofort extends WC_Stripe_UPE_Payment_Method {
 	public function get_retrievable_type() {
 		return WC_Stripe_UPE_Payment_Method_Sepa::STRIPE_ID;
 	}
+
+	/**
+	 * Sofort was discontinued by Stripe on 2025-03-31, so new payments always fail
+	 * and the method must never be offered at checkout. The class is kept (still
+	 * reusable, with subscriptions initialized) because existing Sofort-initiated
+	 * subscriptions renew through their SEPA mandate, which Stripe keeps active,
+	 * and old Sofort orders remain refundable.
+	 *
+	 * @param int|null    $order_id
+	 * @param string|null $account_domestic_currency The account's default currency.
+	 * @return bool
+	 */
+	public function is_enabled_at_checkout( $order_id = null, $account_domestic_currency = null ) {
+		return false;
+	}
 }
--- a/includes/payment-methods/class-wc-stripe-upe-payment-method-sofort.php
+++ b/includes/payment-methods/class-wc-stripe-upe-payment-method-sofort.php
@@ -8,6 +8,9 @@ if ( ! defined( 'ABSPATH' ) ) {
 
 /**
  * Sofort Payment Method class extending UPE base class
+ *
+ * @deprecated 11.0.0 Discontinued by Stripe on 2025-03-31; kept only for refunds and
+ * SEPA-mandate subscription renewals on existing orders.
  */
 class WC_Stripe_UPE_Payment_Method_Sofort extends WC_Stripe_UPE_Payment_Method {
 	use WC_Stripe_Subscriptions_Trait;
@@ -48,11 +51,7 @@ class WC_Stripe_UPE_Payment_Method_Sofort extends WC_Stripe_UPE_Payment_Method {
 	}
 
 	/**
-	 * Sofort was discontinued by Stripe on 2025-03-31, so new payments always fail
-	 * and the method must never be offered at checkout. The class is kept (still
-	 * reusable, with subscriptions initialized) because existing Sofort-initiated
-	 * subscriptions renew through their SEPA mandate, which Stripe keeps active,
-	 * and old Sofort orders remain refundable.
+	 * Sofort is discontinued, so it can never be offered at checkout.
 	 *
 	 * @param int|null    $order_id
 	 * @param string|null $account_domestic_currency The account's default currency.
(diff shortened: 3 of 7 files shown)


tests/phpunit/payment-methods/class-wc-stripe-upe-payment-method-test.php
```diff
@@ -585,7 +598,7 @@ public function test_payment_methods_are_only_enabled_when_capability_is_active(
 
 		$payment_method_ids = array_map( [ $this, 'get_id' ], $this->mock_payment_methods );
 		foreach ( $payment_method_ids as $id ) {
-			if ( WC_Stripe_Payment_Methods::CARD === $id || WC_Stripe_Payment_Methods::BOLETO === $id || WC_Stripe_Payment_Methods::OXXO === $id || WC_Stripe_Payment_Methods::GIROPAY === $id ) {
+			if ( WC_Stripe_Payment_Methods::CARD === $id || WC_Stripe_Payment_Methods::BOLETO === $id || WC_Stripe_Payment_Methods::OXXO === $id || WC_Stripe_Payment_Methods::GIROPAY === $id || WC_Stripe_Payment_Methods::SOFORT === $id ) {
 				continue;
 			}
```
daledupreez: Not blocking, but it might be easier to read this code if we use an array to list the payment methods to skip.
```suggestion
		$payment_methods_to_skip = [
			WC_Stripe_Payment_Methods::CARD,
			WC_Stripe_Payment_Methods::BOLETO,
			WC_Stripe_Payment_Methods::OXXO,
			WC_Stripe_Payment_Methods::GIROPAY,
			WC_Stripe_Payment_Methods::SOFORT,
		];
		foreach ( $payment_method_ids as $id ) {
			if (  in_array( $id, $payment_methods_to_skip, true ) ) {
				continue;
			}
```

includes/payment-methods/class-wc-stripe-upe-payment-method-sofort.php
```diff
@@ -46,4 +49,15 @@ public function __construct() {
 	public function get_retrievable_type() {
 		return WC_Stripe_UPE_Payment_Method_Sepa::STRIPE_ID;
 	}
+
+	/**
+	 * Sofort is discontinued, so it can never be offered at checkout.
+	 *
+	 * @param int|null    $order_id
+	 * @param string|null $account_domestic_currency The account's default currency.
+	 * @return bool
```
daledupreez: Nit: It may be worth specifying the return value as always being false.
```suggestion
	 * @return bool Always return false, as Sofort is no longer available for new purchases.
```

changelog.txt
```diff
@@ -19,6 +19,8 @@ xxxx-xx-xx - version 11.0.0
 * Fix - Name password protection and hidden visibility as their own reasons in the Agentic Commerce feed preview
 * Dev - Make GITHUB_TOKEN optional for the local E2E Docker setup by falling back to the GitHub CLI and then to plugin zips placed in tests/e2e/deps
 * Fix - Reload the plugin's payment method settings when refreshing account details, so the settings screen no longer keeps showing stale values
+* Update - Stop offering Sofort at checkout (discontinued by Stripe on March 31, 2025); refunds and subscription renewals for existing Sofort orders keep working
```
daledupreez: Should this be a `Remove` entry?
```suggestion
* Remove - Stop offering Sofort at checkout (discontinued by Stripe on March 31, 2025); refunds and subscription renewals for existing Sofort orders keep working
```

daledupreez:
@wjrosa, for testing this, is it possible to set up a new subscription using Sofort? I don't have one at the moment, so I am not sure how we can test that.

wjrosa:
You can't create a new Sofort subscription anymore; Stripe discontinued it, and this PR enforces that. You have to recreate an existing one.

It's easy because Sofort renews through a SEPA token (`get_retrievable_type()` returns SEPA, and the renewal just charges the saved source off-session):

1. Keep Sofort enabled in settings so its renewal hook stays registered.
2. Create a real SEPA Direct Debit subscription.
3. Change its payment method to `stripe_sofort` in the DB.
4. Process a renewal.

That gives a real Sofort-initiated subscription: SEPA token, Sofort gateway.

daledupreez (approved):
The changes LGTM and test as described.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-method-test.php
wjrosa: Good call. Done in 30a8797. Committed rather than using the button because the suggestion dropped the `if`'s closing brace, which would have left the braces unbalanced.

changelog.txt
wjrosa: Yes, `Remove` is the right prefix here (matches the giropay entry). Updated in changelog.txt and readme.txt in ef4459978.

includes/payment-methods/class-wc-stripe-upe-payment-method-giropay.php
```diff
@@ -5,6 +5,8 @@
 
 /**
  * The giropay Payment Method class extending UPE base class
+ *
+ * @deprecated 11.0.0 Discontinued by Stripe on 2024-06-30; kept only for refunds on existing orders.
```
jaclync: Might we want to update a couple of "11.0.0" version usage in the docblock to 11.1.0? (Same for a few other files in the PR)

wjrosa: Good catch, thanks! Fixed in #5997.

PR #5835, opened 2026-08-20
Improve the E2E setup flow

diegocurbelo:
Fixes #<github_issue_id>

This PR prevents the E2E environment from starting the Stripe listener when `STRIPE_SECRET_KEY` is incompatible with `stripe listen`. It cannot be used with an OAuth key, it needs a RAK (or legacy shared API Key).

- Use the `wordpress:php7.4` image instead of a custom built one
- Validate the Stripe secret key can be used for webhook listening
    - Run validation from both `test:e2e-setup` and `test:e2e-up`.
    - Fail immediately when validation fails, avoiding confusing webhook-dependent test failures.

It also changes Playwright config when running locally to:
- Use 3 retries (same as the CI env)
    This is for convenience; 99% of the time it's a flaky test that fails, and we want to make sure the full suite runs before pushing and running the tests in GitHub. The only case where this is not ideal is while developing, but in those cases we generally manually run only the "related" test so the impact is very low.
- Auto select workers number (8 workers in current MacBook Pros), CI is still set as 5

## Testing instructions

Make sure all PR checks pass, those checks include the parallel PHPUnit and the E2E suites.

1. Run `npm run test:e2e-setup` with a restricted API key and verify the Stripe listener starts.
2. Repeat with an OAuth key (from your local plugin settings for example) and verify setup warns, skips the listener, and continues.

---

### Changelog entry

<details>

#### Not merchant facing changes

</details>

**Post merge**


tests/e2e/bin/setup.sh
```diff
@@ -107,18 +107,36 @@ if [[ -z "$STRIPE_PUB_KEY" || -z "$STRIPE_SECRET_KEY" ]]; then
...
-    CWD="$CWD" E2E_ROOT="$E2E_ROOT" redirect_output docker compose -p wcstripe-e2e -f "$E2E_ROOT"/env/docker-compose.yml up --build --force-recreate -d
-else
-    CWD="$CWD" E2E_ROOT="$E2E_ROOT" redirect_output docker compose -p wcstripe-e2e --env-file "$E2E_ROOT"/config/local.env -f "$E2E_ROOT"/env/docker-compose.yml up --build --force-recreate -d
+E2E_COMPOSE_ARGS=(-p wcstripe-e2e -f "$E2E_ROOT/env/docker-compose.yml")
+if [ "$CI" != "true" ]; then
+	E2E_COMPOSE_ARGS+=(--env-file "$E2E_ROOT/config/local.env")
 fi
 
+step "Starting E2E docker containers"
+CWD="$CWD" E2E_ROOT="$E2E_ROOT" redirect_output docker compose \
```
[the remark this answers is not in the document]
diegocurbelo: The code contains only one invocation; this is a false positive.

```
CWD="$CWD" E2E_ROOT="$E2E_ROOT" redirect_output docker compose \
	"${E2E_COMPOSE_ARGS[@]}" \
	up --build --force-recreate -d \
	"${E2E_SERVICES[@]}"
```

After replacing the vars:
```
redirect_output docker compose \
	-p wcstripe-e2e -f "/tmp/stripe/tests/e2e/env/docker-compose.yml" \
	up --build --force-recreate -d \
	"wordpress db"
```

tests/e2e/config/playwright.config.js
```diff
@@ -31,10 +31,9 @@ const config = {
 	// Folder for test artifacts such as screenshots, videos, traces, etc
 	outputDir: '../test-results/output',
 
-	/* Retry on CI only */
-	retries: CI ? 3 : 0,
+	retries: 3,
```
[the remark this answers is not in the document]
diegocurbelo: The flakiness is not masked, it's still shown in the run summary and in the webpage report that opens automatically.  And more often that not we want to ensure the test pass, getting an error is usually followed by a retry of that failed test.

Screenshot 2026-08-20 at 19 22 02

Screenshot 2026-08-20 at 19 22 23

tests/e2e/bin/setup.sh
```diff
@@ -107,18 +107,36 @@ if [[ -z "$STRIPE_PUB_KEY" || -z "$STRIPE_SECRET_KEY" ]]; then
...
+if docker run --rm stripe/stripe-cli listen --api-key="$STRIPE_SECRET_KEY" --print-secret > /dev/null 2>&1; then
+	E2E_SERVICES+=(stripe)
+else
+	echo
+	echo -e "\033[0;33mWARNING\033[0m STRIPE_SECRET_KEY cannot be used by the Stripe CLI listener."
+	echo "  The Stripe listener container will not be started. Use a restricted API key instead of an OAuth key to enable it."
+
+	# A previous setup may have left the listener running with different credentials.
+	docker stop wcstripe-e2e-stripe-listener > /dev/null 2>&1 || true
+fi
```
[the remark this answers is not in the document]
diegocurbelo: Changing this makes the code more complex to follow with very little benefits... currently the listener is always started, it just do nothing if the key cannot be used to forward webhook events.

I'm not sure we want to stop the setup in this case, it seems that the same naming problem would happen with the other 2 containers

tests/e2e/env/docker-compose.yml
```diff
@@ -3,16 +3,9 @@ volumes:
   dockerdirectory:
 
 services:
-  stripe:
-    container_name: wcstripe-e2e-stripe-listener
-    image: stripe/stripe-cli
-    restart: unless-stopped
-    depends_on:
-      - wordpress
-    command: "listen --api-key=${STRIPE_SECRET_KEY} --forward-to http://wordpress/?wc-api=wc_stripe"
```
jaclync, line 12: non-blocking: just wondering, is the `stripe:` block move intentional? The code looks the same, and the reordering could add one more step to track git history.

tests/e2e/bin/setup.sh
```diff
@@ -107,18 +107,36 @@ if [[ -z "$STRIPE_PUB_KEY" || -z "$STRIPE_SECRET_KEY" ]]; then
...
+E2E_COMPOSE_ARGS=(-p wcstripe-e2e -f "$E2E_ROOT/env/docker-compose.yml")
+if [ "$CI" != "true" ]; then
+	E2E_COMPOSE_ARGS+=(--env-file "$E2E_ROOT/config/local.env")
 fi
 
+step "Starting E2E docker containers"
+CWD="$CWD" E2E_ROOT="$E2E_ROOT" redirect_output docker compose \
+	"${E2E_COMPOSE_ARGS[@]}" \
+	up --build --force-recreate -d \
+	"${E2E_SERVICES[@]}"
```
jaclync, line 138: non-blocking: both this `setup.sh` and `up.sh` have `step "Starting E2E docker containers"` - do you think it can be shared? Related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5835/changes#r3825614783 to validate the Stripe keys before including the Stripe service.

tests/e2e/bin/setup.sh
```diff
@@ -107,18 +107,36 @@ if [[ -z "$STRIPE_PUB_KEY" || -z "$STRIPE_SECRET_KEY" ]]; then
 	exit 1
 fi
 
+E2E_SERVICES=(wordpress db)
+
+step "Validating Stripe listener credentials"
+if docker run --rm stripe/stripe-cli listen --api-key="$STRIPE_SECRET_KEY" --print-secret > /dev/null 2>&1; then
```
jaclync, line 113: I encountered a false negative with this check on my first run of the branch: my key is a plain `sk_test_*` key, but setup printed `STRIPE_SECRET_KEY cannot be used by the Stripe CLI listener... Use a restricted API key instead of an OAuth key` and skipped the listener. Re-running the exact same `docker run ... stripe listen --print-secret` command a couple of minutes later succeeded and printed the `whsec_ secret` - so the check failed transiently (network/Stripe blip), not because of the key. WDYT of some enhancements here:

- Log the key validation output via `redirect_output` like the rest of the script, so `tests/e2e/e2e-setup.log` shows why it failed.
- Since this also runs in every CI e2e job (where the secrets should always be valid), a transient failure there would silently skip the listener and surface as confusing webhook-dependent spec failures - does it make sense to exit instead of warning and continuing?

jaclync (approved):
Thanks for the improvements to the E2E flow! :shipit: Mostly a comment about a transient failure when validating the Stripe keys.

tests/e2e/bin/setup.sh
diegocurbelo: Fixed in [bf42539](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5835/commits/bf42539e7ae660407b9678f7906c8694ad0e524a)

tests/e2e/bin/setup.sh
diegocurbelo: Now that `E2E_SERVICES` got removed (the script now fails if the key is not valid) I extracted the shared checks in [620aac7](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5835/commits/620aac7dfc5bd9b5dc84467ceb8b22b490b28dce)

tests/e2e/env/docker-compose.yml
diegocurbelo: The listener was the first block, with it at the bottom it's easier to compare with with the main composer file

tests/e2e/bin/setup.sh
```diff
@@ -107,6 +107,15 @@ if [[ -z "$STRIPE_PUB_KEY" || -z "$STRIPE_SECRET_KEY" ]]; then
 	exit 1
 fi
 
+step "Validating Stripe listener credentials"
+# `--print-secret` returns a webhook signing secret on stdout, so only stderr
+# can be persisted safely.
+if ! docker run --rm stripe/stripe-cli listen --api-key="$STRIPE_SECRET_KEY" --print-secret > /dev/null 2>> "$E2E_ROOT/e2e-setup.log"; then
+	error "Stripe listener credential validation failed."
+	echo "  See tests/e2e/e2e-setup.log for details. OAuth keys are not supported by Stripe CLI listeners."
+	exit 1
```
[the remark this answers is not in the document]
diegocurbelo: Updated the PR description, failing the script is the desired behavior

tests/e2e/bin/up.sh
```diff
@@ -3,5 +3,8 @@
 set -e
 . ./tests/e2e/bin/common.sh
 
+load_e2e_local_env
+validate_stripe_listener_credentials
```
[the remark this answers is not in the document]
diegocurbelo: The finding assumes setup should continue without Stripe, but fail-fast is now intentional

PR #5840, opened 2026-08-21
Reconcile stale saved payment tokens when the Stripe customer is missing

wjrosa:
Fixes STRIPE-882

When a customer's Stripe customer object is deleted (or their stored `_stripe_customer_id` is gone), their locally saved payment tokens become unusable, but they could still be displayed at checkout as if they were valid. The token sync (`WC_Stripe_Payment_Tokens::sync_and_retrieve_customer_payment_tokens()`) already reconciles local tokens against Stripe and cleans up stale ones, so for most customers this resolves itself on the first checkout or account view. Two gaps remained, and this PR closes both at the reconciliation layer rather than adding a per-render Stripe call to the display layer (the approach explored and closed in #5590):

1. **Customers at the token cap were never reconciled.** With `posts_per_page` (default 10) or more saved tokens, the sync returned early and skipped verification entirely, so stale cards stayed visible and selectable indefinitely. The cap exists only because the token data store is unpaginated, so it should only prevent *adding* tokens beyond that window. Verification and cleanup of the retrieved tokens now always run; adding new tokens is still skipped at the cap.

2. **A transient Stripe API error could wipe valid tokens.** `WC_Stripe_Customer::get_all_payment_methods()` returned an empty array for both "this customer does not exist" and "the API request failed", and the sync treats an empty list as proof that every local token is stale. A rate limit or auth error at the wrong moment would therefore delete all of a customer's valid saved tokens. The method now accepts an opt-in `$throw_on_error` parameter: a missing customer still returns (and caches) an empty list, which is the authoritative cleanup signal, while generic errors throw so the sync bails out and keeps the local tokens. This becomes more important now that reconciliation also runs for at-cap customers.

Since the sync runs on the `woocommerce_get_customer_payment_tokens` filter, the fix covers every surface that lists saved methods (classic checkout, Blocks checkout, optimized checkout, and My Account), unlike a display-layer check on one gateway hook.

Also guards the error-notice call in the sync's catch block behind a session check: the filter also runs in admin/REST/CLI contexts, where no shopper would see the notice.

**Backward compatibility:** the new `get_all_payment_methods()` parameter is optional and defaults to the existing behavior, so external callers are unaffected. No hooks were added, removed, or re-ordered. The behavior change for at-cap customers is that stale tokens are now deleted (the same cleanup that already applied below the cap); token deletion in this path deliberately skips detaching the payment method from Stripe, as before. Settings and tokens are site-scoped; no multisite behavior changes (not tested under multisite).

## Testing instructions

1. In test mode, as a logged-in customer, save a card at checkout (check "Save payment information to my account").
2. In the Stripe dashboard (test mode), delete that customer object (Customers > select the customer > Actions > Delete).
3. On the store, clear the cached payment methods, e.g. with WP-CLI: `wp transient delete --all` (or wait for the 1-day cache to expire).
4. Visit checkout again as that customer: the saved card must no longer be offered, and the new-card fields must show. The stale token is also gone from My Account > Payment methods.
5. Repeat steps 1–4 with 10 saved cards (save 10 different test cards, e.g. 4242424242424242, 4000056655665556, 5555555555554444, ...). Before this PR the 10 stale cards remained visible; now they are removed as well.
6. Regression check: with a valid customer and saved cards, confirm the cards still display and are usable on classic checkout and Blocks checkout, and that deleting a card from My Account > Payment methods still works.

The transient-API-error path (tokens are kept when Stripe returns e.g. a rate-limit error) is covered by unit tests, as it is impractical to reproduce manually.

---

-   [x] Tested on mobile (or does not apply): server-side token reconciliation, does not apply.

### Changelog entry

Added changelog entries to `changelog.txt` and `readme.txt`.

**Post merge**

### DIFF
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -880,9 +880,9 @@ class WC_Stripe_Customer {
 	 *
 	 * @param string[] $payment_method_types The payment method types to look for using Stripe method IDs. If the array is empty, it implies all payment method types.
 	 * @param int      $limit                The maximum number of payment methods to return. If the value is -1, no limit is applied.
-	 * @param bool     $throw_on_error       Whether to throw on a generic API error instead of returning an empty array. A missing customer still returns an empty array.
+	 * @param bool     $throw_on_error       Throw on generic API errors instead of returning []. A missing customer still returns [].
 	 * @return array
-	 * @throws WC_Stripe_Exception When $throw_on_error is true and the API returns an error other than a missing customer.
+	 * @throws WC_Stripe_Exception When $throw_on_error is true and the error is not a missing customer.
 	 */
 	public function get_all_payment_methods( array $payment_method_types = [], int $limit = -1, bool $throw_on_error = false ) {
 		if ( ! $this->get_id() ) {
@@ -919,9 +919,8 @@ class WC_Stripe_Customer {
 						return [];
 					}
 
-					// A generic API error is not proof the payment methods are gone. Callers that
-					// reconcile local tokens against this list need to distinguish that from a
-					// genuinely empty list, otherwise a transient failure would wipe valid tokens.
+					// A generic error is not proof the payment methods are gone; callers reconciling
+					// local tokens must be able to tell it apart from a genuinely empty list.
 					if ( $throw_on_error ) {
 						throw new WC_Stripe_Exception(
 							print_r( $response->error, true ),
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -659,11 +659,22 @@ class WC_Stripe_Customer {
 	 * error and it is no such customer.
 	 *
 	 * @since 4.1.2
-	 * @param array $error
+	 * @param \stdClass|null $error The error object from a Stripe API response.
+	 * @return bool
 	 */
 	public function is_no_such_customer_error( $error ) {
+		if ( ! $error ) {
+			return false;
+		}
+
+		// Some endpoints reference the customer as a request param and return the
+		// code/param pair; others only carry the message, so both shapes identify
+		// a missing customer.
+		if ( isset( $error->code, $error->param ) && 'resource_missing' === $error->code && 'customer' === $error->param ) {
+			return true;
+		}
+
 		return (
-			$error &&
 			'invalid_request_error' === $error->type &&
 			preg_match( '/No such customer/i', $error->message )
 		);
@@ -674,7 +685,7 @@ class WC_Stripe_Customer {
 	 * error and it is no such customer.
 	 *
 	 * @since 4.5.6
-	 * @param array $error
+	 * @param \stdClass|null $error The error object from a Stripe API response.
 	 * @return bool
 	 */
 	public function is_source_already_attached_error( $error ) {
@@ -909,11 +920,7 @@ class WC_Stripe_Customer {
 				$response = WC_Stripe_API::request( $request_params, 'payment_methods?expand[]=data.sepa_debit.generated_from.charge&expand[]=data.sepa_debit.generated_from.setup_attempt', 'GET' );
 
 				if ( ! empty( $response->error ) ) {
-					if (
-						isset( $response->error->param, $response->error->code )
-						&& 'customer' === $response->error->param
-						&& 'resource_missing' === $response->error->code
-					) {
+					if ( $this->is_no_such_customer_error( $response->error ) ) {
 						// If the customer doesn't exist, cache an empty array.
 						set_transient( $cache_key, [], DAY_IN_SECONDS );
 						return [];
--- a/includes/class-wc-stripe-customer.php
+++ b/includes/class-wc-stripe-customer.php
@@ -880,9 +880,11 @@ class WC_Stripe_Customer {
 	 *
 	 * @param string[] $payment_method_types The payment method types to look for using Stripe method IDs. If the array is empty, it implies all payment method types.
 	 * @param int      $limit                The maximum number of payment methods to return. If the value is -1, no limit is applied.
+	 * @param bool     $throw_on_error       Whether to throw on a generic API error instead of returning an empty array. A missing customer still returns an empty array.
 	 * @return array
+	 * @throws WC_Stripe_Exception When $throw_on_error is true and the API returns an error other than a missing customer.
 	 */
-	public function get_all_payment_methods( array $payment_method_types = [], int $limit = -1 ) {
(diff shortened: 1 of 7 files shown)


includes/class-wc-stripe-customer.php
jaclync: non-blocking, pre-existing: the missing customer check is different from this class' helper `is_no_such_customer_error`. If it's the same kind of error check, might be worth consolidating the error checking helper.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -234,7 +232,9 @@ private function sync_and_retrieve_customer_payment_tokens( array $tokens, int $
 			// Prevent unnecessary recursion, WC_Payment_Token::save() ends up calling 'woocommerce_get_customer_payment_tokens' in some cases.
 			remove_filter( 'woocommerce_get_customer_payment_tokens', [ $this, 'woocommerce_get_customer_payment_tokens' ], 10 );
 
-			$payment_methods    = $customer->get_all_payment_methods( $active_reusable_types );
+			// Throw on generic API errors so a transient failure bails out and keeps the local tokens;
+			// only a missing customer returns the empty list that authorizes cleanup below.
+			$payment_methods    = $customer->get_all_payment_methods( $active_reusable_types, -1, true );
```
jaclync, line 237: non-blocking question: `get_all_payment_methods` has another call site in the deprecated sibling public function `woocommerce_get_customer_upe_payment_tokens`, do you think it's worth updating the call to throw on API error there as well?

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -276,9 +276,11 @@ private function sync_and_retrieve_customer_payment_tokens( array $tokens, int $
 				}
 
 				// Create a new token when:
+				// - The token count is below the retrieval cap.
 				// - The payment method is a valid PaymentMethodID (i.e. only support IDs starting with "src_" when using the card payment method type).
 				// - The payment method belongs to the gateway ID being retrieved or the gateway ID is empty (meaning we're looking for all payment methods).
 				if (
+					$can_add_new_tokens &&
```
jaclync, line 283: For learning purposes: I got stuck reading this "sync" function, which fetches *all* tokens remotely from Stripe but only prunes *the first page (default 10)* of persisted tokens - what are the reasons for not pruning *all* the persisted tokens? If the customer has 10+ payment methods, if ones beyond the first page get outdated, they never get pruned. Just wanted to check if I misunderstood something.

My guess is that the `posts_per_page` limit of the tokens is for UI display, as we don't want to show unlimited number of saved payment methods in the checkout page. However, I thought the `posts_per_page` cap will happen after all the tokens have been synced/pruned.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -314,7 +316,10 @@ private function sync_and_retrieve_customer_payment_tokens( array $tokens, int $
 				}
 			}
 		} catch ( WC_Stripe_Exception $e ) {
-			wc_add_notice( $e->getLocalizedMessage(), 'error' );
+			// wc_add_notice() only guards against a missing session (admin, REST, CLI) since WC 10.5.
+			if ( WC()->session ) {
```
jaclync: non-blocking nit: as the plugin requires WC 10.8+ as in `WC_STRIPE_MIN_WC_VER`, do we still need this guard?

jaclync (approved):
LGTM, worked as expected :shipit: 

Just a note that in case of `get_all_payment_methods` API error, an error banner is also shown on the checkout page in addition to the logged-in user's payment methods page - should be fine, it probably means some network/Stripe errors that might affect checkout down the line anyway.

Image

jaclync:
Also a note on PR description:

> Also guards the error-notice call in the sync's catch block behind a session check, since the filter also runs in admin/REST/CLI contexts and wc_add_notice() only protects against a missing session itself since WC 10.5 (this plugin supports older versions).

The plugin supports WC 10.8+, I don't think older versions are supported? The guard could still be useful for calling it in use cases that don't have a WC session.

includes/class-wc-stripe-customer.php
wjrosa: Good idea. Consolidated in 91c6d8f57: the helper now also matches the resource_missing/customer code-param pair, and this site uses it (with a small data-provider test).

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: Done in 91c6d8f57. It is safe there because the whole function already runs inside a try/catch for `WC_Stripe_Exception`, so on an API error it now keeps the local tokens instead of deleting them.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
wjrosa: You read it right. The cap comes from WC core: the token data store is unpaginated and loads only the first `posts_per_page` tokens, and this filter only receives/returns that list. Those are also the only tokens WooCommerce renders (checkout and My Account read through the same capped call), so a stale token beyond the cap cannot be selected. Once the count drops below the cap it enters the loaded list and gets pruned then. Pre-existing behavior, not changed by this PR.

wjrosa:
Good catch, thanks! We require WC 10.8+, so that rationale was stale. I kept the guard (in sessionless contexts nobody would see the notice) and corrected the code comment in 1cf0b5fec and the PR description.

includes/payment-tokens/class-wc-stripe-payment-tokens.php
```diff
@@ -540,7 +545,9 @@ public function woocommerce_get_customer_upe_payment_tokens( $tokens, $user_id,
 					$active_reusable_payment_method_types[] = WC_Stripe_UPE_Payment_Method_Sepa::STRIPE_ID;
 				}
 			}
-			$payment_methods = $customer->get_all_payment_methods( $active_reusable_payment_method_types );
+			// Throw on generic API errors (caught below) so a transient failure keeps the
+			// local tokens instead of deleting them against an empty list.
+			$payment_methods = $customer->get_all_payment_methods( $active_reusable_payment_method_types, -1, true );
```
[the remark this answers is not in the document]
wjrosa: Not needed for safety: `wc_add_notice()` returns early on a missing session since WC 10.5, and this plugin requires WC 10.8+, so a sessionless caller gets a no-op there, not a failure. The guard in the main sync path is about intent (skip a pointless notice), and this method is deprecated and not hooked in-tree, so I will leave it as is.

PR #5841, opened 2026-08-21
Recreate the Checkout Session when the shopper's login state changes

### LINKED ISSUE #5836: Checkout Session created while logged out is reused after login, so `savePaymentMethod` is rejected and the customer cannot pay by card (10.9.0)
**Plugin version:** 10.9.0 (regression appeared on upgrade from 10.8.5)
**WooCommerce:** 11.0.1 · **WordPress:** 7.1 · **PHP:** 8.3 · Classic (shortcode) checkout
**Settings:** Adaptive Pricing enabled, Optimized Checkout Element enabled, saved cards enabled, single currency (AUD), Stripe Connect
**Stripe API version:** `2026-03-25.dahlia`

## Summary

When Adaptive Pricing is enabled, a Stripe Checkout Session is created on the checkout page. If
the shopper is a **guest** at that moment, the session is created without `customer` and without
`saved_payment_method_options`. If they then **log in or create an account**, the plugin
deliberately preserves ownership of that guest session and `synchronize()` reuses it, because the
cart amount and currency have not changed.

The page now reports `isLoggedIn: true`, so the checkout script sends `savePaymentMethod` on
confirm. Stripe rejects it:

```
savePaymentMethod cannot be used unless the Checkout Session was created with
saved_payment_method_options[payment_method_save]=enabled
```

The message is surfaced as a checkout notice. **The customer cannot pay by card at all** — the
Checkout Session stays `open`, no PaymentIntent is ever created, and retrying does not help
because the session is reused until the cart total changes.

## Root cause

`saved_payment_method_options` is attached **only** when the session is created while logged in —
`includes/class-wc-stripe-checkout-session-manager.php:80-95`:

```php
if ( is_user_logged_in() && WC()->customer instanceof WC_Customer ) {
    ...
    $request['customer']                     = $stripe_customer->get_id();
    $request['saved_payment_method_options'] = [
        'payment_method_save' => 'enabled',
    ];
}
```

The same block appears in
`includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php:70-80`.

`synchronize()` (`class-wc-stripe-checkout-session-manager.php:189-216`) reuses the stored session
whenever the context amount and currency still match the cart and `status === 'success'`. **It does
not consider whether the login state has changed since the session was created**, so it never
recreates the session to add the missing parameter.

The guest session's ownership is explicitly carried across the login by
`WC_Stripe_Checkout_Session_Context::remember_migrated_guest_session()`
(`includes/class-wc-stripe-checkout-session-context.php:148`), hooked to
`woocommerce_guest_session_to_user_id`, which stores `wc_stripe_migrated_guest_session_id` and adds
it to the accepted owner keys at `:427-429`. So validation passes and the stale session is used.

Meanwhile the client sends the parameter purely on the localized login flag —
`build/upe-classic.js`, mirroring `isLoggedIn` from
`includes/payment-methods/class-wc-stripe-upe-payment-gateway.php:558`:

```js
Z()?.isLoggedIn && (_.savePaymentMethod = $(t).find('#wc-stripe-new-payment-method').is(':checked'));
```

There is no check that the current Checkout Session was created with
`saved_payment_method_options` enabled.

## Reproduction

1. Enable Adaptive Pricing, Optimized Checkout Element and saved cards. Classic checkout.
2. As a **logged-out** guest, add a product to the cart and open `/checkout/`. A Checkout Session
   is created — confirm via the API that it has `customer: null` and
   `saved_payment_method_options: null`.
3. **Log in** (or create an account) without changing the cart contents.
4. Return to `/checkout/`, choose card, and tick **"Save payment method"**.
5. Place the order → the error above is displayed and no payment is taken.

## Evidence from a live store

Five distinct customers hit this within the first day after the 10.9.0 upgrade. Each was matched
from the stored `wc_stripe_checkout_session` in their WooCommerce session to the Stripe API. Every
one of the five sessions was created as a guest, and every one of those five sessions is present in
a session record that also carries `wc_stripe_migrated_guest_session_id` — i.e. a guest session
that was migrated to a logged-in user:

| Customer | Session created (UTC) | `customer` | `saved_payment_method_options` | Cart | Session status | Outcome |
|---|---|---|---|---|---|---|
| A | 2026-08-20 02:00:42 | `null` | `null` | $409.15 | `open` / `unpaid`, `payment_intent: null` | Gave up on card, placed the order on bank transfer 3½ min later — **still unpaid** |
| B | 2026-08-20 02:54:35 | `null` | `null` | $108.64 | `open` / `unpaid` | Paid via PayPal 2 min later |
| C | 2026-08-20 06:37:19 | `null` | `null` | $55.57 | `open` / `unpaid` | Paid via Afterpay 2 min later |
| D | 2026-08-20 09:24:28 | `null` | `null` | $16.24 | `complete` / `paid` | **Paid by card** |
| E | 2026-08-20 21:01:52 | `null` | `null` | $98.31 | `open` / `unpaid` | Paid via Zip 2½ min later |

Customer **D** is the useful control and pins down the trigger precisely: their session was
*equally* guest-created, but the save-card box was left unticked, so `savePaymentMethod: false` was
sent — and Stripe **accepted** it. So the API only refuses when the value is `true`. The failure
therefore requires all three of: session created as a guest, subsequent login, and the customer
ticking "Save payment method".

Four of the five could not use a card. One order is still unpaid; three were diverted onto
higher-fee gateways.

## Why it appeared on upgrade

This store had previously had Adaptive Pricing switched off automatically by the plugin. The
10.9.0 upgrade routine turned it back on:

```
2026-08-19T18:50:36+00:00 ERROR Adaptive Pricing re-enabled during plugin update after a previous
automatic disable caused by a Checkout Session amount mismatch.
CONTEXT: {"stripe_version":"10.9.0","previous_version":"10.8.5", ...
  class-wc-stripe-restore-adaptive-pricing-after-amount-mismatch-update.php:63 ...}
```

The first customer failure was 7 hours later. Two observations on that migration, offered
separately from the main bug:

- It re-enables a feature that the plugin's own safety logic had disabled on this site, with no
  evidence that the condition which triggered the disable had been resolved.
- `WC_Stripe_Helper::is_adaptive_pricing_supported()` still consults
  `WC_Stripe_Checkout_Session_Context::was_amount_mismatch_detected()` per session, so the
  per-session guard survives — but the persisted store-level decision was overwritten.

## Suggested fixes

Either would be sufficient; the first is the smaller change.

1. **Record how the session was created and honour it on the client.** Include a flag such as
   `savePaymentMethodAvailable` in the session data already passed to the client via the
   `#wc-stripe-checkout-session-data` fragment / Store API extension, set from whether
   `saved_payment_method_options` was sent, and gate the JS on that instead of on `isLoggedIn`.
   This also fixes the inverse case cleanly and needs no extra Stripe API calls.

2. **Recreate the session when the login state changes.** In `synchronize()`, compare
   `is_user_logged_in()` against whether the stored session was created with a customer, and call
   `create_session()` rather than reusing when they disagree. `remember_migrated_guest_session()`
   is the natural place to mark the record stale, since it already observes the transition.

A defensive third: treat `savePaymentMethod: false` as "omit the parameter" client-side, which
would shrink the blast radius even where the flag is wrong.

## Workaround in place

Adaptive Pricing disabled (`adaptive_pricing = no`), which turns the Checkout Sessions path off
entirely and reverts checkout to the PaymentIntent flow. Verified: no session fragment is emitted,
so no session id is posted, so `process_payment()` skips
`process_payment_with_checkout_session()`, and `isAdaptivePricingEnabled` is false so the script
uses `stripe.confirmPayment`.

wjrosa:
Fixes STRIPE-1400
Fixes #5836

With Adaptive Pricing enabled, a Stripe Checkout Session created while the shopper is logged out is created without a `customer` and without `saved_payment_method_options`, and Stripe does not allow adding them after creation. When the shopper then logs in mid-checkout, the session's ownership is deliberately migrated with the WooCommerce session and `synchronize()` keeps reusing it as long as the cart amount and currency are unchanged. Meanwhile both confirm paths (classic and Blocks) sent `savePaymentMethod` based purely on `isLoggedIn`, so Stripe rejected the confirm with `savePaymentMethod cannot be used unless the Checkout Session was created with saved_payment_method_options[payment_method_save]=enabled` — and the shopper could not pay by card at all until the cart total changed.

Two complementary layers:

1. **Recreate on login-state change (the actual fix).** The session record now stores whether the session was created with payment method saving enabled, and `synchronize()` recreates the session whenever that no longer matches the current login state. This covers both directions (guest → logged in, logged in → guest) and restores the ability to actually save the card, since the recreated session carries the Stripe customer. Records persisted before this change lack the flag, so a logged-in shopper's first sync after updating recreates the session once — which also self-heals any store currently stuck in the broken state.
2. **Gate the client on session capability (defense in depth).** The session's public data (classic fragment + Store API extension) now includes `save_payment_method_enabled`, and both confirm paths gate `savePaymentMethod` on it instead of on `isLoggedIn`, so the parameter can never be sent to a session that cannot accept it.

**Backward compatibility impact:** additive only — a new `save_payment_method_enabled` boolean field on the classic `#wc-stripe-checkout-session-data` fragment payload and the `wc-stripe/checkout-session` Store API extension schema, and a new key in the session record stored in the WooCommerce session. No hooks, option keys, or PHP signatures changed; no existing fields were renamed, removed, or reordered. Session records written by previous versions are handled gracefully (missing flag reads as `false`, triggering at most one extra session creation for logged-in shoppers). Site-scoped state only; not tested under multisite, but no state crosses sites.

## Testing instructions

1. Enable the Optimized Checkout Element and Adaptive Pricing (requires an eligible account), disable Link, and enable saved cards. Use classic (shortcode) checkout.
2. As a logged-out guest, add a product to the cart and open the checkout page. A Checkout Session is created (visible in the Stripe dashboard as unpaid, with no customer).
3. Log in (or create an account) without changing the cart contents, and return to the checkout page.
4. Choose card, tick **Save payment method**, and place the order.
5. Before this change: checkout fails with the `savePaymentMethod cannot be used…` error and no payment is taken. After this change: a new Checkout Session is created on the post-login page load, the payment succeeds, and the card is saved to the account.
6. Repeat on Blocks checkout (steps 2–5).
7. Regression check while logged in throughout: place an order with the save box ticked (card saved) and unticked (card not saved), on both classic and Blocks checkout.

Automated coverage: new PHPUnit tests for `synchronize()` recreating the session on login and logout transitions (including the request payload with/without `saved_payment_method_options`), and Jest tests for both confirm paths verifying `savePaymentMethod` is gated on the session capability flag.

---

### Changelog entry

<details>

#### Comment

Changelog entry added to `changelog.txt` and `readme.txt` in this PR:

> Fix - Recreate the Adaptive Pricing Checkout Session when a shopper logs in or out mid-checkout, so card payments with "Save payment method" no longer fail after a guest logs in

</details>

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,7 +23,7 @@ xxxx-xx-xx - version 11.0.0
 * Dev - Make GITHUB_TOKEN optional for the local E2E Docker setup by falling back to the GitHub CLI and then to plugin zips placed in tests/e2e/deps
 * Fix - Reload the plugin's payment method settings when refreshing account details, so the settings screen no longer keeps showing stale values
 * Update - Fetch express checkout AJAX nonces on demand instead of embedding them in every page
-* Fix - Recreate the Adaptive Pricing Checkout Session when a shopper logs in or out mid-checkout, so card payments with "Save payment method" no longer fail after a guest logs in
+* Fix - Recreate the Adaptive Pricing Checkout Session when a shopper logs in or out mid-checkout, so payment methods can be saved correctly
 * Fix - Keep the wallet billing and shipping addresses on express checkout orders when the address normalization request returns an unusable response
 * Fix - Render the Adaptive Pricing payment element in the store's Stripe locale
 * Fix - Stop sending Level 3 data when paying with a non-card payment method through express checkout (e.g. Amazon Pay), which Stripe rejects and can disable Level 3 data for card payments
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,6 +20,7 @@ xxxx-xx-xx - version 11.0.0
 * Dev - Make GITHUB_TOKEN optional for the local E2E Docker setup by falling back to the GitHub CLI and then to plugin zips placed in tests/e2e/deps
 * Fix - Reload the plugin's payment method settings when refreshing account details, so the settings screen no longer keeps showing stale values
 * Update - Fetch express checkout AJAX nonces on demand instead of embedding them in every page
+* Fix - Recreate the Adaptive Pricing Checkout Session when a shopper logs in or out mid-checkout, so card payments with "Save payment method" no longer fail after a guest logs in
 
 2026-08-17 - version 10.9.0
 * Fix - Re-enable Adaptive Pricing if it was automatically disabled after a Checkout Session amount mismatch
--- a/client/blocks/checkout-sessions/__tests__/hooks.test.js
+++ b/client/blocks/checkout-sessions/__tests__/hooks.test.js
@@ -647,10 +647,9 @@ describe( 'CheckoutSessions hook tests', () => {
 				shippingData
 			);
 			await onCheckoutSuccessResultPromise;
-			expect( confirm ).toHaveBeenCalledWith(
-				expect.not.objectContaining( {
-					savePaymentMethod: expect.anything(),
-				} )
+			expect( confirm ).toHaveBeenCalledTimes( 1 );
+			expect( confirm.mock.calls[ 0 ][ 0 ] ).not.toHaveProperty(
+				'savePaymentMethod'
 			);
 		} );
 	} );
--- a/client/blocks/checkout-sessions/__tests__/hooks.test.js
+++ b/client/blocks/checkout-sessions/__tests__/hooks.test.js
@@ -624,6 +624,35 @@ describe( 'CheckoutSessions hook tests', () => {
 				} )
 			);
 		} );
+
+		it( 'omits savePaymentMethod when the session does not support saving, even with the checkbox checked', async () => {
+			document.body.innerHTML = `
+				<div class="wc-block-components-payment-methods__save-card-info">
+					<input type="checkbox" checked />
+				</div>
+			`;
+			const confirm = jest.fn().mockResolvedValue( {
+				type: 'success',
+			} );
+			const checkoutState = {
+				type: 'success',
+				checkout: { email: '', confirm },
+			};
+			useCheckoutSuccessHandler(
+				checkoutState,
+				onCheckoutSuccess,
+				billing,
+				false,
+				false,
+				shippingData
+			);
+			await onCheckoutSuccessResultPromise;
+			expect( confirm ).toHaveBeenCalledWith(
+				expect.not.objectContaining( {
+					savePaymentMethod: expect.anything(),
+				} )
+			);
+		} );
 	} );
 
 	describe( 'usePaymentFailHandler hook', () => {
--- a/includes/class-wc-stripe-checkout-session-manager.php
+++ b/includes/class-wc-stripe-checkout-session-manager.php
@@ -79,7 +79,8 @@ class WC_Stripe_Checkout_Session_Manager {
 			$request['phone_number_collection'] = [ 'enabled' => 'true' ];
 		}
 
-		if ( is_user_logged_in() && WC()->customer instanceof WC_Customer ) {
+		$save_payment_method_enabled = $this->can_enable_save_payment_method();
+		if ( $save_payment_method_enabled ) {
 			try {
 				// Creation happens before checkout fields are complete, so only the saved account identity can be used here.
 				$stripe_customer = new WC_Stripe_Customer( WC()->customer->get_id() );
@@ -119,11 +120,12 @@ class WC_Stripe_Checkout_Session_Manager {
 		// the source of truth for the amount and currency. That code controls the data sent
 		// to Stripe and is used to validate amounts later.
 		$record = [
-			'session_id'    => (string) $checkout_session->id,
-			'client_secret' => (string) $checkout_session->client_secret,
-			'revision'      => $this->generate_revision_token(),
-			'status'        => 'success',
-			'message'       => '',
+			'session_id'                  => (string) $checkout_session->id,
+			'client_secret'               => (string) $checkout_session->client_secret,
+			'revision'                    => $this->generate_revision_token(),
+			'status'                      => 'success',
+			'message'                     => '',
+			'save_payment_method_enabled' => $save_payment_method_enabled,
 		];
 		$this->set_record( $record );
 
@@ -203,6 +205,12 @@ class WC_Stripe_Checkout_Session_Manager {
 			return $this->create_session();
 		}
 
+		// Stripe cannot add or remove saved_payment_method_options after creation, so a
+		// login-state change (e.g. a migrated guest session) requires a new session.
+		if ( ! empty( $record['save_payment_method_enabled'] ) !== $this->can_enable_save_payment_method() ) {
+			return $this->create_session();
+		}
+
 		// The context is the source of truth for amount and currency, so we need to compare
 		// against the data that was pushed to Stripe.
 		if (
@@ -238,6 +246,16 @@ class WC_Stripe_Checkout_Session_Manager {
 		$this->set_record( $record );
 
 		return $this->get_public_data( $record );
+	}
+
+	/**
+	 * Whether create_session() would attach a customer and saved_payment_method_options
+	 * for the current request.
+	 *
+	 * @return bool
+	 */
+	private function can_enable_save_payment_method(): bool {
+		return is_user_logged_in() && WC()->customer instanceof WC_Customer;
 	}
 
 	/**
@@ -354,11 +372,12 @@ class WC_Stripe_Checkout_Session_Manager {
 	 */
 	private function get_public_data( ?array $record ): array {
 		return [
-			'session_id'    => (string) ( $record['session_id'] ?? '' ),
-			'client_secret' => (string) ( $record['client_secret'] ?? '' ),
-			'revision'      => (string) ( $record['revision'] ?? '' ),
-			'status'        => (string) ( $record['status'] ?? 'uninitialized' ),
-			'message'       => (string) ( $record['message'] ?? '' ),
+			'session_id'                  => (string) ( $record['session_id'] ?? '' ),
+			'client_secret'               => (string) ( $record['client_secret'] ?? '' ),
+			'revision'                    => (string) ( $record['revision'] ?? '' ),
+			'status'                      => (string) ( $record['status'] ?? 'uninitialized' ),
+			'message'                     => (string) ( $record['message'] ?? '' ),
+			'save_payment_method_enabled' => ! empty( $record['save_payment_method_enabled'] ),
 		];
 	}
 }
--- a/includes/class-wc-stripe-checkout-session-manager.php
+++ b/includes/class-wc-stripe-checkout-session-manager.php
@@ -207,7 +207,16 @@ class WC_Stripe_Checkout_Session_Manager {
 
 		// Stripe cannot add or remove saved_payment_method_options after creation, so a
 		// login-state change (e.g. a migrated guest session) requires a new session.
-		if ( ! empty( $record['save_payment_method_enabled'] ) !== $this->can_enable_save_payment_method() ) {
+		$can_save_payment_method = $this->can_enable_save_payment_method();
+		if ( ! empty( $record['save_payment_method_enabled'] ) !== $can_save_payment_method ) {
+			WC_Stripe_Logger::info(
+				'Recreating the Checkout Session after a login-state change.',
+				[
+					'checkout_session_id'         => (string) $record['session_id'],
+					'save_payment_method_enabled' => $can_save_payment_method,
+				]
+			);
+
 			return $this->create_session();
 		}
 
--- a/includes/class-wc-stripe-checkout-session-manager.php
+++ b/includes/class-wc-stripe-checkout-session-manager.php
@@ -249,8 +249,9 @@ class WC_Stripe_Checkout_Session_Manager {
 	}
 
 	/**
-	 * Whether create_session() would attach a customer and saved_payment_method_options
-	 * for the current request.
+	 * Will a newly created session allow a saved payment method to be created.
+	 * This is only available for a logged-in user with an existing WC_Customer definition
+	 * at the start of a session.
 	 *
 	 * @return bool
 	 */
--- a/tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
+++ b/tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
@@ -305,14 +305,22 @@ class WC_Stripe_Checkout_Session_Manager_Test extends WP_UnitTestCase {
 	}
 
 	/**
-	 * A guest-created session cannot accept savePaymentMethod, so a login must recreate it
-	 * rather than reuse the migrated guest one.
+	 * Run synchronize() across a login-state transition with the Stripe API mocked.
+	 *
+	 * Synchronizes once in the starting state, flips the login state (using a
+	 * freshly created user), then synchronizes twice more to cover both the
+	 * recreation and the subsequent reuse of the new session.
+	 *
+	 * @param bool   $start_logged_in Whether the first synchronization happens while logged in.
+	 * @param string $prefix          Fixture prefix for the mocked Stripe identifiers.
+	 * @return array The three synchronize() results, the captured create requests, and the create count.
 	 */
-	public function test_synchronize_recreates_session_when_guest_logs_in(): void {
+	private function run_synchronize_login_transition( bool $start_logged_in, string $prefix ): array {
 		$product = WC_Helper_Product::create_simple_product( true, [ 'regular_price' => 42 ] );
 		WC()->cart->add_to_cart( $product->get_id(), 1 );
 		WC()->cart->calculate_totals();
 
+		$user_id           = self::factory()->user->create();
 		$create_count      = 0;
 		$captured_creates  = [];
 		$original_customer = WC()->customer;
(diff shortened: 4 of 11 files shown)


includes/class-wc-stripe-checkout-session-manager.php
```diff
@@ -240,6 +248,16 @@ public function mark_failed( string $message ): array {
 		return $this->get_public_data( $record );
 	}
 
+	/**
+	 * Whether create_session() would attach a customer and saved_payment_method_options
+	 * for the current request.
```
daledupreez: I don't think this description is very helpful or clear.
```suggestion
	 * Will a newly created session allow a saved payment method to be created.
	 * This is only available for a logged-in user with an existing WC_Customer definition
	 * at the start of a session.
```

changelog.txt
```diff
@@ -20,6 +20,7 @@ xxxx-xx-xx - version 11.0.0
 * Dev - Make GITHUB_TOKEN optional for the local E2E Docker setup by falling back to the GitHub CLI and then to plugin zips placed in tests/e2e/deps
 * Fix - Reload the plugin's payment method settings when refreshing account details, so the settings screen no longer keeps showing stale values
 * Update - Fetch express checkout AJAX nonces on demand instead of embedding them in every page
+* Fix - Recreate the Adaptive Pricing Checkout Session when a shopper logs in or out mid-checkout, so card payments with "Save payment method" no longer fail after a guest logs in
```
daledupreez: I think we should simplify the description.
```suggestion
* Fix - Recreate the Adaptive Pricing Checkout Session when a shopper logs in or out mid-checkout, so payment methods can be saved correctly
```

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
```diff
@@ -304,6 +304,158 @@ public function test_synchronize_repairs_a_context_that_drifted_from_the_cart():
...
+			];
+		};
+		add_filter( 'wc_stripe_request_body', $capture_body, 10, 2 );
+		add_filter( 'pre_http_request', $mock_request, 10, 3 );
+
+		try {
+			$manager  = new WC_Stripe_Checkout_Session_Manager();
+			$as_guest = $manager->synchronize();
+
+			wp_set_current_user( 1 );
```
daledupreez: Is user 1 guaranteed to exist? I think it would be better to create a user dynamically rather than expecting one will exist with a specific user ID.

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
```diff
@@ -304,6 +304,158 @@ public function test_synchronize_repairs_a_context_that_drifted_from_the_cart():
...
+					]
+				),
+			];
+		};
+		add_filter( 'wc_stripe_request_body', $capture_body, 10, 2 );
+		add_filter( 'pre_http_request', $mock_request, 10, 3 );
+
+		try {
+			wp_set_current_user( 1 );
+			WC()->customer = new WC_Customer( 1 );
```
daledupreez: This is also relying on user ID 1 existing, which feels brittle to me.

Also, the main code for the two tests is pretty much the same except for the order in which the user IDs are marked as the current user. It feels like that could all be pushed into a helper/runner method.

includes/class-wc-stripe-checkout-session-manager.php
```diff
@@ -203,6 +205,12 @@ public function synchronize(): array {
 			return $this->create_session();
 		}
 
+		// Stripe cannot add or remove saved_payment_method_options after creation, so a
+		// login-state change (e.g. a migrated guest session) requires a new session.
+		if ( ! empty( $record['save_payment_method_enabled'] ) !== $this->can_enable_save_payment_method() ) {
+			return $this->create_session();
```
daledupreez, line 220: I think this is a pre-existing issue with other branches in this function, but the recreation logic could fail. But I think that failure mode should be better because checkout sessions won't take effect. The one downside is that for logged-in shoppers who had a record _and_ the recreate fails (which should be a small population), they will end up with an error state when the previous state would work fine.

Also, this operation doesn't lock the session, though it's not clear if that makes sense.

client/blocks/checkout-sessions/__tests__/hooks.test.js
```diff
@@ -624,6 +624,35 @@ describe( 'CheckoutSessions hook tests', () => {
...
+				billing,
+				false,
+				false,
+				shippingData
+			);
+			await onCheckoutSuccessResultPromise;
+			expect( confirm ).toHaveBeenCalledWith(
+				expect.not.objectContaining( {
+					savePaymentMethod: expect.anything(),
+				} )
```
daledupreez: My AI tool flagged that `expect.anything()` will match anything except `null` or `undefined`. Maybe we can verify that the passed object does not have the `savePaymentMethod` key?

daledupreez (commented):
I have some feedback from code inspection. I have not had a chance to test yet.

daledupreez:
And some additional AI comments:

### OC save-checkbox visibility still keys on `isLoggedIn` while confirm keys on the session flag — LOW

`client/optimized-checkout/handle-display-of-saving-checkbox.js:161`

No agent raised this; it surfaced in the verification pass. The classic-without-OC branch shows the save-card container when `getStripeServerData()?.isLoggedIn || signupSelected`, but confirm now sends `savePaymentMethod` only when `save_payment_method_enabled` is true. Wherever those two diverge the shopper sees a tickable "Save payment method" checkbox that is silently inert.

In the happy path they converge (the recreate flips the flag true for a logged-in shopper), so the divergence window is narrow — a logged-in shopper whose `WC()->customer` isn't a `WC_Customer`, or the token path in finding 2. Worth aligning the visibility gate with the same flag so the two can't drift.

### No test exercises the legacy-record shape with the missing key — LOW

`tests/phpunit/class-wc-stripe-checkout-session-manager-test.php:311`

The new tests seed records written by the *new* `create_session()`, which always writes the key; no test seeds a record that omits it — the actual pre-upgrade shape behind the self-healing claim.

### Session recreation on login-state change has no distinguishing log signal — LOW

`includes/class-wc-stripe-checkout-session-manager.php:210`

The new branch calls `create_session()` silently, and failures share the generic lines at `class-wc-stripe-checkout-session-lifecycle.php:113`/`:187` with no field naming the cause. There's precedent for doing better: `invalidate_session_after_amount_mismatch()` logs its case at error level for exactly this diagnosability reason. Given the original bug went undetected until five customers hit it in a day — precisely because nothing signalled a session being reused in a broken state — a `reason=login_state_change` context field would pay for itself.

### AJAX endpoint duplicates the save-payment-method logic and never persists the flag — LOW

`includes/class-wc-stripe-checkout-session-manager.php:128` (evidence in `includes/ajax-handlers/class-wc-stripe-checkout-sessions-ajax-handler.php`)

Multiple review threads flagged this, so it's worth noting: `WC_Stripe_Checkout_Sessions_Ajax_Handler::create_checkout_session()` holds a near-identical predicate and issues the same kind of request, but writes no manager record, so a session created there can never carry the flag. It is **not** a hole in your fix — verified: `checkoutSessionsCreateSession()`/`checkoutSessionsUpdateSession()` in `client/api/index.js` are referenced only from Jest files, never from production code, so no shipped client triggers it. The hooks are still registered in `WC_Stripe::init()`, so the endpoint remains reachable by direct request.

**What if we** either delete the unreachable endpoints and their nonces as a follow-up, or have the handler delegate to `WC_Stripe_Checkout_Session_Manager::create_session()`? Either is fine as a separate PR.

_Written using Claude._

includes/class-wc-stripe-checkout-session-manager.php
wjrosa: Applied.

changelog.txt
wjrosa: Applied, and aligned readme.txt with the same wording in 620d0b6d0.

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
wjrosa: Fixed in 055711c4f, the tests now create the user via the factory.

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
wjrosa: Done in 055711c4f, both tests now share a run_synchronize_login_transition() helper.

client/blocks/checkout-sessions/__tests__/hooks.test.js
wjrosa: Good catch. Replaced with a direct toHaveProperty check on the confirm args in 055711c4f.

includes/class-wc-stripe-checkout-session-manager.php
wjrosa: Agreed. A failed recreate goes through the same handling as the other branches: the lifecycle callers catch it, mark_failed() keeps the last usable identifiers, and the next refresh retries. I left locking out of scope since none of the synchronize() paths lock today.

daledupreez:
@wjrosa, I've gone through the basic regression tests, but I want to make sure I am doing the right thing, and the test instructions are somewhat ambiguous, especially these steps:

> 4. Choose card, tick **Save payment method**, and place the order.
> 7. Regression check while logged in throughout: place an order with the save box ticked (card saved) and unticked (card not saved), on both classic and Blocks checkout.

Just to be 100% clear, we show the Link options for saving payment methods, not the standard save payment method box, at least not on my test system where AP is still enabled. Are you describing the Link-based flow or the checkbox that we render in the plugin?

wjrosa:
Thanks! Addressed the two in-scope ones in 0e44dc91f: an info log with the session id when a login-state change forces a recreation, and a test seeding a pre-upgrade record without the key (a guest keeps reusing it, a login recreates it).

On the checkbox visibility: the `isLoggedIn` branch at line 161 only runs on classic without OC, where Checkout Sessions never run (`is_adaptive_pricing_enabled()` requires `should_render_optimized_checkout()`), so the realistic divergence is the OC config-map branch with signup-at-checkout. Agreed it is narrow; I would treat aligning it with the session flag as follow-up hardening.

On the AJAX endpoint: agreed as a separate PR. I would lean toward delegating to `WC_Stripe_Checkout_Session_Manager::create_session()` rather than deleting, since the registered `wc_ajax_*` hooks are public surface.

wjrosa:
> Just to be 100% clear, we show the Link options for saving payment methods, not the standard save payment method box, at least not on my test system where AP is still enabled. Are you describing the Link-based flow or the checkbox that we render in the plugin?

The checkbox we render in the plugin. With Link enabled we intentionally hide it for the card method (Link owns save consent inside the element), so only the Link UI appears on your system. To follow steps 4 and 7, disable Link (with saved cards enabled, logged in); the checkbox then renders and feeds the `savePaymentMethod` confirm argument this PR gates. The Link save flow is Stripe-side and unaffected.

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
```diff
@@ -304,6 +304,183 @@ public function test_synchronize_repairs_a_context_that_drifted_from_the_cart():
...
+		$log_out = static function () use ( $original_customer ) {
+			wp_set_current_user( 0 );
+			WC()->customer = $original_customer;
+		};
+
+		add_filter( 'wc_stripe_request_body', $capture_body, 10, 2 );
+		add_filter( 'pre_http_request', $mock_request, 10, 3 );
+
+		try {
+			$start_logged_in ? $log_in() : $log_out();
```
daledupreez: Not blocking, but this may be easier to follow if you make it an explicit if/else.
```suggestion
			if ( $start_logged_in ) {
				$log_in();
			} else {
				$log_out();
			}
```

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
```diff
@@ -304,6 +304,183 @@ public function test_synchronize_repairs_a_context_that_drifted_from_the_cart():
...
+
+			$manager = new WC_Stripe_Checkout_Session_Manager();
+			$before  = $manager->synchronize();
+
+			$start_logged_in ? $log_out() : $log_in();
```
daledupreez: As above RE using an if/else block.

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
```diff
@@ -304,6 +304,183 @@ public function test_synchronize_repairs_a_context_that_drifted_from_the_cart():
...
+
+	/**
+	 * The inverse transition: logging out must also recreate the session.
+	 */
+	public function test_synchronize_recreates_session_when_user_logs_out(): void {
+		[ $while_logged_in, $after_logout, $still_reused, $captured_creates, $create_count ] =
+			$this->run_synchronize_login_transition( true, 'logout_transition' );
+
+		$this->assertSame( 2, $create_count, 'The logout must trigger exactly one session recreation.' );
+		$this->assertTrue( $while_logged_in['save_payment_method_enabled'] );
```
daledupreez, line 438: Not blocking, but should we also confirm that 2 create calls were captured as we do in the other test case?
```suggestion
		$this->assertCount( 2, $captured_creates );
		$this->assertTrue( $while_logged_in['save_payment_method_enabled'] );
```

daledupreez (approved):
I have some minor feedback on the tests, but the main feature seems to be working well and as described.

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
wjrosa: Applied in 08f454730, to both transitions (the flip below had the same pattern).

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
wjrosa: Done in 08f454730, together with the first one.

tests/phpunit/class-wc-stripe-checkout-session-manager-test.php
wjrosa: Good call. Accepted in d2c60add7.

PR #5843, opened 2026-08-21
Use only one shared Stripe.js instance per page

diegocurbelo:
Fixes STRIPE-1368
Fixes #<github_issue_id>

On a classic (shortcode) checkout, the plugin created **two separate Stripe.js controller instances** on a single page load — one for the Payment Element and one for the Express Checkout Element — each with its own set of Stripe iframes.

The cause is that `WCStripeAPI` caches its Stripe instance on the API object (`this.stripe`), and we construct a `WCStripeAPI` in four different places:

| Bundle | Construction site |
| --- | --- |
| `upe-classic` | `client/classic/upe/deferred-intent.js` (mounts the Payment Element) |
| `upe-classic` | `client/classic/upe/init.js` |
| `express-checkout` | `client/entrypoints/express-checkout/index.js` |
| `upe-blocks` | `client/blocks/upe/index.js` |

`upe-classic`, `express-checkout` and `upe-blocks` are separate webpack entries (`splitChunks: false`), so each ships its own module instances — a module-level cache would still give each bundle its own controller. The cache has to be page scoped.

`client/blocks/load-stripe.js` had the same problem in a different shape: it called `@stripe/stripe-js`'s `loadStripe()`, which returns a **fresh instance per call**, at module scope from both `client/blocks/express-checkout/index.js` and `client/blocks/checkout-sessions/checkout-container.js`. Simply loading `upe-blocks.js` instantiated Stripe.js twice before anything rendered.

This PR adds `client/stripe-utils/shared-stripe-instance.js`, a `window`-scoped registry keyed by publishable key + the resolved constructor options, and routes both creation paths through it:

- `getSharedStripeInstance()` — synchronous, for `WCStripeAPI.createStripe()`.
- `loadSharedStripeInstance()` — promise-returning, for the Blocks loader. It takes the synchronous path when `window.Stripe` is already defined, and only falls back to `@stripe/stripe-js` when a "defer render-blocking JS" optimizer has pushed Stripe.js past our bundle (the script tag is present, so the provenance assertion passes, but `Stripe` isn't defined yet).
- Stripe.js provenance is validated before the first instance is cached for each key/options pair. Reusing that validated instance does not inspect the script tag again, so later DOM changes cannot disrupt an initialized checkout.

**_NOTE:_**  Instances are only shared when the constructor arguments match, so a different publishable key, locale or developer-tools setting still gets its own controller. This currently affects the blocks checkout as it always uses `auto` as the locale (PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5844 fixes it).

**_NOTE:_** This also removes duplicate **controllers** and their iframes only. It does **not** reduce `/v1/elements/sessions` requests — those come from Elements *groups*, and `client/entrypoints/express-checkout/index.js` deliberately creates one `stripe.elements()` group per enabled express payment type. Unchanged here.

### Backward compatibility

- JS only; no PHP hooks, filters, option keys or REST routes are touched.
- `WCStripeAPI` isn't exported to `window`, registered with `wp.hooks`, or exposed as a webpack `library`, so it isn't reachable from outside the plugin; `createStripe()` keeps its signature and return type and now returns the page's shared instance instead of a newly constructed one.

## Testing instructions

**Classic (shortcode) checkout — the reported case**

1. Enable Stripe with card payments, and enable Express Checkout (Apple Pay / Google Pay / Link) under `WooCommerce > Settings > Payments > Stripe`.
2. Set the Checkout page to the classic `[woocommerce_checkout]` shortcode.
3. Add a product to the cart and open `/checkout/` in Chrome.
4. Open DevTools and run:
   ```js
   [ ...document.querySelectorAll( 'iframe[name^="__privateStripeController"]' ) ].length
   ```
   - On `develop`: **2**
   - With this PR: **1**
5. Confirm that the card field mounts and the express checkout buttons render.
6. Enter an invalid card (`4000 0000 0000 0002`) and confirm the decline message still renders when placing the order.
7. Enter a valid test card (`4242 4242 4242 4242`) and place the order.
8. Confirm it completes successfully

**Blocks checkout**

1. Set the Checkout page to the `woocommerce/checkout` block.
2. Repeat step 4 above — the count should drop from **3** to **2** (see the _**NOTE 1**_ above for why it isn't 1 yet).
3. Confirm the Payment Element mounts, the express buttons render, and a card payment completes.

**Optimized Checkout / Checkout Sessions**

1. Enable Optimized Checkout and repeat the Blocks checkout run.
2. Confirm the OC element mounts and a payment completes.

**Other pages that mount Stripe elements** (regression checks)

1. Product page and Cart page with express checkout buttons enabled.
2. `My Account > Payment methods > Add payment method`.
3. Pay for order (`/checkout/order-pay/<id>/`).
4. With Subscriptions active, the change-payment-method page.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/stripe-utils/shared-stripe-instance.js
```diff
@@ -0,0 +1,146 @@
...
+ * @return {Object} The Stripe instance.
+ */
+export const getSharedStripeInstance = ( key, options ) => {
+	const entry = getEntry( key, options );
+
+	if ( ! entry.instance ) {
+		// Asserted before anything is stored, so a provenance failure keeps
+		// throwing instead of being remembered as a resolved entry. Not
+		// re-asserted for a cached instance: that one was already checked, and
+		// a per-call assertion would break checkout if the tag is removed later.
```
jaclync, line 85: non-blocking nit: could this change be worth mentioning in the BC section of the PR description?

client/stripe-utils/__tests__/shared-stripe-instance.test.js
```diff
@@ -0,0 +1,312 @@
...
+			);
+
+			expect( () =>
+				getSharedStripeInstance( 'pk_test_123', { locale: 'en' } )
+			).toThrow( /provenance check failed/ );
+			expect( () =>
+				getSharedStripeInstance( 'pk_test_123', { locale: 'en' } )
+			).toThrow( /provenance check failed/ );
+			expect( global.Stripe ).not.toHaveBeenCalled();
+		} );
```
jaclync, line 104: non-blocking: is it worth covering a case where a valid Stripe tag was added first -> `getSharedStripeInstance` creates a valid instance -> the Stripe tag gets removed or replaced with an invalid tag -> `getSharedStripeInstance`?

changelog.txt
```diff
@@ -23,6 +23,7 @@ xxxx-xx-xx - version 11.0.0
 * Dev - Make GITHUB_TOKEN optional for the local E2E Docker setup by falling back to the GitHub CLI and then to plugin zips placed in tests/e2e/deps
 * Fix - Reload the plugin's payment method settings when refreshing account details, so the settings screen no longer keeps showing stale values
 * Update - Fetch express checkout AJAX nonces on demand instead of embedding them in every page
+* Fix - Load a single instance of Stripe.js per page
```
jaclync, line 25: non-blocking nit: do you think it's worth also mentioning the user-facing impact, I think this change improves performance by reducing the number of Stripe network requests for example?

jaclync (approved):
LGTM and tested well! :shipit:

client/stripe-utils/__tests__/shared-stripe-instance.test.js
daledupreez: I've added a test case for this in 5787c3aee2c1281bffda2df10acc978aab352f7c.

changelog.txt
daledupreez: I think we can take another look at this as we get closer to the release -- I am going to queue this up for merge in the meantime.

PR #5844, opened 2026-08-21
Replace the empty stripe_locale in the blocks Stripe loader with locale

diegocurbelo:
Fixes #<github_issue_id>

The Stripe loader used by the Cart and Checkout blocks read its locale from a config key that the blocks payload has never contained. The lookup silently fell through to `auto`, so Stripe picked the shopper’s browser locale instead of the store’s.

This reads the key that is actually there. The `auto` fallback is unchanged — the store locale still resolves to auto on its own when Stripe doesn’t support it.

The only visible change is the Adaptive Pricing payment element, which now renders in the store locale (if supported by Stripe, if it's not then the browser locale is used). The express checkout buttons are unaffected: they already pass their own locale when creating their Elements group.

## Testing instructions

1. Enable Adaptive Pricing and set the Checkout page to the Checkout block.
2. Set the site language to one Stripe supports (e.g. French) and keep your browser in English.
3. Open the checkout.
4. On develop: the payment element renders in English (browser locale).
5. With this PR: it renders in French (store locale).
6. Confirm the express checkout buttons are unchanged, and that a card payment still completes.
7. Set the site language to one Stripe does not support (e.g. Hindi) and reload — the payment element should fall back to the browser locale.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -20,6 +20,7 @@ xxxx-xx-xx - version 11.0.0
 * Dev - Make GITHUB_TOKEN optional for the local E2E Docker setup by falling back to the GitHub CLI and then to plugin zips placed in tests/e2e/deps
 * Fix - Reload the plugin's payment method settings when refreshing account details, so the settings screen no longer keeps showing stale values
 * Update - Fetch express checkout AJAX nonces on demand instead of embedding them in every page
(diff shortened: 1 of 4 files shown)


wjrosa (approved):
Looks good to me! Thanks for the fix 👍

PR #5846, opened 2026-08-24
Keep express checkout addresses when normalization fails

### LINKED ISSUE #5838: Express Checkout (Apple Pay / Google Pay): billing_address is dropped from the Store API checkout request
# Express Checkout (Apple Pay / Google Pay): `billing_address` is dropped from the Store API checkout request

**Plugin:** WooCommerce Stripe Gateway
**Affected versions:** reproduced on **10.8.5** and still present on **10.9.0**

## Environment

| | |
|---|---|
| WordPress | 7.1 |
| WooCommerce | 11.0.1 |
| WooCommerce Stripe Gateway | 10.9.0 (also 10.8.5) |
| PHP | 8.2.29 |
| Checkout | Blocks (Store API) |
| Optimized Checkout Element | enabled (`optimized_checkout_element = yes`) |
| UPE methods enabled | `card`, `link` (+ Apple Pay / Google Pay) |
| Shipping | disabled |
| Taxes | disabled |
| Guest checkout | disabled (account is created during checkout) |
| Products | virtual + downloadable |

## Symptom

Paying with Apple Pay (triggered from macOS Safari, confirmed on iPhone via Handoff) fails with:

```
Parámetro(s) que falta(n): billing_address
```

i.e. `rest_missing_callback_param` from `POST /wc/store/v1/checkout`. **No order is created**, so the
customer is not charged — but the purchase cannot be completed.

The wallet sheet completes normally; the error appears immediately afterwards.

## What we measured

We instrumented the site with a small read-only plugin that records, for each step, **only the field
names** present in the request (no values), plus the outcome. Two different runs:

**Run A — the request arrives complete**

| time | step | `billing_address` |
|---|---|---|
| 08:15:17 | enters `wc_ajax_wc_stripe_normalize_address` | 11 fields, all populated |
| 08:15:17 | leaves it (filter `wc_stripe_express_checkout_normalize_address`) | 11 fields, all populated |
| 08:15:17 | arrives at `POST /wc/store/v1/checkout` | 11 fields, all populated |

**Run B — the request arrives without the address**

| time | step | `billing_address` |
|---|---|---|
| 08:31:36 | `wc_ajax_wc_stripe_normalize_address` | **never fired** |
| 08:31:36 | arrives at `POST /wc/store/v1/checkout` | **ABSENT** |

In run B the request body contained exactly:

```
payment_method, payment_data, extensions, additional_fields, customer_note
```

with the header `X-WCSTRIPE-EXPRESS-CHECKOUT` present, so the request definitely came from the
Express Checkout path. Note that **both `billing_address` and `shipping_address` are missing
together**, and no other key is missing.

## Where we believe it happens

In `build/express-checkout.js` and `build/upe-blocks.js` (identical in 10.8.5 and 10.9.0):

```js
p = await t.expressCheckoutNormalizeAddress( l.billing_address, l.shipping_address );
p && ( l.billing_address = p.billing_address, l.shipping_address = p.shipping_address );
```

The response shape is not validated. If `p` is truthy but does not contain those two keys, **both
addresses become `undefined` in a single statement**, and `JSON.stringify` then removes them from the
request body. That matches the observed signature exactly: those two keys missing, nothing else.

We could not determine **why** the normalization step is skipped in run B — that is the part we
cannot explain and where we would appreciate your help.

## Additional observation

The Express Checkout Element options are built inline with `emailRequired: true` and
`shippingAddressRequired: <needsShipping>`, but **`billingAddressRequired` is never set**, so it
falls back to Stripe's default. On a store with shipping disabled, the wallet therefore returns very
little address data. We do not know whether this is related, but it seemed worth mentioning.

## Suggested hardening (independent of the root cause)

Guarding the assignment would turn a hard failure into a harmless no-op:

```js
if ( p && p.billing_address ) { l.billing_address = p.billing_address; }
if ( p && p.shipping_address ) { l.shipping_address = p.shipping_address; }
```

## Our current workaround

Because `billing_address` is the **only** `required` argument on that route
(`CheckoutSchema.php:135`), the request is rejected before any of our code can react. We currently:

1. remove `required` from `billing_address` on `/wc/store/v1/checkout` **only** (exact route match —
   using a prefix match also affects `/wc/store/v1/checkout/{id}`, where
   `CheckoutOrder::update_billing_address()` then passes `null` to
   `WC_Order::set_billing_address( array $address )` and PHP raises an uncaught `TypeError`);
2. rebuild the address from the order inside
   `woocommerce_store_api_checkout_update_order_from_request`, and put it back on the request — this
   has to run **before** `process_customer()`, because account creation reads the e-mail from
   `$request['billing_address']['email']` (`Checkout.php:987`);
3. reject the request if no e-mail can be recovered.

This works, but it is a workaround for a defect in the express path and we would rather remove it.

## Impact

Every Apple Pay / Google Pay attempt that hits this condition fails. No order, no charge, but the
sale is lost and the customer sees a technical message they cannot act on.

diegocurbelo:
Fixes STRIPE-1401
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5838

Apple Pay / Google Pay orders intermittently failed with `rest_missing_callback_param: billing_address` from `POST /wc/store/v1/checkout`. No order was created and no charge was made, but the sale was lost and the shopper saw a message they could not act on.

`processOrder()` submitted the addresses it built from the wallet event to `wc_ajax_wc_stripe_normalize_address`, then overwrote both with the response:

```js
if ( normalizedAddress ) {
	normalizedOrderData.billing_address = normalizedAddress.billing_address;
	normalizedOrderData.shipping_address = normalizedAddress.shipping_address;
}
```

The fix merges the response over the wallet-derived addresses instead of replacing them:

```js
normalizedOrderData.billing_address = {
	...normalizedOrderData.billing_address,
	...getUsableAddress( normalizedAddress?.billing_address ),
};
```

## Testing instructions

`npm run test:js -- client/express-checkout/__tests__/payment-flow.test.js`

Manual:

1. On a store with the Blocks checkout and Apple Pay or Google Pay enabled, add a virtual product to the cart and go to the checkout page.
2. Open the browser developer tools Network tab and filter for `wc-ajax=wc_stripe_normalize_address`.
3. Complete a payment through the wallet sheet.
4. Verify the order is created, and the `POST /wc/store/v1/checkout` request body contains a populated `billing_address`.
5. To simulate the failure, block or stub the normalize-address request so it returns `200` with a non-JSON body (e.g.: using devtools request blocking/request conditions), then pay again.
    - In develop the checkout fails with "Missing parameter(s): billing_address" and no order is created
    - In this branch, the order is created using the address the wallet supplied.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/express-checkout/__tests__/payment-flow.test.js
+++ b/client/express-checkout/__tests__/payment-flow.test.js
@@ -656,8 +656,22 @@ describe( 'address normalization', () => {
 				{ billing_address: '', shipping_address: null },
 			],
 			[
+				'addresses encoded as non-empty strings',
+				{
+					billing_address: 'invalid',
+					shipping_address: 'invalid',
+				},
+			],
+			[
 				'addresses encoded as empty PHP arrays',
 				{ billing_address: [], shipping_address: [] },
+			],
+			[
+				'addresses encoded as non-empty arrays',
+				{
+					billing_address: [ 'invalid' ],
+					shipping_address: [ 'invalid' ],
+				},
 			],
 			[
 				'empty addresses',
--- a/client/express-checkout/__tests__/payment-flow.test.js
+++ b/client/express-checkout/__tests__/payment-flow.test.js
@@ -524,3 +524,221 @@ describe( 'handleConfirmationTokenFlow', () => {
 		expect( abortPayment ).not.toHaveBeenCalled();
 	} );
 } );
+
+describe( 'address normalization', () => {
+	let api;
+	let stripe;
+	let elements;
+	let completePayment;
+	let abortPayment;
+	let event;
+
+	// What normalizeOrderData() builds from the wallet event below.
+	const walletBillingAddress = {
+		first_name: 'Jane',
+		last_name: 'Doe',
+		company: '',
+		email: 'jane@example.com',
+		phone: '',
+		country: 'US',
+		address_1: '456 Oak Ave',
+		address_2: '',
+		city: 'San Juan',
+		state: 'PR',
+		postcode: '00901',
+	};
+	const walletShippingAddress = {
+		first_name: 'Jane',
+		last_name: 'Doe',
+		company: '',
+		phone: '',
+		country: 'US',
+		address_1: '456 Oak Ave',
+		address_2: '',
+		city: 'San Juan',
+		state: 'PR',
+		postcode: '00901',
+		method: [ 'flat_rate:1' ],
+	};
+
+	const getCreateOrderPayload = () =>
+		api.expressCheckoutECECreateOrder.mock.calls[ 0 ][ 0 ];
+
+	beforeEach( () => {
+		const paymentResult = {
+			payment_result: {
+				payment_status: 'success',
+				redirect_url: 'https://example.com/order-received',
+			},
+		};
+		api = {
+			expressCheckoutECECreateOrder: jest
+				.fn()
+				.mockResolvedValue( paymentResult ),
+			expressCheckoutECEPayForOrder: jest
+				.fn()
+				.mockResolvedValue( paymentResult ),
+			expressCheckoutNormalizeAddress: jest.fn(),
+			confirmIntent: jest.fn().mockReturnValue( true ),
+		};
+		stripe = {
+			createPaymentMethod: jest
+				.fn()
+				.mockResolvedValue( { paymentMethod: { id: 'pm_test_123' } } ),
+			createConfirmationToken: jest.fn().mockResolvedValue( {
+				confirmationToken: { id: 'ct_test_456' },
+			} ),
+		};
+		elements = {};
+		completePayment = jest.fn();
+		abortPayment = jest.fn();
+		event = {
+			billingDetails: {
+				name: 'Jane Doe',
+				email: 'jane@example.com',
+				address: {
+					city: 'San Juan',
+					country: 'US',
+					line1: '456 Oak Ave',
+					postal_code: '00901',
+					state: 'PR',
+				},
+			},
+			shippingAddress: {
+				name: 'Jane Doe',
+				address: {
+					city: 'San Juan',
+					country: 'US',
+					line1: '456 Oak Ave',
+					postal_code: '00901',
+					state: 'PR',
+				},
+			},
+			shippingRate: { id: 'flat_rate:1' },
+		};
+	} );
+
+	// Apple Pay and Google Pay run through the manual flow; Amazon Pay runs through the
+	// confirmation token flow unless the cart has a free trial. Both share processOrder.
+	describe.each( [
+		[
+			'handleManualPaymentMethodFlow',
+			handleManualPaymentMethodFlow,
+			'apple_pay',
+		],
+		[
+			'handleConfirmationTokenFlow',
+			handleConfirmationTokenFlow,
+			'amazon_pay',
+		],
+	] )( '%s', ( _name, runFlow, expressPaymentType ) => {
+		const flow = ( params = {} ) =>
+			runFlow( {
+				api,
+				stripe,
+				elements,
+				completePayment,
+				abortPayment,
+				event: { ...event, expressPaymentType },
+				...params,
+			} );
+
+		test.each( [
+			[ 'null', null ],
+			[
+				'an HTML page served by a cache or a redirect',
+				'<!DOCTYPE html>',
+			],
+			[ 'an empty array', [] ],
+			[ 'a body without the address keys', { success: true } ],
+			[
+				'addresses that are not objects',
+				{ billing_address: '', shipping_address: null },
+			],
+			[
+				'addresses encoded as empty PHP arrays',
+				{ billing_address: [], shipping_address: [] },
+			],
+			[
+				'empty addresses',
+				{ billing_address: {}, shipping_address: {} },
+			],
+		] )(
+			'keeps the wallet addresses when normalization responds with %s',
+			async ( _label, response ) => {
+				api.expressCheckoutNormalizeAddress.mockResolvedValue(
+					response
+				);
+
+				await flow();
+
+				const payload = getCreateOrderPayload();
+				expect( payload.billing_address ).toEqual(
+					walletBillingAddress
+				);
+				expect( payload.shipping_address ).toEqual(
+					walletShippingAddress
+				);
+				expect( completePayment ).toHaveBeenCalled();
+				expect( abortPayment ).not.toHaveBeenCalled();
+			}
+		);
+
+		test( 'applies the normalized addresses when the response provides them', async () => {
+			// Puerto Rico: the wallet reports it as a US state, and normalization
+			// promotes it to a country.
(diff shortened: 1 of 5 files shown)


client/express-checkout/payment-flow.js
```diff
@@ -43,6 +43,13 @@ const handlePaymentFlowException = ( event, exception, abortPayment ) => {
 	);
 };
 
+const getUsableAddress = ( address ) =>
+	address !== null &&
+	typeof address === 'object' &&
+	! Array.isArray( address )
+		? address
```
daledupreez, line 50: Is it worth being extra careful and ensuring that we get an object returned?
```suggestion
		? Object( address )
```

client/express-checkout/__tests__/payment-flow.test.js
```diff
@@ -524,3 +524,235 @@ describe( 'handleConfirmationTokenFlow', () => {
...
+	} );
+
+	// Apple Pay and Google Pay run through the manual flow; Amazon Pay runs through the
+	// confirmation token flow unless the cart has a free trial. Both share processOrder.
+	describe.each( [
+		[
+			'handleManualPaymentMethodFlow',
+			handleManualPaymentMethodFlow,
+			'apple_pay',
+		],
```
daledupreez, line 628: Just to be explicit and to get full coverage, it is worth adding a `google_pay` entry?
```suggestion
		[
			'handleManualPaymentMethodFlow - Apple Pay',
			handleManualPaymentMethodFlow,
			'apple_pay',
		],
		[
			'handleManualPaymentMethodFlow - Google Pay',
			handleManualPaymentMethodFlow,
			'google_pay',
		],
```

client/express-checkout/__tests__/payment-flow.test.js
```diff
@@ -524,3 +524,235 @@ describe( 'handleConfirmationTokenFlow', () => {
...
+			} );
+
+		test.each( [
+			[ 'null', null ],
+			[
+				'an HTML page served by a cache or a redirect',
+				'<!DOCTYPE html>',
+			],
+			[ 'an empty array', [] ],
+			[ 'a body without the address keys', { success: true } ],
```
daledupreez: Nit: It may be worth mentioning "undefined" in this description.
```suggestion
			[ 'a body without the address keys - addresses are undefined', { success: true } ],
```

client/express-checkout/__tests__/payment-flow.test.js
```diff
@@ -524,3 +524,235 @@ describe( 'handleConfirmationTokenFlow', () => {
...
+			[
+				'addresses that are not objects',
```
daledupreez: Nit: We should be more specific.
```suggestion
				'addresses that are empty strings',
```

client/express-checkout/__tests__/payment-flow.test.js
```diff
@@ -524,3 +524,235 @@ describe( 'handleConfirmationTokenFlow', () => {
...
+				'addresses encoded as non-empty arrays',
+				{
+					billing_address: [ 'invalid' ],
+					shipping_address: [ 'invalid' ],
+				},
+			],
+			[
+				'empty addresses',
+				{ billing_address: {}, shipping_address: {} },
+			],
```
daledupreez, line 691: It might also be worth adding some test cases for numbers and other object-like values like symbols and other object types.
```suggestion
			[
				'empty addresses',
				{ billing_address: {}, shipping_address: {} },
			],
			[
				'addresses encoded as numbers',
				{ billing_address: 1, shipping_address: 2 },
			],
			[
				'addresses encoded as symbols',
				{ billing_address: Symbol( 'billing' ), shipping_address: Symbol( 'shipping' ) },
			],
			[
				'addresses encoded as other object types',
				{ billing_address: new Date(), shipping_address: /test_regex/ },
			],
```

client/express-checkout/__tests__/payment-flow.test.js
```diff
@@ -524,3 +524,235 @@ describe( 'handleConfirmationTokenFlow', () => {
...
+				state: '',
+			} );
+			expect( payload.shipping_address ).toEqual( {
+				...walletShippingAddress,
+				country: 'PR',
+				state: '',
+			} );
+		} );
+
+		test( 'keeps the wallet fields the response omits', async () => {
```
daledupreez, line 736: Noting for the record that I just checked the back end implementation, and our default behaviour is to return both values that are modified/tweaked. However, it is possible that other functions that hook into the `wc_stripe_express_checkout_normalize_address` filter don't do that, which could result in a change in behaviour.
I think we should do two things to cover this:
 1. We should add a compatibility note in `readme.txt`
 2. We should update the `wc_stripe_express_checkout_normalize_address` filter documentation to note that any removed values should be returned as `''`.

daledupreez (approved):
Functionally this is working as described, but I've dropped some suggestions around tightening up the code and the tests. I also think we should document the compatibility impact, as it is unlikely but possible that custom code could see different behaviour after this PR.

daledupreez:
Also for the record, I tested the bad behaviour by adding the following snippet:
```php
function test_break_ece_normalize_address( $data ) {
	return [
		'bad',
		'data'
	];
}
add_filter( 'wc_stripe_express_checkout_normalize_address', 'test_break_ece_normalize_address', 20, 1 );
```

client/express-checkout/__tests__/payment-flow.test.js
diegocurbelo: Fixed in [0cfd553](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5846/commits/0cfd553199302273492d1405b0bc1cfc0efd06dc)

client/express-checkout/__tests__/payment-flow.test.js
diegocurbelo: Fixed in [0cfd553](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5846/commits/0cfd553199302273492d1405b0bc1cfc0efd06dc)

client/express-checkout/__tests__/payment-flow.test.js
diegocurbelo: Fixed in [0cfd553](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5846/commits/0cfd553199302273492d1405b0bc1cfc0efd06dc)

client/express-checkout/__tests__/payment-flow.test.js
diegocurbelo: Fixed in [0cfd553](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5846/commits/0cfd553199302273492d1405b0bc1cfc0efd06dc)

client/express-checkout/__tests__/payment-flow.test.js
diegocurbelo: Updated the readme and filter doc in [db70e43](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5846/commits/db70e4318c57ae553e6c97c478b86c5566389864)

PR #5848, opened 2026-08-24
Prevent invalid express checkout display items

### LINKED ISSUE #5827: Express Checkout: display items exceed total on zero-decimal currencies, wallet sheet rejects resolve() (legacy AJAX path)
### Summary

On stores whose active currency has **0 decimals** (e.g. CZK via WooCommerce Multilingual), the express checkout (Apple Pay / Google Pay / Link) breaks as soon as the wallet sheet requests shipping options. Stripe rejects the `resolve()` with:

```
The amount 76800 is less than the total amount of the line items provided.
```

The sheet then dies and the customer cannot pay. In our case this affected **every** wallet payment that required shipping on that domain, and the failure is silent server-side: no PHP error, no failed order, no `PaymentIntent`. We only found it because we run our own client-side error telemetry.

### Root cause

`WC_Stripe_Helper::build_line_items()` and `WC_Stripe_Helper::get_stripe_amount()` round at different stages:

* Tax and shipping are passed through `wc_format_decimal( $value, wc_get_price_decimals() )` **before** the Stripe scaling. With `wc_get_price_decimals() === 0`, `122.62` becomes `123`, and `get_stripe_amount()` then yields `12300`.
* Product lines / subtotal go straight into `get_stripe_amount()`, where the `* 100` happens **before** `wc_format_decimal()`. The 0-decimal setting therefore never applies and `461.344538` yields `46134` instead of `46100`.

The resulting `displayItems` sum exceeds the (correct) total, and Stripe's validation refuses it.

Real response from `?wc-ajax=wc_stripe_get_shipping_options` (10.8.5, CZK, 19 % tax, prices incl. tax):

| item | amount |
|---|---|
| product line | 46134 |
| tax | 12300 |
| shipping | 18400 |
| **sum** | **76834** |
| `total.amount` | **76800** |

The non-itemized branch shows the same drift (`Subtotal 46134`).

### Why this only affects the legacy AJAX path

The Store API path already guards against exactly this. `transformCartDataForDisplayItems()` in `client/express-checkout/transformers/wc-to-stripe.js` ends with:

```js
// if `totalAmount` is less than the total of `displayItems`, Stripe throws an error
// it can sometimes happen that the total is _slightly_ less, due to rounding errors …
if ( totalAmount < totalAmountOfDisplayItems ) {
    return [];
}
```

That guard is missing on the responses used by the wallet events. `shippingAddressChangeHandler()` and `shippingRateChangeHandler()` in `client/express-checkout/event-handler.js` pass `normalizeLineItems( response.displayItems )` into `event.resolve()` unchecked, and those `displayItems` come from `build_display_items()` in PHP. So the button initializes fine and only dies once the customer opens the sheet.

### Steps to reproduce

1. Store with a 0-decimal currency active (WCML currency option `num_decimals = 0`; ours also uses `rounding = up`, `rounding_increment = 10`).
2. Prices include tax, `woocommerce_tax_round_at_subtotal = yes`, a tax rate that produces a fractional amount.
3. Add a shippable product, open checkout, tap Apple Pay / Google Pay / Link.
4. The sheet requests shipping options and Stripe rejects the resolve with the error above.

Reproducible headlessly without a wallet device: POST to `?wc-ajax=wc_stripe_get_shipping_options` with a valid nonce and a shipping address, then compare `sum(displayItems[].amount)` against `total.amount`.

### Suggested fix

Apply the same semantics the Store API transformer already uses, but server-side in `build_display_items()` (or in the two event handlers before `resolve()`): if the display items sum exceeds the total, drop the display items and let the sheet show the total only. That keeps the charged amount untouched and matches existing behaviour on the other code path.

A more complete fix would be to make `get_stripe_amount()` apply the currency's decimal setting before scaling, so product lines round consistently with tax and shipping. That is the actual asymmetry, but it is also the riskier change.

### Environment

* WooCommerce Stripe Gateway 10.8.5
* WooCommerce 10.9.4, WordPress 7.0.4, PHP 8.2
* WPML / WooCommerce Multilingual 5.5.7 (multi-currency, 0-decimal currency)
* Classic (shortcode) checkout, so the legacy `wc-ajax` endpoints are in use

We are running the server-side guard described above as a must-use plugin and wallet payments work again, with the charged amount unchanged.

diegocurbelo:
Fixes STRIPE-1394
Fixes #5827

Express Checkout initializes cart and checkout buttons from a PHP-rendered cart snapshot before refreshing through the Store API. With zero-decimal currency configurations, independently rounded display items can add up to more than the authoritative cart total, causing Stripe to reject the first wallet click.

This change omits the PHP-generated breakdown when its normalized sum exceeds the total, matching the existing Store API fallback while leaving the amount charged unchanged.

- This change adds `WC_Stripe_Express_Checkout_Helper::get_display_items_for_total()` so cart and Pay for Order payloads use the same display-item validation.

## Testing instructions

_NOTE:_ See STRIPE-1394, for details on the repro status and JN site credentials 

1. Configure a zero-decimal active currency using a [WPML Multilingual & Multicurrency for WooCommerce](https://wordpress.org/plugins/woocommerce-multilingual/)
    - currency = CZK
    - num_decimals = 0
    - rounding = up
    - rounding_increment = 10
2. Enable prices inclusive of tax and tax rounding at subtotal.
3. Add a shippable product whose tax produces a fractional amount.
4. Fresh-load the classic cart or checkout page and immediately launch Apple Pay, Google Pay, or Link.
5. Confirm the wallet opens and displays the total, and no error is shown

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1856,12 +1856,27 @@ public function build_display_items( $itemized_display_items = false ) {
...
+		$display_items    = WC_Stripe_Helper::build_line_items( $display_items );
+
+		$display_items_total = array_reduce(
+			$display_items,
+			static function ( $total, $item ) {
+				$amount = 'total_discount' === ( $item['key'] ?? '' ) ? -$item['amount'] : $item['amount'];
+				return $total + $amount;
+			},
+			0
+		);
```
jaclync, line 1869: non-blocking: it feels like this could be a helper similar to `build_line_items` to calculate the total amount, potentially sharing with pay-for-order as in the PR review comment.

jaclync (commented):
While checking the use cases of `transformLabeledDisplayItems`, pay-for-order seems to be another use case similar to the classic cart/checkout page but it doesn't have the guard that this PR adds (I was able to repro the ECE click stuck indefinitely from pay-for-order page). How about moving the guard from the PR to a shared helper so that both code paths can be fixed?

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
diegocurbelo: Implemented a shared `get_display_items_for_total()` helper in [0e897b8](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5848/commits/0e897b8a8aa1d05bfe9b4fb50cfdaf7a95174341) and applied it to both cart and Pay for Order payloads.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1856,17 +1856,46 @@ public function build_display_items( $itemized_display_items = false ) {
...
+	 *
+	 * Cart discounts use a positive amount with a key, while Pay for Order discounts
+	 * are already negative. Only keyed discounts need their sign adjusted here.
+	 *
+	 * @param array     $display_items Display items in the PHP-generated format.
+	 * @param float|int $total_amount  Total amount in the currency's smallest unit.
+	 *
+	 * @return array The original items, or an empty array when Stripe would reject them.
+	 */
+	public function get_display_items_for_total( array $display_items, $total_amount ): array {
```
jaclync, line 1886: non-blocking: maybe this function can have direct unit testing coverage in `tests/phpunit/payment-methods/class-wc-stripe-express-checkout-helper-test.php`?

jaclync (approved):
LGTM! Worked well in classic cart/checkout and pay-for-order pages now :shipit:

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1856,17 +1856,46 @@ public function build_display_items( $itemized_display_items = false ) {
...
+		$display_items_total = array_reduce(
+			$display_items,
+			static function ( $total, $item ) {
+				$amount = 'total_discount' === ( $item['key'] ?? '' ) ? -$item['amount'] : $item['amount'];
```
jaclync, line 1890: This looks like the same transformation as in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5927/changes#diff-6f71ef0dfdcfd98bd6372ed52980bca4f0a9141c5d98e0f89fa5c665792347b1R154. Could we reuse the logic from there, applying `transformLabeledDisplayItems` before the `array_reduce`?

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -365,6 +365,7 @@ public function localize_pay_for_order_page_scripts( $order ) {
 			];
 		}
```
jaclync, line 366: Do you think it makes sense to apply similar fees handling as in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5942?

diegocurbelo: This PR’s change prevents Stripe rejection by omitting the invalid breakdown, but it does not correct the displayed fee. Adding that normalization here would expand the surface and overlap with #5942. I think we should add Pay for Order coverage there or handling it in a focused follow-up after merging that.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
diegocurbelo: Added in [9fc4ea7](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5848/commits/9fc4ea70ea8a23ebbcd8d5dc1ee3391c1332aa7f)

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
diegocurbelo: Moved the validation to `transformLabeledDisplayItems()`, so `normalizeLineItems()` runs before the item sum is checked. Cart bootstrap and Pay for Order now pass their authoritative totals to that transformer.

jaclync (approved):

includes/payment-methods/class-wc-stripe-express-checkout-element.php
jaclync: Yup makes sense to keep it separate. Tracked in STRIPE-1456.

PR #5851, opened 2026-08-24
Change admin payment method availability to support multi-currency

diegocurbelo:
Fixes STRIPE-1077

Stripe payment methods, including Amazon Pay, are no longer disabled in the admin based only on the store’s base currency. The existing unsupported-currency pill remains visible as guidance, while merchants can enable the method for checkout currencies supplied by a currency-switching plugin.

Other availability failures, including conflicts with official payment-method plugins and Amazon Pay’s billing-address tax restriction, continue to disable methods that are not already enabled. Existing checkout availability checks remain responsible for hiding a method when the shopper’s current currency is unsupported.

#### Alternative solutions

An alternative solution would be to only allow enabling payment methods that need a specific currency based on:
- Known-plugin allowlist: detect WPML/WCML, Aelia, FOX/WOOCS, CURCY, etc. 
    This is reliable for activation but requires maintenance and, ideally, plugin-specific checks to confirm multi-currency is actually configured, so we only show the payment methods that support the actual enabled currencies
- Custom filter; `wc_stripe_is_multi_currency_active` or similar, that allows enabling payment methods even when the currency do not match.

## Testing instructions

1. Set the WooCommerce base currency to USD (without a currency-switching plugin active)
2. Open `WooCommerce > Settings > Payments > Stripe > Payment methods`
3. Confirm that a EUR-only payment method such as SEPA shows the unsupported-currency pill but its checkbox remains clickable
4. Enable SEPA and save the settings
5. Activate and configure a currency-switching plugin that offers EUR(e.g.g: WPML Multilingual & Multicurrency for WooCommerce)
6. Go to the product page, select EUR and add a product to the cart
7. Go to the checkout, and set the country to `France`
8. Confirm that SEPA is available
9. Complete the order

| USD | EUR |
|-|-|
|Screenshot 2026-08-24 at 18 39 18|Screenshot 2026-08-24 at 18 39 47|

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


client/settings/payment-request-section/index.js
```diff
@@ -103,6 +103,8 @@ const PaymentRequestSection = () => {
 
 	const isAmazonPayDisabled =
 		amazonPayUnavailableReason !== null &&
+		amazonPayUnavailableReason !==
+			PAYMENT_METHOD_UNAVAILABLE_REASONS.UNSUPPORTED_CURRENCY &&
```
jaclync, line 107: non-blocking: `*UnavailableReason !== null && *UnavailableReason !== PAYMENT_METHOD_UNAVAILABLE_REASONS.UNSUPPORTED_CURRENCY` is in both use cases here and `client/settings/general-settings-section/payment-method.js`, maybe this can be extracted to a shared helper function that determines if an unavailable reason is blocking?

jaclync (commented):
Worked as expected! During testing, I noticed some UI gap in wp-admin that might be worth addressing (in the same PR or separately) - with multi-currency enabled like with the `WPML Multilingual & Multicurrency for WooCommerce` plugin, there are still places that inform the merchant that the store currency isn't supported for the payment methods that could be supported by multi-currency:

- The admin notice: `%1$s is enabled - it requires store currency to be set to %2$s`

Image

- The payment method pill popover: `%1$s requires store currency to be set to %2$s. Set currency` with a link to the general settings, which reads like the merchant needs to change their base currency to fix a supported configuration.

Image

Now that payment methods can be enabled regardless of the currency, this copy could be misleading. Maybe they can be reworded to be like `%1$s will only be shown at checkout when the customer pays in %2$s` and dropping the `Set currency` link?

daledupreez:
It feels like it might be worth adding a filter that allows the supported currencies to be supplied by other plugins. That way our checks for (un)supported currencies can use that list of available currencies. It feels like that would be relatively easy for other plugins and/or merchants to hook into, while still letting us show valid warnings when certain payment methods won't be available at checkout.

daledupreez:
As we discussed yesterday, what I was thinking about was something like the following:
1. Add a filter that allows supported/available currencies to be returned, e.g.
```php
$store_currency = get_woocommerce_currency();
$available_store_currencies = apply_filters( 'wc_stripe_supported_store_currencies', [ $store_currency ] );
if ( ! is_array( $available_store_currencies ) ) {
	$available_store_currencies = [ $store_currency ];
}
// Ensure that we have an array of upper-case strings before returning.
```
2. Ensure that the various notices and checks that work against the store currency use the result of the code above.
3. Work out if we can contribute upstream PRs to relevant plugins to hook into the filter.

client/settings/payment-request-section/index.js
diegocurbelo: Changed in [586a393](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5851/commits/586a393931a1a7adef6abc766281c5c3952db123)

diegocurbelo:
Updated the message to: `%1$s will only be shown at checkout when the customer pays in %2$s` as suggested and added the `filter` to allow supported/available currencies to be defined.

> Work out if we can contribute upstream PRs to relevant plugins to hook into the filter.

Some popular multi-currency plugins do not have a public repos, maybe we can add an integration/compatibility page in the plugin docs? or maybe a developer blog post would more suitable cc @aheckler

client/settings/general-settings-section/__tests__/general-settings-section.test.js
```diff
@@ -358,31 +360,33 @@ describe( 'GeneralSettingsSection', () => {
 		).not.toBeInTheDocument();
 	} );
 
-	it( 'should disable the payment method checkbox and show the requires currency notice when currency is not supported', () => {
```
daledupreez: Is it worth keeping the notice test for the case where the currency is not known to be supported?

includes/class-wc-stripe-helper.php
```diff
@@ -391,6 +391,50 @@ public static function three_decimal_currencies() {
 		];
 	}
 
+	/**
+	 * Returns the currencies that may be used as the store currency at checkout.
+	 *
+	 * Multi-currency plugins can add their configured currencies so Stripe's admin
+	 * availability checks are not limited to the WooCommerce base currency.
```
daledupreez: Not blocking: Maybe we can rephrase this to be more constructive?
```suggestion
	 * Multi-currency plugins can specify their configured currencies via the
	 * wc_stripe_supported_store_currencies filter so we can
	 * correctly flag which currencies may be used on the store.
```

daledupreez:
@diegocurbelo, can you update the PR description and testing instructions to reflect the changes? The actual behaviour is now driven by the new `wc_stripe_supported_store_currencies` filter, so I want to confirm that I am seeing the expected behaviour.

includes/class-wc-stripe-helper.php
diegocurbelo: Updated in [5c89e31](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5851/commits/5c89e3147625da2212bfc484f6c046cddcab5153)

client/settings/general-settings-section/__tests__/general-settings-section.test.js
diegocurbelo: Restored the notices when the currency is not in the list returned by the new filter in [5c89e31](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5851/commits/5c89e3147625da2212bfc484f6c046cddcab5153)

PR #5852, opened 2026-08-24
Manually scope the account cache to the requested mode

diegocurbelo:
Fixes STRIPE-1403

This PR changes the account cache handling to keep live and test account data fully isolated when both connections are configured, by scoping the account cache to the requested mode using a param:
- Read and write cached Stripe account data using the explicitly requested live or test mode.
- Preserve the existing current-mode behavior when callers do not specify a mode.
- Clear both mode-specific account cache entries when account data is invalidated.

_**NOTE:**_ The public method signatures and option-key format are unchanged.

The main change is that `clear_cache()` now invalidates both site-scoped mode entries, which is backward-compatible for callers and prevents stale account data after credentials or connection state change.

Clearing account data for both modes, should not have any performance impact as switching modes is not frequent, and the only case this would cover is switching to a mode and before the cache expires switch back. Forcing re-fetch of the account data after a mode switch should fix STRIPE-1403 specifically but also keeps the re-connect flow more predictable. 

## Testing instructions

1. Connect a test Stripe account
2. Go to `WooCommerce > Settings > Payments` and select `Stripe`
3. In the `Settings` tab, open the `Configure connection` modal, switch to the Live tab and connect a live account
4. Confirm the connection is successful
5. Check that teh cache entry `wcstripe_cache_test_account_data` was removed (using the Dev Tools, or with `wp option get wcstripe_cache_live_account_data`)

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


changelog.txt
```diff
@@ -24,6 +24,7 @@ xxxx-xx-xx - version 11.0.0
 * Fix - Stop sending Level 3 data when paying with a non-card payment method through express checkout (e.g. Amazon Pay), which Stripe rejects and can disable Level 3 data for card payments
 * Fix - Record the charge-captured state for asynchronously confirmed payments (e.g. ACH) so refunds from wp-admin and the Stripe Dashboard behave correctly
 * Fix - Reject negative refund amounts with an explicit error instead of silently refunding the absolute value
+* Fix - Allow explicitly scoping the account cache by mode and clear cached account data for both modes
```
wjrosa: Nit: this describes the cache internals rather than what a merchant gets, and reads like an Add/Dev entry for what is a Fix. I suggest something along the lines of "Fix - Prevent one mode's cached Stripe account data from being served to the other mode when both test and live connections are configured" (but shorter). Same for the entry in readme.txt

includes/class-wc-stripe-account.php
```diff
@@ -166,7 +167,7 @@ private function cache_account( $mode = null ) {
 		$account_cache = json_decode( wp_json_encode( $account ), true );
 
 		// Create or update the account data cache.
-		WC_Stripe_Database_Cache::set( self::ACCOUNT_CACHE_KEY, $account_cache, self::ACCOUNT_CACHE_EXPIRATION );
+		WC_Stripe_Database_Cache::set_with_mode( self::ACCOUNT_CACHE_KEY, $account_cache, self::ACCOUNT_CACHE_EXPIRATION, $mode );
```
wjrosa, line 171: The matching invalidation in `WC_Stripe_API` (class-wc-stripe-api.php:419, the invalid-key threshold path) still calls the mode-less `delete()`, which resolves to the settings mode. Now that this write is mode-scoped, checking a broken test connection on a live store deletes the valid live entry when the 401 threshold trips, while the stale test entry it meant to invalidate survives. That delete site needs the same `$mode` scoping.

includes/class-wc-stripe-account.php
```diff
@@ -120,7 +120,7 @@ public function get_cached_account_data( $mode = null, bool $force_refresh = fal
 		}
 
 		if ( ! $force_refresh ) {
-			$account = $this->read_account_from_cache();
+			$account = $this->read_account_from_cache( $mode );
```
wjrosa, line 127: Now that the caller's `$mode` reaches the cache layer, two coercion conventions disagree: `add_key_prefix()` coerces any unrecognized string to `test`, while `set_secret_key_for_mode()` and `is_connected()` coerce it to the current mode. So `get_cached_account_data( 'Live' )` on a live store (public API, and the settings controller feeds `$_POST['mode']` toward it) fetches the live account but caches it under `wcstripe_cache_test_account_data`, serving live data to test-mode reads for up to 2 hours. Normalizing `$mode` once at the top of `get_cached_account_data()` closes it.

includes/class-wc-stripe-account.php
```diff
@@ -133,10 +133,11 @@ public function get_cached_account_data( $mode = null, bool $force_refresh = fal
 	/**
 	 * Read the account from the WP option we cache it in.
 	 *
+	 * @param string|null $mode Optional. The mode to get the account data for. 'live' or 'test'. Default will use the current mode.
 	 * @return array empty when no data found, otherwise returns the cached data
 	 */
-	private function read_account_from_cache() {
-		$account_cache = WC_Stripe_Database_Cache::get( self::ACCOUNT_CACHE_KEY );
+	private function read_account_from_cache( $mode = null ) {
+		$account_cache = WC_Stripe_Database_Cache::get_with_mode( self::ACCOUNT_CACHE_KEY, $mode );
```
wjrosa, line 144: The prefetch layer under this read is mode-blind: with prefetch enabled (`wc_stripe_database_cache_prefetch_window` filter), an explicit-mode read of a near-expiry entry queues a prefetch by the unprefixed key, and the handler (class-wc-stripe-database-cache-prefetch.php:267) runs `get_cached_account_data( null, true )`, refreshing the current mode instead of the requested one. The mode-less tracking option then suppresses the correct prefetch for the whole window, so the requested mode expires and the next read blocks on a synchronous fetch.

includes/class-wc-stripe-account.php
```diff
@@ -133,10 +133,11 @@ public function get_cached_account_data( $mode = null, bool $force_refresh = fal
...
 
 		return false === $account_cache ? [] : $account_cache;
```
wjrosa: `get_with_mode()` returns `null` on a miss, never `false`, so this guard is dead and the method returns `null`, contradicting the `@return array` in the docblock this PR just extended. `return $account_cache ?? [];` restores the documented contract.

diegocurbelo: Fixed in 700e0173c

includes/class-wc-stripe-account.php
diegocurbelo: Fixed in 19cb2a87a

includes/class-wc-stripe-account.php
diegocurbelo: Normalized in 700e0173c

includes/class-wc-stripe-account.php
diegocurbelo: Addressed in d76335fb1

includes/class-wc-stripe-api.php
```diff
@@ -344,7 +365,8 @@ public static function request( $request, $api = 'charges', $method = 'POST', $w
 	public static function retrieve( $api ) {
 		// If keep count of consecutive 401 errors, and it exceeds INVALID_API_KEY_ERROR_COUNT_THRESHOLD,
 		// we return null until the cache expires (INVALID_API_KEY_ERROR_COUNT_CACHE_TIMEOUT) or the keys are updated.
-		$invalid_api_key_error_count = WC_Stripe_Database_Cache::get( self::INVALID_API_KEY_ERROR_COUNT_CACHE_KEY );
+		$mode                        = self::get_mode_for_active_secret_key();
+		$invalid_api_key_error_count = WC_Stripe_Database_Cache::get_with_mode( self::INVALID_API_KEY_ERROR_COUNT_CACHE_KEY, $mode );
```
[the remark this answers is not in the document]
diegocurbelo: Rejecting this suggestion because the current-mode fallback is intentional for backward compatibility.
Before this PR, `retrieve()` always scoped invalid-key tracking and account-cache invalidation to the current plugin mode; changing the behavior is outside this fix and potentially backward-incompatible.

wjrosa (approved):
Looks good to me 👍 . Thanks for addressing the suggestions

PR #5853, opened 2026-08-24
Prevent Classic Checkout submitting an empty payment_method when missing a required-field

diegocurbelo:
Fixes STRIPE-1402

This PR changes the Classic Checkout to:

- Treat a required-field wrapper’s checkboxes as one group.
    A "tick at least one" group is complete when any box is ticked, instead of only the first. This is the reported bug: a populated custom field blocked Stripe PaymentMethod creation.
- Stop treating a `<select multiple>` as empty when a placeholder is selected alongside a real option.
    select.value returns only the first selected option, so a selected placeholder masked the real selection.
- Never block on radios.
    Radios stay outside the selector, matching WooCommerce core, so a radio-only wrapper is skipped. A checked radio can still satisfy a wrapper it shares with other controls.

- Reject requests with neither a payment method nor a confirmation token before any Stripe lookup or intent creation.

Compared to 10.9.0 the gate differs in exactly three cases, all of which now allow checkout where it previously blocked: the checkbox group, the multi-select, and the checked radio. Nothing became stricter.

## Testing instructions

_**NOTE:**_ The report mentions [FooEvents for WooCommerce](https://worldpressit.com/downloads/fooevents-for-woocommerce/), but that's a paid plugin not hosted by Woo, so we don't have access to it, so the next steps just simulate the extra fields it would add in the checkout using JS

1. Configure Stripe with a test account
2. Change the checkout to classic (or add a new page for Classic Checkout)
3. Add a product to the cart and go to the Classic Checkout
4. Add the extra checkbox fields
    Open the browser Dev Tools and use this JS snippet:
    ```
    const checkout = document.querySelector( 'form.checkout' );
    const field = document.createElement( 'p' );

    field.id = 'stripe-1402-reproducer';
    field.className = 'form-row form-row-wide validate-required';
    field.innerHTML = `
	    <label>
		    Choose at least one option
		    <abbr class="required" title="required">*</abbr>
	    </label>
	    <label>
		    <input type="checkbox" name="stripe_1402_choice[]" value="first">
		    First option
	    </label>
	    <label>
		    <input type="checkbox" name="stripe_1402_choice[]" value="second" checked>
		    Second option
	    </label>
    `;

    checkout.prepend( field );
    ```
5. Verify the fields were added to the checkout form:
    Screenshot 2026-08-24 at 21 32 11
6. Place the order using the the test CC 4242424242424242
    - In develop this results in an error:
        Screenshot 2026-08-24 at 21 32 00
    - In this branch, the order succeeds (same as in `10.8.5`)
        And it redirects to the Order Received page
7. Repeat with both checkboxes left unchecked and verify that the order is not placed and checkout reports:
    `There was an error processing the payment: Your payment details were not submitted. Please review the checkout form and try again.`

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/classic/upe/__tests__/payment-processing.test.js
+++ b/client/classic/upe/__tests__/payment-processing.test.js
@@ -1779,13 +1779,24 @@ describe( 'payment-processing', () => {
 			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
 		} );
 
-		it( 'returns true when a required radio group has no selection', () => {
+		it( 'returns true when no checkbox in a required group is checked', () => {
+			form.innerHTML =
+				'<p class="validate-required">' +
+				'<input type="checkbox" />' +
+				'<input type="checkbox" />' +
+				'</p>';
+			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+		} );
+
+		// Radios stay outside the selector: blocking on an unselected group would stop
+		// checkout on stores whose radio fields have no server-side validation.
+		it( 'returns false for a required wrapper holding only unselected radios', () => {
 			form.innerHTML =
 				'<p class="validate-required">' +
 				'<input type="radio" name="option" />' +
 				'<input type="radio" name="option" />' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
 		} );
 
 		it( 'returns false when a required radio group has a selection', () => {
@@ -1797,11 +1808,13 @@ describe( 'payment-processing', () => {
 			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
 		} );
 
-		it( 'returns false when a grouped field has a value after an empty placeholder', () => {
+		// The checkbox group check only applies when the group leads the wrapper, so a
+		// filled text field ahead of an unchecked checkbox still passes, as before.
+		it( 'returns false when a filled text input precedes an unchecked checkbox', () => {
 			form.innerHTML =
 				'<p class="validate-required">' +
-				'<select><option value="" selected>Select...</option></select>' +
-				'<input type="radio" name="option" value="yes" checked />' +
+				'<input type="text" class="input-text" value="John" />' +
+				'<input type="checkbox" />' +
 				'</p>';
 			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
 		} );
--- a/client/classic/upe/__tests__/payment-processing.test.js
+++ b/client/classic/upe/__tests__/payment-processing.test.js
@@ -1754,6 +1754,43 @@ describe( 'payment-processing', () => {
 			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
 		} );
 
+		// WooCommerce posts a multiselect as implode( ', ', $value ), so both of the
+		// cases below still arrive as '' and are rejected server-side. Blocking them
+		// costs nothing; the third case posts ', US' and is accepted, so blocking it
+		// would fail the checkout outright.
+		it( 'returns true when a required multiple select has nothing selected', () => {
+			form.innerHTML =
+				'<p class="validate-required">' +
+				'<select multiple>' +
+				'<option value="">Select...</option>' +
+				'<option value="US">US</option>' +
+				'</select>' +
+				'</p>';
+			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+		} );
+
+		it( 'returns true when a required multiple select has only an empty option selected', () => {
+			form.innerHTML =
+				'<p class="validate-required">' +
+				'<select multiple>' +
+				'<option value="" selected>Select...</option>' +
+				'<option value="US">US</option>' +
+				'</select>' +
+				'</p>';
+			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+		} );
+
+		it( 'returns false when a required multiple select has a value selected after an empty option', () => {
+			form.innerHTML =
+				'<p class="validate-required">' +
+				'<select multiple>' +
+				'<option value="" selected>Select...</option>' +
+				'<option value="US" selected>US</option>' +
+				'</select>' +
+				'</p>';
+			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
+		} );
+
 		it( 'returns true when a required checkbox is unchecked', () => {
 			form.innerHTML =
 				'<p class="validate-required">' +
--- a/client/classic/upe/__tests__/payment-processing.test.js
+++ b/client/classic/upe/__tests__/payment-processing.test.js
@@ -1705,10 +1705,13 @@ describe( 'payment-processing', () => {
 			document.body.appendChild( form );
 		} );
 
-		const getWrappers = () => form.querySelectorAll( '.validate-required' );
+		const getRequiredFieldWrappers = () =>
+			form.querySelectorAll( '.validate-required' );
 
 		it( 'returns false when there are no required fields', () => {
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				false
+			);
 		} );
 
 		it( 'returns true when a required text input is empty', () => {
@@ -1716,7 +1719,9 @@ describe( 'payment-processing', () => {
 				'<p class="validate-required">' +
 				'<input type="text" class="input-text" value="" />' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				true
+			);
 		} );
 
 		it( 'returns true when a required text input has only whitespace', () => {
@@ -1724,7 +1729,9 @@ describe( 'payment-processing', () => {
 				'<p class="validate-required">' +
 				'<input type="text" class="input-text" value="   " />' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				true
+			);
 		} );
 
 		it( 'returns false when all required text inputs have values', () => {
@@ -1735,7 +1742,9 @@ describe( 'payment-processing', () => {
 				'<p class="validate-required">' +
 				'<input type="text" class="input-text" value="john@example.com" />' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				false
+			);
 		} );
 
 		it( 'returns true when a required select has no value', () => {
@@ -1743,7 +1752,9 @@ describe( 'payment-processing', () => {
 				'<p class="validate-required">' +
 				'<select><option value="">Select...</option><option value="US">US</option></select>' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				true
+			);
 		} );
 
 		it( 'returns false when a required select has a value', () => {
@@ -1751,7 +1762,9 @@ describe( 'payment-processing', () => {
 				'<p class="validate-required">' +
 				'<select><option value="">Select...</option><option value="US" selected>US</option></select>' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				false
+			);
 		} );
 
 		// WooCommerce posts a multiselect as implode( ', ', $value ), so both of the
@@ -1766,7 +1779,9 @@ describe( 'payment-processing', () => {
 				'<option value="US">US</option>' +
 				'</select>' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				true
+			);
 		} );
 
 		it( 'returns true when a required multiple select has only an empty option selected', () => {
@@ -1777,7 +1792,9 @@ describe( 'payment-processing', () => {
 				'<option value="US">US</option>' +
 				'</select>' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				true
+			);
 		} );
 
 		it( 'returns false when a required multiple select has a value selected after an empty option', () => {
@@ -1788,7 +1805,9 @@ describe( 'payment-processing', () => {
 				'<option value="US" selected>US</option>' +
 				'</select>' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				false
+			);
 		} );
 
 		it( 'returns true when a required checkbox is unchecked', () => {
@@ -1796,7 +1815,9 @@ describe( 'payment-processing', () => {
 				'<p class="validate-required">' +
 				'<input type="checkbox" />' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				true
+			);
 		} );
 
 		it( 'returns false when a required checkbox is checked', () => {
@@ -1804,7 +1825,9 @@ describe( 'payment-processing', () => {
 				'<p class="validate-required">' +
 				'<input type="checkbox" checked />' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				false
+			);
 		} );
 
 		it( 'returns false when one checkbox in a required group is checked', () => {
@@ -1813,7 +1836,9 @@ describe( 'payment-processing', () => {
 				'<input type="checkbox" />' +
 				'<input type="checkbox" checked />' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( false );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				false
+			);
 		} );
 
 		it( 'returns true when no checkbox in a required group is checked', () => {
@@ -1822,7 +1847,9 @@ describe( 'payment-processing', () => {
 				'<input type="checkbox" />' +
 				'<input type="checkbox" />' +
 				'</p>';
-			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
+			expect( hasEmptyRequiredFields( getRequiredFieldWrappers() ) ).toBe(
+				true
+			);
 		} );
 
 		// Radios stay outside the selector: blocking on an unselected group would stop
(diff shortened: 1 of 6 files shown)


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3517,6 +3517,13 @@ protected function prepare_payment_information_from_request( WC_Order $order ) {
 		} else {
 			$payment_method_id = sanitize_text_field( wp_unslash( $_POST['wc-stripe-payment-method'] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
 
+			if ( '' === $payment_method_id && empty( $_POST['wc-stripe-confirmation-token'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
+				throw new WC_Stripe_Exception(
+					'Payment method ID is missing from the request.',
+					__( "The selected payment method isn't valid.", 'woocommerce-gateway-stripe' )
+				);
+			}
```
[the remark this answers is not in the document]
diegocurbelo: > Validate the request before updating an existing intent.

Not fixing in this PR. Hoisting the check into `process_payment()` means duplicating the saved-token branch and running it on the checkout-session path too — wider than the bug this PR fixes. The ordering predates this PR.

`update_intent()` only runs when `wc_payment_intent_id` is posted and the selected method is non-deferred (BLIK and ACSS) and both Classic and Blocks _always_ post `wc-stripe-payment-method` anyway, so no real submission reaches that state.

And it also isn’t unvalidated: `update_intent()` calls `validate_intent_for_order()` first, which enforces amount, currency, payment-method type and order to intent linkage before mutating anything

> Normalize and validate wc-stripe-confirmation-token before selecting its flow.

The raw empty() is deliberately correct: it mirrors the two pre-existing raw checks at lines 1561 and 3594 that select the flow. Sanitizing in the guard while flow selection stays raw is exactly how you get a guard that throws on requests the rest of the method would happily route to the token branch

_Drafted with AI; reviewed by me_

client/classic/upe/payment-processing.js
```diff
@@ -1688,23 +1688,27 @@ export const confirmWalletPayment = async ( api, jQueryForm ) => {
...
-			if ( ! input.checked ) {
-				return true;
+		const hasValue = [ ...inputs ].some( ( input ) => {
+			if ( input.type === 'checkbox' || input.type === 'radio' ) {
+				return input.checked;
 			}
-		} else if ( ! input.value || input.value.trim() === '' ) {
+
+			return input.value?.trim() !== '';
+		} );
```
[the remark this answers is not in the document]
daledupreez: Is this worth adding in this pass? If not, we should create a follow-up issue to tackle this relatively soon.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -870,6 +870,31 @@ public function provide_process_payment_deferred_intent_returns_valid_response()
...
+			->expects( $this->never() )
+			->method( 'create_and_confirm_payment_intent' );
+
+		$this->mock_gateway
+			->expects( $this->never() )
+			->method( 'stripe_request' );
+
+		$response = $this->mock_gateway->process_payment( $order->get_id() );
+
+		$this->assertSame( 'failure', $response['result'] );
```
daledupreez, line 899: Please reset `$_POST` in this test.

client/classic/upe/payment-processing.js
```diff
@@ -1688,23 +1688,27 @@ export const confirmWalletPayment = async ( api, jQueryForm ) => {
...
-		if ( input.type === 'checkbox' ) {
-			if ( ! input.checked ) {
-				return true;
+		const hasValue = [ ...inputs ].some( ( input ) => {
+			if ( input.type === 'checkbox' || input.type === 'radio' ) {
+				return input.checked;
 			}
-		} else if ( ! input.value || input.value.trim() === '' ) {
+
+			return input.value?.trim() !== '';
```
daledupreez: I worked with an AI reviewer to flag some issues with this updated approach. I'll transcribe the findings with light edits, but the core issue is that including the radio items _and_ the `.some()` check has some unexpected impacts relative to the code in WooCommerce core.

- Adding `input[type="radio"]` to the selector makes the gate *stricter* than
  both its previous self and WooCommerce core, in a fix whose whole purpose was to loosen it. A
  `.validate-required` wrapper containing only unselected radios used to be invisible to the gate; it now
  blocks PaymentMethod creation. The new Jest case asserts that as the desired behaviour.
- Claude reproduced the predicate side-by-side against a transcription of WC core's own `.validate-required` scan to check this rather than reason about it — matrix below. Your intended fix (row A) works exactly as designed. Rows B, E and F are the changes that occur as side-effects.
- The `.some()` rewrite also *loosens* the gate for wrappers where the first control is empty but a later
  one is filled (a day/month/year select triple, say). Not necessarily bad, but it could have unexpected side-effects, and the existing "multiple fields" test doesn't cover it, because it uses two separate wrappers.
- Worth a line in the PR description: this will fail on a gate/server disagreement, which reverses the
  tradeoff #4556 chose explicitly ("Creating a couple of extra payment methods is harmless compared to blocking users from paying"). While that change is defensible, we should make sure we note it explicitly.

### Behaviour matrix

`coreFn` is a jQuery-free transcription of WC core's `.validate-required` scan
(`client/legacy/js/frontend/checkout.js`, identical in released 10.6.1 and trunk); `oldFn`/`newFn` are the
literal pre- and post-PR implementations. _(I can share the script for this if needed.)_

| `.validate-required` wrapper contains | WC core | 10.9.0 | this PR | |
|---|---|---|---|---|
| one-of-many checkbox group, 1 checked | pass | BLOCK | pass | **A — the intended fix; now matches core** |
| radio group, none selected | pass | pass | **BLOCK** | **B — new blocking class, stricter than core** |
| radio group, one selected | pass | pass | pass | C |
| 2 required text inputs, only 1st filled | pass | pass | pass | D — unchanged |
| 2 required text inputs, only 2nd filled | BLOCK | BLOCK | **pass** | **E — now looser than core** |
| day/month/year selects, only year defaulted | BLOCK | BLOCK | **pass** | **F — now looser than core** |
| lone placeholder select | BLOCK | BLOCK | BLOCK | unchanged |
| terms checkbox unchecked + hidden `terms-field` | BLOCK | BLOCK | BLOCK | unchanged, correct |
| filled text + unchecked required checkbox | BLOCK | pass | pass | pre-existing divergence |
| lone required `textarea.input-text` | pass | pass | pass | pre-existing blind spot on both sides |


_Written by AI; edited by me._

daledupreez (commented):
My AI review tool also wrote the following about the handling for radio buttons:

#### Radio-only required wrappers now block PaymentMethod creation

`client/classic/upe/payment-processing.js:1697` · confidence 0.85

The widened selector matches radios, and the wrapper is evaluated as one group. For a wrapper containing only radios with none selected, this was previously invisible to the gate — WC core renders radios with `class="input-radio"`, never `input-text` (`wc-template-functions.php:3563`), so the old selector matched nothing and the wrapper was skipped. Now the selector matches, nothing is checked, `hasEmptyRequiredFields()` returns `true`, `deferred-intent.js` returns early, no `wc-stripe-payment-method` is appended — and because WooCommerce's own AJAX submit proceeds regardless, the new server guard at `class-wc-stripe-upe-payment-gateway.php:3535` throws. `handle_process_payment_error()` marks the order FAILED and shows the shopper a message that names nothing they can fix.

This is the STRIPE-1402 disagreement class in the opposite direction, and the new Jest case `returns true when a required radio group has no selection` enshrines it rather than guarding against it.

**Scope, verified rather than assumed:** if the radio field is *also* a registered WC checkout field with `required => true`, `WC_Checkout::validate_posted_data()` (`class-wc-checkout.php:961`) errors first and `process_payment()` never runs, so everything works as expected. The problematic path needs `.validate-required` radio markup that is **not** server-enforced as a WC checkout field: third-party markup carrying the class for styling, or conditionally-required fields. That is a smaller set of situations than "every radio group", but this PR is specifically aimed at these specific issues.

The reason this is worth fixing is as follows:
 * A radio field that genuinely must be filled and *is* registered with WC gets caught by `validate_posted_data()` regardless, so blocking it in the client buys nothing. However a radio field that isn't server-enforced now results in a failed purchase that would previously have completed successfully. The change is weakly negative wherever it has any effect at all.

**What if we** decided *whether to evaluate* a wrapper using core's selector, and *whether it's satisfied* using the widened one?

```js
const coreControls = wrapper.querySelectorAll( 'input.input-text, select, input[type="checkbox"]' );
if ( ! coreControls.length ) {
    continue; // core skips these wrappers too
}
const controls = wrapper.querySelectorAll(
    'input.input-text, select, input[type="checkbox"], input[type="radio"]'
);
const hasValue = [ ...controls ].some( ... );
```

The AI tool checked this variant against the same script used for the behaviour matrix above: it preserves every behaviour currently in the PR, including the `empty select + checked radio` test case, and removes only the new radio-only blocking. The `returns true when a required radio group has no selection`
case would need to be inverted.

_Written by AI; edited and approved by me._

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
diegocurbelo: Done in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5853/commits/3f9a45499d9a223bdebf73e0bb941d8fffc1bde9

client/classic/upe/payment-processing.js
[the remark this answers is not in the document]
diegocurbelo: Fixed in [16b0aeb](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5853/commits/16b0aeb198b5713da173075ab92d46835bfdc83b)

client/classic/upe/payment-processing.js
diegocurbelo: Adopted the two-selector suggestion and re-ran your matrix. An independent transcription reproduces your `core`, `10.9.0` and `this PR` columns exactly; `now` is the current branch.

| `.validate-required` wrapper contains | core | 10.9.0 | reviewed | now | |
|---|---|---|---|---|---|
| one-of-many checkbox group, 1 checked | pass | BLOCK | pass | **pass** | A |
| radio group, none selected | pass | pass | **BLOCK** | **pass** | B — resolved |
| radio group, one selected | pass | pass | pass | pass | C |
| 2 text inputs, only 1st filled | pass | pass | pass | pass | D |
| 2 text inputs, only 2nd filled | BLOCK | BLOCK | **pass** | **BLOCK** | E — resolved |
| d/m/y selects, only year defaulted | BLOCK | BLOCK | **pass** | **BLOCK** | F — resolved |
| lone placeholder select | BLOCK | BLOCK | BLOCK | BLOCK | |
| terms unchecked + hidden `terms-field` | BLOCK | BLOCK | BLOCK | BLOCK | |
| filled text + unchecked checkbox | BLOCK | pass | pass | pass | pre-existing |
| lone `textarea.input-text` | pass | pass | pass | pass | pre-existing |
| multi-select: placeholder **+** real selected | pass | BLOCK | BLOCK | **pass** | K — new fix |
| multi-select: nothing selected | pass | BLOCK | BLOCK | BLOCK | L |
| multi-select: only placeholder selected | pass | BLOCK | BLOCK | BLOCK | M |
| checked radio + empty placeholder select | BLOCK | BLOCK | pass | **pass** | N — your suggestion |
| unchecked radio + empty placeholder select | BLOCK | BLOCK | BLOCK | BLOCK | O |

Three rows change from 10.9.0 — A, K, N — all BLOCK → pass. Nothing gets stricter.

B, E and F are back to 10.9.0 values. Radios are out of the selector so a radio-only wrapper is skipped as core skips it, but a checked radio still satisfies a wrapper (N, with O as its negative) — your suggestion, as one extra query rather than a second `querySelectorAll`. Values keep first-element semantics, which is what keeps E and F out.

L and M are stricter than core deliberately: core never flags a multi-select because jQuery 3's `.val()` returns an array, but `get_posted_data()` posts one as `implode( ', ', $value )` — `''` for both, so the server rejects them regardless. K posts `', US'`, which the server accepts, so blocking it was a hard failure. That's the row that changed.

client/classic/upe/__tests__/payment-processing.test.js
```diff
@@ -1754,6 +1754,43 @@ describe( 'payment-processing', () => {
...
+		// would fail the checkout outright.
+		it( 'returns true when a required multiple select has nothing selected', () => {
+			form.innerHTML =
+				'<p class="validate-required">' +
+				'<select multiple>' +
+				'<option value="">Select...</option>' +
+				'<option value="US">US</option>' +
+				'</select>' +
+				'</p>';
+			expect( hasEmptyRequiredFields( getWrappers() ) ).toBe( true );
```
daledupreez: Not blocking, but it feels like we could rename this helper to something more self-descriptive like `getRequiredDOMElements()`.

client/classic/upe/payment-processing.js
```diff
@@ -1693,18 +1693,40 @@ export const confirmWalletPayment = async ( api, jQueryForm ) => {
...
-				return true;
-			}
-		} else if ( ! input.value || input.value.trim() === '' ) {
+		const first = inputs[ 0 ];
+		let isEmpty;
+
+		if ( first.type === 'checkbox' ) {
+			// A required wrapper can hold a "tick at least one" checkbox group, so the
+			// first box being unchecked does not mean the group is incomplete.
+			isEmpty = ! inputs.some( ( input ) => input.checked );
```
daledupreez: Is it worth checking that we have checkbox fields in `inputs`?
```suggestion
			isEmpty = ! inputs.some( ( input ) => input.type === 'checkbox' && input.checked );
```

client/classic/upe/payment-processing.js
```diff
@@ -1693,18 +1693,40 @@ export const confirmWalletPayment = async ( api, jQueryForm ) => {
...
+		} else if ( first.type === 'select-multiple' ) {
+			// `value` is only the first selected option, so a selected empty placeholder
+			// would hide the real selections behind it.
+			isEmpty = ! [ ...first.selectedOptions ].some(
+				( option ) => option.value.trim() !== ''
+			);
+		} else {
+			isEmpty = first.value.trim() === '';
+		}
```
daledupreez: The `! expression.some()` checks here are pretty complex to parse. Might it be worth flipping the condition so we negate in the check below? (Not blocking.)
```suggestion
		let hasValidValue;

		if ( first.type === 'checkbox' ) {
			// A required wrapper can hold a "tick at least one" checkbox group, so the
			// first box being unchecked does not mean the group is incomplete.
			hasValidValue = inputs.some( ( input ) => input.type === 'checkbox' && input.checked );
		} else if ( first.type === 'select-multiple' ) {
			// `first.value` is only the first selected option, so a selected empty placeholder
			// would hide the real selections behind it.
			hasValidValue = [ ...first.selectedOptions ].some(
				( option ) => option.value.trim() !== ''
			);
		} else {
			hasValidValue = first.value.trim() !== '';
		}
```

client/classic/upe/payment-processing.js
```diff
@@ -1693,18 +1693,40 @@ export const confirmWalletPayment = async ( api, jQueryForm ) => {
...
+
+		// Radios are outside the selector above, so a wrapper holding only radios is
+		// skipped entirely, as it is by WC core. They can still satisfy a wrapper
+		// though: a checked radio means the shopper filled the field in, and blocking
+		// on it would fail a checkout the server would have accepted.
```
daledupreez: I think we can rephrase this slightly.
```suggestion
		// Radios are not checked by the selector above, so a wrapper holding only radios is
		// skipped entirely, as it is by WooCommerce core. Radios can still cause a required
		// element to be present: a checked radio means the shopper filled the field in, so
		// we should let the server validate the submission further.
```

daledupreez (commented):
Given the changes you made in [this thread](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5853#discussion_r3877844034), I am not sure as to what updates we still need for the PR description, as the changes are now more nuanced, especially in terms of radio support.

I am also seeing what I think is a related failure mode. If I run your test script and then ensure that both of the custom fields are unchecked, classic checkout still tries to submit incomplete checkout data. That then results in a somewhat misleading `There was an error processing the payment: The selected payment method isn't valid.` error message (which seems to stem from the server changes here). The PR description doesn't capture what the correct or expected behaviour is for this one.

Should we be returning false from the `$( 'form.checkout' ).on( generateCheckoutEventNames()` even handler in the case where we have a Stripe payment method selected and the required fields are missing? This would be instead of the basic `return` when we have missing required fields?

client/classic/upe/payment-processing.js
diegocurbelo: Changed in [ce6f197](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5853/commits/ce6f197ce33fb9212c96669da5c41e432324ef0c)

client/classic/upe/payment-processing.js
diegocurbelo: Updated in [ce6f197](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5853/commits/ce6f197ce33fb9212c96669da5c41e432324ef0c)

client/classic/upe/__tests__/payment-processing.test.js
diegocurbelo: Renamed to `getRequiredFieldWrappers ` in [937179f](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5853/commits/937179f634f9cfbde9e6f98d933321f624673be6)

client/classic/upe/payment-processing.js
diegocurbelo: Changed in [62d6337](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5853/commits/62d6337827a2705dfa52acd0dcdd8d54bd2fdb29)

diegocurbelo:
> Given the changes you made in [this thread](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5853#discussion_r3877844034), I am not sure as to what updates we still need for the PR description, as the changes are now more nuanced, especially in terms of radio support.
> 
> I am also seeing what I think is a related failure mode. If I run your test script and then ensure that both of the custom fields are unchecked, classic checkout still tries to submit incomplete checkout data. That then results in a somewhat misleading `There was an error processing the payment: The selected payment method isn't valid.` error message (which seems to stem from the server changes here). The PR description doesn't capture what the correct or expected behaviour is for this one.
> 
> Should we be returning false from the `$( 'form.checkout' ).on( generateCheckoutEventNames()` even handler in the case where we have a Stripe payment method selected and the required fields are missing? This would be instead of the basic `return` when we have missing required fields?

On return false — I dug into this and it would cause two problems.

- The gate runs before `isUsingSavedPaymentMethod()`, so returning false would also abort saved-token checkouts, which work fine today.
- Core puts its entire AJAX submit and all its error rendering (`submit_error`, `show_inline_errors`) inside the `triggerHandler( … ) !== false` block, so returning false shows the shopper nothing at all. To make it viable we’d need to move the gate inside the saved-method check, mark the wrappers ourselves, add a notice and manage focus (basically rebuilding the client-side gate that #4556 removed).

What I’ve pushed instead is a fix to the message itself, which was simply describing the wrong thing:
Instead of:
`There was an error processing the payment: The selected payment method isn’t valid.`

It now shows:
`There was an error processing the payment: Your payment details were not submitted. Please review the checkout form and try again.`

The gateway the shopper picked is fine; what’s missing are the payment details the browser should have attached. The new wording _deliberately names no cause_, because that guard is shared by classic, Blocks, ECE and the subscription change-payment flow (and also used when Stripe.js fails to load).

client/classic/upe/payment-processing.js
```diff
@@ -1693,18 +1693,44 @@ export const confirmWalletPayment = async ( api, jQueryForm ) => {
...
+		// still satisfy a wrapper: it is not proof the other controls are filled, but
+		// it is reason enough to let the server validate the submission rather than
+		// block it here.
+		const hasCheckedRadio = !! wrapper.querySelector(
+			'input[type="radio"]:checked'
+		);
+
+		if ( ! hasValue && ! hasCheckedRadio ) {
 			return true;
 		}
```
daledupreez, line 1744: Not blocking, but I think we can simplify this by handling the `hasValue` check before we look at the radios.
```suggestion
		if ( hasValue ) {
			return false;
		}

		// Radios are not matched by the selector above, so a wrapper holding only
		// radios is skipped entirely, as it is by WooCommerce core. A checked radio can
		// still satisfy a wrapper: it is not proof the other controls are filled, but
		// it is reason enough to let the server validate the submission rather than
		// block it here.
		const hasCheckedRadio = !! wrapper.querySelector(
			'input[type="radio"]:checked'
		);

		if ( ! hasCheckedRadio ) {
			return true;
		}
```

daledupreez (approved):
Thanks for updating the description and confirming the expected behaviour. I think this is now ready to merge, though I think it's worth expanding the testing instructions to add the fields you recommended using your script _and_ leaving them both unchecked to confirm the new error message is displayed as expected. (I tested this, but I think it should be part of the PR.)

I do have one minor superficial comment that's not blocking.

client/classic/upe/payment-processing.js
diegocurbelo: Applied, with `continue` rather than `return false`.

`return false` exits `hasEmptyRequiredFields()` entirely, so the first satisfied wrapper would report “nothing is empty” and the rest would never be inspected.

I also updated the testing instructions with the both-unchecked case.

PR #5856, opened 2026-08-25
Mark orders awaiting payment in the session so carts clear after redirect payments

### LINKED ISSUE #5854: Amazon Pay redirect flow does not clear cart on mobile, can cause duplicate orders
### Description

When a customer uses Amazon Pay on mobile (Chrome/Android), the payment completes successfully but the cart is not cleared. When this occurs, the customer sees a full cart, assumes the payment failed, and places the order again with a different payment method, resulting in two paid orders for the same items.

### Steps to reproduce

On a mobile device (Android/Chrome), add items to cart
Proceed to checkout and select Amazon Pay
Complete the Amazon Pay authorization on Amazon's site
Get redirected back to the store

### Expected behavior

Cart is cleared, customer sees the order confirmation page.

### Actual behavior
Payment succeeds (charge confirmed, order status moves to Processing), but the cart retains all items. Customer navigates back to checkout and can place (and pay for) the same order again.

### Root cause
WC()->cart->empty_cart() is session-dependent. It requires the customer's browser session cookie to identify which cart to clear. On redirect-based payment methods like Amazon Pay, the return redirect on mobile can lose the session cookie (new tab, cross-domain cookie restrictions, mobile browser context switching). The Stripe webhook confirms the payment server-side, but webhooks run without a browser session and cannot clear the cart.

The redirect return handler should not silently fall through when the session is missing. It has access to the payment_intent ID from the return URL, which maps to the order, which has the customer's session key. It should look up and clear the session-stored cart even when the cookie is absent.

### Evidence from a real occurrence

Two orders placed 2.5 minutes apart, same customer, same items, same cart session ID (wcal_abandoned_cart_id is identical on both)
Order 1: Amazon Pay via store-api, paid at 14:31 EDT
Order 2: Credit card via checkout, paid at 14:33 EDT
Same session_start_time on both — customer never left the session
Session pages went from 4 (order 1) to 7 (order 2) — customer navigated back to checkout after the redirect
Different IPv6 addresses between orders (mobile redirect changed network path)
Billing name slightly differs on different payment method (auto populated by Amazon Pay, manually entered on Credit card second checkout): confirms different input flows

### System report

WooCommerce Gateway Stripe: 10.8.5
WooCommerce: 11.0.1
WordPress: 7.0.4
PHP: 8.2.30
Server: LiteSpeed
Payment method: Amazon Pay (stripe_amazon_pay)
Device: Android 10, Chrome 151 (mobile)

</details>

wjrosa:
Fixes STRIPE-1410
Fixes #5854

When a redirect payment method (Amazon Pay, Klarna, Alipay, …) completes but the return redirect lands in a browser context without the shopper's session cookie (common on mobile app/webview handoffs), the order is paid but the original cart is never cleared — inviting a duplicate order.

WC core self-heals this on any later page load via the `order_awaiting_payment` session key, but only classic checkout sets it — Store API checkout never does (reported upstream as woocommerce/woocommerce#68022). This PR sets the key from `WC_Stripe_UPE_Payment_Gateway::process_payment()` for orders needing payment, mirroring classic checkout, so core clears the cart on the shopper's next page load once the order is paid. Skipped when changing a subscription's payment method, where the order ID is a subscription whose non-pending status would make core clear an unrelated cart.

Trade-off: the cart clears on the next page load, not at webhook time — clearing another session's persisted cart would require direct `woocommerce_sessions` writes.

**Backward compatibility:** no signatures, hooks, or option keys change; we only write a session key core already owns and sets on the classic path. Not tested under multisite (all state involved is site-scoped).

## Testing instructions

Automated: `npm run test:php -- --filter test_process_payment_sets_order_awaiting_payment_in_session`

Manual (any redirect method; Alipay in test mode is easiest):

1. In browser A, add a product to the cart and check out as a guest via the block checkout with Alipay.
2. When redirected to the Stripe test page, copy its URL into a private/incognito window (browser B) and click *Authorize Test Payment* there.
3. Browser B shows the order-received page; the order moves to Processing.
4. In browser A, open the Cart page. **Expected:** empty cart. Without the fix it still contains the product, and the shopper can pay again.

---

-   [x] Tested on mobile (or does not apply) — backend-only change; the mobile context switch is simulated by the incognito repro

### Changelog entry

<details>

#### Comment

Changelog entry added to `changelog.txt` and `readme.txt` under 11.0.0.

</details>

**Post merge**


includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1543,6 +1543,21 @@ public function add_email_currency_conversion_notice( $order, bool $sent_to_admi
...
+		// subscription payment-method changes, where $order_id is a non-pending subscription and
+		// core would clear an unrelated cart.
+		$order = wc_get_order( $order_id );
+		if ( $order instanceof WC_Order
+			&& $order->needs_payment()
+			&& WC()->session
+			&& is_callable( [ WC()->session, 'set' ] )
+			&& ! $this->is_changing_payment_method_for_subscription() ) {
+			WC()->session->set( 'order_awaiting_payment', $order_id );
+		}
```
daledupreez, line 1578: I have multiple comments on this:
1. Should we only set the value in the session when it is not already set? And by that I mean should we check if the value is `null`, `false`, or `0` before overwriting a value that may be set by WooCommerce Core. Or we can maybe check if the value doesn't match the current order ID. It feels like we should only add the handling when it was not already done by WooCommerce Core, so we are acting as more of a fallback.
2. I think we should also save the session data after setting this value, as that is what the upstream WooCommerce code does.

https://github.com/woocommerce/woocommerce/blob/232f8c942f78d0caefc6187f84251a9ec2756488/plugins/woocommerce/includes/class-wc-checkout.php#L1150-L1156

3. Nit: With the current indentation, it's a bit tricky to distinguish the conditions from the code.
4. Nit: We can stash the result of `WC()->session` so we're not calling the function multiple times.
```suggestion
		$wc_session = WC()->session;
		if ( $order instanceof WC_Order
			&& $order->needs_payment()
			&& $wc_session
			&& is_callable( [ $wc_session, 'get' ] )
			&& is_callable( [ $wc_session, 'set' ] )
			&& ! $this->is_changing_payment_method_for_subscription()
			&& empty( $wc_session->get( 'order_awaiting_payment' ) ) // Use empty() to check for 0, false, and null
		) {
			$wc_session->set( 'order_awaiting_payment', $order_id );
			if ( is_callable( [ $wc_session, 'save_data' ] ) ) {
				// Make sure the order ID is saved before we can trigger errors.
				$wc_session->save_data();
			}
		}
```

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -870,6 +870,67 @@ public function provide_process_payment_deferred_intent_returns_valid_response()
...
+				],
+			],
+			self::MOCK_CARD_PAYMENT_INTENT_TEMPLATE
+		);
+
+		$_POST = [
+			'payment_method'               => 'stripe',
+			'wc-stripe-payment-method'     => 'pm_mock',
+			'wc-stripe-confirmation-token' => '',
+		];
```
daledupreez, line 910: I don't see any code that cleans these values up. Can we ensure that happens before we run any assertions?

daledupreez (commented):
This seems like a reasonable approach, but I think we should see how the WooCommerce core folks approach this, as I am not clear if this flag was intentionally skipped for blocks.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Updated it in eeab9f08d. Two changes from your suggestion:

1. Kept the set unconditional. Core itself overwrites ([class-wc-checkout.php#L1150](https://github.com/woocommerce/woocommerce/blob/232f8c942f78d0caefc6187f84251a9ec2756488/plugins/woocommerce/includes/class-wc-checkout.php#L1150)), and the only case where `empty()` changes behavior is a stale key from an earlier failed/abandoned attempt. Skipping the write there leaves the key pointing at a dead order, so the cart-clear self-heal never fires for the new order; Store API checkout has no equivalent of classic's "resume or overwrite" recovery.
2. While verifying, I noticed `process_payment()` also runs on pay-for-order, where core never sets this key; setting it there would clear the shopper's unrelated cart after paying an old order. Added a `did_action( 'woocommerce_before_pay_action' )`

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Added a clear for the session key in `tear_down()` (17e63c6a9)

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1543,6 +1543,31 @@ public function add_email_currency_conversion_notice( $order, bool $sent_to_admi
 	 * @return array|null An array with result of payment and redirect URL, or nothing.
 	 */
 	public function process_payment( $order_id, $retry = true, $force_save_source = false, $previous_error = false, $use_order_source = false ) {
+		// Store API checkout, unlike classic, never sets `order_awaiting_payment`, so WC core's
+		// wc_clear_cart_after_payment() can't clear the cart when a redirect payment's return
+		// misses the session cookie (mobile app/browser handoffs), leaving a paid-for cart that
+		// invites a duplicate order. Mirror WC_Checkout::process_order_payment() here. Skipped for
+		// subscription payment-method changes, where $order_id is a non-pending subscription and
+		// core would clear an unrelated cart.
```
daledupreez: Note that this is in the process of being changed upstream in https://github.com/woocommerce/woocommerce/pull/68102, so I think we should rephrase the commentary here to reflect that this change is to serve as a fallback.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1543,6 +1543,31 @@ public function add_email_currency_conversion_notice( $order, bool $sent_to_admi
...
+		$order      = wc_get_order( $order_id );
+		$wc_session = WC()->session;
+		if ( $order instanceof WC_Order
+			&& $order->needs_payment()
+			&& $wc_session
+			&& is_callable( [ $wc_session, 'set' ] )
+			&& ! $this->is_changing_payment_method_for_subscription()
+			// Core only sets this key on the checkout path; pay-for-order must not claim the
+			// session's cart, or it gets cleared after a payment for an unrelated order.
+			&& ! did_action( 'woocommerce_before_pay_action' )
```
daledupreez, line 1570: Given that this will be something that happens in WooCommerce Core, I think we should only set the value if it is not already set.
```suggestion
			&& ! did_action( 'woocommerce_before_pay_action' )
			&& (string) $order_id !== (string) ( $wc_session->get( 'order_awaiting_payment' ) ?? '' )
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Reworded in 9993575ad. The comment now says this is a fallback for WooCommerce before 11.2.0, where core does not set the key for Store API checkout, and that the block can go once WC 11.2 is the oldest version we support.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Thanks, the suggestion is in. PHPCS asked for the Yoda order on that comparison, so I flipped the two sides in 16104279f.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1551,6 +1551,34 @@ public function add_email_currency_conversion_notice( $order, bool $sent_to_admi
...
 	 */
 	public function process_payment( $order_id, $retry = true, $force_save_source = false, $previous_error = false, $use_order_source = false ) {
+		// Fallback for WooCommerce versions before 11.2.0, where Store API checkout never sets
+		// `order_awaiting_payment` (classic checkout always has). Without it, WC core's
+		// wc_clear_cart_after_payment() can't clear the cart when a redirect payment's return
+		// misses the session cookie (mobile app/browser handoffs), leaving a paid-for cart that
+		// invites a duplicate order. Mirror WC_Checkout::process_order_payment() here, leaving the
+		// key alone when core already set it; drop this block once WC 11.2 is the oldest supported
+		// version. Skipped for subscription payment-method changes, where $order_id is a
+		// non-pending subscription and core would clear an unrelated cart.
```
daledupreez: I think we should cut this comment down. This is _really_ long!
```suggestion
		// Fallback for WooCommerce versions before 11.2.0, where Store API checkout never sets
		// `order_awaiting_payment` (classic checkout always has). We need to set the flag so WC core's
		// wc_clear_cart_after_payment() clears the cart when a redirect payment's return
		// doesn't have the session cookie (mobile app/browser handoffs). This removes a paid-for cart
		// from the session.
		// The logic matches WC_Checkout::process_order_payment(), and only adds the value when unset.
		// Skipped for subscription payment-method changes, where $order_id is a
		// non-pending subscription ID, not an order ID.
```
I think we should add the TODO as an issue.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1551,6 +1551,34 @@ public function add_email_currency_conversion_notice( $order, bool $sent_to_admi
...
+		$order      = wc_get_order( $order_id );
+		$wc_session = WC()->session;
+		if ( $order instanceof WC_Order
+			&& $order->needs_payment()
+			&& $wc_session
+			&& is_callable( [ $wc_session, 'set' ] )
+			&& ! $this->is_changing_payment_method_for_subscription()
+			// Core only sets this key on the checkout path; pay-for-order must not claim the
+			// session's cart, or it gets cleared after a payment for an unrelated order.
```
daledupreez: Let's make this less verbose as well.
```suggestion
			// Don't set the flag for the pay-for-order path.
```

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1551,6 +1551,34 @@ public function add_email_currency_conversion_notice( $order, bool $sent_to_admi
...
+			&& ! did_action( 'woocommerce_before_pay_action' )
+			&& (string) ( $wc_session->get( 'order_awaiting_payment' ) ?? '' ) !== (string) $order_id
+		) {
+			$wc_session->set( 'order_awaiting_payment', $order_id );
+			// Persist now, like core does: if the gateway request below hangs, the session is
+			// never saved at shutdown and a resubmit can create a duplicate order.
```
daledupreez:
```suggestion
			// Ensure the flag is saved rather than waiting until the end of the request.
```

daledupreez (approved):
The changes look good to me, but I have some complaints about the verbose comments! 🙃

PR #5857, opened 2026-08-25
Route 3DS card payments past the wallet/voucher flow under Dynamic Payment Methods

wjrosa:
Fixes STRIPE-1411

#5683 made eligible on-session Optimized Checkout payments create intents with `automatic_payment_methods`, so the intent's `payment_method_types` lists every PMC-enabled method instead of the customer's selection. Two places still read that list as the selection: the voucher/wallet branches in `get_redirect_url()`, which sent 3DS card payments into the wallet/voucher client flow ("Invalid wallet payment type", order stuck pending), and the intersect in `process_payment_with_payment_method()` that suppressed returning `payment_method` for Blocks to save.

Both sites now gate on the resolved selected payment type via a shared private helper, keeping the intent-types intersect only as a fallback when no selection reference is available.

**Backward compatibility:** no public symbols, hooks, or option keys change. Explicit-types intents (non-OC, subscriptions, saved tokens, express) list only the selected method, so their behavior is unchanged.

**Note on the 3DS experience:** under Dynamic Payment Methods the 3DS challenge is a full-page redirect to Stripe's hosted page, not the on-page modal. That comes from #5683: the intent is confirmed server side with a `return_url` and without `use_stripe_sdk`, and for that shape Stripe returns `next_action.redirect_to_url`. This PR does not change that behavior, it is just the first place it becomes visible, because before this fix the 3DS payment failed before reaching the redirect. Restoring the modal is tracked in #5892.

Verified locally: 3DS card `4000 0027 6000 3184` completes on classic and Blocks checkout (order reaches `processing`); selecting WeChat Pay still shows the QR modal. Follow-up under STRIPE-1411: add an SCA scenario to the `optimized-checkout` E2E project.

## Testing instructions

1. Use a test account whose PMC has WeChat Pay (or Cash App Pay / a voucher method) active, with Optimized Checkout on and Adaptive Pricing not active for the session.
2. On classic checkout, pay with the 3DS test card `4000 0027 6000 3184` using a new payment method. The hosted 3D Secure page should open; click **Complete** and confirm the order is `processing`.
3. Repeat on Blocks checkout.
4. Pay with WeChat Pay: the QR modal should still appear.

---

### Changelog entry

<details>

#### Comment

</details>

**Post merge**


tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -957,6 +957,105 @@ public function test_process_payment_deferred_intent_with_required_action_return
 		$this->assertMatchesRegularExpression( "/#wc-stripe-confirm-pi:{$order_id}:{$mock_intent->client_secret}/", $response['redirect'] );
 	}
 
+	/**
+	 * A 3DS card payment must not be routed into the wallet/voucher client flow when the intent's
+	 * `payment_method_types` also lists wallet/voucher methods, as happens under Dynamic Payment
+	 * Methods where the intent carries every PMC-enabled method rather than the customer's selection.
```
daledupreez: Nit: We should use Optimized Checkout Suite.
```suggestion
	 * `payment_method_types` also lists wallet/voucher methods, as happens under Optimized Checkout
	 * Suite where the intent lists every PMC-enabled method rather than the customer's selection.
```

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -957,6 +957,105 @@ public function test_process_payment_deferred_intent_with_required_action_return
...
+	 *
+	 * @param object|null $next_action               The intent's next_action.
+	 * @param string      $expected_redirect_pattern  Regex the response redirect must match.
+	 * @param bool        $expects_payment_method     Whether the response must expose `payment_method` for Blocks to save.
+	 *
+	 * @dataProvider provide_process_payment_deferred_intent_with_dpm_intent_types_uses_card_flow
+	 */
+	public function test_process_payment_deferred_intent_with_dpm_intent_types_uses_card_flow( $next_action, $expected_redirect_pattern, $expects_payment_method ) {
```
daledupreez: Nit: As above, `ocs` instead of `dpm`.
```suggestion
	 * @dataProvider provide_process_payment_deferred_intent_with_ocs_intent_types_uses_card_flow
	 */
	public function test_process_payment_deferred_intent_with_ocs_intent_types_uses_card_flow( $next_action, $expected_redirect_pattern, $expects_payment_method ) {
```

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
daledupreez: My AI reviewer flagged that we should test that things still work for WeChat Pay when card and other elements are in this list. So it could work to expand this list to include card and Link. Though maybe this test case belongs on the OCS gateway test class?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1810,7 +1810,7 @@ private function process_payment_with_payment_method( int $order_id ) {
 			if ( in_array( $payment_intent->status, WC_Stripe_Intent_Status::REQUIRES_CONFIRMATION_OR_ACTION_STATUSES, true )
 				&& WC_Stripe_Payment_Methods::BLIK !== $selected_payment_type ) {
 				$wallet_and_voucher_methods        = array_merge( WC_Stripe_Payment_Methods::VOUCHER_PAYMENT_METHODS, WC_Stripe_Payment_Methods::WALLET_PAYMENT_METHODS );
-				$contains_wallet_or_voucher_method = isset( $payment_intent->payment_method_types ) && count( array_intersect( $wallet_and_voucher_methods, $payment_intent->payment_method_types ) ) !== 0;
+				$contains_wallet_or_voucher_method = $this->is_payment_using_method_types( $wallet_and_voucher_methods, $this->get_selected_payment_type_from_info( $payment_information ), $payment_intent );
```
daledupreez: Should we re-derive `$selected_payment_type` before this code block? It feels like `$selected_payment_type` might not match the return value from `$this->get_selected_payment_type_from_info( $payment_information )`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -4437,6 +4437,27 @@ protected function get_redirect_url( $return_url, $payment_intent, $payment_info
...
+	 * `payment_method_types` lists every PMC-enabled method, not the customer's selection.
+	 * The intent-types intersect remains as a fallback when there is no selection reference.
+	 *
+	 * @param string[] $payment_method_types  Payment method types to match against.
+	 * @param string   $selected_payment_type The resolved selected payment type, '' when unknown.
+	 * @param object   $payment_intent        The Stripe payment intent.
+	 *
+	 * @return bool
+	 */
+	private function is_payment_using_method_types( array $payment_method_types, string $selected_payment_type, $payment_intent ): bool {
```
daledupreez, line 4453: My AI review tool flagged the following concern with the pre-existing code. I am not sure that it's worth tackling this in this PR.

#### Under OC, a degraded Stripe PM fetch can resolve the selection to `'card'`**

`includes/payment-methods/class-wc-stripe-upe-payment-gateway.php:4453`

`WC_Stripe_API::retrieve()` can return `null` (401 error-count threshold) or a `WP_Error` (outage) without throwing; `prepare_payment_information_from_request()` normalizes that to `(object) []` (`:3557-3560`), so `payment_method_details->type` is absent, the OCS override falls through to the parent, and under OC the form value maps to `'card'` (`:3846`). A WeChat Pay payment then resolves to `'card'` and skips the wallet branch.

Being straight about the delta: **the pre-change code was equally broken in this state** — it emitted `#wc-stripe-wallet-{id}:card:…`, which `confirmWalletPayment`'s `default` case rejects outright (`client/classic/upe/payment-processing.js:1633-1637`). So the change makes an already-broken path *quieter*, it does not create one. It also needs a stale 401 counter or an outage affecting GETs while POSTs still succeed. Recording it for awareness rather than as an ask.

*A tempting fix that I'd advise against:* having the OCS override return `''` when `payment_method_details->type` is missing. That condition is not specific to API degradation — `payment_method_details` is `(object) []` on **every** confirmation-token/express OCS payment (whenever `payment_method_id` is empty). Returning `''` there reaches `intent-controller:1281-1282`, where `get_payment_method_instance( '' )` returns `null` and `$payment_method->is_reusable()` is then dereferenced whenever `has_subscription` is truthy — a fatal on an OCS express subscription purchase. It would also emit `#wc-stripe-voucher-{id}::{secret}:{url}`, which the client hash parser expects to have non-empty segments. If this is ever worth closing, it needs scoping to "PM id was present but the fetch failed" *plus* a null guard at the intent-controller site.

_Written by AI; approved by me._

daledupreez (commented):
Overall, this is working as described. That said, there is one behavioural change that is not documented in this PR: 3DS challenges now require a full-page redirect and no longer show a modal window. Is that intentional or expected?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: They read the same array key and nothing changes it in between, so they cannot diverge. But I switched to the local variable to make that clear: 47af77b49

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
wjrosa: Done in 47af77b49. I kept the test here because the routing logic lives in the UPE gateway and the OCS gateway inherits it.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Thanks for recording this. Agreed it is pre-existing and fails the same way before this change, so I will keep it out of this PR and note it as a possible follow-up.

daledupreez:
@wjrosa, this is working as described, but I still don't know whether the "loss" of the 3DS modal in favour of a full-page redirect is expected or intentional, as that is not documented in your comments or the PR description.

wjrosa:
It is not intentional in this PR, it comes from #5683. The DPM intent is confirmed server side with a `return_url` and without `use_stripe_sdk`, and for that shape Stripe returns a hosted redirect for 3DS instead of the modal. I confirmed this against the API in test mode: adding `use_stripe_sdk: true` to the same request brings back the modal next action.

So the modal can be restored with a small change (we already send `use_stripe_sdk` for setup intents, just not for payment intents). I would do that in a follow-up to keep this PR scoped to the bug, but I can fold it in here if you prefer.

daledupreez:
@wjrosa, I think it's worth a note in the PR description that the upstream PR was the original source of the change, and this PR doesn't modify it. The key issue is that this PR is the first place where this new behaviour for 3DS is visible, so we should have some explicit documentation for that.

And I am more than happy to have this be tackled in a follow-up rather than this PR.

wjrosa:
Done. I added a note to the PR description explaining that the redirect behavior comes from #5683 and is only first visible here, and I opened #5892 for restoring the modal in a follow-up.

daledupreez (approved):
Thanks for making the updates to the PR description. 🚢

Issue #5734, opened 2026-07-23
In confirmation token flow, UPE payment type is not set when checking level 3 data


daledupreez:
### Description

https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5681 added checks to prevent Level 3 data being sent for non-card purchases. However, there is one code path where that doesn't work as expected: when we're handling the confirmation-token flow, the UPE payment type has not yet been persisted, so we don't prevent level 3 data from being included.

The specific path is via `WC_Stripe_UPE_Payment_Gateway->process_payment_with_confirmation_token()` , where the code can call either `create_and_confirm_payment_intent()` or `update_and_confirm_payment_intent()`. For the `create*` path, we call `request_with_level3_data()` before writing the payment method meta. For the `update*` path, we don't write the payment method meta.

_This finding was AI-generated._

### Steps to reproduce

I didn't reproduce this yet, but it should be reproducible by paying using Amazon Pay or Link via Express Checkout.

### Expected behavior

We should ensure that we don't send Level3 data for non-card payment methods.

PR #5858, opened 2026-08-26
Stop caching Level 3 rejections caused by non-card requests

jaclync:
Follow-up to #5826 (see "Follow-up hardening" in its description). Scope was narrowed after review — see the review threads for the descoping rationale.

One rejected non-card Level 3 request can disable Level 3 account-wide for 3 months (the `wc_stripe_level3_not_allowed` transient), silently dropping Level 3 data for legitimate card payments — that's how #5734 manifested. This PR scopes that caching without changing when Level 3 is *sent*:

- **Don't cache rejections caused by the request itself.** When the rejected request declares only non–Level 3 payment method types (e.g. `['sepa_debit']`, `['amazon_pay']`), the rejection reflects that one request, not the account's Level 3 gating, so the transient is no longer set. The payment still succeeds via the existing retry without Level 3. Rejections with card-family types, or with no declared types at all, cache for 3 months exactly as before — zero change for any healthy flow.
- **Persist the selected payment type earlier** in `process_payment_with_confirmation_token()`, mirroring the payment-method flow, so parallel readers (e.g. webhooks fired by the confirmation) see it in the database.

Earlier revisions also made the send-decision prefer the request's `payment_method_types` over the order meta; that was reverted per review to avoid behavior changes (Level 3 is sent exactly as on `develop`).

**BC impact:** no public signatures, hooks, or option keys change; the new helpers are private. The only behavior delta is the narrower rejection caching described above. Known limitation: a flow that neither writes the order meta nor declares request types could still cache an ambiguous rejection the old way.

## Testing instructions

The rejection path requires a Stripe account that is *not* gated for Level 3, so it's covered by PHPUnit rather than manual steps: a 7-row rejection matrix in `tests/phpunit/class-wc-stripe-api-test.php` (card/mixed/absent types cache; non-card-only types don't; malformed type lists count as absent) plus an early-persistence test in the UPE gateway tests.

Manual regression on a US store in test mode:

1. A card payment succeeds and its `payment_intents` request still carries `level3` (Level 3–gated accounts).
2. An Amazon Pay express payment succeeds, its request carries no `level3`, and no `_transient_wc_stripe_level3_not_allowed` row appears in `wp_options`.

---

### Changelog entry

<details>

#### Comment

</details>

**Post merge**


includes/class-wc-stripe-api.php
```diff
@@ -550,6 +554,69 @@ public static function request_with_level3_data( $request, $api, $level3_data, $
...
+	 * @return string[] The declared payment method types, or an empty array when the request
+	 *                  carries none (e.g. capture or confirm calls).
+	 */
+	private static function get_request_payment_method_types( $request ) {
+		if ( isset( $request['payment_method_types'] ) && is_array( $request['payment_method_types'] ) ) {
+			return $request['payment_method_types'];
+		}
+
+		return [];
+	}
```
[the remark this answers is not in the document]
jaclync: Good catch, the request array passes through public filters before the gate reads it, so the entries are untrusted. Fixed in 290262263.

includes/class-wc-stripe-api.php
```diff
@@ -522,9 +522,13 @@ public static function request_with_level3_data( $request, $api, $level3_data, $
 		);
 
 		if ( $is_level3_param_not_allowed ) {
-			// Set a transient so that future requests do not add level 3 data.
-			// Transient is set to expire in 3 months, can be manually removed if needed.
-			set_transient( 'wc_stripe_level3_not_allowed', true, 3 * MONTH_IN_SECONDS );
+			// Only cache the rejection for payments known to support level3: for an unknown
+			// payment type it's ambiguous and would disable level3 account-wide.
+			if ( self::is_payment_known_to_support_level3_data( $request, $order ) ) {
```
wjrosa, line 527: The gate and the cache resolve the unknown case in opposite directions: when the request has no `payment_method_types` and the order has no payment-type meta, the gate still attaches Level 3 (the empty-meta default is true), but the rejection isn't cached because the type is "unknown". So flows like legacy `confirm_intent` and the order-handler captures re-send Level 3 and get rejected on every payment, where pre-PR the first rejection cached for 3 months.

Maybe cache ambiguous rejections with a short TTL, or read the type from the rejected intent's response?

includes/class-wc-stripe-api.php
```diff
@@ -550,6 +554,76 @@ public static function request_with_level3_data( $request, $api, $level3_data, $
...
+	private static function payment_supports_level3_data( $request, $order ) {
+		$payment_method_types = self::get_request_payment_method_types( $request );
+
+		if ( [] !== $payment_method_types ) {
+			// A supported type among the candidates is enough (e.g. [card, link]): Stripe
+			// only rejects level3 when no type on the intent supports it.
+			return [] !== array_intersect( $payment_method_types, WC_Stripe_Payment_Methods::LEVEL3_SUPPORTED_PAYMENT_METHODS );
+		}
+
+		return WC_Stripe_Helper::order_supports_level3_data( $order );
```
wjrosa, line 577: This fallback doesn't hold on Optimized Checkout + PMC, the scenario the changelog targets: under PMC the intent controller strips `payment_method_types`, and the OC confirmation-token flow always writes `card` meta (the consolidated form value; the OCS override only replaces it when a payment method id is present). So a shopper picking Klarna/Affirm inside the OC element still attaches Level 3 via the wrong `card` meta, and a rejection still sets the 3-month account-wide transient. The meta behavior is pre-existing, but it's a hole in this PR's stated fix, and the new gateway test pins `oc_enabled = false`, so the OC path is untested. Worth either handling or scoping the claim.

includes/class-wc-stripe-api.php
```diff
@@ -550,6 +554,76 @@ public static function request_with_level3_data( $request, $api, $level3_data, $
...
+	 *
+	 * @return bool Whether Level 3 data applies to the payment.
+	 */
+	private static function payment_supports_level3_data( $request, $order ) {
+		$payment_method_types = self::get_request_payment_method_types( $request );
+
+		if ( [] !== $payment_method_types ) {
+			// A supported type among the candidates is enough (e.g. [card, link]): Stripe
+			// only rejects level3 when no type on the intent supports it.
+			return [] !== array_intersect( $payment_method_types, WC_Stripe_Payment_Methods::LEVEL3_SUPPORTED_PAYMENT_METHODS );
```
wjrosa, line 574: Behavior change worth confirming: off-session renewals with a saved Link payment method send `payment_method_types: ['link']` (from the source object's type), and renewal orders don't carry the payment-type meta, so these card-rails charges now lose Level 3 where they previously sent it. If Stripe rejects L3 on link-typed intents this is a fix; otherwise `link` may belong in `LEVEL3_SUPPORTED_PAYMENT_METHODS`. Either way, worth stating in the PR description.

includes/class-wc-stripe-api.php
```diff
@@ -550,6 +554,76 @@ public static function request_with_level3_data( $request, $api, $level3_data, $
...
+	 *
+	 * Prefers the payment_method_types the request itself declares; the order meta is the
+	 * fallback for requests that carry no payment method info (e.g. capture-time calls).
+	 *
+	 * @param array    $request The request parameters.
+	 * @param WC_Order $order   The order associated with the payment.
+	 *
+	 * @return bool Whether Level 3 data applies to the payment.
+	 */
+	private static function payment_supports_level3_data( $request, $order ) {
```
wjrosa, line 568: This gate can now pass without ever touching `$order` (request types alone), which newly exposes the `invalid_request_error` logging branch to a non-`WC_Order` `$order`: that branch matches every error of that type and dereferences `$order->get_items()`. Internal callers are guarded, but `request_with_level3_data()` is public static API, so a third-party caller passing card types with a false order (previously safe, since the old gate returned false for it) now fatals there. An `instanceof` guard in the logging branch would close it.

includes/class-wc-stripe-api.php
jaclync: Good catch, I decided to reduce the scope in ec415596f. Rejections with no declared types now cache for 3 months exactly as on develop, so there's no per-payment retry regression for legacy flows. The only remaining change is that a rejection whose request declared **exclusively non-card types** skips the cache, since it reflects that one request rather than the account’s Level 3 gating.

includes/class-wc-stripe-api.php
jaclync: Thanks for pointing out some scenarios I didn't think of. I decided to revert most of the behavior changes here in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5858/changes/ec415596f3b917f5028a8ec663afa88028573ced, keeping just one behavior change to skip setting the transient when the request's `payment_method_types` do not support level 3 data.

Please feel free to point out any new issues from the latest implementation, I'm also happy to drop this change.

includes/class-wc-stripe-api.php
jaclync: The diffs were reverted from https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5858#discussion_r3977646869 to minimize the behavior changes without clear benefits.

includes/class-wc-stripe-api.php
jaclync: Also deleted, related to https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5858#discussion_r3977646869.

jaclync:
Thanks for the review! I reverted and scoped down the changes to reduce the risk of regressions, since the request and order's payment methods data have lots of edge cases. Ready for another pass!

wjrosa (approved):
Looks good to me and works as intended! 👍

PR #5863, opened 2026-08-26
Exploratory work on Transactions UI

daledupreez:
Fixes #<github_issue_id>

## Testing instructions

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**


.nvmrc
```diff
@@ -1 +1 @@
-v20.18.1
+v20.20.2
```
jaclync, line 1: Is the Node bump to v20.20.2 intentional for this PR? If so, we might want to update the version (or refer to this source of truth) in `AGENTS.md`.

client/entrypoints/payments-admin/constants.js
```diff
@@ -0,0 +1,62 @@
...
+	canceled: 'gray',
+};
+
+export const PAYOUT_STATUS_LABELS = {
+	paid: __( 'Paid', 'woocommerce-gateway-stripe' ),
+	pending: __( 'Pending', 'woocommerce-gateway-stripe' ),
+	incomplete: __( 'Incomplete', 'woocommerce-gateway-stripe' ),
+	upcoming: __( 'Upcoming', 'woocommerce-gateway-stripe' ),
+	canceled: __( 'Canceled', 'woocommerce-gateway-stripe' ),
+};
```
jaclync, line 29: The status keys don't seem to completely match the `status` values in the [Stripe doc](https://docs.stripe.com/api/payouts/object?api-version=2026-08-26.preview&rds=1#payout_object-status) (`paid, pending, in_transit, canceled, failed`), is it based on another reference? This might explain why my payouts in `failed` status, the status had a gray color rather than a color that indicates a failure:

Image

client/entrypoints/payments-admin/style.scss
```diff
@@ -0,0 +1,26 @@
...
+
+		.dataviews-wrapper {
+			background-color: $stripe-payments-background-color;
+		}
+	}
+}
+
+body.wp-admin.payments_page_wc-stripe-payments {
+	background-color: $stripe-payments-background-color;
+}
```
jaclync: super nit: new line for EOF.

client/entrypoints/payments-admin/payouts-fields.js
```diff
@@ -0,0 +1,91 @@
...
+			),
+	},
+	{
+		id: 'bank_details',
+		label: __( 'Bank details', 'woocommerce-gateway-stripe' ),
+		enableSorting: false,
+		enableHiding: true,
+		filterBy: false,
+		getValue: ( { item } ) =>
+			sprintf(
```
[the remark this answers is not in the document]
jaclync: +1 to this fallback.

client/entrypoints/payments-admin/payouts-table.js
```diff
@@ -0,0 +1,105 @@
...
+	const onChangeView = useCallback(
+		( nextView ) => {
+			// A different page size invalidates every cursor we collected.
+			if ( nextView.perPage !== view.perPage ) {
+				setCursors( [ null ] );
+				setView( { ...nextView, page: 1 } );
+				return;
+			}
+
+			if ( nextView.page > view.page ) {
```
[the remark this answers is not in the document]
jaclync: I believe this is a real issue to fix.

client/entrypoints/payments-admin/use-payouts.js
```diff
@@ -0,0 +1,81 @@
...
+				if ( cancelled || requestId !== requestIdRef.current ) {
+					return;
+				}
+
+				// Keep the last good page on screen so a transient failure does
+				// not blank the table out from under the reader.
+				setState( ( previous ) => ( {
+					...previous,
+					isLoading: false,
+					error: error?.message ?? null,
```
jaclync: Two potential issues:

- A rejection without a message results in null, which the table treats as success - no notice, and on first load it shows "No payouts found." to a merchant who has payouts. Maybe we could use a generic fallback message instead of falling back to success, e.g. `__( 'Unable to load payouts.', 'woocommerce-gateway-stripe' )`.
- Found during testing from a local PHP fatal error (not caused by the PR): `error.message` isn't always plain text: when the REST route hits a PHP fatal, WordPress responds with a message containing HTML (e.g. `<p>There has been a critical error…</p>`), and the notice renders the tags literally:

Image

client/entrypoints/payments-admin/payouts-table.js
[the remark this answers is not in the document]
jaclync: Also, does it make sense to early return when `isLoading` is true?

client/entrypoints/payments-admin/payouts-table.js
```diff
@@ -0,0 +1,105 @@
...
+	const [ cursors, setCursors ] = useState( [ null ] );
+
+	const cursor = cursors[ view.page - 1 ] ?? null;
+
+	const { data, hasMore, isLoading, error } = usePayouts( {
+		perPage: view.perPage,
+		cursor,
+	} );
+
+	const paginationInfo = useMemo( () => {
```
jaclync, line 67: Non-blocking: the pagination code feels a bit awkward here, given Stripe's cursor-based pagination in the API (no total pages/items). Instead of the "Page N of N+1" display in the UI where the `N+1` might change, maybe we can explore showing just the previous/next CTAs for the data view (if feasible) like in the Stripe dashboard.

client/entrypoints/payments-admin/payouts-fields.js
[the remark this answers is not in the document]
daledupreez: I improved the handling here in 14508c48bb5bbbc9429b44533726409c9c3a55c7.

client/entrypoints/payments-admin/payouts-table.js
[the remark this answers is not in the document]
daledupreez: I updated the implementation in 9292cf67e8b137c9986ff7db7aaa30f61255b882, which also addresses a timing gap.

client/entrypoints/payments-admin/style.scss
```diff
@@ -0,0 +1,26 @@
...
+			font-weight: 600;
+		}
+
+		.dataviews-wrapper {
+			background-color: $stripe-payments-background-color;
+		}
+	}
+}
+
+body.wp-admin.payments_page_wc-stripe-payments {
```
jaclync: In `payments_page_wc-stripe-payments` body class, the prefix `payments` depends on the localized parent page. I tested in German and the body class became `zahlungen_page_wc-stripe-payments`. We might want to use a class that doesn't depend on a localizable string.

includes/admin/class-wc-stripe-payments-ui-controller.php
```diff
@@ -0,0 +1,211 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Registers the Payments -> Stripe admin submenu and renders the Stripe Payments UI.
+ *
+ * @since 11.1.0
+ */
+class WC_Stripe_Payments_UI_Controller {
```
jaclync: Non-blocking nit: does it make sense to mark it as final?

includes/admin/class-wc-stripe-payments-ui-controller.php
```diff
@@ -0,0 +1,211 @@
...
+	 * Builds the params for the Payments UI.
+	 *
+	 * The minor-unit currency lists are passed through rather than duplicated in
+	 * JS so PHP stays the single source of truth for Stripe's exponent rules.
+	 *
+	 * @return array
+	 */
+	private function get_script_params() {
+		return [
+			'locale'                 => str_replace( '_', '-', get_user_locale() ),
```
jaclync, line 214: Non-blocking: could the client use `document.documentElement.lang` instead?

jaclync (commented):
The Payouts tab worked well! Left some initial questions/comments. I saw that there were some new commits after my initial review, feel free to ignore any comments that might be outdated.

.nvmrc
daledupreez: Yes, this is intentional -- one of the `@wordpress/dataviews` dependencies requires node 20.20.2. I've updated the PR description and `AGENTS.md` to reflect the bump.

client/entrypoints/payments-admin/payouts-table.js
daledupreez: I think that's a fair comment. In some of my earlier commits, the pagination was handled via manual buttons outside of the table. I definitely think that's something worth exploring from the UX/design perspective, though it should be pretty easy to handle in a follow-up.

client/entrypoints/payments-admin/use-payouts.js
daledupreez: I've added the fallback message in 3ed00ce71ea59f340e2389b854861c6f8c5032dc.

For non-string messages, I'd rather have an ugly representation than try to allow HTML, especially because the only failure modes where that might be expected would be fatal-equivalent errors.

client/entrypoints/payments-admin/style.scss
daledupreez: Excellent catch! I've updated the implementation to use the `admin_body_class` filter in 9f07543d4ff3e697e09e1b61db78e5d8f74bb1c8 so we can guarantee a specific class for this selector.

client/entrypoints/payments-admin/style.scss
daledupreez: Addressed in 9f07543d4ff3e697e09e1b61db78e5d8f74bb1c8.

client/entrypoints/payments-admin/constants.js
daledupreez: I missed checking the values -- the original set was AI-generated and I didn't confirm them. I've updated the values in d5bd5b7cb54145c332c0787d9d9c4023002a9279.

includes/admin/class-wc-stripe-payments-ui-controller.php
daledupreez: Good suggestion. Done in 9e9889d4412a362331ab5e108865fdb822104c2d.

PR #5864, opened 2026-08-26
Stop defining WOOCOMMERCE_CART during full cart/checkout page renders

wjrosa:
Fixes STRIPE-1414

Since 10.9.0, `get_cart_render_data()` (#5606) runs `build_display_items()` during full cart/checkout page renders, and its unconditional `define( 'WOOCOMMERCE_CART', true )` makes core's `is_cart()` return true on the checkout page for the rest of the request. Plugins branching on `is_cart()` misread checkout views as cart views (reported: GTM4WP firing `view_cart` instead of `begin_checkout`), and our own localized `is_cart_page` flag became `"1"` on checkout, making ECE init early instead of waiting for `updated_checkout`.

Fix:

- Remove the define from `build_display_items()`.
- Add it to the deprecated `ajax_get_shipping_options()`, the only AJAX entry point that relied on it transitively; the other ECE AJAX handlers already define it themselves, so all AJAX paths keep their behavior.

**Backward compatibility:** restores the constant's pre-10.9.0 scope (ECE AJAX requests only, never full page renders). No hooks, signatures, or option keys change.

## Testing instructions

1. Enable Apple Pay/Google Pay express checkout with the checkout button location enabled (default).
2. Add this snippet as a mu-plugin (or via a code snippets plugin):
   ```php
   add_action( 'wp_footer', function () {
       printf( '<!-- probe is_cart=%s const=%s -->', is_cart() ? 'YES' : 'no', defined( 'WOOCOMMERCE_CART' ) ? 'DEFINED' : 'undef' );
   }, PHP_INT_MAX );
   ```
3. Add a product to the cart, open the checkout page, and find the `probe` comment in the page source.
   - Without this PR: `is_cart=YES const=DEFINED`, and `wc_stripe_express_checkout_params` contains `"is_cart_page":"1"`.
   - With this PR: `is_cart=no const=undef`, and `"is_cart_page":""`.
4. On the cart page, the probe shows `is_cart=YES` and the express checkout button still renders with the correct total.
5. Pay for an order via Apple Pay or Google Pay from the cart or checkout page to confirm express checkout still works.

---

### Changelog entry

<details>

#### Comment

</details>

**Post merge**

https://claude.ai/code/session_018SExFbpGjvB7Eqw8JkNNhb

### DIFF
--- a/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
@@ -352,6 +352,10 @@ class WC_Stripe_Express_Checkout_Ajax_Handler {
 		check_ajax_referer( 'wc-stripe-express-checkout-shipping', 'security' );
 		_deprecated_function( __METHOD__, '10.9.0' );
 
+		if ( ! defined( 'WOOCOMMERCE_CART' ) ) {
+			define( 'WOOCOMMERCE_CART', true );
+		}
+
 		$shipping_address          = filter_input_array(
 			INPUT_POST,
 			[
--- a/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-helper.php
@@ -1794,10 +1794,9 @@ class WC_Stripe_Express_Checkout_Helper {
 	 * }
 	 */
 	public function build_display_items( $itemized_display_items = false ) {
-		// Must not define WOOCOMMERCE_CART here: this also runs during full page
-		// renders (get_cart_render_data()), where the constant would make core's
-		// is_cart() return true on the checkout page for the rest of the request.
-		// The ECE AJAX entry points define it themselves.
+		// Don't define WOOCOMMERCE_CART here: this also runs on full checkout page
+		// renders, where the constant would flip is_cart() to true. The ECE AJAX
+		// entry points define it themselves.
 
 		/**
 		 * Filters whether Express Checkout itemization should be hidden.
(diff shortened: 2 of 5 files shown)


includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
daledupreez, line 373: We should add a note to the PHPDoc for get_shipping_options() that the caller is responsible for specifying whether we are running in a WooCommerce cart context.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
```diff
@@ -1794,9 +1794,9 @@ protected function build_shipping_methods( $shipping_methods ) {
 	 * }
 	 */
 	public function build_display_items( $itemized_display_items = false ) {
```
daledupreez, line 1804: I think we should add a note to the PHPDoc to mention that callers are responsible for specifying whether we are running in a WooCommerce cart context, and this is a new requirement in 11.0.0.

readme.txt
daledupreez, line 40: I think we should add a compatibility note something like the following:

```suggestion
* 11.0.0
   - Some express checkout helpers now require the caller to specify whether they are in the WooCommerce Cart context
* 10.9.0
```

daledupreez (approved):
The changes look good to me overall, though I think we need some PHPDoc changes and a minor compatibility note to ensure that we make the changes explicit.

includes/payment-methods/class-wc-stripe-express-checkout-ajax-handler.php
wjrosa: Added in 5f38cd3e6.

includes/payment-methods/class-wc-stripe-express-checkout-helper.php
wjrosa: Added in 5f38cd3e6.

Issue #5811, opened 2026-08-14
Render Express Checkout buttons on page load for variable products, regardless of attribute selection


pierorocca:
On variable products, Stripe's Express Checkout buttons currently only render after a variation has been selected, rather than on page load whenever the product is otherwise eligible. This causes a layout shift as buttons appear once an attribute is chosen. WooPayments already renders its Express Checkout buttons on page load for variable products regardless of attribute selection — Woo Stripe should match that.

## Proposal

* Render Express Checkout buttons on page load whenever the product is otherwise eligible, without requiring a variation to be pre-selected.
* When a shopper attempts an express checkout before completing the attribute selection, defer to WooCommerce core's standard validation UX (highlighting unselected attribute fields) once available (see woocommerce/woocommerce#67749), instead of hiding/disabling the buttons.
* Validate across classic checkout, Blocks checkout, and the block-based Product page (Add to Cart + Options block).

## Related

* woocommerce/woocommerce#67749 (core: standardized validation pattern)
* Automattic/woocommerce-payments#12046 (WooPayments: adopt the same validation pattern)

PR #5868, opened 2026-08-27
Render express checkout buttons on page load for variable products and validate the selection at click time

jaclync:
Part of STRIPE-1366
Part of STRIPE-1367 / https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5811

> [!NOTE]
> **Size:** ~half of this diff is tests (~375 lines). Non-test code is ~355 changed lines: a new ~205-line DOM-reader module (`variable-product-page.js`) and a ~150-line rewire of the express-checkout entrypoint (net +1 there). The work is intentionally split for reviewability - this PR covers visibility + template compatibility only; the click-path amount-integrity hardening follows in a stacked PR.

## What

Express Checkout buttons on variable product pages now render on page load, before a variation is selected, on both product templates. ("Both templates" = the two ways WooCommerce renders the add-to-cart form on a product page: the **classic form** - the `.variations_form` markup driven by `wc-add-to-cart-variation.js`, used by classic themes and by block themes that haven't upgraded the block - and the **blockified Add to Cart + Options block**, the Interactivity-API-based form used in block-theme Single Product templates, still labeled beta but the default on new block-theme installs.) Clicking before completing the attribute selection shows the "Please select your product options before proceeding." prompt (WooPayments parity) instead of opening the wallet sheet; validation moved from gating *visibility* to gating the *click*.

Previously, visibility was driven by jQuery variation events that only the classic `.variations_form` script fires. The blockified **Add to Cart + Options** (iAPI) block fires none of them, so on that template the buttons never appeared for variable products without a default variation (STRIPE-1366); on the classic template they popped in after selection, causing a layout shift (STRIPE-1367).

## How

- `ready` no longer requires a selected variation to show the buttons.
- New template-agnostic DOM readers in `client/express-checkout/utils/variable-product-page.js`, built on the compatibility markup WooCommerce renders "to be used by extensions or Express Payment methods" (`AddToCartWithOptions::render()`): `.single_variation_wrap`, the bound `variation_id` input, and `attribute_*` fields.
- Supported markup generations: classic form; blockified WC 10.9+ (hidden `attribute_*` inputs); blockified WC 10.7/10.8 (pill radios / dropdown selects - no hidden inputs); blockified "legacy mode" (form-field extensions or redirect-after-add flip the block to a `form.cart` POST form).
- Variation changes on the blockified template are observed via MutationObserver on the `variation_id` input's value attribute; stepper quantity changes via a click listener (the steppers update the value property only - no events, no attribute writes).

## Alternatives considered

Please feel free to suggest the best approach for ECE in variable/general product pages.

- **Bridging the missing events** (observe the block's internals and re-fire the jQuery variation events, keeping the hide-until-selected model): not pursued - it preserves the initialization race and the layout shift, and it would be discarded when the render-on-load behavior (STRIPE-1367) landed anyway.
- **Reading the block's Interactivity API store directly**: not viable - the store is private and its shape has already changed across WooCommerce releases; the rendered compatibility markup is the surface the block maintainers recommend for express payment methods, so the DOM readers build on that instead.
- **WooPayments' model** (no live preview refresh at all; read the DOM once at click time): simpler - no observers or listeners - but the classic template has shipped live preview updates for years and dropping them would be a merchant-visible regression, and a click-time-only design still needs a fallback for Stripe's 1-second resolve deadline. We keep the live preview here and make the click path cart-authoritative in the follow-up PR, which converges on WooPayments' robustness where it matters without the UX regression.

## Backward compatibility

- Behavior change (intended): buttons are visible before variation selection on the classic template too; hide-on-deselect removed. No hooks, signatures, or option keys changed. Server endpoints untouched.
- Reach: mostly variable product pages, with two deliberate improvements on **simple** products - the click guard is now null-safe and blocks blockified forms reporting an invalid quantity, and blockified stepper (− / +) quantity changes now refresh the preview amount (they previously went unnoticed; steppers fire no input events). Cart, checkout, and pay-for-order pages are untouched.
- The charged amount is always computed server-side from the real cart at confirmation; this PR does not change the click/authorization path.
- Not tested under multisite; no site-scoped state is read or written beyond existing options.

## Testing

- Jest: DOM readers across all three markup generations; existing entrypoint harness (classic refresh contract) passes unchanged.
- E2E: variable-product spec updated to the new contract (visible pre-selection + options prompt on click). A blockified-template E2E path needs a block theme, i.e. a dedicated Playwright project - proposed as a follow-up.
- Live verification matrix (both templates, including blockified legacy mode, stepper/typed quantity, variation switching, Link sheet open) in the workspace test-matrix doc.

### Manual testing steps

Setup: Stripe test mode with express checkout enabled for product pages; a variable product with two attributes and **no default variation**, including one attribute combination with no variation; a simple product. Test each flow on **both templates** - classic (e.g. Storefront, or a block theme whose Single Product template uses the classic Add to Cart with Options block) and blockified (block theme, Site Editor → Templates → Single Product → replace the block with **Add to Cart + Options (Beta)**).

Variable products:
1. Load the page with nothing selected → express buttons are visible immediately (previously: hidden on classic until selection; never shown on blockified).
2. Click a button before selecting anything → "Please select your product options before proceeding." alert; the wallet does not open.
3. Select only one of the two attributes and click → same prompt.
4. Complete the selection → click → wallet sheet opens at the variation's price; complete a test payment and confirm the order total and line items.
5. Switch from variation A to B, click → sheet and order reflect B.
6. Clear/deselect the attributes → buttons stay visible; clicking prompts again.
7. Change the quantity - typed, and on blockified also via the − / + steppers → the amount refreshes.
8. Select the combination that matches no variation → "Sorry, this product is unavailable. Please choose a different combination." (classic via the button marker; blockified via inference).
9. Repeat step 4 on a variable product **with** a default variation → correct price on load and through checkout.

Non-variable products:
10. Simple product, both templates → buttons render on load, clicking opens the sheet with no prompt, payment completes; on blockified, stepper quantity changes refresh the amount.
11. Blockified simple product with an invalid quantity (clear the field or exceed max) → click is blocked with the options prompt.
12. If available: bookings / deposits products → behavior unchanged from develop (smoke).

Unchanged surfaces (smoke test): cart and checkout (shortcode + block) and pay-for-order → express buttons render and pay as on develop.

Legacy mode: enable WooCommerce → Settings → Products → "Redirect to the cart page after successful addition" → the blockified form re-renders as a classic-style `form.cart` POST form → repeat steps 1–4 → restore the setting.

## Follow-ups

- **PR 2 (upcoming, stacked on this one):** amount-integrity hardening - latest-wins refresh gating, `event.reject()` click contract, cart-authoritative sheet totals with reject-on-deadline. Also covers a pre-existing display gap surfaced during testing: on simple products with quantity > 1, the sheet total is quantity-aware but the line-item breakdown shows the quantity-1 items.
- Blockified-template E2E coverage: requires a block theme plus a Single Product template override - global state per `tests/e2e` rules, i.e. a dedicated Playwright project and CI matrix entry. Working spec logic exists as local probes; needs porting once the project scaffolding lands.
- STRIPE-1367 remainder: swap the interim alert for core's standardized field validation once WOOPLUG-7528 ships.


jaclync:
@coderabbitai review
@claude review

client/express-checkout/utils/variable-product-page.js
```diff
@@ -0,0 +1,206 @@
...
+	return count > 0 && chosenCount === count && ! getSelectedVariationId();
+};
+
+/**
+ * Reads the currently selected variation attributes.
+ *
+ * @return {Object} count The number of attribute fields.
+ *                  chosenCount The number of fields with a selected value.
+ *                  data Map of attribute field name to selected value.
+ */
```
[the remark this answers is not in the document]
jaclync: Good catch, fixed in 2299fc3c3.

client/express-checkout/utils/variable-product-page.js
```diff
@@ -0,0 +1,206 @@
...
+ * Watches the blockified template's quantity stepper buttons.
+ *
+ * The steppers update the input's value property only — no input event, no
+ * attribute write — so the button press is the only observable signal.
+ * Typed changes fire input events and are handled by callers directly.
+ *
+ * @param {Function} onChange Called whenever a stepper button is pressed.
+ * @return {{disconnect: Function}|null} Disconnectable handle, or null on
+ *                                       the classic template.
+ */
```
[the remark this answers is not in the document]
jaclync: Agreed, the docblock overstated it. Fixed in 2299fc3c3 - it now says null is returned whenever no blockified quantity container is present (classic template, products sold individually, non-product pages).

client/express-checkout/utils/variable-product-page.js
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Whether the page renders the classic `.variations_form` template (as
+ * opposed to the blockified Add to Cart + Options block).
+ *
+ * @return {boolean} True on the classic template.
+ */
+const isClassicTemplate = () =>
+	Boolean( document.querySelector( '.variations_form' ) );
```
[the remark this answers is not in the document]
jaclync: Legitimate concern, deferring it as a future enhacenement follow-up rather than folding it in here. Context: every selector in this lineage has queried document-wide since before this PR - on develop and in WooPayments' equivalent module alike - so pages with a second purchase form (bundle/bought-together widgets) were never a supported configuration. This PR doesn't widen that exposure, it inherits it. Scoping correctly is its own design task: the natural anchor is the form containing `#wc-stripe-express-checkout-element`, but that container isn't inside the form under every theme/hook configuration, and the change threads through all the readers plus the entrypoint across four markup states.

tests/e2e/tests/express-checkout/express-checkout.spec.js
```diff
@@ -47,19 +47,33 @@ test.describe( 'customer can use Link express checkout', () => {
...
+				timeout: 5 * 1000,
+			} );
+			await linkButton.click();
+			const dialog = await dialogPromise;
+			alertMessage = dialog.message();
+			await dialog.dismiss();
+		} ).toPass( { timeout: 45 * 1000 } );
+
+		expect( alertMessage ).toContain(
+			'select your product options before proceeding'
```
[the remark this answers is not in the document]
jaclync: Agreed on the goal, and it's the first item this PR's Follow-ups section proposes. The blockified template requires a block theme plus a Single Product template override, which is global store state; per tests/e2e conventions that means a dedicated Playwright project and CI matrix entry rather than additions to default. The spec logic already exists as working local Playwright probes (visible-before-selection, click-time prompt, selection --> sheet) and will be ported once the main work lands. Until then the blockified flow is covered by Jest fixtures for all three supported markup generations.

tests/e2e/tests/express-checkout/express-checkout.spec.js
```diff
@@ -47,19 +47,33 @@ test.describe( 'customer can use Link express checkout', () => {
...
+		let alertMessage = '';
+		await expect( async () => {
+			const dialogPromise = page.waitForEvent( 'dialog', {
+				timeout: 5 * 1000,
+			} );
+			await linkButton.click();
+			const dialog = await dialogPromise;
+			alertMessage = dialog.message();
+			await dialog.dismiss();
+		} ).toPass( { timeout: 45 * 1000 } );
```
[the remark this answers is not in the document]
jaclync: I'd prefer keeping the retry deliberately. The root cause isn't in the code or fixable by waiting: Stripe’s Express Checkout iframe silently drops clicks while it re-syncs its position after scroll-into-view, and it exposes no readiness signal to await - visibility/enabled checks pass before and during the window. If Stripe ever exposes a sync signal, both call sites can drop the pattern together.

client/entrypoints/express-checkout/index.js
```diff
@@ -898,113 +926,86 @@ jQuery( function ( $ ) {
...
+						) {
+							// Refresh stored items so the click breakdown matches the new qty.
+							wcStripeECE.refreshTotals( response );
+							wcStripeECE.updateExpressCheckoutAmount(
+								response.total.amount
+							);
+						} else {
+							wcStripeECE.reInitExpressCheckoutElement(
+								response
+							);
```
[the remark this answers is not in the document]
jaclync: Confirmed and fixed in eecd82b7f. The handler now bails on `!response || typeof response !== 'object'`, recording the error (surfaced as an alert on the next click) and keeping the current preview.

jaclync:
@coderabbitai Re https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5868#pullrequestreview-5037675591:

> Serialize product-data refreshes.

This will be addressed in the follow-up PR.

> Guard the localized global before reading wallet flags.

This seems out of the scope of the PR changes.

changelog.txt
```diff
@@ -30,6 +30,7 @@ xxxx-xx-xx - version 11.0.0
 * Fix - Hide the Apple Pay and Google Pay express buttons on pages unchecked in their own locations setting, instead of following other wallets' locations
 * Fix - Leave the Expires column blank instead of showing "N/A" on My Account → Payment methods for saved payment methods that have no expiry date, such as Link, SEPA and Cash App Pay
 * Fix - Remove the empty box shown under "Use a new payment method" on classic checkout when the save-payment-method checkbox is hidden
+* Fix - Render express checkout buttons on variable product pages before a variation is selected and prompt for product options on early clicks. Previously the buttons never appeared on the Add to Cart + Options block and appeared only after selecting a variation on the classic template
```
wjrosa: I think the entries could be much shorter

tests/e2e/tests/express-checkout/express-checkout.spec.js
```diff
@@ -47,19 +47,33 @@ test.describe( 'customer can use Link express checkout', () => {
...
-		// might cause the test to pass only because the element is not yet loaded.
-		const linkContainer = page.locator(
-			'#wc-stripe-express-checkout-element-link iframe[name^="__privateStripeFrame"]'
-		);
-		await expect( linkContainer ).toHaveCount( 1 );
-		await expect( linkContainer ).toBeHidden();
+		const linkButton = await getLinkButton( page );
+		await expect( linkButton ).toBeVisible( { timeout: 60 * 1000 } );
+
+		// Clicking before completing the attribute selection must not open
```
wjrosa: Inline comments could also be reduced

client/express-checkout/utils/variable-product-page.js
```diff
@@ -0,0 +1,208 @@
...
+		'input[name="variation_id"]'
+	);
+	if ( ! variationIdInput ) {
+		return null;
+	}
+
+	const observer = new MutationObserver( onChange );
+	observer.observe( variationIdInput, {
+		attributes: true,
+		attributeFilter: [ 'value' ],
```
wjrosa, line 175: This observer never fires on a real blockified page: `data-wp-bind--value` writes the input's `value` property only (on init and on Preact re-render), never the attribute, so nothing mutates what `attributeFilter: [ 'value' ]` watches. That leaves `onVariationChanged` dead on the blockified template: the sheet keeps the page-load amount while the cart gets the selected variation. The Jest test passes because it calls `setAttribute( 'value', ... )` directly, a write production never performs. WC dispatches maintained synthetic `change` events on the form (e.g. `setQuantity` in `add-to-cart-with-options.js`); listening for those, or polling `getSelectedVariationId()`, would be a signal that actually fires. Worth verifying in a real browser.

client/entrypoints/express-checkout/index.js
```diff
@@ -281,22 +285,16 @@ jQuery( function ( $ ) {
...
-				const addToCartButton = document.querySelector(
-					'.single_add_to_cart_button'
-				);
-
-				// First check if product can be added to cart.
-				if ( addToCartButton.classList.contains( 'disabled' ) ) {
+				// The buttons render before a variation is selected, so this
+				// guard is what prompts the shopper for their options instead
+				// of opening the wallet sheet.
+				if ( isAddToCartUnavailable() ) {
```
wjrosa, line 291: On the blockified template `isAddToCartUnavailable()` is also true for a resolved but out-of-stock variation and for an invalid quantity (both toggle the form's `is-invalid` class), while `isSelectedVariationUnavailable()` stays false, so the shopper gets "select your product options" with everything selected (or on a simple product with qty 0), and re-selecting fixes nothing. Classic shows the correct unavailable text via its marker. Consider routing the resolved-but-invalid case to the unavailable/quantity message.

client/entrypoints/express-checkout/index.js
```diff
@@ -883,6 +855,62 @@ jQuery( function ( $ ) {
...
+				// handled at click time, where the shopper is prompted to
+				// choose their options.
+				return;
+			}
+
+			wcStripeECE.blockExpressCheckoutButton();
+
+			$.when( wcStripeECE.getSelectedProductData() )
+				.then( ( response ) => {
+					if ( response.error ) {
```
wjrosa, line 893: Missing the null/primitive guard eecd82b7f added to `onQuantityChanged`: a null response throws on `response.error`, rejects into `.catch( () => hide() )`, and hides the buttons for the page view; a primitive falls into the success branch and corrupts the cached product data via `reInitExpressCheckoutElement`. Same guard applies here.

client/entrypoints/express-checkout/index.js
```diff
@@ -883,6 +855,62 @@ jQuery( function ( $ ) {
 			displayExpressCheckoutNotice( message, 'error' );
 		},
 
+		// Refresh the express checkout amount/items for the selected variation.
+		// Wired to both templates' change signals: the classic
+		// `woocommerce_variation_has_changed` jQuery event and the blockified
+		// template's `variation_id` input observer.
+		onVariationChanged: () => {
+			if ( isVariationSelectionNeeded() ) {
```
wjrosa, line 871: This early return assumes the buttons are visible, but the error paths below call `hide()`: after an errored refresh (qty over the variation's stock, stale nonce, network failure), deselecting an attribute lands here and the buttons stay hidden, breaking the visible-plus-prompt contract. A `show()` in this branch covers it.

client/entrypoints/express-checkout/index.js
```diff
@@ -883,6 +855,62 @@ jQuery( function ( $ ) {
 			displayExpressCheckoutNotice( message, 'error' );
 		},
 
+		// Refresh the express checkout amount/items for the selected variation.
+		// Wired to both templates' change signals: the classic
+		// `woocommerce_variation_has_changed` jQuery event and the blockified
+		// template's `variation_id` input observer.
+		onVariationChanged: () => {
```
wjrosa, line 870: `wcStripeECEError` is only reset in `onQuantityChanged`, so one failed quantity refresh blocks every express click with the stale error alert even after a later successful variation change updates the preview; only another quantity change clears it. Clearing it here, mirroring `onQuantityChanged`, fixes that.

jaclync:
@wjrosa Thanks for the first pass on this PR! While addressing the feedback, observing variation/quantity changes in JS for both product templates to update the preview/cache felt quite fragile as there is no official way to subscribe to these changes. Also, performance wise, if the shopper never clicks on the ECE buttons, the variation/quantity observations and preview updates felt unnecessary. Therefore, I opened a separate PR https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5886 to try a different approach by removing all variation/quantity observations and determining the ECE data from Store API cart response upon ECE button click. Please feel free to comment/correct anything on the direction first, like if the preview/cache is required for some reasons.

wjrosa:
Thank you! I agree with the direction in #5886. The preview/cache is not needed: the shopper only sees an amount inside the sheet, and the sheet now gets its amount from the cart response on click. Two things to check in #5886: with the 700ms budget, slow stores may reject the first tap very often. And the description says a late response primes the next attempt without re-adding, but the code seems to call addToCart() on every click. I think we should move forward with #5886.

jaclync:
Thanks for the feedback! Updates:

> And the description says a late response primes the next attempt without re-adding, but the code seems to call addToCart() on every click.

This was addressed in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5886 today.

However, after more extensive testing for https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5886, I still hit the timeout case once in a while in a pretty good network condition. The Store API request (even after batching the empty-cart + add-to-cart requests) could take more than 700ms to complete, and without a preview/cache like in `develop`/this PR, https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5886 currently doesn't have a way to open the payment modal. While we can throw an error and the next click will work after the cart response resolves, this doesn't feel like a good UX and would be a regression for stores where cart requests take more than 700ms.

I'm going to look into other ways to open the express checkout modal before the cart response comes back, with some approximate data to start with + refresh when the cart response is back.

jaclync:
Update after digging further into the Stripe docs and codebase - the fuller picture, and a trade-off I’d like the team’s feedback on.

### How `develop` works today

On product pages we observe variation + quantity changes (debounced) and call our own ajax endpoint (`wc_stripe_get_selected_product_data`), which recomputes a price preview in parallel to what the cart would compute (the implementation could drift from the Store API version). The observation approach is unofficial: the classic template’s jQuery events are fine while still fragile, but the blockified "Add to Cart + Options" block exposes no JS API or events, so we watch DOM internals that already changed between WC 10.7 and 10.9 and can silently break on any WC update.

On EC button click we call the Store API anyway - two sequential requests (clear cart + add-to-cart) - but only to mutate the cart for the later shipping/confirm steps: the sheet is priced from the preview in both outcomes of the 700ms race ([Stripe requires `resolve()`/`reject()` within 1s of the click](https://docs.stripe.com/js/elements_object/express_checkout_element_click_event)). A fast response's pricing data is discarded - only `items_count` is read, as a success flag - and on timeout the sheet opens before we even know the add succeeded.

### Potential issues about the current approach

- The preview is a second, weaker pricing implementation (quantity x price + client-summed add-ons + tax imitation). It could miss anything that's in Store API's cart price calculation, if it doesn't already.
- A missed/broken observation opens the sheet at a stale amount, and non-shippable products have no address step to correct it, so the shopper could be approving an incorrect amount.
- On a slow Store API cart add, failure surfaces at confirm, after the shopper filled in the whole sheet.

#### Performance notes

It spends one preview ajax per selection change (including selections never purchased) - and measured locally, the preview endpoint costs similar to a real cart call: it isn't cheaper, just earlier. If the shopper never clicks on the EC buttons, the requests were still made.

### https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5886 approach

Delete the observers + preview and price the sheet from the Store API response the click already makes - correct by construction, nothing to observe, nothing to drift, and no new requests. The cost: the same 700ms budget now gates the sheet itself. When the response misses it, the click is rejected and the retry (primed from the settled response) opens instantly - so on slow sites, it can take shoppers more than 1 tap. We can alert on timeout ("still preparing your payment - check your connection and try again"), but I'm hesitant to ship that as the primary UX: my localhost hit the timeout repeatedly on a normal connection, e.g. whenever a few due Action Scheduler tasks (feed sync, webhook deliveries) were draining on the first request after idle. Please feel free to share any thoughts on the UX.

### The dilemma

Observations have worked in production for years, and [WooPayments still uses them](https://github.com/Automattic/woocommerce-payments/blob/d0eae22708dc14761db2a97329ea5e089475a39f/client/express-checkout/shortcode-buttons-express/compatibility/wc-product-page.js) - paired with a tokenized one-off cart, so express taps never touch the shopper's real cart. From my understanding, cart-authoritative pricing is the only approach that can't show a wrong amount, but it exposes real-world latency directly in the first tap.

### Improvements regardless of the new approach

- In the EC payments on product page, use a one-off cart instead of resetting the shopper's existing cart.
- Batch the current Store API empty-cart + add-to-cart requests so that it's one request.

daledupreez:
@jaclync, I think there are a few things worth flagging independently.

#### Browser pop-up permissions and timing

This piece of the puzzle is driving large portions of the implementation. Browsers only permit popups to be opened in direct response to a user interaction (like a click/tap), where "direct response" includes a 1 second time limit. (More on the basics in the [MDN docs for `window.open()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open#description).) The Stripe Express checkout docs for the click event also document this timeout, as you noted above. I also checked and there is no way to pass in alternative line items/amounts via the `update()` API ([docs](https://docs.stripe.com/js/elements_object/update_express_checkout_element)).

As a result of all the above, our click handling is triggering the add to cart call while also adding a time limit to ensure that we can successfully launch the express checkout modal. In addition, it also means we're relying on our product details data that we generated before the click occurred (or at least that we tried to generate).

In terms of approach, I am wondering if something like the following could work:
 * Remove the custom AJAX call to generate a possibly-faulty list of line items
 * Update the UX to have the click event handling follow an algorithm more like the following:
    - Trigger the add to cart API call as we do currently, with batching when we are also removing items from the cart, as well as the 700ms time restriction
    - If the cart is returned within 700ms, then we use the cart data to drive the express checkout UI - this is definitely the best outcome for the good case
    - If the cart is not returned within 700ms, then we should do the following:
      * Reject the initial express checkout click, as we can't support it
      * Show a new/separate modal with some sort of loading state while we're waiting for the cart details to come back
         - When we get the definitive cart details back, we might be able to render a secondary CTA or express checkout button to retry the purchase with the received cart details

I think we might need to experiment with the UX, as it's not clear to me that this would be better when the API call times out, or whether we would be able to trigger the express checkout modal correctly.

jaclync:
Thanks for sharing the insights and suggestions! I experimented with the UX when the cart response times out using a modal and showing a newly generated express checkout button when the cart response is ready. It seems to work, in mobile web as well. Example screencast (from #5886):

https://github.com/user-attachments/assets/736d1833-f190-4746-8d5f-ada6a442917a

We can't programmatically open the express checkout sheet, as it only launches from a real gesture on Stripe's own button. The modal has to stay open while the wallet sheet is active (the new button's Elements state is mounted inside it, and removing it would detach the frame mid-payment). It closes when the sheet is dismissed or the payment completes.

It still requires some work to iterate on the copy, verify in different templates, and testing.

In parallel, I can combine the cart-clearing and add-to-cart calls into a single Store API batch request as an optimization.

PR #3551, opened 2024-10-28
Miscellaneous tasks on ECE

Mayisha:
Handled some miscellaneous tasks in this PR.

## Changes and testing instructions
- Passed `locale` to ECE when initializing ([c2e3a6f](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3551/commits/c2e3a6fbc1ff28c77f40bf153f61aecb4f056811))
- Replaced instances of `source` with `payment method` on the edit subscription page ([c6f730b](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3551/commits/c6f730b040ede139c3146701ab8536b42af115c8))
    - Go to the subscription edit page and click the pencil icon in the billing address section.
    - Before the payment method field used to be `Stripe Sources ID`. Now it is changed to `Stripe Payment Method ID`

Screenshot 2024-10-30 at 3 29 00 PM

- Not calling the ECE specific `paymentFailed` function for order creation failures ([071cd30](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3551/commits/071cd3083370a2bad032cfa4ac1c65e4e5819130))
    - Fixed the `abortPayment` function. When `abortPayment` is called due to order related failure (not payment failure with ECE), the `paymentFailed( { reason: 'fail' } )` call throws an error.

Screenshot 2024-10-28 at 4 14 57 PM

- Fixed issue in purchasing subscriptions when the store has no shipping options ([148b75b](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/3551/commits/148b75b1e00bdbbcefeba2e76df6397b460e39cb))

    - Remove all the shipping methods from your store.
    - Create a simple subscription product.
    - As a shopper, go the the simple subscription product page.
    - Try to pay with Google Pay.
    - In `develop`, you should see the above error, but in this branch payment should work fine.

Screenshot 2024-10-28 at 11 37 12 AM

PR #5869, opened 2026-08-27
Fix Express Checkout silent drop after wallet sheet

### LINKED ISSUE #5845: 10.9.0: Express Checkout (Apple Pay) sheet approved but no PaymentIntent created — customer re-prompted, order left pending until auto-cancelled
### Description

Since updating to 10.9.0 (from 10.8.x, on 2026-08-19), a meaningful share of Apple Pay / express checkout attempts fail silently after the customer approves the payment sheet. The approved wallet authorization never reaches Stripe — no PaymentIntent is created at all — the page re-prompts the customer to pay again, and Apple Pay declines or the customer abandons the second sheet. The order remains pending with no payment activity and is later cancelled by WooCommerce’s unpaid-order timer.
A customer described it directly: “I clicked the first Apple Pay link, approved it, then the site asked for it again. Apple doesn’t like that.”

Evidence from our store (digital goods, ~15 orders/day, US, Stripe):
	•	Timeline pivots exactly on the 10.9.0 install (2026-08-19). Before: 0–2 failed orders/day, zero cancellations. After: failed orders stop entirely and unpaid-pending → cancelled orders appear at ~2/day. (We understand the failed→cancelled relabel is intentional per the 10.9.0 changelog; the regression is the express-checkout drop, not the relabel.)
	•	For all 9 affected orders (44509, 44511, 44577, 44593, 44612, 44617, 44626, 44628, 44633): zero PaymentIntents in Stripe — not failed intents, none. The authorization is lost client-side before intent creation.
	•	Wallet success rates from Stripe charge data: apple_pay ~3.0/day before → ~1.2/day after; google_pay 0.8 → 0.5; while plain-card volume roughly doubled (customers retrying with card after the wallet flow fails). Total volume flat — this is substitution plus abandonment, not a traffic change.
	•	One customer attempted the same product twice, 12 minutes apart; both attempts left pending orders with no Stripe activity.
	•	Failure is intermittent — some Apple Pay payments still succeed daily — which points at the 10.9.0 express-checkout changes (“Render the Express Checkout button on the cart and checkout from page-bootstrapped data, removing a cart-details request”; “Memoize the Blocks Express Checkout button…”) leaving the button with stale state in some sequence of page/cart updates.

Environment: WooCommerce 11.0.1, Stripe gateway 10.9.0, WordPress with Kadence child theme, both shortcode and block (Store API) checkout paths affected (created_via shows both), express checkout buttons enabled on product/cart/checkout, US store, USD.
Expected behavior: An approved Apple Pay sheet results in a confirmed PaymentIntent, or at minimum a failed intent with an error surfaced — never a silent drop and re-prompt.

diegocurbelo:
Fixes STRIPE-1406
Fixes #5845

For order-side errors (`isOrderError` opt-out from #3551) `abortPayment()` skipped `event.paymentFailed()`, so the approved Apple Pay / Google Pay sheet never got a _terminal_ result and stayed open. 

And `displayExpressCheckoutNotice()` silently no-ops on pages with no notices wrapper (which is common on single-product templates). No PaymentIntent was created and nothing reached the gateway log, so the attempt was invisible to the shopper and to support.

This PR changes `abortPayment()` to always fail the sheet, after putting the message on screen; aborts always carry a message; and the notice falls back to the express checkout container.

Separately, `Legacy::process_legacy_payment()` (priority 999) returns without setting a status when it can’t resolve the payment method to an available gateway — after WooCommerce has already moved the order to pending — so the Store API answers 200 with an empty payment status and nothing logged anywhere. `WC_Stripe_Blocks_Support::fail_unprocessed_payment()` now catches that at priority 10000, logs it, and throws.

This makes the failure visible and diagnosable; it does not change why a store’s gateway becomes unresolvable, which needs a store-specific reproduction the new log line should now supply.

**_Backward compatibility impact_**:
`abortPayment()`’s third argument is gone, but it is module-private in both entrypoints and not exposed on window. The new PHP method and JS export are additive; no hook, filter, option key or public signature changed. The new guard is scoped to payment method ids this gateway actually registers, so it can’t claim another vendor’s stripe_-prefixed gateway, and it only fires when the payment result is still empty — which only happens when the gateway was never invoked.

## Testing instructions

1. Add this to a mu-plugin to make the order fail after the wallet sheet is approved:
    ```php
    add_action( 'woocommerce_store_api_checkout_update_order_from_request', function () {
        throw new \Automattic\WooCommerce\StoreApi\Exceptions\RouteException(
            'test_failure', 'Simulated order failure', 400
        );
    } );
    ```
2. On a product page, click the Apple Pay / Google Pay button and approve the sheet.
    - in develop: the sheet does not close and no error appears anywhere on the page.
    - in this PR: the wallet sheet now closes with a failure
3. Repeat on the cart page and on classic checkout:
    - the message renders in the store’s normal notices area, as it did before.
4. Repeat on a Blocks checkout:
    - the message renders in the Blocks error area.
5. Remove the snippet and complete a normal Apple Pay / Google Pay purchase from the product, cart and checkout pages.
6. Verify they all still succeed and land on the order-received page.
7. With Optimized Checkout enabled, confirm express checkout still completes and card payments are unaffected.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/blocks/express-checkout/hooks.js
+++ b/client/blocks/express-checkout/hooks.js
@@ -56,15 +56,16 @@ export const useExpressCheckout = ( {
 
 	const abortPayment = useCallback(
 		( onConfirmEvent, message ) => {
-			// The wallet sheet only closes once the confirm event gets a
-			// terminal result, so this must run for order errors too.
-			onConfirmEvent.paymentFailed( { reason: 'fail' } );
-
 			// If we have a multiline message using newlines, replace them with <br>.
 			const formattedMessage = message.replace( /\n/g, '<br>' );
 			setExpressPaymentError( formattedMessage );
 
 			onAbortPaymentHandler( onConfirmEvent, message );
+
+			// The wallet sheet only closes once the confirm event gets a
+			// terminal result, so this must run for order errors too. Last,
+			// so a rejected late call can't cost the shopper the message.
+			onConfirmEvent.paymentFailed( { reason: 'fail' } );
 		},
 		[ setExpressPaymentError ]
 	);
--- a/client/blocks/express-checkout/hooks.js
+++ b/client/blocks/express-checkout/hooks.js
@@ -62,9 +62,9 @@ export const useExpressCheckout = ( {
 
 			onAbortPaymentHandler( onConfirmEvent, message );
 
-			// The wallet sheet only closes once the confirm event gets a
-			// terminal result, so this must run for order errors too. Last,
-			// so a rejected late call can't cost the shopper the message.
+			// The wallet sheet only closes once the confirm event gets a terminal
+			// result, so order errors must fail it too. A late call rejects an
+			// internal Stripe promise asynchronously, after the message is shown.
 			onConfirmEvent.paymentFailed( { reason: 'fail' } );
 		},
 		[ setExpressPaymentError ]
--- a/client/blocks/express-checkout/hooks.js
+++ b/client/blocks/express-checkout/hooks.js
@@ -55,10 +55,10 @@ export const useExpressCheckout = ( {
 	}, [] );
 
 	const abortPayment = useCallback(
-		( onConfirmEvent, message, isOrderError = false ) => {
-			if ( ! isOrderError ) {
-				onConfirmEvent.paymentFailed( { reason: 'fail' } );
-			}
+		( onConfirmEvent, message ) => {
+			// The wallet sheet only closes once the confirm event gets a
+			// terminal result, so this must run for order errors too.
+			onConfirmEvent.paymentFailed( { reason: 'fail' } );
 
 			// If we have a multiline message using newlines, replace them with <br>.
 			const formattedMessage = message.replace( /\n/g, '<br>' );
--- a/client/express-checkout/utils/index.js
+++ b/client/express-checkout/utils/index.js
@@ -39,15 +39,20 @@ export const getErrorMessageFromNotice = ( notice ) => {
  * non-success payment status with nothing attached, or a Stripe error whose `message`
  * is unset — and aborting with an empty one leaves the shopper with no explanation.
  *
+ * Takes the message as-is: callers that receive a WooCommerce notice run it through
+ * `getErrorMessageFromNotice()` first, and parsing a plain Stripe message as HTML would
+ * truncate it at anything tag-like (e.g. an email address in angle brackets).
+ *
  * @param {string|undefined} message The message reported by Stripe or the server, if any.
  * @return {string} The message to display.
  */
 export const getExpressCheckoutErrorMessage = ( message ) =>
-	getErrorMessageFromNotice( message ) ||
-	__(
-		'There was a problem processing the order.',
-		'woocommerce-gateway-stripe'
-	);
+	message?.trim()
+		? message
+		: __(
+				'There was a problem processing the order.',
+				'woocommerce-gateway-stripe'
+		  );
 
 /**
  * Retrieves express checkout data from global variable.
--- a/client/express-checkout/utils/index.js
+++ b/client/express-checkout/utils/index.js
@@ -1,5 +1,6 @@
 /* global wc_stripe_express_checkout_params */
 import jQuery from 'jquery';
+import { __ } from '@wordpress/i18n';
 import { isAmazonPayEnabled, isLinkEnabled } from 'wcstripe/stripe-utils';
 import { EXPRESS_CHECKOUT_NOTICE_DELAY } from 'wcstripe/data/constants';
 import {
@@ -30,6 +31,23 @@ export const getErrorMessageFromNotice = ( notice ) => {
 	div.innerHTML = notice.trim();
 	return div.firstChild?.textContent || '';
 };
+
+/**
+ * Resolves the message shown when an express checkout payment is aborted.
+ *
+ * Stripe errors and Store API responses can both arrive without a usable message — a
+ * non-success payment status with nothing attached, or a Stripe error whose `message`
+ * is unset — and aborting with an empty one leaves the shopper with no explanation.
+ *
+ * @param {string|undefined} message The message reported by Stripe or the server, if any.
+ * @return {string} The message to display.
+ */
+export const getExpressCheckoutErrorMessage = ( message ) =>
+	getErrorMessageFromNotice( message ) ||
+	__(
+		'There was a problem processing the order.',
+		'woocommerce-gateway-stripe'
+	);
 
 /**
  * Retrieves express checkout data from global variable.
@@ -399,22 +417,23 @@ export const displayExpressCheckoutNotice = (
 		`<div class="${ classNames.join( ' ' ) }" role="note" />`
 	).html( safeMessage );
 
-	if ( $container.length ) {
-		if ( isBlockCheckout ) {
-			$container.prepend( note );
-		} else {
-			$container.append( note );
-		}
-	} else {
+	if ( ! $container.length ) {
 		// Product pages, and themes that render no notices wrapper, would drop the
-		// message entirely. Fall back to the button, which is on every page that
-		// can raise one of these errors.
+		// message entirely. Fall back to the classic express checkout container; it
+		// sits where the shopper just tapped, so no scrolling is needed (and none is
+		// wanted, with a wallet sheet opening over the page).
 		const $button = jQuery( '#wc-stripe-express-checkout-element' );
-		if ( ! $button.length ) {
-			return;
+		if ( $button.length ) {
+			$button.before( note );
 		}
 
-		$button.before( note );
+		return;
+	}
+
+	if ( isBlockCheckout ) {
+		$container.prepend( note );
+	} else {
+		$container.append( note );
 	}
 
 	// Scroll to the notice.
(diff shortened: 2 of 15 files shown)


includes/class-wc-stripe-blocks-support.php
```diff
@@ -584,6 +585,72 @@ public function add_stripe_intents( PaymentContext $context, PaymentResult &$res
...
+
+		WC_Stripe_Logger::error(
+			'Payment was never processed for order: ' . $context->order->get_id(),
+			[
+				'payment_method' => $payment_method,
+				'reason'         => 'WooCommerce could not resolve the payment method to an available gateway.',
+			]
+		);
+
+		throw new Exception( __( 'This payment method is not available right now. Please try again or use a different one.', 'woocommerce-gateway-stripe' ) );
```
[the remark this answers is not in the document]
diegocurbelo: Confirmed — the only empty-status path is the one this targets.

`Legacy::process_legacy_payment()` sets a status unconditionally once the gateway runs: `$gateway_result['result'] ?? 'failure'`, coerced into `['success','failure','pending','error']`. Redirect-based UPE `returns ['result' => 'success', 'redirect' => …]`, so iDEAL/Bancontact are covered.

Skipping the suggested test: it would assert WooCommerce’s invariant rather than ours. The existing data provider already covers the inverse — we don’t throw when a status is set.

Updated the docblock and the log reason in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5869/commits/8529bc77d76cebdc544b069ce62ae9e84368b8b8, `get_request_payment_method()` already rejects an unavailable gateway before this hook fires, so reaching an empty status is only possible if gateway availability changes mid-request.

> Drafted with Claude Code; reviewed by me

includes/class-wc-stripe-blocks-support.php
```diff
@@ -584,6 +585,72 @@ public function add_stripe_intents( PaymentContext $context, PaymentResult &$res
...
+		if ( 0 !== strpos( $payment_method, $prefix ) ) {
+			return false;
+		}
+
+		$main_gateway = WC_Stripe::get_instance()->get_main_stripe_gateway();
+		if ( ! $main_gateway instanceof WC_Stripe_UPE_Payment_Gateway ) {
+			return false;
+		}
+
+		return isset( $main_gateway->payment_methods[ substr( $payment_method, strlen( $prefix ) ) ] );
```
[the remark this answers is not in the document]
diegocurbelo: Done in ac5a2da

client/blocks/express-checkout/hooks.js
```diff
@@ -55,16 +55,17 @@ export const useExpressCheckout = ( {
...
 			// If we have a multiline message using newlines, replace them with <br>.
 			const formattedMessage = message.replace( /\n/g, '<br>' );
 			setExpressPaymentError( formattedMessage );
 
 			onAbortPaymentHandler( onConfirmEvent, message );
+
+			// The wallet sheet only closes once the confirm event gets a
+			// terminal result, so this must run for order errors too. Last,
+			// so a rejected late call can't cost the shopper the message.
+			onConfirmEvent.paymentFailed( { reason: 'fail' } );
```
wjrosa, line 68: I think `paymentFailed()` can throw if the confirm event is already terminal (for example, the shopper dismissed the sheet while the request was still running). If that happens, the catch in `payment-flow.js` calls `abortPayment` again and the shopper sees Stripe's IntegrationError text instead of the real error. 

Maybe we should wrap this call in a try/catch, here and in the classic `abortPayment` (`entrypoints/express-checkout/index.js`)?

client/express-checkout/utils/index.js
```diff
@@ -31,6 +32,23 @@ export const getErrorMessageFromNotice = ( notice ) => {
...
+ * Resolves the message shown when an express checkout payment is aborted.
+ *
+ * Stripe errors and Store API responses can both arrive without a usable message — a
+ * non-success payment status with nothing attached, or a Stripe error whose `message`
+ * is unset — and aborting with an empty one leaves the shopper with no explanation.
+ *
+ * @param {string|undefined} message The message reported by Stripe or the server, if any.
+ * @return {string} The message to display.
+ */
+export const getExpressCheckoutErrorMessage = ( message ) =>
```
wjrosa, line 49: I think `getErrorMessageFromNotice` was made for WooCommerce notices (HTML), but here it also receives raw Stripe messages. Since it parses the string as HTML and keeps only `div.firstChild?.textContent`, a message containing something that looks like a tag (for example `Invalid email <user@example.com>`) gets cut at that point, and some can even collapse to the generic fallback. Maybe we should only parse server notices and pass Stripe's `error.message` through as plain text?

client/blocks/express-checkout/hooks.js
```diff
@@ -55,16 +55,17 @@ export const useExpressCheckout = ( {
...
 
 	const abortPayment = useCallback(
-		( onConfirmEvent, message, isOrderError = false ) => {
-			if ( ! isOrderError ) {
-				onConfirmEvent.paymentFailed( { reason: 'fail' } );
-			}
-
+		( onConfirmEvent, message ) => {
 			// If we have a multiline message using newlines, replace them with <br>.
 			const formattedMessage = message.replace( /\n/g, '<br>' );
```
wjrosa, line 60: Two small things here: this path injects the message with only the `\n` to `<br>` replacement, while the classic path escapes it first (`jQuery( '<div>' ).text( message ).html()` in `displayExpressCheckoutNotice`), so markup in a server/Stripe message renders unescaped only on Blocks. Also, the show-message + `paymentFailed()` sequence and its comment are now duplicated in both entrypoints; could we move it into the shared handler so the invariant lives in one place?

includes/class-wc-stripe-blocks-support.php
```diff
@@ -584,6 +574,75 @@ public function add_stripe_intents( PaymentContext $context, PaymentResult &$res
...
+	 *
+	 * @param PaymentContext $context Holds context for the payment.
+	 * @param PaymentResult  $result  Result object for the payment.
+	 *
+	 * @throws Exception When the payment was never processed.
+	 *
+	 * @return void
+	 */
+	public function fail_unprocessed_payment( PaymentContext $context, PaymentResult &$result ) {
+		if ( '' !== (string) $result->status ) {
```
wjrosa, line 623: `PaymentResult` has a public `get_status()` getter, so I think we can use it here and drop the new `phpstan-baseline.neon` entry for the protected `$status` read (maybe with a `method_exists` guard if the oldest supported WC does not have it).

diegocurbelo: `PaymentResult::get_status()` does exist (WC 10.5+, we require 10.8), but I’d rather not use it here.

The pinned `php-stubs/woocommerce-stubs` is v10.4.2, which predates the getter, so PHPStan reports `method.notFound` — the baseline entry gets swapped for an inline ignore rather than removed.

It’d also be the only getter call in the file. The other 14 reads of `PaymentContext/PaymentResult` there — `payment_method` ×5, `payment_details` ×4, `order` ×3, `redirect_url`, `payment_data` — all go through the magic `__get` and are baselined the same way, because no getters exist for them.

So it’s the same suppression count, in two different forms, for one read out of fifteen.

client/blocks/express-checkout/hooks.js
diegocurbelo: A try/catch wouldn’t catch it. In the shipped `Stripe.js` bundle the ECE `paymentFailed()` throws inside a detached `.then()`:
```js
m.action.getIsAwaitingConfirmCallback({ groupId: _ }).then(function (t) {
  if (!t) throw new I.N("Unexpected call to paymentFailed(). …");
})
```

No `return`, no `.catch()`. The call returns `undefined` synchronously and the `IntegrationError` surfaces as an unhandled rejection, so it never reaches the catch in `payment-flow.js` —no second `abortPayment`, and the message isn’t replaced.

I updated the comment in [72d3cae](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5869/commits/72d3cae97b79bbd48a69d502b03b03355f45f654)

client/express-checkout/utils/index.js
diegocurbelo: Good catch, fixed in [e9d8371](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5869/commits/e9d83711fce4519345765b90e547f264d6eb8abb)

client/blocks/express-checkout/hooks.js
diegocurbelo: Checked, and Blocks sanitizes it: `setExpressPaymentError` → `createErrorNotice` → `StoreNoticesContainer` → `<RawHTML>{ sanitizeHTML( decodeEntities( notice.content ) ) }</RawHTML>`, where `sanitizeHTML` is `DOMPurify` allowlisting `a`, `b`, `em`, `i`, `strong`, `p`, `br`, `abbr`. So `<br>` works and anything dangerous is stripped.

The `\n` → `<br>` line is also byte-for-byte unchanged from develop — this PR only moved the statements around it.

Only the paymentFailed() call is actually common — showing the message genuinely differs (`setExpressPaymentError` vs `displayExpressCheckoutNotice`), so a shared helper would wrap a single method call.

wjrosa (approved):
Code is good and works as expected 🎉

Issue #634, opened 2018-05-05
Allow use with a restricted Stripe API key


Calrion:
#### What I expected

After configuring the WooCommerce Stripe gateway with a Stripe restricted API key, I attempted to checkout on the corresponding website using the Stripe payment method and a Stripe test card number.

I expected the checkout process to succeed and for a new test transaction to appear in my Stripe account.

I also expected WooCommerce to require only a restricted key with read and write access to charges and customers, and perhaps events; and maybe read-only access to Apple Pay domains.

My goal here, as a Stripe customer who integrates multiple applications, is to ensure that I can _always_: ensure apps have only the access they need and not a bean more; and that I can revoke access to a compromised or misbehaving app without affecting the others. The combination of requiring an unrestricted API key and Stripe's restriction of one unrestricted key per account confounds this.

On a related note, I expected all Stripe API actions to be available to restricted API keys (providing they were configured with sufficient permissions); I have sent Stripe feedback to this effect.

#### What happened instead

I received the following error message on the `/checkout` page, after clicking the _Place order_ button (line endings added for readability):

````
The provided key 'rk_test_************************' does not have the required 
permissions for this endpoint on account 'acct_****************'. This is a restricted 
API key, but the required permissions are not available for use by restricted keys.
````

I would first like to know what actions are being performed that can't be done using a restricted API key, and secondly would like to suggest enhancing the Stripe gateway so that it _can_ use a Stripe restricted API key.

#### Steps to reproduce the issue

While in **test mode** I created a Stripe restricted API key in my account and (for the purposes of this ticket) gave it all the available permissions. Specifically, I clicked the "read and write" option next to the major categories of "All core resources", "All connect resources", "All relay resources", and "All subscription resources".

I provided my test publishable key and the restricted secret key I created above to WooCommerce. WooCommerce had Stripe test mode enabled.

I'm running WooCommerce 3.3.5 on WordPress 4.9.5 with the Stripe gateway plugin 4.1.2.

-------------------

- [ ] Issue assigned to next milestone.
- [x] Issue assigned a priority (will be assessed by maintainers).

roykho:
Closing this for now as it seems to be account related. @thorsten-stripe do you know anything about this? If it is something we need to adjust in the plugin let me know and I will re-open.

Calrion:
@thorsten-stripe I'm not prepared to disclose that detail here, but I contacted Stripe and asked they pass the information to you. Thanks for looking-in to this.

[...]

PR #5872, opened 2026-08-27
Support defining Stripe secret keys as wp-config.php constants

wjrosa:
Towards STRIPE-3
Towards #634

Merchants who want to run the plugin on a [Stripe Restricted API Key](https://docs.stripe.com/keys#create-restricted-api-secret-key) currently have no supported way to keep that key out of the database: there is no UI field for entering keys (the settings screen is OAuth-only), and a key written directly to the `woocommerce_stripe_settings` option is still a database-stored credential, which defeats the main reason for using a restricted key.

This PR adds first-class support for supplying the secret keys from `wp-config.php`:

```php
define( 'WC_STRIPE_SECRET_KEY', 'rk_live_...' );      // or sk_live_...
define( 'WC_STRIPE_TEST_SECRET_KEY', 'rk_test_...' ); // or sk_test_...
```

A new `WC_Stripe_Key_Constants` class hooks `option_woocommerce_stripe_settings` and `pre_update_option_woocommerce_stripe_settings`:

- **On read**, a defined constant replaces the stored `secret_key` / `test_secret_key`. The filters are used (rather than the plugin's own settings accessors) because WooCommerce core also reads and writes this option directly via `WC_Settings_API::init_settings()` / `process_admin_options()`, and every reader must see the same key. This also means the override survives the Stripe App OAuth refresh job, which rewrites the stored key on `app`-type connections.
- **On write**, a value that round-tripped from the constant is reverted to the stored value, so a settings save never persists the constant into the database. A deliberately changed key (e.g. via the account keys REST endpoint) is written as-is; the constant simply keeps winning at read time while defined.
- Blank or non-string constants are ignored rather than applied, so a typo cannot silently disconnect the store.

**Backward compatibility:** purely additive. No existing symbol, hook, or option key changes. Stores that do not define the constants are unaffected: with no constants defined, both filters return their input unchanged. The stored keys are never modified by this feature (reverting on write restores what was already stored).

## Testing instructions

1. Connect a store to Stripe as usual (OAuth or keys via the REST endpoint).
2. In your Stripe dashboard, create a restricted key (or use any other valid secret key for the account).
3. Add to `wp-config.php`, above the "That's all, stop editing!" line: `define( 'WC_STRIPE_TEST_SECRET_KEY', 'rk_test_...' );`
4. Load the Stripe settings page and the checkout page: the account still shows as connected and payment methods render. Place a test order; in the Stripe dashboard request logs, the requests are made with the restricted key.
5. Save any Stripe setting (e.g. toggle test mode off/on), then check the stored option: `wp option get woocommerce_stripe_settings --format=json | jq '.test_secret_key'`: it still holds the originally stored key, not the constant.
6. Remove the constant from `wp-config.php`: the store immediately falls back to the stored key.

---

### Changelog entry

<details>

#### Comment

</details>

**Post merge**

cc @aheckler

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -30,7 +30,7 @@ xxxx-xx-xx - version 11.0.0
 * Fix - Hide the Apple Pay and Google Pay express buttons on pages unchecked in their own locations setting, instead of following other wallets' locations
 * Fix - Leave the Expires column blank instead of showing "N/A" on My Account → Payment methods for saved payment methods that have no expiry date, such as Link, SEPA and Cash App Pay
 * Fix - Remove the empty box shown under "Use a new payment method" on classic checkout when the save-payment-method checkbox is hidden
-* Add - Support defining the Stripe secret keys as constants in wp-config.php (WC_STRIPE_SECRET_KEY, WC_STRIPE_TEST_SECRET_KEY), so a restricted API key can be used without storing it in the database
+* Add - Allow defining the Stripe secret keys as wp-config.php constants, keeping them out of the database
 
 2026-08-17 - version 10.9.0
 * Fix - Re-enable Adaptive Pricing if it was automatically disabled after a Checkout Session amount mismatch
--- a/changelog.txt
+++ b/changelog.txt
@@ -30,6 +30,7 @@ xxxx-xx-xx - version 11.0.0
 * Fix - Hide the Apple Pay and Google Pay express buttons on pages unchecked in their own locations setting, instead of following other wallets' locations
 * Fix - Leave the Expires column blank instead of showing "N/A" on My Account → Payment methods for saved payment methods that have no expiry date, such as Link, SEPA and Cash App Pay
 * Fix - Remove the empty box shown under "Use a new payment method" on classic checkout when the save-payment-method checkbox is hidden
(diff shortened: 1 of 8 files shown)


aheckler:
> Remove the constant from wp-config.php: the store immediately falls back to the stored key.

I'm confused. 

If we're not documenting how to remove the full access OAuth keys from the db (or automating their removal) as a part of these 4 PRs, then I'm not sure I understand the point of using RAKs to begin with?

The root problem, as I see it, is not "the plugin doesn't support RAKs." The problem is "My site got hacked and now someone has my full access Stripe keys." So any solution that allows me to use RAKs but leaves the original keys sitting in the db seems like it's missing the point.

The second option you included in #5875 (the account keys REST endpoint) seems better, since it actually does swap out the full access keys in favor of RAKs. IMHO it could benefit from a UI, but it's better than what we have today I suppose.

wjrosa:
That's a good point. Rolling the key in the Stripe dashboard is the only way to really revoke a stolen key. Removing the DB copies protects against the next leak, and that is where the constant helps: with the endpoint swap alone, the RAK is still stored in the DB.

The idea is to combine both: swap the stored key for the RAK via the endpoint (#5874 makes that stick on app connections), optionally define the constant at the top, then roll the old key in Stripe.

The doc was indeed vague about this. I added a "Decommission the standard key" section to #5875 with those steps. A more complete cleanup guide is planned separately. Agreed, a UI would help too; out of scope for these PRs, but worth a follow-up issue.

PR #5880, opened 2026-08-31
Prevent Classic Checkout submitting: followup to show an error instead of submitting when required fields are empty

diegocurbelo:
Follow-up to merged PR #5853 (STRIPE-1402).

#5853 stops the required-field gate misreading fields that are filled. This PR handles genuinely empty fields and related submission failures.

WooCommerce submits Classic Checkout unless a `checkout_place_order_*` handler returns exactly `false`. The gate returned `undefined`, so WooCommerce submitted the form without the payment method that Stripe withheld. This created an order and then marked it failed.

This PR:

- Returns `false` and shows "Please fill in all required fields." when a visible required field is empty.
- Resets the one-shot checkout state when that validation blocks a programmatic retry.
- Recognizes required textareas rendered by `woocommerce_form_field()`.
- Blocks submission with an error when the selected Stripe component is unavailable.
- Falls back to rendering errors in `form.checkout` when a custom template omits `.woocommerce-notices-wrapper`.
- Applies required-field validation before saved-token handling, while still leaving payment processing for a valid saved-token submission to WooCommerce.

The new-card path now shows one generic required-fields message instead of WooCommerce's field-specific server messages. The affected fields are still marked through the existing checkout validation events. The DOM does not identify which fields WooCommerce knows how to validate server-side, so narrowing the gate to third-party fields would be unreliable.

## Testing instructions

1. Configure Stripe with a test account and use Classic Checkout.
2. Add a product to the cart, fill the billing fields and card details, then add the checkbox group using the snippet from #5853.
3. Leave both injected checkboxes unchecked and place the order.
   - No checkout request or order should be created.
   - "Please fill in all required fields." should appear.
   - The shopper should remain on the checkout page.
4. Check either injected checkbox and place the order again. The order should complete.
5. Start another checkout, run the #5853 snippet again, then replace the injected field with a required textarea:

   ```js
   const field = document.querySelector( '#stripe-1402-reproducer' );
   field.innerHTML = `
       <label>Required details <abbr class="required" title="required">*</abbr></label>
       <textarea class="input-text" name="stripe_1402_details"></textarea>
   `;
   ```

6. Place the order with the textarea empty. The same required-fields error should appear and no order should be created.
7. Fill the textarea and retry. The order should complete.
8. Repeat the empty-textarea check after running:

   ```js
   document.querySelector( '.woocommerce-notices-wrapper' )?.remove();
   ```

   The error should appear at the top of the checkout form.
9. On a new checkout, inject the same client-only required field and leave it empty, then pay with a saved card. The required-fields error should appear and no order should be created.
10. Fill the injected field and retry with the saved card. The order should complete without creating a new Stripe payment method in the browser.
11. On a new checkout with a new card, leave a native billing field empty. Confirm that the generic required-fields message appears and the field is marked invalid.

Covered by Jest tests for handler return values, retry state, textarea handling, missing components, and notice fallback. An E2E test verifies that an empty third-party field causes no checkout request and that filling it allows the order to complete.

---

### Changelog entry

<details>

#### Comment

Merged PR #5853 already contains the STRIPE-1402 changelog entry.

</details>

**Post merge**

🤖 Generated with AI assistance.

### DIFF
--- a/client/classic/upe/deferred-intent.js
+++ b/client/classic/upe/deferred-intent.js
@@ -7,6 +7,7 @@ import {
 	isPaymentMethodRestrictedToLocation,
 	isUsingSavedPaymentMethod,
 	paymentMethodSupportsDeferredIntent,
+	showErrorCheckout,
 	togglePaymentMethodForCountry,
 } from '../../stripe-utils';
 import './style.scss';
@@ -23,6 +24,7 @@ import {
 	processPayment,
 	trackMountInProgress,
 } from './payment-processing';
+import { __ } from '@wordpress/i18n';
 
 jQuery( function ( $ ) {
 	const stripeServerData = getStripeServerData();
@@ -102,27 +104,37 @@ jQuery( function ( $ ) {
 
 	function processPaymentIfNotUsingSavedMethod( $form ) {
 		const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
-		if ( ! isUsingSavedPaymentMethod( paymentMethodType ) ) {
-			return processPayment( api, $form, paymentMethodType );
+
+		// A saved token travels in the POST and needs no payment method, so the
+		// required-field state is none of our business on that path.
+		if ( isUsingSavedPaymentMethod( paymentMethodType ) ) {
+			return;
 		}
-	}
 
-	$( 'form.checkout' ).on( generateCheckoutEventNames(), function () {
-		const $form = $( this );
-
-		// Don't create a Stripe payment method if required checkout fields are empty.
-		// This prevents unnecessary Stripe API calls before WC's server-side validation.
-		// jQuery :visible filters out fields hidden by conditional checkout logic
-		// (e.g. shipping fields when "Ship to different address" is unchecked).
+		// Returning false is what stops WooCommerce submitting: returning early without
+		// it would let the form POST with no payment method, and the order would be
+		// created and then failed. jQuery :visible filters out fields hidden by
+		// conditional checkout logic (e.g. shipping fields when "Ship to different
+		// address" is unchecked).
 		if (
 			hasEmptyRequiredFields(
 				$form.find( '.validate-required:visible' ).toArray()
 			)
 		) {
-			return;
+			showErrorCheckout(
+				__(
+					'Please fill in all required fields.',
+					'woocommerce-gateway-stripe'
+				)
+			);
+			return false;
 		}
 
-		return processPaymentIfNotUsingSavedMethod( $form );
+		return processPayment( api, $form, paymentMethodType );
+	}
+
+	$( 'form.checkout' ).on( generateCheckoutEventNames(), function () {
+		return processPaymentIfNotUsingSavedMethod( $( this ) );
 	} );
 
 	// Mount the Stripe Payment Elements onto the Add Payment Method page and Pay for Order page.
--- a/client/classic/upe/deferred-intent.js
+++ b/client/classic/upe/deferred-intent.js
@@ -22,6 +22,7 @@ import {
 	maybeUpdateOptimizedCheckoutExclusions,
 	mountStripePaymentElement,
 	processPayment,
+	resetCheckoutCompletionState,
 	trackMountInProgress,
 } from './payment-processing';
 import { __ } from '@wordpress/i18n';
@@ -103,14 +104,6 @@ jQuery( function ( $ ) {
 	} );
 
 	function processPaymentIfNotUsingSavedMethod( $form ) {
-		const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
-
-		// A saved token travels in the POST and needs no payment method, so the
-		// required-field state is none of our business on that path.
-		if ( isUsingSavedPaymentMethod( paymentMethodType ) ) {
-			return;
-		}
-
 		// Returning false is what stops WooCommerce submitting: returning early without
 		// it would let the form POST with no payment method, and the order would be
 		// created and then failed. jQuery :visible filters out fields hidden by
@@ -121,6 +114,7 @@ jQuery( function ( $ ) {
 				$form.find( '.validate-required:visible' ).toArray()
 			)
 		) {
+			resetCheckoutCompletionState();
 			showErrorCheckout(
 				__(
 					'Please fill in all required fields.',
@@ -130,7 +124,10 @@ jQuery( function ( $ ) {
 			return false;
 		}
 
-		return processPayment( api, $form, paymentMethodType );
+		const paymentMethodType = getSelectedUPEGatewayPaymentMethod();
+		if ( ! isUsingSavedPaymentMethod( paymentMethodType ) ) {
+			return processPayment( api, $form, paymentMethodType );
+		}
 	}
 
 	$( 'form.checkout' ).on( generateCheckoutEventNames(), function () {
--- a/client/classic/upe/payment-processing.js
+++ b/client/classic/upe/payment-processing.js
@@ -65,6 +65,14 @@ const gatewayUPEComponents = {};
 let hasCheckoutCompleted = false;
 
 /**
+ * Clears the one-shot completion flag when a later checkout check blocks the
+ * programmatic submit before processPayment can consume it.
+ */
+export function resetCheckoutCompletionState() {
+	hasCheckoutCompleted = false;
+}
+
+/**
  * OC exclusions last applied to each Elements instance (as a sorted key), so
  * redundant `elements.update()` calls can be skipped. Keyed by instance: a
  * re-mounted element starts fresh and stale state can't leak across mounts.
@@ -1163,17 +1171,19 @@ export const processPayment = (
 		return;
 	}
 
+	const genericErrorMessage = __(
+		'Payment failed. Please try again.',
+		'woocommerce-gateway-stripe'
+	);
+
 	if ( ! gatewayUPEComponents[ paymentMethodType ] ) {
-		return;
+		showErrorCheckout( genericErrorMessage );
+		return false;
 	}
 
 	blockUI( jQueryForm );
 
 	const getErrorMessage = ( err ) => {
-		const genericErrorMessage = __(
-			'Payment failed. Please try again.',
-			'woocommerce-gateway-stripe'
-		);
 		if ( ! err ) {
 			return genericErrorMessage;
 		}
@@ -1700,7 +1710,7 @@ export const hasEmptyRequiredFields = ( requiredWrappers ) => {
 	for ( const wrapper of requiredWrappers ) {
 		const inputs = [
 			...wrapper.querySelectorAll(
-				'input.input-text, select, input[type="checkbox"]'
+				'input.input-text, textarea.input-text, select, input[type="checkbox"]'
 			),
 		];
 		if ( ! inputs.length ) {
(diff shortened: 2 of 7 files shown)


client/classic/upe/deferred-intent.js
```diff
@@ -102,27 +104,37 @@ jQuery( function ( $ ) {
...
-		}
-	}
 
-	$( 'form.checkout' ).on( generateCheckoutEventNames(), function () {
-		const $form = $( this );
+		// A saved token travels in the POST and needs no payment method, so the
+		// required-field state is none of our business on that path.
+		if ( isUsingSavedPaymentMethod( paymentMethodType ) ) {
+			return;
+		}
```
daledupreez, line 112: I am not sure that we can safely bypass required field checks for saved payment tokens. In many cases, the required fields are not related to billing, but to other customized sections of the checkout flow/UX. So I think we should still validate the required fields for this case. Either way, it feels like this may need some deeper thought as to how this should work.

client/classic/upe/deferred-intent.js
```diff
@@ -102,27 +104,37 @@ jQuery( function ( $ ) {
...
+		// created and then failed. jQuery :visible filters out fields hidden by
+		// conditional checkout logic (e.g. shipping fields when "Ship to different
+		// address" is unchecked).
 		if (
 			hasEmptyRequiredFields(
 				$form.find( '.validate-required:visible' ).toArray()
 			)
 		) {
-			return;
+			showErrorCheckout(
```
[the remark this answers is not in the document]
daledupreez: Are we guaranteed to run our code after any WooCommerce Core event handlers? Or is this indeterminate?

tests/e2e/tests/checkout/shortcode/third-party-required-fields.spec.js
```diff
@@ -0,0 +1,85 @@
+import { test, expect } from '@playwright/test';
+import config from 'config';
+import { payments } from '../../../utils';
```
daledupreez, line 3: ❤️ for adding e2e tests.

tests/e2e/tests/checkout/shortcode/third-party-required-fields.spec.js
```diff
@@ -0,0 +1,85 @@
...
+const injectThirdPartyRequiredField = ( page ) =>
+	page.evaluate( ( id ) => {
+		const field = document.createElement( 'p' );
+		field.id = id;
+		field.className = 'form-row form-row-wide validate-required';
+		field.innerHTML =
+			'<label>Third party field <abbr class="required" title="required">*</abbr></label>' +
+			'<input type="text" class="input-text" name="third_party_required_field" value="" />';
+		document.querySelector( 'form.checkout' ).prepend( field );
+	}, FIELD_ID );
```
daledupreez, line 32: Is it worth adding one or more types/structures of fields so we can test multiple behaviours rather than just one?

changelog.txt
```diff
@@ -33,7 +32,6 @@ xxxx-xx-xx - version 11.0.0
 * Fix - Prevent classic checkout from processing an empty payment method when custom required fields are present
```
wjrosa, line 32: Is the double-entry intentional?

client/classic/upe/payment-processing.js
```diff
@@ -1726,10 +1721,6 @@ export const hasEmptyRequiredFields = ( requiredWrappers ) => {
 			hasValue = first.value.trim() !== '';
 		}
 
-		if ( hasValue ) {
-			continue;
-		}
-
 		// Radios are not matched by the selector above, so a wrapper holding only
```
wjrosa, line 1724: I would reduce this and other comments a bit

client/classic/upe/payment-processing.js
wjrosa, line 1705: This only checks the first control inside the wrapper. Example: if a wrapper has a select with an empty placeholder and then a filled text input, we treat it as empty, but the server would accept it. 

Before this PR, that was ok because the form still submitted. Now we return `false`, so the customer gets "Please fill in all required fields." and cannot move forward. I think we should check all controls in the wrapper, not just the first one.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -3532,6 +3532,17 @@ protected function prepare_payment_information_from_request( WC_Order $order ) {
 		} else {
 			$payment_method_id = sanitize_text_field( wp_unslash( $_POST['wc-stripe-payment-method'] ?? '' ) ); // phpcs:ignore WordPress.Security.NonceVerification.Missing
 
+			if ( '' === $payment_method_id && empty( $_POST['wc-stripe-confirmation-token'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Missing
+				// The gateway the shopper picked is fine; what is missing are the payment details the
+				// browser should have attached. This path is shared by every deferred-intent flow, so
+				// the message must not name a cause: it is also reached when Stripe.js fails to load,
+				// when the payment element is remounting, and on tampered requests.
+				throw new WC_Stripe_Exception(
```
wjrosa, line 3540: This throw runs before `lock_order_payment()`, but the catch in `process_payment_with_payment_method()` always calls `unlock_order_payment()` and marks the order as failed. So a duplicate POST without a payment method (for example a second tab where Stripe.js failed to load) can release the lock held by the real request and set the order to failed, sending failure emails while the real payment is still completing. Before this change, that second request hit the lock check and got the "already being processed" message with no side effects. I think we should either move this check after the lock, or only unlock in the catch when this request took the lock.

client/classic/upe/deferred-intent.js
```diff
@@ -102,27 +104,37 @@ jQuery( function ( $ ) {
...
+				__(
+					'Please fill in all required fields.',
+					'woocommerce-gateway-stripe'
+				)
+			);
+			return false;
```
wjrosa, line 130: This gate also runs on the second submit that `processPayment` triggers after setting `hasCheckoutCompleted`. If an `updated_checkout` re-render empties a required field in that window (for example resetting the terms checkbox), we return `false` here and `processPayment` never consumes the flag, and nothing else resets it. When the customer fixes the field, changes the card and clicks again, the stale flag makes the handler submit the first payment method and the new card details are ignored. I think we should reset `hasCheckoutCompleted` when we block here.

client/classic/upe/deferred-intent.js
wjrosa, line 124: `showErrorCheckout` returns early when the page has no `.woocommerce-notices-wrapper`, with no fallback. On a theme or custom template that removed `woocommerce_output_all_notices`, we block the submit and show nothing, so the Place order button looks dead. Before this PR the form still submitted and WC's AJAX handler prepended the notice directly to `form.checkout`, which did not depend on that wrapper. Maybe we need a fallback container for this case.

client/classic/upe/payment-processing.js
```diff
@@ -1693,18 +1693,44 @@ export const confirmWalletPayment = async ( api, jQueryForm ) => {
  */
 export const hasEmptyRequiredFields = ( requiredWrappers ) => {
 	for ( const wrapper of requiredWrappers ) {
-		const input = wrapper.querySelector(
-			'input.input-text, select, input[type="checkbox"]'
-		);
-		if ( ! input ) {
+		const inputs = [
+			...wrapper.querySelectorAll(
+				'input.input-text, select, input[type="checkbox"]'
```
wjrosa, line 1698: This selector cannot match a `<textarea class="input-text">`, which is what `woocommerce_form_field` renders for the `textarea` type. So a required third-party textarea counts as filled, the form submits, and we get the same order-created-then-failed result this PR is fixing. WC core validates with `.input-text, select, input:checkbox`, which matches textareas. I think we should align with that selector.

client/classic/upe/deferred-intent.js
```diff
@@ -102,27 +104,37 @@ jQuery( function ( $ ) {
...
 		}
 
-		return processPaymentIfNotUsingSavedMethod( $form );
+		return processPayment( api, $form, paymentMethodType );
```
wjrosa, line 133: `processPayment` still returns `undefined` on some early exits, for example when `gatewayUPEComponents[ paymentMethodType ]` is missing (Stripe.js failed to load, or the type is not registered). That `undefined` propagates through this handler and WooCommerce submits the form with no payment method, hitting the new server-side guard and creating a failed order. The pay-for-order and add-payment-method handlers have the same hole. Maybe we should coerce those early returns to `false` too.

client/classic/upe/deferred-intent.js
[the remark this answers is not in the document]
diegocurbelo: WooCommerce’s submit handler triggers `checkout_place_order` before the gateway-specific `checkout_place_order_{gateway}` event. Our handler uses the gateway-specific event, so WooCommerce’s global place-order handlers run first.

`showErrorCheckout()` still triggers WooCommerce’s field validation events, so the empty fields are marked.

changelog.txt
diegocurbelo: Fixed when rebasing onto develop

client/classic/upe/deferred-intent.js
diegocurbelo: The behavior is the same as before, with the refactor it got moved to its own guard, the original was:
```
if ( ! isUsingSavedPaymentMethod( paymentMethodType ) ) {
	return processPayment( api, $form, paymentMethodType );
}
```

I refactored the function to keep the original flow that was easier to understand

client/classic/upe/payment-processing.js
diegocurbelo: Fixed in 5046f1dd1. The selector includes `textarea.input-text`, with empty and filled Jest cases.

client/classic/upe/deferred-intent.js
diegocurbelo: Fixed in 5046f1dd1. A missing component now shows an error and returns false from the shared function

client/classic/upe/deferred-intent.js
diegocurbelo: Fixed in 5046f1dd1

client/classic/upe/deferred-intent.js
diegocurbelo: Fixed in 5046f1dd1

tests/e2e/tests/checkout/shortcode/third-party-required-fields.spec.js
diegocurbelo: Text inputs, textareas, selects, multi-selects, checkbox groups, and radios are already covered in Jest

client/classic/upe/payment-processing.js
diegocurbelo: A `.validate-required` wrapper does not tell us whether sibling controls are alternatives or separate fields. Treating any filled sibling as sufficient could let an optional text input hide an empty required select.

The current implementation only uses group-wide checks where the semantics are clear: `checkbox` groups, `multi-select` values, and checked `radios`. For normal wrappers, it keeps the primary-control behavior. Supporting other mixed-control structures would need an explicit contract or a concrete plugin case so we do not allow incomplete checkout data.

wjrosa (commented):
Nice work! Looks good to me 🎉

wjrosa (approved):

PR #5884, opened 2026-08-31
Guard the express checkout availability probe against an unhydrated Blocks cart

### LINKED ISSUE #5876: Express checkout fails and breaks regular checkout form for free trials of WooCommerce subscription products
### Summary
On the block checkout, when the cart contains a free-trial subscription ($0 due today) and the page is loaded cold (uncached), the Express Checkout availability probe calls Stripe.js `elements()` before WooCommerce Blocks has hydrated the cart, so `currency` is an empty string:

```
Uncaught IntegrationError: Invalid value for elements(): currency should be one of
the following strings: aed, afn, … usd, … You specified: .
```

The error is thrown inside the `<Elements>` wrapper with no error boundary. The probe is memoized, so it never retries — a full page reload is the only recovery.  This also breaks the regular payment form below the contact information and billing fields.

### Environment

- WooCommerce Stripe Gateway 10.9.0
- WooCommerce 11.0.1, WooCommerce Subscriptions 9.1.0
- Block-based cart & checkout
- Optimized Checkout Element: enabled
- Express Checkout (Apple Pay / Google Pay / Link): enabled, with Checkout in the button locations
- Adaptive Pricing: enabled in settings, but `WC_Stripe_Helper::is_adaptive_pricing_supported()` returns `false` for the account
- Reproduced in Chrome 151 (headless), cache disabled

### Steps to reproduce

- Create a Simple subscription product with a free trial (e.g. 7 days) and no sign-up fee, so the amount due today is `$0.00`.
- Ensure Express Checkout is enabled for the Checkout location and the block checkout is in use.
- In a fresh browser session (cold cache), add that product to the cart and land on `/checkout/`.
- Observe the console error and the missing/blank payment areas on first load.
- Reload — it renders correctly.

### Expected
Express Checkout renders on a $0 free-trial cart since an automatic renewal takes place at the end of the trial.

### Root cause
In `build/upe-blocks.js`, the consolidated `stripe` express method's `canMakePayment:`

```
canMakePayment: ({ cart }) =>
  typeof wc_stripe_express_checkout_params !== "undefined"
  && !(0 === parseFloat(cart.cartTotals.total_price) && !getData("has_free_trial"))
  && !!getData()?.shouldShowExpressCheckoutButton
  && probe(method, api, cart)   // `uM`, lodash.memoize'd
```

The `!(total === 0 && !has_free_trial)` guard normally short-circuits a $0 cart before any Stripe element is built. But `has_free_trial ` comes from the server-localized `wc_stripe_express_checkout_params` (correctly `true`), so for a trial cart the guard is bypassed and `probe()` runs.
`probe()` is `lodash.memoize`d (keyed on the method name) and does:

```
root.render(
  createElement(Elements, {
    stripe: api.loadStripe(),
    options: {
      mode: hasFreeTrial ? "subscription" : "payment",
      amount: Number(...),
      currency: cart.cartTotals.currency_code.toLowerCase(),  // <-- ""
      ...
    }
  }, ...)
);
```

On the first render, WooCommerce Blocks passes its default cart totals — `@woocommerce/block-data` seeds them as `{ currency_code: "", total_price: "0", … } `— before the Store API data (or the page-bootstrapped cart) is applied. So `currency` is `""` and `elements()` throws.

Because `probe()` is memoized, the failed result sticks for the life of the page. On reload the cart is bootstrapped before the probe runs, so `currency_code` is populated and it works.

### Suggested fixes

- In `probe()` (and `vt`), bail out early when `cart.cartTotals.currency_code` is falsy — return `canMakePayment = false` rather than calling `elements()` with an empty currency.
- Don't memoize a negative/failed probe, or include `currency_code` in the memo key so it re-evaluates once the cart hydrates.
- Gate the `has_free_trial` exception on the cart actually being hydrated (e.g. also require a non-empty `currency_code`).
- Consider an error boundary around the express payment block so a throw there can't blank the whole checkout.

wjrosa:
Fixes STRIPE-1420
Fixes #5876

On the block checkout, `canMakePayment` for the express methods short-circuits $0 carts before probing, but a free-trial subscription cart is allowed through (`has_free_trial`). On the first render, WooCommerce Blocks passes its seeded default cart totals (empty `currency_code`) before the Store API data is applied, so the availability probe mounted a Stripe Elements group with `currency: ''`. Stripe's `elements()` throws an `IntegrationError`, and because the probe promise is memoized and never settles, express checkout and every regular payment option stay hidden until a full page reload.

This change puts a hydration guard in front of the memoized probe: when `cart.cartTotals.currency_code` is empty, `checkPaymentMethodIsAvailable` resolves `false` without mounting anything and without entering the memo cache. When Blocks re-runs `canMakePayment` after the cart hydrates, the probe runs normally with the real currency and is memoized as before.

The exported function keeps its signature and semantics; only the pre-hydration call changes behavior (it used to crash). No BC impact: this file is bundled frontend code with no public hooks or handles touched.

Verified in a local environment against a free-trial subscription cart on a cold block-checkout load: without the guard, 3 `IntegrationError` throws and an empty payment options block; with it, no errors and all payment options render.

## Testing instructions

1. Enable Express Checkout (Apple Pay / Google Pay) with the Checkout button location, and use the block-based cart and checkout pages.
2. With WooCommerce Subscriptions active, create a Simple subscription product with a free trial (e.g. 7 days) and no sign-up fee, so the amount due today is $0.00.
3. In a fresh browser session (or with the cache disabled in devtools), add that product to the cart and go to the checkout page.
4. Before this fix: the console shows `IntegrationError: Invalid value for elements(): currency should be one of the following strings: …` and the payment options area stays blank.
5. With this fix: no console error, and the payment options (and express checkout buttons, when the device supports them) render on the first load.
6. Regression check: with a regular (non-trial) paid product, confirm express checkout buttons and payment options still render on block checkout.

---

### Changelog entry

<details>

#### Comment

</details>

**Post merge**

### DIFF
--- a/client/express-checkout/utils/check-payment-method-availability.js
+++ b/client/express-checkout/utils/check-payment-method-availability.js
@@ -1,3 +1,4 @@
+import { Component } from 'react';
 import { createRoot } from 'react-dom/client';
 import { ExpressCheckoutElement, Elements } from '@stripe/react-stripe-js';
 import { memoize } from 'lodash';
@@ -15,10 +16,32 @@ import {
 import { transformPriceWithMinorUnits } from 'wcstripe/express-checkout/transformers/wc-to-stripe';
 
 /**
+ * React error boundaries must be class components. This one lets a failure
+ * while mounting the probe (e.g. Stripe rejecting the Elements options) be
+ * reported to the caller instead of becoming an uncaught render error.
+ */
+class ProbeErrorBoundary extends Component {
+	constructor( props ) {
+		super( props );
+		this.state = { hasError: false };
+	}
+
+	static getDerivedStateFromError() {
+		return { hasError: true };
+	}
+
+	componentDidCatch() {
+		this.props.onError();
+	}
+
+	render() {
+		return this.state.hasError ? null : this.props.children;
+	}
+}
+
+/**
  * Renders an invisible Stripe Express Checkout Element and waits for its ready
  * event to learn whether the given express payment method is available.
- *
- * Results are memoized so the availability check is only performed once per payment method.
  *
  * @param {string} paymentMethod The express payment method identifier (e.g. 'googlePay', 'applePay').
  * @param {Object} api           The WCStripeAPI instance used to load Stripe.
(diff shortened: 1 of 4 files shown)


client/express-checkout/utils/check-payment-method-availability.js
```diff
@@ -25,7 +25,7 @@ import { transformPriceWithMinorUnits } from 'wcstripe/express-checkout/transfor
  * @param {Object} cart          The WooCommerce cart object containing totals and currency info.
  * @return {Promise<boolean>} Promise that resolves to true if the payment method is available, false otherwise.
  */
-export const checkPaymentMethodIsAvailable = memoize(
+const probePaymentMethodAvailability = memoize(
```
jaclync: super nit, non-blocking: can keep the `check*` for consistency.
```suggestion
const checkPaymentMethodAvailability = memoize(
```

client/express-checkout/utils/check-payment-method-availability.js
```diff
@@ -25,7 +25,7 @@ import { transformPriceWithMinorUnits } from 'wcstripe/express-checkout/transfor
...
 	( paymentMethod, api, cart ) => {
 		return new Promise( ( resolve ) => {
```
jaclync, line 53: non-blocking: in case there's another error mounting the Stripe elements like this validation error, is there some way to catch it so that it doesn't get memoized?

client/express-checkout/utils/check-payment-method-availability.js
```diff
@@ -15,8 +15,8 @@ import {
 import { transformPriceWithMinorUnits } from 'wcstripe/express-checkout/transformers/wc-to-stripe';
 
 /**
- * Checks whether a given express payment method is available in the current context
- * by rendering an invisible Stripe Express Checkout Element and waiting for its ready event.
+ * Renders an invisible Stripe Express Checkout Element and waits for its ready
+ * event to learn whether the given express payment method is available.
  *
  * Results are memoized so the availability check is only performed once per payment method.
```
jaclync: non-blocking nit: maybe this line can be moved to the public docblock on `checkPaymentMethodIsAvailable` for visibility?

jaclync (approved):
Worked as expected! :shipit: For some reason, I couldn't reproduce it in Chrome incognito but was able to in Safari + clearing local site cookies/storage.

Site question from testing: is it expected that the Link EC sheet has no CTA to continue with the free trial? The Continue CTA is disabled as in the screenshot:

Image

client/express-checkout/utils/check-payment-method-availability.js
wjrosa: Good call. Added an error boundary around the probe in eb88f48a1. Any mount or load error now resolves false and evicts the memoized entry, so the next call probes again.

client/express-checkout/utils/check-payment-method-availability.js
wjrosa: Moved it to the public docblock in eb88f48a1.

wjrosa:
Thank you for testing! I don't think it is expected, but the sheet UI comes from Stripe, and this PR only changes when the availability probe runs. For free trials, we pass `mode: 'subscription'` and `amount: 0`, which the Stripe docs support exactly for free trials, so it looks like a Link limitation with a 0 amount. It should behave the same on `develop`. I will open a follow-up issue to confirm.

PR #5890, opened 2026-09-01
Accept shipping addresses in the express checkout sheet for free-trial carts with deferred shipping

### LINKED ISSUE #5889: Link express checkout: Continue button is disabled for free trial subscriptions
### Description

On a cart that contains a subscription with a free trial, the Link express checkout sheet opens but the Continue button is disabled, so the purchase cannot be completed with Link. Found by @jaclync while testing #5884: https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5884#pullrequestreview-5075735332

### Steps to reproduce

1. Enable express checkout methods, including Link.
2. Create a subscription product with a free trial period.
3. Add it to the cart and go to the checkout page.
4. Click the Link express checkout button and log in to Link.
5. The Continue button in the sheet is disabled.

### Expected behavior

The shopper can complete the free trial purchase with Link, like with Apple Pay and Google Pay. If Link cannot support this flow, we should hide the Link button for free trial carts instead of showing a sheet that cannot continue.

### Screenshots

See the screenshot in the review linked above.

### Additional context

- Free trial support for ECE was added in #4557. For free trials we create the elements with `mode: 'subscription'`, `amount: 0` and `paymentMethodCreation: 'manual'`.
- The [Stripe docs](https://docs.stripe.com/js/elements_object/create_without_intent) say `amount: 0` is supported in subscription mode exactly for coupons and free trials, and the `paymentMethodCreation: 'manual'` restrictions only mention Amazon Pay and Klarna, not Link.
- The sheet UI is rendered by Stripe, so nothing on our side disables the button. This looks like a Link limitation with a 0 amount, but it needs to be confirmed on `develop` first (it is not related to #5884, which only changes when the availability probe runs). If confirmed, we should raise it with Stripe.

wjrosa:
Fixes STRIPE-1427
Fixes #5889

On a cart with a free-trial subscription that needs shipping, Woo Subscriptions charges shipping with the recurring payments, so the Store API returns no shipping package for the sign-up cart. The express checkout `shippingaddresschange` handler treated "no rates" as an unserviceable address and rejected every address. In the Link sheet this shows "Your shipping address is unavailable for this purchase" and the final Continue button can never be enabled, so the purchase is impossible (found by @jaclync while testing #5884).

This PR:

- Falls back to the same `Pending` placeholder shipping option the sheet was opened with (from `get_default_shipping_option()`) when the cart has no shipping package at all. A package with empty rates still rejects, because that address genuinely cannot be served.
- Short-circuits the `shippingratechange` handler for the `pending` placeholder, since it is not a real WooCommerce rate and selecting it through the Store API would fail.

The handler is shared by the classic and Blocks express checkout paths, so both are covered.

Backward compatibility: no public surface changes. Only the internal behavior of the two shipping event handlers changes, and only for carts where the Store API returns no shipping package.

Verified end to end locally on Blocks checkout: before the fix the Link sheet rejected every address for a physical free-trial subscription; after the fix the sheet shows "Delivery: Pending (Free)", the Continue button enables, and completing the flow creates the order (total 0.00, paid via Link) with an active subscription and the next payment scheduled after the trial. A regular physical product still gets real rates ("Flat rate") in the same flow.

## Testing instructions

1. Enable express checkout methods, including Link, and make sure a shipping method exists (any zone with a flat rate).
2. With Woo Subscriptions active, create a **physical** subscription product with a free trial period.
3. As a shopper, add it to the cart and go to the Blocks checkout page.
4. Click the Link express checkout button, sign up or log in to Link (test mode), and enter a shipping address.
5. Before this PR: the sheet shows "Your shipping address is unavailable for this purchase" and Continue stays disabled. After: the delivery shows "Pending (Free)" and Continue enables.
6. Enter a test card (4242 4242 4242 4242) and continue. Confirm the order is created with total 0 and the subscription is active.
7. Regression check: repeat with a regular physical product and confirm real shipping rates still appear in the sheet and the purchase completes.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

https://claude.ai/code/session_01WeKekL9qsfgxNT1YVCFgzf

### DIFF
--- a/client/express-checkout/event-handler.js
+++ b/client/express-checkout/event-handler.js
@@ -47,12 +47,10 @@ export const shippingAddressChangeHandler = async ( event, elements ) => {
 
 		let shippingRates = transformCartDataForShippingRates( cartData );
 
-		// For a subscription with a free trial, Woo Subscriptions charges
-		// shipping with the recurring payments, so the sign-up cart has no
-		// shipping package at all. Keep the placeholder option the sheet was
-		// opened with instead of rejecting, which would block the purchase
-		// for every address. A package with empty rates still rejects below:
-		// that address genuinely cannot be served.
+		// No package at all means nothing ships now (Woo Subscriptions defers
+		// free-trial shipping to the recurring payments), so keep the
+		// placeholder the sheet was opened with. A package with empty rates
+		// still rejects below: that address cannot be served.
 		if (
 			shippingRates.length === 0 &&
 			( cartData?.shipping_rates?.length ?? 0 ) === 0
@@ -95,9 +93,8 @@ export const shippingAddressChangeHandler = async ( event, elements ) => {
  * @return {Promise<void>} Resolves when the shipping rate has been updated.
  */
 export const shippingRateChangeHandler = async ( event, elements ) => {
-	// The 'pending' placeholder (see get_default_shipping_option() in PHP) is
-	// not a real WooCommerce rate; selecting it through the Store API would
-	// fail, and there is nothing to persist anyway.
+	// The 'pending' placeholder from get_default_shipping_option() is not a
+	// real WooCommerce rate; the Store API would reject selecting it.
 	if ( event.shippingRate?.id === 'pending' ) {
 		event.resolve();
 		return;
--- a/client/express-checkout/event-handler.js
+++ b/client/express-checkout/event-handler.js
@@ -45,7 +45,24 @@ export const shippingAddressChangeHandler = async ( event, elements ) => {
 			),
 		} );
 
-		const shippingRates = transformCartDataForShippingRates( cartData );
+		let shippingRates = transformCartDataForShippingRates( cartData );
+
+		// For a subscription with a free trial, Woo Subscriptions charges
+		// shipping with the recurring payments, so the sign-up cart has no
+		// shipping package at all. Keep the placeholder option the sheet was
+		// opened with instead of rejecting, which would block the purchase
+		// for every address. A package with empty rates still rejects below:
+		// that address genuinely cannot be served.
+		if (
+			shippingRates.length === 0 &&
+			( cartData?.shipping_rates?.length ?? 0 ) === 0
+		) {
+			const defaultShippingOption =
+				getExpressCheckoutData( 'checkout' )?.default_shipping_option;
+			if ( defaultShippingOption ) {
+				shippingRates = [ defaultShippingOption ];
+			}
+		}
 
 		if ( shippingRates.length === 0 ) {
 			event.reject();
@@ -78,6 +95,14 @@ export const shippingAddressChangeHandler = async ( event, elements ) => {
  * @return {Promise<void>} Resolves when the shipping rate has been updated.
  */
 export const shippingRateChangeHandler = async ( event, elements ) => {
+	// The 'pending' placeholder (see get_default_shipping_option() in PHP) is
+	// not a real WooCommerce rate; selecting it through the Store API would
+	// fail, and there is nothing to persist anyway.
+	if ( event.shippingRate?.id === 'pending' ) {
+		event.resolve();
+		return;
+	}
+
 	try {
 		const cartData = await cartApi.selectShippingRate( {
 			package_id: 0,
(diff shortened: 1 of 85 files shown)


jaclync:
Hi @wjrosa 👋 since the E2E test is now in trunk and expecting failure, maybe this PR can remove the failure expectation line when it's on the latest `develop`:

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/5dba184c2b79999505f1b99cbfd8a09ba282aa97/tests/e2e/tests/express-checkout/free-trial-link.spec.js#L155

wjrosa:
Good call, done in c901e616. Merged develop and removed the `test.fail()` marker (and the outdated comment above it) from the saved-address free-trial test, so it now asserts the passing behavior this PR provides.

client/express-checkout/event-handler.js
```diff
@@ -78,6 +93,13 @@ export const shippingAddressChangeHandler = async ( event, elements ) => {
  * @return {Promise<void>} Resolves when the shipping rate has been updated.
  */
 export const shippingRateChangeHandler = async ( event, elements ) => {
+	// The 'pending' placeholder from get_default_shipping_option() is not a
+	// real WooCommerce rate; the Store API would reject selecting it.
+	if ( event.shippingRate?.id === 'pending' ) {
```
jaclync: non-blocking: could we use `getExpressCheckoutData( 'checkout' )?.default_shipping_option?.id` here instead of the hard-coded `pending` string that could diverge from the source of truth if the default ID ever changes on the server side?

client/express-checkout/event-handler.js
```diff
@@ -45,7 +45,22 @@ export const shippingAddressChangeHandler = async ( event, elements ) => {
...
+		// still rejects below: that address cannot be served.
+		if (
+			shippingRates.length === 0 &&
+			( cartData?.shipping_rates?.length ?? 0 ) === 0
+		) {
+			const defaultShippingOption =
+				getExpressCheckoutData( 'checkout' )?.default_shipping_option;
+			if ( defaultShippingOption ) {
+				shippingRates = [ defaultShippingOption ];
+			}
```
jaclync: non-blocking: this is now the third copy of the "fall back to `default_shipping_option`" pattern, as in two other flows:

Blocks: https://github.com/woocommerce/woocommerce-gateway-stripe/blob/e53c892b6ed4df7d5076227fe4ef7ce348fab9da/client/blocks/express-checkout/hooks.js#L93-L99
Classic: https://github.com/woocommerce/woocommerce-gateway-stripe/blob/e53c892b6ed4df7d5076227fe4ef7ce348fab9da/client/entrypoints/express-checkout/index.js#L96-L101

WDYT about extracting the default shipping rates getter to a shared helper (e.g. `getDefaultShippingOptions()` in `client/express-checkout/utils`) and use it in all three places?

tests/e2e/tests/express-checkout/free-trial-link.spec.js
jaclync: The E2E test now failed consistently with the fix, as the Link sheet's shipping state became "Delivery - pending":

Image

Instead of the shipping addresses form. Maybe it can be updated with:

```suggestion
await expect(
    popup.getByText( 'Pending', { exact: true } )
).toBeVisible( { timeout: 60 * 1000 } );
```

jaclync (approved):
The fix LGTM :shipit: good to merge after the E2E test is updated for it to pass consistently.

client/express-checkout/event-handler.js
wjrosa: Done — it now reads `default_shipping_option?.id` from the server instead of the hard-coded string.

client/express-checkout/event-handler.js
wjrosa: Good call — extracted `getDefaultShippingOptions()` into `express-checkout/utils` and used it in all three flows.

tests/e2e/tests/express-checkout/free-trial-link.spec.js
wjrosa: Applied, but scoped to just this assertion — the suggestion's range also covered the two checks and the closing brace below, so I kept those.

PR #5897, opened 2026-09-02
Claim an atomic owner row for order payment locks

kamlesh08:
Fixes STRIPE-1187

This PR builds on #5591. Merge order: #5591, then this PR, then #5901 and #5900. Change the base to `develop` after #5591 merges.

`WC_Stripe_Order_Helper::lock_order_payment()` reads the lock meta and then writes it. Two workers can both see no lock and both charge the customer. The method also cannot tell which worker owns the lock it removes.

**Scope.** This changes the lock for every caller. The callers are classic and Blocks checkout, Optimized Checkout, express checkout, the webhook handler, the order handler, the intent controller, and subscription renewals. A lock adds three queries to the meta write it already did: a row insert into `wp_options` with `autoload = 'no'`, a second read of the meta, and a read of the row. An owned unlock adds two: a row update and a row delete. On a test site with HPOS, a full lock ran 34 queries and an unlock 4; almost all of the lock's queries come from WooCommerce saving the order meta, which `develop` does too. The code reads the rows with direct queries, not `get_option()`, so it does not touch the option cache. A row that a crashed request leaves behind stays until the next lock on that order reclaims it, or until #5900 sweeps it.

- Before it writes the lock meta, the helper claims a row in `wp_options`. The row has a unique key. `INSERT IGNORE` claims a new row. A compare-and-set `UPDATE` claims an expired row. The lock meta now has the format `{expiry}|{owner_token}`. Old locks with only a timestamp still work.
- New methods: `acquire_order_payment_lock()`, `unlock_order_payment_if_owned()`, `is_order_payment_lock_owned()`, and `renew_order_payment_lock_if_owned()`. #5901 uses them on the renewal path. `lock_order_payment()` writes the lock on the order object it gets, as before. `acquire_order_payment_lock()` works on a clone, so the caller's unsaved order changes stay unsaved, and it returns the lock value that the caller passes to the `_if_owned` methods.
- `unlock_order_payment()` keeps its signature. It releases a lock that the same helper instance acquired. For any other lock it only clears an expired lock. All current callers lock and unlock in the same request, so their behaviour does not change.
- The renewal code reads the expiry from the new lock format. It does not change in any other way here.
- The deprecated gateway lock methods keep their old bodies and contracts. They write and delete a bare timestamp, and the helper honours such a lock while it is active. The deprecated reader treats a value that is not a scalar as locked, as the helper does.

**Multisite.** The owner row lives in the current site's options table. The in-memory key includes that table name. A `switch_to_blog()` call cannot mix locks between sites.

**Backward compatibility.** No public signature changes. When another worker owns the row, `lock_order_payment()` returns `true` without a call to `is_order_payment_locked()`; an override of that method cannot free a contended lock. Lock meta that is not a scalar, such as an array, counts as locked. A scalar that is not a timestamp counts as expired, as before, so a bad value cannot lock an order for good. Code that parses `_stripe_lock_payment` as a bare number still gets the expiry from an `(int)` cast.

## Testing instructions

**Environment:** Docker (`npm run up`), PHP dependencies installed.

**Action:** Run `npm run test:php`, `npm run phpstan`, and `npm run lint:php`.

**Validation:** The suite passes, apart from seven tests that also fail on `develop`. The helper tests cover these cases:

- The first of two workers gets the lock.
- The helper takes over an expired owner row.
- The helper does not release a lock that another worker replaced.
- Unsaved order meta survives a lock write.

The webhook test shows a checkout session event waits while another request holds the lock. The intent controller tests on `develop` still pass, so the checkout status update keeps an existing lock.

---

### Changelog entry

<details>

#### Comment

`Fix - Claim an atomic owner row before writing an order payment lock, and only release locks this request acquired`

</details>


daledupreez (commented):
@kamlesh08, I think it may be better to push the locking through the `WC_Stripe_Option_Lock` helper, at least for the atomic operations that we're handling at present.

PR #5904, opened 2026-09-03
Add E2E coverage for express checkout with free-trial subscriptions

jaclync:
Part of #5889

### What

Adds E2E coverage for express checkout (Link) with free-trial subscriptions — flagged in #5889 as under-tested — across the classic and Blocks cart/checkout surfaces, with and without shipping, including completing a purchase with Link.

### Coverage added

**`tests/e2e/tests/express-checkout/free-trial.spec.js` (`default` project)**

- **Cart & checkout render (8 tests):** classic cart, classic checkout, block cart, and block checkout — each for a virtual (no shipping) and a physical (needs shipping) free-trial product — assert the Link button renders and its sheet opens. Free-trial carts total 0, which normally hides express checkout; these pin the deliberate exception (`mode: 'subscription'`, `amount: 0`). They cover only that the button appears; the shipping-address code path (the #5889 code) runs after Link login and is covered below.
- **Product page (2 tests):** assert express checkout is hidden (see the APFS constraint below).

**`tests/e2e/tests/express-checkout/free-trial-link.spec.js` (`express-checkout-link` project)**

These drive the real sandbox Link enrollment/login flow, so they live in their own Playwright project + CI job — a slow Link popup can't push the shared `default` job over its timeout and cancel every other spec.

- **Completing the purchase with Link (2 tests, serial):** enroll a fresh Link account, save the test card, and complete the $0 free-trial purchase through to the order-received page; then log back in and assert the returning-account sheet keeps Continue enabled. Uses Stripe test-mode Link (any email enrolls; sandbox one-time passcode `000000`).
- **Saved-shipping-address regression (2 tests, serial):** enroll a Link account and save a shipping address via a regular purchase, then open a shipping-required free-trial cart, log in, wait for the shipping-address Store API round trip, and assert the saved address is accepted and Continue stays enabled. **Marked `test.fail()`** because it currently reproduces #5889 — the run errors ("Expected to fail, but passed") when the fix lands, prompting removal of the marker. This is the test that exercises the #5889 shipping-address code path.

### Test infrastructure

- Fixtures use WooCommerce Subscriptions **APFS** (All Products for Subscriptions) plans — the same `products/<id>/subscription-plans` REST approach as the existing `subscriptionData()` fixture — with a 14-day trial. No product-type manipulation needed.
- **Why APFS, and its constraint:** a native `subscription`-type product can't be created through the WC REST API (it only accepts the core `simple`/`grouped`/`external`/`variable` types), so using one would require flipping the `product_type` term via WP-CLI — which the fixtures can't do cleanly, since WP-CLI isn't reachable through the REST API the other fixtures use and instead means shelling out to the Docker container (and over SSH for remote `--base_url` runs), coupling the fixtures to the environment and duplicating the setup's `cli()` invocation. APFS attaches a plan to a `simple` product, so it works through the existing REST fixture path — at the cost of one limitation: APFS hides express checkout on the *product page* for any product carrying a subscription plan (it returns `true` from the `wc_stripe_hide_payment_request_on_product_page` filter via `WCS_ATT_Stripe_Compatibility::hide_stripe_quickpay`). So the product-page tests assert the button is hidden, and the free-trial express-checkout flow is exercised on the cart/checkout surfaces, where the trial is detected from the cart (`WC_Subscriptions_Cart::cart_contains_free_trial()`).
- New `express-checkout-link` Playwright project + CI matrix entry (`e2e-tests.yml`, `run-e2e-tests.yml`) for the Link enrollment specs, following the existing per-project pattern (`isk`, `acss`, …).
- New Link popup helpers in `tests/e2e/tests/express-checkout/utils.js` (`openLinkPopup`, `signUpForLink`, `fillLinkPaymentDetails`, `fillLinkShippingAddress`, `fillLinkCardDetails`, `loginToLink`).

### Testing

- `npm run test:e2e-run -- --project=default free-trial` — 10 render/product-page tests green locally (multiple runs).
- `npm run test:e2e-run -- --project=express-checkout-link` — 4 Link enrollment tests green locally (multiple runs), including the expected `test.fail()`.
- Both run in existing/new Playwright projects wired into the CI matrix.

### Backward compatibility

Test-only change; no runtime/public surface touched.


tests/e2e/tests/express-checkout/free-trial.spec.js
```diff
@@ -0,0 +1,250 @@
...
+
+// Free trial carts total 0 at checkout time, which is normally a condition for
+// hiding express checkout; free trials are the deliberate exception (the
+// element is created with mode: 'subscription' and amount: 0). These tests pin
+// that exception across the classic and Blocks cart/checkout surfaces, with and
+// without shipping, since a regression there silently removes the buttons.
+//
+// Coverage is limited to the cart/checkout surfaces: these are APFS
+// (subscribe-and-save) products, and APFS intentionally hides express checkout
+// on the product page, so that surface is asserted hidden rather than driven.
```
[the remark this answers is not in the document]
jaclync: PR description updated.

tests/e2e/tests/express-checkout/free-trial.spec.js
```diff
@@ -0,0 +1,250 @@
...
+			await waitForOrderReceivedPage( page );
+		} );
+
+		// Asserts the intended behavior; expected to fail until #5889 is
+		// fixed, so remove the test.fail() marker then.
+		// https://github.com/woocommerce/woocommerce-gateway-stripe/issues/5889
+		test( 'accepts the saved shipping address on a free-trial cart @blocks @express-checkout @subscriptions', async ( {
+			page,
+		} ) => {
+			test.fail();
```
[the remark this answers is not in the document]
jaclync: I'd prefer keeping `test.fail()` to make the failure explicit, and it will be removed when the fix is in (if this PR ships earlier, in the fix PR).

tests/e2e/tests/express-checkout/utils.js
```diff
@@ -36,4 +36,180 @@ export const assertLinkModalLoads = async ( page, isBlockPage = false ) => {
 	await expect( popup.getByTestId( 'pay-button' ) ).toBeVisible( {
 		timeout: 60 * 1000,
 	} );
+
+	return popup;
```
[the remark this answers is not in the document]
jaclync: This retry pattern is pre-existing - it's verbatim from the existing `assertLinkModalLoads` helper, which this PR just extracts into `openLinkPopup`. I kept the behavior identical rather than change a helper that the existing `express-checkout.spec.js` also depends on. If we see actual flakiness here I’m happy to tune the inner/overall timeouts as a separate change.

tests/e2e/tests/express-checkout/free-trial.spec.js
```diff
@@ -0,0 +1,255 @@
...
+				page,
+			} ) => {
+				await addSubscriptionToCart( page, variant.id() );
+				await page.goto( '/checkout' );
+				await assertLinkModalLoads( page, true );
+			} );
+		} );
+	}
+
+	test.describe( 'completing the purchase with Link', () => {
```
wjrosa, line 137: The `default` E2E job has a 20-minute timeout, and the last run of this PR took 18 minutes, including one free-trial test that retried twice. 

These four tests have a 240s timeout each, and a serial retry runs the whole group again, so one slow Link popup can push the job over the limit and cancel it, and we lose the results of all the other specs too. 

I think we should move the Link enrollment tests to their own Playwright project and matrix entry, as we do for `isk` and `acss`.

tests/e2e/tests/express-checkout/free-trial.spec.js
```diff
@@ -0,0 +1,255 @@
...
+	// Both product variants run the same cart/checkout coverage. The product
+	// IDs are read lazily because they are assigned in beforeAll, after these
+	// blocks are collected.
+	const productVariants = [
+		{
+			title: 'without shipping (virtual product)',
+			id: () => virtualProductId,
+		},
+		{
+			title: 'with shipping (physical product)',
```
wjrosa, line 78: The four tests for this variant only check that the Link popup opens with a pay button on the first email screen. Link fires `shippingaddresschange` only after login, so the shipping code path for a trial cart (the same code as #5889) is never reached here, and a regression there still passes all four. The only shipping assertion in the file is inside the `test.fail()` test, where a failure is reported as expected. I think we should either drive the physical variant past the login step, or change the header comment and the PR description so they do not claim shipping coverage.

tests/e2e/tests/express-checkout/free-trial.spec.js
```diff
@@ -0,0 +1,255 @@
...
+			const popup = await openLinkPopup( page, true );
+			await loginToLink( popup, linkEmail );
+
+			await expect( popup.getByText( 'Shipping addresses' ) ).toBeVisible(
+				{ timeout: 60 * 1000 }
+			);
+			await expect(
+				popup.getByText( 'Unavailable for this purchase' )
+			).toBeHidden();
+			await expect( popup.getByTestId( 'pay-button' ) ).toBeEnabled();
```
wjrosa, line 252: These two assertions are single snapshots, and on a physical cart they can run while the `shippingaddresschange` round trip is still pending. `toBeHidden` passes right away because the text is not there yet, and `toBeEnabled` can pass before the reject lands. Today this shows up as a flaky "Expected to fail, but passed". After the fix lands and the marker is removed, a regression that brings the reject back is a silent green. I think we should wait for the `/wc/store/v1/cart/update-customer` response, or poll, before asserting.

tests/e2e/tests/express-checkout/utils.js
```diff
@@ -36,4 +36,180 @@ export const assertLinkModalLoads = async ( page, isBlockPage = false ) => {
...
+		.selectOption( address.state_iso );
+	await popup.locator( 'input[name="locality"]' ).fill( address.city );
+	await popup
+		.locator( 'input[name="addressLine1"]' )
+		.fill( address.address_1 );
+
+	const closeSuggestions = popup.getByTestId(
+		'autocomplete-suggestion-list-close'
+	);
+	if ( await closeSuggestions.isVisible().catch( () => false ) ) {
```
wjrosa, line 159: `isVisible()` does not wait, and the `Escape` runs right after the `fill()` on the street field. Link loads the suggestions asynchronously, so the list can appear after the `Escape` and sit over the "Continue to payment" button, and the click then retries until the action timeout. I think we should wait for the close button with a short timeout before pressing `Escape`, or fill the street before another field so the next `fill()` blurs it.

tests/e2e/tests/express-checkout/free-trial.spec.js
jaclync: Done in b0f2bc179 - `free-trial-link.spec.js` now runs in its own `express-checkout-link` project and matrix entry, excluded from default. The fast test cases that don't go through the payment flow stay in `free-trial.spec.js`. b18429508 also sizes the Link popup helper's retries to fit inside the 120s/240s test budgets.

tests/e2e/tests/express-checkout/free-trial.spec.js
jaclync: Addressed in b0f2bc179 - the header comment and PR description now scope these tests to button rendering only, and the shipping path past Link login is driven in `free-trial-link.spec.js`.

tests/e2e/tests/express-checkout/utils.js
jaclync: Updated in 2d02fa4e3 - the street field is filled last so it keeps focus for the Escape, and the Continue click retries Escape + click in a short `toPass` loop so a late-appearing overlay can't intercept it.

tests/e2e/tests/express-checkout/free-trial.spec.js
jaclync: Done in b0f2bc179 - the test now waits for the `/wc/store/v1/cart/update-customer` response triggered by Link login before asserting.

wjrosa (approved):
Looks good to me and works as expected! Thanks for addressing the suggestions 👍

jaclync:
Thanks for the review! Merging now, with CI GitHub token updated after it [stopped working a few hours ago](https://github.com/woocommerce/woocommerce-gateway-stripe/actions/runs/34314181043/job/102346834486).

Changes since the approval, also ref p1788933602920399-slack-C055WHLA98D - WooCommerce 11.1.0 (what `WC=latest` in CI used as of in the last day) broke the subscription specs suite-wide, plus two hardening commits on the Link popup helper.

WC 11.1.0 breakage: WC no longer relabels the APFS add-to-cart button to "Sign up" when a subscription plan is selected, so every spec clicking 'Sign up' failed with element-not-found, including specs this PR never touched.

0b45bf15b: `addSubscriptionToCart` clicks `form.cart button[type="submit"]` instead of matching button copy. The subscribe path is still asserted via APFS's subscribe-to-action-input before the click, so the selector is label- and locale-proof.
4960450ad: the five specs that inlined the same select-plan → add-to-cart sequence now use the shared helper.

Also Link popup hardening in b18429508 and 1703feb2c. However, they don't completely fix the pre-existing flaky `openLinkPopup`.

PR #5905, opened 2026-09-03
Make e2e suite theme-agnostic and run a block theme (purple) in CI

jaclync:
## What & why

The e2e suite was coupled to the Storefront (classic) theme, so it could not validate the plugin under block/FSE themes. This makes the suite theme-agnostic and adds block-theme coverage to CI, using the WooCommerce **purple** starter theme as the block-theme case.

Three self-contained commits:

1. **Make e2e checkout helpers theme-agnostic** — resolve cart state via the Store API instead of shop-loop XPath and the header cart counter; match notices/headings by text/role (or dual selectors) rather than classic wrapper classes; re-verify add-to-cart against the cart (block themes drop a click that lands before Interactivity API hydration); scope the shortcode card frame to the payment-input iframe to avoid a races-to-multiple-matches strict-mode failure; dispatch the Place Order click so a post-hydration keep-in-view scroll can't move the button out from under a physical click.
2. **Isolate e2e Docker stacks** via `E2E_PROJECT` / `E2E_WP_PORT` / `E2E_DB_PORT` so two checkouts can run the Docker e2e flow side by side (defaults unchanged).
3. **Parameterize the e2e theme and run a block theme in CI** — `WP_THEME` (default `storefront`) selects the theme at setup; WordPress.org slugs install by slug, and a block theme not on WordPress.org (purple) is fetched from a **pinned** woo-themes commit tarball, zipped, and installed. The CI matrix gains a theme axis: all projects run on Storefront as before, plus the theme-sensitive checkout projects (`default`, `optimized-checkout`) also run on purple.

## Testing

- Under purple, the default-project failure set became **identical** to the Storefront baseline (no purple-only failures).
- Storefront still passes (the shortcode iframe fix also repaired a pre-existing classic-path flake).
- Ran the real `WP_THEME=purple … npm run test:e2e-setup` end-to-end: it fetches, zips, and activates purple (exit 0), and pre-order specs complete real orders on the provisioned site.
- Residual suite failures are pre-existing, theme-independent environment effects (empty Stripe fee metadata, cold account-cache races) that occur identically on both themes; CI (real secrets + working webhook listener) should not hit them.

## Backward compatibility

Test infrastructure only — no changes to plugin runtime code, public APIs, hooks, options, or assets. Default behavior of the e2e scripts is unchanged (Storefront, ports 8088/6789), so existing local and CI usage is unaffected.

## Notes

- The purple tarball is pinned to a woo-themes commit SHA so trunk drift can't silently change CI; bump the SHA in both workflows to pick up newer revisions.
- The tarball path requires `tar` and `zip` on the runner (present on `ubuntu-latest`).

### DIFF
--- a/tests/e2e/bin/common.sh
+++ b/tests/e2e/bin/common.sh
@@ -3,6 +3,14 @@
 CWD=$(pwd)
 E2E_ROOT="$CWD/tests/e2e"
 
+# The compose project (and its container-name prefix) and the host ports are
+# global to the machine: two checkouts using the defaults clobber each other's
+# stacks. Override these to run an isolated second stack side-by-side.
+# Exported so docker compose can interpolate them in env/docker-compose.yml.
+export E2E_PROJECT=${E2E_PROJECT:-wcstripe-e2e}
+export E2E_WP_PORT=${E2E_WP_PORT:-8088}
+export E2E_DB_PORT=${E2E_DB_PORT:-6789}
+
 ADMIN_USER=${ADMIN_USER-admin}
 ADMIN_PASSWORD=${ADMIN_PASSWORD-admin}
 ADMIN_EMAIL=${ADMIN_EMAIL-admin@example.com}
@@ -38,5 +46,5 @@ redirect_output() {
 # --user xfs forces the wordpress:cli container to use a user with the same ID as the main wordpress container.
 # See: https://hub.docker.com/_/wordpress#running-as-an-arbitrary-user
 cli() {
-	docker run -i --rm --user 33:33 --env-file ${E2E_ROOT}/env/default.env --volumes-from "wcstripe-e2e-wordpress" --network container:"wcstripe-e2e-wordpress" wordpress:cli "$@"
+	docker run -i --rm --user 33:33 --env-file ${E2E_ROOT}/env/default.env --volumes-from "${E2E_PROJECT}-wordpress" --network container:"${E2E_PROJECT}-wordpress" wordpress:cli "$@"
 }
--- a/tests/e2e/bin/common.sh
+++ b/tests/e2e/bin/common.sh
@@ -3,14 +3,6 @@
 CWD=$(pwd)
 E2E_ROOT="$CWD/tests/e2e"
 
-# The compose project (and its container-name prefix) and the host ports are
-# global to the machine: two checkouts using the defaults clobber each other's
-# stacks. Override these to run an isolated second stack side-by-side.
-# Exported so docker compose can interpolate them in env/docker-compose.yml.
-export E2E_PROJECT=${E2E_PROJECT:-wcstripe-e2e}
-export E2E_WP_PORT=${E2E_WP_PORT:-8088}
-export E2E_DB_PORT=${E2E_DB_PORT:-6789}
-
 ADMIN_USER=${ADMIN_USER-admin}
 ADMIN_PASSWORD=${ADMIN_PASSWORD-admin}
 ADMIN_EMAIL=${ADMIN_EMAIL-admin@example.com}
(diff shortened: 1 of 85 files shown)


tests/e2e/bin/common.sh
```diff
@@ -3,6 +3,14 @@
 CWD=$(pwd)
 E2E_ROOT="$CWD/tests/e2e"
 
+# The compose project (and its container-name prefix) and the host ports are
+# global to the machine: two checkouts using the defaults clobber each other's
+# stacks. Override these to run an isolated second stack side-by-side.
+# Exported so docker compose can interpolate them in env/docker-compose.yml.
+export E2E_PROJECT=${E2E_PROJECT:-wcstripe-e2e}
+export E2E_WP_PORT=${E2E_WP_PORT:-8088}
+export E2E_DB_PORT=${E2E_DB_PORT:-6789}
```
daledupreez: Might it be worth pushing this logic into a dedicated PR that automates this as part of the `npm run worktree:setup` script? That should give us better default behaviour when multiple worktrees are in use.

That should also address some subtle issues that can come up when you either have old/stale variables in the current terminal or have a new terminal that doesn't have the variables available yet.

It also feels like pulling this out will reduce the scope of this PR, which should be focused on block theme support.

tests/e2e/tests/checkout/shortcode/retries.spec.js
```diff
@@ -57,8 +57,15 @@ test( 'customer can retry payment, with a different card @smoke', async ( {
...
-	// Expect the order to fail
-	await expect( page.locator( '.woocommerce-error' ) ).toBeVisible();
+	// Expect the order to fail. Classic themes render the error as
+	// `.woocommerce-error`, block themes as an error notice banner.
+	await expect(
+		page
+			.locator(
+				'.woocommerce-error, .wc-block-components-notice-banner.is-error'
+			)
+			.first()
```
daledupreez: Not blocking: is it worth pushing this logic into a helper function, e.g. `getErrorNotices( page )`?

tests/e2e/tests/express-checkout/express-checkout.spec.js
```diff
@@ -8,12 +8,34 @@ const addProductToCart = async ( page, productSlug = 'beanie' ) => {
...
+	const dropdown = page.getByLabel( attributeLabel, { exact: true } );
+	const isSelect =
+		( await dropdown.count() ) > 0 &&
+		( await dropdown
+			.first()
+			.evaluate( ( el ) => el.tagName === 'SELECT' ) );
+	if ( isSelect ) {
+		await dropdown.selectOption( value );
+	} else {
+		await page.getByRole( 'radio', { name: value, exact: true } ).click();
```
daledupreez: Not blocking, but is it worth requiring the radio element to be a child of the product cart form? We're using the top-level `page` object to locate the radio button, which could be flaky in some situations.

tests/e2e/utils/payments.js
```diff
@@ -9,11 +9,53 @@ import { verifyOrderChargedAmount } from './admin';
...
+	const expectedCount = initialCount + 1;
+
+	// Block themes submit the add via the Interactivity API, and a click that
+	// lands before its scripts hydrate is silently dropped — so confirm against
+	// the cart itself and re-click when the count did not move.
+	await retryWithBackoff(
+		async () => {
+			// Skip the click when a previous slow add already landed, so a
+			// retry cannot add the item twice.
+			if ( ( await getCartItemsCount( page ) ) !== expectedCount ) {
```
daledupreez: Nit: It feels like a `<` would be safer, especially in cases where we end up with too many items.
```suggestion
			if ( ( await getCartItemsCount( page ) ) < expectedCount ) {
```

tests/e2e/bin/common.sh
jaclync: Yes, I removed the related E2E code and created a separate issue STRIPE-1442 that I'm working on. I included these diffs in this draft PR to allow multiple worktrees running E2E tests in parallel locally, with this PR to test the Purple theme originally.

This PR is still in draft with E2E issues that I'm fixing, I will mark it ready for review when it is.

PR #5907, opened 2026-09-04
Add admin Stripe disputes  rest api endpoint

en-adobrescu:

Adds a simple  admin REST endpoint for retrieving a Stripe API disputes listing and an endpoint for retrieving one dispute details.

The responses received from Stripe API are returned after being filtered against a whitelist.

## Testing instructions
1. Navigate to wp-admin while logged in as a user with admin permissions.
2. Open your browser console
3. Run the following snippet to call the API:
```
const response = await wp.apiFetch( { path: '/wc/v3/wc_stripe/disputes' } );
response;
```
```
{
    "object": "list",
    "has_more": false,
    "data": [
        {
            "id": "du_1TtPY7ABBA0dQbSBIBMx5Rmj",
            "created": 1784109379,
            "amount": 7900,
            "currency": "usd",
            "status": "lost",
            "payment_intent": "pi_3TtPYABBAACDCbSB0Jb3hqkX",
            "reason": "fraudulent",
            "evidence_details": {
                "due_by": 1784851199
            }
        }
    ]
}
```

4. The endpoint supports the following parameters:

- limit
- starting_after
- ending_before
- created 
- payment_intent
- charge

They are documented here: 
https://docs.stripe.com/api/disputes/list?api-version=2026-06-24.preview&rds=1

5. To test the dispute details endpoint send the request to:

[WordPress instance URL]/wp-json/wc/v3/wc_stripe/disputes/[dispute_id].

---

### Changelog entry

<details>

#### Message
Add a Stripe REST proxy endpoint for querying payout endpoints of Stripe API.

</details>

**Post merge**

### DIFF
--- a/includes/admin/class-wc-stripe-rest-disputes-controller.php
+++ b/includes/admin/class-wc-stripe-rest-disputes-controller.php
@@ -1,0 +1,384 @@
+<?php
+/**
+ * Class WC_Stripe_REST_Disputes_Controller
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * REST controller exposing Stripe dispute details to the admin UI.
+ *
+ * @since 11.1.0
+ */
+class WC_Stripe_REST_Disputes_Controller extends WC_Stripe_REST_Base_Controller {
+
+	protected const DISPUTE_ID_PATTERN        = 'du_[A-Za-z0-9_]+';
+	protected const PAYMENT_INTENT_ID_PATTERN = 'pi_[A-Za-z0-9_]+';
+	protected const CHARGE_ID_PATTERN         = 'ch_[A-Za-z0-9_]+';
+
+	/**
+	 * Endpoint path.
+	 *
+	 * @var string
+	 */
+	protected $rest_base = 'wc_stripe/disputes';
+
+	protected const STRIPE_SINGLE_RESPONSE_ALLOWED_FIELDS = [
+		'object',
+		'id',
+		'created',
+		'amount',
+		'currency',
+		'status',
+		'description',
+		'payment_intent.latest_charge.balance_transaction',
+		'payment_intent.latest_charge.billing_details',
+		'payment_intent.latest_charge.payment_method_details',
+	];
+
+	protected const STRIPE_SINGLE_EXPAND_PARAM = [
+		'payment_intent.latest_charge.balance_transaction',
+		'payment_intent.latest_charge.billing_details',
+		'payment_intent.latest_charge.payment_method_details',
+	];
+
+	protected const STRIPE_LIST_RESPONSE_ALLOWED_FIELDS = [
+		'object',
+		'has_more',
+		'data.id',
+		'data.created',
+		'data.amount',
+		'data.currency',
+		'data.status',
+		'data.payment_intent',
+		'data.reason',
+		'data.evidence_details.due_by',
+	];
+
+	protected const STRIPE_LIST_EXPAND_PARAM = [];
+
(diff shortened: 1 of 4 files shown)


includes/admin/class-wc-stripe-rest-disputes-controller.php
```diff
@@ -0,0 +1,384 @@
...
+	/**
+	 * Validate that a parameter is a non-empty string.
+	 *
+	 * @param string $param_value The parameter value.
+	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 * @param string $param_name The parameter name.
+	 *
+	 * @return bool
+	 */
+	public static function validate_non_empty_string( $param_value, $request, $param_name ) {
```
jaclync: ❓ Is this used in this PR? If it's unused, let's remove.

includes/admin/class-wc-stripe-rest-disputes-controller.php
```diff
@@ -0,0 +1,384 @@
...
+	 *
+	 * @param string $param_value The parameter value.
+	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 * @param string $param_name The parameter name.
+	 *
+	 * @return bool
+	 */
+	public static function validate_charge_id( $param_value, $request, $param_name ) {
+		return 1 === preg_match( '/^' . self::CHARGE_ID_PATTERN . '$/', $param_value );
+	}
```
jaclync, line 295: For `preg_match`, it expects the value for matching a `string` as in [the doc](https://www.php.net/manual/en/function.preg-match.php), which will likely throw in PHP 8 if the type mismatches. Since the param value can be any type, if it's an array like `?payment_intent[]=pi_x`, this will result in a 500 error. I think it's worth checking the type first, and these 3 similar validation functions can perhaps share a private helper.

includes/admin/class-wc-stripe-rest-disputes-controller.php
```diff
@@ -0,0 +1,384 @@
...
+	 *
+	 * @param string $param_value The parameter value.
+	 * @param WP_REST_Request<array<string, mixed>> $request The incoming REST request.
+	 * @param string $param_name The parameter name.
+	 *
+	 * @return bool
+	 */
+	public static function validate_non_empty_string( $param_value, $request, $param_name ) {
+		return '' !== trim( $param_value );
+	}
```
jaclync: 💭 There are 6 identical functions between this and the [payouts controller](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/2072e7f4ef1573a0e2e28beb283da8d969aa1bc6/includes/admin/class-wc-stripe-rest-payouts-controller.php#L261-L356):

- fetch_from_stripe
- build_params_to_forward
- sanitize_unix_timestamp
- validate_unix_timestamp
- is_valid_timestamp
- validate_non_empty_string (unused as in the comment above)

I think it's worth sharing these functions to avoid duplication. WDYT of extracting the validators into a static `WC_Stripe_REST_Request_Validation` helper mirroring `WC_Stripe_REST_Response_Filter`, and `fetch_from_stripe` can be in an intermediate abstract class or a base controller? cc @diegocurbelo (the other reviewer) @daledupreez for architectural input.

includes/admin/class-wc-stripe-rest-disputes-controller.php
en-adobrescu: Thank you, Jaclync. Removed the unused code.

includes/admin/class-wc-stripe-rest-disputes-controller.php
en-adobrescu: Done, added is_string(...) check at the beginning of each method

PR #3150, opened 2024-05-27
Update the billing address on saved card before creating the payment intent/processing payment

mattallan:
Fixes #1682 

## Changes proposed in this Pull Request:

When a customer saves a payment method to their account, the billing address provided on the checkout is saved to that payment method. Say this customer moves/changes addresses and proceeds through checkout using their saved payment method Stripe could block the payment due to zip code verification erros.

In Stripe, we currently update the billing address on the saved method asynchronously on an action scheduler hook, but this doesn't update the payment method address quick enough and it needs to happen before the intent is created.

This PR fixes this issue by:
1. Removing code which updates the saved payment methods billing address on an action scheduler hook
2. Updating the saved payment method before creating the payment/setup intent
3. Move the existing `update_saved_payment_method()` to the abstract Stripe gateway base class so it can be called when UPE is enabled and disabled.

## Testing instructions

I wasn't able to replicate zip code verification errors so to test these changes I just confirmed that the payment method billing address wasn't being updated at the time of processing the payment and didn't match the shipping address.

The two addresses I used for testing were:

Address A:
```
Australia
123 Brisbane Street
Brisbane City, Queensland, 4000
```

Address B:
```
Australia
123 Sydney Street
Sydney, New South Wales, 2000
```

1. Make sure UPE is enabled and also the "Enable payments via saved cards" setting is selected
2. Add a product to your cart and while on the checkout page, enter Address A
3. Enter a new card and make sure to save it to your account
4. Open Stripe and confirm the payment method has the "Brisbane Street" address:
image
5. While on `develop`, add another product the cart, visit the checkout.
6. Enter in Address B into the billing details and use the same saved card that was saved above.
7. Submit the checkout and view the new payment in the stripe dashboard.
8. Notice that the payment method still has the old customer address, but the payment itself has the new address: image
9. Check out this branch and run through the checkout again with a different Address.
10. Confirm that the payment method billing address has been updated and matches the payment method address.

Here's a full list of tests I ran:
- [x] UPE enabled
    - [x] Existing saved card
    - [x] New saved card - shouldn't update payment method as it will be set when creating
    - [x] No saved payment method
- [x] UPE disabled
    - [x] Existing saved card
    - [x] New saved card
    - [x] No saved payment method
- [x] SEPA saved payment method

---

-   [ ] Covered with tests (or have a good reason not to test in description ☝️)
-   [x] Added changelog entry **in both** `changelog.txt` and `readme.txt` (or does not apply)
[...]

Issue #1682, opened 2021-08-13
Gateway not updating billing address for a saved card when placed an order


kuldeep3:
**Describe the bug**
When a user places an order and adds a card as a payment method and saves that card for a future transaction, stripe uses the billing address added by the user while checking out. If the user again places another order using the same card but with a different billing address, then the stripe doesn't update the user billing address from the checkout fields. Instead, it fetches the billing address that is attached to the saved card on stripe's servers which results in a zip code validation error for users.

**To Reproduce**
1. Place an order and proceed to checkout.
2. On the checkout page, fill in the billing address and add a new payment method and then click on save this payment method for the future.
3. After successfully placing an order, create another order with a different billing address and choosing the same card that was used in the previous order.
4. Check stripe logs, there you will find that the billing address on the latest order is not reflected correctly, it shows the billing address that has been already attached to the saved card on stripe's server.

**Expected behavior**
Whenever the customer creates an order and uses a previously saved card on the order, then the billing address should be updated as per the address added by the customer while checking out or that is stored in the customer profile.

**Screenshots**

https://user-images.githubusercontent.com/37093310/129411904-4e655547-bb36-4d37-8fa5-47934f4728db.mp4

**Environment (please complete the following information):**
 - WordPress Version - 5.7.2
 - WooCommerce Version - 5.3.1
 - Stripe Plugin Version - 5.2.0
 - Browser [e.g. chrome, safari] and Version - Chrome 92.0.4515.131

**Additional context**
--NA--

nagpai:
Can you confirm if you did these tests with only WooCommerce core plugin and WooCommerce Stripe Payment Gateway plugin active on either StoreFront or TwentyTwenty one theme?

A bug should be logged here in GH only if you are able to reproduce this under the test conditions mentioned above.  If you are unable to replicate it in the test set-up. It would point to a conflict caused by a third party theme or a plugin.

nagpai:
I am unable to replicate the error on a test site with Storefront theme, WooCommerce 5.5.2 and Stripe 5.3.0.

kuldeep3:
@nagpai I am able to reproduce this with only the WooCommerce core plugin and WooCommerce Stripe Payment Gateway plugin active on the Twenty Twenty-One v1.2 theme.

https://user-images.githubusercontent.com/37093310/129445589-218887a1-f6fa-42fa-aa58-e20d955510f4.mp4

kuldeep3:
[...]

PR #5910, opened 2026-09-04
Stop rewriting a saved payment method's identity fields at checkout

wjrosa:
Fixes STRIPE-1390

When a shopper pays with a saved payment method, `update_saved_payment_method()` copied the order's full billing details (name, email, phone, address) onto the PaymentMethod in Stripe before the intent was created. Any checkout attempt, including declined ones or ones run under the wrong account, could overwrite the original cardholder's identity on the saved method.

This PR limits the write to the billing address. Name, email, and phone on an existing saved method are no longer touched at checkout. The write is also skipped when the order has no billing address, instead of blanking the stored one.

The address still goes on the payment method before the intent: #3150 placed it there because Stripe's AVS check reads the address from the PaymentMethod, and moving it after confirmation would bring back the postal-code declines from #1682.

**Backward compatibility.** `update_saved_payment_method( $payment_method_id, $order )` keeps its signature; only the payload changes. No hooks, options, or site-scoped state are touched. Stale PHPStan baseline entries for the removed accessors were dropped.

## Testing instructions

1. Enable "Enable payments via saved cards" in the Stripe settings.
2. As a logged-in shopper, save a card at checkout using address A and note the name and email on the order.
3. In the Stripe dashboard, confirm the PaymentMethod's billing details show that name, email, and address A.
4. Check out again with the saved card, using a different billing name and email and address B. Place the order.
5. Reload the PaymentMethod in Stripe. The address should be B; the name and email should be unchanged from step 2.
6. Repeat step 4 with a saved card that declines (e.g. `4000 0000 0000 0002`). Name and email must still be unchanged.

Unit tests in `tests/phpunit/class-wc-stripe-payment-gateway-test.php` assert the request body sent to Stripe. The manual steps above were not run against a live Stripe test account.

---

### Changelog entry

**Post merge**

### DIFF
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -2810,13 +2810,8 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 	/**
 	 * Copies the order's billing address onto a saved payment method in Stripe.
 	 *
-	 * Only the address is written. The saved method's name, email, and phone identify the person
-	 * who originally saved it and must survive later checkouts: a checkout running under the wrong
-	 * account (a shared login, a session mix-up at a caching layer) would otherwise overwrite the
-	 * cardholder's identity with someone else's, and Stripe's copy is what fraud reviews rely on.
-	 * The address is the exception because Stripe's AVS check reads it from the payment method,
-	 * so a shopper who moved would be declined on postal-code mismatch unless it is refreshed
-	 * before the charge.
+	 * Only the address is written, since AVS reads it from the payment method. Name, email, and
+	 * phone identify the original cardholder and must not be overwritten by a later checkout.
 	 *
 	 * @param string       $payment_method_id The payment method to update.
 	 * @param WC_Order|int $order             Order object or id.
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -2837,6 +2837,16 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 				return;
 			}
 
+			$payment_method          = WC_Stripe_API::get_payment_method( $payment_method_id );
+			$payment_method_customer = is_object( $payment_method ) && isset( $payment_method->customer ) ? $payment_method->customer : null;
+			$order_customer          = $this->get_stripe_customer_id( $order );
+
+			// Stripe is authoritative here because public callers are not required to enter through the saved-token ownership checks.
+			if ( ! is_string( $payment_method_customer ) || ! is_string( $order_customer ) || $payment_method_customer !== $order_customer ) {
+				WC_Stripe_Logger::warning( 'Skipped saved payment method address update because its Stripe customer does not match the order.' );
+				return;
+			}
+
 			WC_Stripe_API::update_payment_method(
 				$payment_method_id,
 				[
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -2810,8 +2810,8 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 	/**
 	 * Copies the order's billing address onto a saved payment method in Stripe.
 	 *
-	 * Only the address is written, since AVS reads it from the payment method. Name, email, and
-	 * phone identify the original cardholder and must not be overwritten by a later checkout.
+	 * Only the address is written, since the Stripe Address Validation Service needs those details.
+	 *  Name, email, and phone should not be overwritten during checkout.
 	 *
 	 * @param string       $payment_method_id The payment method to update.
 	 * @param WC_Order|int $order             Order object or id.
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -2819,7 +2819,7 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 	public function update_saved_payment_method( $payment_method_id, $order ) {
 		$order = ! is_a( $order, 'WC_Order' ) ? wc_get_order( $order ) : $order;
 
-		if ( ! $order || ! $this->is_type_payment_method( $payment_method_id ) ) {
+		if ( ! $order instanceof WC_Order || ! $this->is_type_payment_method( $payment_method_id ) ) {
 			return;
 		}
 
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -2808,7 +2808,15 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 	}
 
 	/**
-	 * Update the saved payment method information with updated billing values.
+	 * Copies the order's billing address onto a saved payment method in Stripe.
+	 *
+	 * Only the address is written. The saved method's name, email, and phone identify the person
+	 * who originally saved it and must survive later checkouts: a checkout running under the wrong
+	 * account (a shared login, a session mix-up at a caching layer) would otherwise overwrite the
+	 * cardholder's identity with someone else's, and Stripe's copy is what fraud reviews rely on.
+	 * The address is the exception because Stripe's AVS check reads it from the payment method,
+	 * so a shopper who moved would be declined on postal-code mismatch unless it is refreshed
+	 * before the charge.
 	 *
 	 * @param string       $payment_method_id The payment method to update.
 	 * @param WC_Order|int $order             Order object or id.
@@ -2821,31 +2829,25 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 		}
 
 		try {
-			// Get the billing details from the order.
-			$billing_details = [
-				'address' => [
-					'city'        => $order->get_billing_city(),
-					'country'     => $order->get_billing_country(),
-					'line1'       => $order->get_billing_address_1(),
-					'line2'       => $order->get_billing_address_2(),
-					'postal_code' => $order->get_billing_postcode(),
-					'state'       => $order->get_billing_state(),
-				],
-				'email'   => $order->get_billing_email(),
-				'name'    => trim( $order->get_formatted_billing_full_name() ),
-				'phone'   => $order->get_billing_phone(),
+			$address = [
+				'city'        => $order->get_billing_city(),
+				'country'     => $order->get_billing_country(),
+				'line1'       => $order->get_billing_address_1(),
+				'line2'       => $order->get_billing_address_2(),
+				'postal_code' => $order->get_billing_postcode(),
+				'state'       => $order->get_billing_state(),
 			];
 
-			$billing_details = array_filter( $billing_details );
-			if ( empty( $billing_details ) ) {
+			if ( empty( array_filter( $address ) ) ) {
 				return;
 			}
 
-			// Update the billing details of the selected payment method in Stripe.
 			WC_Stripe_API::update_payment_method(
 				$payment_method_id,
 				[
-					'billing_details' => $billing_details,
+					'billing_details' => [
+						'address' => $address,
+					],
 				]
 			);
 
--- a/includes/abstracts/abstract-wc-stripe-payment-gateway.php
+++ b/includes/abstracts/abstract-wc-stripe-payment-gateway.php
@@ -2815,8 +2815,9 @@ abstract class WC_Stripe_Payment_Gateway extends WC_Payment_Gateway_CC {
 	 *
 	 * @param string       $payment_method_id The payment method to update.
 	 * @param WC_Order|int $order             Order object or id.
+	 * @param object|null  $payment_method    The PaymentMethod object when the caller already retrieved it, so the ownership check needs no extra request.
 	 */
-	public function update_saved_payment_method( $payment_method_id, $order ) {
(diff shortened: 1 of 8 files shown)


daledupreez (commented):
@wjrosa, I am not convinced this is the right approach here. I can think of valid use cases where some of these details may have been updated, like their phone number or email address. For cards, I am also not sure if the same number can remain in use after name changes. But I can see name changes being relevant for other saved payment methods like SEPA, ACH, Bacs, and ACSS where the underlying bank account remains stable. The same is likely true for other payment methods like Klarna and Cash App.

wjrosa:
Thanks, @daledupreez. I went with address-only because AVS only reads the address, and the Stripe Customer already gets the name, email, and phone from the order on every checkout, so contact changes still reach Stripe. For SEPA, Bacs, and ACSS, the name on the payment method is the mandate holder, so copying the order's billing name onto it looked risky to me. 

But you are right that there are valid updates, and this PR still rewrites the address on failed attempts. I think we should keep the address write before the intent (AVS needs it) and move name, email, and phone to after the intent succeeds. WDYT?

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2808,7 +2808,10 @@ protected function get_risk_outcome( $event_data ) {
 	}
 
 	/**
-	 * Update the saved payment method information with updated billing values.
+	 * Copies the order's billing address onto a saved payment method in Stripe.
+	 *
+	 * Only the address is written, since AVS reads it from the payment method. Name, email, and
+	 * phone identify the original cardholder and must not be overwritten by a later checkout.
```
daledupreez: It may be worth using "Address Validation Service" rather than the acronym.
```suggestion
	 * Only the address is written, since the Stripe Address Validation Service needs those details.
	 *  Name, email, and phone should not be overwritten during checkout.
```

tests/phpunit/class-wc-stripe-payment-gateway-test.php
```diff
@@ -1885,4 +1885,102 @@ public function provide_test_save_payment_method_checkbox(): array {
...
+		$this->assertSame(
+			[
+				'billing_details' => [
+					'address' => [
+						'city'        => 'WooCity',
+						'country'     => 'US',
+						'line1'       => 'WooAddress',
+						'line2'       => 'Unit 4',
+						'postal_code' => '12345',
+						'state'       => 'NY',
```
daledupreez, line 1959: Nit: I think we should explicitly set the initial fields for the order, as this approach has an implicit dependency on the way the test order was created.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
daledupreez: Not blocking: is it worth verifying that we have a `WC_Order` instance here?
```suggestion
		if ( ! $order instanceof WC_Order || ! $this->is_type_payment_method( $payment_method_id ) ) {
```

daledupreez (commented):
I haven't completed testing yet, but I did run an AI-assisted analysis of the change, and I don't think we should ship this in 11.0.0 for the following reasons:
 * `WC_Stripe_UPE_Payment_Gateway->is_order_associated_to_setup_intent()` currently requires the order email to match the email on the setup intent, which may break some flows due to this change.
 * For logged-in users, what fields do we normally use to populate the Stripe customer data? We may use a blend of data from WP_User and user metadata rather than checkout details, which makes it trickier to understand exactly what is going on and what should go on for saved tokens (as they're only available to logged-in users).
 
I am also concerned that we have other flows or expectations that depend on the full set of fields being extracted at this time, so changing this behaviour so close to the release feels a bit risky to me.

tests/phpunit/class-wc-stripe-payment-gateway-test.php
wjrosa: Makes sense. Applied in https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5910/commits/88e068a6d31c980d374dde41b89c5ae0db3a0f1d

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2808,44 +2808,41 @@ protected function get_risk_outcome( $event_data ) {
 	}
 
 	/**
-	 * Update the saved payment method information with updated billing values.
+	 * Copies the order's billing address onto a saved payment method in Stripe.
+	 *
+	 * Only the address is written, since the Stripe Address Validation Service needs those details.
+	 *  Name, email, and phone should not be overwritten during checkout.
```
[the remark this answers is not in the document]
wjrosa: Thanks for flagging this. I added a PaymentMethod-scoped lock around the billing address update and intent creation/confirmation sequence. This preserves the existing order-scoped lock while preventing different orders using the same saved PaymentMethod from racing and confirming against the wrong AVS address.

The lock uses ownership-safe release and stale-lock recovery. I also added coverage verifying that concurrent reuse cannot update the address or create an intent, that the competing lock is preserved, and that the order lock is released correctly.

Validated with PHPStan, PHPCS, and the saved-payment-method test suite (40 tests, 115 assertions).

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2808,44 +2808,41 @@ protected function get_risk_outcome( $event_data ) {
...
 	 *
 	 * @param string       $payment_method_id The payment method to update.
 	 * @param WC_Order|int $order             Order object or id.
 	 */
 	public function update_saved_payment_method( $payment_method_id, $order ) {
 		$order = ! is_a( $order, 'WC_Order' ) ? wc_get_order( $order ) : $order;
 
-		if ( ! $order || ! $this->is_type_payment_method( $payment_method_id ) ) {
+		if ( ! $order instanceof WC_Order || ! $this->is_type_payment_method( $payment_method_id ) ) {
```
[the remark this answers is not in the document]
wjrosa: Thanks for flagging this. I added a Stripe-side ownership check before updating the saved PaymentMethod. The update now proceeds only when the PaymentMethod’s customer matches the Stripe customer associated with the order; missing or mismatched ownership skips the mutation and logs a warning.

I also added regression coverage confirming that a mismatched PaymentMethod is retrieved for validation but never updated.

Validated with PHPUnit, PHPStan, PHPCS, and `git diff --check`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -17,6 +17,9 @@ class WC_Stripe_UPE_Payment_Gateway extends WC_Stripe_Payment_Gateway {
 
 	public const ID = 'stripe';
 
+	private const PAYMENT_METHOD_LOCK_OPTION_PREFIX = 'wc_stripe_payment_method_lock_';
+	private const PAYMENT_METHOD_LOCK_TTL           = 5 * MINUTE_IN_SECONDS;
```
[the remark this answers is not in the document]
wjrosa: Thanks for flagging the duplicated lock prefix. I removed the test-local constant and now derive the option name from `WC_Stripe_UPE_Payment_Gateway::PAYMENT_METHOD_LOCK_OPTION_PREFIX` via `WC_Stripe_Test_Helper::get_class_const_value()`.

The production constant remains private because the test helper already supports non-public constants, so we can keep the test aligned without expanding the subclass API. The lock success/contention tests, PHPCS, PHPStan, and `git diff --check` all pass.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1781,36 +1784,53 @@ private function process_payment_with_payment_method( int $order_id ) {
...
+			$payment_method_lock_owner = null;
 			if ( $is_using_saved_payment_method ) {
-				$this->update_saved_payment_method( $payment_method_id, $order );
+				$payment_method_lock_owner = $this->acquire_payment_method_lock( $payment_method_id );
+				if ( null === $payment_method_lock_owner ) {
+					throw new WC_Stripe_Exception(
+						'Saved payment method is already being used by another checkout.',
+						__( 'This saved payment method is already being used. Please try again.', 'woocommerce-gateway-stripe' )
+					);
+				}
```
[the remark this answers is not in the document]
wjrosa: Thanks for catching this. PaymentMethod lock contention now follows the existing order-lock contract: it releases the order lock and returns a retryable failure response instead of entering `handle_process_payment_error()`.

I also strengthened the contention regression test to verify that the order remains pending. The focused test, PHPCS, PHPStan, and `git diff --check` all pass.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -2308,24 +2400,14 @@ private function is_order_associated_to_payment_intent( int $order_id, string $i
...
 		}
 
-		if ( $order->get_billing_email() !== $intent->payment_method->billing_details->email ) {
-			return false;
-		}
+		$stored_intent_id = WC_Stripe_Order_Helper::get_instance()->get_stripe_setup_intent_id( $order );
 
-		return true;
+		// The persisted Stripe identifier binds the redirect to this order even when checkout details differ from the saved cardholder identity.
+		return is_string( $stored_intent_id ) && '' !== $stored_intent_id && $stored_intent_id === $intent_id;
```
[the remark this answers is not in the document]
wjrosa: Fixed in 1767d41c8. The subscription redirect branch now persists the exact SetupIntent ID before saving and redirecting, since `save_intent_to_order()` intentionally skips subscriptions. This lets the return handler validate the SetupIntent against the subscription without weakening the association check. PHPCS and PHPStan pass; the focused PHPUnit suite could not start locally because the WordPress test library is not installed.

tests/phpunit/payment-methods/class-wc-stripe-upe-payment-gateway-test.php
```diff
@@ -3492,6 +3537,45 @@ function ( $passed_order ) use ( $order ) {
...
+
+		$_POST['payment_method']           = 'stripe';
+		$_POST['wc-stripe-payment-method'] = 'pm_mock';
+
+		$order             = WC_Helper_Order::create_order();
+		$payment_method_id = $token->get_token();
+		$lock_option       = self::PAYMENT_METHOD_LOCK_OPTION_PREFIX . md5( $payment_method_id );
+		$existing_owner    = time() . ':other-order';
+
+		add_option( $lock_option, $existing_owner, '', false );
```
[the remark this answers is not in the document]
wjrosa: Added in 5a9437100. The new test seeds a PaymentMethod lock older than the configured TTL, verifies that the stale owner is replaced while checkout is processing, asserts successful intent creation, and confirms the replacement lock is released afterward. PHPCS and PHPStan pass; the focused PHPUnit suite cannot start in this checkout because the WordPress test library is not installed.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
```diff
@@ -2808,44 +2808,51 @@ protected function get_risk_outcome( $event_data ) {
...
+				'state'       => $order->get_billing_state(),
 			];
 
-			$billing_details = array_filter( $billing_details );
-			if ( empty( $billing_details ) ) {
+			if ( empty( array_filter( $address ) ) ) {
+				return;
+			}
+
+			$payment_method          = WC_Stripe_API::get_payment_method( $payment_method_id );
```
diegocurbelo: This adds another synchronous Stripe retrieval, but this checkout path already retrieved the same `PaymentMethod` before calling this method. Can we pass and reuse that object for the check. An optional argument would preserve compatibility for external callers while avoiding another request during checkout.

includes/compat/trait-wc-stripe-subscriptions.php
```diff
@@ -364,6 +364,9 @@ public function process_change_subscription_payment_with_deferred_intent( $subsc
 
 			// If the payment intent requires confirmation or action, redirect the customer to confirm the intent.
 			if ( in_array( $payment_intent->status, WC_Stripe_Intent_Status::REQUIRES_CONFIRMATION_OR_ACTION_STATUSES, true ) ) {
+				// Subscription intents are excluded from save_intent_to_order(), but redirect validation needs the exact SetupIntent ID.
+				WC_Stripe_Order_Helper::get_instance()->update_stripe_setup_intent_id( $subscription, $payment_intent->id );
```
diegocurbelo, line 368: Should we add a test for this new udpate/write?

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
```diff
@@ -1895,6 +1920,85 @@ private function process_payment_with_payment_method( int $order_id ) {
...
+		$wpdb->delete(
+			$wpdb->options,
+			[
+				'option_name'  => $option_name,
+				'option_value' => $lock_owner,
+			]
+		);
+		wp_cache_delete( $option_name, 'options' );
+	}
+
```
diegocurbelo, line 1932: Instead of duplicating `acquire_payment_method_lock()` and `release_payment_method_lock()` methods we should use `WC_Stripe_Option_Lock`

`WC_Stripe_Option_Lock` `acquire/release` already provides the same atomic insertion, timestamped owner token, TTL handling, conditional stale-lock reclamation, and ownership-safe release.

includes/abstracts/abstract-wc-stripe-payment-gateway.php
wjrosa: Done in a117cec2: `update_saved_payment_method()` takes an optional PaymentMethod object and checkout passes the one it already retrieved. Test added.

includes/compat/trait-wc-stripe-subscriptions.php
wjrosa: Added in a117cec2: `test_change_subscription_payment_persists_setup_intent_id_before_redirect`.

includes/payment-methods/class-wc-stripe-upe-payment-gateway.php
wjrosa: Agreed, done in a117cec2: both lock calls now go through `WC_Stripe_Option_Lock` and the duplicated methods are gone.

PR #5914, opened 2026-09-07
Improve Express Checkout required custom fields error logging

diegocurbelo:
Towards STRIPE-1434

This PR adds an ERROR log entry when missing required custom fields reject an express checkout request.
The entry includes the full validation messages and is recorded even when Stripe debug logging is disabled (error type), but does not log any submitted field values.

It also adds a new filter `wc_stripe_express_checkout_log_missing_required_fields` so merchants can disable this extra logging: `add_filter( 'wc_stripe_express_checkout_log_missing_required_fields', '__return_false' );`

## Testing instructions

1. Add this temporary snippet using a snippets plugin:
    ```php
    add_filter( 'woocommerce_checkout_fields', function ( $fields ) {
        $fields['billing']['billing_custom_reference'] = [
            'type'     => 'text',
            'label'    => 'Custom reference',
            'required' => true,
        ];

        return $fields;
    } );
    ```
2. Add a product to the cart, open classic checkout, and leave Custom reference empty.
3. Attempt payment using Apple Pay/Google Pay.
4. Confirm the existing required-field error appears. 
    In the browser’s Network panel, the Store API checkout request should return HTTP 400 with code `wc_stripe_express_checkout_missing_required_fields`
5. Open `WooCommerce → Status → Logs` and find the latest `woocommerce-gateway-stripe` log.
6. Confirm a new new ERROR log entry is present:
    Screenshot 2026-09-07 at 12 12 46

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
@@ -94,7 +94,15 @@ class WC_Stripe_Express_Checkout_Custom_Fields {
 				$required_field_errors[] = __( 'Please go to the checkout page, fill in the required fields, and complete your order from there.', 'woocommerce-gateway-stripe' );
 			}
 			$error_messages = implode( "\n", $required_field_errors );
-			WC_Stripe_Logger::error( 'Missing required custom fields in express checkout.', [ 'error_message' => $error_messages ] );
+			/**
+			 * Whether to log missing required custom fields during express checkout.
+			 *
+			 * @since 11.0.0
+			 * @param bool $should_log Return false to disable this error log. Default true.
+			 */
+			if ( false !== apply_filters( 'wc_stripe_express_checkout_log_missing_required_fields', true ) ) {
+				WC_Stripe_Logger::error( 'Missing required custom fields in express checkout.', [ 'error_message' => $error_messages ] );
+			}
 			throw new RouteException( 'wc_stripe_express_checkout_missing_required_fields', $error_messages, 400 );
 		}
 
--- a/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
@@ -102,7 +102,7 @@ class WC_Stripe_Express_Checkout_Custom_Fields {
 			 * @since 11.0.0
 			 * @param bool $should_log Return false to disable this error log. Default true.
 			 */
-			if ( false !== apply_filters( 'wc_stripe_express_checkout_log_missing_required_fields', true ) ) {
+			if ( apply_filters( 'wc_stripe_express_checkout_log_missing_required_fields', true ) ) {
 				WC_Stripe_Logger::error(
 					'Missing required custom fields in express checkout.',
 					[
--- a/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
+++ b/includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
@@ -106,8 +106,8 @@ class WC_Stripe_Express_Checkout_Custom_Fields {
 				WC_Stripe_Logger::error(
 					'Missing required custom fields in express checkout.',
 					[
-						'error_message'      => $error_messages,
 						'missing_field_keys' => $missing_field_keys,
+						'error_message'      => $error_messages,
 					]
 				);
 			}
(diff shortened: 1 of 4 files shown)


includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
```diff
@@ -94,6 +94,15 @@ public function process_custom_checkout_data( $order, $request ) {
 				$required_field_errors[] = __( 'Please go to the checkout page, fill in the required fields, and complete your order from there.', 'woocommerce-gateway-stripe' );
 			}
 			$error_messages = implode( "\n", $required_field_errors );
+			/**
+			 * Whether to log missing required custom fields during express checkout.
+			 *
+			 * @since 11.0.0
+			 * @param bool $should_log Return false to disable this error log. Default true.
+			 */
+			if ( false !== apply_filters( 'wc_stripe_express_checkout_log_missing_required_fields', true ) ) {
```
diegocurbelo: In other places we use this convention, but we could remove the `!==`

```suggestion
			if ( apply_filters( 'wc_stripe_express_checkout_log_missing_required_fields', true ) ) {
```

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
```diff
@@ -94,6 +94,15 @@ public function process_custom_checkout_data( $order, $request ) {
...
+				WC_Stripe_Logger::error( 'Missing required custom fields in express checkout.', [ 'error_message' => $error_messages ] );
```
daledupreez: Just to be explicit, is it worth storing the field keys/names in addition to the full message? It feels like that would be simpler to parse through compared to the full error message.

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
daledupreez: I don't have a strong opinion on this one.

daledupreez (commented):
The changes LGTM and test well. My only nit is that it would help to have the missing field keys logged in the context, not just the combined error message.

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
diegocurbelo: Added in [8e87636](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5914/commits/8e876367764576cddfffb4b98ce6ad39ee7ab419)

tests/phpunit/payment-methods/class-wc-stripe-express-checkout-custom-fields-test.php
```diff
@@ -442,7 +455,12 @@ static function ( $context ) use ( &$logged_error_message ) {
 			$message = $e->getMessage();
 			$this->assertSame( 'wc_stripe_express_checkout_missing_required_fields', $e->getErrorCode() );
 			$this->assertSame( 400, $e->getCode() );
-			$this->assertSame( $expects_logging ? $message : null, $logged_error_message );
+			$this->assertSame( $expects_logging ? $message : null, $logged_context['error_message'] ?? null );
+			$this->assertSame(
+				$expects_logging ? [ 'billing_custom_field1', 'shipping_custom_field' ] : null,
+				$logged_context['missing_field_keys'] ?? null
+			);
```
daledupreez, line 476: Not blocking, but is it worth testing that we get the expected values when one field is present and the other is missing?

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
```diff
@@ -101,7 +103,13 @@ public function process_custom_checkout_data( $order, $request ) {
 			 * @param bool $should_log Return false to disable this error log. Default true.
 			 */
 			if ( false !== apply_filters( 'wc_stripe_express_checkout_log_missing_required_fields', true ) ) {
-				WC_Stripe_Logger::error( 'Missing required custom fields in express checkout.', [ 'error_message' => $error_messages ] );
+				WC_Stripe_Logger::error(
+					'Missing required custom fields in express checkout.',
+					[
+						'error_message'      => $error_messages,
+						'missing_field_keys' => $missing_field_keys,
```
daledupreez: Nit: can we flip the order so we have the easy-to-read key values before the longer error message?
```suggestion
						'missing_field_keys' => $missing_field_keys,
						'error_message'      => $error_messages,
```

daledupreez (approved):
The changes LGTM based on inspection.

daledupreez:
Also, I think this should be `Towards STRIPE-1434`, as I am not 100% sure this is all we want to ship in response to that issue.

PR #5915, opened 2026-09-07
Add a link to the default Checkout page in the required custom field errors message

diegocurbelo:
Towards STRIPE-1434

This PR adds a link to the checkout page in the error message when there are missing required custom fields during Express Checkout on the product or cart page.

| Before | After |
|-|-|
|Screenshot 2026-09-07 at 12 12 20|Screenshot 2026-09-07 at 15 59 01|
|Screenshot 2026-09-07 at 21 52 07|Screenshot 2026-09-07 at 16 01 23|
|Screenshot 2026-09-07 at 21 51 00|Screenshot 2026-09-07 at 16 00 00|

_**NOTE:**_ Classic and Blocks notices preserve the link while stripping unsafe markup and attributes, and the required-field validation and the existing HTTP 400 response remain unchanged.

## Testing instructions

1. Add this temporary snippet using a snippets plugin:
    ```php
    add_filter( 'woocommerce_checkout_fields', function ( $fields ) {
        $fields['billing']['billing_custom_reference'] = [
            'type'     => 'text',
            'label'    => 'Custom reference',
            'required' => true,
        ];

        return $fields;
    } );
    ```
2. Go to a product page.
3. Click a Express Checkout button (Apple Pay, Google Pay, Link, etc)
4. Confirm the required-field error appears and the error message has a link to the default checkout
5. Add the product to the card and navigate to the Cart.
6. Click a Express Checkout button (Apple Pay, Google Pay, Link, etc)
7. Confirm the required-field error appears and the error message has a link to the default checkout

Repeat with classic and Blocks cart layouts.

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/client/blocks/express-checkout/hooks.js
+++ b/client/blocks/express-checkout/hooks.js
@@ -10,6 +10,7 @@ import {
 } from 'wcstripe/express-checkout/event-handler';
 import {
 	displayExpressCheckoutNotice,
+	formatExpressCheckoutNotice,
 	getExpressCheckoutButtonStyleSettings,
 	getExpressCheckoutData,
 	normalizeLineItems,
@@ -55,9 +56,12 @@ export const useExpressCheckout = ( {
 	}, [] );
 
 	const abortPayment = useCallback(
-		( onConfirmEvent, message ) => {
+		( onConfirmEvent, message, options = {} ) => {
 			// If we have a multiline message using newlines, replace them with <br>.
-			const formattedMessage = message.replace( /\n/g, '<br>' );
+			const formattedMessage =
+				options?.preserveLinks === true
+					? formatExpressCheckoutNotice( message, true )
+					: message.replace( /\n/g, '<br>' );
 			setExpressPaymentError( formattedMessage );
 
 			onAbortPaymentHandler( onConfirmEvent, message );
--- a/client/entrypoints/express-checkout/__tests__/index.test.js
+++ b/client/entrypoints/express-checkout/__tests__/index.test.js
@@ -515,38 +515,56 @@ describe( 'Express Checkout order failures', () => {
 	// The order-side abort used to skip paymentFailed(), leaving the approved wallet
 	// sheet open with nothing on screen. The third argument is the removed
 	// `isOrderError` opt-out: passing it must change nothing.
-	it( 'fails the wallet sheet and shows the message when the order errors', async () => {
-		const handlers = stubStripeButton();
-		loadEntrypoint();
+	it.each( [
+		[ 'plain message', 'Order creation error', true ],
+		[
+			'checkout link',
+			'Required field.\nPlease go to the <a href="https://example.com/checkout/">checkout page</a>.',
+			{ preserveLinks: true },
+		],
+	] )(
+		'fails the wallet sheet and shows the %s when the order errors',
+		async ( _name, message, options ) => {
+			const handlers = stubStripeButton();
+			loadEntrypoint();
 
-		// Resolve the mocks from the same module registry the entrypoint loaded from;
-		// `jest.resetModules()` hands each test its own copy.
-		// eslint-disable-next-line global-require
-		const jq = require( 'jquery' );
-		const {
-			onAbortPaymentHandler,
-			onConfirmHandler,
+			// Resolve the mocks from the same module registry the entrypoint loaded from;
+			// `jest.resetModules()` hands each test its own copy.
 			// eslint-disable-next-line global-require
-		} = require( 'wcstripe/express-checkout/event-handler' );
+			const jq = require( 'jquery' );
+			const {
+				onAbortPaymentHandler,
+				onConfirmHandler,
+				// eslint-disable-next-line global-require
+			} = require( 'wcstripe/express-checkout/event-handler' );
 
-		jq( document.body ).trigger( 'updated_checkout' );
+			jq( document.body ).trigger( 'updated_checkout' );
 
-		const event = { paymentFailed: jest.fn() };
-		await handlers.confirm( event );
+			const event = { paymentFailed: jest.fn() };
+			await handlers.confirm( event );
 
-		const { abortPayment } = onConfirmHandler.mock.calls[ 0 ][ 0 ];
-		abortPayment( event, 'Order creation error', true );
+			const { abortPayment } = onConfirmHandler.mock.calls[ 0 ][ 0 ];
+			abortPayment( event, message, options );
 
-		expect( event.paymentFailed ).toHaveBeenCalledWith( {
-			reason: 'fail',
-		} );
-		expect(
-			document.querySelector( '.woocommerce-error' ).textContent
-		).toBe( 'Order creation error' );
+			expect( event.paymentFailed ).toHaveBeenCalledWith( {
+				reason: 'fail',
+			} );
+			const notice = document.querySelector( '.woocommerce-error' );
+			expect(
+				notice.querySelector( 'a' )?.getAttribute( 'href' ) ?? null
+			).toBe(
+				options.preserveLinks ? 'https://example.com/checkout/' : null
+			);
+			expect( notice.textContent ).toBe(
+				options.preserveLinks
+					? 'Required field.Please go to the checkout page.'
+					: message
+			);
 
-		// The message has to be in front of the shopper before the sheet closes.
-		expect(
-			onAbortPaymentHandler.mock.invocationCallOrder[ 0 ]
-		).toBeLessThan( event.paymentFailed.mock.invocationCallOrder[ 0 ] );
-	} );
+			// The message has to be in front of the shopper before the sheet closes.
+			expect(
+				onAbortPaymentHandler.mock.invocationCallOrder[ 0 ]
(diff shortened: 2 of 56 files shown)


jaclync (commented):
💭 I was wondering, also for https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5915/changes#r3953627542, could we move the checkout link construction to the client side? Like:

Server side: keep the message plain text, and pass the "link to checkout" signal through `RouteException`'s 4th argument like `[ 'redirect_to_checkout' => true ]` when the request lacked the custom-data payload. It surfaces client-side as `exception.data.redirect_to_checkout`, right next to the code we already match on. (The error code alone isn't enough as it also fires on the checkout page itself.)

Client side: when the `redirect_to_checkout` flag is set, append a localized linkable message `please go to the [checkout page]` sentence, building the link from `getExpressCheckoutData( 'checkout' )?.url`, the same `wc_get_checkout_url()` value, already available on every page where the button renders from testing. That would let us drop the `formatExpressCheckoutNotice` sanitizer and the `preserveLinks` threading, and it avoids some side effects of parsing the whole message as HTML. For example, a field label like `Size <XL>` currently results in an error like `Size is a required field` from testing.

includes/payment-methods/class-wc-stripe-express-checkout-custom-fields.php
```diff
@@ -93,7 +93,12 @@ public function process_custom_checkout_data( $order, $request ) {
 			// without the custom-data payload came from a page without that form
 			// (e.g. product or cart), where the buyer has no way to fill the fields in.
 			if ( ! $this->request_has_custom_checkout_data( $request ) ) {
-				$required_field_errors[] = __( 'Please go to the checkout page, fill in the required fields, and complete your order from there.', 'woocommerce-gateway-stripe' );
+				$required_field_errors[] = sprintf(
+					/* translators: 1: opening checkout link, 2: closing checkout link */
+					__( 'Please go to the %1$scheckout page%2$s, fill in the required fields, and complete your order from there.', 'woocommerce-gateway-stripe' ),
+					'<a href="' . esc_url( wc_get_checkout_url() ) . '">',
```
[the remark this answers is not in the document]
diegocurbelo: The “sensitive query params” scenario needs a hypothetical filter injecting a secret into a URL that’s public by design.

diegocurbelo:
> 💭 I was wondering, also for https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5915/changes#r3953627542, could we move the checkout link construction to the client side? Like:
> 
> Server side: keep the message plain text, and pass the "link to checkout" signal through `RouteException`'s 4th argument like `[ 'redirect_to_checkout' => true ]` when the request lacked the custom-data payload. It surfaces client-side as `exception.data.redirect_to_checkout`, right next to the code we already match on. (The error code alone isn't enough as it also fires on the checkout page itself.)
> 
> Client side: when the `redirect_to_checkout` flag is set, append a localized linkable message `please go to the [checkout page]` sentence, building the link from `getExpressCheckoutData( 'checkout' )?.url`, the same `wc_get_checkout_url()` value, already available on every page where the button renders from testing. That would let us drop the `formatExpressCheckoutNotice` sanitizer and the `preserveLinks` threading, and it avoids some side effects of parsing the whole message as HTML. For example, a field label like `Size <XL>` currently results in an error like `Size is a required field` from testing.

Makes sense, looks cleaner this way. Changed in ab9e47b.

The sentence is localized server-side with `esc_url( wc_get_checkout_url() ))` rather than via `__()` on the client because classic has no `wp_set_script_translations`, so a client string would have been English-only on classic pages.

client/entrypoints/express-checkout/__tests__/index.test.js
```diff
@@ -547,38 +551,59 @@ describe( 'Express Checkout order failures', () => {
...
+				notice.querySelector( 'a' )?.getAttribute( 'href' ) ?? null
+			).toBe(
+				options.redirectToCheckout
+					? 'https://example.com/checkout/'
+					: null
+			);
+			// A tag-like label stays visible as text.
+			expect( notice.textContent ).toBe(
+				options.redirectToCheckout
+					? 'Size <XL> is a required field.Please go to the checkout page.'
```
daledupreez: Nit: Should we not require that there is a space between `.Please`?
```suggestion
					? 'Size <XL> is a required field. Please go to the checkout page.'
```

client/express-checkout/utils/index.js
```diff
@@ -404,17 +404,33 @@ export const getPaymentMethodTypesForExpressMethod = ( paymentMethodType ) => {
...
+	return sentence ? `${ noticeHtml }<br>${ sentence }` : noticeHtml;
+};
+
 /**
  * Display a notice on the checkout page (for Express Checkout Element).
  *
  * @param {string} message           The message to display.
  * @param {string} type              The type of notice.
  * @param {Array}  additionalClasses Additional classes to add to the notice.
+ * @param {Object} options           Set `redirectToCheckout` to append a link to the checkout page.
```
daledupreez: Nit: We are not actually _redirecting_ to checkout, we are only _linking_ to checkout. I think we should tweak the flag name to be more accurate, as I think we should make it clear that we are only adding a link.

jaclync: Agreed, maybe it can be renamed to `linkToCheckout`.

client/blocks/express-checkout/hooks.js
```diff
@@ -56,9 +57,12 @@ export const useExpressCheckout = ( {
...
 
 	const abortPayment = useCallback(
-		( onConfirmEvent, message ) => {
+		( onConfirmEvent, message, options = {} ) => {
 			// If we have a multiline message using newlines, replace them with <br>.
-			const formattedMessage = message.replace( /\n/g, '<br>' );
+			let formattedMessage = message.replace( /\n/g, '<br>' );
+			if ( options?.redirectToCheckout ) {
+				formattedMessage = appendCheckoutLink( formattedMessage );
+			}
```
jaclync, line 65: The message in the blocks checkout isn't escaped like in classic checkout, with example label like `Size <XL> <a href="https://…/test">free gift</a> is a required field`:

blocks | classic
-- | -- 
Image | Image

We probably want to escape in both paths?

includes/payment-methods/class-wc-stripe-express-checkout-element.php
```diff
@@ -237,6 +237,12 @@ public function javascript_params() {
 				'no_prepaid_card'  => __( 'Sorry, we\'re not accepting prepaid cards at this time.', 'woocommerce-gateway-stripe' ),
 				/* translators: Do not translate the [option] placeholder */
 				'unknown_shipping' => __( 'Unknown shipping option "[option]".', 'woocommerce-gateway-stripe' ),
+				'go_to_checkout'   => sprintf(
+					/* translators: 1: opening checkout link, 2: closing checkout link */
+					__( 'Please go to the %1$scheckout page%2$s, fill in the required fields, and complete your order from there.', 'woocommerce-gateway-stripe' ),
+					'<a href="' . esc_url( wc_get_checkout_url() ) . '">',
+					'</a>'
+				),
```
jaclync: Since the classic path injects it via `.html()` with no sanitizer, do you think it's worth:

1. Wrapping the sentence in `wp_kses` so a malicious translation can't cause issues here
2. Guarding the empty-URL case, as `wc_get_checkout_url()` can return `''` (unassigned checkout page, or a filter), which yields a self-linking `<a href="">`

client/express-checkout/utils/index.js
diegocurbelo: Renamed to linkToCheckout in [3f00db7](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5915/commits/3f00db744de043ece4ec8f28aa146b64e2f319d1)

client/blocks/express-checkout/hooks.js
diegocurbelo: Restored `develop` behavior (using `wp_strip_all_tags()`) in [3f00db7](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5915/commits/3f00db744de043ece4ec8f28aa146b64e2f319d1)

includes/payment-methods/class-wc-stripe-express-checkout-element.php
diegocurbelo: Fixed in [3f00db7](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5915/commits/3f00db744de043ece4ec8f28aa146b64e2f319d1)

client/entrypoints/express-checkout/__tests__/index.test.js
diegocurbelo: It reads wrong, but the assertion is right.
The notice is _"...is a required field.`<br>`Please go to the..."_, and `<br>` is ignored by `textContent`, changed it to `innerHTML` in [3f00db7](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/5915/commits/3f00db744de043ece4ec8f28aa146b64e2f319d1) so the `<br>` is visible in the test.

jaclync (approved):
Thanks for the updates, worked as expected (tags stripped) :shipit:

PR #5918, opened 2026-09-08
Handle one word shipping names in Express Checkouts

### LINKED ISSUE #5913: Express Checkout: one-word wallet name leaves shipping last_name empty, Store API rejects the order (400 woocommerce_rest_invalid_address)
### Describe the bug

When a wallet (Link, Google Pay, Apple Pay) returns a single-word name, the Express Checkout client splits it with `approximateFirstName()` / `approximateLastName()` in `client/express-checkout/utils/normalize.js`. For the billing address a `-` placeholder is used (`approximateLastName( name, '-' )`), for the shipping address there is no placeholder (`approximateLastName( name )`), so `shipping_address.last_name` is sent as an empty string.

WooCommerce requires the shipping last name by default. `OrderController::validate_order_before_payment()` rejects the Store API checkout with HTTP 400 `woocommerce_rest_invalid_address` ("Last name is required") before any payment is attempted. The wallet sheet offers no way to add a last name, so the shopper cannot complete an express payment at all. The only way out is the classic form.

Because 10.9.1's `abortPayment()` now calls `event.paymentFailed()` after the Store API error, the shopper additionally sees `IntegrationError: Unexpected call to paymentFailed()` in the console (same symptom as #3446).

### To Reproduce

1. WooCommerce 11.0.1, Stripe gateway 10.9.1, classic (shortcode) checkout, Express Checkout enabled, physical product in cart.
2. Pay with a Link account (or Google Pay / Apple Pay contact) whose name is a single word, e.g. `Cher`.
3. Confirm the wallet sheet.

Result: `POST /wp-json/wc/store/v1/checkout` returns 400 `woocommerce_rest_invalid_address`, the draft order stays `checkout-draft`, no PaymentIntent is created, error notice "There was a problem with the provided shipping address: Last name is required".

### Expected behavior

Either use the same `-` placeholder for the shipping name as for the billing name, or better: send the full wallet name so that WooCommerce can store it, and only split when the wallet provides separate given/family names.

### Environment

WooCommerce 11.0.1, woocommerce-gateway-stripe 10.9.1, WordPress 7.1, PHP 8.2, classic checkout with WPML multi-domain. Observed in production on 2026-09-01 (Firefox Mobile 154, Android 14); reproduced with a plain Store API request.

### Additional context

Related: #5827 (display items on zero-decimal currencies, same legacy classic-checkout ECE path). We work around this one with a small mu-plugin hooked to `woocommerce_store_api_checkout_update_order_from_request` that fills the empty shipping last name in express-checkout requests before validation.

diegocurbelo:
Fixes STRIPE-1435
Fixes #5913

Express Checkout now uses `-` as the shipping last name when a wallet provides a non-empty, one-word name. This prevents WooCommerce from rejecting the shipping address because the last name is empty. 

This solution follows the existing billing-address behavior, it reuses the billing convention while keeping the change narrowly scoped (using `name?.trim() ? '-' : ''` to keep existing behavior when shipping name is empty).

## Testing instructions

1. Add a physical product to the cart
2. Use Link, Apple Pay, or Google Pay with a one-word shipping name
3. Complete the payment
4. Confirm that checkout succeeds and the order contains:
    Shipping first name: `<one-word shipping name>`
    Shipping last name: `-`
5. Repeat with a multi-word name and confirm it is split as before

---

### Changelog entry

<details>

#### Comment 

</details>

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 11.0.0
+* Fix - Allow express checkout payments when a wallet provides a one-word shipping name
 * Add - Show a placement simulator on each Customize express checkouts tab that previews where the express checkout button would and wouldn't appear, with the reason
 * Tweak - Dim the button size hint on the Amazon Pay and Link customize tabs so it matches the Apple Pay/Google Pay tab
 * Add - Show the fuller sync-eligibility verdict in the Products list Agentic Commerce column and link the excluded-products view from the settings page
--- a/client/express-checkout/utils/normalize.js
+++ b/client/express-checkout/utils/normalize.js
@@ -398,11 +398,11 @@ const getBillingAddressData = ( event ) => {
  */
 const getShippingAddressData = ( event ) => {
 	const shipping = event?.shippingAddress ?? {};
-	const name = shipping?.name;
+	const name = shipping?.name?.trim() ?? '';
 
 	const data = {
 		first_name: approximateFirstName( name ),
-		last_name: approximateLastName( name, name?.trim() ? '-' : '' ),
+		last_name: approximateLastName( name, name === '' ? '' : '-' ),
 		company: shipping?.organization ?? '',
 		phone: getPhone( event ),
 		country: shipping?.address?.country ?? '',
--- a/client/express-checkout/utils/normalize.js
+++ b/client/express-checkout/utils/normalize.js
@@ -402,7 +402,7 @@ const getShippingAddressData = ( event ) => {
 
 	const data = {
 		first_name: approximateFirstName( name ),
-		last_name: approximateLastName( name ),
+		last_name: approximateLastName( name, name?.trim() ? '-' : '' ),
 		company: shipping?.organization ?? '',
 		phone: getPhone( event ),
 		country: shipping?.address?.country ?? '',
--- a/client/express-checkout/utils/__tests__/normalize.test.js
+++ b/client/express-checkout/utils/__tests__/normalize.test.js
@@ -566,6 +566,37 @@ describe( 'Express checkout normalization', () => {
 			} );
 		} );
 
+		test( 'should use the shipping last name fallback for a one-word shipping name', () => {
+			const oneWordShippingNameEvent = {
+				shippingAddress: {
+					name: 'Cher',
+				},
+			};
+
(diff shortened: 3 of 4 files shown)


client/express-checkout/utils/normalize.js
```diff
@@ -402,7 +402,7 @@ const getShippingAddressData = ( event ) => {
 
 	const data = {
 		first_name: approximateFirstName( name ),
-		last_name: approximateLastName( name ),
+		last_name: approximateLastName( name, name?.trim() ? '-' : '' ),
```
jaclync: If the `name` is ` Cher ` (leading space), it's returning `There was a problem with the provided shipping address: First name is required` error on the first name now.

How about trimming the `name` first before this const, and pass `-` as the default value for both first and last name fields? Plus unit testing coverage.

changelog.txt
```diff
@@ -1,6 +1,7 @@
 *** Changelog ***
 
 xxxx-xx-xx - version 11.0.0
+* Fix - Allow express checkout payments when a wallet provides a one-word shipping name
```
jaclync, line 4: Just checking that we are targeting 11.0.0 with "code freeze" yesterday?

jaclync (commented):
I think both first and last name can be handled the same way, WDYT?

client/express-checkout/utils/normalize.js
daledupreez: Agreed with @jaclync -- it feels like we should trim the incoming value first.
```suggestion
	const name = shipping?.name?.trim();

	const data = {
		first_name: approximateFirstName( name ),
		// Default the last name to `-` when a non-empty name has been supplied.
		last_name: approximateLastName( name,  name ? '-' : ''),
```

changelog.txt
daledupreez: We haven't merged the updates from `release/11.0.0` back into `trunk` or `develop` yet, so this will get cleaned up at that stage.

client/express-checkout/utils/normalize.js
daledupreez: I pushed this up in baa3a2db7d5e3e6a491a0da03bdf4297f2729715 and confirmed that it was working manually.

daledupreez (approved):
The changes LGTM with my minor tweak for the whitespace handling.

